PR #287, opened 2015-01-18
Feature instance launch time

chrishenry:
I have some code that looks at instance launch times, and looks for either the oldest or newest instance. This allows me to call `run_instances` with a sleep in between and ensure everything works correctly.

### DIFF
--- a/moto/ec2/models.py
+++ b/moto/ec2/models.py
@@ -299,6 +299,7 @@ class Instance(BotoInstance, TaggedEC2Resource):
         self.subnet_id = kwargs.get("subnet_id")
         self.key_name = kwargs.get("key_name")
         self.source_dest_check = "true"
+        self.launch_time = datetime.now().strftime('%y-%m-%dT%H:%M:%S%z')
 
         self.block_device_mapping = BlockDeviceMapping()
         self.block_device_mapping['/dev/sda1'] = BlockDeviceType(volume_id=random_volume_id())
--- a/moto/ec2/models.py
+++ b/moto/ec2/models.py
@@ -299,7 +299,7 @@ class Instance(BotoInstance, TaggedEC2Resource):
         self.subnet_id = kwargs.get("subnet_id")
         self.key_name = kwargs.get("key_name")
         self.source_dest_check = "true"
-        self.launch_time = datetime.now().strftime('%y-%m-%dT%H:%M:%S%z')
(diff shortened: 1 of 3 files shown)


spulec:
Thanks Chris!

tests/test_ec2/test_instances.py
```diff
@@ -50,6 +51,7 @@ def test_instance_launch_and_terminate():
     instances.should.have.length_of(1)
     instances[0].id.should.equal(instance.id)
     instances[0].state.should.equal('running')
+    datetime.datetime.strptime(instances[0].launch_time, '%Y-%m-%dT%H:%M:%S.%f')
```
hltbra: It looks like it is missing an expectation/assertion.

moto/ec2/models.py
```diff
@@ -299,6 +299,7 @@ def __init__(self, ec2_backend, image_id, user_data, security_groups, **kwargs):
         self.subnet_id = kwargs.get("subnet_id")
         self.key_name = kwargs.get("key_name")
         self.source_dest_check = "true"
+        self.launch_time = datetime.now().isoformat()
```
hltbra: Should UTC be used? I am not sure how other codes in moto handle dates.

tests/test_ec2/test_instances.py
chrishenry: `strptime` will return a `ValueError` if the format doesn't match the string given. See [here](https://docs.python.org/2/library/datetime.html#datetime.datetime.strptime). It seemed like that was sufficient to ensure that `launch_time` was the correct format.

moto/ec2/models.py
chrishenry: I don't think it matters much in this case, only that the time is in the correct format.

tests/test_ec2/test_instances.py
spulec: I've used freezegun to make a slightly stronger assertion here: https://github.com/spulec/moto/commit/b1c41820f70e8d9c8464f5fc380fb8a2fae1c1d5#diff-314909133cd9b2301075555c88d76574R56

moto/ec2/models.py
spulec: I think it is probably safer to go with `utcnow()` in case this runs on a machine that isn't set to UTC. Updated in master.

PR #291, opened 2015-01-29
Folder results should include the delimiter.

jbalogh:
This matches actual AWS results, the behavior in the [`if prefix` branch](https://github.com/jbalogh/moto/blob/c7bf6ffc91b378a14c653d6cbf7594f379e38504/moto/s3/models.py#L329), and the [docs](http://docs.aws.amazon.com/AmazonS3/latest/dev/ListingKeysHierarchy.html) which say:

> Amazon S3 groups these keys and return a single CommonPrefixes element with prefix value photos/ that is a substring from the beginning of these keys to the first occurrence of the specified delimiter.


spulec:
Great, thank you!

PR #293, opened 2015-02-02
Make RDS interface work with us-east-1 region

hltbra:
At the moment moto does not work with us-east-1 because boto endpoint does not specify that region: https://github.com/boto/boto/blob/e271ff09364ea18d9d8b6f4d63d6b0ac6cbc9b75/boto/endpoints.json#L285

This pull request fixes that issue by changing the RDS regex pattern to have the region as an optional value.


spulec:
Thanks!

PR #294, opened 2015-02-02
Add StorageType and Iops elements to DBInstance XML response

hltbra:
API Version 2014-10-31: http://docs.aws.amazon.com/AmazonRDS/latest/APIReference/API_DBInstance.html

Unfortunately boto does not support accessing StorageType through `storage_type` yet: https://github.com/boto/boto/issues/2923


spulec:
Thank you!

PR #297, opened 2015-02-10
Support for client supplied part order in multipart uploads

kouk:
The multipart code ignores the request body when completing a multipart upload and just assume the client wants all the parts it uploaded (and in that order). This PR adds support for using and validating the request body which contains the list of parts that should be included (possibly a subset of total uploaded parts).

I have also done some refactoring in the REST error code generation, inspired by the work done in the ec2 backend.

Last, I have a commit, not pushed yet, that patches the S3 backend during the multipart tests so that the minimum part size is smaller (e.g. 256 bytes instead of 5 MB). This speeds the tests up significantly. Do you think it's a good idea? If so, I'll push it here.

### DIFF
--- a/moto/core/exceptions.py
+++ b/moto/core/exceptions.py
@@ -1,0 +1,28 @@
+from werkzeug.exceptions import HTTPException
+from jinja2 import DictLoader, Environment
+
+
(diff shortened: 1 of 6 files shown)


spulec:
The additional commit would be great. Everything else looks good so I'm happy to merge after you push that.

spulec:
Thank you!

PR #307, opened 2015-02-17
Escaping tags individually for building the describe_tags template

dbrandenburg:
Calling "get_all_tags" of an ec2 connection fails, because the template doesn't get escaped. Iterating through the results to escape them individually. Issue happens when using tags including special characters.


spulec:
Thank you!

PR #309, opened 2015-02-17
Add support to create AWS::EC2::VPCPeeringConnection through the CloudFormation interface

hltbra:
It was missing an entry at `MODEL_MAP`.


spulec:
Thank you!

PR #310, opened 2015-02-20
allow running mock_s3 and mock_s3bucket_path one after the other

kouk:
Trying to get the full test suite running for me, and I've ran into some issues. One, fixed for me with this PR, is that the s3bucket_path tests all fail when run together with the s3 tests. The reason is that the URL patterns between them conflict in some cases (such as listing all buckets). If someone is using moto to test code that needs to handle both calling formats then they might have to do call reset on the backend manually. Instead I thought we could just call it when disabling the backend. I added a test case so you can see the problem.

Also, is there any reason there are no `__init__.py` files in the test suite? If I want to run a specific test I have to `touch` the needed files for `nose` to find the module.

### DIFF
--- a/moto/core/models.py
+++ b/moto/core/models.py
@@ -61,6 +61,7 @@ class MockAWS(object):
 
         if self.__class__.nested_count == 0:
             HTTPretty.disable()
+            HTTPretty.reset()
 
     def decorate_callable(self, func):
         def wrapper(*args, **kwargs):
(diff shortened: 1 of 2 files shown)


kouk:
The other problem I had was this: https://github.com/gabrielfalcao/HTTPretty/issues/221

spulec:
Awesome, thanks.

Nope, no reason for the missing `__init__.py`. Feel free to add in master.

kouk:
cool, will do.

PR #312, opened 2015-02-23
Support tag-key, tag-value and instance_type filters

aaltepet:
I added support for the instance filters listed above, including tests.


spulec:
Code looks good, but Travis seems to be giving some errors. If we can get those fixed, I'll be happy to merge.

PR #314, opened 2015-02-23
Update minimum support boto version.

jetlag11235:
boto 2.2.20 introduces kinesis.  alternatively, this requirement could be
relaxed by using conditional imports.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -4,7 +4,7 @@ from setuptools import setup, find_packages
 
 install_requires = [
     "Jinja2",
-    "boto",
+    "boto>=2.20.0",
     "flask",
     "httpretty>=0.6.1",
     "requests",


spulec:
I'm a little confused. What was the scenario that this became an issue with? Were you trying to setup the repo for development?

Ideally, I would like to allow people to use Moto even if they are using an older Boto version.

jetlag11235:
created an issue with more details.  apologies for not starting with that.
also updated this pull request with the correct minimum version based on
Kinesis.  feel free to close if you'd prefer to address via a less drastic
option (I was thinking conditional imports based on what features are
available in boto ... perhaps you'll have an even more clever solution!)

spulec:
Ah, got it. I forgot that we actually import boto in some places. This makes sense then.

Thank you!

PR #316, opened 2015-02-24
Support tag-key, tag-value and instance_type filters (second try)

aaltepet:
I added support for the instance filters listed above, including tests.

This time I've verified that travis-ci passes.  Sorry, I've never contributed to a project using travis-ci before. I needed to rewrite the failed test.  Thanks!

PR #317, opened 2015-02-24
Add support to tag filtering to Security Groups

hltbra:
I think that https://github.com/spulec/moto/pull/316 would be a great addition to this PR


spulec:
Thanks!

PR #319, opened 2015-02-25
Add support to AWS::EC2::SecurityGroupIngress creation

hltbra:
Can you please review, @spulec?

I noticed that CloudFormation resources do not get tags, this PR adds tags to security groups only. We need to review all models in `parsing.MODEL_MAP` to add proper tagging.


spulec:
Thanks. Opened a new issue and I will go back through all of the existing models later.

PR #322, opened 2015-02-27
Added in test for the boto IAM method: list_instance_profiles_for_role()

mikegrima:
Hello:

I have added in a test for the IAM method `list_intance_profiles_for_role()`

The most "jarring" change would be `moto/core/responses.py`, as I needed to pass in a parameter to one of my response methods -- Not sure of another way of doing this...

Thank You

### DIFF
--- a/moto/core/responses.py
+++ b/moto/core/responses.py
@@ -137,7 +137,10 @@ class BaseResponse(_TemplateEnvironmentMixin):
         if action in method_names:
             method = getattr(self, action)
             try:
-                response = method()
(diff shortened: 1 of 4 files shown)


mikegrima:
I am going to investigate the failed test, and see if I can get it fixed.  

Wondering if it is related to the test running in Python 3.x.

mikegrima:
Amended my commit with the fixes.  Re-opening...

mikegrima:
Another Python 2->3 bug...  Will fix... (I'm new to Python, so it may take me a few attempts :/)

PR #323, opened 2015-02-27
Added in test for the boto IAM method: list_instance_profiles_for_role()

mikegrima:
Hello:

I have added in a test for the IAM method `list_intance_profiles_for_role()`

The most "jarring" change would be `moto/core/responses.py`, as I needed to pass in a parameter to one of my response methods -- Not sure of another way of doing this...

Thank You

Note: This is my second attempt at the pull-request... Still learning how to use GitHub as well...

### DIFF
--- a/moto/iam/responses.py
+++ b/moto/iam/responses.py
@@ -87,7 +87,8 @@ class IamResponse(BaseResponse):
         template = self.response_template(LIST_INSTANCE_PROFILES_TEMPLATE)
         return template.render(instance_profiles=profiles)
 
-    def list_instance_profiles_for_role(self, role_name=None):
+    def list_instance_profiles_for_role(self):
+        role_name = self._get_param('RoleName')
         profiles = iam_backend.get_instance_profiles_for_role(role_name=role_name)
 
         template = self.response_template(LIST_INSTANCE_PROFILES_FOR_ROLE_TEMPLATE)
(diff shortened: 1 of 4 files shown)


mikegrima:
WooHoo! It passed! :)

moto/iam/responses.py
```diff
@@ -87,6 +87,12 @@ def list_instance_profiles(self):
         template = self.response_template(LIST_INSTANCE_PROFILES_TEMPLATE)
         return template.render(instance_profiles=profiles)
 
+    def list_instance_profiles_for_role(self, role_name=None):
+        profiles = iam_backend.get_instance_profiles_for_role(role_name=role_name)
```
spulec: Instead of special-casing `list_instance_profiles_for_role`, this method already has access to the querystring so we can  just do something like
```python
role_name = self.querystring['RoleName'][0])
```

or using a little helper we've added:

```python
role_name = self._get_param('RoleName')
```

Apologies that none of the internals are very well documented for this type of thing.

If we can get that cleaned up, I think everything else looks good.

mikegrima: Ahhh ok.  Will make the change and undo the change to `responses.py`.

Thank you!

mikegrima: Made the change on my end, and will publish it soon.

spulec:
Awesome. Thank you!

PR #335, opened 2015-03-22
Add support for ELB attributes.

mrucci:
Enable support for retrieving and modifying ELB attributes such as connection draining, cross zone load balancing, idle timeout and s3 access logs.


spulec:
This is great. Thank you!

PR #336, opened 2015-03-22
ELB can describe instances health.

mrucci:
Add support for describing instances health to ELB according to [API docs].(http://docs.aws.amazon.com/ElasticLoadBalancing/latest/APIReference/API_DescribeInstanceHealth.html)


spulec:
Merged with https://github.com/spulec/moto/commit/fcf5cdb9606b2de7e9d32021fac6916934d81131

Thank you!

PR #338, opened 2015-03-26
Update responses.py

Tinche:
Add a dummy NextToken to sns.list_platform_applications. Ideally the library would actually paginate, but this should be an alright change in the mean time.


spulec:
Thank you!

Issue #303, opened 2015-02-12
Mock S3 patching causes other uses of httplib2 to fail


wiltzius:
First of all, thanks for creating this library -- it's been extremely useful to us testing our app.

If a test is using the mock S3 instance and then tries to make an unrelated request with httplib2, that request fails with a BadStatusLine exception.

Here's a complete reduced example:

``` python
import unittest
import httplib2
from moto import mock_s3

class MyTestCase(unittest.TestCase):

  @classmethod
  def setUpClass(cls):
    mock_s3_obj = mock_s3()
    mock_s3_obj.start()

  def test_foo(self):
    http = httplib2.Http(timeout=60)
    http.follow_redirects = False
    http.request('https://google.com/', 'GET')
    return

if __name__ == '__main__':
  unittest.main()
```

the http.request line will raise a BadStatusLine exception deep in httplib somewhere.

I think ideally it should be possible to use Moto without it interfering with other services -- in our case, we want to mock S3 but use live Twilio for a certain integration test. Twilio's client, which uses httplib internally, is nonfunctional when s3 is mocked.

spulec:
This is a bit of a known issue. See https://github.com/spulec/moto/issues/32 and https://github.com/gabrielfalcao/HTTPretty/issues/65.

I think there has been some recent progress on the issue in HTTPretty, but I think there unfortunately is still more work to do. I haven't tried it, but you can try using the fixes suggested in https://github.com/gabrielfalcao/HTTPretty/pull/216.

wiltzius:
Ah yeah, I had not realized that #32 applied here.

For now we've worked around the issue by splitting our tests and not mocking S3 if we need any other live HTTP services. 

It looks like HTTPretty just merged that branch in, so maybe it's fixed with the latest. If I get a chance I'll give it a try -- are there any other known issues with newere HTTPretty versions? Moto is using 0.6.x which looks quite a bit older.

spulec:
I'm not aware of any issues.

Moto just requires `httpretty>=0.6.1` so you should be able to upgrade. Please let me know if you see anything unusual though.

porterjamesj:
just bumping to say that this still happens with the latest HTTPretty

argos83:
Hi, this is still happening with `httpretty 0.8.14` and `moto 0.4.30` (the latest versions atm). Any updates on this issue?

Thanks

spulec:
My understanding is that this is an issue with httpretty. Are you seeing otherwise?

[...]

PR #339, opened 2015-04-03
allow starting without reseting

porterjamesj:
The ultimate motivation for this is working around HTTPretty bugs that cause it to interfere with other network calls when moto is enabled, specifically [this](https://github.com/gabrielfalcao/HTTPretty/issues/182), [this](https://github.com/gabrielfalcao/HTTPretty/issues/65), etc. Relatedly https://github.com/spulec/moto/issues/303.

I'm using moto in some Big Fat Integration Tests™, and HTTPretty was causing other http calls in my tests to fail. Obviously the right thing to do is fix HTTPretty, but in the meantime I figured I would just `mock.patch` out my calls to functions that do S3 interaction so that they turn moto on and off before and after running respectively.

Unfortunately this doesn't work, since the backends get reset every time a `MockAWS` is `start`ed. For now I'm monkeypatching around this, but it would be nice to be able to do it explicitly.

I basically need the following to work:

```python

s3 = mock_s3()

s3.start()

conn = boto.connect_s3()
conn.create_bucket("test_bucket")

s3.stop()

# . . . other network calls that I need HTTPretty not to interfere with

s3.start(reset=False)

# . . . boto calls that use test_bucket

s3.stop()
```

I'm not attached to this interface, it was just the quickest way to get what I wanted. If you want me to change things / refactor I'm happy to do so.

### DIFF
--- a/moto/core/models.py
+++ b/moto/core/models.py
@@ -29,10 +29,11 @@ class MockAWS(object):
     def __exit__(self, *args):
         self.stop()
 
-    def start(self):
+    def start(self, reset=True):
         self.__class__.nested_count += 1
-        for backend in self.backends.values():
-            backend.reset()
+        if reset:
+            for backend in self.backends.values():
+                backend.reset()
 
         if not HTTPretty.is_enabled():
             HTTPretty.enable()


spulec:
Thanks, I think this is a good compromise for now.

PR #344, opened 2015-04-23
Adding support for comments on hosted zones.

pcorliss:
Paired with @kpdonn

Route53 Hosted Zones can be created with a comment in boto. This is currently the only way to differentiate two hosted zones via boto when retrieving a list. Wanted to get support in this for moto.


spulec:
Great, thank you!

PR #345, opened 2015-04-28
Use dummy date instead of an invalid date

chicofranchico:
It would really help and avoid problems if the returning dates are actually valid ones. I used the same date as found in other response bodies.


spulec:
Thank you!

PR #351, opened 2015-05-20
Fix bug creating elb with a listener certificate

kpdonn:
Added test to cover that case.

Paired with @jszwedko


spulec:
Great, thank you!

PR #354, opened 2015-05-29
threadsafe fix

jgelens:
Dispatching is not threadsafe. We had request bodies overwriting each other with a high number of simultaneous requests. This pull requests fixes all cases we've found until now.


spulec:
Happy to merge this, but I should note that HTTPretty is not thread-safe in general. See https://github.com/gabrielfalcao/HTTPretty/issues/209

PR #361, opened 2015-06-12
Bad XML from S3_BUCKET_VERSIONING and S3_BUCKET_GET_VERSIONING

mruwnik:
The '<?xml version="1.0" encoding="UTF-8"?>' bit must be on the first line, but the templates used for versioning were generating strings that started with a newline


spulec:
Thank you!

PR #364, opened 2015-06-25
Fix unhandled KeyError

ivanhernandezplangrid:
When S3 delete method is called on a key that is not present a KeyError is raised causing the moto server to freeze until timeout.


spulec:
Thanks, I decided a slightly different fix to ensure we return a 404. I appreciate you opening the issue!

ivanhernandezplangrid:
Thanks a lot!

PR #365, opened 2015-06-28
Add support for alias records in Route53.

mrucci:
[API reference](http://docs.aws.amazon.com/Route53/latest/APIReference/CreateAliasRRSAPI.html)


spulec:
Thank you!

PR #369, opened 2015-06-30
Improves range request support.

MurphyMarkW:
Just a couple small modifications to range request handling.

The current implementation of range request handling raises errors on byte-range-spec range requests with explicit last-byte-pos that are greater than or equal to the length of the underlying data. The expected behavior of such a request is that it will be interpreted as a request for all remaining bytes after and including the first-byte-pos, similar to a byte-range-spec range request where no last-byte-pos was specified.

Likewise, suffix-byte-range-spec range requests with last-byte-pos that are greater than the length of the underlying data raise errors. The expected behavior of such a request is that it will be interpreted as a request for the entire length of the data.

These behaviors are also observable using AWS S3. You can find this described in RFC 7233 section 2.1, and the older RFC 2616 section 14.3.1 that is referenced in the AWS S3 specs.

This also adds a couple tests for different range requests, checks for some off-by-one errors and groups the individual tests by what they're looking for.

Thanks!  o7

### DIFF
--- a/moto/s3/responses.py
+++ b/moto/s3/responses.py
@@ -252,7 +252,7 @@ class ResponseObject(_TemplateEnvironmentMixin):
         toint = lambda i: int(i) if i else None
         begin, end = map(toint, rspec.split('-'))
         if begin is not None:  # byte range
-            end = last if end is None else end
+            end = last if end is None else min(end, last)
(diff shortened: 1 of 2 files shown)


spulec:
This is great. Thank you!

PR #371, opened 2015-07-07
Adds the ConditionalCheckFailedException to put_item

apjaffe:
If the Item‘s original data is inconsistent with what's in DynamoDB,
the request should fail (unless overwrite is set to True).

http://boto.readthedocs.org/en/latest/ref/dynamodb2.html#boto.dynamodb2.table.Table.put_item

### DIFF
--- a/moto/dynamodb2/models.py
+++ b/moto/dynamodb2/models.py
@@ -191,7 +191,7 @@ class Table(object):
                     keys.append(key['AttributeName'])
         return keys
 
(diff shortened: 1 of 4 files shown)


spulec:
Can you add a couple tests for this and then I would be happy to merge?

apjaffe:
Okay, will do

spulec:
Great work! Thank you!

PR #373, opened 2015-07-08
feature emr: set_termination_protection  and instance_groups param in run_jobflow constructor

evgenyshulman:
feature emr: support for instance_groups param in run_jobflow ( in case user has provided instance_groups argument)
feature emr: set_termination_protection implementation


spulec:
This is great. Thank you!

Issue #347, opened 2015-05-15
boto raises BadStatusLine exception when using moto.emr and boto.emr.connect_to_region()


achiku:
Hi! First of all thank you so much for open sourcing this wonderful library!

I have a question related to ```moto.emr``` module. Is it possible to use ```boto.emr.connect_to_region()``` to get EmrConnection while using moto? 

What I wanted to do
-------------------
```boto.emr.connect_to_region()``` is used in my code, and I wanted to mock EMR in my test code.

Errors
------

When I run the test above, it raises BadStatusLine exception.

```
/Users/achiku/.virtualenvs/dashboard/src/boto/boto/emr/connection.py:580: in run_jobflow
    'RunJobFlow', params, RunJobFlowResponse, verb='POST')
/Users/achiku/.virtualenvs/dashboard/src/boto/boto/connection.py:1192: in get_object
    response = self.make_request(action, params, path, verb)
/Users/achiku/.virtualenvs/dashboard/src/boto/boto/connection.py:1116: in make_request
    return self._mexe(http_request)
/Users/achiku/.virtualenvs/dashboard/src/boto/boto/connection.py:1030: in _mexe
    raise ex
E   BadStatusLine: ''
```

Error Reproduction
------------------

moto EMR test cases are using ```boto.connect_emr()```, and when I clone moto repo, make init, and run the tests(make test), all tests successfully passed. Then I silightly modify ```tests/test_emr/test_emr.py``` and run the tests again to reproduce the error above.

```
$ python --version
Python 2.7.9
$ pip freeze
You are using pip version 6.0.8, however version 6.1.1 is available.
You should consider upgrading via the 'pip install --upgrade pip' command.
boto==2.38.0
coverage==3.7.1
Flask==0.10.1
freezegun==0.3.3
gnureadline==6.3.3
httpretty==0.8.9
ipython==3.1.0
itsdangerous==0.24
Jinja2==2.7.3
MarkupSafe==0.23
mock==1.0.1
-e git+git@github.com:spulec/moto.git@be5f0414160ca254d9c6df5e13917d52969c269b#egg=moto-master
nose==1.3.6
python-dateutil==2.4.2
requests==2.7.0
six==1.9.0
sure==1.2.3
Werkzeug==0.10.4
xmltodict==0.9.2
```

```
diff --git a/tests/test_emr/test_emr.py b/tests/test_emr/test_emr.py
index 9157a82..d6065b9 100644
--- a/tests/test_emr/test_emr.py
[...]

PR #375, opened 2015-07-12
Make EMR multi region

achiku:
@spulec 

This pr makes EMR mock support multi region.
https://github.com/spulec/moto/issues/347

Note:
Since this issue https://github.com/boto/boto/pull/3186 can't be resolved, moto's EMR multi region can only support ```us-east-1``` and ```eu-west-1```. These two regions are only regions which have compatible endpoints (```elasticmapreduce.eu-west-1.amazonaws.com``` and ```elasticmapreduce.us-east-1.amazonaws.com```) with moto's ```emr/urls.py```. However, we can use other regions with moto when boto3 is used instead of boto.

### DIFF
--- a/moto/emr/__init__.py
+++ b/moto/emr/__init__.py
@@ -1,3 +1,12 @@
 from __future__ import unicode_literals
-from .models import emr_backend
-mock_emr = emr_backend.decorator
+from .models import emr_backends
+from ..core.models import MockAWS
+
+emr_backend = emr_backends['us-east-1']
+
+
+def mock_emr(func=None):
+    if func:
+        return MockAWS(emr_backends)(func)
+    else:
+        return MockAWS(emr_backends)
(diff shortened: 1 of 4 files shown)


achiku:
Since this issue https://github.com/boto/boto/pull/3186 can't be resolved, moto's EMR multi region can only support ```us-east-1``` and ```eu-west-1```. These two regions are only regions which have compatible endpoints (```elasticmapreduce.eu-west-1.amazonaws.com``` and ```elasticmapreduce.us-east-1.amazonaws.com```) with moto's ```emr/urls.py```. However, we can use other regions with moto when boto3 is used instead of boto.

spulec:
Great work! Thank you for your contribution.

achiku:
No problem at all. Thank you for quick response!

PR #379, opened 2015-07-16
implement load balancer policies mocking

devmage:
Hey @spulec ! I'm doing some test work involving setting load balancer policies. My colleague and I found that these weren't implemented in moto, so we tried to accomplish that here.

For reference, the methods we sought to implement were:

* `create_app_cookie_stickiness_policy` [doc](http://docs.aws.amazon.com/ElasticLoadBalancing/latest/APIReference/API_CreateAppCookieStickinessPolicy.html)
* `create_cookie_stickiness_policy` [doc](http://docs.aws.amazon.com/ElasticLoadBalancing/latest/APIReference/API_CreateLBCookieStickinessPolicy.html)
* **`create_lb_policy`** [doc](http://docs.aws.amazon.com/ElasticLoadBalancing/latest/APIReference/API_CreateLoadBalancerPolicy.html)
* `set_policies_of_listener` [doc](http://docs.aws.amazon.com/ElasticLoadBalancing/latest/APIReference/API_SetLoadBalancerPoliciesOfListener.html)
* **`set_policies_of_backend_server`** [doc](http://docs.aws.amazon.com/ElasticLoadBalancing/latest/APIReference/API_SetLoadBalancerPoliciesForBackendServer.html)

The two methods in bold are those we actually needed for our tests. We felt it made sense to implement the remaining three, since they appeared to be tightly related. All five are also described in [boto's elb documentation](http://boto.readthedocs.org/en/latest/ref/elb.html).

Some use-cases for these methods include [configuring sticky sessions](http://docs.aws.amazon.com/ElasticLoadBalancing/latest/DeveloperGuide/elb-sticky-sessions.html), [enabling proxy protocol support](http://docs.aws.amazon.com/ElasticLoadBalancing/latest/DeveloperGuide/enable-proxy-protocol.html), and [setting SSL security policies](http://docs.aws.amazon.com/ElasticLoadBalancing/latest/DeveloperGuide/elb-create-https-ssl-load-balancer.html).

We were not sure as to the best way to mock the backend server description portion, so that may not reflect AWS reality very well. There also are arguments to be made for implementing `delete_load_balancer_policy` and `describe_load_balancer_policies` as part of this work, but we felt these methods were a good start for supporting policies (and they satisfy our needs for the moment).

### DIFF
--- a/moto/elb/responses.py
+++ b/moto/elb/responses.py
@@ -220,7 +220,7 @@ class ELBResponse(BaseResponse):
         policy.policy_name = policy_name
         cookie_expirations = [value[0] for key, value in self.querystring.items() if "CookieExpirationPeriod" in key]
         if cookie_expirations:
-            policy.cookie_expiration_period = long(cookie_expirations[0])
+            policy.cookie_expiration_period = int(cookie_expirations[0])
         else:
             policy.cookie_expiration_period = None
 
--- a/moto/elb/responses.py
+++ b/moto/elb/responses.py
@@ -5,6 +5,12 @@ from boto.ec2.elb.attributes import (
     AccessLogAttribute,
     CrossZoneLoadBalancingAttribute,
 )
+from boto.ec2.elb.policies import (
+    Policies,
+    AppCookieStickinessPolicy,
+    LBCookieStickinessPolicy,
+    OtherPolicy,
+)
 
 from moto.core.responses import BaseResponse
 from .models import elb_backends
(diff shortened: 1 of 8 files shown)


devmage:
Looking forward to feedback! Thanks for your work here @spulec .

tests/test_elb/test_elb.py
```diff
@@ -315,6 +321,107 @@ def test_connection_settings_attribute():
...
+    # with a given listener
+
+    # in a real flow, it is necessary first to create a policy,
+    # then to set that policy to the listener
+    lb.create_cookie_stickiness_policy(None, policy_name)
+    lb.set_policies_of_listener(listener_port, [policy_name])
+
+    lb = conn.get_all_load_balancers()[0]
+    print lb.listeners
+    print lb.backends
```
devmage: Could have sworn I got rid of these!

moto/elb/responses.py
```diff
@@ -261,7 +385,22 @@ def describe_instance_health(self):
...
+                {% if backend.policy_names %}
+                    <PolicyNames>
+                        {% for policy in backend.policy_names %}
+                            <member>{{ policy }}</member>
+                        {% endfor %}
+                    </PolicyNames>
+                    {% endif %}
+            </member>
+          {% endfor %}
+          </BackendServerDescriptions>
```
devmage: This is the portion I'm talking about, when I say that I'm not sure that I've correctly modeled the backend server description objects.

moto/elb/responses.py
```diff
@@ -177,6 +183,101 @@ def parse_attribute(attribute_name):
...
+                try:
+                    policy = self.querystring['PolicyNames.member.{0}'.format(policy_index)][0]
+                except KeyError:
+                    break
+
+                policy_index += 1
+                policies.append(str(policy))
+
+            self.elb_backend.set_load_balancer_policies_of_listener(load_balancer_name, load_balancer_port, policies)
+        # else: explode?
```
devmage: Silently failing didn't quite seem like the best approach for not finding a load balancer, but I wasn't sure as to what would be better.

devmage:
Sorry about the merge commits from `master` showing up in here. I can cherrypick my work onto a new branch and reopen the PR if you'd prefer.

moto/elb/responses.py
spulec: Based on the docs, it looks like Boto will return a 400 status code. Not a big deal, we can add that later.

spulec:
This is great. Thanks for you work!

PR #382, opened 2015-07-23
CloudWatch DescribeAlarm filters.

IlyaSukhanov:
Adds support for filtering by:
action_prefix
alarm_name_prefix
alarm_names

And throw NotImplementedError when filtering by:
state_value


spulec:
This is great! Thanks, will push a new version in a minute

PR #383, opened 2015-07-23
S3: Bucket policy

spg:


laurrentt:
+1 I could use it in a dozen of cases

spulec:
This is great. Thank you!

PR #384, opened 2015-07-26
volmods

rocky4570:
Moto is cool! Added changes to show attached volumes on an instance (handling ghost volume as well), search volumes/snapshots by their IDs and show placement and start/create times based on API usage.

### DIFF
--- a/moto/ec2/models.py
+++ b/moto/ec2/models.py
@@ -3,6 +3,7 @@ from __future__ import unicode_literals
 from collections import defaultdict
 import copy
 from datetime import datetime
+
 import itertools
 import re
 
@@ -96,6 +97,9 @@ from .utils import (
     is_tag_filter,
 )
 
+
+def utc_date_and_time():
+    return datetime.utcnow().strftime('%Y-%m-%dT%H:%M:%SZ')
 
 def validate_resource_ids(resource_ids):
     for resource_id in resource_ids:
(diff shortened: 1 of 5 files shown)


moto/ec2/models.py
```diff
@@ -1444,6 +1457,13 @@ def attach_volume(self, volume_id, instance_id, device_path):
             return False
 
         volume.attachment = VolumeAttachment(volume, instance, device_path)
+        # Modify instance to capture mount of block device.
+        bdt = BlockDeviceType(volume_id=volume_id, status=volume.status, size=volume.size,
+                              attach_time=utc_date_and_time())
+        try:
+            instance.block_device_mapping[device_path] = bdt
+        except:
```
spulec: When does the exception happen? Can we specify the Exception type?

spulec:
This is great! I just had one comment/question and then happy to get this merged.

moto/ec2/models.py
rocky4570: The setdefault doesn't clear out the /dev/sda1 key which is introduced on a launch instance in moto (i call it a ghost volume - as it doesn't exist when describe_all_volumes() but shows up when looking at block_device_mapping of the instance - initially was going to deleted but decided to keep for backward compatability). So I just used a try/except to get around it, but could have dealt with it specifically for /dev/sda1 I guess.

spulec:
Thanks!

PR #390, opened 2015-08-03
Added server backends with tests for IAM and SNS services.

jotes:
Hi, I've added two new backends and started a little bit refactoring in server tests for them. I'll appreciate any constructive comments :)


spulec:
Awesome, thanks!

PR #394, opened 2015-08-04
Fixed regex

collinstocks:
The regex in url_bases in moto.s3.urls was too broad. Bucket names cannot be zero characters, and the dot separating the bucket name from the rest of the domain name is not optional. This overly-broad regex caused URLs that should have gone to the moto.s3bucket_path backend go to the moto.s3 backend instead.

### DIFF
--- a/moto/s3/urls.py
+++ b/moto/s3/urls.py
@@ -2,7 +2,7 @@ from __future__ import unicode_literals
 from .responses import S3ResponseInstance
 
 url_bases = [
-    "https?://(?P<bucket_name>[a-zA-Z0-9\-_.]*)\.?s3(.*).amazonaws.com"
+    "https?://(?P<bucket_name>[a-zA-Z0-9\-_.]+)\.s3(.*).amazonaws.com"
 ]
 
 url_paths = {


collinstocks:
It looks as though `test_s3.test_get_all_buckets` is failing. I don't think I caused this error.

jotes:
It looks like your change causes get_all_buckets to fail, because boto requests data about buckets from endpoint with following address: http://s3(.*).amazonaws.com .

collinstocks:
Right, and the s3bucket_path backend should support this. I think this reveals an existing bug in s3bucket_path.

collinstocks:
That said, I made the same regex change in my locally installed copy of the release version of moto (installed using pip), and I am able to list all buckets, so I wonder if this is a problem with the test, or was a recently-introduced bug.

collinstocks:
Hmm, I think I know why the tests are failing. The s3bucket_path backend keeps a different state than the s3 backend. Is this behavior intended? If I make a call to S3, no matter which calling format I use, I read/write the same set of buckets on the same account....

collinstocks:
I'll be happy to make the appropriate changes if this is not the intended behavior. I'd really like to use this library.

spulec:
I agree that the s3bucket_path backend and the normal s3 backend should use the same behavior. If that isn't the case, happy to accept a PR to fix

collinstocks:
@spulec Which approach would you prefer:
1) Change the normal s3 backend to correctly handle the URLs that s3bucket_path currently handles, and either remove s3bucket_path, or make it an alias for the normal s3 backend.
2) Change the normal s3 backend and s3bucket_path to use the same S3 namespace.
3) Something else?

collinstocks:
Also, is there an API reference somewhere so I can determine which parts of the s3 backend are internal and which need to be preserved as part of the API?

spulec:
We should change them to both use the same backend.

All of the internal API should be considered private.

collinstocks:
Alright, I'll spend some time on this when I get a chance.

JackDanger:
@collinstocks This still seems pretty useful. I'm closing this PR due to its age. Please reopen when you get a chance!

PR #399, opened 2015-08-12
Use extras_require to build a wheel usable on Python 2.6

### LINKED ISSUE #340: Python 2.6 issues with wheels and dependencies
My Travis tests failed on Python 2.6: piskvorky/smart_open#15 .

After some digging around it appears this is because of `moto`. Moto apparently depends on some `ordereddict` package, but that package is not installed (nor mentioned anywhere in the docs, AFAICS).

Do you think you could make `ordereddict` a dependency for moto, when installing on Python 2.6?

In other words, after I successfully run `pip install moto`, I'd expect moto to work, even on Python 2.6.

And thanks for the great package!

rouge8:
By using PEP 426 markers with `extras_require` in `setup.py`, wheel can parse the marker and install ordereddict only on Python 2.6.

ref: https://wheel.readthedocs.org/en/latest/#defining-conditional-dependencies

Fixes #340.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -13,11 +13,10 @@ install_requires = [
     "werkzeug",
 ]
 
-import sys
-
-if sys.version_info < (2, 7):
-    # No buildint OrderedDict before 2.7
-    install_requires.append('ordereddict')
+extras_require = {
+    # No builtin OrderedDict before 2.7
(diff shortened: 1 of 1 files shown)


spulec:
Perfect! Thank you.

PR #400, opened 2015-08-13
Began Documentation

ZuluPro:
Began a documentation inspired from boto's one.
I don't think it is ready for ReadTheDocs, but a basis for further work and encourage contributors.


spulec:
Awesome, thanks!

PR #401, opened 2015-08-13
Add more boto3 s3 tests

adepue:
I was debugging a mis configuration on my end and ended up writing these tests for head_object in boto3.  Figured I would offer them back.


spulec:
Excellent, thank you!

PR #404, opened 2015-08-20
Implement XML responses for SNS (for Boto3)

pior:
Hi @spulec,

I partially implemented the XML responses for the SNS backend (the topic related responses).
There is a new Boto3 specific test: https://github.com/ludia/moto/blob/feature/implement-sns-xml-response/tests/test_sns/test_topics_boto3.py 

If you think this is the right direction, I could probably finish this today.

### DIFF
--- a/moto/sns/models.py
+++ b/moto/sns/models.py
@@ -2,6 +2,7 @@ from __future__ import unicode_literals
 
 import datetime
 import uuid
+import json
 
 import boto.sns
 import requests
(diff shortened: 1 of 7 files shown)


moto/sns/models.py
```diff
@@ -284,9 +285,9 @@ def set_endpoint_attributes(self, arn, attributes):
             }
         }
     }]
-}
+})
```
pior: Even in the JSON representation, the *_policy are represented as string (not as nested structure)

moto/sns/responses.py
```diff
@@ -335,10 +411,101 @@ def set_endpoint_attributes(self):
...
+                    }
                 }
-            }
-        })
+            })
+
+        template = self.response_template(_TEMPLATE)
+        return template.render()
+
+_TEMPLATE = ""
```
pior: This will go away.

spulec:
This is great!

I'm happy to merge now, but can wait if you want to keep working on the rest on this branch.

pior:
@spulec Great! 

I'll just finish the complete implementation of SNS.

pior:
@spulec Done!

spulec:
Awesome! Thanks for all of your work.

Issue #403, opened 2015-08-18
Support ELB tags


pitluga:
I'd like to be able to add tags to our ELBs. There is a boto issue that's been open for about a year adding tags https://github.com/boto/boto/issues/2549, so it doesn't seem like it will be added there. Boto3 has this feature, but only supports API versions `2014-09-01`, `2014-10-01`, `2015-03-01`, and `2015-04-15`. Moto's responses use API version `2012-06-01` so boto3 cannot parse moto's responses.

Any guidance on how to add this feature? Should we upgrade the ELB responses to the most recent version provided it works with boto 2.36? I didn't want to do a bunch of work that wouldn't get merged in.

I was thinking of doing two PRs, one upgrading moto to support boto3 for ELBs and a second adding ELB tags. Sound like a plan?

spulec:
Sounds perfect

spulec:
Fixed with #408

PR #408, opened 2015-08-24
Make all existing ELB endpoints compatible with boto3

pitluga:
Turns out I was wrong in #403, the issue was not API versions, but some of the existing responses did not have the `<FooBarResponse>` wrappers and only had `<FooBarResult>` which did not work with boto3. 

This commit adds all the appropriate wrappers and tests for the endpoints that changed. I'm happy to go back and fill in another set of boto3 compatibility for the rest of the endpoints. Let me know.


spulec:
Perfect, thank you!

PR #410, opened 2015-08-31
Return messages once they are gathered

djmitche:
If one or more messages are available, stop waiting and return them.

This avoids waiting the full `wait_seconds_timeout` when there are messages in the queue

More importantly, it avoids returning more than `count` messages.  Boto's `Queue.read` passes `count=1`, an then returns the message if `len(rs) == 1`, otherwise None.  So if `len(rs) > 1`, it returns `None` and discards the multiple messages returned.

### DIFF
--- a/moto/sqs/models.py
+++ b/moto/sqs/models.py
@@ -281,7 +281,7 @@ class SQSBackend(BaseBackend):
                 if len(result) >= count:
                     break
 
-            if time.time() > polling_end:
+            if result or time.time() > polling_end:
                 break
 
         return result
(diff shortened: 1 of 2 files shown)


spulec:
Thanks for opening!

This makes sense to me, though it looks like we may need to modify one of the tests.

djmitche:
From the AWS docs: "The duration (in seconds) for which the call will wait for a message to arrive in the queue before returning. If a message is available, the call will return sooner than WaitTimeSeconds."

So I think the implementation is correct, but `test_messages_polling` goes somewhat out of its way to make sure that ReceiveMessage does *not* return sooner than WaitTimeSeconds.

spulec:
Yes, I agree with your assessment. I'm trying to figure out if we can change that test to be meaningful or if we should just drop it. Thoughts?

djmitche:
howzat?

spulec:
Perfect, thanks for your contribution!

PR #412, opened 2015-09-04
Tweak ResponseObject to let bucket.delete_keys work with s3bucket_path.

ybastide:
Add `ResponseObject.is_delete_keys`, so `_bucket_response_post` also recognize delete_keys response with s3bucket_path: The response path is `/bucket/?delete`, not `/?delete`.


spulec:
Awesome, thank you!

PR #413, opened 2015-09-08
Add support for latency based route53 records

jszwedko:
Store and marshal the region field of records

Similar to the support of weighted records.

Signed-off-by: Kevin Donnelly <kevin.donnelly@getbraintree.com>


spulec:
Great, thank you!

PR #414, opened 2015-09-10
DynamoDB2 is overwriting the table instead of returning an error

2mf:
This fix returns 400 and ResourceInUseException, in case  you try to create the table but the table already exists. According to AWS DynamoDB documentation.


2mf:
bump

spulec:
thanks

PR #417, opened 2015-09-16
Datapipeline

spulec:
First version, I still need to finish the implementation of Pipeline.activate.

@hltbra thoughts?

### DIFF
--- a/moto/datapipeline/models.py
+++ b/moto/datapipeline/models.py
@@ -28,13 +28,14 @@ class Pipeline(object):
         self.pipeline_id = get_random_pipeline_id()
         self.creation_time = datetime.datetime.utcnow()
         self.objects = []
+        self.status = "PENDING"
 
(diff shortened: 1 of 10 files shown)


moto/datapipeline/models.py
```diff
@@ -0,0 +1,122 @@
...
+        self.pipeline_id = get_random_pipeline_id()
+        self.creation_time = datetime.datetime.utcnow()
+        self.objects = []
+
+    def to_json(self):
+        return {
+            "Description": self.description,
+            "Fields": [{
+                "key": "@pipelineState",
+                "stringValue": "SCHEDULED"
```
hltbra: I think that in order to be more accurate, the initial state should be `PENDING`, and when `activate()` is called, it becomes `SCHEDULED`. This is a minor and I am happy with the current implementation, but that change would make it more accurate.

PR #425, opened 2015-09-24
Remade ec2.utils.random_key_pair for be really random

ZuluPro:
The function ``ec2.utils.random_key_pair`` returned fixed material and fingerprint. I've made a simple system for produce it randomly like written in the comment above.

I use ``list()ˋ` around ranges because, they are generators in Python 3 and can't be sum together.

I've also added unittest and test perf:
```
>>> %timeit random_key_pair()
1000 loops, best of 3: 1.69 ms per loop
```
(On a 1CPU VM on MacBook Pro)

Let me know if you want to create real keys with paramiko, it will be more realist but more expensive in performance. Paramiko is included in boto.

### DIFF
--- a/moto/ec2/utils.py
+++ b/moto/ec2/utils.py
@@ -453,27 +453,22 @@ def simple_aws_filter_to_re(filter_string):
     return tmp_filter
 
 
-# not really random ( http://xkcd.com/221/ )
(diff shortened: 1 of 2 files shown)


spulec:
I'm happy with this, thanks!

PR #431, opened 2015-10-05
Fixes #430: MD5 hashing should be done to the real body, not an escaped one

### LINKED ISSUE #430: SQS MD5 Hashing Issues
I've started using Moto as a standalone server to aid testing a PHP stack.  I've discovered that once I create a message which contains encapsulated (escaped) JSON - it starts to fail with the AWS PHP SDK, although it works fine with Boto2.

The issue appears to be in and around the calculation of the MD5 sum.  I suspect Boto must be ignoring the MD5, otherwise I'd think it would be reporting the same problem.

The exception I get from PHP SDK:
```
PHP Warning:  Uncaught Aws\Sqs\Exception\SqsException: AWS Error Code: , Status Code: , AWS Request ID: , AWS Error Type: , AWS Error Message: Body MD5 mismatch for array (
  'MessageId' => '97f171c9-b7a5-b764-f3e0-4234555f509f',
  'ReceiptHandle' => 'nntoxkpevzvvbvbvylufszndstdeplilaxnckhsceeztjvmdqtzpxptfoeyndfgscncydyntjilbppbgsrwlldsjpksxklybpayijnoewirfexullvcdtmbvuablunaykrqomudptfmnznriseoegwopnaxidtwwsmoikjndpaxilaicgcbpisdpt',
  'MD5OfBody' => '08ab38f810e137a6cce4990c3952be77',
  'Body' => '{
```

Trying to reproduce that MD5 of the body using the same body contents from a json file:

PHP:
```
php > $body = file_get_contents(__DIR__ . '/test.json');
php > echo md5($body);
6d8dc937d72f4cdfad4b76be545dda6b
```

Python:
```
>>> import hashlib
>>> with open('git_src/api/data/sqs/ses/temp_bounce.json') as myfile:
...     data=myfile.read()
>>> hashlib.md5(data).hexdigest()
'6d8dc937d72f4cdfad4b76be545dda6b'
>>> from xml.sax.saxutils import escape
>>> hashlib.md5(escape(data).encode('utf-8')).hexdigest()
'08ab38f810e137a6cce4990c3952be77'
```

So it seems the XML escaping is causing the problem.  Before I put together a PR I'll confirm how the real AWS SQS service calculates this MD5.

araines:
Resolves compatibility issues with Amazon PHP SDK for SQS.

### DIFF
--- a/moto/sqs/models.py
+++ b/moto/sqs/models.py
@@ -34,7 +34,7 @@ class Message(object):
     @property
     def md5(self):
         body_md5 = hashlib.md5()
-        body_md5.update(self.body.encode('utf-8'))
+        body_md5.update(self._body.encode('utf-8'))
         return body_md5.hexdigest()
 
     @property


spulec:
Ah, great. Thanks!

PR #433, opened 2015-10-05
Change CloudFormationResponse.get_template() to return `GetTemplateResponse/GetTemplateResult/TemplateBody`

### LINKED ISSUE #432: CloudFormation's GetTemplate response is inaccurate
The GetTemplate function returns through `boto` the following format:
```python
>>> import boto.cloudformation
>>> cf = boto.cloudformation.connect_to_region('us-east-1')
>>> cf.get_template('<my-stack-id>')
{u'GetTemplateResponse': {u'GetTemplateResult': {u'TemplateBody': u'{"Resources": "<RESOURCES HERE>"}', u'ResponseMetadata': {u'RequestId': u'aaaaaaaa-bbbb-cccc-dddd-eeeeeeeeeeee'}}}
```

But moto returns just the template body... Moto's source: https://github.com/spulec/moto/blob/1dcfcbca5711ac50f9c3314b8a7e8812e3025156/moto/cloudformation/responses.py#L87-L91

AWS Docs: http://docs.aws.amazon.com/AWSCloudFormation/latest/APIReference/API_GetTemplate.html

hltbra:
This pull request fixes #432.

I noticed that some methods use a constant as XML template and other methods use an inline dict (and then call `json.dumps()` on it). This should probably be all constants, right? I followed what existed, but I think it can be improved.
Also, this works perfectly with boto2, [which has JSON as content type hardcoded for `GetTemplate`](https://github.com/boto/boto/blob/dc33c30ee24623c3cc6418bcdaa5c1254ae45b21/boto/cloudformation/connection.py#L692-L715), but it will not work with any client that relies on XML.

Can you please review, @spulec?

### DIFF
--- a/moto/cloudformation/responses.py
+++ b/moto/cloudformation/responses.py
@@ -86,9 +86,19 @@ class CloudFormationResponse(BaseResponse):
 
     def get_template(self):
         name_or_stack_id = self.querystring.get('StackName')[0]
-
         stack = self.cloudformation_backend.get_stack(name_or_stack_id)
-        return stack.template
+
+        response = {
+            "GetTemplateResponse": {
+            "GetTemplateResult": {
+                "TemplateBody": stack.template,
+                "ResponseMetadata": {
+                    "RequestId": "2d06e36c-ac1d-11e0-a958-f9382b6eb86bEXAMPLE"
+                }
+            }
+          }
+        }
+        return json.dumps(response)
 
     def update_stack(self):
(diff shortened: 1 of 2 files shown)


spulec:
The inconsistency about constants vs inline is known. I go back and forth depending on the size, but you are probably correct that it should get standardized.

Thanks for this. A bit embarrassed it was broken in the first place.

PR #435, opened 2015-10-07
Add S3 ACL support.

mikekap:
For writes, this adds support for "x-amz-acl" headers as well as "x-amz-grant-*" headers. For reads this adds support for the ?acl endpoint. ACLs aren't actually enforced.

### DIFF
--- a/moto/s3/models.py
+++ b/moto/s3/models.py
@@ -24,6 +24,7 @@ class FakeKey(object):
         self.name = name
         self.value = value
         self.last_modified = datetime.datetime.utcnow()
+        self.acl = get_canned_acl('private')
         self._storage_class = storage
         self._metadata = {}
         self._expiry = None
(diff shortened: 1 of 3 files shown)


spulec:
This is great, thanks!

bibby:
A big thanks for this changeset (and moto in general).
But I'm in need of ACLs for Bucket objects, which isn't implemented.
My library code is starting to accrue null checks on `bucket.get_acl().acl` for the sake of the test suite.

I'm using this with ceph/radosgw instead of AWS, but the same bucket policies should be in play.
Hopefully, we can revisit this.

spulec:
I just pushed some support for buckets with c38731ecbbfbbfc66fe41c0cf14d6d86ee17ebd2. Can you let me know if that is sufficient?

bibby:
YESSS!
So far, that's exactly what I needed.
Tremendous thanks.

PR #437, opened 2015-10-14
Set snapshots to be in a completed state after being created and added

nuwang:
This would mirror the functionality of volumes when snapshot.update() is called.


spulec:
Thanks!

PR #423, opened 2015-09-24
Added fixture system for images

ZuluPro:
Added the ability to have images set after mock starting.
Fixtures are set by a dictionary example:

```
{
  'us-east-1': {
    'images': [
      {'ami_id': 'foo_id', 'name': 'foo_name'},
      {'ami_id': 'bar_id', 'name': 'fbar_name'}
    ]
  }
}
```

PR #438, opened 2015-10-19
Added describe instance types

ZuluPro:


ZuluPro:
I think next step should be to use fixtures system I implemented here https://github.com/spulec/moto/pull/423 ?

spulec:
thanks!

PR #439, opened 2015-10-20
Use correct kwarg for master_user_password when creating rds2 database

yorinasub17:
This uses the correct [kwarg to set the `master_user_password`](https://github.com/spulec/moto/blob/master/moto/rds2/models.py#L40) in the rds2 backend.


spulec:
thanks!

PR #443, opened 2015-10-29
When adding security group rules, allow a source group only to be spe…

afgane:
…cified - as per boto docs & functionality: https://github.com/boto/boto/blob/develop/boto/ec2/securitygroup.py#L150

Adding a rule as follows should work: `security_group.authorize(src_group=sg)` while without this patch, it breaks.


spulec:
Thank you!

Issue #429, opened 2015-09-30
Add SWF support


jbbarth:
I'd really like to see SWF support in moto, so I started hacking on the easiest endpoints here: https://github.com/botify-labs/moto/tree/feature/add-swf, with some implementation notes here: https://github.com/botify-labs/moto/wiki

I have a few questions before continuing:
- do you think it would be possible to include it in moto? or should I make a separate library?
- is there any kind of automatation to ease endpoint integration from boto docs or source code? not that it's a problem but if there's something...
- are there general guidelines for implementing errors? I saw some parts of moto use boto exceptions directly, while other reimplement their own ones
- if an integration upstream is possible, should I open a PR now as "WIP" for early comments? should I make separate incremental pull requests as I implement endpoints?

Thanks for any feedback and keep up the good work with moto ;-)

spulec:
- Yes, I would love it.
- Nothing easy now, all manual :/
- We try not to use boto exceptions anymore. For a example of what we do now, see [here](https://github.com/spulec/moto/blob/master/moto/ec2/exceptions.py).
- If you want to open it now, feel free and I can try to give some early feedback.

Let me know if there is anything else I can do to help!

spulec:
This was added with #451

PR #451, opened 2015-11-09
[wip] Add SWF support

jbbarth:
*Work in progress, do NOT merge!*

See #429. This branch implements mocks for SWF endpoints, mapping directly to `boto.swf.Layer1`, which is the low-level layer used by `boto.swf.Layer2` and libraries above. As SWF is mostly a coordination service with complex logic, so making moto/swf useful boils down to reimplementing part of the logic behind SWF (much more easy though: non distributed, few limits, no permissions).

## Implementation progress

Implemented endpoints ([by category](http://docs.aws.amazon.com/amazonswf/latest/developerguide/swf-api-by-category.html)):
- **Actions Related to Activities**
  - [x] PollForActivityTask
  - [x] RespondActivityTaskCompleted
  - [x] RespondActivityTaskFailed
  - [ ] RespondActivityTaskCanceled
  - [x] RecordActivityTaskHeartbeat
- **Actions Related to Deciders**
  - [x] PollForDecisionTask
  - [x] RespondDecisionTaskCompleted (partial; see below)
- **Actions Related to Workflow Executions**
  - [ ] RequestCancelWorkflowExecution
  - [x] StartWorkflowExecution
  - [ ] SignalWorkflowExecution
  - [x] TerminateWorkflowExecution
- **Actions Related to Administration**
  - *Activity Management*
    - [x] RegisterActivityType
    - [x] DeprecateActivityType
  - *Workflow Management*
    - [x] RegisterWorkflowType
    - [x] DeprecateWorkflowType
  - *Domain Management*
    - [x] RegisterDomain
    - [x] DeprecateDomain
- **Visibility Actions**
  - *Activity Visibility*
    - [x] ListActivityTypes
    - [x] DescribeActivityType
  - *Workflow Visibility*
    - [x] ListWorkflowTypes
    - [x] DescribeWorkflowType
  - *Workflow Execution Visibility*
    - [x] DescribeWorkflowExecution
    - [ ] ListOpenWorkflowExecutions
    - [ ] ListClosedWorkflowExecutions
    - [ ] CountOpenWorkflowExecutions
    - [ ] CountClosedWorkflowExecutions
    - [x] GetWorkflowExecutionHistory
  - *Domain Visibility*
    - [x] ListDomains
    - [x] DescribeDomain
  - *Task List Visibility*
    - [x] CountPendingActivityTasks
    - [x] CountPendingDecisionTasks

Implemented decision types for [RespondDecisionTaskCompleted endpoint](http://docs.aws.amazon.com/amazonswf/latest/apireference/API_RespondDecisionTaskCompleted.html):
- [ ] CancelTimer
- [ ] CancelWorkflowExecution
- [ ] ContinueAsNewWorkflowExecution
- [x] CompleteWorkflowExecution
- [x] FailWorkflowExecution
- [ ] RecordMarker
- [ ] RequestCancelActivityTask
- [ ] RequestCancelExternalWorkflowExecution
- [x] ScheduleActivityTask
- [ ] ScheduleLambdaFunction
- [ ] SignalExternalWorkflowExecution
- [ ] StartChildWorkflowExecution
- [ ] StartTimer

Implemented [timeout types](http://docs.aws.amazon.com/amazonswf/latest/developerguide/swf-timeout-types.html):
- [x] Workflow Execution: Execution Start to Close
- [x] Decision Task: Start to Close
- [x] Activity Task: Heartbeat
- [ ] Activity Task: Schedule to Close
- [ ] Activity Task: Schedule to Start
- [ ] Activity Task: Start to Close

Implemented [limits](http://docs.aws.amazon.com/amazonswf/latest/developerguide/swf-dg-limits.html) (most are hard to test or implement, this section is only there as reference):

- *General limits*
  - [ ] Maximum registered domains: 100
  - [ ] Maximum workflow and activity types: 10,000 each per domain [likely never]
  - [ ] Maximum request size: 1 MB per request
- *Limits on Workflow Executions*
  - [ ] Maximum open workflow executions: 100,000 per domain [likely never]
  - [ ] Maximum workflow execution time: 1 year [likely never]
  - [ ] Maximum workflow execution history size: 25,000 events [will do]
  - [ ] Maximum child workflow executions: 1,000 per workflow execution. [will do]
  - [ ] Workflow execution idle time limit: 1 year [likely never]
- *Limits on Task Executions*
  - [ ] Maximum task execution time: 1 year (constrained by workflow execution time limit) [likely never]
  - [ ] Maximum time SWF will keep a task in the queue: 1 year (constrained by workflow execution time limit) [likely never]
  - [ ] Maximum open activity tasks: 1,000 per workflow execution. [will do]
  - [ ] Maximum open timers: 1,000 per workflow execution [will do]
  - [ ] Maximum input/result data size: 32,000 characters [will do]
  - Maximum decisions in a decision task response [will do]
- *Throttling limits*
  - none: too hard and few benefits in the general case

## Known limitations

- permissions and request limits are not implemented, namely `LimitExceededFault` and `OperationNotPermittedFault` ; I don't think it's needed for now
- pagination and max number of results are generally not implemented (`nextPageToken` is omitted in responses ; `nextPageToken` and `maximumPageSize` have no effect in requests) ; but I think I'll implement them because I need to test SWF limits in our client library
- in case of timeouts, the timeout events are properly timestamped but the successive decisions may not be ; this is because we don't have a running daemon to handle things at regular interval, so history is computed on demand afterwards ; this will eventually be fixed
- + some TODO comments in the code

## Some remarks

- I wanted to have a working service before submitting a PR, but I do understand that it is a lot of code to review ; you can mostly follow the progress commit by commit, should be easier to handle
- I'm not happy with `moto.swf.responses.SWFResponse.call_action()`, which is completely copied from `BaseResponse` except one modified line (see comment inside) ; I don't know if it could be changed though,  what do you think?
- I had trouble working with freezed dates + timestamps ; Travis assumes dates are in UTC it seems, so I personnally run the tests with `TZ=GMT`; don't know if it should be part of the Makefile or something else
- I'm not sure which names should have been namespaced, let me know ; "Timeout" for instance may be too generic (?)
- I wonder where to keep some documentation, especially where to keep track of a status of what is implemented like I did above

As a general remark, I'm just a beginner in Python so let me know if you see any improvement, I'd be happy to learn.

Also I realize the end result might be a bit complicated for now. It could be simplified or released as a separate lib if you tell me so.

I'll continue to work on this PR in the next few days. Thanks!


spulec:
This is looking great!

I have no problem merging now and then we can have subsequent pull requests for additional features. Your call. Let me know when you would like me to start reviewing.

jbbarth:
Thanks! The current state allows for running administrative tasks as long as running a basic workflow. I'd be happy if you can start reviewing now and tell me what you think of the limitations/remarks I mentionned above. This will help me to improve things in the subsequent pull requests. I'm also open if you want to discuss some things on IRC or a Slack channel, let me know.

NB: I edited the initial PR text with a new section about implemented limits, and one more point in "remarks"

moto/swf/exceptions.py
```diff
@@ -0,0 +1,126 @@
+from __future__ import unicode_literals
+
+from boto.exception import JSONResponseError
+
+
+class SWFClientError(JSONResponseError):
```
spulec: We've been trying to move to using exceptions so that HTTPretty will return the error code instead of these exceptions jumping up the stack. This allows for the exceptions to work in stand-alone server mode as well. See [here](https://github.com/spulec/moto/blob/master/moto/sns/exceptions.py#L5-L10) for an example.

moto/swf/models/__init__.py
```diff
@@ -0,0 +1,405 @@
...
+        _types = sorted(_types, key=lambda domain: domain.name)
+        if reverse_order:
+            _types = reversed(_types)
+        return _types
+
+    def register_type(self, kind, domain_name, name, version, **kwargs):
+        self._check_string(domain_name)
+        self._check_string(name)
+        self._check_string(version)
+        for _, value in kwargs.items():
```
spulec: you can use `kwargs.values()` to iterate through just the values.

moto/swf/models/__init__.py
```diff
@@ -0,0 +1,405 @@
...
+
+    def start_workflow_execution(self, domain_name, workflow_id,
+                                 workflow_name, workflow_version,
+                                 tag_list=None, **kwargs):
+        self._check_string(domain_name)
+        self._check_string(workflow_id)
+        self._check_string(workflow_name)
+        self._check_string(workflow_version)
+        self._check_none_or_list_of_strings(tag_list)
+        for _, value in kwargs.items():
```
spulec: Same `kwargs.values()`

moto/swf/models/__init__.py
```diff
@@ -0,0 +1,405 @@
...
+        self._check_string(status)
+        domain = self._get_domain(domain_name)
+        _types = domain.find_types(kind, status)
+        _types = sorted(_types, key=lambda domain: domain.name)
+        if reverse_order:
+            _types = reversed(_types)
+        return _types
+
+    def register_type(self, kind, domain_name, name, version, **kwargs):
+        self._check_string(domain_name)
```
spulec: This isn't a huge deal, but it feels like this validation should be getting done in SWFResponse instead of in the model. We like to try to keep the models clean so they can be reused in other places (for example Cloudformation).

moto/swf/models/__init__.py
```diff
@@ -0,0 +1,405 @@
...
+        self._check_string(name)
+        self._check_string(version)
+        for _, value in kwargs.items():
+            self._check_none_or_string(value)
+        domain = self._get_domain(domain_name)
+        _type = domain.get_type(kind, name, version, ignore_empty=True)
+        if _type:
+            raise SWFTypeAlreadyExistsFault(_type)
+        _class = globals()["{0}Type".format(kind.capitalize())]
```
spulec: The use of `globals()` is generally frowned upon in python. I would probably just create a module-level mapping for these.

moto/swf/models/activity_task.py
```diff
@@ -0,0 +1,84 @@
...
+        self.scheduled_event_id = scheduled_event_id
+        self.started_event_id = None
+        self.state = "SCHEDULED"
+        self.task_token = str(uuid.uuid4())
+        self.timeouts = timeouts
+        self.timeout_type = None
+        self.workflow_execution = workflow_execution
+        # this is *not* necessarily coherent with workflow execution history,
+        # but that shouldn't be a problem for tests
+        self.scheduled_at = datetime.now()
```
spulec: Would probably change to `datetime.utcnow()`

moto/swf/models/domain.py
```diff
@@ -0,0 +1,124 @@
...
+
+    def add_to_activity_task_list(self, task_list, obj):
+        if not task_list in self.activity_task_lists:
+            self.activity_task_lists[task_list] = []
+        self.activity_task_lists[task_list].append(obj)
+
+    @property
+    def activity_tasks(self):
+        _all = []
+        for _, tasks in self.activity_task_lists.items():
```
spulec: I think we can just `return self.activity_task_lists.values()`

moto/swf/models/domain.py
```diff
@@ -0,0 +1,124 @@
...
+
+    def add_to_decision_task_list(self, task_list, obj):
+        if not task_list in self.decision_task_lists:
+            self.decision_task_lists[task_list] = []
+        self.decision_task_lists[task_list].append(obj)
+
+    @property
+    def decision_tasks(self):
+        _all = []
+        for _, tasks in self.decision_task_lists.items():
```
spulec: Same as above

moto/swf/models/history_event.py
```diff
@@ -0,0 +1,169 @@
...
+            "eventTimestamp": self.event_timestamp,
+            self._attributes_key(): self.event_attributes()
+        }
+
+    def _attributes_key(self):
+        key = "{0}EventAttributes".format(self.event_type)
+        return decapitalize(key)
+
+    def event_attributes(self):
+        if self.event_type == "WorkflowExecutionStarted":
```
spulec: Not a big deal, but maybe worth creating a property for each of these conditions to clean this up a bit.

moto/swf/responses.py
```diff
@@ -0,0 +1,311 @@
...
+    # "x-amz-target", in the form of com.amazonaws.swf.service.model.SimpleWorkflowService.<action>
+    # This is not supported directly in BaseResponse sor for now we override
+    # the call_action() method
+    # See: http://docs.aws.amazon.com/amazonswf/latest/developerguide/UsingJSON-swf.html
+    def call_action(self):
+        headers = self.response_headers
+        # Headers are case-insensitive. Probably a better way to do this.
+        match = self.headers.get('x-amz-target') or self.headers.get('X-Amz-Target')
+        if match:
+            # TODO: see if we can call "[-1]" in BaseResponse, which would
```
spulec: Agreed. I changed this with 32dd72f6b7289620fb149bd876858d5293a118f0

spulec:
After more review, I still think this is great :)

A few small comments, but pretty solid code (especially for a Python beginner). If you address a few of the comments, I would be happy for us to merge this now. I don't think any of the limitations are blockers. 

You should also feel free to add yourself to `AUTHORS.md`

moto/swf/models/__init__.py
jbbarth: Done, great!

moto/swf/models/__init__.py
jbbarth: Done too

moto/swf/models/__init__.py
jbbarth: I'll move them no problem, makes sense!

moto/swf/models/__init__.py
jbbarth: Probably a Ruby habit, I removed it in favor of a mapping in a dict (plus it won't change very often, so no point in being dynamic)

moto/swf/models/activity_task.py
jbbarth: Changed, as long as other calls to datetime.now()

moto/swf/models/domain.py
jbbarth: Nope, unfortunately: the inner loop doesn't perform a `list.append()` but a `+` ; if I replace by `.values()` I get nested lists (like `[["foo"], ["bar", "baz"]]`) instead of a flat list (`["foo", "bar", "baz"]`). I didn't found a way to easily "flatten" a list.

moto/swf/models/domain.py
jbbarth: See above

moto/swf/models/history_event.py
jbbarth: Sure, I'll do that! Honestly the logic is nearly always the same, it could probably be trimmed down to a dozen lines. At first I thought I would do a lot of complex formatting here, but in fact no.

moto/swf/responses.py
jbbarth: Awesome! I removed this accordingly, and I rebased the whole PR against master. I'll `git push -f` now, hope you don't mind (my current pythonistas team seem to favor that for topic branches, don't know if it's a general practice in python projects).

moto/swf/exceptions.py
jbbarth: I must admit I cannot wrap my head around this one: when I use `werkzeug.exceptions.HTTPException` as a base class for my exceptions, the tests now say that the calls raise a `boto.exception.SWFResponseError`, thought I didn't use it anywhere explicitly. Not sure if it's OK or not, I'd prefer discussing it in another PR if you're OK, so I can focus on it better and not pollute the commit history with trials/errors here.

jbbarth:
@spulec: thanks a lot for your review, lots of insights!

I'll re-push now with:
- all the small things fixed
- TODO now: move type checks into response to keep models clean
- TODO now: refactor the history events handling (I agree it was a bit ugly)
- TODO later: use werkzeug exceptions instead of boto ones ; not sure how I'll do that, see my comment above ; but I'll dig into that

I should have the 2 "TODO now" done before tomorrow, if you're OK we could merge this once they're done.

Thanks again!

jbbarth:
@spulec: done as for the two minor "TODOs", sorry for the delay. I let you review the last commits when you have some time. Let me know what you think for the errors base exception. Thanks!

moto/swf/exceptions.py
spulec: Yup, that is the goal. Raising a `werkzeug.exceptions.HTTPException` will have the response return a `4xx` status code back to the boto client which will then raise the `SWFResponseError`.

spulec:
Awesome, thanks again for all your work.

pior:
Wow! Awesome job!
I wish I had seen this PR before jumping directly to httpretty for my tests.

Maybe SWF should be mentioned in the README ?

jbbarth:
Thanks @prior! Forgot about the README yes, I'll update it in the next PR.

Issue #450, opened 2015-11-09
Moto should respond with a 400 when trying to send a message > 256KiB


iandees:
SQS API responds like this when sending a message > 262144 bytes:

```
SQSError: SQSError: 400 Bad Request
<?xml version="1.0"?><ErrorResponse xmlns="http://queue.amazonaws.com/doc/2012-11-05/"><Error><Type>Sender</Type><Code>InvalidParameterValue</Code><Message>One or more parameters are invalid. Reason: Message must be shorter than 262144 bytes.</Message><Detail/></Error><RequestId>xxxxxx-xxxx-xxxx-xxxx-xxxxxxx</RequestId></ErrorResponse>
```

PR #452, opened 2015-11-10
Check SQS message size

iandees:
Fixes https://github.com/spulec/moto/issues/450

### DIFF
--- a/moto/sqs/exceptions.py
+++ b/moto/sqs/exceptions.py
@@ -11,11 +11,6 @@ class ReceiptHandleIsInvalid(Exception):
     status_code = 400
 
 
-class InvalidParameterValue(Exception):
-    description = "One or more parameters are invalid. Reason: Message must be shorter than 262144 bytes."
-    status_code = 400
-
-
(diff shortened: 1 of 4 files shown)


spulec:
Awesome. Would you mind adding a small test?

iandees:
Definitely!

thedrow:
SQS now supports large messages by storing them in S3.
See https://aws.amazon.com/about-aws/whats-new/2015/10/now-send-payloads-up-to-2gb-with-amazon-sqs/
I think that only boto3 supports that but I'm not 100% sure since I'm not using boto3.

iandees:
Wow, thanks for pointing this out. I added this because I was implementing that in boto :smile:.

iandees:
@spulec Test added and it's passing.

spulec:
Great, thanks for your work!

iandees:
I don't suppose you could cut a new tag/release for this? :smile:

spulec:
Just pushed moto 0.4.19

PR #453, opened 2015-11-12
Fix Kinesis API

2mf:
Kinesis:
add put_records
fix create_stream to emit right response in case of stream already exists


spulec:
Thanks!

PR #456, opened 2015-11-15
Implement JSON responses for EMR (for Boto3)

achiku:
@spulec 

I started adding Boto3 support for EMR module, and will be happy if you can review the modification. 

Most of my modification is pretty much the same with https://github.com/spulec/moto/pull/404, but one difference is switching JSON and XML response by looking at `User-Agent` instead of `querystring` since I couldn't get `JSON` string from QueryString created by boto3 emr client, and IMHO it makes a bit more clear this is for Boto3.

### DIFF
--- a/moto/core/responses.py
+++ b/moto/core/responses.py
@@ -131,7 +131,7 @@ class BaseResponse(_TemplateEnvironmentMixin):
         else:
             self.region = self.default_region
 
-        self.headers = dict(request.headers)
+        self.headers = request.headers
         self.response_headers = headers
         return self.call_action()
 
--- a/moto/emr/responses.py
+++ b/moto/emr/responses.py
@@ -1,4 +1,5 @@
 from __future__ import unicode_literals
+import json
 
 from moto.core.responses import BaseResponse
 from .models import emr_backends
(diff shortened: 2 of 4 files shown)


achiku:
This is failing because Python2 and Python3's `requests` module return different keys in `BaseResponse.headers`, Py2 returns 'user-agent' and Py3 returns 'User-Agent'. I can workaround this problem using if statement for Py2 and Py3, but trying to find better way. Any idea?

In `moto/core/responses.py`, requests's header is converted to plane dict.
https://github.com/spulec/moto/blob/master/moto/core/responses.py#L134

Is there any reason to do this? Since `requests.Response.headers` is case-insensitive dict-like object, it seems compatible to common use case.

http://docs.python-requests.org/en/v1.0.0/api/#requests.Response.headers

achiku:
Py3 tests all passed with `BaseResponse.headers` as `requests`'s header, not a dict. But Py2 test failed (` tests/test_kms/test_kms.py:test_describe_missing_key`).

spulec:
I would prefer to use the `content-type` header. Although moto is mostly designed for use with Boto, we aim to provide an interface that is agnostic of the client.

It seems that something like `'json' in self.headers['content-type']` should work.

achiku:
@spulec 
Thanks for the comment! PR updated!

spulec:
Great, thank you!

Issue #458, opened 2015-11-17
Error in SQS send_message with non-ascii messages


carldunham:
Will also work on a PR for this one:

``` python
# -*- coding: utf-8 -*-

import boto3
from moto import mock_sqs


with mock_sqs():
    session = boto3.session.Session(aws_access_key_id='x', aws_secret_access_key='y', region_name='us-east-1')
    sqs = session.resource('sqs')

    q = sqs.create_queue(QueueName='testq')
    q.send_message(MessageBody='hello😀')

    print(q.receive_messages())
```

Results running in p3:

```
Traceback (most recent call last):
  File "/usr/local/lib/python3.5/site-packages/botocore/vendored/requests/packages/urllib3/connectionpool.py", line 372, in _make_request
    httplib_response = conn.getresponse(buffering=True)
TypeError: getresponse() got an unexpected keyword argument 'buffering'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "<stdin>", line 12, in <module>
  File "/usr/local/lib/python3.5/site-packages/boto3/resources/factory.py", line 394, in do_action
    response = action(self, *args, **kwargs)
  File "/usr/local/lib/python3.5/site-packages/boto3/resources/action.py", line 77, in __call__
    response = getattr(parent.meta.client, operation_name)(**params)
  File "/usr/local/lib/python3.5/site-packages/botocore/client.py", line 310, in _api_call
    return self._make_api_call(operation_name, kwargs)
  File "/usr/local/lib/python3.5/site-packages/botocore/client.py", line 384, in _make_api_call
    operation_model, request_dict)
  File "/usr/local/lib/python3.5/site-packages/botocore/endpoint.py", line 112, in make_request
    return self._send_request(request_dict, operation_model)
  File "/usr/local/lib/python3.5/site-packages/botocore/endpoint.py", line 141, in _send_request
    success_response, exception):
  File "/usr/local/lib/python3.5/site-packages/botocore/endpoint.py", line 214, in _needs_retry
    caught_exception=caught_exception)
  File "/usr/local/lib/python3.5/site-packages/botocore/hooks.py", line 226, in emit
    return self._emit(event_name, kwargs)
  File "/usr/local/lib/python3.5/site-packages/botocore/hooks.py", line 209, in _emit
    response = handler(**kwargs)
  File "/usr/local/lib/python3.5/site-packages/botocore/retryhandler.py", line 183, in __call__
    if self._checker(attempts, response, caught_exception):
[...]

PR #460, opened 2015-11-17
Fix for #458 - Allow unicode in SQS messages

carldunham:
The fix is in `core/responses.py`, which may impact non-sqs calls as well. If you would rather, I can isolate it to sqs. If you are OK with leaving it there, I wonder if there are places that could benefit from tests that would include unicode or other non-ascii content in responses. I would be happy to add those, if needed.

### DIFF
--- a/moto/core/responses.py
+++ b/moto/core/responses.py
@@ -115,7 +115,10 @@ class BaseResponse(_TemplateEnvironmentMixin):
         if not querystring:
             querystring.update(parse_qs(urlparse(full_url).query, keep_blank_values=True))
         if not querystring:
-            querystring.update(parse_qs(self.body, keep_blank_values=True))
+            body = self.body
+            if isinstance(body, six.binary_type):
+                body = body.decode('utf-8')
+            querystring.update(parse_qs(body, keep_blank_values=True))
         if not querystring:
             querystring.update(headers)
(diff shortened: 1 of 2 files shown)


moto/core/responses.py
```diff
@@ -115,7 +115,10 @@ def _dispatch(self, request, full_url, headers):
         if not querystring:
             querystring.update(parse_qs(urlparse(full_url).query, keep_blank_values=True))
         if not querystring:
-            querystring.update(parse_qs(self.body, keep_blank_values=True))
+            body = self.body
+            if six.PY3 and isinstance(body, bytes):
```
spulec: Any reason to do this as opposed to `if isinstance(body, six.binary_type)`? We want to do the decoding in Py2, right?

spulec:
Thanks!

One small question, but will be happy to merge after.

moto/core/responses.py
carldunham: Good point. Done.

spulec:
Looks like Travis isn't happy

spulec:
I Pulled these changes in with https://github.com/spulec/moto/commit/a0a205328d44f6f423fe8f4f25be6b2e0af8b7b1

Thanks!

PR #466, opened 2015-11-25
Adding Support for Get All DHCP Options

EarthmanT:
add DescribeDhcpOptionsResponse
support filtering describe_dhcp_options
add get_all_dhcp_options

### DIFF
--- a/moto/ec2/utils.py
+++ b/moto/ec2/utils.py
@@ -185,14 +185,6 @@ def network_acl_ids_from_querystring(querystring_dict):
     return network_acl_ids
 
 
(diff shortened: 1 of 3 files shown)


moto/ec2/utils.py
```diff
@@ -185,6 +185,14 @@ def network_acl_ids_from_querystring(querystring_dict):
     return network_acl_ids
 
 
+def dhcp_opt_ids_from_querystring(querystring_dict):
```
spulec: I think we can use the `sequence_from_querystring` below

EarthmanT:
@spulec made the requested change

spulec:
thanks!

PR #468, opened 2015-11-25
Adding Support for VPN Connections

EarthmanT:
support create, delete, describe


moto/ec2/utils.py
```diff
@@ -193,6 +197,14 @@ def vpc_ids_from_querystring(querystring_dict):
     return vpc_ids
 
 
+def vpn_connection_ids_from_query_string(querystring_dict):
```
spulec: I think we can use the `sequence_from_querystring` below instead.

spulec:
Could you add a couple tests?

EarthmanT:
@spulec added tests

spulec:
thanks!

PR #478, opened 2015-12-06
Add support for default VPCs and subnets.

mrucci:
The first VPC created will be elected as default.
All subnets of the default VPC are considered default for their availability zone.


spulec:
Great, thank you!

PR #481, opened 2015-12-09
Manage instances of ELBs attached to autoscaling group.

mrucci:
This replicate the AWS functionality of keeping the ELB instances in sync with the autoscaling group instances.


spulec:
Awesome, thank you!

PR #482, opened 2015-12-09
Add new endpoints to ELB (fix #479)

### LINKED ISSUE #479: boto3 + elb + describe_tags = NotImplementedError in moto
Hi,

I'm hitting an error when trying to add tests to some existing code that is using boto3 client library with moto: `NotImplementedError: The describe_tags action has not been implemented`: https://travis-ci.org/zalando/zmon-aws-agent/jobs/95541858

This happens with the ELB client.  The problem appears to be that even though the XML request/response for this API are well defined, the old boto interface doesn't provide a method for this particular call.

Do you see a good way around this sort of issues short of re-implementing moto with boto3 interfaces?

2mf:
This will add the endpoints AddTags, RemoveTags and DescribeTags to ELB.
This should fix #479

### DIFF
--- a/moto/elb/exceptions.py
+++ b/moto/elb/exceptions.py
@@ -1,0 +1,29 @@
+from __future__ import unicode_literals
+from moto.core.exceptions import RESTError
+
+
+class ELBClientError(RESTError):
+    code = 400
+
+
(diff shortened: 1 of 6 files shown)


spulec:
Thank you!

Issue #484, opened 2015-12-13
ELB response gives empty list for missing load balancer instead of thrown exception


gilzellner:
# Boto behaviour:

In [9]: conn.get_all_load_balancers()
Out[9]: [LoadBalancer:STICKYLB]

In [10]: conn.get_all_load_balancers(load_balancer_names='aaa')
BotoServerError: BotoServerError: 400 Bad Request
        <Message>Cannot find Load Balancer aaa</Message>

# moto behaviour: 
In [14]: conn.get_all_load_balancers()
Out[14]: [LoadBalancer:STICKYLB]

In [14]: conn.get_all_load_balancers(load_balancer_names='aaa')
Out[14]: []

spulec:
Thanks for opening!

Fixed with https://github.com/spulec/moto/commit/5d421dc343046bda58220b047133d5b418248ee8

PR #485, opened 2015-12-13
ELB response gives empty list for missing load balancer instead of th…

gilzellner:
ELB response gives empty list for missing load balancer instead of thrown exception #484


spulec:
Thanks for this. I fixed it with https://github.com/spulec/moto/commit/5d421dc343046bda58220b047133d5b418248ee8 so that the standalone server version can handle the exception as well.

PR #486, opened 2015-12-14
Add query index support for boto3 via dynamodb2

pcraciunoiu:
This adds a bunch of support and tests for boto3 dynamodb.

I've tested this as part of a larger upgrade for one of the projects I work on.

### DIFF
--- a/moto/dynamodb2/responses.py
+++ b/moto/dynamodb2/responses.py
@@ -291,7 +291,8 @@ class DynamoHandler(BaseResponse):
                             range_comparison = None
                             range_values = []
 
-        items, last_page = dynamodb_backend2.query(name, hash_key, range_comparison, range_values)
+        index_name = self.body.get('IndexName')
+        items, last_page = dynamodb_backend2.query(name, hash_key, range_comparison, range_values, index_name=index_name)
         if items is None:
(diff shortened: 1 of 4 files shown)


moto/dynamodb2/responses.py
```diff
@@ -239,11 +242,31 @@ def query(self):
...
+            else:
+                index = table.schema
+
+            key_map = [column for _, column in sorted((k, v) for k, v in self.body['ExpressionAttributeNames'].items())]
+
             if " AND " in key_condition_expression:
                 expressions = key_condition_expression.split(" AND ", 1)
-                hash_key_expression = expressions[0]
+
+                index_hash_key = [key for key in index if key['KeyType'] == 'HASH'][0]
```
pcraciunoiu: This also fixes an issue where the hash key would have to be before the range key in the expressions.

pcraciunoiu:
@spulec I hope the tests explain what this all does. I'm not 100% sure it doesn't break anything, but I am beginning to use it in a project. I'll update if I find any bugs.

Would appreciate your thoughts anyways, and it would be great to get this upstream if it seems reasonably well written to you.

spulec:
this is great, thanks!

PR #487, opened 2015-12-14
Fix ELB.describe_tags with multiple LoadBalancerNames.

a1exsh:


a1exsh:
@spulec @2mf thanks for #482, but that was not enough apparently ;-)

2mf:
Ah I missed that one ;-)
The other Add and Remove endpoint only allowed "1" ELB specified per request

spulec:
Thanks!

Issue #498, opened 2015-12-31
Partial Save Not Supported on DynamoDB2 Tables with Range Key


im-auld:
Partial Save doesn't seem to be supported for DynamDB2 tables that have a range key.

spulec:
Fixed with #499

PR #499, opened 2015-12-31
Partial save support with range key

im-auld:
Added support for using `Item.partial_save()` with DynamoDB tables that have both a range and hash key. Previously an exception was raised. Issue #498


im-auld:
@spulec Anyway I can help expedite the review for this PR? Would love to be able to use this library at work and need this feature.

spulec:
this is great, thanks!

PR #504, opened 2016-01-09
Support for other UPDATE DynamoTypes and query Select=COUNT

pcraciunoiu:
@spulec Thanks for taking #499 and my PR #486 as well.

This is more work I did as part of a boto3 upgrade for DynamoDB. Improvements to UPDATE and support for `Select=COUNT`.

Don't hesitate to leave comments!


spulec:
Looks great, thanks!

PR #506, opened 2016-01-11
Add filters for describeVolumes and describeSnapshots.

DenverJ:
This adds filters for volumes and snapshots. 

The following filters are implemented for volumes.
* attachment.attach-time
* attachment.device
* attachment.instance-id
* create-time
* size
* snapshot-id
* status
* tag
* tag-key
* tag-value

The following filters are not implemented for volumes due to missing information etc.
* attachment.delete-on-termination
* availability-zone (would currently only work for us-east AZs)
* encrypted
* volume-type

The following filters are implemented for snapshots.
* description
* snapshot-id
* start-time
* tag
* tag-key
* tag-value
* volume-id
* volume-size

The following filters are not implemented for snapshots.
* owner-alias
* owner-id
* progress
* status

### DIFF
--- a/moto/ec2/models.py
+++ b/moto/ec2/models.py
@@ -1430,6 +1430,36 @@ class Volume(TaggedEC2Resource):
         else:
             return 'available'
 
+    def get_filter_value(self, filter_name):
+
+        if filter_name.startswith('attachment') and not self.attachment:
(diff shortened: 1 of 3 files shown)


spulec:
Great, thanks!

PR #507, opened 2016-01-11
Delete missing attribute results in KeyError.

pcraciunoiu:
Found a small issue with one of my previous changes, if you try to delete an attribute that's already deleted.


spulec:
thanks!

PR #509, opened 2016-01-14
Bug fix secondary indexes ignored

im-auld:
Basic support for secondary indexes (`AllIndex`). Will need to ass support for QueryFiltering to get full support.

Currently when creating a table like this one:

```
Table.create(settings.NOTIFICATIONS_TABLE_NAME, schema=[
            HashKey('user_id', data_type=NUMBER),
            RangeKey('uniq', data_type=NUMBER),
        ], throughput={
            'read': 5,
            'write': 5,
        }, indexes=[
            AllIndex('timestamp_index', parts=[
                HashKey('user_id', data_type=NUMBER),
                RangeKey('timestamp'),
            ])
        ])
```

When querying this table like so:

```
notification_table.query_2(
    recipient_id__eq=user_id,
    timestamp__gt=timestamp_cutoff,  # this isn't supported yet, I will be working on that next
    index='recipient_timestamp_index',
    reverse=True
)
```

You will get an error in your test since currently the key `LocalSecondaryIndexes` isn't added to the table. This result in call to `query2()` failing due to there being no available indexes.

### DIFF
--- a/moto/dynamodb2/models.py
+++ b/moto/dynamodb2/models.py
@@ -184,6 +184,7 @@ class Table(object):
                 'ItemCount': len(self),
                 'CreationDateTime': unix_time(self.created_at),
                 'GlobalSecondaryIndexes': [index for index in self.global_indexes],
+                'LocalSecondaryIndexes': [index for index in self.indexes]
             }
         }
         return results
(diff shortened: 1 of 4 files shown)


spulec:
Thank you!

PR #510, opened 2016-01-14
Add value and UPDATE fixes

pcraciunoiu:
* Supporrt ADD Value for numeric types
* Create item when it doesn't exist on update (to match real DynamoDB behavior)
* When hash/range key overlap, range key may end up null


pcraciunoiu:
@spulec this is a bit of a mishmash of features, but I ran into them when working on a different project. Some tests illustrate. The hash/range key overlap was tricky, but if you look at the code you can tell.

spulec:
Great, thank you!

PR #512, opened 2016-01-15
Adds support for authorizing and revoking egress permissions

kevgliss:
Adds basic support for egress permission on aws security groups (vpc only)


spulec:
This is great; thank you!

PR #518, opened 2016-01-22
Fix delete keys with moto_server s3

millerdev:
Was getting errors when running tests against `moto_server s3` and boto3 client:
- first on the client: `ClientError: An error occurred (400) when calling the DeleteObjects operation: Bad Request`
- then on the server (after fixing first error): `AttributeError: 'Request' object has no attribute 'body'`


spulec:
thanks!

PR #531, opened 2016-02-03
Adding additional params to response

kevgliss:
Boto3 was complaining about not being able to find these attributes.


spulec:
Great, thank you!

PR #533, opened 2016-02-05
Ensuring boto3 can correctly identify instance security groups

kevgliss:
This is another subtle boto3 issue, it seems that it no longer expects that a dict will be passed ```{'id': '', 'name': ''}```. Instead it only requires (and passes) the security groupId. My guess was that the name was no longer needed. I believe the changes I made will be backward compatible with boto. 

There are probably a few other places this is an issue. I will try to update them as I find them.


spulec:
thanks again!

PR #535, opened 2016-02-09
Add method describe_clusters and fix create_cluster,  list_clusters, list_task_definitions,  list_services for service ecs

beeva-antonioirizar:
Add method describe_clusters and I fix create_cluster when it doesn't have name. 

It should create a cluster with name default and not with name None. See documentation http://docs.aws.amazon.com/cli/latest/reference/ecs/create-cluster.html

Remove nexToken from methods:
* list_clusters
* list_task_definitions
* list_services

Because the library doesn't support pagination and it  doesn't send nextToken or put it to None if it doesn't exceed maxResults. See documentation http://docs.aws.amazon.com/cli/latest/reference/ecs/list-clusters.html

### DIFF
--- a/moto/ecs/models.py
+++ b/moto/ecs/models.py
@@ -117,6 +117,20 @@ class EC2ContainerServiceBackend(BaseBackend):
         """
         return [cluster.arn for cluster in self.clusters.values()]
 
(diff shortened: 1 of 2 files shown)


spulec:
thanks!

PR #536, opened 2016-02-10
Support placement when starting new instances

andyjones:
Thanks for this module, finding it very useful!

Boto/AWS lets you specify the availability zone when starting [new instances](http://boto.cloudhackers.com/en/latest/ref/ec2.html#boto.ec2.connection.EC2Connection.run_instances)


spulec:
Great, thanks!

PR #543, opened 2016-02-17
Include tests module and test files in sdist

koobs:
Allows downstream users and operating system packagers and porters to test released package within their own QA, build or other environments


spulec:
thanks!

PR #544, opened 2016-02-17
Fix route53

beeva-antonioirizar:
* Add IsTruncated to False in template list_hosted_zones_response in route53. Because it always has it.

* add private zone in templates. Because it should have. It is always to False because it doesn't support for private zones.

* Comment's params is optional in create_hosted_zones. It is fix to support and templates adapted to it.

* add responses for methods don't have implemented in route53


spulec:
thanks!

PR #545, opened 2016-02-18
Cloudwatch support multiple regions

beeva-antonioirizar:
Update Cloudtwatch to support multiple regions. Now it can have different metrics in multiple regions.


spulec:
Perfect, thanks!

PR #546, opened 2016-02-20
boto3 SQS set attributes fix.

scttl:
In boto3, the SQS `set_queue_attributes` items are now presented as a list of values of the form `Attribute.1.Name`, `Attribute.1.Value`, `Attribute.2.Name`, and so on.

This patch ensures these are properly handled.


spulec:
thanks!

PR #549, opened 2016-02-23
remove NextMarker in response list_functions (Lambda)

beeva-antonioirizar:
I remove NextMarker in response for list_functions (Lambda), because It doesn't have a correctly support to call with markers and it shouldn't send if it doesn't have more functions to response. Reference to documentation: http://docs.aws.amazon.com/cli/latest/reference/lambda/list-functions.html 

Best.

### DIFF
--- a/moto/awslambda/responses.py
+++ b/moto/awslambda/responses.py
@@ -32,7 +32,7 @@ class LambdaResponse(BaseResponse):
         lambda_backend = self.get_lambda_backend(full_url)
         return 200, headers, json.dumps({
             "Functions": [fn.get_configuration() for fn in lambda_backend.list_functions()],
-            "NextMarker": str(uuid.uuid4()),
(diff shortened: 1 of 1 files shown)


hltbra:
@beeva-antonioirizar, the API specifies `NextMarker` and `MaxItems`: http://docs.aws.amazon.com/lambda/latest/dg/API_ListFunctions.html

You can specify a maximum number of functions you want to retrieve, then the `NextMaker` is used to paginate. It might not be present in the CLI page, but it exists in the API.

hltbra:
Sorry @beeva-antonioirizar, I misunderstood your pull request. Yes, all functions are always returned and it doesn't make sense to return `NextMarker` and it will only make sense when `MaxItems` is implemented.

beeva-antonioirizar:
No problem. I didn't explain it right and my english is not so good.

Best.

spulec:
thanks!

PR #550, opened 2016-02-24
Add XML support for cloudformation commands that lacked it

2rs2ts:
This lets boto3's cloudformation API work with moto.

A lot of this work was inspired by https://github.com/spulec/moto/pull/404.

fixes #444 

Signed-off-by: Scott Greene <scott.greene@getbraintree.com>

PR #551, opened 2016-02-24
Pin boto3 and botocore (fix the builds!)

2rs2ts:
botocore 1.3.29 (new as of yesterday) breaks a few of the s3 and
lambda tests; something about a StringIO being closed prematurely.

¯\\\_(ツ)\_/¯

boto3 doesn't pin the botocore version; as a result the new version of
botocore gets pulled in transitively, breaking today's builds.

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -5,6 +5,8 @@ sure>=1.2.24
 coverage
 freezegun
 flask
-boto3>=1.2.3
-botocore>=1.3.26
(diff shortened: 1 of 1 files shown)


2rs2ts:
@spulec My other PR (#550) aside, this seems pretty important as it will unbreak builds.

spulec:
thanks, opened a [new issue](https://github.com/spulec/moto/issues/552) where I'll work on fixing stuff to be able to unpin the version.

PR #553, opened 2016-03-01
Implement NAT Gateways

hltbra:
Can you please review, @spulec? The NatGateway `__init__` method is the most important part IMO because it creates the ENI. From the CreateNatGateway docs:
> This action creates a network interface in the specified subnet with a private IP address from the IP address range of the subnet

References:
* http://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_DescribeNatGateways.html
* http://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_CreateNatGateway.html
* http://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_DeleteNatGateway.html
* http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-ec2-natgateway.html

### DIFF
--- a/moto/ec2/responses/nat_gateways.py
+++ b/moto/ec2/responses/nat_gateways.py
@@ -6,14 +6,14 @@ from moto.ec2.utils import filters_from_querystring
 class NatGateways(BaseResponse):
 
     def create_nat_gateway(self):
-        subnet_id = self.querystring.get('SubnetId')[0]
-        allocation_id = self.querystring.get('AllocationId')[0]
+        subnet_id = self._get_param('SubnetId')
+        allocation_id = self._get_param('AllocationId')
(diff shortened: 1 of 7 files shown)


moto/ec2/responses/nat_gateways.py
```diff
@@ -0,0 +1,73 @@
+from __future__ import unicode_literals
+from moto.core.responses import BaseResponse
+from moto.ec2.utils import filters_from_querystring
+
+
+class NatGateways(BaseResponse):
+
+    def create_nat_gateway(self):
+        subnet_id = self.querystring.get('SubnetId')[0]
```
spulec: The new way of grabbing params is `self._get_param('SubnetId')`

spulec:
LGTM

PR #555, opened 2016-03-01
Fix a typo in DESCRIBE_ROUTE_TABLES_RESPONSE

yannlambret:
There are two 'tagSet' nodes in the DESCRIBE_ROUTE_TABLES_RESPONSE, so responses to 'DescribeRouteTables' requests are buggy.


yannlambret:
After digging a little deeper, it looks like the problem only affects boto3 client. I've just added two tests (boto and boto3) to validate the fix. The "boto" test always passes, the "boto3" test only passes with the fix.

spulec:
Great, thanks!

PR #557, opened 2016-03-02
Allow cloudformation to delete Route 53 RecordSets

2rs2ts:
While there isn't an API method exposed for directly deleting a Route 53 RecordSet [it's performed via `POST` that acts more like a `PATCH` than anything else](http://docs.aws.amazon.com/Route53/latest/APIReference/API_ChangeResourceRecordSets.html), CloudFormation can have templates which contain RecordSets which refer to zones that don't exist inside the template. Ergo, we need a way to effect a delete upon these RecordSets when we don't have a direct reference to the zone.

This exposes a delete method that isn't hooked up to any response (and rightfully so), it just enables the ~polymorphic deletion behavior that we've written into the CloudFormation implementation.

### DIFF
--- a/moto/route53/models.py
+++ b/moto/route53/models.py
@@ -70,6 +70,8 @@ class RecordSet(object):
         self.weight = kwargs.get('Weight')
         self.region = kwargs.get('Region')
         self.health_check = kwargs.get('HealthCheckId')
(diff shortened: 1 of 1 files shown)


spulec:
thanks!

PR #559, opened 2016-03-07
Support HostedZoneId in RecordSet templates

2rs2ts:
Before this commit, using `HostedZoneId` would result in you not being able to find the hosted zone.


spulec:
thanks!

PR #560, opened 2016-03-09
update SES backend to support domain identities and multiple recipients

jstewmon:
includes test cases for ses client with boto3

### DIFF
--- a/moto/ses/exceptions.py
+++ b/moto/ses/exceptions.py
@@ -1,0 +1,10 @@
+from __future__ import unicode_literals
+from moto.core.exceptions import RESTError
+
+
+class MessageRejectedError(RESTError):
+    code = 400
+
+    def __init__(self, message):
+        super(MessageRejectedError, self).__init__(
+            "MessageRejected", message)
--- a/moto/ses/models.py
+++ b/moto/ses/models.py
@@ -1,70 +1,96 @@
 from __future__ import unicode_literals
+
+import email
+
 from moto.core import BaseBackend
+from .exceptions import MessageRejectedError
 from .utils import get_random_message_id
 
 
+RECIPIENT_LIMIT = 50
+
+
 class Message(object):
-    def __init__(self, message_id, source, subject, body, destination):
+    def __init__(self, message_id):
         self.id = message_id
-        self.source = source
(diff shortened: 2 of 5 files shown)


jstewmon:
I can't seem to get the tests passing for python 2.6. On 2.6 only, the parsed email message doesn't return anything for the header fields. I've tinkered with it for a few hours now with no luck getting it to work. I suppose it may be a bug in 2.6. Is there a switch I can use to disable the quota check for send_raw_email tests on python 2.6?

jstewmon:
I finally figured out that PY2 and PY3 expect different types for email.message_from_string, so I was able to get all the tests passing by conditionally decoding the raw data.

spulec:
this is great, thanks!

jstewmon:
no problem. Any idea when the next release might be? I'd like to leverage these changes in another project.

spulec:
Version `0.4.24` is now out

njl:
Removing sent_messages from SESBackend just broke a bunch of my unit tests :/. Was there a reason that was removed?

jstewmon:
@njl sorry about that. I removed it because there is no SES Api for retrieving sent messages, and my supposition was that the only expectation for backends is that they implement the corresponding AWS Api.

Since moto may be used either as a library or as an http service, I thought it made sense that the available features be consistent between the two interfaces.

I'll defer to @spulec for an authoritative opinion on whether or not sent_messages should have been preserved.

I guess you were using the sent_messages list to assert expectations about the message attributes in the list after invoking your code? Another way, albeit less convenient, to make those assertions is to mock the boto method being called (using [mock][mock] or similar) and assert expectations on the passed arguments.

[mock]: https://docs.python.org/dev/library/unittest.mock.html

njl:
Yeah, it's pretty helpful to make sure code sent the messages you expect it sent! When I get some bandwidth, would the be objection to a pull request to add it back in?

spulec:
No objection here

aop:
Why were other data than the message_id removed from the message models? I'd really like to have raw_data in the RawMessage so I can create tests to check that the actual data in the emails are in good shape. Just as in https://github.com/spulec/moto/pull/604 those should be added back in

PR #565, opened 2016-03-15
Propagate stack-level tags to resources

LarsFronius:
According to
http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-resource-tags.html

"All stack-level tags, including automatically created tags, are
propagated to resources that AWS CloudFormation supports. Currently,
tags are not propagated to Amazon EBS volumes that are created from
block device mappings."

I didn't take care of the EBS volume case from block device mappings, but generally fixed that stack level tags are propagated to any taggable ec2 resource.

### DIFF
--- a/moto/cloudformation/models.py
+++ b/moto/cloudformation/models.py
@@ -26,7 +26,7 @@ class FakeStack(object):
         self.output_map = self._create_output_map()
 
     def _create_resource_map(self):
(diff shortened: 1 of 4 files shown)


LarsFronius:
Please don't merge this just yet, I figured that tags with the reserved prefix "aws:" don't count for the limit of 10 tags and thus should not throw a TagLimitExceeded exception. I'll fix this in a bit. :) However, could also turn that into a second PR.

spulec:
thanks!

Issue #164, opened 2014-08-20
Dynamodb2 table query to support QueryFilter


nadersoliman:
A feature request for Dynamodb2 table to support QueryFilter

Thanks,

im-auld:
@spulec What part of the code would need to be patched to make QueryFilters work? I'll see if I can get it working.

spulec:
https://github.com/spulec/moto/blob/master/moto/dynamodb2/models.py#L281

im-auld:
Added a bug fix for this issue, tests are passing: https://travis-ci.org/spulec/moto/builds/224523178

PR #567, opened 2016-03-17
Query filters issue 164

im-auld:
Added support for query filters as requested in issue #164


im-auld:
@spulec I feel like I may be missing some edge cases in this implementation so any feedback would be great. This would be a big feature to add and would probably be useful for a lot of users.

spulec:
Looks great, thanks!

PR #570, opened 2016-03-22
Add ip-address and private-ip-address to the list of filters for instance filtering

digital-abyss:
Please let me know if I'm missing anything. 

These filters are described in
http://boto3.readthedocs.org/en/latest/reference/services/ec2.html#EC2.Client.describe_instances

and map to the instance methods of public_ip() and private_ip() respectively.


spulec:
thanks!

PR #574, opened 2016-03-29
Fixed DynamoDB 'IN' comparison function

DarthLorenzo:
Previously experiencing this error when using the "IN" comparison function:

```
File "[path]/moto/dynamodb2/models.py", line 65, in compare
    return comparison_func(self.value, *range_values)
TypeError: <lambda>() takes exactly 2 arguments (3 given)
```

Modified the lambdas to take an argument list.


spulec:
thanks!

PR #577, opened 2016-04-04
Add datetime to cloudwatch's alarms

beeva-antonioirizar:
it appears in template's response but it didn't have it in model.
Boto3 needs a non-empty date.

Best.


spulec:
thanks!

PR #579, opened 2016-04-05
Added availability-zone filtering to instance collections

roger-lai:
Code Changes:
* Adds the ability to filter by availability zones as described in: http://boto3.readthedocs.org/en/latest/reference/services/ec2.html#EC2.ServiceResource.instances
* maps to placement function which describes the availability zone

Please let me know if anything else can be done.
Thanks!


spulec:
Thanks!

PR #583, opened 2016-04-11
Throw exception if an update is attempted on a stack in ROLLBACK_COMP…

jszwedko:
…LETE

If a stack has a status of ROLLBACK_COMPLETE and you attempt to update
it, the AWS API throws a validation error. This updates moto to have the
same behvaior. We also uncommented a test which tests updating a stack
which passed without any additional modification -- it is unclear why
this test was commented.

Signed-off-by: Jack Lund <jack.lund@getbraintree.com>


spulec:
Merged with 0578f4f041b2ccc6decd447a0d4e14807e72d3e2. Thanks for your work!

PR #585, opened 2016-04-12
Fix multiple bugs encountered with boto3

declanshanaghy:
* Fix path detection for deleting keys in S3 bucket
* Fix stack deletion ensure delete method exists on object
  * Previous tests were using a stack with no resources
* Fix DESCRIBE_STACK_RESOURCES_RESPONSE,
  * Previously untested code path

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,9 +2,11 @@ moto.egg-info/*
 dist/*
 .tox
 .coverage
+cover/
 *.pyc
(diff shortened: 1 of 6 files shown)


declanshanaghy:
What's the ETA on integrating this so we can get back on the main line?  ;-D

spulec:
Thanks!

PR #586, opened 2016-04-13
Filtering implementation for DHCP options sets

yannlambret:
Implement specific filters for DHCP options set objects, and use moto generic_filter's capabilities for 'regular' filters.


spulec:
Thanks!

Issue #575, opened 2016-03-31
cloudformation update_stack response ResponseParserError


aiell0:
This is a follow-up to #550:

The update_stack mocking function was not updated with the fixes made in #550, and as a result, is attempting to parse JSON as XML, obviously returning an error. I don't have the resources to fix this at the moment, but from looking at the rest of the changes made in that ticket, doesn't seem like it would be very tough to fix.

I'll take a crack at it when I get a chance, but that may not be for a couple weeks.

spulec:
Fixed with #590

PR #587, opened 2016-04-15
Fix for cloudformation update_stack response ResponseParserError #575

aiell0:
This is a fix for [#575](https://github.com/spulec/moto/issues/575).

The following changes were made:

* In models.py, added a one-liner that updated the stack status appropriately when the stack is updated.
* In responses.py, modified the update_stack function to fix the ResponseParserError.
* Added some tests for these fixes


aiell0:
Not sure why the Python 2.6 checks fail....I'll look at it in a bit.

aiell0:
Closing so that I can create a fork and branch this, to make it more organized.

PR #588, opened 2016-04-18
Improved the ELBBackend

dhepper:
This change implements subnet and tag handling for the CreateLoadBalancer call.

* when subnets are given, the corresponding VPC is assigned to the ELB
* tags can also be set on creation

Before, subnets and tags passed on creation where silently ignored.


spulec:
Thanks!

PR #589, opened 2016-04-18
Add trailing slash to route53 rrset url path

### LINKED ISSUE #529: botocore.exceptions.ConnectionClosedError raised when calling change_resource_record_sets (Boto3)
I am not sure whether or not I should expect this to work, but I see there are currently similar tests in moto against boto so I thought I would inquire.

When using the Route53 client from boto3, a call to the change_resource_record_set method raises a botocore.exceptions.ConnectionClosedError. (botocore.exceptions.ConnectionClosedError: Connection was closed before we received a valid response from endpoint URL: "https://route53.amazonaws.com/2013-04-01/hostedzone/cc11c883/rrset/")

A test case to reproduce is below.

```python
import boto3
import uuid
from moto import mock_route53

def guid():
    return str(uuid.uuid4())

@mock_route53
def test_route53_rrset_fail():

    client = boto3.client('route53')

    # Create a new zone
    zone_name = '{0}.com'.format(guid())
    zone = client.create_hosted_zone(
        Name=zone_name,
        CallerReference=guid(),
        HostedZoneConfig={'Comment': guid()}
    )
    zone_id = zone['HostedZone']['Id']

    # Verify the zone is retrievable
    z = client.get_hosted_zone(Id=zone_id)
    assert z['HostedZone']['Id'] == zone_id

    # Try to create a record set
    # Raises botocore.exceptions.ConnectionClosedError
    client.change_resource_record_sets(
        HostedZoneId=zone_id,
        ChangeBatch={
            'Comment': guid(),
            'Changes': [{
                'Action': 'CREATE',
                'ResourceRecordSet': {
                    'Name': 'foo.{0}'.format(zone_name),
                    'Type': 'A',
                    'ResourceRecords': [{'Value': '1.2.3.4'}]
                }
            }]
        }
    )
```

leg100:
Fixes #529. The [API docs](http://docs.aws.amazon.com/Route53/latest/APIReference/API_ChangeResourceRecordSets.html) make no mention of a trailing slash but adding one seems to make it work.

### DIFF
--- a/moto/route53/urls.py
+++ b/moto/route53/urls.py
@@ -8,7 +8,7 @@ url_bases = [
 url_paths = {
     '{0}hostedzone$': responses.list_or_create_hostzone_response,
     '{0}hostedzone/[^/]+$': responses.get_or_delete_hostzone_response,
-    '{0}hostedzone/[^/]+/rrset$': responses.rrset_response,
+    '{0}hostedzone/[^/]+/rrset/?$': responses.rrset_response,
     '{0}healthcheck': responses.health_check_response,
     '{0}tags|trafficpolicyinstances/*': responses.not_implemented_response,
 }


spulec:
Thanks!

PR #590, opened 2016-04-18
Fixes issue #575: cloudformation update_stack response ResponseParser…

aiell0:
Changes in this release:

* Added a one-liner in cloudformation/models.py to update the status of the stack correctly (UPDATE_COMPLETE when it is updated.).
* Added response template in cloudformation/responses.py to return the correct template when the user requests JSON.
* Added tests to test the new feature.


spulec:
Thank you!

Issue #581, opened 2016-04-06
Filter dicts have not been implemented in Moto for '%s' yet - mock_ec2


aiell0:
I am attempting to use the describe_instances function in boto3. I am supplying a filter that returns a string, and am receiving the following message:

NotImplementedError: (u"Filter dicts have not been implemented in Moto for '%s' yet. Feel free to open an issue at https://github.com/spulec/moto/issues", u'Architecture')

So I am following the suggestion and opening an issue. I will take a crack at this if I have time, but figured the community should know this is not supported at the current moment.

spulec:
Fixed with #592

spulec:
Can you try using `architecture` with a lowercase `a`? Does an upper-case `a` work with the Amazon API?

PR #592, opened 2016-04-18
Added ec2 describe-instances filter: 'architecture'

aiell0:
This is a fix for issue [#581](https://github.com/spulec/moto/issues/581).

* Modified moto/ec2/utils.py to add 'architecture' to the filter dict, as well as change the exception message to format correctly.
* Added a test for filtering by architecture in tests/test_ec2/test_instances.py

There are still plenty more filters that need to be added, but this was the one that our team needed. If we need more in the future, I will add them.


spulec:
Thanks!

PR #594, opened 2016-04-20
Fix security groups

yannlambret:
The boto documentation is pretty vague about egress rules management, but according to boto3 documentation, it looks pretty clear that the same logic should apply for both ingress and egress rules, because the API expects the same payloads. See:

http://boto3.readthedocs.org/en/latest/reference/services/ec2.html#EC2.SecurityGroup.authorize_egress
http://boto3.readthedocs.org/en/latest/reference/services/ec2.html?#EC2.SecurityGroup.revoke_egress

and:

http://boto3.readthedocs.org/en/latest/reference/services/ec2.html?#EC2.SecurityGroup.authorize_ingress
http://boto3.readthedocs.org/en/latest/reference/services/ec2.html?#EC2.SecurityGroup.authorize_ingress

These modifications make it possible to authorize or revoke an outbound rule for another security group.

It also add the default outbound rule for a security group.

### DIFF
--- a/moto/ec2/exceptions.py
+++ b/moto/ec2/exceptions.py
@@ -136,7 +136,7 @@ class InvalidPermissionNotFoundError(EC2ClientError):
     def __init__(self):
         super(InvalidPermissionNotFoundError, self).__init__(
             "InvalidPermission.NotFound",
-            "Could not find a matching ingress rule")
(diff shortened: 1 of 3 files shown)


spulec:
Thank you!

PR #597, opened 2016-04-26
Fix security group ingress authorization for all protocols with no port spec

romanek-adam:


spulec:
thanks!

romanek-adam:
I'm glad I could help.

PR #599, opened 2016-04-28
Add support for src_version_id to bucket copy_key op.

gwax:
This fixes an issue where copy operations fail when the source object is versioned.


spulec:
Awesome, thank you!

PR #600, opened 2016-04-28
Add S3 versioned bucket support to key head requests.

d-felix:
This PR adds support for passing version IDs in key HEAD requests. This fixes an issue where the boto3 client head_object method would only return information about the current object version.


spulec:
Great, thank you!

PR #602, opened 2016-05-02
Implement support for sns delete_endpoint()

mvantellingen:
Adds missing implementation for deleting endpoints in sns

Thanks for this library btw, cool stuff 👍


spulec:
Awesome, thank you!

PR #608, opened 2016-05-07
Extend filtering support for elastic network interfaces

yannlambret:
Add support for tag filtering, as well as for the following filters:

* addresses.private-ip-address
* private-ip-address
* subnet-id
* vpc-id

Also check if the interface is associated to return the correct status ('available' vs. 'in-use') in the DESCRIBE_NETWORK_INTERFACES_RESPONSE template, which is very useful when testing code that uses boto3 waiters.


spulec:
Awesome, thank you!

PR #617, opened 2016-05-15
Disallow creation of ELBs with duplicate names.

mrucci:
As the AWS API does: http://docs.aws.amazon.com/ElasticLoadBalancing/latest/APIReference/API_CreateLoadBalancer.html

This commit also makes sure that the ELB dns_name is always unique.


spulec:
Thanks!

PR #621, opened 2016-05-31
Respect VisibilityTimeout when calling receive_message

mattlong:
Previously, receive_message would always use the queue's default
visibility timeout instead of the value passed as a query parameter when
calling the receive_message method on an SQS connection.


spulec:
thanks!

PR #624, opened 2016-06-02
Save the sns messages in the PlatformEndpoint

mvantellingen:
This makes it easier to test the correct sns messages are sent.

For example in my test cases I can now do:

```python
    from moto.sns.models import sns_backends
    messages = sns_backends['eu-west-1'].platform_endpoints.values()[0].messages
    
    # check number of messages
    assert len(messages) == 1
    assert 'Hello' in messages.values()[0]
```

### DIFF
--- a/moto/sns/models.py
+++ b/moto/sns/models.py
@@ -119,6 +119,7 @@ class PlatformEndpoint(object):
         self.token = token
         self.attributes = attributes
         self.id = uuid.uuid4()
+        self.messages = OrderedDict()
 
     @property
     def arn(self):
@@ -130,8 +131,9 @@ class PlatformEndpoint(object):
         )
 
     def publish(self, message):
-        message_id = six.text_type(uuid.uuid4())
         # This is where we would actually send a message
+        message_id = six.text_type(uuid.uuid4())
+        self.messages[message_id] = message
         return message_id
 
 


spulec:
thanks!

tzusman:
@mvantellingen Is it possible to read these messages when moto is in server mode?

bblommers:
Hi @tzusman, it is of course possible to use the `receive_message` to verify whether a message has been sent.
Directly accessing the backend to read the messages is not possible at the moment, though.

Slightly related feature request: https://github.com/spulec/moto/issues/2882

tzusman:
Hey @bblommers thanks for getting back to me.  I just wanted to confirm that you're talking about using SQS's receive_message to get these push notification payloads?  It doesn't look like the messages are being published anywhere, but rather just [stored in the Dict](https://github.com/spulec/moto/blob/master/moto/sns/models.py#L346)

bblommers:
Ah, my bad @tzusman - I misread it as SQS. You're right, in that case, the SNS messages are not published anywhere.

tzusman:
I had to hack it (now it sends an http request whenever a message is sent to a PlatformEndpoint).  Obviously this is less than ideal.  @bblommers what do you think would be the least dirty way of pulling that `messages` data out in server mode?

bblommers:
@tzusman Probably a custom HTTP endpoint in the MotoServer to download the internal state, with filters to retrieve the state of a specific service. 
That would be the most generic solution, and usable for any language/platform/service.

PR #625, opened 2016-06-02
Strip extra whitespace and newlines from SQS response templates.

### LINKED ISSUE #623: Whitespace and newlines in some message responses fields
Due to the way the response templates are declared, some response fields get extra whitespace/newlines which should not be there.

Example response for SQS create_queue:

`{u'MD5OfMessageBody': '\n            58fd9edd83341c29f1aebba81c31e257\n        ', 'ResponseMetadata': {'HTTPStatusCode': 200, 'RequestId': '\n            27daac76-34dd-47df-bd01-1f6e873584a0\n        '}, u'MessageId': '\n            a915de12-97f9-cf50-edf3-43b1ff062927\n        '}`

The fix easy though,
```
<RequestId>7a62c49f-347e-4fc4-9331-6e8e7a96aa73</RequestId>
```
instead of
```
<RequestId>
    7a62c49f-347e-4fc4-9331-6e8e7a96aa73
</RequestId>
````

nfvs:
Whitespace and newlines in the template strings are significant, and it seems boto is usually not expecting it. This leads to things like `'RequestId': '\n 27daac76-34dd-47df-bd01-1f6e873584a0\n '`.

The solution is to remove newlines/whitespace completely when not using variables (i.e. RequestId blocks), or use `{{-` and `-}}` when using variables in a new line, so that Jinja2 ignores them.

I tried a generic solution in `core/responses.py`, both using `replace()` and regexes, but none of them satisfied all tests, so perhaps the best solution is to fix all template strings instead.

Fixes #623

### DIFF
--- a/moto/core/responses.py
+++ b/moto/core/responses.py
@@ -81,7 +81,8 @@ class _TemplateEnvironmentMixin(object):
         template_id = id(source)
         if not self.contains_template(template_id):
             self.loader.update({template_id: source})
-            self.environment = Environment(loader=self.loader, autoescape=self.should_autoescape)
+            self.environment = Environment(loader=self.loader, autoescape=self.should_autoescape, trim_blocks=True,
+                                           lstrip_blocks=True)
         return self.environment.get_template(template_id)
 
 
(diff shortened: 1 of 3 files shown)


spulec:
great, thank you!

PR #631, opened 2016-06-07
Route53 private zone support

hummus:
support for `HostedZone.Config.PrivateZone=True`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 moto.egg-info/*
 dist/*
+.cache
 .tox
 .coverage
 cover/
--- a/moto/route53/models.py
+++ b/moto/route53/models.py
@@ -147,12 +147,12 @@ class RecordSet(object):
 
 class FakeZone(object):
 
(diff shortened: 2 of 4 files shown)


moto/emr/__init__.py
```diff
@@ -6,7 +6,4 @@
 
 
 def mock_emr(func=None):
-    if func:
-        return MockAWS(emr_backends)(func)
-    else:
-        return MockAWS(emr_backends)
+    return MockAWS(emr_backends)
```
spulec: Why this change?

spulec:
Thanks for this. One question, but otherwise happy to merge

moto/emr/__init__.py
hummus: Oh my. Sorry, this disabled the emr testing while I somehow thought it had just fixed my issue where `test_emr.test_create_job_flow_in_multiple_regions` would not pass. 

passes in CI now though I can't get it pass locally....

spulec:
Great, thanks!

PR #632, opened 2016-06-09
Updating RDS instance status after deleting it

jcastillocano:
In AWS RDS API, after deleting a RDS instance the status of the instance should be 'deleting' (http://docs.aws.amazon.com/AmazonRDS/latest/APIReference/API_DeleteDBInstance.html)

Adding database.status = 'deleting' will create the right response using the DELETE_DATABASE_TEMPLATE since the main block is the xml rep of the db instance.


spulec:
If you can un-bump the version, I'll be happy to merge

spulec:
thanks!

PR #633, opened 2016-06-09
KMS Encrypt/Decrypt/GenerateDataKey actions

austinmoore-:
This PR adds some basic support for KMS Encrypt, Decrypt, and GenerateDataKey actions. The functions do not check if the `key_usage` property of the keys, so that might be something someone wants to do down the road. Also, the NumberOfBytes property for GenerateDataKey is not implemented.


spulec:
Thanks for this.

Looks like Travis is failing, but would be happy to get this merged if we can fix.

JackDanger:
@austinmoore- I'm closing this PR due to its age. Please reopen if you can. Some of this has already been implemented but the `EncryptionContext` hasn't been. It would be super helpful if you could rebase this on latest master.

PR #634, opened 2016-06-09
Fix cfn error message and delete stack

hiteshghia-splunk:
- Fix typo in exception message for cloudformation.
- Fix delete stack functionality for stacks with resources for which 'delete' attribute is not defined.


spulec:
Great, thanks!

PR #642, opened 2016-06-21
Fix update item on dynamodb2

victorblaga:
Added the possibility to call dynamodb_v2 by supplying an additional ExpressionAttributeNames and ExpressionAttributeValues dictionaries.

Also fixed some wrong tests related to the update_item functionality.


spulec:
thanks

Issue #629, opened 2016-06-06
Lambda invoke function is not mocked


mikegrima:
Hello:

I am attempting to mock out a call to the AWS Lambda `invoke()` function.  Instead of hitting Moto, it appears to be reaching out to AWS directly and failing as a result:

```
self = <botocore.retryhandler.ExceptionRaiser object at 0x10349b090>
attempt_number = 5
caught_exception = ConnectionClosedError(u'Connection was closed before we received a valid respo...lambda.us-west-2.amazonaws.com/2015-03-31/functions/MYFUNCTION/invocations".',)

    def _check_caught_exception(self, attempt_number, caught_exception):
        # This is implementation specific, but this class is useful by
        # coordinating with the MaxAttemptsDecorator.
        # The MaxAttemptsDecorator has a list of exceptions it should catch
        # and retry, but something needs to come along and actually raise the
        # caught_exception.  That's what this class is being used for.  If
        # the MaxAttemptsDecorator is not interested in retrying the exception
        # then this exception just propogates out past the retry code.
>       raise caught_exception
E       ConnectionClosedError: Connection was closed before we received a valid response from endpoint URL: "https://lambda.us-west-2.amazonaws.com/2015-03-31/functions/MYFUNCTION/invocations".
```

I was looking at the moto code for Lambda, and wasn't quite sure where the issue lies or how to fix it.  I recall making a PR in the past, and for that PR, I recall adding in the XML response returned from AWS.  Things appear to be very different with the newer AWS APIs, so I'm not entirely sure how to implement moto fixes now...

I was looking to see if adding an `invoke` method to https://github.com/spulec/moto/blob/master/moto/awslambda/models.py would work, but it doesn't seem to get executed.

Do you know what may be causing the `invoke` method to get skipped over, and where to place the code to fix it?

Thank You

spulec:
It looks like invocation hasn't been added yet.

We would need to add the [url endpoint](https://github.com/spulec/moto/blob/master/moto/awslambda/urls.py), map it to a new [response]( https://github.com/spulec/moto/blob/master/moto/awslambda/responses.py), and add a method on the [lambda function model](https://github.com/spulec/moto/blob/master/moto/awslambda/models.py) to execute.

I probably won't have time for this in the near future, but would be happy to accept a PR if anyone has time.

mikegrima:
Hi @spulec I wouldn't mind submitting a PR for now.  Thank you for providing some insight on how it works.

[...]

PR #644, opened 2016-06-22
Added Lambda invoke capability.

mikegrima:
Addresses #629

### DIFF
--- a/moto/awslambda/models.py
+++ b/moto/awslambda/models.py
@@ -3,6 +3,7 @@ from __future__ import unicode_literals
 import base64
 import datetime
 import hashlib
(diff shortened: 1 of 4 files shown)


mikegrima:
`test_lambda.test_invoke_function ... ok`

My test passes... Not sure why all the others are failing...

2rs2ts:
@mikegrima It's not your fault - botocore isn't pinned to an exact revision in boto3, so new botocore releases sometimes break boto3. I've already had to address it once in this repo before, this makes it the second time. #654 should fix your build.

spulec:
This is great, thanks!

PR #645, opened 2016-06-23
Support Filters in describe_security_groups() for boto3

paulci:
Filters in boto3 are passed in a list, which means any comparisons will currently fail.  Update to model and tests to resolve this issue.


spulec:
thanks!

kenden:
Thanks for fixing this issue.
This is blocking us, would you know when it is going to be released?

Issue #628, opened 2016-06-04
NotImplementedError: The describe_stack_events action has not been implemented


gauravve:
I am trying to test the fix for #595 but getting a new error much earlier in the moto code. The error is shown below. 

```python
File "/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/httplib.py", line 390, in __init__
    self.fp = sock.makefile('rb', 0)
  File "/Users/gauravverma/.virtualenvs/moto-mocks/lib/python2.7/site-packages/httpretty/core.py", line 332, in makefile
    self._entry.fill_filekind(self.fd)
  File "/Users/gauravverma/.virtualenvs/moto-mocks/lib/python2.7/site-packages/httpretty/core.py", line 591, in fill_filekind
    status, headers, self.body = self.callable_body(self.request, self.info.full_url(), headers)
  File "/Users/gauravverma/.virtualenvs/moto-mocks/lib/python2.7/site-packages/moto/core/responses.py", line 95, in dispatch
    return cls()._dispatch(*args, **kwargs)
  File "/Users/gauravverma/.virtualenvs/moto-mocks/lib/python2.7/site-packages/moto/core/responses.py", line 138, in _dispatch
    return self.call_action()
  File "/Users/gauravverma/.virtualenvs/moto-mocks/lib/python2.7/site-packages/moto/core/responses.py", line 164, in call_action
    raise NotImplementedError("The {0} action has not been implemented".format(action))
NotImplementedError: The describe_stack_events action has not been implemented
```

Wondering where is describe_stack_events not implemented? Earlier when I raised #595 I did not get this error.

spulec:
That endpoint hasn't been implemented yet. Happy to accept a pull request if anyone gets some time.

spulec:
Closed with #653

PR #653, opened 2016-06-29
Add CloudFormation:DescribeStackEvents

2rs2ts:
This only adds the events for the stack itself, but it's possible to add events for the other resources (I just didn't have a need to implement events for any of them.)


2rs2ts:
This will close #628 incidentally!

spulec:
thanks!

PR #654, opened 2016-06-29
Pin boto3 and botocore (again)

2rs2ts:
Builds started failing 7 days ago. Turns out it was because of the bump from 1.4.28 to 1.4.29. It may not be worth playing whac-a-mole - I'll file an issue in either the boto3 or botocore repo in the meantime.

Affected PRs:

* #644 
* #645 
* #653

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -5,8 +5,21 @@ sure>=1.2.24
 coverage
 freezegun
 flask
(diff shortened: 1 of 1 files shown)


2rs2ts:
It's funny how just changing a dependency version can change the code path tests take...

spulec:
thanks!

PR #655, opened 2016-06-29
Remove HTTPHeaders from ResponseMetadata in some tests

2rs2ts:
botocore 1.4.29 started adding this. I'd match against it, but it contains a date object and it's just not worth rewriting the assertions for a key-by-key approach.

If this is acceptable it will eliminate the need for #654

### DIFF
--- a/tests/test_apigateway/test_apigateway.py
+++ b/tests/test_apigateway/test_apigateway.py
@@ -76,6 +76,7 @@ def test_create_resource():
         restApiId=api_id,
         resourceId=root_id,
     )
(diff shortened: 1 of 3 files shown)


spulec:
Thanks for working on this, but I would rather we change the tests. I want Moto to resemble the AWS API as closely as possible.

2rs2ts:
@spulec those headers are an artifact of botocore, not moto. It's not something the API returns and not something you'd have to worry about emulating. You could change the tests but you'd have to pull some crazy monkey patching or find a fuzzy matcher for `datetime`s.

spulec:
Oh, okay. I see.

spulec:
thanks!

PR #659, opened 2016-07-05
Fix tag filter

kierandoonan:
Currently the tag filter fails when describing resources and using wildcards.

e.g. an instance with the tags ("Key1", "Value1") wouldn't get picked up with a filter of {'tag:Key1': 'Value*'}


spulec:
thanks!

PR #661, opened 2016-07-08
Update format of S3 client errors

jszwedko:
To match
http://docs.aws.amazon.com/AmazonS3/latest/API/ErrorResponses.html#RESTErrorResponses
which documents that it should be at the top level rather than nested
under `Errors`.


spulec:
thanks!

Issue #330, opened 2015-03-12
moto vs flask/socketserver


memory:
I was _delighted_ to discover moto, but am running into some ugly issues trying to use it for our project, which is a Flask-restful application backed by S3 and DynamoDB.

Moto very successfully mocks out the boto.s3 and dynamodb calls when running unit tests on individual functions, but if we try to test the actual API endpoints of the app while using the mocked backends, terrible things happen:

```
Exception in thread Thread-1:
Traceback (most recent call last):
  File "/usr/lib/python2.7/threading.py", line 810, in __bootstrap_inner
    self.run()
  File "/usr/lib/python2.7/threading.py", line 763, in run
    self.__target(*self.__args, **self.__kwargs)
  File "/vagrant/target/python/local/lib/python2.7/site-packages/werkzeug/serving.py", line 602, in inner
    passthrough_errors, ssl_context).serve_forever()
  File "/vagrant/target/python/local/lib/python2.7/site-packages/werkzeug/serving.py", line 462, in serve_forever
    HTTPServer.serve_forever(self)
  File "/usr/lib/python2.7/SocketServer.py", line 238, in serve_forever
    self._handle_request_noblock()
  File "/usr/lib/python2.7/SocketServer.py", line 290, in _handle_request_noblock
    request, client_address = self.get_request()
  File "/vagrant/target/python/local/lib/python2.7/site-packages/werkzeug/serving.py", line 475, in get_request
    con, info = self.socket.accept()
  File "/usr/lib/python2.7/socket.py", line 203, in accept
    return _socketobject(_sock=sock), addr
TypeError: __init__() got an unexpected keyword argument '_sock'
```

(platform: ubuntu trusty, python 2.7.6)

Are we swimming against the tide by trying to do this all in-app?  We could set up a pair of moto-servers, but that seems like it would make for a much more complicated and fragile test harness.

spulec:
It certainly should not be swimming against the tide to do this.

It looks like HTTPretty is interfering with Werkzeug. HTTPretty patches `_socketobject`, but it appears that the object it is using to patch does not have the same `__init__` arguments as the real `_socketobject`.

If I get some time later in the week, I'll try to take a deeper look. If you want to take a shot, I would try adding a `_sock` argument to the HTTPretty [`fakesock.socket.__init__`](https://github.com/gabrielfalcao/HTTPretty/blob/691865d48bf322111a2f80731563b3b05ce76318/httpretty/core.py#L251)

memory:
[...]

PR #664, opened 2016-07-18
Bump up httpretty version to 0.8.14

maurodoglio:
As discussed in issue #330 this is required to avoid problems with 3rd party libraries

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -6,7 +6,7 @@ install_requires = [
     "Jinja2",
     "boto>=2.36.0",
     "flask",
-    "httpretty==0.8.10",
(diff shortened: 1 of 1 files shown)


spulec:
Thanks for opening.

You can feel free to manually override this requirement yourself, but newer versions of HTTPretty are not necessarily going to [work with Python3](https://github.com/gabrielfalcao/HTTPretty#disclaimer). For that reason, I would prefer to not change the default (as can be seen with some of the failing tests).

I understand that this causes other things to break, but it is the best compromise I can see. I'm open to other suggestions.

PR #670, opened 2016-07-22
Add volume attachment status

andrewgross:
Adds volume attachment status for filtering. 

Current possible statuses are `attached` or `detached`, the intermediate states (`attaching` and `detaching`) are not implemented yet, because I couldn't figure out a good implementation.  The `detached` status is tricky to test, since you have to keep the `VolumeAttachment` from `detach_volume`.


spulec:
thanks!

PR #676, opened 2016-08-10
Add support for requestTemplates parameter in apigateway integrations mock

whummer:
This PR adds support for the requestTemplates=... parameter for the put_integration method in apigateway mock. See http://boto3.readthedocs.io/en/latest/reference/services/apigateway.html#APIGateway.Client.put_integration

This is a non-breaking change. New test cases have been added to ensure we keep backward compatibility.

@spulec Thanks for maintaining this awesome project! :)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -11,6 +11,8 @@ install_requires = [
     "xmltodict",
     "six",
     "werkzeug",
+    "sure",
+    "freezegun"
 ]
 
 extras_require = {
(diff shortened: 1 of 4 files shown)


setup.py
```diff
@@ -11,6 +11,8 @@
     "xmltodict",
     "six",
     "werkzeug",
+    "sure",
```
whummer: I had to add these two lines, as these packages were not installed on my machine. To avoid relying on global packages, maybe we should consider setting up a Python virtualenv at some point? http://docs.python-guide.org/en/latest/dev/virtualenvs/

spulec:
Thanks!

PR #678, opened 2016-08-14
additional filters to get_all_vpcs

rocky4570:
Additional filters on get_vpcs (isDefault and state)

First VPC created by moto is the default VPC. Was required by user to remove default VPCs prior to rebuilding.

state is difficult to filter in moto as state is only pending on initial create; subsequent describe results is available. However this may show up when creating a VPC when running live (loop until a vpc is available for instance). Unit tests for this are lame as not sure how to test for pending state.


spulec:
Thanks!

PR #679, opened 2016-08-15
Causes get_server_certificate to fail correctly

phobologic:
When no certificate with the name exists, the API should return a 404
(NoSuchEntity).

### DIFF
--- a/moto/iam/models.py
+++ b/moto/iam/models.py
@@ -318,6 +318,10 @@ class IAMBackend(BaseBackend):
             if name == cert.cert_name:
                 return cert
 
+        raise IAMNotFoundException(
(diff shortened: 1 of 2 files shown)


spulec:
Thanks!

phobologic:
NP, thanks for the tool!  One question, if you have a moment - any idea why the coverage decreased when I added a test?

spulec:
No idea. There is certainly no way it fell by 2%. Probably a bug.

Thanks for your work on troposphere! We use it significantly at my work.

phobologic:
Awesome!  Glad you guys like it - I'm using moto with stacker now, looking forward to adding more to it, so thank you! :)

PR #681, opened 2016-08-17
Use region from ctor to set queue_arn.

ronin13:
Without this, any region passed to boto3.resource is ignored and unit
tests depending on it break.


spulec:
Thanks!

PR #682, opened 2016-08-22
Initial support for apigateway stages

whummer:
This PR adds initial support for deployment stages for the apigateway service.

When a new deployment is created, a new stage is automatically created for the `stageName` that has been specified for this deployment. Using the `update_stage` operation (which is an HTTP `PATCH` method), the stage information can be modified. Currently this works only for very simple examples, and the `"/*/*/<fieldName>"` path syntax is not supported yet. Will be added as part of a future PR.

Additional tests have been added to cover the new functionality.

### DIFF
--- a/moto/apigateway/models.py
+++ b/moto/apigateway/models.py
@@ -149,6 +149,26 @@ class Resource(object):
         return self.resource_methods[method_type].pop('methodIntegration')
 
 
+class Stage(dict):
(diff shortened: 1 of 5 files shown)


spulec:
This is great. Thank you!

whummer:
What is the current release cadence, when can we expect to get this released? Thanks!

spulec:
I just released version 0.4.27

PR #689, opened 2016-09-01
Make testing a bit safer binding to 127.0.0.1

michael-dev2rights:
Please accept this patch which changes to bind by default to 127.0.0.1 so that mistakes during testing are less likely to cause problems.  I've updated also the documentation and this passes tests on travis.  

Thanks for moto in any case.  Very nice.


spulec:
Good catch, thanks!

PR #692, opened 2016-09-03
Fix unicode S3 key listing in Python 2

imran2140:
Unicode keys returned by `bucket.list()` have malformed name in Python 2.

Using separating key name unquoting logic for Python 2 fixes the problem (Call `unquote` with `str`, then decode to `unicode` again). Updated the test case for unicode key to reproduce the error (without the fix).


spulec:
thanks!

PR #693, opened 2016-09-07
ALWAYS use utcnow(), never now()

collinstocks:
Moto only returns the right CreationDateTime for DynamoDB tables if the current timezone is UTC. This PR fixes that behavior.


spulec:
Thanks!

PR #698, opened 2016-09-13
Stop stripping the trailing slash off keys listed from s3 buckets

### LINKED ISSUE #684: Unable to create a key with a trailing slash using OrdinaryCallingFormat
When using OrdinaryCallingFormat, it's not possible to create a key ending with a slash (e.g. when mimicking directory creation), since this is stripped off when parsing the key name. I can't comment on S3, but this is at least different behaviour from Ceph.

For example, the below fails as is, but works if the connection uses SubdomainCallingFormat instead.

```
import boto
import moto
import unittest


class TestCreatingKeyEndingWithSlash(unittest.TestCase):

    @moto.mock_s3
    def test_ordinary_calling_format(self):
        bucket_name = 'testbucket'
        key_name = 'key_ending_with_slash/'

        conn = boto.connect_s3('access_key', 'secret_key',
                               calling_format=boto.s3.connection.OrdinaryCallingFormat())
        bucket = conn.create_bucket(bucket_name)

        key = boto.s3.key.Key(bucket)
        key.key = key_name
        key.set_contents_from_string('')

        self.assertIn(key_name, [k.name for k in bucket.get_all_keys()])
```

SpoonMeiser:
fixes #684

### DIFF
--- a/moto/s3bucket_path/utils.py
+++ b/moto/s3bucket_path/utils.py
@@ -12,7 +12,7 @@ def bucket_name_from_url(url):
 
 
 def parse_key_name(path):
-    return "/".join(path.rstrip("/").split("/")[2:])
+    return "/".join(path.split("/")[2:])
 
 
 def is_delete_keys(request, path, bucket_name):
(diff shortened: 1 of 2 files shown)


spulec:
Thanks!

PR #699, opened 2016-09-13
added more api gateway coverage

tootedom:
Hi there,

Firstly, thank you very much for all the work on moto.  Great library.

I've added extra support for apigateway, as a result of developing ansible modules using boto3 from provisioning api gateways.  This pull request covers the features we are intending to use at work, so I built the additional mocking around the testing of our ansible modules.

The features added are:

- get_stages, 
- stage not found exception 
- update stage configuration
- descriptions on deployments
- creation date on deployments
- setting stage variables on deployments and stage creation

Hopefully the changes make sense.  Please let me know if you need us to make any changes.
thanks
/dom

### DIFF
--- a/moto/apigateway/models.py
+++ b/moto/apigateway/models.py
@@ -155,8 +155,10 @@ class Resource(object):
 class Stage(dict):
 
 
-    def __init__(self, name=None, deployment_id=None, variables={}):
+    def __init__(self, name=None, deployment_id=None, variables=None):
         super(Stage, self).__init__()
+        if variables is None:
+            variables = {}
         self['stageName'] = name
         self['deploymentId'] = deployment_id
         self['methodSettings'] = {}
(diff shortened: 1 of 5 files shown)


moto/apigateway/models.py
```diff
@@ -150,24 +153,124 @@ def delete_integration(self, method_type):
 
 
 class Stage(dict):
-    def __init__(self, name=None, deployment_id=None):
+
+
+    def __init__(self, name=None, deployment_id=None, variables={}):
```
spulec: Generally not a good idea to have mutable things as default arguments. See [here](http://docs.python-guide.org/en/latest/writing/gotchas/#mutable-default-arguments) for why. It is one of the bigger gotchas in python IMO. It looks like you have a similar thing in a couple other places too.

spulec:
This is great!

I had one comment about mutable things as default arguments, but would be happy to merge if we can get that cleaned up.

tootedom:
>>
>> Generally not a good idea to have mutable things as default arguments
>>

You learn something new everyday.  Thanks for that.  
Think I've updated them all.

cheers
/dom

tootedom:
Hi there,

I just noticed a bug in the former commit that was always setting 'cacheClusterSize'.  It appears this does not come back on a newly created stage, until the caching is enabled.  So I've catered for this by setting 'cacheClusterSize' to 0.5 if cacheClusterEnabled is set to true (and it's not previously defined).  At the same time the cache cluster size is a string (no idea why - seems that way returned from apigw).

I've also whilst fixing the above edge case.  Extended create stage to allow for taking:

- description, cacheClusterEnabled and cacheClusterSize.


Sorry for the update.  Only, just noticed this edge case this morning.

/dom

tootedom:
Hi there,

I'm also a moron, and typo'd the NotFoundException as NonFoundException.  I've fixed this.

Apologies
/dom

spulec:
Perfect, thanks!

PR #700, opened 2016-09-15
Invalid resource types would return all resources if no other filters were included

monkeyrevenge:
If the resource-type in describe_tags was invalid it would skip adding that filter. 

This behavior differed from that of AWS behavior which would instead return 0 results.

### DIFF
--- a/moto/ec2/models.py
+++ b/moto/ec2/models.py
@@ -846,8 +846,7 @@ class TagBackend(object):
                             resource_id_filters.append(re.compile(simple_aws_filter_to_re(value)))
                     if tag_filter == 'resource-type':
                         for value in filters[tag_filter]:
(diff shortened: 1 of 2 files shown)


monkeyrevenge:
This pull request will fix #701

spulec:
Can you revert the change to the version number? If yes, I would be happy to merge

monkeyrevenge:
Thank you.  Secondarily I noticed that tags are not cleared between mocks. For instance in some unit tests that use tags each test cumulatively adds more tags, Is this intentional?  If not I would open an issue for that as well and a fix.

monkeyrevenge:
That other bug I mentioned is not a bug...just me messing up in my tests.  Sorry about that.

spulec:
thanks!

PR #704, opened 2016-09-19
Fix an S3 multipart upload bug

### LINKED ISSUE #703: S3 multipart upload failure due to ETag mismatch with boto3
A basic multipart upload on S3 with a script as follows work on unmocked S3:

```
import logging

import boto3
import moto


logging.basicConfig(level='WARNING')


def multipart_upload():
    bucket = 'mytestbucket'
    key = 'uploaded'

    s3 = boto3.client('s3')

    upload_id = s3.create_multipart_upload(
        Bucket=bucket, Key=key)['UploadId']
    try:
        parts = []

        parts.append(
            s3.upload_part(
                Bucket=bucket,
                Key=key,
                PartNumber=1,
                UploadId=upload_id,
                Body='somebody')['ETag'][1:-1])

        s3.complete_multipart_upload(
            Bucket=bucket, Key=key, UploadId=upload_id, MultipartUpload={
                'Parts': [{'ETag': etag, 'PartNumber': i}
                          for i, etag in enumerate(parts, 1)]})
    except Exception:
        s3.abort_multipart_upload(
            Bucket=bucket, Key=key, UploadId=upload_id)
        raise


@moto.mock_s3
def test_upload():
    s3 = boto3.client('s3')
    s3.create_bucket(Bucket='mytestbucket')
    multipart_upload()


if __name__ == '__main__':
    multipart_upload()
```

but the test with mock fails with a ClientError:

```
ClientError: An error occurred (InvalidPart) when calling the CompleteMultipartUpload operation: One or more of the specified parts could not be found. The part might not have been uploaded, or the specified entity tag might not have matched the part's entity tag.
```

This appears to be due to unstripped double quotes in ETag in the S3 model code.

okomestudio:
Fixes #703.

### DIFF
--- a/moto/s3/models.py
+++ b/moto/s3/models.py
@@ -7,12 +7,17 @@ import copy
 import itertools
 import codecs
 import six
+import logging
 
 from bisect import insort
 from moto.core import BaseBackend
 from moto.core.utils import iso_8601_datetime_with_milliseconds, rfc_1123_datetime
 from .exceptions import BucketAlreadyExists, MissingBucket, MissingKey, InvalidPart, EntityTooSmall
 from .utils import clean_key_name, _VersionedKeyStore
+
+
+log = logging.getLogger(__name__)
+
 
 UPLOAD_ID_BYTES = 43
 UPLOAD_PART_MIN_SIZE = 5242880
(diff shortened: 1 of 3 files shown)


okomestudio:
Upon further investigation, it looks like the issue happens only with `boto3` and not with `boto`, and the change in this PR breaks the `boto` version. I'll either try to make this work with `boto3` or discard the PR.

PR #706, opened 2016-09-21
Add support and unittests for Route53 tags

kryptek:
This commit adds support for Route53 tagging. Unit tests included.


spulec:
Awesome, thank you!

PR #708, opened 2016-09-22
Parse JSON request body from boto3

### LINKED ISSUE #707: EMR run_job_flow does not retain cluster info when used with boto3
The simplest way to reproduce the issue is to add the following line to one of the existing tests, `test_emr/test_emr_boto3.py:test_list_clusters`:

```python
@mock_emr
def test_list_clusters():
    client = boto3.client('emr', region_name='us-east-1')
    client.run_job_flow(
        Name='cluster',
        Instances={
            'MasterInstanceType': 'c3.xlarge',
            'SlaveInstanceType': 'c3.xlarge',
            'InstanceCount': 3,
            'Placement': {'AvailabilityZone': 'us-east-1a'},
            'KeepJobFlowAliveWhenNoSteps': True,
        },
        VisibleToAllUsers=True,
    )
    summary = client.list_clusters()
    clusters = summary['Clusters']
    clusters.should.have.length_of(1)
    cluster = clusters[0]
    cluster['NormalizedInstanceHours'].should.equal(0)
    cluster['Status']['State'].should.equal("RUNNING")
    cluster['Name'].should.equal('cluster')  # this line does not exist in the existing test
```

If mocked correctly, the last assert should be true, and the `boto` version of the tests pass this fine. However, with `boto3`, the backend has its `cluster.name` attribute set to `None`, and the above test fails.

This appears to be caused by the request body being JSON, but `moto.core.responses.BaseResponse.setup_class` does not parse `request.body` into `querystring` when the body is in JSON.

I've tried patching `moto.core.responses.BaseResponse.setup_class` so that `request.body` gets parsed into `querystring`, e.g., 

```python
    def setup_class(self, request, full_url, headers):
        querystring = {}
        if hasattr(request, 'body'):
            # Boto
            self.body = request.body
            if 'json' in request.headers.get('content-type', []):
                jd = json.loads(self.body)
                for key, value in jd.items():
                    querystring[key] = [value]
        ...
```

and this appears to make the `boto3` EMR tests pass, but it breaks `boto` version. Apparently, I'm doing the patching in a wrong place...

Our code base is mostly in `boto3` so I'd be happy to create PRs to check in fixes for use with `boto3`, but I would like to get some guidance as to where I should patch things like this.

okomestudio:
This revision adds code to handle cases when request body is JSON
(typically from boto3). A couple EMR endpoints are updated with boto3
responses and corresponding test cases are added to demonstrate that
it works.

This resolves #707.

### DIFF
--- a/moto/core/responses.py
+++ b/moto/core/responses.py
@@ -92,18 +92,18 @@ def flatten_json_request_body(dict_body):
         if isinstance(value, dict):
             r = flatten_json_request_body(value)
             for k, v in r.items():
-                flat[key + '.' + k] = [v]
+                flat[key + '.' + k] = v
         elif isinstance(value, list):
             for i, v in enumerate(value, 1):
                 memberstr = '.member.' + str(i)
                 if isinstance(v, dict):
                     r = flatten_json_request_body(v)
                     for k, vv in r.items():
-                        flat[key + memberstr + '.' + k] = [vv]
+                        flat[key + memberstr + '.' + k] = vv
                 else:
-                    flat[key + memberstr] = [v]
+                    flat[key + memberstr] = v
         else:
-            flat[key] = [value]
+            flat[key] = value
     return flat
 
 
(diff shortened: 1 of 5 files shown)


okomestudio:
Closing this in hope of creating another PR that takes care of boto3/JSON requests/responses more generically.

PR #710, opened 2016-09-23
Fix Route53 tagging code

kryptek:
Hey Steve, 

I suppose I should have tried using the built library locally. Although the Unittests were passing, the underlying implementation was not producing the correct results. I've refactored the models file to correct the issue.

Tags for a resourceID were all being placed under `self.resource_tags[UUID]['Value']`. Keys are now correctly added under the UUID with the proper key name and the delete logic works. 

I also tested this build code against the project I'm working on. Looks solid to me!


kryptek:
Ping! Can we get this pulled in, please?

spulec:
Thank you!

PR #712, opened 2016-09-23
fix: change dynamodb2 batch write response

skygeo:
Updating the Dynamodb2 Batch Write response format to match the official documentation:
http://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_BatchWriteItem.html


spulec:
Thanks!

PR #714, opened 2016-09-26
setup.py - use extras_require for flask

adamchainz:
Adding `moto` to the requirements file for my Django project pulls in `flask`, even though I don't need the server component. By making the `server` 'extra' then there is no hard requirement on flask.


spulec:
Thanks!

Issue #717, opened 2016-09-29
GPL Dependency


jzelinskie:
At some point you guys added `sure` as a dependency. Sure is licensed under the GPL and thus cannot be use in a number of projects. We depend on this library, but cannot ship GPL code. Are you interested in working around this issue at all?

andrewgross:
IANAL. Is it an issue to use `moto`, even though it is not used in the packaged version of the projects(available on PyPI?  It is only used in the development of the project, but it is not bundled with the code.

jzelinskie:
If that's the case, can you remove `sure`'s existence from moto's `setup.py`? Right now we basically have a script that searches for the GPL in the PyPi license data of each package in our requirements.txt and the latest `moto` is bringing `sure` into our project.

andrewgross:
Hmm, I think that could be doable.  A quick search doesn't have `sure` anywhere but in `test_` files.

andrewgross:
We already have `sure` listed in `requirements-dev.txt`, but it seems like we re-list it in `setup.py`.  Given that we don't even lock a `sure` version in `setup.py`, and that it is the only testing requirement, I think it can be safely removed.  I will try a PR with it.

andrewgross:
Same with `freezegun`.

spulec:
Fixed with #719

PR #719, opened 2016-09-29
Remove sure and freezegun from setup.py

andrewgross:
They still exist in requirements-dev.txt. Opening for testing.

This addresses #717.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -11,8 +11,6 @@ install_requires = [
     "xmltodict",
     "six",
     "werkzeug",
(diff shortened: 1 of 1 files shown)


andrewgross:
My local testing shows this seems to work still, without `freezegun` and `sure` in the `setup.py`.  Thoughts @spulec ?

jzelinskie:
LGTM! This change will let us use moto again.

spulec:
Thanks!

PR #720, opened 2016-09-30
The modify_network_interface_attribute method should work with updating other attributes too

choosy:
The method is currently failed if for example the Security group is not changed for that network interface, but the SourceDestCheck attribute is changed instead.

Although this commit doesn't completely take an action of when the SourceDestCheck is changed, it doesn't fail either.


spulec:
Thanks for bringing this up. It looks like the tests are failing. Any idea why?

JackDanger:
@choosy This PR still seems useful but I'm going to close it because of how old it is. Please reopen it if you get a chance!

PR #721, opened 2016-10-02
Add ARN and Alias functionality to KMS

rocky4570:
Doing some work with KMS and noticed moto only supported KeyId rather than the ARN or Aliases (describe_key only). This will allow testing to include:

Key ARN Example - arn:aws:kms:us-east-1:123456789012:key/12345678-1234-1234-1234-123456789012
Alias ARN Example - arn:aws:kms:us-east-1:123456789012:alias/MyAliasName
Globally Unique Key ID Example - 12345678-1234-1234-1234-123456789012
Alias Name Example - alias/MyAliasName

One area I found which I didnt change was the _assert_valid_key_id in kms/response.py is a regex search on the beginning of the string which results in it failing for the Key ARN example above. For this I use the backend get_key_id routine which determines the KeyId from Key ARN or returns the KeyID if used.

Unit tests have been included as always.

### DIFF
--- a/moto/kms/models.py
+++ b/moto/kms/models.py
@@ -108,7 +108,7 @@ class KmsBackend(BaseBackend):
         return self.key_to_aliases
 
     def get_key_id_from_alias(self, alias_name):
-        for key_id, aliases in dict(self.key_to_aliases).iteritems():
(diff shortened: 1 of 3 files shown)


spulec:
This is great! Thanks!

PR #723, opened 2016-10-06
lambderize the moto lambda

rocky4570:
This patch allows someone to run some rudimentary python lambda functions for unit testing. For example one can have a lambda function working within moto which can get a list of EC2 instances and return them to an existing moto'd lambda function (as per test).

Some minor tweaks were required on cloud-formation unit test as it was developed for node and well this just uses the python exec function to invoke the retrieved zip function and runs it.

Hopefully this will work for python 3, but I haven't caught up with it yet ;)

Also a few issues found. regions is retrieved from boto and for some reason only US and EU are in the latest boto release so have an append in for AUS, may require others or maybe use boto3 regions collection or something?

Also noticed 202 is returned on invoke rather than 200 - possibly when using boto?

### DIFF
--- a/moto/awslambda/models.py
+++ b/moto/awslambda/models.py
@@ -1,13 +1,18 @@
 from __future__ import unicode_literals
 
+import StringIO
 import base64
 import datetime
 import hashlib
 import io
 import json
-import StringIO
 import sys
 import zipfile
+
+try:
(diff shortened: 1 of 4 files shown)


spulec:
This is great, thanks!

PR #724, opened 2016-10-08
Return the right dhcp_options id in the vpc response

AgarFu:
After associate a vpc with a dhcp_options moto wasn't updating the id of the dhcp_options in the vpc describe response.


spulec:
Thanks for this. If you can revert the change to the version, I'll be happy to merge.

AgarFu:
Reverted the changes to the version. Thank you for this amazing work!

spulec:
Thanks!

Issue #726, opened 2016-10-10
post-provision EIP allocations not added to instance data


majuscule:
I believe I have found an issue with dynamically allocated and associated EIPs not being added to the compute instance description endpoint. The standalone script below provisions an instance and eip, associates the resources, and looks for the missing network interface.

I have also found that if the vpc subnet is not explicitly declared in the `create_instances` configuration dictionary, this test will pass. There seems to be some differences between a "moto default vpc/subnet" and a "boto default vpc/subnet" that would explain this.

Finally, I have noticed that in the successful case, the instance description still differs from the current AWS response which provides PublicIpAddress and PublicDnsName as top level keys instead of nesting them under NetworkInterfaces.

I will continue to investigate and attempt to streamline the basic test cases, but wanted to raise this issue for comment immediately in case I am missing something.

`pytest standalone.py`

```python
import pytest
import boto3
from moto import mock_ec2

# Force a singular moto backend to be shared across fixtures
mock_ec2_server = mock_ec2()
mock_ec2_server.start()


@pytest.fixture(scope='module')
def session():
    return boto3.session.Session(region_name='us-west-1')


@pytest.fixture(scope='module')
def service(session):
    return session.resource('ec2')


@pytest.fixture(scope='module')
def client(session):
    return session.client('ec2')


@pytest.fixture(scope='module')
def network(client):
    vpc_res = client.create_vpc(CidrBlock='10.0.0.0/24')
    subnet_res = client.create_subnet(
            VpcId=vpc_res['Vpc']['VpcId'], CidrBlock='10.0.0.0/24')
    return {
        'vpc' : vpc_res['Vpc']['VpcId'],
        'subnet' : subnet_res['Subnet']['SubnetId']
    }


@pytest.fixture(scope='module')
def instance(service, network):
    return service.create_instances(**{
        'InstanceType': 't2.micro',
        'ImageId': 'ami-test',
        'MinCount': 1,
        'MaxCount': 1,
        'SubnetId': network['subnet']
    })[0]


@pytest.fixture(scope='module')
def boto3_associate_ip(client, instance):
    allocation_id = client.allocate_address(Domain='vpc')['AllocationId']

    association_id = client.associate_address(
        InstanceId=instance.id,
        AllocationId=allocation_id,
        AllowReassociation=False)

    return association_id


def test_boto3_ip_association(instance, boto3_associate_ip, client):
    description = client.describe_instances(InstanceIds=[instance.id])\
[...]

PR #727, opened 2016-10-10
add test case for eip association within a new vpc

majuscule:
I've written a new test case with regard to issue https://github.com/spulec/moto/issues/726. I'm not sure if I'm missing something basic here, but I can't figure out why the instance does not receive the associated EIP as a public address as this test attempts and fails. I do not believe that an instance should require an ENI for public addressing.


majuscule:
The reason for this behavior is that AWS does not refresh instance data without explicit subsequent API calls. Re-fetching the instance after EIP association provides the `.ip_address` attribute.

PR #730, opened 2016-10-11
add test case for instance.load after eip association

majuscule:
This test case demonstrates the failure of a `boto3.resources.factory.ec2.Instance` to refresh `public_ip_address` and `public_dns_name` attributes after EIP allocation and association.


spulec:
Thanks, this was added with 27f1bece9b2aab2ad7e0d2bf705101b28c48eab6

PR #732, opened 2016-10-12
Implementing KMS encrypt/decrypt

JackDanger:
This adds support for testing the /encrypt and /decrypt endpoints
of Amazon KMS

### DIFF
--- a/moto/kms/responses.py
+++ b/moto/kms/responses.py
@@ -1,5 +1,6 @@
 from __future__ import unicode_literals
 
+import base64
 import json
 import re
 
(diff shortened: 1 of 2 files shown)


spulec:
Thanks for opening this!

It looks like there is an issue with Python3 and the tests. If we can get that fixed, would be happy to merge.

JackDanger:
Sadly, these changes are impossible to get working under Py3 because the travis build hardcodes the Boto version to 2.36 and Boto didn't support Py3 until 2.40+ for this feature.
Putting the tests behind a version check for now but I've verified that it works properly under modern versions of Boto.

JackDanger:
@spulec If we add more recent versions of Boto to the .travis.yml test matrix then I can add a condition to run this test against fixed versions of Boto. Otherwise this is ready to go as-is.

spulec:
That works for now. Thanks!

PR #735, opened 2016-10-15
Add dry_run to a number of EC2 services

rocky4570:
Add dry_run capability for a number of boto EC2 calls and also provided unit tests where applicable.

### DIFF
--- a/moto/core/responses.py
+++ b/moto/core/responses.py
@@ -3,6 +3,8 @@ import datetime
 import json
 import re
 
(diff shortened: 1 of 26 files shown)


rocky4570:
I should admit that possibly some of the responses from this commit are not likely to be accurate with what AWS respond with as I havent had a need or seen all DryRun responses that were committed here. Please update any that may come along the way.

spulec:
This is great, thanks!

PR #736, opened 2016-10-17
Remove botocore version pinning

okomestudio:
The boto3 and botocore packages in the install requirement (for development) are pinned to older versions, due to a recently added field `RetryAttempts` in JSON response making existing tests to fail. That field doesn't appear important for mocking purposes, and can therefore safely be ignored for testing. This revision modifies tests to ignore it, so that more recent version of boto3 and botocore can be used while testing.

This revision also includes a minor update to JSON response generation routine to align its behavior more closely to the unmocked behavior.


spulec:
Thanks!

PR #738, opened 2016-10-19
Implement IAM managed policy

okomestudio:
This revision adds mock objects and a few end points for IAM managed policies, which are not mocked yet (only inline policies are mocked in the existing code base).

### DIFF
--- a/moto/iam/models.py
+++ b/moto/iam/models.py
@@ -1,10 +1,57 @@
 from __future__ import unicode_literals
-
-from moto.core import BaseBackend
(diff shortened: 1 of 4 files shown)


okomestudio:
It looks like the tests are failing because the boto version for tests is pinned at version 2.36.0, and one of the functions (`create_policy`) implemented in the revision appears to be added in a recent version. The tests pass fine when I run locally (with boto 2.42.0).

I am actually not sure what the guideline is for dealing with an issue like this. I certainly would need that function to be mocked by moto for my project, but then I also understand the desire to stay a little conservative for existing users.

Any suggestion?

spulec:
This is great. Thank you!

PR #739, opened 2016-10-20
Implement filters and pagers for some EMR end points

okomestudio:
This revision implements proper filtering and paging (with marker) for relevant EMR end points.


spulec:
This is great. Thanks!

PR #742, opened 2016-10-20
made the Security Group backend throw the same error as AWS if the nb of sec groups limit is hit

costypetrisor:
I have used the default limits from:
- [https://docs.aws.amazon.com/general/latest/gr/aws_service_limits.html#limits_ec2](EC2-classic sec group limits)
- [https://docs.aws.amazon.com/AmazonVPC/latest/UserGuide/VPC_Appendix_Limits.html#vpc-limits-security-groups](EC-VPC sec group limits)

### DIFF
--- a/moto/ec2/models.py
+++ b/moto/ec2/models.py
@@ -1265,6 +1265,16 @@ class SecurityGroup(TaggedEC2Resource):
     def add_egress_rule(self, rule):
         self.egress_rules.append(rule)
 
+    def get_number_of_ingress_rules(self):
+        return sum(
+            len(rule.ip_ranges) + len(rule.source_groups)
+            for rule in self.ingress_rules)
+
+    def get_number_of_egress_rules(self):
+        return sum(
+            len(rule.ip_ranges) + len(rule.source_groups)
+            for rule in self.egress_rules)
+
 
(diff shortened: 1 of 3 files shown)


costypetrisor:
I need to refactor this a bit. I'll let you know when it's ok

costypetrisor:
Apologies for the long delay in regards to updating this PR :(

I did indeed had a an oversight: grants to other security groups do weigh in the overall limit of grants that are allowed for a security group.
I've fixed that, and extended the 2 unit tests I've added to cover this as well.

moto/ec2/models.py
```diff
@@ -1415,6 +1429,19 @@ def authorize_security_group_egress(self,
                 if not is_valid_cidr(cidr):
                     raise InvalidCIDRSubnetError(cidr=cidr)
 
+        max_nb_rules = 50 if group.vpc_id else 100
```
spulec: Can you generalize the logic between these two places a bit?

costypetrisor: @spulec I've added a new commit. Have a look and let me know if you like the way I did it or not.
:)

spulec:
Looks great, thanks!

PR #744, opened 2016-10-25
Fix failure of scan() after has_item()

### LINKED ISSUE #731: Dynamodb has_item causes scan problems
Calling `has_item` for an item that doesn't exist, and then calling `scan()` causes `AttributeError: 'dict' object has no attribute 'attrs'`

Example of behaviour that causes this:

``` python
import boto


with mock_dynamodb():
    conn = boto.connect_dynamodb()
    table = conn.create_table(
        name='mytable',
        schema=conn.create_schema(
            hash_key_name='name',
            hash_key_proto_value=str,
        ),
        read_units=1,
        write_units=1
    )
    print(list(table.scan()))
    table.has_item('something')
    print(list(table.scan()))
```

`has_item` calls down to `Table.get_item` in `dynamodb/models.py`. This accesses `self.items`, which is a `defaultdict` whose default value is an empty dict. This results in an empty dict being set in `self.items`. The `scan()` call in `dynamodb/responses.py` does not handle dicts (or anything other than an Item)

Replacing the error line in `dynamodb/responses.py` with the following code appears to work:

``` python
"Items": [item.attrs for item in items if item],
```

I'm not very familiar with this library (only started using it today) but happy to submit a PR if that seems like the right thing to do.

helenst:
Fixes #731 

It wasn't causing problems with range key (perhaps `has_item` follows a different path there) but I suppose good to test it in any case?

### DIFF
--- a/moto/dynamodb/responses.py
+++ b/moto/dynamodb/responses.py
@@ -278,7 +278,7 @@ class DynamoHandler(BaseResponse):
 
         result = {
             "Count": len(items),
-            "Items": [item.attrs for item in items],
+            "Items": [item.attrs for item in items if item],
             "ConsumedCapacityUnits": 1,
             "ScannedCount": scanned_count
         }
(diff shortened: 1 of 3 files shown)


spulec:
Great, thanks!

PR #757, opened 2016-11-04
Attempt to parse simple ConditionExpressions in DynamoDB put_item().

nfvs:
Internally these get converted into simpler Expected={} parameters, so
for now only `attribute_exists` and `attribute_not_exists` are supported;
OR is also not suported since `expected` does not support it.


spulec:
Thanks!

PR #763, opened 2016-11-08
Fix SNS application endpoint to match AWS return format

mpappas86:
SNS endpoints currently only hold exactly the `attributes` dict that is passed into the constructor. While this naively seems correct, it turns out it doesn't match the behavior of AWS typically, which automatically sets 2 attributes - `Token` and `Enabled`. My change simply auto-sets those attributes if they weren't explicitly included, and modifies the tests accordingly.


spulec:
Thanks!

PR #764, opened 2016-11-09
Fixed time formatting in ec2/models.py

SectorNine50:
The format returned by Amazon's API follows what was created in the `utc_date_and_time()` function already in this file, not `datetime.utcnow().isoformat()`. Made running tests on comparisons by creation date difficult since the test format would return different than the real API. Haven't checked if this problem occurs anywhere else.

### DIFF
--- a/moto/ec2/models.py
+++ b/moto/ec2/models.py
@@ -917,7 +917,7 @@ class Ami(TaggedEC2Resource):
         self.architecture = None
         self.kernel_id = None
         self.platform = None
-        self.creation_date = datetime.utcnow().isoformat()
+        self.creation_date = utc_date_and_time()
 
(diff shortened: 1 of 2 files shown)


SectorNine50:
The test that failed (`test_layers.test_describe_layers`) appears to be due to a race condition in the test itself:
>X['ResponseMetadata']['HTTPHeaders']['date'] is 'Mon, 14 Nov 2016 17:58:36 GMT' whereas Y['ResponseMetadata']['HTTPHeaders']['date'] is 'Mon, 14 Nov 2016 17:58:37 GMT'

spulec:
We should be able to use freezegun for this test in order to get it passing.

Issue #532, opened 2016-02-04
s3bucket_path broken from 0.4.19 to 0.4.20


leplatrem:
I couldn't find any issue (appart the old #55) for this regression.

I run `moto_server s3bucket_path  -H 0.0.0.0 -p 5000`.

With 0.4.19:

```
$ curl http://127.0.0.1:5000/            

<ListAllMyBucketsResult xmlns="http://s3.amazonaws.com/doc/2006-03-01">
  <Owner>
    <ID>bcaf1ffd86f41161ca5fb16fd081034f</ID>
    <DisplayName>webfile</DisplayName>
  </Owner>
  <Buckets>
 </Buckets>
</ListAllMyBucketsResult>
```

With 0.4.20:

```
$ curl http://127.0.0.1:5000/            

<?xml version="1.0" encoding="UTF-8"?>
  <Response>
    <Errors>
      <Error>
        <Code>NoSuchBucket</Code>
        <Message>The specified bucket does not exist</Message>
        <BucketName>127</BucketName>
      </Error>
    </Errors>
  <RequestID>7a62c49f-347e-4fc4-9331-6e8eEXAMPLE</RequestID>
</Response>
```

I cannot invest much time into this (right now) and just wanted to let you know about this regression (_plus, 0.4.19 is working fine_). So I'm sorry to be the one adding stuff to the pile without contributing...

Thank for your project! Vital for our functional tests!

spulec:
No problem, thanks for opening the issue!

If you pass the bucketname as a header, does it work. ie

`$ curl http://127.0.0.1:5000/ -H 'Host: bucketname.s3.amazonaws.com'`

leplatrem:
I have the same behaviour:

```
$ pip install moto
Downloading/unpacking moto
  Downloading moto-0.4.23-py2.py3-none-any.whl (253kB): 253kB downloaded
...
...


$ moto_server s3bucket_path -H 0.0.0.0 -p 5000 
 * Running on http://0.0.0.0:5000/ (Press CTRL+C to quit)
127.0.0.1 - - [29/Apr/2016 12:46:13] "GET / HTTP/1.1" 404 -
```

```
curl http://127.0.0.1:5000/ -H 'Host: bucketname.s3.amazonaws.com'
<?xml version="1.0" encoding="UTF-8"?>
  <Response>
    <Errors>
      <Error>
        <Code>NoSuchBucket</Code>
        <Message>The specified bucket does not exist</Message>
[...]

PR #765, opened 2016-11-10
Fix s3bucketpath handling for IP based requests

2mf:
I've made some additions to the 'Host' field checking to determine whether it is subdomain based or url path based.
With this change it works with plain IP-Addresses.

This partially fixes issue #532

### DIFF
--- a/moto/s3/responses.py
+++ b/moto/s3/responses.py
@@ -46,23 +46,28 @@ class ResponseObject(_TemplateEnvironmentMixin):
 
     def subdomain_based_buckets(self, request):
         host = request.headers.get('host', request.headers.get('Host'))
-        if host.startswith("localhost"):
+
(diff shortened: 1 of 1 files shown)


jgelens:
Great, this fixes our use-case.

spulec:
Thanks

leplatrem:
Thanks for tackling this!

rocky4570:
This change causes problems when using moto on Windows machines as inet_pton is not part of the socket module. 

Results in AttributeError: 'module' object has no attribute 'inet_pton'

Requires installing (win-inet-pton) and placing:

import os
if os.name == 'nt':
    import win_inet_pton
import socket

But this isn't really a good option IMHO, as forces requirements to contain a library for architecture type for  the 1 check being done to determine if sub-domain.

PR #768, opened 2016-11-11
Add support for login profile deletion

smarlowucf:
Support the creation of login profiles using boto3 and deletion of login profiles.

A KeyError was raised when attempting to create a login profile using boto3:
- Fixed typo in XML result.
- Add test to cover boto3.

delete_login_profile:
- Resets user password to None.
- If user password is already none (login profile does not exist) raise exception.


spulec:
Great, thanks!

PR #773, opened 2016-11-17
Add more availability zones and implement default VPC

okomestudio:
This PR includes the following revisions on `mock_ec2`:

- Add more regions
- Add three availability zones for each region, and create default VPC (and its default subnet). This is more inline with the current EC2 behavior. (With the prior code base default VPC/subnet needed to be created manually)
- Bug fix: Add `availability-zone` as an appropriate filter name.
- Bug fix: Add more filter name variety (e.g., 'vpc-id' and 'vpcId')
- Bug fix: Use `subnet_ids` parameter if provided through query string. Currently `subnet_ids` filter results appropriately only when they are provided as filter, and it is not compatible with boto/botocore interface.
- Appropriate fixes for affected tests

The overall the assumptions on default VPC/availability zones/subnets change quite a bit, so please review the change and code carefully when evaluating whether to merge into the upstream.

### DIFF
--- a/tests/test_ec2/test_network_acls.py
+++ b/tests/test_ec2/test_network_acls.py
@@ -7,41 +7,37 @@ from moto import mock_ec2
 
 @mock_ec2
 def test_default_network_acl_created_with_vpc():
-
     conn = boto.connect_vpc('the_key', 'the secret')
     vpc = conn.create_vpc("10.0.0.0/16")
-
-    all_network_acls = conn.get_all_network_acls()
-    all_network_acls.should.have.length_of(1)
-
-@mock_ec2
-def test_network_acls():
-
-    conn = boto.connect_vpc('the_key', 'the secret')
-    vpc = conn.create_vpc("10.0.0.0/16")
-
-    network_acl = conn.create_network_acl(vpc.id)
-
     all_network_acls = conn.get_all_network_acls()
     all_network_acls.should.have.length_of(2)
 
-@mock_ec2
-def test_new_subnet_associates_with_default_network_acl():
 
+@mock_ec2
+def test_network_acls():
     conn = boto.connect_vpc('the_key', 'the secret')
     vpc = conn.create_vpc("10.0.0.0/16")
+    network_acl = conn.create_network_acl(vpc.id)
+    all_network_acls = conn.get_all_network_acls()
+    all_network_acls.should.have.length_of(3)
+
+
+@mock_ec2
+def test_new_subnet_associates_with_default_network_acl():
+    conn = boto.connect_vpc('the_key', 'the secret')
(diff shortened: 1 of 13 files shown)


okomestudio:
Oops some tests failed. Fixes are on the way...

tests/test_ec2/test_network_acls.py
```diff
@@ -7,41 +7,37 @@
 
 @mock_ec2
 def test_default_network_acl_created_with_vpc():
-
     conn = boto.connect_vpc('the_key', 'the secret')
     vpc = conn.create_vpc("10.0.0.0/16")
-
     all_network_acls = conn.get_all_network_acls()
-    all_network_acls.should.have.length_of(1)
+    all_network_acls.should.have.length_of(2)
```
spulec: Why the extra ACL?

spulec:
Thanks for this. One question, but excited to get this merged

tests/test_ec2/test_network_acls.py
okomestudio: Sorry for not catching up to this sooner!

So this is because every time a new VPC is created, the default ACL for that VPC is created as well:

https://github.com/spulec/moto/blob/master/moto/ec2/models.py#L1832-L1833

With the change in this PR, the default VPC (and the corresponding default network ACL) already exists when the mock starts. But in this test case, a new VPC and (its default ACL) is explicitly created. So in total, there are two network ACLs found with `get_all_network_acls()` since by default the method list all the ACLs found in the region.

In general, I think I had to increment the count of ACLs by one for tests of this kind due to the pre-existing default VPC. So this could be confusing depending on the assumption. The original intention might be clearer if the tests get network ACLs associated only with the VPC newly created within the test?

yannlambret:
I'm not a big fan of the 'default VPC' feature. A lot of people or companies just remove the default VPC,  so it wouldn't fit their environments.

Besides, it's not really hard to "emulate" the default VPC in tests, something like:

```python
class VpcTestCase(unittest.TestCase):

    credentials = {
        'aws_access_key_id': 'access_key',
        'aws_secret_access_key': 'secret_access_key',
        'region_name': 'us-west-2'
    }

    def setUp(self):
        self.mock = mock_ec2()
        self.mock.start()

        self.session = Session(**self.credentials)
        self.ec2 = self.session.resource('ec2')

        # Create a default VPC for each test
        self.ec2.create_vpc(CidrBlock='172.31.0.0/16')

    def tearDown(self):
        self.mock.stop()
```

yannlambret:
Looks like I spoke too soon. The current limitation is that first VPC created when mocking is always created as the default VPC. So there's no way to simulate a default VPC free environment anyway.

spulec:
Thanks for your work on this!

PR #774, opened 2016-11-18
Make Cloudformation/parsing#load_parameters handle lists

Dawil:
We were using this and noticed that list parameters weren't getting properly turned into lists by moto. This short addition will check the `Type` of a resolved parameter and if it's a `CommaDelimitedList` will split it on commans (',').


Dawil:
Sorry for the coverage spam. I continued playing around in the branch and didn't realize the extra commits were ending up here. I reverted back to just the necessary commit and will continue playing in a separate branch.

spulec:
Thanks!

PR #776, opened 2016-11-21
rename duplicate rds/models db_instance_identifier to physical_resource_id

Dawil:
This is needed for [clean_json](https://github.com/spulec/moto/blob/master/moto/cloudformation/parsing.py#L98) to correctly handle 'Ref's for DBInstanceIdentifiers.

You'll notice that in the changed file an assignment happens to `self.db_instance_identifier` twice, the same value each time. I suspect that this may have been a copy paste error that forgot to rename the second occurrence to physical_resource_id.

### DIFF
--- a/moto/rds/models.py
+++ b/moto/rds/models.py
@@ -130,7 +130,7 @@ class Database(object):
         source_db_identifier = properties.get("SourceDBInstanceIdentifier")
         if source_db_identifier:
             # Replica
-            db_kwargs["source_db_identifier"] = source_db_identifier.db_instance_identifier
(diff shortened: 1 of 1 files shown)


Dawil:
I updated some code to make the two breaking tests pass. However it really needs someone knowledgeable on the codebase to confirm whether or not the change will make a breaking change.

spulec:
Ah, the way we have generally be handling this is to create a property called this. See https://github.com/spulec/moto/blob/81d489ce46535e98b379a8f3d00503d9870307d9/moto/route53/models.py#L24-L26

spulec:
Thanks!

Issue #748, opened 2016-10-30
By default variables in XML templates aren't escaped


tomviner:
Doing some mocking with moto, and I have `&` in my test data. This kind of thing fails:

```
@mock_s3
def test_boto3_list_keys_xml_escaped():
    s3 = boto3.client('s3', region_name='us-east-1')
    s3.create_bucket(Bucket='mybucket')
    s3.put_object(Bucket='mybucket', Key='Q&A.txt', Body=b'is awesome')
    s3.list_objects_v2(Bucket='mybucket')
```

errors because the `&` isn't escaped `<Key>Q&A.txt</Key>`:

``` python
  File "/home/me/dev/moto/.tox/py34/lib/python3.4/site-packages/botocore/parsers.py", line 386, in _parse_xml_string_to_dom
    "invalid XML received:\n%s" % (e, xml_string))
nose.proxy.ResponseParserError: Unable to parse response (not well-formed (invalid token): line 9, column 18), invalid XML received:
b'<?xml version="1.0" encoding="UTF-8"?>\n<ListBucketResult xmlns="http://s3.amazonaws.com/doc/2006-03-01/">\n  <Name>mybucket</Name>\n  <Prefix>None</Prefix>\n  <MaxKeys>1000</MaxKeys>\n  <Delimiter>None</Delimiter>\n  <IsTruncated>false</IsTruncated>\n    <Contents>\n      <Key>Q&A.txt</Key>\n      <LastModified>2016-10-30T19:53:45.011Z</LastModified>\n      <ETag>"d32bda93738f7e03adb22e66c90fbc04"</ETag>\n      <Size>10</Size>\n      <StorageClass>STANDARD</StorageClass>\n      <Owner>\n        <ID>75aa57f09aa0c8caeab4f8c24e99d10f8e7faeebf76c078efc7c6caea54ba06a</ID>\n        <DisplayName>webfile</DisplayName>\n      </Owner>\n    </Contents>\n  </ListBucketResult>'
```

Similarly listing by prefix, regardless of objects in s3 fails:

``` python
s3.list_objects_v2(Bucket='mybucket', Prefix='<>')
# writes
<Prefix><></Prefix>
```

This is because in `moto/core/responses.py` [`_TemplateEnvironmentMixin.should_autoescape` was set to False](https://github.com/spulec/moto/commit/d3e4c2c4b5bf900ac1c32cbb8469ef718e14c807#diff-427c67d098e415aae0a327e597f8c7c7R65).

[...]

PR #779, opened 2016-11-24
Turn on XML escaping by default

tomviner:
Fix for https://github.com/spulec/moto/issues/748

This turns on XML escaping by default.


spulec:
Thanks!

Issue #780, opened 2016-11-27
Modifications: Fix s3bucketpath handling for IP based requests #765


rocky4570:
With regards to change https://github.com/spulec/moto/pull/765.

This change causes problems when using moto (s3) on Windows machines as inet_pton is not part of the socket module. Results in AttributeError: 'module' object has no attribute 'inet_pton'

Fix is to use inet_pton but requires installing (win-inet-pton) and placing in moto/s3/responses.py:

import os
if os.name == 'nt':
    import win_inet_pton
import socket

But this isn't really a good option IMHO, as forces additional requirements to contain a library for architecture type for the 1 check being done to determine if sub-domain.

jgelens:
Replacing the ip logic with a regex for ipv4 and ipv6 IP's would maybe work?

spulec:
Fixed with #784

Issue #769, opened 2016-11-14
Invoking Lambda function does not generate the correct payload


RealSalmon:
Is the payload delivered when invoking a lambda function incorrect? Instead of delivering the function's return value, it looks like it is delivering a dictionary with a single 'result' key that contains the log output.

e.g. I would expect test_invoke() below to pass . . .

```python

import io
import zipfile
import boto3
import json
from moto import mock_lambda

lambda_fx = """
def lambda_handler(event, context):
    return event
"""

def init_lambda(fx):
    zip_output = io.BytesIO()
    zip_file = zipfile.ZipFile(zip_output, 'w', zipfile.ZIP_DEFLATED)
    zip_file.writestr('lambda_function.zip', fx)
    zip_file.close()
    zip_output.seek(0)

    boto3.client('lambda').create_function(
        FunctionName='test-lambda-fx',
        Runtime='python2.7',
        Role='test-iam-role',
        Handler='lambda_function.lambda_handler',
        Description='test lambda function',
        Timeout=3,
        MemorySize=128,
        Publish=True,
        Code={
            'ZipFile': zip_output.read()
        },
    )


@mock_lambda
def test_invoke():
    init_lambda(lambda_fx)
    result = boto3.client('lambda').invoke(
        FunctionName='test-lambda-fx',
        InvocationType='RequestResponse',
        LogType='Tail',
        Payload=json.dumps({'hello': 'world'})
    )

    assert json.loads(result['Payload'].read()) == {'hello': 'world'}

```

rocky4570:
Ah ok - the patch I put up was putting data to LogResult (http://boto3.readthedocs.io/en/latest/reference/services/lambda.html#Lambda.Client.invoke) so I will have a look at this.

spulec:
Fixed with #781

PR #781, opened 2016-11-28
Fixup lambda for ResponseRequest

rocky4570:
Issues with #780 aside,

Here is attempt 2 or 5 (forget) for fixing Lambda. Issue #769 - Invoking Lambda function does not generate the correct payload.

As mentioned there was only putting data out for Cloud Watch Logs not in actual payload - hopefully this will address it.


spulec:
Great, thanks!

PR #784, opened 2016-12-01
Fix s3bucket_path

2mf:
This pull request fixes a issue within s3 when creating a bucket and the URL ends with a slash.

It as well fixes the compatibility issue mentioned in issue #780 .

Testcases were extended to cover these cases.


spulec:
Awesome, thank you!

PR #787, opened 2016-12-05
sqs: Use unix_time in place of time.time()

ronin13:
unix_time() from moto.core.utils is used as the time source through moto,
and it is identical to time.time() in output. Hence, using unix_time()
since it makes mocking easier during testing (when time is mocked out).


spulec:
Thanks!

PR #789, opened 2016-12-07
Various RDS, RDS/Cloudformation, RDS/KMS fixes and additions

mnussbaum:
We need to mock out deploying RDS instances with full disk encryption
and detailed tagging. We also need to be able do deploy these instances
with Cloudformation, and then access them with both boto and boto3.
The following changes were made to support this use case:

* Join RDS and RDS2 backends - this makes RDS resources created via
  either of the two boto RDS APIs visible to both, more closely
  mirroring how AWS works
* Fix RDS responses that were returning JSON but should be returning XML
* Add mocking of RDS Cloudformation calls
* Add mocking of RDS full disk encryption with KMS
* Add mocking of RDS DBParameterGroups
* Fix mocking of RDS DBSecurityGroupIngress rules
* Make mocking of RDS OptionGroupOptions more accurate
* Fix mocking of RDS cross-region DB replication
* Add tag support to RDS:
  * DBs
  * DBSubnetGroups
  * DBSecurityGroups

@ritsok and @2rs2ts helped with this work!

### DIFF
--- a/moto/cloudformation/parsing.py
+++ b/moto/cloudformation/parsing.py
@@ -12,6 +12,7 @@ from moto.elb import models as elb_models
 from moto.iam import models as iam_models
 from moto.kms import models as kms_models
 from moto.rds import models as rds_models
+from moto.rds2 import models as rds2_models
 from moto.redshift import models as redshift_models
 from moto.route53 import models as route53_models
 from moto.s3 import models as s3_models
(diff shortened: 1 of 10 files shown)


mnussbaum:
I've amended the commit to fix the test failures that were occurring under python 2.6

spulec:
Awesome, thank you for all of the work on this!

mnussbaum:
Thanks for merging!

PR #790, opened 2016-12-08
Fixes for dynamodb2 mocking

tideline:
We started using moto for our unit tests, and the following were needed to make our code work with the mocked dynamodb2:

- dynamodb item updates allow whitespace around the `=` in update expression, so splitting the expression by whitespace does not guarantee grouping the value expression together
- dynamodb attribute name replacement allows names to be substrings of other names (e.g. we programatically construct `c1` and `c10`, and the mock library would replace these in sequence, thus could replace `c1` before `c10` leading to faulty replacement
- the code to find the hash key condition for a query assumed that the indexes of a sorted list of attribute expression names match the corresponding indexes into the key conditions; this is not generally true, as there can be any number of attributes in the attribute names, leading to the wrong conditions being popped off the list. This PR replaces that assumption with actually searching for the attribute name in the expressions.


moto/dynamodb2/models.py
```diff
@@ -110,27 +111,27 @@ def describe_attrs(self, attributes):
...
-        ACTION_VALUES = ['SET', 'set', 'REMOVE', 'remove']
-
-        action = None
-        for value in update_expression.split():
-            if value in ACTION_VALUES:
-                # An action
-                action = value
-                continue
-            else:
+        parts = [p for p in re.split(r'\b(SET|REMOVE|ADD|DELETE)\b', update_expression) if p]
```
georgepsarakis: Why are you filtering values in the list comprehension with `if p`? As far as I understand a valid expression will not yield e.g. an empty string.

moto/dynamodb2/models.py
```diff
@@ -110,27 +111,27 @@ def describe_attrs(self, attributes):
...
+        for action, valstr in zip(parts[:-1:2], parts[1::2]):
```
georgepsarakis: `zip` will not raise an error if the iterables are not of the same length: 

```python
zip(range(10), range(20)) # [(0, 0) ... (9, 9)]
```

Perhaps it would be better to use a list comprehension?

```python
for operation, attributes in ((parts[index], parts[index + 1]) for index in xrange(len(parts))):
    ...
```

JackDanger:
@tideline & @sumpfork There are some comments on this PR thanks to @georgepsarakis, could you please take a look and either reply or implement the changes? Then we can merge this to master!

moto/dynamodb2/models.py
sumpfork: `'SET #c0=:v0, #c1=:v1, #c...2, #c3=:v3, #c4=:v4, #c5=:v5, #c6=:v6, #c7=:v7, #c8=:v8, #c9=:v9, #c10=:v10'`, for example, generates the list `[u'', u'SET', u' #c0=:v0, #c1=:v1, #c2=:v2, #c3=:v3, #c4=:v4, #c5=:v5, #c6=:v6, #c7=:v7, #c8=:v8, #c9=:v9, #c10=:v10']`, so this `if p` filters out the empty starting element. I guess it's true that only that starting element should ever be empty - I could change the code to check that explicitly and then throw it out?

moto/dynamodb2/models.py
sumpfork: While true that `zip` behaves that way, we are zipping a list against itself here, so the lengths can't be radically different. At worst, the length is odd and we'll miss the final element. I added an assert to catch that.

moto/dynamodb2/models.py
georgepsarakis: @sumpfork how about replacing with `match`, I think it increases readability:

```
>>> print re.match(r'(?P<action>SET|REMOVE|ADD|DELETE)\s(?P<attribute_pairs>.+)', s).groupdict()
>>> {'action': 'SET',
 'attribute_pairs': '#c0=:v0, #c1=:v1, #c...2, #c3=:v3, #c4=:v4, #c5=:v5, #c6=:v6, #c7=:v7, #c8=:v8, #c9=:v9, #c10=:v10'}
```

sumpfork: As far as I understand the DynamoDB docs (I haven't used this feature myself) it is legal to include multiple update statements in one update expressions as long as they don't use the same operator. Using `match` in this straightforward manner would include any following ones in the `attribute_pairs` group. We could design a fancier regex and use `findall`, but I believe the only way expressions are identified is by occurrence of one of the operator keywords, so I doubt that this expression would be more readable.

sumpfork: I think this would be the `findall` version:
```
        operator_expr = r'\b(?:SET|REMOVE|ADD|DELETE)\b'
        parts = re.findall('({})((?:.(?!{}))+)'.format(operator_expr, operator_expr), update_expression)
```

georgepsarakis: @sumpfork I agree with you, there seems to be virtually no gain and the readability is better as the `split` version is much easier for someone to understand. Perhaps what is needed is just a comment for the `if` filtering in the list comprehension, for future readers to understand.

JackDanger:
@georgepsarakis does this look good to you?

JackDanger:
@tideline Thanks for your contribution. Apologies it took a while to review and merge but we're on the case now and we'd love to have more great improvements like this from you. Welcome to Moto!

PR #796, opened 2016-12-16
Add service ARNs support to the `DescribeServices` ECS API

hltbra:
Before this PR, moto only worked with service names, but AWS also supports service ARNs.


spulec:
Thanks!

PR #725, opened 2016-10-09
Some fixes for Route53 mocks

joshuaspence:
Two fixes for mocking Route53 API calls:
1. When calling `created_hosted_name`, a period will be appended to the zone name if it did not already end in a period.
2. The zone IDs in the response from the `list_hosted_zones` are prefixed with `/hostedzone/`, similar to the responses from other Route53 methods.

PR #799, opened 2017-01-05
Added '/hostedzone/' to hosted zone ID as they are in both boto2 and …

AkhterAli:
Added the correct output of zone ID from boto2 and boto3.

### DIFF
--- a/moto/route53/models.py
+++ b/moto/route53/models.py
@@ -231,7 +231,7 @@ class Route53Backend(BaseBackend):
         self.resource_tags = defaultdict(dict)
 
     def create_hosted_zone(self, name, private_zone, comment=None):
-        new_id = get_random_hex()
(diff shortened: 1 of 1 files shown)


AkhterAli:
I'll elaborate on this.  On Boto3 and on boto2 (when calling get_zones or get_hosted_zone_by_name), the hosted zone ID is returned with a '/hostedzone/' prefix.  Thus many of us strip this out.  If possible, could we also add the '/hostedzones/' prefix when calling the same API's on Moto?

If need be, I'll build a utility trimmer to trim it out so the tests do not fail.

spulec:
I'm confused. Does boto add this or is it part of the return data in the API?

Any idea why the tests are failing?

spulec:
Going to close this in favor of #725

PR #802, opened 2017-01-12
IAM: raise error if requiested instance profile does not exist

jszwedko:
Borrowed the same approach missing roles takes

Signed-off-by: Andrew Harris <andrew.harris@getbraintree.com>


spulec:
Thanks!

PR #804, opened 2017-01-12
Fix Route53 response and tests

AkhterAli:
This commit is to make sure #725 and #799 pass.  This commit should not pass tests as the tests need to be changed.  If tests change before commit then it should pass.

get_all_zones() does not return just the zone ID, it returns '/hostedzone/zoneid'


AkhterAli:
As I said, this will fail because of the conditions put on the tests.  The tests need to be changed, I've added the new changes on my PR.

spulec:
Thanks!

PR #806, opened 2017-01-13
sts: Implement get_caller_identity

jszwedko:
Return a canned response

Signed-off-by: Andrew Harris <andrew.harris@getbraintree.com>


spulec:
Thanks!

PR #809, opened 2017-01-13
Allow ELB policies to be set via Cloudformation.

mnussbaum:
Also more closely mirror AWS by not returning ELB
BackendServerDescriptions that don't have policies.

Signed-off-by: Michael Nussbaum <michael.nussbaum@getbraintree.com>


spulec:
Cool, thank you!

PR #811, opened 2017-01-19
Add resource IDs to ECS Cluster and Service models

hltbra:
This enhances metadata retrieval and it's more complete. We should do physical resource IDs for anything that supports CloudFormation at some point.

I found that Cluster uses name and Service uses ARNs looking at existing CFN stacks with those resources.


spulec:
Thanks!

PR #814, opened 2017-01-24
Implement list_objects_v2 for S3 buckets

### LINKED ISSUE #734: Boto3 S3 list_objects_v2 response is different with moto
With Boto3, the response returned by the S3 client list_objects_v2 method while mocking is missing the 'KeyCount' attribute that you get with the real response from the S3 client.

Example without mocking:

```
session = boto3.Session( profile_name='default' )
s3_client = session.client('s3')

response = s3_client.list_objects_v2(
                Bucket=s3_bucket,
                Delimiter='/',
                Prefix=prefix,
                MaxKeys=1
            )
print(response)

Response:
{
 'Prefix': 'prefix',
 'Delimiter': '/',
 'Name': 'test',
 'MaxKeys': 1000,
 'Contents': [
    {'StorageClass': 'STANDARD',
     'ETag': '"d42d8ce98f00h204e5800998edf8427e"',
     'Size': 0,
     'LastModified':
         datetime.datetime(2016, 8, 10, 18, 14, 11, tzinfo=tzutc()),
     'Key': 'prefix'}
    ],
 'KeyCount': 3,
 'IsTruncated': False,
 'ResponseMetadata': {
        'HTTPStatusCode': 200,
        'RequestId': 'C09582522733BC30',
        'HostId': 
            'Y61Adc1makUntal84YC9kWFnZm3t63U1nZEaGdO2g3kFMzYu3Mf18wb+zNfjue3J3V+OWBTL2rI='
    },
 'CommonPrefixes': [
     {'Prefix': 'prefix/test1/'},
     {'Prefix': 'prefix/test2/'}
    ]
}
```

With mocking:

```
Response:
{
 'Contents': [
    {'LastModified':
        datetime.datetime(
            2016, 10, 13, 16, 13, 45, 371000, tzinfo=tzutc()),
     'ETag': '"b11a8db164e9754105b7a99be72e3fe5"',
     'StorageClass': 'STANDARD',
     'Owner': {
        'DisplayName': 'webfile', 
        'ID': '75aa57f09aa0c8caeab4f8c24e99d10f8e7faeebf76c078efc7c6caea54ba06a'},
     'Key': 'prefix/testfile.txt',
     'Size': 11}
    ],
 'Delimiter': '/',
 'IsTruncated': False,
 'Prefix': 'prefix/',
 'Name': 'test',
 'MaxKeys': 1000,
 'ResponseMetadata': {'HTTPStatusCode': 200}
}
```

snordhausen:
This change adds a separate implementation for list_objects_v2. Previously, it used the same implementation as list_objects.

The unit tests still work if you run them without the mock_s3 decorator (and change the bucket name) using boto3 version 1.4.4 and botocore version 1.5.4 against the real AWS. Which means they only check for behavior that Amazon/boto3 actually implement.

This adds/fixes the following things:
- Add missing KeyCount in result (fixes #734).
- Do not hard code MaxKeys to 1000.
- Truncate result if it has more than MaxKeys items. Set IsTruncated
    and NextContinuationToken accordingly.
- Support the StartAfter parameter.
- Return Owner information only when FetchOwner=True is given.
- "Prefix" in response is now "" instead of None when omitted in request.
- "Delimiter" is now omitted from response when not given in request.

### DIFF
--- a/moto/s3/responses.py
+++ b/moto/s3/responses.py
@@ -215,6 +215,8 @@ class ResponseObject(_TemplateEnvironmentMixin):
                 delimiter='',
                 is_truncated='false',
             )
+        elif querystring.get('list-type', [None])[0] == '2':
+            return 200, headers, self._handle_list_objects_v2(bucket_name, querystring)
 
         bucket = self.backend.get_bucket(bucket_name)
         prefix = querystring.get('prefix', [None])[0]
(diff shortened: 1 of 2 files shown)


spulec:
Great, thank you!

PR #817, opened 2017-01-25
Reload the server on a file change

jgelens:
Using the default Werkzeug reloader. I prefer a reload using SIGHUP, but that's kinda difficult with Werkzeug. This works by "touching" one of the moto module files. This can be used for resetting a Moto server without actually killing it (useful for Docker for example)


spulec:
Awesome, thanks!

PR #819, opened 2017-01-31
make instanceTenancy configurable for VPCs

detiber:
Adds the ability to set InstanceTenancy for VPCs.


detiber:
@spulec It looks like coveralls permissions need to be re-enabled, it's not able to connect to github to show the actual lines covered/missed by tests.

spulec:
Thanks!

PR #822, opened 2017-02-02
make get_all_security_groups filter AND match group ids, not OR them

Dawil:
Currently, if you call `get_all_security_groups()` with values for `group_ids` and `filters`, security groups will be returned if they match the `group_ids` OR the `filters`. This patch changes it so that they will be returned if they match the `group_ids` AND the `filters`.  A test case is also included so you can confirm the old behavior fails the unit test.


spulec:
Thanks!

PR #824, opened 2017-02-03
install server version of moto for Dockerfile

joekiller:
Use the server install of moto for Dockerfile.

This allows the ~~flake~~ flask prerequisite of the server to be installed.


spulec:
Thanks!

PR #826, opened 2017-02-10
KMS encryption under Python 3

JackDanger:
This upgrades the KMS encrypt and decrypt endpoints to work
under both Python 2 and 3


spulec:
Thanks!

PR #829, opened 2017-02-14
Ensure boto3's managed transfer won't use threads

### LINKED ISSUE #816: upload file with boto, download it with boto3: file gets corrupted (wrong md5 sum)
Hi,

The following code uploads a file to a mock S3 bucket using boto, and downloads the same file to the local disk using boto3. I apologize for bringing both of the libraries into this, but the code I am testing in real life still uses both (definitely trying to get rid of all the boto code and fully migrate to boto3 but that isn't going to happen right away).

What happens is that the resulting file does not have the same md5 sum as the original file so it has been corrupted at some point (not sure if it was during the boto upload or the boto3 download).

This seems to be an issue with moto because if I comment out the line `@moto.mock_s3` (using 'real' S3) the script works fine (I also need to change the bucket name to a unique one to avoid collisions).

The script keeps looping (doing the upload/download/md5sum comparison) until it fails (because in my real project this would not happen every time) but this test script seems to fail (for me anyway) on the first attempt every time.

The test file that it uploads/downloads is available [here]( https://s3-us-west-2.amazonaws.com/demonstrate-moto-problem/K158154-Mi001716_S1_L001_R1_001.fastq.gz).

You can download it with:

    curl -O  https://s3-us-west-2.amazonaws.com/demonstrate-moto-problem/K158154-Mi001716_S1_L001_R1_001.fastq.gz

At this point if you run md5sum on it you should get `6083801a29ef4ebf78fbbed806e6ab2c`:

```
$ md5sum K158154-Mi001716_S1_L001_R1_001.fastq.gz
6083801a29ef4ebf78fbbed806e6ab2c  K158154-Mi001716_S1_L001_R1_001.fastq.gz
```

Here is the test script (`motoprob.py`):

```python
import sys
import os
import hashlib
import moto
import boto
import boto3

def md5(fname):
    hash_md5 = hashlib.md5()
    with open(fname, "rb") as f:
        for chunk in iter(lambda: f.read(4096), b""):
            hash_md5.update(chunk)
    return hash_md5.hexdigest()



@moto.mock_s3
def doit():
    # upload file to s3
    conn = boto.connect_s3()
    bkt = conn.create_bucket("mybucket")
    key = boto.s3.key.Key(bkt)
    key.key = "foo/bar.fastq.gz"
    print("Uploading...")

    # You can get this file from:
    #  https://s3-us-west-2.amazonaws.com/demonstrate-moto-problem/K158154-Mi001716_S1_L001_R1_001.fastq.gz
    key.set_contents_from_filename("K158154-Mi001716_S1_L001_R1_001.fastq.gz")

    # download it again
    dlfile = "bar.fastq.gz"
    if os.path.exists(dlfile):
        os.remove(dlfile)

    print("Downloading...")

    client = boto3.client('s3')
    client.download_file(Bucket="mybucket",
      Key="foo/bar.fastq.gz", Filename="bar.fastq.gz")


    md5sum = md5(dlfile)
    if not md5sum == "6083801a29ef4ebf78fbbed806e6ab2c":
        print("Incorrect md5sum! {}").format(md5sum)
        sys.exit(1)


while True:
    doit()

```

Version info:
```
$ pip freeze |grep oto
boto==2.42.0
boto3==1.4.0
botocore==1.4.48
moto==0.4.29

$ python --version
Python 2.7.12

$ uname -a
Linux f51bec2ad3be 4.9.4-moby #1 SMP Wed Jan 18 17:04:43 UTC 2017 x86_64 x86_64 x86_64 GNU/Linux

$ more /etc/lsb-release
DISTRIB_ID=Ubuntu
DISTRIB_RELEASE=16.04
DISTRIB_CODENAME=xenial
DISTRIB_DESCRIPTION="Ubuntu 16.04.1 LTS"
```

Other ways to see that the resulting file is not the same as the original:

```
$ diff bar.fastq.gz /data/2016-10-27-PT140/K158154-Mi001716_S1_L001_R1_001.fastq.gz
Binary files bar.fastq.gz and /data/2016-10-27-PT140/K158154-Mi001716_S1_L001_R1_001.fastq.gz differ


$ zcat bar.fastq.gz > bar.fastq # this works for the original file

gzip: bar.fastq.gz: invalid compressed data--crc error

gzip: bar.fastq.gz: invalid compressed data--length error
```

snordhausen:
fixes #816 (and potentially some other tickets related to managed transfer methods)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,7 +4,7 @@ ADD . /moto/
 ENV PYTHONUNBUFFERED 1
 
 WORKDIR /moto/
-RUN python setup.py install
+RUN pip install ".[server]"
 
 CMD ["moto_server"]
 
--- a/moto/apigateway/__init__.py
+++ b/moto/apigateway/__init__.py
@@ -1,12 +1,6 @@
 from __future__ import unicode_literals
 from .models import apigateway_backends
-from ..core.models import MockAWS
+from ..core.models import MockAWS, base_decorator
 
 apigateway_backend = apigateway_backends['us-east-1']
-
-
-def mock_apigateway(func=None):
-    if func:
-        return MockAWS(apigateway_backends)(func)
-    else:
-        return MockAWS(apigateway_backends)
+mock_apigateway = base_decorator(apigateway_backends)
--- a/moto/autoscaling/__init__.py
+++ b/moto/autoscaling/__init__.py
@@ -1,12 +1,6 @@
 from __future__ import unicode_literals
 from .models import autoscaling_backends
-from ..core.models import MockAWS
+from ..core.models import MockAWS, base_decorator
 
 autoscaling_backend = autoscaling_backends['us-east-1']
-
-
-def mock_autoscaling(func=None):
-    if func:
-        return MockAWS(autoscaling_backends)(func)
-    else:
-        return MockAWS(autoscaling_backends)
+mock_autoscaling = base_decorator(autoscaling_backends)
--- a/moto/awslambda/__init__.py
+++ b/moto/awslambda/__init__.py
@@ -1,13 +1,7 @@
 from __future__ import unicode_literals
 from .models import lambda_backends
-from ..core.models import MockAWS
+from ..core.models import MockAWS, base_decorator
 
 
 lambda_backend = lambda_backends['us-east-1']
-
-
-def mock_lambda(func=None):
-    if func:
-        return MockAWS(lambda_backends)(func)
-    else:
-        return MockAWS(lambda_backends)
+mock_lambda = base_decorator(lambda_backends)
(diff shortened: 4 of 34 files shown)


JackDanger:
@snordhausen This looks useful! Can you please rebase on the latest code since the httpretty patching has changed significantly?

snordhausen:
I'm not sure if this feature is still needed since #848 talks about deprecating HTTPretty. Can @spulec please shed some light on this?

spulec:
My guess is that this is no longer needed with the current system

JackDanger:
I'm closing this until someone mentions they need it.

leopepe:
It is 2021 and this issue still happening. In my setup I am using pyspark 3.1.2 and moto 2.0.2

PR #831, opened 2017-02-14
Feat: ECS container status updating

gjtempleton:
Features:
* Added initial method to set container instance states to `DRAINING` or `ACTIVE`
* Updated boto3 dependency to the corresponding version introducing the new method

Fixes:
* PEP8 fixes - minor issues as well as a couple of duplicate methods - have kept the newer version where appropriate

References:
* https://aws.amazon.com/about-aws/whats-new/2017/01/amazon-ecs-supports-container-instance-draining/
* https://boto3.readthedocs.io/en/latest/reference/services/ecs.html#ECS.Client.update_container_instances_state

I've updated the run_task method to only place tasks on active instances for now but the chunk of work to fully implement container draining as well as resource checking when placing tasks/services will be made in a different merge request as it's a fairly substantial piece.

### DIFF
--- a/moto/ecs/models.py
+++ b/moto/ecs/models.py
@@ -363,9 +363,10 @@ class EC2ContainerServiceBackend(BaseBackend):
         container_instances = list(self.container_instances.get(cluster_name, {}).keys())
         if not container_instances:
             raise Exception("No instances found in cluster {}".format(cluster_name))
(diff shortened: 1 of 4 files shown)


gjtempleton:
I realise the test for instance state setting is a bit unwieldy, but I wanted to cover testing idempotency as well as switching between the two states to cover the real world behaviour as closely as possible.

spulec:
This is great, thanks!

PR #832, opened 2017-02-15
Simple event driven framework and some fixes

argos83:
Hi!
I've added a simple event driven functionality  (observer pattern), to allow users to extend, control, and customize what happens when AWS endpoints are invoked, specially when running moto in server mode. In relation to this, I've also added another cli swicth (for server mode): `-s path/to/script.py` or `--script path/to/script.py` which causes the given script to be loaded, and if it has a `start` method it will be invoked passing the event publisher as an argument, so users can subscribe to events of interest.

Perhaps an example (why I personally need this for) can help understand what the use case is:
I need to mock RDS (using the rds2 backend) but also, whenever a `create rds instance` request is issued, as a result I want to create and start a real postgress instance, and provide the right address in the response so the SUT can connect to it.

So I can define the following script somewhere. E.g. `postgres_provision.py`

```
import subprocess

from moto.core.publisher import EventTypes

def start(publisher):
    publisher.subscribe(provision_database, EventTypes.RDS2_DATABASE_CREATED)

def provision_database(event, database):
    # observers are invoked passing the event id (so you can tell from different events 
    # in the case that you subscribe the same observer to multiple events), and data related
    # to the event, in this case the database model instance

    database.address_gen = lambda *_args: '127.0.0.1'

    args = ["docker", "run", "--rm", "-p", "%s:%s" % (database.port, 5432),
            "-e", "POSTGRES_PASSWORD=%s" % database.master_user_password,
            "-e", "POSTGRES_USER=%s" % database.master_username,
            "postgres:%s" % database.engine_version]
    subprocess.Popen(args)
```

and run the server this way: `moto_server --script path/to/postgres_provision.py rds2`

In this PR, `RDS2_DATABASE_CREATED` is the only event I've created, observers will be invoked with the instance of the `Database` model. But it's easy to keep adding more and more events anywhere where functionality could be extended.

This PR also includes some other fixes, but I'll add comments inline about them.
I haven't bumped the version yet, nor updated `README.md`. I'll wait for your input first.


moto/rds2/models.py
```diff
@@ -27,19 +25,23 @@ def __init__(self, **kwargs):
...
-                                        "oracle-ee": "11.2.0.4.v3",
-                                        "sqlserver-ee": "11.00.2100.60.v1",
-                                        "sqlserver-se": "11.00.2100.60.v1",
-                                        "sqlserver-ex": "11.00.2100.60.v1",
-                                        "sqlserver-web": "11.00.2100.60.v1",
-                                        "postgres": "9.3.3"
-                                        }
-        if not self.engine_version and self.engine in self.default_engine_versions:
-            self.engine_version = self.default_engine_versions[self.engine]
+
```
argos83: if the port was not specified in the request, moto was returning None instead of the default port for the selected engine (which triggered and exception in the client libraries)

moto/rds2/models.py
```diff
@@ -207,7 +212,7 @@ def to_xml(self):
 
     @property
     def address(self):
-        return "{0}.aaaaaaaaaa.{1}.rds.amazonaws.com".format(self.db_instance_identifier, self.region)
+        return self.address_gen(self.db_instance_identifier, self.region)
```
argos83: it behaves exactly the same as before, but now we can control the address if necessary (e.g. I want to return the address of a real database)

moto/rds2/models.py
```diff
@@ -832,12 +838,6 @@ def modify_db_parameter_group(self, db_parameter_group_name, db_parameter_group_
 
         return db_parameter_group
 
-    def delete_db_parameter_group(self, db_parameter_group_name):
```
argos83: this exact same method was defined twice in this class (copy paste error I guess)

moto/server.py
```diff
@@ -14,9 +14,13 @@
 from werkzeug.serving import run_simple
 
 from moto.backends import BACKENDS
+from moto.compat import load_module
+from moto.core.publisher import default_publisher
 from moto.core.utils import convert_flask_to_httpretty_response
 
+
 HTTP_METHODS = ["GET", "POST", "PUT", "DELETE", "HEAD", "PATCH"]
+ADDITIONAL_URL_BASES = [r"https?://{0}\.(.+)\.xip\.io"]  # see http://xip.io/
```
argos83: This is a small feature. I'd like to start moto in server mode with support for all backends (instead of single service).
However for that to be usable it will require hacking around with DNS or /etc/hosts.. so all the *.amazon.aws names resolve to localhost.
With this change I can set up the endpoints in the client like `http://rds2.127.0.0.1.xip.io`, `http://ec2.127.0.0.1.xip.io`, etc, etc. And moto server will be able to forward requests to the corresponding backends.

tests/test_rds2/test_rds2.py
```diff
@@ -31,6 +30,32 @@ def test_create_database():
...
+                                       Port=1234,
+                                       DBSecurityGroups=["my_sg"])['DBInstance']
+
+    instance['MasterUsername'].should.equal("toor")
+    instance['Endpoint']['Port'].should.equal(4567)
+    instance['Endpoint']['Address'].should.equal("db-master-1.us-west-2.somewhere-else.com")
+    default_publisher.reset()
+
+
+@disable_on_py3()
```
argos83: FYI, all these tests seem to work with python 3 (at least python 3.5). I've kept them like that since I don't know the reason why they've been disabled.

JackDanger:
@argos83 This is good work, sorry I'm just seeing it (I'm new around here). I have two dumb questions for you:

1) Can this same feature be implemented via SNS within Moto?
2) Is this still useful to you?

JackDanger:
@argos83 I'm going to close this PR for now because it's stale but that is in no way a reflection of it's usefulness. If you have a minute to come back and address some questions in this PR please reopen it.

argos83:
Hi @JackDanger, sorry for disappearing.

My answers to your questions:

> @argos83 This is good work, sorry I'm just seeing it (I'm new around here). I have two dumb questions for you:
> 
> 1. Can this same feature be implemented via SNS within Moto?

To be honest, I'm not sure what's the current state of SNS in Moto. But anyways I doubt it would allow us to hook into very specific operations. E.g. we could bind SNS with RDS in moto to get a notification when an RDS instance is created, but in my particular case I need to hook my function right after I get the creation request message, but before the response is sent back to the client. 

> 2. Is this still useful to you?

Not to me anymore since I moved to a different team in my company, but my former team is using this (from the branch in my fork atm).

PR #833, opened 2017-02-16
Route53: allow hosted zone id as well when creating record sets

Dawil:
Currently a `KeyError: u'HostedZoneName'` is thrown when supplying a `HostedZoneId` to a RecordSetGroup in a CloudFormation template and only a `HostedZoneName` will work. Both of these options are supposed to work, as per [here](http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-route53-recordsetgroup.html#cfn-route53-recordsetgroup-hostedzoneid).

This PR updates `RecordSetGroup.create_from_cloudformation_json()` to perform an identical check to the one in `RecordSet`. This behavior could be abstracted into a function if desired but all occurrences in `moto/route53/models.py` appear to be handled now.

### DIFF
--- a/moto/route53/models.py
+++ b/moto/route53/models.py
@@ -213,8 +213,11 @@ class RecordSetGroup(object):
     def create_from_cloudformation_json(cls, resource_name, cloudformation_json, region_name):
         properties = cloudformation_json['Properties']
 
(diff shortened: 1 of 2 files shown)


spulec:
Great, thanks!

PR #836, opened 2017-02-20
Correct IAM list_server_certs template that was based off incorrect docs

Dawil:
This removes the `<ServerCertificateMetadata>` tag from within the <member> elements of the server certificate list.

This occurred because the documentation for this method here https://docs.aws.amazon.com/IAM/latest/APIReference/API_ListServerCertificates.html

say the return type is this  `ServerCertificateMetadataList.member.N`

but the sample response incorrectly include a `<ServerCertificateMetadata>`. I've sent feedback to the AWS docs telling them to fix their stuff but this also needs to be fixed. I haven't checked other templates with <member> tags in them, as they may be prone to this same problem.


spulec:
Thanks!

PR #841, opened 2017-02-22
Elastic IP PhysicalResourceId should always be its public IP

rouge8:
According to the [CloudFormation `Ref` docs][docs], the `Ref` return value (and physical ID of the resource) for an Elastic IP is its public IP address.

[docs]: http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/intrinsic-function-reference-ref.html


spulec:
Yes, thanks for fixing that!

PR #843, opened 2017-02-23
make cloudformation update stack use parameters provided

Dawil:
Currently when updating a stack, if parameters are provided they are ignored. This patch reads the parameters from the Response in the same way that `create_stack` does, passes it on through to the underlying `ResourceMap.update()` method and updates the `input_parameters` attribute. I'm not sure if there might be a preferred way to do this, but this seems like a reasonably small change.

As always, I'm happy to discuss or incorporate feedback to make this PR more conforming.


spulec:
Great, thank you!

PR #846, opened 2017-02-24
Cast desired capacity for cloudformation asg to int

william-richard:
Cloudformation passes MaxSize, MinSize and DesiredCapacity as strings, but
we want to store them as ints.  Also includes tests of this fix, to help
avoid regression.


spulec:
Great, thank you!

PR #851, opened 2017-03-02
Fix lambda stdout/stderr mocking

2rs2ts:
Originally, the code was setting sys.stdout and sys.stderr back to the
original, official forms, but this breaks idioms like mocking stdout to
capture printing output for tests. So instead, we will reset sys.stdout
and sys.stderr to what they were before running the lambda function, so
that in case someone is mocking stdout or stderr, their tests won't
break.

I realize that technically someone could write code that overwrites these variables, but I think it is unlikely. Let me know if this is a dealbreaker.

### DIFF
--- a/moto/awslambda/models.py
+++ b/moto/awslambda/models.py
@@ -135,6 +135,8 @@ class LambdaFunction(object):
             print("Exception %s", ex)
 
         try:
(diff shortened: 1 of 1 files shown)


spulec:
Looks good, thanks!

PR #852, opened 2017-03-03
Add event IDs to CF Stack events

2rs2ts:
So that events can be uniquely identified.

I tried to match the format documented here: http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/using-cfn-listing-event-history.html


spulec:
Thanks!

PR #853, opened 2017-03-03
Return CF Stack events in reverse chronological order

2rs2ts:
This is how the AWS API works: http://boto3.readthedocs.io/en/latest/reference/services/cloudformation.html#CloudFormation.Client.describe_stack_events

I wasn't sure whether it would be better to make the internal representation be in reverse chronological order instead of just changing the response. I could go either way. Let me know if you want me to change it!


spulec:
This is great, thanks!

PR #858, opened 2017-03-13
Add a dashboard

spulec:
moto_and_new_tab


moto/core/models.py
```diff
@@ -208,9 +210,36 @@ def dec(f):
         return dec
 
 
+model_data = defaultdict(dict)
+class InstanceTrackerMeta(type):
```
hltbra: This registry implementation looks great, very concise.

PR #860, opened 2017-03-13
Add ListHostedZonesByName

2rs2ts:
I implemented both the lookup and the sorting for when you don't specify a `dnsname`.

Added a new test and uncommented/fixed a piece of test code that can now run because the method is now supported.


spulec:
Thanks!

PR #861, opened 2017-03-13
Ensure SQS property WaitTimeSeconds is an integer

scawley-r7:
This just ensure that the `WaitTimeSeconds` property is in integer.


spulec:
Thanks!

PR #865, opened 2017-03-13
update RDS responses to return DBInstanceArn

### LINKED ISSUE #864: RDS responses don't return DBInstanceArn
RDS/RDS2 endpoints such as `create_db_instance`, `describe_db_instances`, etc. do not return the DBInstanceArn attribute in the response.

This makes it difficult to use moto when attempting to test the following code:

```
import boto3

rds_conn = boto3.client('rds')
for instance in rds_conn.describe_db_instances()['DBInstances']:
    for tag in rds_conn.list_tags_for_resource(ResourceName=instance['DBInstanceArn'])['TagList']:
        print(tag)
```

Because `DBInstanceArn` isn't returned in the response from `describe_db_instances()`, it is difficult to test code that relies on listing tags for resources that are listing (unless constructing the ARN based on user input).

This is present in moto==0.4.31

adamstauffer:
Fixes #864

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -44,3 +44,4 @@ Moto is written by Steve Pulec with contributions from:
 * [Jean-Baptiste Barth](https://github.com/jbbarth)
 * [Tom Viner](https://github.com/tomviner)
 * [Justin Wiley](https://github.com/SectorNine50)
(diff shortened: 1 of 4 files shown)


spulec:
Great, thank you!

PR #867, opened 2017-03-14
Be flexible with Route53 Hosted Zone IDs with /hostedzone/ prefix

2rs2ts:
We will continue to store just the unique ID, but since the AWS API
returns `/hostedzone/<id>`, we should accept attempts to pass that back.
For example, both just the ID as well as `/hostedzone/<id>` work for
specifying the `HostedZoneId` of a `ResourceRecordSet` in CloudFormation. So
we should support that too.

### DIFF
--- a/moto/route53/models.py
+++ b/moto/route53/models.py
@@ -277,7 +277,7 @@ class Route53Backend(BaseBackend):
         return self.zones.values()
 
     def get_hosted_zone(self, id_):
-        return self.zones.get(id_.lstrip("/hostedzone/"))
+        return self.zones.get(id_.replace("/hostedzone/", ""))
 
     def get_hosted_zone_by_name(self, name):
         for zone in self.get_all_hosted_zones():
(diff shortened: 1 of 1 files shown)


2rs2ts:
I'm not sure why the build is failing with DnsServerErrors. On my local machine the errors happen too but they are sporadic. I never get the same set of failures. Sometimes I get none.

2rs2ts:
@spulec Can you help me figure out why these tests are failing? I'm not sure why it would happen, but I'm starting to believe moto has race conditions in it because my tests in my own projects are intermittently failing when trying to look up hosted zones. Seems like a race condition. If you can't think of a reason why I'll try to make a minimal reproducible example...

moto/route53/models.py
```diff
@@ -277,18 +277,15 @@ def get_all_hosted_zones(self):
         return self.zones.values()
 
     def get_hosted_zone(self, id_):
-        return self.zones.get(id_)
+        return self.zones.get(id_.lstrip("/hostedzone/"))
```
spulec: It looks like `lstrip()` doesn't behave like this. See [here](https://docs.python.org/2/library/string.html#string.lstrip).

In the cases where `id_` starts with one of the letters in `/hostedzone/`, it is getting removed. For example, `'e67ec7b9'.lstrip("/hostedzone/")` ->`'67ec7b9'`

One option is to change to something like `id_.lstrip("/hostedzone")[0]`.

FWIW, I've been bitten by the unintuitiveness of `lstrip()` in the past many times as well.

2rs2ts: Ah, of course. I literally just bumped into this earlier today too, and hadn't learned my lesson.

spulec:
Looks great, thanks!

PR #872, opened 2017-03-18
Allow CloudFormation stack tags to be updated

2rs2ts:
Limitations:

* does not update the tags of the resources in the stack. that can be
implemented later.
* does not support the supposed feature of clearing tags by passing an
empty value that boto3 mentions in its documentation. I could not find
anything in the request body to indicate when an empty value was passed.


spulec:
Awesome. Thank you!

PR #874, opened 2017-03-21
Catch KeyError instead of ValueError

bkochendorfer:
Not catching the correct error here I believe. This allows us to run moto_server locally without passing in a `HTTP_AUTHORIZATION` environment variable.


2rs2ts:
There's a chance that the ValueError was for tuple unpacking.

spulec:
Thanks, I've added the KeyError.

PR #881, opened 2017-03-27
Add mfa device endpoints to iam backend.

smarlowucf:
- Add mfa device class
- Add mfa devices dictionary to user class
- Add responses, endpoints and tests


smarlowucf:
Closes #879

spulec:
This is great, thanks!

PR #891, opened 2017-04-06
Fix SQS message polling to abort after wait_seconds_timeout

whummer:
This PR fixes SQS message polling to abort after `wait_seconds_timeout`. In the current implementation, we loop until a new message becomes available in the queue, and never actually abort the loop based on the timeout value.

Steps to reproduce the issue:
1) create queue
2) send message to the queue
3) run `sqs receive-message` against the queue -> returns the message immediately
4) run `sqs receive-message --wait-time-seconds 1` against the queue (immediately after previous command)
   - expected behavior: call returns after 1 second and returns empty result
   - actual behavior: call waits roughly 30 seconds (until `visibility_timeout` expires and the message gets re-added to the queue) and returns the message

### DIFF
--- a/moto/sqs/models.py
+++ b/moto/sqs/models.py
@@ -289,6 +289,10 @@ class SQSBackend(BaseBackend):
 
         # queue.messages only contains visible messages
         while True:
+
+            if result or (wait_seconds_timeout and unix_time() > polling_end):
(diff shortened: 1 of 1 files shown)


spulec:
thanks!

PR #895, opened 2017-04-12
Container Instance Resource implementation

gjtempleton:
*Features:*
* Now implements a basic model of Container Instance Resources (providing defaults based on C4.Xlarges - future work to base this on the EC2 instance type to follow) - existing tests updated and new ones added for this
* Check for ability to place a task on each instance - refuse to place on a task without the `MEMORY`, `CPU` or `PORTS` reminding to place the task on the instance - new tests for this
* Reserve resources on start of a task, free up on removal - new tests added for this

*Bug Fixes:*
* Now able to use container instance ARNs for `describe_container_instances` - one new test covers this case

I'm not sure I've followed your preferred practice in all the cases, creating a few private helper methods for calculating resources given the AWS implementation as a list of dicts rather than duplicate code, happy to rework if need be.

### DIFF
--- a/moto/ecs/models.py
+++ b/moto/ecs/models.py
@@ -1,6 +1,6 @@
 from __future__ import unicode_literals
 import uuid
-from random import randint, random
+from random import random
 
 from moto.core import BaseBackend, BaseModel
 from moto.ec2 import ec2_backends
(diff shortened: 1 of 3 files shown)


spulec:
This is great! Thanks for the contribution.

Issue #821, opened 2017-02-01
Feature request: support S3 TemporaryRedirect and PermanentRedirect responses


Fedorof:
I would like to simulate S3 behavior for redirects. It could be achieved with a server that always returns the same type of response with information of the real moto server endpoint.

Say, if moto server listens to `localhost:5001`, the redirect server always returns 301 or 307 response with headers and contents provided below.

An example of the response headers:

    HTTP/1.1 301 Moved Permanently
    x-amz-bucket-region: us-east-1
    x-amz-request-id: CD12xxxxxxxxxxxx
    x-amz-id-2: IVdExxxxxxxxxxxxxxxxxxxxxxx
    Content-Type: application/xml
    Transfer-Encoding: chunked
    Date: Mon, 30 Jan 2017 13:07:10 GMT
    Server: AmazonS3

The body for PermanentRedirect (301 status code):

    <Error>
      <Code>PermanentRedirect</Code>
      <Message>The bucket you are attempting to access must be addressed using the specified endpoint. Please send all future requests to this endpoint.</Message>
      <Bucket>foo</Bucket>
      <RequestId>CD12xxxxxxxxxxxx</RequestId>
      <HostId>IVdExxxxxxxxxxxxxxxxxxxxxxx</HostId>
      <Endpoint>localhost:5001</Endpoint>
    </Error>

The body for TemporaryRedirect (307 status code):

    <Error>
      <Code>TemporaryRedirect</Code>
      <Message>Please re-send this request to the specified temporary endpoint. Continue to use the original request endpoint for future requests.</Message>
      <RequestId>63B2xxxxxxxxxxxx</RequestId>
      <Bucket>foo</Bucket>
      <HostId>S3nuxxxxxxxxxxxxxxxxxxxxxxx</HostId>
      <Endpoint>localhost:5001</Endpoint>
    </Error>

spulec:
Noted. I don't have a particular use for this so it isn't high on my list, but pull requests are very much welcome if someone has time.

spulec:
Fixed with #901

PR #901, opened 2017-04-14
s3: handle WebsiteRedirectLocation

niamster:
S3 keys has a special header x-amz-website-redirect-location/WebsiteRedirectLocation which results into permanent redirection when a key is accessed.

boto3 put_object/get_object key: WebsiteRedirectLocation
s3 header name: x-amz-website-redirect-location

spulec/moto#821


spulec:
Thanks!

PR #902, opened 2017-04-14
Return the revision in ecs.register_task_definition

mvantellingen:
This matches boto, see
http://boto3.readthedocs.io/en/latest/reference/services/ecs.html#ECS.Client.register_task_definition


spulec:
Thanks!

PR #907, opened 2017-04-18
Implement support for SSM parameter store

mvantellingen:
This commit adds initial support for the Simple System Manager client.

It currently only mocks the following api endpoints:
 - delete_parameter()
 - put_parameter()
 - get_parameters()


mvantellingen:
For reference, i'm using this in https://github.com/LabD/python-param-store/blob/master/tests/test_stores.py

spulec:
This is great! Thank you for your work!

PR #909, opened 2017-04-21
Implementation of delete_pipeline

apiasetskyi:
Also added basic tags for Data Pipeline


spulec:
This is great! Thanks for your help!

PR #916, opened 2017-04-25
Added validation for record names in route53 (boto3)

code-haven:
Changes in this PR:
----

- Added validation for record set creation in route53 (boto3)
- Added a few test cases for ```change_resource_record_sets``` function(boto3) which is used to create, update and delete record sets in Route53.

Steps to reproduce the issue with record set creation
---
* Create a hosted zone with ```Name=database.```
* Create a invalid record```Name=prod.redis.scooby.doo``` - moto treats it as a valid request and returns a 201.
 
  The same request using boto3(with real credentials) would give:
```
InvalidChangeBatch: An error occurred (InvalidChangeBatch) when calling the 

ChangeResourceRecordSets operation: RRSet with DNS name prod.redis.scooby.doo. is not 

permitted in zone database.
```

### DIFF
--- a/moto/route53/responses.py
+++ b/moto/route53/responses.py
@@ -112,6 +112,17 @@ class Route53(BaseResponse):
             for value in change_list:
                 action = value['Action']
                 record_set = value['ResourceRecordSet']
+
+                cleaned_record_name = record_set['Name'].strip('.')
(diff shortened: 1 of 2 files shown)


spulec:
This is great, thank you!

PR #919, opened 2017-04-27
MD5 calculation of SQS message attributes

### LINKED ISSUE #918: Hardcoded Message Attributes MD5 checksum causes error in AWS Java SDK
I've been trying to get set up to use Atlassian's localstack for local development of a Java application. When trying to send SQS messages that use MessageAttributes I receive errors from the SDK that the returned MD5 checksum does not match what was expected:

```
com.amazonaws.AmazonClientException: MD5 returned by SQS does not match the calculation on the original request. (MD5 calculated by the message attributes: "235c5c510d26fb653d073faed50ae77c", MD5 checksum returned: "324758f82d026ac6ec5b31a3b192d1e3")
```

In the [issue I filed with localstack](https://github.com/atlassian/localstack/issues/78), one of the maintainers noted that moto is using a [hard-coded MD5 checksum](https://github.com/spulec/moto/blob/896f040fca975865c1d441d90016446c72db3ed4/moto/sqs/responses.py#L343).

Would it be possible for moto to calculate and return the expected checksum value?

JackDanger:
This implements the same MD5 hashing pattern as implemented in the Ruby
and Java AWS SDKs

Fixes #918

### DIFF
--- a/tests/test_sqs/test_sqs.py
+++ b/tests/test_sqs/test_sqs.py
@@ -43,10 +43,44 @@ def test_get_inexistent_queue():
 def test_message_send():
     sqs = boto3.resource('sqs', region_name='us-east-1')
     queue = sqs.create_queue(QueueName="blah")
-    msg = queue.send_message(MessageBody="derp")
-
+    msg = queue.send_message(
+        MessageBody="derp",
+        MessageAttributes={
+            'timestamp': {
+                'StringValue': '1493147359900',
+                'DataType': 'Number',
+            }
+        }
+    )
     msg.get('MD5OfMessageBody').should.equal(
         '58fd9edd83341c29f1aebba81c31e257')
+    msg.get('MD5OfMessageAttributes').should.equal(
+        '235c5c510d26fb653d073faed50ae77c')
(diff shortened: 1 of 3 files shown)


spulec:
Hmmmm, Travis seems to be unhappy about something we are base64 decoding.

tests/test_sqs/test_sqs.py
```diff
@@ -199,7 +234,7 @@ def test_send_message_with_attributes():
     message = queue.new_message(body)
     message_attributes = {
         'test.attribute_name': {'data_type': 'String', 'string_value': 'attribute value'},
-        'test.binary_attribute': {'data_type': 'Binary', 'binary_value': 'binary value'},
+        'test.binary_attribute': {'data_type': 'Binary', 'binary_value': base64.b64encode('binary value')},
```
whummer: @JackDanger For Python 3.x compatibility, I think this should be `base64.b64encode(b'binary value')` - see failing test on travis-ci: https://travis-ci.org/spulec/moto/jobs/231174489#L1635

JackDanger: Thanks for your help, @whummer !

PR #921, opened 2017-04-27
Fixing metadata key on s3 operation response

### LINKED ISSUE #920: Metadata being returned with some keys with wrong format
**Problem description**
When using AWS Java SDK against Moto an issue was uncovered. Metadata is being returned with keys like "Etag" when they should be returned with a specific format (in this example, "ETag").

**Symptoms**
The above mentioned problem leads the Java SDK to generate an NPE as it's counting with a header called "ETag" and not "Etag".

**Steps to reproduce**
- Download https://github.com/awslabs/aws-java-sample.
- Change the sample to use Moto Server by setting the endpoint. Locally, it would be something like ```s3.setEndpoint("http://127.0.0.1:5000");```
- Start Moto Server at ```127.0.0.1:5000```
- Run the sample

**Output**
```
Exception in thread "main" java.lang.NullPointerException
	at com.amazonaws.util.BinaryUtils.fromHex(BinaryUtils.java:61)
	at com.amazonaws.services.s3.AmazonS3Client.putObject(AmazonS3Client.java:1450)
	at com.amazonaws.samples.S3Sample.main(S3Sample.java:107)
```

picadoh:
ETag metadata key is being returned as "Etag" instead of "ETag". This leads to issues in some AWS SDKs using MotoServer. This change fixes the issue by updating the key to the correct format.

This closes #920

### DIFF
--- a/moto/s3/models.py
+++ b/moto/s3/models.py
@@ -91,7 +91,7 @@ class FakeKey(BaseModel):
     @property
     def response_dict(self):
         res = {
-            'Etag': self.etag,
+            'ETag': self.etag,
             'last-modified': self.last_modified_RFC1123,
             'content-length': str(len(self.value)),
         }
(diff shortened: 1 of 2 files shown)


JackDanger:
@spulec this LGTM if you have a second to check it out. Nice work, @picadoh !

spulec:
Thanks!

PR #922, opened 2017-04-29
ValueError when calling boto3 list_object_versions

georgepsarakis:
When calling S3 [list_object_versions](http://boto3.readthedocs.io/en/latest/reference/services/s3.html?highlight=s3#S3.Client.list_object_versions) the following exception is raised:

```
ValueError: invalid literal for int() with base 10: ''
```

- [x] Adding a failing test to veify bug and avoid regression.
- [x] Fix error by specifying correct type for `MaxKeys` in response template.

### DIFF
--- a/moto/s3/responses.py
+++ b/moto/s3/responses.py
@@ -224,7 +224,7 @@ class ResponseObject(_TemplateEnvironmentMixin):
                 key_list=versions,
                 bucket=bucket,
                 prefix='',
-                max_keys='',
+                max_keys=1000,
                 delimiter='',
                 is_truncated='false',
             )
--- a/tests/test_s3/test_s3.py
+++ b/tests/test_s3/test_s3.py
@@ -14,6 +14,7 @@ from boto.exception import S3CreateError, S3ResponseError
 from boto.s3.connection import S3Connection
 from boto.s3.key import Key
 from freezegun import freeze_time
+import six
 import requests
 import tests.backport_assert_raises  # noqa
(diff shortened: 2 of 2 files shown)


thehesiod:
@georgepsarakis actually there's another issue as well, it doesn't filter by the `Prefix`.  Mind adding that to your patch?

thehesiod:
another issue is that it doesn't seem to obey pagination as `max_keys` is not used in `get_bucket_versions`

thehesiod:
ack, yet another issue is that it doesn't implement DeleteMarkers

georgepsarakis:
@thehesiod can you please add a reference to the issue numbers?

thehesiod:
I haven't created issues for these, I just ran into them.  Want me to add them?

georgepsarakis:
No need probably, let me see what I can do to implement your suggestions.

thehesiod:
this: https://github.com/thehesiod/moto/commit/4da99724b9d802dc6012b610786e424549047582 is what I did until I ran into `DeleteMarker` support missing from `list-object-versions`...unfortunately unless `DeleteMarker` support is implementing I'm giving up on moto.  I have a feeling `CopySourceIfUnmodifiedSince` from `copy_object` is probably not implemented either.  I would have expected at least an error thrown if a particular feature wasn't implemented yet.

thehesiod:
given this is a critical feature, I just created https://github.com/spulec/moto/issues/936

spulec:
Thanks!

PR #924, opened 2017-05-01
Require content-length header

### LINKED ISSUE #908: S3: PUT request should fail when no 'Content-Length' header is present, but doesn't
I've discovered an inconsistency between the AWS behaviour and the moto one when trying to upload an empty file to S3 with a PUT request: as stated in http://docs.aws.amazon.com/AmazonS3/latest/API/RESTObjectPUT.html#RESTObjectPUT-requests, the Content-Length header is required in such request, and should fail with a 411 Length Required response if not provided. This is happening if I contact the "real" AWS API using the requests Python library (which, for some reason, is not setting this header when the file is empty), but the moto mocked one just returns 200 OK.

My version of moto is 0.4.27.

JackDanger:
fixes #908

### DIFF
--- a/moto/s3/responses.py
+++ b/moto/s3/responses.py
@@ -294,6 +294,8 @@ class ResponseObject(_TemplateEnvironmentMixin):
         )
 
     def _bucket_response_put(self, request, body, region_name, bucket_name, querystring, headers):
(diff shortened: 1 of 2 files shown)


spulec:
Awesome, thanks!

PR #932, opened 2017-05-07
Update list IAM AccessKeys

grahamhar:
Add the creation date, including timezone info, to the data returned
when requesting all AccessKeys for an IAM user.

This fixes #75


JackDanger:
This LGTM!

JackDanger:
Waiting on @spulec who I assume is super busy

spulec:
Thanks for your work!

PR #941, opened 2017-05-14
S3 DeleteMarker support

### LINKED ISSUE #936: S3 backend missing `DeleteMarker` support
This touches various APIs like `list-object-versions`, `delete-object`, `copy-object` and the like.  From what I see it will require a re-architecture of the S3 backend.  It's a critical feature for versioned buckets.

georgepsarakis:
- [x] Add `DeleteMarker` entries in the response for listing object versions.
- [x] Enable `IsLatest` in response, becomes `true` only for the latest version or delete marker.
- [x] Return `NoSuchKey` if `DeleteMarker` is the latest version of a versioned key.
- [x] Add test cases for versioned object deletions and key version deletion.

This closes #936.

### DIFF
--- a/moto/s3/models.py
+++ b/moto/s3/models.py
@@ -18,6 +18,17 @@ UPLOAD_ID_BYTES = 43
 UPLOAD_PART_MIN_SIZE = 5242880
 
 
+class FakeDeleteMarker(BaseModel):
+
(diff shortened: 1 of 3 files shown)


spulec:
Excellent! Thank you for your help!

PR #947, opened 2017-05-15
Shorter sleeps in SQS test

JackDanger:
One of these tests actually waited the entire 60 seconds of the
visibility timeout but that value appears to have been copied
from a previous test that didn't. Updating all tests with shorter
timeouts so folks who copy setup code in the future don't fall
into this trap


JackDanger:
I waited a week for comments but I guess this is pretty straightforward so I'll merge it solo.

Issue #949, opened 2017-05-16
Missing import for mock


acmiyaguchi:
The import for mock is missing for the repository. This can be reproduced from the head commit of the repository.

```
$ git show --oneline -s
6ff8d50 Merge pull request #946 from heddle317/add_policy_version
$ virtualenv venv
$ source venv/bin/activate
(venv) $ pip install .
(venv) $ python
Python 2.7.13 (default, Jan 19 2017, 14:48:08) 
[GCC 6.3.0 20170118] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import moto
>>> moto.__version__
u'1.0.0'
>>> with moto.mock_s3():
...     pass
... 
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "moto/core/models.py", line 44, in __enter__
    self.start()
  File "moto/core/models.py", line 55, in start
    self.enable_patching()
  File "moto/core/models.py", line 132, in enable_patching
    responses.start()
  File "moto/packages/responses/responses.py", line 297, in start
    import mock
ImportError: No module named mock

>>> 
```

PR #950, opened 2017-05-16
Fixes issue #949 - Add mock as dependency

acmiyaguchi:
This fixes issue #949. The fix can be verified by running the following:

```
import moto
with moto.mock_s3():
    pass
```


JackDanger:
Thanks for finding this! Interestingly, if I run this under `py.test` it has `mock` vendored in `pytest` but if I run it under `nosetests` it doesn't. Surprises are no fun - nice fix!

Issue #953, opened 2017-05-17
S3 Mock's error code is not 404 when a key does not exist


gordonshieh94:
I'm writing a unit test for a wrapper to S3. One of my tests will attempt to fetch a Key `Object` that does not exist on S3. I get an exception as expected, but the `ClientError.response['Error']['Code']` is not 404. But instead the string "NoSuchKey"

Our wrapper follows the example from here http://boto3.readthedocs.io/en/latest/guide/migrations3.html#accessing-a-bucket

![image](https://cloud.githubusercontent.com/assets/7716105/26172832/8e0a955a-3afe-11e7-87ba-0d013da0d5c3.png)

peterbe:
For what it's worth, the `exception.response['Error']['Code']` is `'404'` if you do a `s3client.head_object(Bucket='...', Key='doesnotexist')`
But if you instead do a `s3client.get_object(Bucket='...', Key='doesnotexist')` then `exception.response['Error']['Code']` is `'NoSuchKey'`. 

Isn't moto respecting that?

JackDanger:
@gordonshieh94 This is now fixed on Moto master. Thanks for your help!

gordonshieh94:
Thank's for fixing this so quickly! Working as expected now.

PR #954, opened 2017-05-17
NoSuchKey error in S3 is actually '404'

### LINKED ISSUE #571: S3 not found error differs from AWS response
When trying to access a key that is not present, moto returns this:

```python
{
    'ResponseMetadata': {
        'HTTPStatusCode': 404, 'HostId': '', 'RequestId': ''
    },
    'Error': {
        'Message': 'Not Found',
        'Code': '404'
    }
}
```
vs AWS S3 returning this:

```python
{
    'ResponseMetadata': {
        'HTTPStatusCode': 404, 'HostId': 'my-host-id', 'RequestId': 'some-request-id'
    },
    'Error': {
        'Message': 'The specified key does not exist.',
        'Code': 'NoSuchKey',
        'Key': 'my-s3-prefix/object-id'
    }
}
```
`Error.Code` is were the important difference lies. Handling of '404' vs 'NoSuchKey' will be very different. moto's '404'  also doesn't let you differentiate between bucket not found and key not found.

JackDanger:
Fixes #571 and #953

### DIFF
--- a/moto/s3/models.py
+++ b/moto/s3/models.py
@@ -11,7 +11,7 @@ import six
 from bisect import insort
 from moto.core import BaseBackend, BaseModel
 from moto.core.utils import iso_8601_datetime_with_milliseconds, rfc_1123_datetime
-from .exceptions import BucketAlreadyExists, MissingBucket, MissingKey, InvalidPart, EntityTooSmall
+from .exceptions import BucketAlreadyExists, MissingBucket, InvalidPart, EntityTooSmall
 from .utils import clean_key_name, _VersionedKeyStore
 
 UPLOAD_ID_BYTES = 43
@@ -473,7 +473,7 @@ class S3Backend(BaseBackend):
         if isinstance(key, FakeKey):
             return key
         else:
-            raise MissingKey(key_name=key_name)
+            return None
 
     def initiate_multipart(self, bucket_name, key_name, metadata):
         bucket = self.get_bucket(bucket_name)
--- a/moto/s3/responses.py
+++ b/moto/s3/responses.py
@@ -12,7 +12,7 @@ from moto.core.responses import _TemplateEnvironmentMixin
 from moto.s3bucket_path.utils import bucket_name_from_url as bucketpath_bucket_name_from_url, parse_key_name as bucketpath_parse_key_name, is_delete_keys as bucketpath_is_delete_keys
 
 
-from .exceptions import BucketAlreadyExists, S3ClientError, InvalidPartOrder
+from .exceptions import BucketAlreadyExists, S3ClientError, MissingKey, InvalidPartOrder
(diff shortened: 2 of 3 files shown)


JackDanger:
@georgepsarakis Would you mind taking a look at this? It appears AWS does not, despite the docs, return a 'NoSuchKey' error code, it returns '404'. I reproduced this but would love if you could try it as well in case you see something I missed.

georgepsarakis:
As you mentioned above the documentation says that the HTTP status code is 404, but the 
`<Code>` field contains the value `NoSuchKey`. Have you tried it with AWS directly and it does not return this response?

JackDanger:
> Have you tried it with AWS directly and it does not return this response?

Yes, and that's the most confusing part about this. Here's a recreation:

```
try:
  boto3.session.Session(region_name='us-west-1').resource('s3').Object('real-bucket', key='does-not-exist').load()
except Exception as e:
  pass
print(e.response)
{'Error': {'Code': '404', 'Message': 'Not Found'},
 'ResponseMetadata': {'HTTPHeaders': {'content-type': 'application/xml',
   'date': 'Wed, 17 May 2017 19:55:45 GMT',
   'server': 'AmazonS3',
   'transfer-encoding': 'chunked',
   'x-amz-id-2': 'CqX9MueZUQpD1oQGSxUrRRXpleR4o0oa3SKE6WPDMry/IKecB5m2FOAcNMbUPxT8j1DpVwbY9tU=',
   'x-amz-request-id': 'AC981B0B8540A48E'},
  'HTTPStatusCode': 404,
  'HostId': 'CqX9MueZUQpD1oQGSxUrRRXpleR4o0oa3SKE6WPDMry/IKecB5m2FOAcNMbUPxT8j1DpVwbY9tU=',
  'RequestId': 'AC981B0B8540A48E',
  'RetryAttempts': 1}}
```

peterbe:
I'm running `boto3==1.4.4` and `botocore==1.5.48`. 

The tricky thing is that the Code is different depending on if you do a `get_object` or `head_object`.

Consider the following...
```python
import boto3
s3_client = boto3.client('s3')
try:
    s3_client.get_object(Bucket='real-bucket', Key='doesnotexist')
except ClientError as exc:
    print(exc.response['Error']['Code'])

try:
    s3_client.head_object(Bucket='real-bucket', Key='doesnotexist')
except ClientError as exc:
    print(exc.response['Error']['Code'])
```

The output is...:
```
NoSuchKey
404
```
`moto` just needs to do the same. I think it did prior to 1.0.0.

JackDanger:
@peterbe I've changed it such that HEAD requests and GET requests return different error codes. This appears to be how it worked before May 1st when the exceptions were refactored. The tests have been updated to prevent a regression.

PR #960, opened 2017-05-19
feat(s3) HeadObject: honor If-Modified-Since header

spg:
See http://docs.aws.amazon.com/AmazonS3/latest/API/RESTObjectHEAD.html


JackDanger:
Thanks for adding this feature, @spg!

PR #964, opened 2017-05-24
Add extended CloudFormation models for Lambda and DynamoDB

whummer:
This PR adds extended CloudFormation models for Lambda and DynamoDB. We encountered that these classes are missing when deploying a simple sample project via *serverless* (https://github.com/serverless/serverless)

* new class `EventSourceMapping`
* new class `LambdaVersion`
* new class `LogGroup`
* new mappings in `parsing.py`

Additionally, this PR introduces an environment variable `VALIDATE_LAMBDA_S3` to allow us to configure whether creating an instance of `LambdaFunction` should necessarily raise an exception if the code does not exist in S3. This is also required to get serverless working locally. Note that the default behavior has been left unchanged (i.e., do validate if `VALIDATE_LAMBDA_S3` is not set), so there should be no issues with backwards compatibility whatsoever.

### DIFF
--- a/moto/cloudformation/parsing.py
+++ b/moto/cloudformation/parsing.py
@@ -7,7 +7,9 @@ import warnings
 
 from moto.autoscaling import models as autoscaling_models
 from moto.awslambda import models as lambda_models
+from moto.cloudwatch import models as cloudwatch_models
 from moto.datapipeline import models as datapipeline_models
+from moto.dynamodb import models as dynamodb_models
 from moto.ec2 import models as ec2_models
 from moto.ecs import models as ecs_models
 from moto.elb import models as elb_models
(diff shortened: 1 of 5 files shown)


moto/cloudformation/parsing.py
```diff
@@ -133,7 +139,7 @@ def clean_json(resource_json, resources_map):
             try:
                 return resource.get_cfn_attribute(resource_json['Fn::GetAtt'][1])
             except NotImplementedError as n:
-                logger.warning(n.message.format(
+                logger.warning(str(n).format(
```
JackDanger: Thanks for this

JackDanger:
@whummer Your changes look great and all that's missing is tests for the new code. If you get stuck implementing them please say so and I'll help out.

whummer:
Thanks for the reminder @JackDanger . I've added a set of tests to cover the new functionality, please take a look. Thanks

JackDanger:
@whummer Nice! Thanks so much. Merging

PR #965, opened 2017-05-26
Use region list from Boto.

hardyj:
Boto can be configured with extra regions, but moto will fail to import
if they are not in the hardcoded list in ec2/models.py. Instead, use the
region list from boto to build the ec2_backends dict to ensure all
regions are available.


JackDanger:
Great work, @hardyj ! This is much appreciated.

PR #969, opened 2017-06-01
Bug fix for infinite loop when WaitTimeSeconds was set to 0

Brett55:
There was a bug when WaitTimeSeconds was set to zero causing an infinite loop if no messages came in. Added two unit tests

### DIFF
--- a/moto/sqs/models.py
+++ b/moto/sqs/models.py
@@ -334,6 +334,8 @@ class SQSBackend(BaseBackend):
         :param string queue_name: The name of the queue to read from.
         :param int count: The maximum amount of messages to retrieve.
         :param int visibility_timeout: The number of seconds the message should remain invisible to other queue readers.
+        :param int wait_seconds_timeout:  The duration (in seconds) for which the call waits for a message to arrive in
+         the queue before returning. If a message is available, the call returns sooner than WaitTimeSeconds
         """
         queue = self.get_queue(queue_name)
         result = []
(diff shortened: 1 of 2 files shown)


JackDanger:
Fantastic work, @Brett55!

JackDanger:
Merged, thank you for your contribution!

Brett55:
@JackDanger How do I bump the version so I can install it locally with new changes?

JackDanger:
> @JackDanger How do I bump the version so I can install it locally with new changes?

If you just need it installed locally this'll work: `pip install -U git+https://github.com/spulec/moto.git@master`

If you need it to work with setup.py you'll have to google for 'process dependency links'. I'm unsure how to read the latest code with requirements.txt but I'm sure there's a way.

Brett55:
When does pypi typically get updated?

JackDanger:
@Brett55 Historically, not often. The 1.0.0 release just happened and I bet @spulec will release another one in the next month or two but it's not something we do after each bugfix. Personally, I use `setup.py` and `--process-dependency-links` and have my primary app use the latest master commit of Moto at all times. I recommend that approach until the next stable release.

Brett55:
Thanks, I'm familiar with --process-dependency-links, I've noticed its
deprecated, have you found any other alternatives?

JackDanger:
I really haven't. There's an issue on Pip about un-deprecating it: https://github.com/pypa/pip/issues/4187

spulec:
@Brett55 I've released `1.0.1` (I'm trying to get better about releasing versions more often)

Brett55:
Thanks!

PR #974, opened 2017-06-07
add pass_through option to responses

### LINKED ISSUE #948: can't use responses
with v1, since moto vendors in `responses`, it's impossible to use `responses` for mocking non-aws calls. example:

```py
with _responses.RequestsMock() as rsps:
    rsps.add(rsps.GET, 'http://foo')
    with moto.mock_sns():
       requests.get('http://foo')
```

this used to work on < v1, on v1 it raises a connection refused error because basically requests is patched twice by two different modules

itajaja:
fixes #948

### DIFF
--- a/moto/packages/responses/responses.py
+++ b/moto/packages/responses/responses.py
@@ -10,6 +10,7 @@ import six
 from collections import namedtuple, Sequence, Sized
 from functools import update_wrapper
 from cookies import Cookies
+from requests.adapters import HTTPAdapter
 from requests.utils import cookiejar_from_dict
 from requests.exceptions import ConnectionError
 from requests.sessions import REDIRECT_STATI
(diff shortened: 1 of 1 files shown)


itajaja:
ping :)

JackDanger:
@itajaja I *think* I understand the reason for this change but could you explain it to us just the same? I'm not sure quite when I'd use this but I suspect it's really useful.

itajaja:
look at the first message on #948, this will enable using responses for additional mocking outside moto

JackDanger:
Ah, this looks pretty good then. Passing to @spulec for approval but this looks good to me because it allows developers to work around the hard-wired dependency.

itajaja:
I actually opened a PR upstram to allow that, but didn't receive a response yet: getsentry/responses#150

spulec:
Looks great! Thanks everyone.

itajaja:
Thanks! After a new release is cut, we can finally update to Moto 1 :)

PR #975, opened 2017-06-08
Add cf parsing

nadlerjessie:
@spulec This PR adds a few parsing options, specifically: `Fn::Split`, `Fn::Select`, `Fn::Sub`, and `Fn::ImportValue`. 
To implement `Fn::ImportValue`, I added a field, `cross_stack_resources` to `FakeStack`.

Additionally, I refactored two things. 
(1) `Fn::Join`:
It wasn't working correctly when joining on another intrinsic functions. For example, previously 
`{"Fn::Join": ['"-", {"Fn::Split": ["_", "foo_bar"]}]}` evaluated to `"Fn::Split"`, rather than `foo-bar` due to the iteration over ` {"Fn::Split": ["_", "foo_bar"]}`.
(2) `Export`:
I parsed the name and values before creating the `Export` object. It is likely one or both of those values will use some intrinsic functions.

Please let me know if you have any feedback. Thanks!
cc @hltbra

### DIFF
--- a/moto/cloudformation/parsing.py
+++ b/moto/cloudformation/parsing.py
@@ -162,6 +162,15 @@ def clean_json(resource_json, resources_map):
                                  if cleaned_val else '{0}'.format(val))
             return resource_json['Fn::Join'][0].join(join_list)
 
+        if 'Fn::Split' in resource_json:
+            to_split = clean_json(resource_json['Fn::Split'][1], resources_map)
+            return to_split.split(resource_json['Fn::Split'][0])
+
+        if 'Fn::Select' in resource_json:
+            select_index = int(resource_json['Fn::Select'][0])
+            select_list = clean_json(resource_json['Fn::Select'][1], resources_map)
+            return select_list[select_index]
+
         cleaned_json = {}
         for key, value in resource_json.items():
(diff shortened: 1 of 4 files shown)


spulec:
This is great! Thanks!

moto/cloudformation/parsing.py
```diff
@@ -155,12 +156,42 @@ def clean_json(resource_json, resources_map):
...
+                for sub in to_sub:
+                    if '.' in sub:
+                        cleaned_ref = clean_json({'Fn::GetAtt': re.findall('(?<=\${)[^"]*?(?=})', sub)[0].split('.')}, resources_map)
+                    else:
+                        cleaned_ref = clean_json({'Ref': re.findall('(?<=\${)[^"]*?(?=})', sub)[0]}, resources_map)
+                    fn_sub_value = fn_sub_value.replace(sub, cleaned_ref)
+                for literal in literals:
+                    fn_sub_value = fn_sub_value.replace(literal, literal.replace('!', ''))
+                return fn_sub_value
+            pass
```
hltbra: This `pass` is dead code. (I know this was already merged, but it there's any subsequent PRs, this is an opportunity to remove this line)

PR #977, opened 2017-06-10
Support Expected in dynamoDB updateItem

gregsterin:
DynamoDB supports Expected on UpdateItem http://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_UpdateItem.html.

moto does not have that implemented, so this PR adds support for providing Expected on update_item requests.


gregsterin:
@spulec

spulec:
Thanks!

PR #978, opened 2017-06-12
create_load_balancer requires port definitions

JackDanger:
Throw the appropriate error when defining a loadbalancer with no ports


spulec:
👍

PR #984, opened 2017-06-20
Re-enabling tests on Python3

JackDanger:


spulec:
this is awesome, thanks for cleaning this up!

PR #987, opened 2017-06-20
Ecr describe images filter with image

william-richard:
Added the ability to mock describe_images and pass an image tag or image digest to filter the results.

This is branched off of https://github.com/spulec/moto/pull/985, so it would probably be easier to merge that first before looking at this.


JackDanger:
This is great! Nice work @william-richard!

PR #989, opened 2017-06-21
Boto3 has different keys for auto scaling tags - handle that gracefully

william-richard:


tests/test_autoscaling/test_autoscaling.py
```diff
@@ -8,7 +8,8 @@
 import boto.ec2.elb
 import sure  # noqa
 
-from moto import mock_autoscaling, mock_ec2_deprecated, mock_elb_deprecated, mock_autoscaling_deprecated, mock_ec2
+from moto import mock_autoscaling, mock_ec2_deprecated, mock_elb_deprecated, mock_autoscaling_deprecated, \
```
JackDanger: If you're cleaning this line up there's probably a clearer way these imports could be expressed. IMO they should be all on their own lines for readability

william-richard: ugh yeah, my editor did that - I can split them up to your liking, or revert the change.  Whichever you prefer.

moto/autoscaling/models.py
```diff
@@ -262,10 +258,17 @@ def set_desired_capacity(self, new_capacity):
         if self.desired_capacity > curr_instance_count:
             # Need more instances
             count_needed = int(self.desired_capacity) - \
-                int(curr_instance_count)
+                           int(curr_instance_count)
```
JackDanger: The tests are failing because of the indentation here

JackDanger:
LGTM once the tests are passing. Thanks for doing this @william-richard!

moto/autoscaling/models.py
william-richard: Yeah I have a fix queued up, as well as a fix for a test that I didn't realize was failing - should be pushing soon.

tests/test_autoscaling/test_autoscaling.py
JackDanger: Let's revert the change for now purely because I don't think I have a sense for what the Moto style guidelines should be here. If you feel strongly one way or another please adjust the code accordingly.

PR #990, opened 2017-06-27
Add support for recursive emr settings

bobo333:
- Updates _RecursiveDictRef to implement `__getitem__`, avoiding errors when using recursive settings for an emr job flow.

Previously, when using settings like the following, `moto` would throw a `TypeError: '_RecursiveDictRef' object has no attribute '__getitem__'` exception.

```
{
            "Classification": "spark-env",
            "Configurations": [
                {
                    "Classification": "export",
                    "Configurations": [],
                    "Properties": {
                        "JAVA_HOME": "/usr/lib/jvm/java-1.8.0"
                    }
                }
            ],
            "Properties": {}
        }
```

From the `boto3` docs, these recursive settings should be supported: http://boto3.readthedocs.io/en/latest/reference/services/emr.html#EMR.Client.run_job_flow

I also tested this and `boto3` correctly launched an emr cluster with these settings, so `moto` should support mocking it.

### DIFF
--- a/moto/core/responses.py
+++ b/moto/core/responses.py
@@ -414,6 +414,9 @@ class _RecursiveDictRef(object):
     def __getattr__(self, key):
         return self.dic.__getattr__(key)
 
+    def __getitem__(self, key):
+        return self.dic.__getitem__(key)
+
     def set_reference(self, key, dic):
         """Set the RecursiveDictRef object to keep reference to dict object
         (dic) at the key.
--- a/tests/test_emr/test_emr_boto3.py
+++ b/tests/test_emr/test_emr_boto3.py
@@ -64,7 +64,18 @@ def test_describe_cluster():
     args['Configurations'] = [
         {'Classification': 'yarn-site',
          'Properties': {'someproperty': 'somevalue',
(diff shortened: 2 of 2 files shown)


JackDanger:
Great work, @bobo333! Thanks for your contribution and welcome to the Moto project

bobo333:
Thanks very much @JackDanger! I'm not familiar with the release cadence, any idea when this will make it into an official release? I'd like to use the functionality 😄

JackDanger:
We're releasing more often but I expect you need it right now. I recommend doing what I do in between releases and adding this to setup.py:

```
setup(
    name='bobo333_app',
    include_package_data=True,
    dependency_links=[
'git+https://github.com/spulec/moto.git@master#egg=moto-0.0.0.0.1'
    ],
    install_requires=[
        "moto==0.0.0.0.1",
    ]
)
```
Then `python setup.py --process-dependency-links` will get you going.

Eventually everybody needs to switch to something like [Pipfile](https://github.com/pypa/pipfile) but we're not there as a Python community yet.

bobo333:
Sounds good, I appreciate it!

PR #992, opened 2017-06-29
SQSResponse: include MD5OfMessageAttributes parameter only when there are message attributes

ingluisjimenez:
bug fix for https://github.com/spulec/moto/issues/993

The `MD5OfMessageAttributes` parameter should only be included in the response when there is at least 1 message attribute.

### DIFF
--- a/moto/sqs/responses.py
+++ b/moto/sqs/responses.py
@@ -339,7 +339,9 @@ SEND_MESSAGE_RESPONSE = """<SendMessageResponse>
         <MD5OfMessageBody>
             {{- message.body_md5 -}}
         </MD5OfMessageBody>
+        {% if message.message_attributes.items()|count > 0 %}
         <MD5OfMessageAttributes>{{- message.attribute_md5 -}}</MD5OfMessageAttributes>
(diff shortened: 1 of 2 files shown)


ingluisjimenez:
cc: @JackDanger who pushed https://github.com/spulec/moto/pull/919

JackDanger:
Great patch, @ingluisjimenez! This is much appreciated, welcome to Moto!

ingluisjimenez:
Thanks @JackDanger, much appreciated. Is there a process to make a release for this patch?

JackDanger:
@ingluisjimenez There's so much activity on Moto that I've been recommending people just bundle the `master` branch (always stable). For most projects that means using `--process-dependency-links` in setup.py but for localstack is there a way to bundle using the head revisions of the Moto dependency?

If not, the release schedule is up to @spulec

ingluisjimenez:
I'll ping the folks at localstack and see if they can switch to use master. That makes sense, I can see there is a lot of activity here, which is great. 

Keep the good work coming!

PR #994, opened 2017-06-30
Add CLI flag to enable SSL for moto_server

whummer:
This PR adds an `--ssl` flag for the `moto_server` CLI to enable encrypted connections. This is an important requirement for users that want to test their service calls via HTTPS.

The SSL certificate is generated on the fly and does not require any additional configuration. (However, as this approach uses a self-signed certificate, most clients using the server will have to be configured with a flag to skip the certificate validation).


JackDanger:
Great idea, @whummer! Merging.

PR #1001, opened 2017-07-07
Test boto3 elb listener deletion

JackDanger:


JackDanger:
The `boto` path was exercised and this ensures we're testing the `boto3` path as well

Issue #818, opened 2017-01-25
Implement verify_email_address method in ses


ncknuna:
While the `send_email` method in ses is mocked by moto, responses come back with:

botocore.errorfactory.MessageRejected: An error occurred (MessageRejected) when calling the SendEmail operation: Email address not verified

When attempting to use the `verify_email_address` method to fix this issue, a NotImplementedError gets raised:

moto/core/responses.py:193: NotImplementedError

This makes it very difficult to test the core functionality of ses using moto.

spulec:
Noted. Pull requests are very much welcome if anyone has time.

terrycain:
Pretty sure this works now, its in the codebase. @JackDanger close please

PR #1003, opened 2017-07-10
verify_email_address

fdfk:
#818


JackDanger:
@fdfk Thank you for this! This code needs a corresponding test to cover it as well before it can be merged. Please try to add a test and if you get stuck just mention it here and another Moto contributor will step in and help out.

PR #1011, opened 2017-07-14
Avoid to override SocketType when disabling the mock and bad_socket_s…

gilgamezh:
…hadow is True

In older versions of python3.4 (debian jessie) when enabling the mock SocketType is skipped. But in the disable() isn't it. So the result is SocketType overrided. 

Small fix to avoid that.

### DIFF
--- a/moto/packages/httpretty/core.py
+++ b/moto/packages/httpretty/core.py
@@ -72,6 +72,10 @@ from datetime import datetime
 from datetime import timedelta
 from errno import EAGAIN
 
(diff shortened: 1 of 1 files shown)


JackDanger:
Nice fix, @gilgamezh! I appreciate the comment explaining the use. Thanks for jumping in to help with this communal tool. Welcome to Moto! :)

gilgamezh:
@JackDanger thanks you!

PR #1018, opened 2017-07-20
Make HEAD bucket throw ClientError instead of NoSuchBucket on boto3

### LINKED ISSUE #1004: Error code confusion in mocked responses?
It appears an error response including `Code` and `Message` is rendered using `moto.core.exceptions.ERROR_RESPONSE`, which is passed to the template via `RESTError(error_type, message)` where `error_type` gets rendered into `Code` of the template.

In `exceptions.py` of service submodules, `error_type` often is a short descriptive text, instead of HTTP status code which appears to be the case in many non-mocked responses. (In fact, real responses are indeed inconsistent in `Code` being a numeric status code vs. short text...)

Unfortunate side effect of the difference is that an idiomatic code like the following does not work properly (this is from a `boto3` example page):

```
import botocore
bucket = s3.Bucket('mybucket')
exists = True
try:
    s3.meta.client.head_bucket(Bucket='mybucket')
except botocore.exceptions.ClientError as e:
    # If a client error is thrown, then check that it was a 404 error.
    # If it was a 404 error, then the bucket does not exist.
    error_code = int(e.response['Error']['Code'])
    if error_code == 404:
        exists = False
```

since `e.response['Error']['Code']` of mocked response, being a text, cannot be cast properly to `int`.

I'm wondering if this is a design decision of some sort of oversight? Is there a reason why `Code` and `Message` are not mocked closely to the actual responses?

okomestudio:
With `boto` and `boto3`, the responses are different when HEAD bucket operation returns 404, the former returning `S3ResponseError` but the latter returning `ClientError`. `moto` returns `S3ResponseError` for both `boto` and `boto3`, which makes exception handling for the specific case to be incompatible assuming the real API behavior. This PR should fix the issue and resolves #1004.

### DIFF
--- a/moto/s3/responses.py
+++ b/moto/s3/responses.py
@@ -13,7 +13,7 @@ from moto.core.responses import _TemplateEnvironmentMixin
 from moto.s3bucket_path.utils import bucket_name_from_url as bucketpath_bucket_name_from_url, parse_key_name as bucketpath_parse_key_name, is_delete_keys as bucketpath_is_delete_keys
 
 
-from .exceptions import BucketAlreadyExists, S3ClientError, MissingKey, InvalidPartOrder
+from .exceptions import BucketAlreadyExists, S3ClientError, MissingBucket, MissingKey, InvalidPartOrder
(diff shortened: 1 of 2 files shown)


JackDanger:
@okomestudio This is a great fix! Thank you so much

PR #1023, opened 2017-07-25
Add ecr get_authorization_token response and tests

william-richard:
Pretty straightforward - not sure what else to add.


william-richard:
This isn't correct - it should be returning a b64 encoded string of the form `AWS:<password`.  I'm going to reopen once I have that.

JackDanger:
This looks great, @william-richard - thanks!

PR #1027, opened 2017-07-30
Send JSON message to HTTP endpoint of SNS

sodastsai:
By the documentation from AWS - http://docs.aws.amazon.com/sns/latest/dg/SendMessageToHttp.html , SNS would send messages to HTTP/HTTPS endpoint in JSON format. But current implementation of `moto` sends messages in form-data format.


JackDanger:
Thanks for this great work, @sodastsai ! Welcome to Moto

PR #1028, opened 2017-07-31
SNS delete_topic should also delete subscriptions.

ptrus:
By the documentation from AWS (http://docs.aws.amazon.com/cli/latest/reference/sns/delete-topic.html), deleting SNS should also delete all of the subscriptions. Current implementation of moto only deletes the topic.

Although in https://forums.aws.amazon.com/thread.jspa?threadID=219168 they state:

> 
> When you delete a topic, subscriptions to the topic will not be "deleted" immediately, 
> but become orphans.  SNS will periodically clean these orphans, usually every 10 hours, but not 
> guaranteed. 
> If you create a new topic with the same topic name before these orphans are cleared up, 
> the new topic will not inherit these orphans. So, no worry about them.
> 

Currently in moto, recreating the topic with same topic-name actually inherits the orphaned subscriptions.  Therefore i purpose to just delete subscriptions as topic gets deleted.

### DIFF
--- a/moto/sns/models.py
+++ b/moto/sns/models.py
@@ -193,10 +193,17 @@ class SNSBackend(BaseBackend):
             next_token = None
         return values, next_token
 
+    def _get_topic_subscriptions(self, topic):
+        return [sub for sub in self.subscriptions.values() if sub.topic == topic]
+
     def list_topics(self, next_token=None):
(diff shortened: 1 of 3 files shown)


JackDanger:
This is fantastic, @ptrus! Your work is much appreciated. Welcome to Moto!

PR #1032, opened 2017-08-04
Fix SWF name in docs

andrewhillcode:
Docs had `@mock_sfw` written when it should be `@mock_swf`.

Let me know if those .rst files didn't need touching.


JackDanger:
Nice catch @andrewhillcode!

Issue #866, opened 2017-03-13
Queue names ending with '.fifo' yield 404


tilthouse:
We are going to use FIFO queues, which just have a queue name ending in `.fifo`. When I change the names of my queues accordingly, moto only will give 404 errors. 

The create queue calls appear to work, but then the queues are not found.

spulec:
Can you give some more details? The following code works for me

```python
@mock_sqs
def test_create_queue_with_dot_in_name():
    sqs = boto3.resource('sqs', region_name='us-east-1')
    new_queue = sqs.create_queue(QueueName='test.queue')
    new_queue.should_not.be.none
    new_queue.should.have.property('url').should.contain('test.queue')

    queue = sqs.get_queue_by_name(QueueName='test.queue')
    queue.attributes.get('QueueArn').should_not.be.none
    queue.attributes.get('QueueArn').split(':')[-1].should.equal('test.queue')
```

PR #1034, opened 2017-08-05
Fix receiving of messages from queues with a dot character in their name

whummer:
This PR changes the `url_path` regex for SQS to allow the dot character (`.`) in queue names. This change fixes https://github.com/spulec/moto/issues/866

Tests have been extended to cover the change and avoid possible regressions in the future.


JackDanger:
This looks great, @whummer

PR #1035, opened 2017-08-05
Allow non-ascii characters in request URLs

whummer:
This PR adds support for non-ASCII characters in request URLs. This is required to support, for instance, unicode bucket or object names in S3 (when using S3 path encoding).

The change has been manually tested in Python 2.x and 3.x.

A simple test has been added to test the new functionality. Note that the `PATH_INFO` contains Japanese unicode characters.

### DIFF
--- a/moto/server.py
+++ b/moto/server.py
@@ -3,6 +3,7 @@ import json
 import re
 import sys
 import argparse
+import six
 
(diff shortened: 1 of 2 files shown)


JackDanger:
Great idea, @whummer!

PR #1037, opened 2017-08-09
updating documentation to describe support for boto2

ygrosu:


JackDanger:
@ygrosu this is super helpful! Thank you very much. Welcome to Moto!

PR #1039, opened 2017-08-10
Redshift Updates

bpandola:
I'm working on a project involving Redshift and implemented a few more endpoints as well as a bit of miscellaneous clean up and bug fixes.  Let me know what you think. I could probably commit to more work on the Redshift endpoints, if you're happy with what you see.

Here's a summary of my changes:
- Implement create_cluster_snapshot endpoint
- Implement describe_cluster_snapshots endpoint
- Implement delete_cluster_snapshot endpoint
- Implement restore_from_cluster_snapshot endpoint
- Implement limited support for describe_tags endpoint
- Correctly serialize errors to json (for boto) or xml (for boto3)
- Simulate cluster spin up by returning initial status as 'creating' and subsequent statuses as 'available'
- Fix issue with modify_cluster endpoint where cluster values get set to None when omitted from request
- Add 'Endpoint' key to describe_clusters response syntax

### DIFF
--- a/moto/redshift/exceptions.py
+++ b/moto/redshift/exceptions.py
@@ -56,3 +56,18 @@ class InvalidSubnetError(RedshiftClientError):
         super(InvalidSubnetError, self).__init__(
             'InvalidSubnet',
             "Subnet {0} not found.".format(subnet_identifier))
+
+
+class ClusterSnapshotNotFoundError(RedshiftClientError):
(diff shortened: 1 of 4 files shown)


JackDanger:
@bpandola This is extremely good! Thank you so much. This is going to be a big help to so many thousands of people. I'll merge this right away and, yes, we would love if you could help implement any of the other endpoints. Welcome to Moto! :)

PR #1045, opened 2017-08-10
fix list-object-versions with prefix

thehesiod:


JackDanger:
@thehesiod This is great! It needs a test on this behavior before we can merge it. Mind making an attempt at adding a test for this and if you get stuck just push what you have and ask for help here?
I recommend starting around here: https://github.com/thehesiod/moto/blob/e1aa42cf7df129a1bf618e5e599d3e6d30b674a8/tests/test_s3/test_s3.py#L749

thehesiod:
done

JackDanger:
Nice! Great work. Merging.

JackDanger:
It's merged. Thanks for the contribution @thehesiod. Welcome to Moto! :)

thehesiod:
I'm working on a scenario where I have a SNS topic that's supposed to trigger Lambdas, but I'm guessing lambda execution is not supported as that would require moto using docker or something under the covers.  It's unfortunate there doesn't seem to be a botocore API to be able to get SNS notifications, but I saw the trick of hooking a SQS queue to get them that way.  thanks for this module!

JackDanger:
@thehesiod Please share with the Moto community (in an issue or wherever is convenient) whatever setup you come up with. Everybody is trying to solve very similar problems!

thehesiod:
logged https://github.com/spulec/moto/issues/1047

PR #1046, opened 2017-08-11
Add proper errors to ECR calls

hltbra:
This pull requests adds proper error handling to missing ECR images and ECR repositories. I didn't add error handling to validate the parameters. The error messages were copied from real boto3 calls.

I updated botocore to the current latest version to get the latest features from the ECR client.

P.S.: I had to change the `struct` calls in the SQS models file because of older versions of Python (I am running 2.7.6): https://bugs.python.org/issue19099


spulec:
Looks great, thanks!

Issue #1047, opened 2017-08-11
[enhancement] Support Lambda Execution via Docker containers


thehesiod:
I think something like this can be used: https://github.com/lambci/docker-lambda

to provide true support for lambda functions...I think right now it expects everything to be in a exec'able string, which means it doesn't support multiple files as noted in https://github.com/spulec/moto/issues/981

terrycain:
This is done right?

thehesiod:
yep!

Issue #981, opened 2017-06-15
Feature Request: Lambda Function with multiple files in a Zip File


RAbraham:
Hi,
I have a zip file which contains two files. I can successfully unzip the zip file using `unzip` in the terminal.

```python
code_paths = ['lambda.py', 'handler.dill']
zip_path = '/tmp/' + 'moto-rajiv' + '.zip'
make_local_zipfile(code_paths)
with zipfile.ZipFile(zip_path, "r", zipfile.ZIP_DEFLATED) as f:
        pprint(f.namelist()) #This prints ['lambda.py', 'handler.dill']

with open(zip_path, 'rb') as ziper:
       conn.create_function(
            FunctionName=...,
            ...
            Code={
                'ZipFile': ziper.read()
            },
            ...
        )
```

I get the following error. Note that the files get conjugated into one single line:
```
Traceback (most recent call last):
  File "/Users/rabraham/Documents/dev/thescore/analytics/jaya/jaya/mock/in_memory_harness.py", line 121, in <module>
    deploy_lambda.deploy_lambda_package_local(print_lambda)
  File "/Users/rabraham/Documents/dev/thescore/analytics/jaya/jaya/deployment/deploy_lambda.py", line 329, in deploy_lambda_package_local
    Publish=True,
  File "/Users/rabraham/Documents/dev/thescore/analytics/jaya/venv/lib/python3.6/site-packages/botocore/client.py", line 253, in _api_call
    return self._make_api_call(operation_name, kwargs)
  File "/Users/rabraham/Documents/dev/thescore/analytics/jaya/venv/lib/python3.6/site-packages/botocore/client.py", line 544, in _make_api_call
    operation_model, request_dict)
  File "/Users/rabraham/Documents/dev/thescore/analytics/jaya/venv/lib/python3.6/site-packages/botocore/endpoint.py", line 141, in make_request
    return self._send_request(request_dict, operation_model)
  File "/Users/rabraham/Documents/dev/thescore/analytics/jaya/venv/lib/python3.6/site-packages/botocore/endpoint.py", line 170, in _send_request
    success_response, exception):
  File "/Users/rabraham/Documents/dev/thescore/analytics/jaya/venv/lib/python3.6/site-packages/botocore/endpoint.py", line 249, in _needs_retry
    caught_exception=caught_exception, request_dict=request_dict)
[...]

PR #1048, opened 2017-08-11
lambda + SNS enhancements

thehesiod:
## Features
- support lambda messages from SNS
- run lambda in docker container
- support all lambda environments via docker
- fixes lambda region + arn
- adds partial cloudwatchlogs endpoint
- fixes environment variable propagation to lambda

## Resolves
- https://github.com/spulec/moto/issues/1047
- https://github.com/spulec/moto/issues/981
- escaping issue with urls that contain backslashes


thehesiod:
btw, you can see what other variables can be populated by looking at:
/var/runtime/awslambda/bootstrap.py + /var/runtime/awslambda/runtime.py in the docker container

thehesiod:
ok, new lambda w/ new cloudwatchlogs is working.  One side-effect of this is that if you're using patching the lambda exec will no longer obey it since it's being run in a container.  For my use-case I did something like this:

```python
def _wrapt_boto_create_client(wrapped, instance, args, kwargs):
    def unwrap_args(service_name, region_name=None, api_version=None,
                      use_ssl=True, verify=None, endpoint_url=None,
                      aws_access_key_id=None, aws_secret_access_key=None,
                      aws_session_token=None, config=None):

        if endpoint_url is None:
            endpoint_url = os.environ.get('{}_endpoint_url'.format(service_name))

        return wrapped(service_name, region_name, api_version, use_ssl, verify,
                       endpoint_url, aws_access_key_id, aws_secret_access_key,
                       aws_session_token, config)

    return unwrap_args(*args, **kwargs)

    async with MotoService('sns') as sns_svc, MotoService('lambda') as lambda_svc, \
        MotoService('s3') as s3_svc, MotoService('kms') as kms_svc, MotoService('logs') as logs_svc:

        os.environ['lambda_endpoint_url'] = lambda_svc.endpoint_url
        os.environ['kms_endpoint_url'] = kms_svc.endpoint_url
        os.environ['sns_endpoint_url'] = sns_svc.endpoint_url
        os.environ['s3_endpoint_url'] = s3_svc.endpoint_url
        os.environ['logs_endpoint_url'] = logs_svc.endpoint_url
```
where MotoService is: https://gist.github.com/thehesiod/2e4094a1db1190f7e122e7043f1973a0  I'm thinking of something similar in my lambda function and hooking it up via the SNS message

thehesiod:
@JackDanger any ideas about this CI issue?

toshitanian:
you should fix syntax error.
```
9.63s$ make test
flake8 moto
moto/awslambda/models.py:180:21: E123 closing bracket does not match indentation of opening bracket's line
```

thehesiod:
actually it was a different error, let me get it back to that point

thehesiod:
ok, let me know what you think of the error that's now happening

toshitanian:
hmm, `(module 'boto' has no attribute 'plugin')` looks the main error.
I will take a look at it .

toshitanian:
I tried to find out the reason for error. but couldn't make it  😢

thehesiod:
Weird eh? From what I saw online it could be a problem with Travis and boto

JackDanger:
I tried to `git bisect` but there were PEP8 errors on the first commit and it was tough to isolate this one failure. I gave up before I found the root cause

thehesiod:
I have a feeling it's related to adding the docker service

thehesiod:
ran across this: https://github.com/travis-ci/travis-ci/issues/7940

thehesiod:
yay, I fixed it :)

thehesiod:
ok, what's left are unittests, if anyone wants to help pitch in let me know!  I'll take care of the lambda broken  lambda tests.

thehesiod:
ok, fixed the base issue of lambda invoke, next thing is tests like `test_invoke_function_get_ec2_volume` where it's trying to use mocked boto calls in the lambda function.  To do this it would need to use the endpoint_url redirect and run in server mode, what do you guys think?

thehesiod:
not sure why `test_invoke_requestresponse_function` isn't passing in travis, it works locally

thehesiod:
ok, have a way to do the invoke from the lambda, but need to think of how to securely expose the host's IP

thehesiod:
`test_lambda.test_invoke_function_get_ec2_volume` also passes locally <sigh>

thehesiod:
hmm, I think docker volume mounting may be broken?  https://github.com/travis-ci/worker/issues/171

thehesiod:
so the last issue is getting the external IP from the container w/o exposing the service to the world.

oharaandrew314:
Will this fix `get_function` when the function was created with a `ZipFile`?  If so, then I'll hold off on fixing it myself.

thehesiod:
@oharaandrew314 can you provide a unittest to add to test_lambda.py? I can add it to my PR and make sure it works

oharaandrew314:
It's really simple.  At the end of `test_create_function_from_zipfile`, just add the line `conn.get_function(FunctionName='testFunction')` and you'll get an error.

thehesiod:
so I checked, if you upload a zip, it ends up making Code = something like: https://awslambda-us-west-2-tasks.s3-[REGION].amazonaws.com/snapshots/[ACCOUNT]/email_lambda-[LAMBDA_GUIDp/...   this will not be easy to implement.  What do you propose?

thehesiod:
I added a skeleton for the impl to fix the simple test along with some TODOs, let me know what you think.

oharaandrew314:
Well, if the tests pass, then that works for me!  I feel like this "fix" could be useful now, i.e. before your WIP is completed.  I don't want to take the credit for your work, do do you want to make a PR with just the changes for that fix?

JackDanger:
@thehesiod @oharaandrew314 @toshitanian I haven't been following Alexander's work very closely so if you three can agree on how to proceed I'll happily merge the work. Thanks for putting so much care into this @thehesiod!

thehesiod:
@JackDanger Here's the status of this PR: Lambdas now work as they do in AWS however the test fails in travis (works locally) due to IP mapping issues (need to spend a few hours to resolve this, accessing outer container from inner container).  This also implements cloudwatchlogs however there are no unittest tests for it.  Not sure at what level you want this PR to be before merging.  I can't submit one without the other as the lambda changes now populate cloudwatch logs (yay).  I've been using this branch for local unittests at my company for awhile now using both new features.

JackDanger:
@thehesiod I'd love if we could pull out the Cloudwatch Logs part (I'd be happy to add the tests for that) and merge it right away. Then we can focus on the ip mapping part of this in isolation while not blocking the other work. Sound good?

JackDanger:
@thehesiod Also, THANK YOU. This is awesome.

thehesiod:
hmm.  This travis host/port thing is a PITA.  It works locally on mac + debian...starting to pull my hair out trying to get this to work.  It's going to take some work pulling out the cloudwatch given integration with lambda and my local test usage relying on both, I'll try adding some tests.

oharaandrew314:
Getting Cloudwatch logs merged first would be great.  I have some methods that I would like to add to it once it's in.

thehesiod:
ah ok, I see merge just cloudwatch, then keep this open, k, I'll create a PR with just the cloudwatch part

thehesiod:
from what I understand travis runs the base test in a docker container, which would explain why reaching out to the host doesn't work. ok, I think I know how to get this working, run moto server in a docker container, and then alias the two together.

oharaandrew314:
Yeah... docker in docker is tricky.

thehesiod:
this may be the build where the server tests pass!  \*crossing fingers\*!

thehesiod:
I swear, I think travis is trying to kill me

JackDanger:
@thehesiod You're fighting the good fight. Don't let this CI system win the day!

JackDanger:
(Watching you do this makes me so grateful that _I'm_ not doing it)

thehesiod:
lol, I dunno WTF travis did on the last one, somehow the JSON result and prints got reversed, but that should be impossible because the result is not available until after the print

thehesiod:
o-m-g, it worked, now to re-enable everything :)

thehesiod:
@JackDanger  IT PASSED, FINALLY!  OMG =)  Let me know if anything else is needed to get this merged.

thehesiod:
ugh, got a random ordering issue and I have no idea who's doing it..note how it worked in 2.7 and not 3.6, and last time in 3.6 it worked

thehesiod:
ooo, stdout + stderr are merged, I bet that's the issue

thehesiod:
cool, hopefully that does it

.travis.yml
```diff
@@ -1,23 +1,34 @@
...
     - TEST_SERVER_MODE=true
+before_install:
+  - export BOTO_CONFIG=/dev/null
 install:
-  - travis_retry pip install boto==2.45.0
-  - travis_retry pip install boto3
-  - travis_retry pip install .
-  - travis_retry pip install -r requirements-dev.txt
-  - travis_retry pip install coveralls==1.1
+  # We build moto first so the docker container doesn't try to compile it as well, also note we don't use
```
JackDanger: @thehesiod This comment is really useful, thanks

.travis.yml
```diff
@@ -1,23 +1,34 @@
...
+  # -d for docker run so the logs show up in travis
   - |
+    python setup.py sdist
     if [ "$TEST_SERVER_MODE" = "true" ]; then
-      AWS_SECRET_ACCESS_KEY=server_secret AWS_ACCESS_KEY_ID=server_key moto_server -p 5000&
+      docker run --rm -t --name motoserver -e TEST_SERVER_MODE=true -e AWS_SECRET_ACCESS_KEY=server_secret -e AWS_ACCESS_KEY_ID=server_key -v `pwd`:/moto -p 5000:5000 -v /var/run/docker.sock:/var/run/docker.sock python:3.6.2-stretch /moto/travis_moto_server.sh &
```
JackDanger: Is there a way we can use the current python version here instead of hardcoding to '3.6.2'? Currently we run in `SERVER_MODE=true` under all of the pythons listed in the `python:` YAML section above.

Maybe something like?:
```
python_version=$(python --version)
if [[ $python_version =~ " 3\." ]]; then
  docker_image="python:3.6.2-stretch"
else
  docker_image="python:????"
fi

docker run --rm -t --name motoserver \
  -e TEST_SERVER_MODE=true \
  -e AWS_SECRET_ACCESS_KEY=server_secret \
  -e AWS_ACCESS_KEY_ID=server_key \
  -v `pwd`:/moto -p 5000:5000 \
  -v /var/run/docker.sock:/var/run/docker.sock \
  ${docker_container} \
  /moto/travis_moto_server.sh &
```

setup.py
```diff
@@ -16,12 +19,21 @@
     "pytz",
     "python-dateutil<3.0.0,>=2.1",
     "mock",
+    "docker>=2.5.1"
```
JackDanger: Forgive my ignorance but I don't see any `import docker` or `from docker import ...` lines in this PR. How is the `docker` library used?

tests/test_awslambda/test_lambda.py
```diff
@@ -27,21 +29,23 @@ def get_test_zip_file1():
 def lambda_handler(event, context):
     return event
 """
-    return _process_lamda(pfunc)
+    return _process_lambda(pfunc)
```
JackDanger: oh, nice catch!

wait_for.py
```diff
@@ -0,0 +1,31 @@
+import time
```
JackDanger: This helper is great, thanks for adding it

setup.py
thehesiod: see https://github.com/spulec/moto/pull/1048/files#diff-2df06bdd95e3c78b9335da9f260a3472R6

tests/test_awslambda/test_lambda.py
thehesiod: ya did a double take until I realized it :)

thehesiod:
for some reason I didn't see a github notification on your comment, working on review changes now

thehesiod:
ok, I think I got all the requested changes

.travis.yml
thehesiod: ya, actually looks like we can re-use an existing travis env var

setup.py
JackDanger: Thanks for helping me understand!

JackDanger:
Great work, @thehesiod !

thehesiod:
very happy it's finally "done" :)

PR #1049, opened 2017-08-11
Added support for SSM delete_parameters

aleerizw:
Added support for [delete_parameters](https://docs.aws.amazon.com/systems-manager/latest/APIReference/API_DeleteParameters.html)


JackDanger:
@aleerizw This is so good! Thanks so much for adding this, it'll make a lot of people's jobs better. It's nice to have you here, welcome to Moto!

PR #1053, opened 2017-08-13
Rds stop start

rocky4570:
Basic Stop and Start RDS Instances implementation - currently doesn't create snapshot as per AWS documentation and doesn't check all instance types which are invalid for stopping/starting.


JackDanger:
@rocky4570 This is fantastic. Would you like to take a stab at implementing those snapshots? I would normally just say we should remove the commented-out `create_rds_snapshot` method (to avoid confusion that if somebody greps the code base for this method they won't know how much is implemented) and merge this but it seems like you're pretty close to finishing the whole thing!

JackDanger:
@rocky4570 This is so good! Thanks so much for this. I'll merge it right now, thanks for this contribution!

PR #1057, opened 2017-08-16
Implement methods related to listener rules at ELBv2

kawasakitoshiya:
# Implemented
- create_rule
- delete_rule
- describe_rules
- modify_rule
- set_rule_priorities

advanced validation like checking content of condition values is not implemented.


kawasakitoshiya:
test failed at `test_elbv2.test_create_listener_rules`. will check.

kawasakitoshiya:
fixed

JackDanger:
@kawasakitoshiya you are amazing! This is so, so good! I'm going to fix the failing test case around test_s3 in server mode separately. That should not block your excellent work. Thank you!

PR #1059, opened 2017-08-17
[DNM] Test travis

aleerizw:
I suspect that the pyOpenSSL package (required by the coveralls 1.2 version) is causing the builds to fail. Testing to check if the build passes.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,7 @@ install:
   - travis_retry pip install boto3
   - travis_retry pip install .
   - travis_retry pip install -r requirements-dev.txt
-  - travis_retry pip install coveralls
+  - travis_retry pip install coveralls==1.1
   - |
     if [ "$TEST_SERVER_MODE" = "true" ]; then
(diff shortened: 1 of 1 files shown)


aleerizw:
I suspect that the pyOpenSSL package (required by the coveralls 1.2 version) is causing the builds to fail. Testing to check if the build passes.

aleerizw:
The build passed so it is caused by the pyOpenSSL package. 
The following [commit](https://github.com/coveralls-clients/coveralls-python/commit/c5541263a220ff4347244d1aa70e409be115ae01) in coveralls adds urllib3[secure] which requires pyOpenSSL.

I think botocore doesn't work nicely with pyOpenSSL. I am not sure what the actual fix should be but we can either remove pyOpenSSL after installing coveralls or pin coveralls to version 1.1.

@JackDanger Please let me know what you think.

JackDanger:
@aleerizw You've just saved the day. Thanks so much for this. Great sleuthing!

Now that the build is working we can take our time to upgrade coveralls and pyopenssl and botocore more carefully. 👍

JackDanger:
Apologies for merging a PR titled '[DNM]' but master is broken and this fixes it so your caution has been overridden by you being a hero.

PR #1060, opened 2017-08-17
escape json string when sending message from sns mock to sqs mock

kawasakitoshiya:
When acquiring messages published to SQS via SNS, the message structure that can be acquired differs between when using real AWS and when mocking with moto.


kawasakitoshiya:
ok. new test failed.

kawasakitoshiya:
all tests passed except for s3 ssl error.

JackDanger:
@kawasakitoshiya This is a good fix. Thank you for your hard work on Moto!

kawasakitoshiya:
@JackDanger Thank you for your review and merging 😉

PR #1061, opened 2017-08-17
use requests version more than 2.5.0

kawasakitoshiya:
use requests version more than 2.5.0, because requests.request started supporting json argument from v2.4.2


kawasakitoshiya:
all tests passed except for  s3 ssl error.

kawasakitoshiya:
all tests passed except for s3 error on server mode.

JackDanger:
@kawasakitoshiya Thanks for this!

Issue #1062, opened 2017-08-18
In some case, elbv2 target group mock might not work well


kawasakitoshiya:
## Reporting Bugs
I will look for the cause for this error.
```
  File "/Users/kawasakitoshiya/.pyenv/versions/3.6.0/lib/python3.6/site-packages/botocore/client.py", line 312, in _api_call
    return self._make_api_call(operation_name, kwargs)
  File "/Users/kawasakitoshiya/.pyenv/versions/3.6.0/lib/python3.6/site-packages/botocore/client.py", line 588, in _make_api_call
    operation_model, request_dict)
  File "/Users/kawasakitoshiya/.pyenv/versions/3.6.0/lib/python3.6/site-packages/botocore/endpoint.py", line 141, in make_request
    return self._send_request(request_dict, operation_model)
  File "/Users/kawasakitoshiya/.pyenv/versions/3.6.0/lib/python3.6/site-packages/botocore/endpoint.py", line 170, in _send_request
    success_response, exception):
  File "/Users/kawasakitoshiya/.pyenv/versions/3.6.0/lib/python3.6/site-packages/botocore/endpoint.py", line 249, in _needs_retry
    caught_exception=caught_exception, request_dict=request_dict)
  File "/Users/kawasakitoshiya/.pyenv/versions/3.6.0/lib/python3.6/site-packages/botocore/hooks.py", line 227, in emit
    return self._emit(event_name, kwargs)
  File "/Users/kawasakitoshiya/.pyenv/versions/3.6.0/lib/python3.6/site-packages/botocore/hooks.py", line 210, in _emit
    response = handler(**kwargs)
  File "/Users/kawasakitoshiya/.pyenv/versions/3.6.0/lib/python3.6/site-packages/botocore/retryhandler.py", line 183, in __call__
    if self._checker(attempts, response, caught_exception):
  File "/Users/kawasakitoshiya/.pyenv/versions/3.6.0/lib/python3.6/site-packages/botocore/retryhandler.py", line 251, in __call__
    caught_exception)
  File "/Users/kawasakitoshiya/.pyenv/versions/3.6.0/lib/python3.6/site-packages/botocore/retryhandler.py", line 269, in _should_retry
    return self._checker(attempt_number, response, caught_exception)
  File "/Users/kawasakitoshiya/.pyenv/versions/3.6.0/lib/python3.6/site-packages/botocore/retryhandler.py", line 317, in __call__
    caught_exception)
[...]

PR #1064, opened 2017-08-18
fix bug that Names filter dont work at describe_target_groups

kawasakitoshiya:
Fixing this. https://github.com/spulec/moto/issues/1062

### DIFF
--- a/moto/elbv2/models.py
+++ b/moto/elbv2/models.py
@@ -354,7 +354,7 @@ class ELBv2Backend(BaseBackend):
             matched = []
             for name in names:
                 found = None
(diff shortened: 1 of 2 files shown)


moto/elbv2/models.py
```diff
@@ -354,7 +354,7 @@ def describe_target_groups(self, load_balancer_arn, target_group_arns, names):
             matched = []
             for name in names:
                 found = None
-                for target_group in self.target_groups:
+                for target_group in self.target_groups.values():
```
JackDanger: Great catch!

PR #1065, opened 2017-08-18
validate target group name

kawasakitoshiya:
API doc

http://docs.aws.amazon.com/elasticloadbalancing/latest/APIReference/API_CreateTargetGroup.html

```
Name
The name of the target group.

This name must be unique per region per account, can have a maximum of 32 characters, must contain only alphanumeric characters or hyphens, and must not begin or end with a hyphen.
```

### DIFF
--- a/moto/elbv2/exceptions.py
+++ b/moto/elbv2/exceptions.py
@@ -170,7 +170,7 @@ class DuplicatePriorityError(ELBClientError):
 
 class InvalidTargetGroupNameError(ELBClientError):
 
-    def __init__(self, invalid_name):
+    def __init__(self, msg):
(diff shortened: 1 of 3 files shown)


JackDanger:
@kawasakitoshiya Just leave a comment whenever you'd like this merged - this looks great!

kawasakitoshiya:
I added undocumented validation :) pls merge if it's okay.

kawasakitoshiya:
And, am gonna make some small PRs, but it's not so hurry. please review just when you have time :)

JackDanger:
@kawasakitoshiya thanks so much!

PR #1072, opened 2017-08-21
[WIP]Add if_none option to _get_list_prefix at moto.core.response

toshitanian:
*I can include this change  in my working PR, but this change can affect. So I made this independently.*

_get_list_prefix returns empty list(`[]`) even if the param_prefix not is in querystring.
We cannot recognise whether the parameter is empty list or not specified.

By adding if_none option, we can recognise whether the parameter is not passed or it is empty list.

In the case of ELBv2..

## Actions parameter is not specified

```
# method to call
conn.modify_rule(
    RuleArn=first_rule['RuleArn']
)

# querystring
{'Action': ['ModifyRule'], 'Version': ['2015-12-01'], 'RuleArn': ['...']}

# get_list_prefix
self._get_list_prefix('Actions.member', if_none=None)
> None

# returns empty list by default
self._get_list_prefix('Actions.member')
> []
```
## Actions parameter is specified but empty list
```
# method to call
conn.modify_rule(
     RuleArn=first_rule['RuleArn'],
     Actions=[]
)

#  querystring
{'Action': ['ModifyRule'], 'Version': ['2015-12-01'], 'RuleArn': ['...'], 'Actions': ['']}

# get_list_prefix
self._get_list_prefix('Actions.member', if_none=None)
> []
```

### DIFF
--- a/moto/core/responses.py
+++ b/moto/core/responses.py
@@ -272,7 +272,7 @@ class BaseResponse(_TemplateEnvironmentMixin):
                     key.replace(param_prefix, ""))] = value[0]
         return params
 
-    def _get_list_prefix(self, param_prefix):
+    def _get_list_prefix(self, param_prefix, if_none=[]):
         """
         Given a query dict like
         {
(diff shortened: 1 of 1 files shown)


toshitanian:
With my work, I didn't need this change. So, close this pr.

Issue #1070, opened 2017-08-21
elbv2 modify_rule should change attributes partically


toshitanian:
## Reporting Bugs
When only Conditions is specifies, Actions should not be empty. Now the method is implemented to be so.

### Spec
Either Actions or Conditions is specified, modify partially
Both is specified modify boto

PR #1073, opened 2017-08-21
ELBv2 modify_rule changes listener rule partially

toshitanian:
Fixing: https://github.com/spulec/moto/issues/1070

When modify_rule is called without either Actions or Conditions parameter,  or one of them is empty list, the parameter which is empty/None should be remained in the original record.


JackDanger:
Great work @toshitanian !

Issue #1075, opened 2017-08-21
NotImplementedError: Filter dicts for 'network-interface.private-dns-name'


dlutsch:
In trying to make a describe_instances call to ec2 I'm getting the following error:

NotImplementedError: Filter dicts have not been implemented in Moto for 'network-interface.private-dns-name' yet.

JackDanger:
Fixed thanks to your work, @dlutsch !

PR #1076, opened 2017-08-23
Add private-dns-name filter and fix regional DNS

dlutsch:
This is a fix for https://github.com/spulec/moto/issues/1075

- Added the 'network-interface.private-dns-name' and 'private-dns-name' ec2 filters
- Fixed public/private DNS values. This should be set depending on the region ([aws documentation](http://docs.aws.amazon.com/AmazonVPC/latest/UserGuide/VPC_DHCP_Options.html); see _Amazon DNS Server_)
- Added the region_name attribute to Instance class so that it is available to the dns setters.
- Added/updated tests

### DIFF
--- a/moto/ec2/models.py
+++ b/moto/ec2/models.py
@@ -434,7 +434,11 @@ class Instance(TaggedEC2Resource, BotoInstance):
 
     @property
     def private_dns(self):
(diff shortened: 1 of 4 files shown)


JackDanger:
@dlutsch Great work! Thanks for adding these filters - this'll be really useful for folks. Welcome to Moto! :)

Issue #998, opened 2017-07-05
Filter dicts have not been implemented in Moto for '%s' yet for instance.group-name


Prasadchitnis26:
I am attempting to use the describe_instances function in boto3. I am supplying a filter that returns a string, and am receiving the following message:

NotImplementedError: (u"Filter dicts have not been implemented in Moto for '%s' yet. Feel free to open an issue at https://github.com/spulec/moto/issues", u'instance.group-name')

So I am opening an issue.Thank you for your help.

spulec:
Thanks for opening! I'll add this as a feature request.

There are a lot of feature requests right now though so I can't make any promises on timing.

bblommers:
This was fixed by #1083, so I'll close this

PR #1083, opened 2017-08-25
Add list support to filters

dlutsch:
- allows filtering on attributes with a list value 
- fixes the ec2 'group-id' and 'instance.group-id' filters
- adds the 'instance.group-name' filter
- addresses https://github.com/spulec/moto/issues/998


JackDanger:
Great work, @dlutsch !

JackDanger:
This is now released with Moto 1.1.1

PR #1085, opened 2017-08-26
Create git tag on release

JackDanger:
I'm pretty forgetful so I doubt I'd remember to do this

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -17,3 +17,5 @@ test_server:
 
 publish:
 	python setup.py sdist bdist_wheel upload
+	git tag $(python setup.py version)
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -17,3 +17,5 @@ test_server:
 
 publish:
 	python setup.py sdist bdist_wheel upload
+	git tag $(python setup.py version)
+	git push push origin $(python setup.py version)
```
mrucci: `git push push`?

mrucci: Nevermind, fixed in https://github.com/spulec/moto/pull/1090

PR #1088, opened 2017-08-26
Support IAM account alias

toshitanian:


JackDanger:
Wow, thank you so much, @toshitanian ! You're adding so much good stuff!

Issue #1091, opened 2017-08-27
1.1.x broke update_expression in mock_dyanmodb2


ssolari:
Problem
======

1.1.x seems to have introduced a change that cause exceptions using mock_dynamodb2 that work fine with 1.0.x.   

Issue
====

It appears that the check for expression operators was changed to only accept UPPERCASE.  Both lowercase and UPPERCASE are valid however. 

Example Error
===========

```
../auth_storage.py:135: in store_login_info
    val = update_dynamo_table(self._dynamodb, USERINFO_TABLE, {USERINFO_KEY: userid}, {USERINFO_INFO: infobytes})
../../../simpaaws/simpaaws/dynamodb_access.py:109: in update_dynamo_table
    ExpressionAttributeValues=expression_values,
../../../../../../anaconda/envs/simpa/lib/python3.6/site-packages/botocore/client.py:312: in _api_call
    return self._make_api_call(operation_name, kwargs)
../../../../../../anaconda/envs/simpa/lib/python3.6/site-packages/botocore/client.py:588: in _make_api_call
    operation_model, request_dict)
../../../../../../anaconda/envs/simpa/lib/python3.6/site-packages/botocore/endpoint.py:141: in make_request
    return self._send_request(request_dict, operation_model)
../../../../../../anaconda/envs/simpa/lib/python3.6/site-packages/botocore/endpoint.py:170: in _send_request
    success_response, exception):
../../../../../../anaconda/envs/simpa/lib/python3.6/site-packages/botocore/endpoint.py:249: in _needs_retry
    caught_exception=caught_exception, request_dict=request_dict)
../../../../../../anaconda/envs/simpa/lib/python3.6/site-packages/botocore/hooks.py:227: in emit
    return self._emit(event_name, kwargs)
../../../../../../anaconda/envs/simpa/lib/python3.6/site-packages/botocore/hooks.py:210: in _emit
    response = handler(**kwargs)
../../../../../../anaconda/envs/simpa/lib/python3.6/site-packages/botocore/retryhandler.py:183: in __call__
    if self._checker(attempts, response, caught_exception):
../../../../../../anaconda/envs/simpa/lib/python3.6/site-packages/botocore/retryhandler.py:251: in __call__
    caught_exception)
../../../../../../anaconda/envs/simpa/lib/python3.6/site-packages/botocore/retryhandler.py:269: in _should_retry
    return self._checker(attempt_number, response, caught_exception)
[...]

PR #1092, opened 2017-08-27
fix uppercase only issue in update_expression #1091

ssolari:
fix for #1091


JackDanger:
@ssolari Thanks for this patch! I poked at the Travis build and it went green so I'll merge this now. Your work is much appreciated, welcome to Moto!

PR #1097, opened 2017-08-30
Fixed Dynamo mocking STS bug

### LINKED ISSUE #1095: Pytest fixtures break STS if dynamodb is mocked first
I have encountered a very bizarre bug with pytest fixtures and Moto that is difficult to debug (and find the root cause of -- still haven't gotten there yet):

With the following versions:
Python: 3.6.2
OS: macOS Sierra
boto3: 1.4.7 (Newest versions also affected)
botocore: 1.7.0 (Newest versions also affected)
moto: 1.0.1 (Newest version also affected)
pytest: 3.1.3 (Newest version also affected)

There is a very bizarre bug where if you create fixtures for `dynamodb` and `sts`, using the `dynamodb` fixture before the `sts` fixture will result in the breakage of the STS `assume_role` API.  Here is an example of this in action:

**Fixtures**:
```
@pytest.yield_fixture(scope="function")
def sts():
    with mock_sts():
        yield boto3.client("sts")

@pytest.yield_fixture(scope="function")
def iam():
    with mock_iam():
        yield boto3.client("iam")

@pytest.yield_fixture(scope="function")
def dynamodb():
    with mock_dynamodb2():
        yield boto3.client("dynamodb", region_name="us-east-1")
```

**The tests:**
```
import boto3

def test_works(iam, sts, dynamodb):
    """dynamo fixture is last PASSES"""
    iam_client = boto3.client("iam")
    iam.create_role(RoleName="somerole", AssumeRolePolicyDocument="{}")

    sts_client = boto3.client("sts")
    sts_client.assume_role(RoleArn="arn:aws:iam::012345678910:role/somerole", RoleSessionName="asdf")

def test_fails(dynamodb, iam, sts):
    """dynamo fixture is first FAILS"""
    iam_client = boto3.client("iam")
    iam.create_role(RoleName="somerole", AssumeRolePolicyDocument="{}")

    sts_client = boto3.client("sts")
    sts_client.assume_role(RoleArn="arn:aws:iam::012345678910:role/somerole", RoleSessionName="asdf")
```

**The Exception**:
Boto3 will raise a `JSONDecodeError` with the message: `Expecting value: line 1 column 1 (char 0)`:
```
self = <json.decoder.JSONDecoder object at 0x10ae77208>
s = 'Action=AssumeRole&Version=2011-06-15&RoleArn=arn%3Aaws%3Aiam%3A%3A012345678910&RoleSessionName=asdf'
idx = 0

    def raw_decode(self, s, idx=0):
        """Decode a JSON document from ``s`` (a ``str`` beginning with
            a JSON document) and return a 2-tuple of the Python
            representation and the index in ``s`` where the document ended.
    
            This can be used to decode a JSON document from a string that may
            have extraneous data at the end.
    
            """
        try:
>           obj, end = self.scan_once(s, idx)
E           StopIteration: 0
```

For some reason, if the dynamo fixture is first, then the body for the `assume_role` API request isn't in a JSON format (which it apparently needs to be). This causes boto3 to flip out. However, this **only** happens if the dynamo fixture is first.

Also odd: this only happens with fixtures. Simply mocking within the function causes no issues.

At the very least, you can work around this by having the `dynamodb` fixture after `sts`, but it is still very odd that mocking DynamoDB as a fixture first breaks `assume_role`, and maybe even other APIs.

Still not sure where the cause of the bug lies, and debugging this has not proven to be that simple, unfortunately.

mikegrima:
Fixes #1095

Also removed some older STS related logic from Dynamo that wasn't relevant.

### DIFF
--- a/moto/dynamodb/responses.py
+++ b/moto/dynamodb/responses.py
@@ -7,33 +7,6 @@ from moto.core.utils import camelcase_to_underscores
 from .models import dynamodb_backend, dynamo_json_dump
 
 
-GET_SESSION_TOKEN_RESULT = """
-<GetSessionTokenResponse xmlns="https://sts.amazonaws.com/doc/2011-06-15/">
- <GetSessionTokenResult>
- <Credentials>
- <SessionToken>
- AQoEXAMPLEH4aoAH0gNCAPyJxz4BlCFFxWNE1OPTgk5TthT+FvwqnKwRcOIfrRh3c/L
- To6UDdyJwOOvEVPvLXCrrrUtdnniCEXAMPLE/IvU1dYUg2RVAJBanLiHb4IgRmpRV3z
- rkuWJOgQs8IZZaIv2BXIa2R4OlgkBN9bkUDNCJiBeb/AXlzBBko7b15fjrBs2+cTQtp
- Z3CYWFXG8C5zqx37wnOE49mRl/+OtkIKGO7fAE
- </SessionToken>
- <SecretAccessKey>
- wJalrXUtnFEMI/K7MDENG/bPxRfiCYzEXAMPLEKEY
- </SecretAccessKey>
- <Expiration>2011-07-11T19:55:29.611Z</Expiration>
- <AccessKeyId>AKIAIOSFODNN7EXAMPLE</AccessKeyId>
- </Credentials>
- </GetSessionTokenResult>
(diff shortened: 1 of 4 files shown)


JackDanger:
Great fix, @mikegrima! Thanks for figuring this out and fixing it.

PR #1099, opened 2017-08-31
fix TargetGroupNotFoundError is not defined correctly

toshitanian:
error log
```
  File "/Users/kawasakitoshiya/workspace/moto/moto/elbv2/exceptions.py", line 45, in __init__
    super(TooManyTagsError, self).__init__(
TypeError: super(type, obj): obj must be an instance or subtype of type
```


JackDanger:
@toshitanian I owe you one for fixing this bug I introduced - thank you!

PR #1103, opened 2017-09-01
handle short form function in cfn yaml template.

toshitanian:
fix bug that create_stack fails when call it with yaml template which includes shorthand intrinsic function such as `!Join`.


moto/cloudformation/responses.py
```diff
@@ -391,8 +391,7 @@ def list_exports(self):
 
 GET_TEMPLATE_RESPONSE_TEMPLATE = """<GetTemplateResponse>
   <GetTemplateResult>
-    <TemplateBody>{{ stack.template }}
-    </TemplateBody>
+    <TemplateBody>{{ stack.template }}</TemplateBody>
```
JackDanger: Ooooh, good fix.

JackDanger:
I am very impressed. Thanks for this, @toshitanian!

PR #1104, opened 2017-09-01
set None to Volumes property of ecs task-definition when called from cfn template

toshitanian:
Volumes property of ecs task-definition is not required paramter in cfn template

http://docs.aws.amazon.com/ja_jp/AWSCloudFormation/latest/UserGuide/aws-resource-ecs-taskdefinition.html#cfn-ecs-taskdefinition-volumes


JackDanger:
Thanks for this!

PR #1110, opened 2017-09-06
Fixed S3 lifecycle error message.

### LINKED ISSUE #1109: S3 GetBucketLifecycleConfiguration does not return proper error
AWS's API returns an HTTP 404 if an S3 bucket does not contain a lifecycle configuration. The error message returned by the API is a `NoSuchLifecycleConfiguration`.  However, it needs to be returned as an XML object as defined here: http://docs.aws.amazon.com/AmazonS3/latest/API/ErrorResponses.html.

In https://github.com/spulec/moto/blob/master/moto/s3/responses.py#L191, because this is not an XML string, `botocore` will not properly pass down the `NoSuchLifecycleConfiguration` error.

I am going to submit a PR to resolve this to be consistent with the AWS API (mostly going to copypasta the `S3_NO_POLICY` response).

mikegrima:
Fixes #1109 

Tested internally and works properly.

### DIFF
--- a/moto/s3/exceptions.py
+++ b/moto/s3/exceptions.py
@@ -91,3 +91,23 @@ class EntityTooSmall(S3ClientError):
             "EntityTooSmall",
             "Your proposed upload is smaller than the minimum allowed object size.",
             *args, **kwargs)
+
+
+class InvalidRequest(S3ClientError):
+    code = 400
+
+    def __init__(self, method, *args, **kwargs):
+        super(InvalidRequest, self).__init__(
+            "InvalidRequest",
+            "Found unsupported HTTP method in CORS config. Unsupported method is {}".format(method),
+            *args, **kwargs)
+
+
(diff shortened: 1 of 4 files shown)


mikegrima:
While I'm at it...I'm also adding in the `put_bucket_tagging` functionality in.

mikegrima:
Added `put_bucket_tagging`... There are a bunch of other API calls that are missing.

I guess I should just add them all in.  @spulec Would you be able to quickly merge and push to PyPI once I'm done? I'm working on major S3 related things that we intend on open sourcing, and extensively using moto for my tests.  The only other way for me to continue is to mock out all of the S3 related calls and skip moto.

mikegrima:
Added CORS support.

JackDanger:
This is awesome, @mikegrima! I'll release a new version of Moto immediately with your fixes

PR #1112, opened 2017-09-06
add support for ssm get_parameter

oharaandrew314:
Added support for ssm [GetParameter](https://docs.aws.amazon.com/systems-manager/latest/APIReference/API_GetParameter.html)


oharaandrew314:
I'm guessing that server mode doesn't like how I'm raising an exception.  I'll try to find examples of how it's done in other places.

Issue #1113, opened 2017-09-06
request.body should be request.data?


kefo:
I'm using `moto_server glacier` (or trying to) and repeatedly saw the following stack trace from `moto_server` whenever I tried to create a glacier archive:

```
127.0.0.1 - - [06/Sep/2017 14:40:32] "POST /kefo/vaults/testvault/archives HTTP/1.1" 500 -
Error on request:
Traceback (most recent call last):
  File "/Users/kford1/Work/motoenv/lib/python3.5/site-packages/werkzeug/serving.py", line 209, in run_wsgi
    execute(self.server.app)
  File "/Users/kford1/Work/motoenv/lib/python3.5/site-packages/werkzeug/serving.py", line 197, in execute
    application_iter = app(environ, start_response)
  File "/Users/kford1/Work/motoenv/lib/python3.5/site-packages/moto/server.py", line 92, in __call__
    return backend_app(environ, start_response)
  File "/Users/kford1/Work/motoenv/lib/python3.5/site-packages/flask/app.py", line 1997, in __call__
    return self.wsgi_app(environ, start_response)
  File "/Users/kford1/Work/motoenv/lib/python3.5/site-packages/flask/app.py", line 1985, in wsgi_app
    response = self.handle_exception(e)
  File "/Users/kford1/Work/motoenv/lib/python3.5/site-packages/flask/app.py", line 1540, in handle_exception
    reraise(exc_type, exc_value, tb)
  File "/Users/kford1/Work/motoenv/lib/python3.5/site-packages/flask/_compat.py", line 33, in reraise
    raise value
  File "/Users/kford1/Work/motoenv/lib/python3.5/site-packages/flask/app.py", line 1982, in wsgi_app
    response = self.full_dispatch_request()
  File "/Users/kford1/Work/motoenv/lib/python3.5/site-packages/flask/app.py", line 1614, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "/Users/kford1/Work/motoenv/lib/python3.5/site-packages/flask/app.py", line 1517, in handle_user_exception
    reraise(exc_type, exc_value, tb)
  File "/Users/kford1/Work/motoenv/lib/python3.5/site-packages/flask/_compat.py", line 33, in reraise
    raise value
  File "/Users/kford1/Work/motoenv/lib/python3.5/site-packages/flask/app.py", line 1612, in full_dispatch_request
    rv = self.dispatch_request()
[...]

PR #1117, opened 2017-09-07
Use 'data' attribute of Flask Request object.

kefo:
This PR addresses the issue raised here:  https://github.com/spulec/moto/issues/1113

In short, the Flask Request object does not support a 'body' attribute. Changed here to 'data'.

### DIFF
--- a/tests/test_glacier/test_glacier_jobs.py
+++ b/tests/test_glacier/test_glacier_jobs.py
@@ -84,7 +84,7 @@ def test_get_job_output():
     })
     job_id = job_response['JobId']
 
-    time.sleep(62)
+    time.sleep(6)
     
     output = conn.get_job_output(vault_name, job_id)
     output.read().decode("utf-8").should.equal("some stuff")
--- a/tests/test_glacier/test_glacier_jobs.py
+++ b/tests/test_glacier/test_glacier_jobs.py
@@ -40,25 +40,11 @@ def test_describe_job():
     job_id = job_response['JobId']
 
     job = conn.describe_job(vault_name, job_id)
-    json.loads(job.read().decode("utf-8")).should.equal({
-        'CompletionDate': '2013-03-20T17:03:43.221Z',
-        'VaultARN': None,
-        'RetrievalByteRange': None,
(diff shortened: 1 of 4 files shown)


JackDanger:
@kefo This PR is good! Please take a look at the test failures and adjust the code, there isn't much further to go.

kefo:
Over to you @JackDanger

tests/test_glacier/test_glacier_jobs.py
```diff
@@ -96,5 +84,7 @@ def test_get_job_output():
     })
     job_id = job_response['JobId']
 
+    time.sleep(62)
```
JackDanger: @kefo Why do we need this? Is there a way to accomplish the same thing without adding a minute to all test runs?

terrycain: We could use freezegun to test this (though it would have to raise SkipTest when in server mode)
E.g. https://github.com/spulec/moto/blob/master/tests/test_sqs/test_sqs.py#L760-L796

kefo: I wasn't crazy about it either; I'm open to ideas.

The pull request, in part, makes moto glacier work more like aws glacier, which is to say that when you now initiate a job to retrieve an archive, moto glacier does not immediately return the content.  Instead, a job is created (like aws), complete with an InProgress status.  You can query the job, etc.  After a minute (the 62 seconds you see in the test), it flips to a 'Completed' status and the client can fetch the archived content.

JackDanger: What if we just used a shorter timeout? So keep the same functionality but change
```
self.et = self.st + datetime.timedelta(minutes=1)
```
to
```
self.et = self.st + datetime.timedelta(seconds=5)
```

JackDanger: @kefo I have just one question about whether we can lower the sleep from 62 seconds to, say, 5 seconds. If we can, great. If not, that's fine too - either way we can merge this once we resolve that. Thanks!

PR #1123, opened 2017-09-09
Handle starting token in the 'list-objects' query

julienduchesne:


JackDanger:
@julienduchesne Thank you so much for this! You've just improved this library for many thousands of users. Welcome to Moto!

julienduchesne:
Thanks!

JackDanger:
This is now available in Moto version 1.1.4

Issue #1126, opened 2017-09-11
Dynamodb2 update_item does not support ADD attribute update for string sets


brianrower:
When if doing an "ADD" to a string set, moto's dynamodb2 raises NotImplementedError:
"NotImplementedError: ADD not supported for SS"

Example code:

table.put_item(Item={ 
        'forum_name': 'the-key', 
        'subject': '123', 
        'string_set': set(['str1', 'str2']), 
    }) 

table.update_item( 
        Key=item_key, 
        AttributeUpdates={ 
            'string_set': { 
                'Action': u'ADD', 
                'Value': set(['str3']), 
            }, 
        }, 
    ) 

moto 0.4.31
Python 3.6.1

PR #1127, opened 2017-09-11
Dynamodb2 - Allow doing an ADD update_item of a string set

brianrower:
Fix issue #1126 
Add support for doing an update_item ADD for a string set in dynamodb2


JackDanger:
@brianrower This is very helpful. Thanks for the great contribution and welcome to Moto!

JackDanger:
This is now available in Moto version 1.1.4

PR #1133, opened 2017-09-13
WIP - do not merge. Please review my refactor for cloudwatch put_metric_data

chrisevett:
I refactored the way put_metric_data works so I can implement get_metric_statistics. I figured I'd get you guys' input on what I have in place already before I get started on the rest. 

This is babby's first python project so keep that in mind if you see anything that isnt idiomatic python. If it looks funny then it probably is 😆  🍼 

Some todos that I already know about:

Util.py needs tests
There are a bunch of commented out tests, these will be repurposed for get_metric_statistics and probably end up in their own file.
I have several test_fixture methods that will definitely change so its not so clunky 
The __str__ methods were for testing, i don't know if they should stay or not.

Anyway, please review and let me know if there are any major issues before I continue. Again, this refactor was to add support for all of the data in get_metric_statistics that didnt exist before.

### DIFF
--- a/moto/cloudwatch/models.py
+++ b/moto/cloudwatch/models.py
@@ -9,12 +9,28 @@ class Dimension(object):
         self.name = name
         self.value = value
 
+    def __str__(self):
+        return "{} {}".format(self.name, self.value)
+
 
 class FakeAlarm(BaseModel):
 
-    def __init__(self, name, namespace, metric_name, comparison_operator, evaluation_periods,
-                 period, threshold, statistic, description, dimensions, alarm_actions,
-                 ok_actions, insufficient_data_actions, unit):
+    def __init__(
+            self,
+            name,
+            namespace,
+            metric_name,
+            comparison_operator,
+            evaluation_periods,
+            period,
+            threshold,
+            statistic,
+            description,
+            dimensions,
+            alarm_actions,
+            ok_actions,
+            insufficient_data_actions,
+            unit):
         self.name = name
         self.namespace = namespace
         self.metric_name = metric_name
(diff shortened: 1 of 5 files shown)


moto/cloudwatch/models.py
```diff
@@ -44,18 +60,127 @@ def __init__(self, namespace, name, value, dimensions):
...
+
+    def __init__(self, namespace, data):
+        self.namespace = namespace
+        self.data = data
+
+
+class MetricDatumv2(BaseModel):
+
+    def __init__(self, namespace, metricdata):
+        self.Namespace = namespace
```
JackDanger: It's unusual to have object attributes be uppercased like this. In this situation I'd expect `Namespace` to be a hardcoded class name. It makes me think this is how the code works:
```
class ANamespace:
    pass
class MetricDatumv2(BaseModel)
    def __init__(self):
        print(self.Namespace)
MetricDatumv2.Namespace = ANamespace
MetricDatumv2()  # prints 'ANamespace'
```

So I recommend using lowercase values for this and all similar variables throughout.

moto/cloudwatch/models.py
```diff
@@ -44,18 +60,127 @@ def __init__(self, namespace, name, value, dimensions):
...
+    # parse the metricdata object
+
+        self.MetricData = [
```
JackDanger: Same here about the uppercase. `Metric` is great as a class name but I'd recommend `self.metric_data = Metric(...`

moto/cloudwatch/models.py
```diff
@@ -44,18 +60,127 @@ def __init__(self, namespace, name, value, dimensions):
...
+            for metric in metricdata]
+
+    def __str__(self):
+        return "{} {}".format(self.Namespace, self.MetricData)
+
+
+class Metric():
+    def __init__(
+            self,
+            MetricName,
```
JackDanger: Same here for the variable names. In the boto3 api the keyword arguments to many of the API calls are camel cased but it's pretty unusual in Python and you won't find it in the Moto code base generally.

moto/cloudwatch/models.py
```diff
@@ -102,10 +227,13 @@ def delete_alarms(self, alarm_names):
         for alarm_name in alarm_names:
             self.alarms.pop(alarm_name, None)
 
-    def put_metric_data(self, namespace, metric_data):
-        for name, value, dimensions in metric_data:
-            self.metric_data.append(MetricDatum(
-                namespace, name, value, dimensions))
+    def put_metric_data(self, Namespace, MetricData):
+        for datapoint in MetricData:
```
JackDanger: Do you need this instead?:
```
for datapoint in MetricData:
    self.metric_data.append(MetricDatumv2(Namespace, datapoint))
```

JackDanger:
@chrisevett This is super good work. Thank you for tackling this! If you could make sure that old functionality still works so we don't go breaking somebody's test suite then your improvements are mergeable. I believe this means uncommenting the old tests and ensuring those code paths still work. This is great though and it's a huge help to everybody who uses AWS!

chrisevett:
Thanks for the feed back. I will make the changes you requested and then send an update with the get_metric_statistics implementation included.

moto/cloudwatch/models.py
thehesiod: ya it's against PEP, @JackDanger perhaps should enable some PEP flake tests?

moto/cloudwatch/models.py
```diff
@@ -44,18 +60,127 @@ def __init__(self, namespace, name, value, dimensions):
...
+    def __init__(self, namespace, data):
+        self.namespace = namespace
+        self.data = data
+
+
+class MetricDatumv2(BaseModel):
+
+    def __init__(self, namespace, metricdata):
+        self.Namespace = namespace
+    # parse the metricdata object
```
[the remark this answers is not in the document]
thehesiod: another thing that can be caught by linter

moto/cloudwatch/models.py
JackDanger: The test suite runs `flake8 ./moto` before every run and I'm not sure why this isn't being caught. Any ideas? Yes, we should absolutely tighten the flake8 constraints for developers. Maybe even adding autopep8 to help out would be useful.

thehesiod: just created https://github.com/spulec/moto/pull/1232

thehesiod:
not sure how I didn't see this PR before I implemented something similar, @chrisevett could you check out what recently merged?

chrisevett:
@thehesiod Sure thing. For some reason I thought we had merged this already but it looks like i was mistaken.

chrisevett:
@JackDanger @thehesiod since this is so old I'm going to close this and open a new PR so I don't have to deal with funky merges/rebasing

PR #1134, opened 2017-09-13
Add support for the ADD and DELETE operations in dynamodb update expressions

### LINKED ISSUE #766: DynamoDB update_item() does not support ADD and DELETE actions
The action list here https://github.com/spulec/moto/blob/master/moto/dynamodb2/models.py#L113 only includes SET and REMOVE.

Mostly this fails silently (that is, the code runs but the field is not updated), but it can also raise `ValueError: need more than 1 value to unpack` from the decorator if I have an UpdateExpression like `SET set_field=:set_val ADD add_field :add_val`.

brianrower:
Add support for the ADD and DELETE operations in dynamodb update expressions
Fixes #766

### DIFF
--- a/moto/dynamodb2/models.py
+++ b/moto/dynamodb2/models.py
@@ -150,7 +150,7 @@ class Item(BaseModel):
                     else:
                         self.attrs[key] = DynamoType({"S": value})
                 elif action == 'ADD':
-                    key, value = value.split(" ", maxsplit=1)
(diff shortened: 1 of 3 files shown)


JackDanger:
@brianrower This is really great. Thanks for this, you're making Moto's DynamoDB support really first rate. I'm merging and will release a new version immediately.

JackDanger:
Release as part of Moto 1.1.6

PR #1140, opened 2017-09-14
Added keypair filtering

nuwang:
This PR adds keypair filtering support by key-name and fingerprint.
Also improves error reporting if an InvalidKeyPairNameError occurs

### DIFF
--- a/moto/ec2/models.py
+++ b/moto/ec2/models.py
@@ -795,16 +795,29 @@ class InstanceBackend(object):
         return reservations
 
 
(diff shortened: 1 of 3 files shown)


JackDanger:
@nuwang This is really great, thank you for this!

JackDanger:
This is now released as Moto 1.1.7

nuwang:
@JackDanger Thanks for the super quick release - made my life much easier!

PR #1143, opened 2017-09-15
add partial add_permission and remove_permission to lambda client

oharaandrew314:
Adds a partial implementation of the lambda client `add_permission` and `remove_permission` methods.

The permissions do not have any effect on the lambda, and only the required attributes of the methods are supported.


moto/awslambda/urls.py
```diff
@@ -12,5 +12,7 @@
     '{0}/(?P<api_version>[^/]+)/functions/(?P<function_name>[\w_-]+)/?$': response.function,
     '{0}/(?P<api_version>[^/]+)/functions/(?P<function_name>[\w_-]+)/invocations/?$': response.invoke,
     '{0}/(?P<api_version>[^/]+)/functions/(?P<function_name>[\w_-]+)/invoke-async/?$': response.invoke_async,
-    '{0}/(?P<api_version>[^/]+)/tags/(?P<resource_arn>.+)': response.tag
+    '{0}/(?P<api_version>[^/]+)/tags/(?P<resource_arn>.+)': response.tag,
+    '{0}/(?P<api_version>[^/]+)/functions/(?P<function_name>[\w_-]+)/policy?$': response.policy,
```
oharaandrew314: Is there a way I can merge these last two statements?  I've tried
`{0}/(?P<api_version>[^/]+)/functions/(?P<function_name>[\w_-]+)/policy(/?P<sid>[\w_-]+)?

but it doesn't capture the case when the sid is provided.

dbfr3qs:
Did you try something like in urls.py:
```python
'{0}/(?P<api_version>[^/]+)/functions/(?P<function_name>[\w_-]+)/policy/?$': response.policy
```

This is the endpoint that I see while using a debugger to trace through the call to add_permission.
```bash
'https://lambda.us-west-2.amazonaws.com/2015-03-31/functions/testFunction/policy'
```

PR #1146, opened 2017-09-15
Ensure root device is listed in block device mappings

nuwang:


JackDanger:
I'm not very familiar with this API but the Example #1 on this page shows a root device of `/dev/sda` and then a deviceName of `/dev/sda1`: https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_DescribeImages.html

JackDanger:
We can merge this PR as-is because Example #2 on that page matches what your changes suggest. I just wonder if either the AWS docs are wrong (which happens) or if there's something else going on here.

nuwang:
@JackDanger I suspect that the example may be wrong. The docs [here](http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/block-device-mapping-concepts.html#ami-block-device-mapping) state the following:
"Each AMI has a block device mapping that specifies the block devices to attach to an instance when it is launched from the AMI. An AMI that Amazon provides includes a root device only. To add more block devices to an AMI, you must create your own AMI."

The docs [here](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/Storage.html?icmpid=docs_ec2_console) say that "Every time you launch an instance from an AMI, a root storage device is created for that instance. The root storage device contains all the information necessary to boot the instance"

Putting the two things together, I think it's safe to assume that the root device always appears in block device mappings for the AMI, although I couldn't find an unambiguous statement stating this explicitly. It's also consistent with how things work through the console and APIs.

JackDanger:
@nuwang Thanks for figuring this out! Once again Moto developers eclipse the AWS docs in accuracy 👍

PR #1148, opened 2017-09-15
Improved support for VPC Address filtering

nuwang:
This PR improves filtering for DescribeAddresses.
Also fixes issue where the address's eni is None despite being attached.
Includes tests.

### DIFF
--- a/moto/ec2/models.py
+++ b/moto/ec2/models.py
@@ -2959,6 +2959,27 @@ class ElasticAddress(object):
             return self.allocation_id
         raise UnformattedGetAttTemplateException()
 
+    def get_filter_value(self, filter_name):
+        if filter_name == 'allocation-id':
+            return self.allocation_id
+        elif filter_name == 'association-id':
+            return self.association_id
+        elif filter_name == 'domain':
+            return self.domain
+        elif filter_name == 'instance-id' and self.instance:
+            return self.instance.id
+        elif filter_name == 'network-interface-id' and self.eni:
+            return self.eni.id
+        elif filter_name == 'network-interface-owner-id':
+            msg = "The filter '{0}' for DescribeAddresses has not been" \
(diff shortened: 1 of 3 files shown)


JackDanger:
This is great, @nuwang!

moto/ec2/models.py
```diff
@@ -2959,6 +2959,27 @@ def get_cfn_attribute(self, attribute_name):
...
+            return self.allocation_id
+        elif filter_name == 'association-id':
+            return self.association_id
+        elif filter_name == 'domain':
+            return self.domain
+        elif filter_name == 'instance-id' and self.instance:
+            return self.instance.id
+        elif filter_name == 'network-interface-id' and self.eni:
+            return self.eni.id
+        elif filter_name == 'network-interface-owner-id':
```
dbfr3qs: This is fantastic. For future proofing I'm wondering if this could be a catch all final else, instead of specific to 'network-interface-id', to catch any future filters added to Boto3. So simply remove this code and replace at the bottom of the function with:

```python
else:
    msg = "The filter '{0}' for DescribeAddresses has not been" \
          " implemented in Moto yet. Feel free to open an issue at" \
          " https://github.com/spulec/moto/issues".format(filter_name)
    raise NotImplementedError(msg)

nuwang: @dbfr3qs Thanks for highlighting this. I was wondering whether to do this, but then didn't because util.generic_filter() doesn't. Looking at other implementations of get_filter_value, some are repeating the exception handling, and others ignore unknown filters. What do you think of moving the unrecognised filter exception handling to the generic filter, so that it's made consistent across the board? I can issue a PR if that sounds ok.

nuwang: Done in PR: https://github.com/spulec/moto/pull/1150

JackDanger:

nuwang:
Excellent! Thanks for letting me know.

PR #1149, opened 2017-09-16
Fix for instance public ip not being cleared on eip disassociation

nuwang:
This PR fixes the following issues
1. When disassociating an EIP, the instance's public ip address does not get cleared
2. Cleans up a duplicated assignment of instance public ip during EIP association
3. Adds tests to enforce these


JackDanger:
Wonderful!

JackDanger:

PR #1152, opened 2017-09-16
Refactor querystring params

nuwang:
This PR cleans up duplication of logic between base response and ec2 util classes.
It replaces functions such as sequence_from_querystring with the inherited _get_multi_param.
It also replaces direct querystring index based access with _get_param where possible.

### DIFF
--- a/moto/ec2/utils.py
+++ b/moto/ec2/utils.py
@@ -174,62 +174,6 @@ def split_route_id(route_id):
     return values[0], values[1]
 
 
-def instance_ids_from_querystring(querystring_dict):
-    instance_ids = []
-    for key, value in querystring_dict.items():
-        if 'InstanceId' in key:
-            instance_ids.append(value[0])
-    return instance_ids
-
-
-def image_ids_from_querystring(querystring_dict):
-    image_ids = []
-    for key, value in querystring_dict.items():
-        if 'ImageId' in key:
-            image_ids.append(value[0])
-    return image_ids
-
-
-def executable_users_from_querystring(querystring_dict):
-    user_ids = []
-    for key, value in querystring_dict.items():
-        if 'ExecutableBy' in key:
-            user_ids.append(value[0])
-    return user_ids
-
-
-def route_table_ids_from_querystring(querystring_dict):
-    route_table_ids = []
-    for key, value in querystring_dict.items():
-        if 'RouteTableId' in key:
-            route_table_ids.append(value[0])
-    return route_table_ids
-
-
-def network_acl_ids_from_querystring(querystring_dict):
-    network_acl_ids = []
-    for key, value in querystring_dict.items():
-        if 'NetworkAclId' in key:
-            network_acl_ids.append(value[0])
-    return network_acl_ids
-
-
-def vpc_ids_from_querystring(querystring_dict):
-    vpc_ids = []
-    for key, value in querystring_dict.items():
-        if 'VpcId' in key:
-            vpc_ids.append(value[0])
-    return vpc_ids
-
-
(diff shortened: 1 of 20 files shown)


moto/ec2/utils.py
```diff
@@ -174,62 +174,6 @@ def split_route_id(route_id):
     return values[0], values[1]
 
 
-def instance_ids_from_querystring(querystring_dict):
```
JackDanger: This is the best part!!!

JackDanger:
@nuwang I'm speechless. This is going to make this part of the code base (arguably the most central and important) more accessible for all other developers. This is just so useful!

nuwang:
@JackDanger Thanks for reviewing. I think that the following enhancements could also be made, and I'll note them here for future reference.
1. Since all AWS parameters have a fixed set of formats, a standard utility class for parsing them would facilitate reuse. I think the following primitives would be beneficial.
    a. get_param() - get a single parameter
    b. get_param_array() - to get an array of parameters
    c. get_param_dict() - to get a compound dict structure, including all nested values.
While all of the above exist in various forms, I think that they need to be documented and moved to a central, reusable location. (The inherited methods are less usable when outside of the BaseResponse class)
2. If the above can be done, many of the methods in ec2.util could be removed, or composed out of the above primitives. The methods that can be refactored include: filters_from_querystring, tags_from_query_string, dhcp_configuration_from_querystring, dict_from_querystring, tags_from_query_string.
3. Doubtless this will apply to the services other than ec2.

Perhaps this should be filed as an issue tagged "Future enhancements" instead?

JackDanger:
I think that's a great proposal and worth creating an issue for that (or even multiple issues for each piece of it).

JackDanger:

tardyp:
Hi guys, I got regressions of buildbot tests with this code..
```
    def _get_multi_param(self, param_prefix):
        """
        Given a querystring of ?LaunchConfigurationNames.member.1=my-test-1&LaunchConfigurationNames.member.2=my-test-2
        this will return ['my-test-1', 'my-test-2']
        """
```
This does not look equivalent to instance_ids_from_querystring, which finds InstanceId without ".1", ".2" suffixing.

my tests are fixed if I replace

        instance_id = self._get_multi_param('InstanceId')

by

        instance_id = self._get_param('InstanceId')

in moto/ec2/responses/general.py

nuwang:
@tardyp Just checked the [docs](http://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_GetConsoleOutput.html) and it should indeed be get_param. I think the tests failed without _get_multi_param though, let me check and get back to you.

nuwang:
@tardyp @JackDanger So it looks like boto is sending in a list param:
https://github.com/boto/boto/blob/abb38474ee5124bb571da0c42be67cd27c47094f/boto/ec2/connection.py#L1061
Although the latest docs suggest that it should be sending in a single param.
Boto3 seems to be sending in a single param: https://github.com/boto/boto3/blob/d6f3c2bc02562d83d42226bea3b46e121eb60b2f/boto3/data/ec2/2016-11-15/resources-1.json#L689

I guess we could special case this to work in both scenarios? To prevent future regressions, adding a test for boto3 would probably sort out both.

nuwang:
@tardyp @JackDanger I've issued a PR with a fix: https://github.com/spulec/moto/pull/1201

tardyp:
Thanks for quick fix! When do you intend to push this to pypi?

JackDanger:
@tardyp I just released it as Moto version 1.1.15

Enjoy!

JackDanger:
And thanks for saving the day, @nuwang!

PR #1156, opened 2017-09-18
Added SSM GetParameter support

terrycain:
SSM now has GetParameter (boto added get_parameter) so I added it here.

Any issues with this?

### DIFF
--- a/moto/ssm/models.py
+++ b/moto/ssm/models.py
@@ -75,6 +75,11 @@ class SimpleSystemManagerBackend(BaseBackend):
                 result.append(self._parameters[name])
         return result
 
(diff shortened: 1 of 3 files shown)


oharaandrew314:
Can you add a test case which gets a parameter that does not exist?  This gave me trouble the last time I tried to implement it.

oharaandrew314:
Ah.  I see what I was missing now.

JackDanger:
Great work, @terrycain ! This is super helpful. Welcome to Moto!

terrycain:
No problem, I'll throw you some more PRs if I come across anything else :smile:

JackDanger:
This is now released as part of Moto 1.1.11

PR #1157, opened 2017-09-18
Raise InvalidGroup.NotFound in DescribeSecurityGroups

nuwang:
This PR fixes an issue where DescribeSecurityGroups does not raise an InvalidSecurityGroup exception when invalid ids/names are supplied, returning an empty list instead. Also fixes the same for snapshots, volumes, vpcs and subnets.

### DIFF
--- a/moto/ec2/models.py
+++ b/moto/ec2/models.py
@@ -1360,22 +1360,25 @@ class SecurityGroupBackend(object):
         return group
 
     def describe_security_groups(self, group_ids=None, groupnames=None, filters=None):
-        all_groups = itertools.chain(*[x.values()
-                                       for x in self.groups.values()])
-        groups = []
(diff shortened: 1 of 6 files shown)


nuwang:
I've found a few more places where the appropriate error is not being raised. I'll update the pull request to include these too.

nuwang:
Fixed in commits: https://github.com/spulec/moto/pull/1157/commits/08c4eff0b26518cf14c7eacaa1eb060bfc17ea50 and: https://github.com/spulec/moto/pull/1157/commits/ca56955a97c98057bffb7b3b9bba9f49fd8212bf

JackDanger:
This is really great!

JackDanger:
This is now released as part of Moto 1.1.11

nuwang:
That's great! Thanks again for the super fast response.

PR #1160, opened 2017-09-19
Shrink Dockerfile

terrycain:
Have switched the Dockerfile from using Debian to Alpine which is far smaller. Also switched to Python3 but seeing as we run tests against both I see no issue with that.

After switching to Alpine the image size has gone down from 298MB to 34MB. 
My test repo on DockerHub - https://hub.docker.com/r/terrycain/moto/

It would be nice for someone with write access or possibly it might need to be the owner to setup the project with DockerHub, its pretty easy, you link it with GitHub, then pushes to master will add to the "latest" image, and you can configure it to make tagged released based on tags. Perhaps setting up an organisation on DockerHub might be better, haven't done that so unsure if it'll be of any use.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,11 +1,15 @@
-FROM python:2
+FROM alpine:3.6
 
 ADD . /moto/
 ENV PYTHONUNBUFFERED 1
 
 WORKDIR /moto/
-RUN pip install ".[server]"
+RUN  apk add --no-cache python3 && \
+     python3 -m ensurepip && \
+     rm -r /usr/lib/python*/ensurepip && \
+     pip3 --no-cache-dir install --upgrade pip setuptools && \
+     pip3 --no-cache-dir install ".[server]"
 
-CMD ["moto_server"]
+ENTRYPOINT ["/usr/bin/moto_server"]
 
 EXPOSE 5000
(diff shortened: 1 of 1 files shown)


terrycain:
Don't merge this yet. It all works but I want to get moto to run on 0.0.0.0 by default inside the container so then when you expose ports it works nicer.

terrycain:
Example run command is `docker run -p 5000:5000 --name moto terrycain/moto:test ec2`

JackDanger:
Thank you for this, @terrycain!

JackDanger:
This is now released as Moto 1.1.12

(I assume the addition of that last commit meant it was mergeable - if not, let me know and I'll jump in and help)

terrycain:
Yeah it's good to go. All we really need now is either possibly you or the owner to setup a dockerhub account. As I can only automate docker builds of my fork

JackDanger:
I've created https://hub.docker.com/r/motoserver/moto/ and added you as an organization owner, @terrycain. @spulec I tried to add you but it looks like your email isn't connected to Dockerhub. No worries here, I just want to make sure you have access to everything.

I've pushed the current version and this should now be usable (!!): `docker pull motoserver/moto`

I wish I could have registered `docker pull moto/server` but the "moto" username is taken and unused :(

terrycain:
Nice just saw. What we really want is an automated build on dockerhub, (click create in the upper right > automated build) if it shows moto when you link your github account then it will build and tag without us needing to push to it (though we could just get travis to do that). But I think to get dockerhub to do it, we'd need either @spulec to add it, or for moto to be in an organisation and us added to said org.

PR #1161, opened 2017-09-19
[WIP] Initialize script for service/operation like scaffold in Ruby on Rails

toshitanian:
Making a script like RoR's scaffold.

Implementing new operation of boto3 to moto3 is done by:
- adding function, which name s same as boto3's,  to `models.py` and `responses.py`
- adding response template
- adding test of service
- etc...

We can make it more effective by making a script like RoR's scaffold. It'll generates appropriate functions and templates in library and test directories, and at README if possible.

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -9,3 +9,5 @@ flask
 boto3>=1.4.4
 botocore>=1.5.77
 six>=1.9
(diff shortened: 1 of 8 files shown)


toshitanian:
Made issue https://github.com/spulec/moto/issues/1163, closes this pr.
will make another pr when it's complete.

Issue #1114, opened 2017-09-07
Add support for acm


dlutsch:
Currently moto does not have support for the ACM endpoints. This is a request to add that functionality.  

http://boto3.readthedocs.io/en/latest/reference/services/acm.html

terrycain:
What functions do you need right away?

terrycain:
I believe this can be closed @JackDanger

JackDanger:
> I believe this can be closed @JackDanger

Thanks to you, @terrycain

PR #1172, opened 2017-09-21
Initialization script for service/operation like scaffold in Ruby on Rails

toshitanian:
By executing `setup_new_function.py` and selecting service and operation,  you'll be able to add auto-generated code and response template to appropriate file.

```
$ python setup_new_function.py
Select service: elbv2
==Current Implementation Status==
[ ] add_tags
[X] create_listener
[X] create_load_balancer
[X] create_rule
[X] create_target_group
[X] delete_listener
[X] delete_load_balancer
[X] delete_rule
[X] delete_target_group
[X] deregister_targets
[ ] describe_account_limits
[X] describe_listeners
[ ] describe_load_balancer_attributes
[X] describe_load_balancers
[X] describe_rules
[ ] describe_ssl_policies
[ ] describe_tags
[ ] describe_target_group_attributes
[X] describe_target_groups
[X] describe_target_health
[ ] modify_listener
[ ] modify_load_balancer_attributes
[X] modify_rule
[ ] modify_target_group
[ ] modify_target_group_attributes
[X] register_targets
[ ] remove_tags
[ ] set_ip_address_type
[X] set_rule_priorities
[ ] set_security_groups
[ ] set_subnets
=================================
Select Operation: set_subnets
        Initializing service    elbv2
        skip creating   moto/elbv2
        skip creating   moto/elbv2/__init__.py
        skip creating   moto/elbv2/exceptions.py
        skip creating   moto/elbv2/models.py
        skip creating   moto/elbv2/responses.py
        skip creating   tests/test_elbv2
        skip creating   tests/test_elbv2/test_server.py
        skip creating   tests/test_elbv2/test_elbv2.py
        inserting code  moto/elbv2/responses.py
        inserting code  moto/elbv2/models.py
```
The inserted codes are...
```
diff --git a/moto/elbv2/models.py b/moto/elbv2/models.py
index 3c6afe7..08afedb 100644
--- a/moto/elbv2/models.py
+++ b/moto/elbv2/models.py
@@ -539,7 +539,10 @@ class ELBv2Backend(BaseBackend):
             modified_rules.append(given_rule)
         return modified_rules

+    def set_subnets(self, load_balancer_arn, subnets):
+        # implement here
+        return availability_zones

 elbv2_backends = {}
 for region in ec2_backends.keys():
-    elbv2_backends[region] = ELBv2Backend(region)
+    elbv2_backends[region] = ELBv2Backend(region)
\ No newline at end of file
[...]

PR #1174, opened 2017-09-21
ACM Implementation

terrycain:
Started this near empty PR so no one else starts working on the same thing whilst I do :wink:
Should help out #1114 

## Implementation:
- [x] AddTagsToCertificate
- [x] DeleteCertificate
- [x] DescribeCertificate
- [x] GetCertificate
- [x] ImportCertificate
- [x] ListCertificates
- [x] ListTagsForCertificate
- [x] RemoveTagsFromCertificate
- [x] RequestCertificate
- [x] ResendValidationEmail

## Tests:
All tests done.


JackDanger:
Cool! Check out #1172 if you want to auto-generate the scaffolding

terrycain:
Ahh too late, I saw that just after I created some. Wasnt too bad as there isn't too much in this one.

terrycain:
There any issue me adding https://github.com/pyca/cryptography to dependencies?

JackDanger:
> There any issue me adding https://github.com/pyca/cryptography to dependencies?

As long as we use APIs within Cryptography that are stable enough that users can use a wide range of versions I think that should be fine. When I implemented KSM I just had the `encrypt` method base64-encode the data so if there's a way to do this without generating certs during a user's test run that might work. I'll leave it to you to figure out since you're doing the actual work :)

terrycain:
I've pregenerated some certs, but its more the acm does lots of cert checking, being a certificate manager and all :). Yeah I think cryptography should be fine. Worst case if someone needs a lower version we can always reassess the situation.

JackDanger:
That sounds like the right tradeoff. I'm glad you're on the case!

terrycain:
Getting there, should be done by tomorrow :smile:

terrycain:
Ok, theres 2 tests I've commented out as rely on some extra functionality I PR'd to @spulec's freezegun repo which allows you to shift the time to different timestamps whilst in a decorated test function.

Test1: The first test validated the auto-validate logic I added. As like with SNS, ACM would send you an email to validate a certificate, this way it will just validate itself after 1 minute.

Test2: This tested the IdempotencyToken which allows you to request the same certificate with that token for an hour, after that it will make another certificate. The time jumping allows testing that too... without waiting an hour.

Rest is pretty standard.

terrycain:
FYI Coveralls is broken, you cant view the source with the annotated coverage until someone re-oauth's to coveralls so it can access the repo.

tests/test_acm/test_acm.py
```diff
@@ -0,0 +1,357 @@
...
+    client = boto3.client('acm', region_name='eu-central-1')
+
+    resp = client.request_certificate(
+        DomainName='google.com',
+        SubjectAlternativeNames=['google.com', 'www.google.com', 'mail.google.com'],
+    )
+    resp.should.contain('CertificateArn')
+
+
+# # Also tests the SAN code
```
JackDanger: I normally don't like commented-out code but this seems like a great trade-off to me. Any future developer can take it from here if they notice the FreezeGun PR is merged

JackDanger:
@terrycain This is great. I'm very happy to merge this and we'll deal with any fixes in future PRs. It's so good we can get this into the hands of developers now

tests/test_acm/test_acm.py
terrycain: Yeah agreed, but then this would be worth it. If that code gets merged then it would allow cutting time off many tests.

terrycain:
Yeah my thoughts exactly, I'd rather a half baked implementation out there so boto doesnt scream pages of stacktrace at you, and if the logic is off a bit thats what the community are for :)

terrycain:
No problem, happy to help

tests/test_acm/test_acm.py
nom3ad: FreezeGun PR has been merged already.    Bringing back these tests by  https://github.com/spulec/moto/pull/3372

PR #1182, opened 2017-09-22
Fixed bug in S3 put_bucket_tagging. Closes #1181.

### LINKED ISSUE #1181: S3 PutBucketTagging moto bug if only 1 tag is supplied
As it turns out, if you only supply 1 tag in the `s3.put_bucket_tagging()` operation, the `_bucket_tagging_from_xml()` in S3's`responses.py` breaks because the data type is a `dict` and not a `list` apparently.

I'm going to submit a PR to fix very soon.

mikegrima:
Fixes bug outlined in #1181 .


JackDanger:
@mikegrima nice fix! Will merge when green

JackDanger:

Issue #1006, opened 2017-07-12
Add support for AWS X-Ray service


hwine:
The [AWS X-Ray](https://aws.amazon.com/xray/) can be very helpful for debugging on AWS. This is especially true of [Lambda functions](http://docs.aws.amazon.com/xray/latest/devguide/xray-services-lambda.html)

I was hoping to use moto with a lambda function written in Python 2.7. The [fleece project](https://github.com/racker/fleece) provide Python support for X-Ray, but the two projects do not work well together at the moment.

terrycain:
Have started on X-Ray support. Do you have anything that generates decent X-Ray data as I could do with some "real world" data to mock it out right

hwine:
I don't -- for my use case, I'd just be happy if the calls were intercepted by moto so boto doesn't try to connect to AWS during testing :)

terrycain:
Ahh no problem, should be done soon then :smile:

terrycain:
Ahh just thinking about this. Does your app directly make calls to the Xray service using boto3? As the AWS xray client by default just throws UDP packets to a local daemon which then sends batches to the AWS service. If the latter then you'll need to point the aws xray daemon to a moto server.

hwine:
No, I'm using [racker/fleece](https://github.com/racker/fleece) (see [docs](https://github.com/racker/fleece#x-ray-integration) and some [additional notes about X-Ray](https://github.com/racker/fleece/pull/45#issuecomment-317906853)).

> then you'll need to point the aws xray daemon to a moto server.

n00b question - does that mean I would need to change to using ``moto`` in standalone mode?

terrycain:
Yeah it sounds like fleece will monkeypatch the aws-xray sdk into the botocore library (which will then be able to trace anything your app then does). But the way xray works is a daemon runs on a port, it gets loads of udp segments from your lambda then the daemon uploads bulk segments into xray.

Moto doesnt mock the dameon that xray would post to, only what the daemon would then bulk upload to.
So I guess you would need to set `AWS_XRAY_DAEMON_ADDRESS` which would point the lambda to a custom xray daemon, which would then need to be pointed to a moto server.

https://github.com/racker/fleece/blob/master/fleece/xray.py#L97-L108

https://github.com/spulec/moto#stand-alone-server-mode

terrycain:
@JackDanger this can be closed after merge of #1255

JackDanger:
Closing as fixed. Thanks @terrycain

PR #1183, opened 2017-09-22
AWS X-Ray

terrycain:
Started work on X-Ray cuz it looked cool and #1006 wanted it. Now I regret it, its like one of AWS's secrets giving out little documentation. 

Will probably need actual xray data from the community to get the data structures and logic locked down but I have make a good crack at it which will be a decent starting point. Also helps AWS are using a different API structure yet again :smiley:.

Want to implement the last 3 endpoints, to return correct but empty data, least that way boto wont blow up.

TODO:
- [x] BatchGetTraces
- [x] GetServiceGraph
- [x] GetTraceGraph
- [x] GetTraceSummaries
- [x] PutTelemetryRecords
- [x] PutTraceSegments

### DIFF
--- a/moto/xray/responses.py
+++ b/moto/xray/responses.py
@@ -1,0 +1,133 @@
+from __future__ import unicode_literals
+from urllib.parse import urlsplit
+import json
+import six
+import datetime
+
+from moto.core.responses import BaseResponse
+from moto.core.utils import camelcase_to_underscores, method_names_from_class
+from werkzeug.exceptions import HTTPException
+
+from .models import xray_backends
+from .exceptions import AWSError, BadSegmentException
+
+
+class XRayResponse(BaseResponse):
+
+    def _error(self, code, message):
+        return json.dumps({'__type': code, 'message': message}), dict(status=400)
+
+    @property
+    def xray_backend(self):
+        return xray_backends[self.region]
+
+    @property
+    def request_params(self):
+        try:
+            return json.loads(self.body)
+        except ValueError:
+            return {}
+
+    def _get_param(self, param, default=None):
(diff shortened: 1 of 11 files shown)


terrycain:
All the methods are mocked. GetServiceGraph, GetTraceGraph just return acceptable json so boto doesnt go mental.

Finished for now :smile:

moto/xray/responses.py
```diff
@@ -0,0 +1,150 @@
...
+    def request_params(self):
+        try:
+            return json.loads(self.body)
+        except ValueError:
+            return {}
+
+    def _get_param(self, param, default=None):
+        return self.request_params.get(param, default)
+
+    def _get_action(self):
```
JackDanger: Nice work pulling out `_get_action`

JackDanger:
This is a fantastic start to XRay, @terrycain - you're on a damned roll

moto/xray/responses.py
terrycain: Yeah I had just reimplemented the call function but then it was bugging me any time I looked at that class :-)

terrycain:
Yeah, looking at X-ray and some of the microservice stuff I do at work, I reckon it'll be useful soon so we should see a decent implementation in the not so distant future.

Can we add a UDP mode to moto? or branch out to another repo to either mock the aws xray daemon or spin up the aws xray daemon in test mode, as thats the real thing that needs mocking

JackDanger:
@terrycain That's an interesting idea. I don't know why Moto server can't bind to a UDP socket. If AWS does it it seems reasonable to have Moto server do it.

(this PR is now released as Moto 1.1.14)

Issue #1044, opened 2017-08-10
route53 mock uses incorrect format for HostedZoneId


pll:
Using Moto 1.0.1

```
(Pdb) print json.dumps(zones, indent=4)
{
    "HostedZones": [
        {
            "ResourceRecordSetCount": 0,
            "Config": {
                "PrivateZone": false
            },
            "Id": "/hostedzone/478c0b5c",
            "Name": "subdomain.toplevel.domain."
        }
    ],
```

AWS uses this form: 
```
"Id": "/hostedzone/Z1RWWTK7Y8UDDQ"
```

Not a huge deal. I'm guessing they've changed since the route53 code for moto was written.

terrycain:
^^ Should fix that

terrycain:
@JackDanger this can be closed

pll:
Thanks!

PR #1184, opened 2017-09-22
Updated R53 ID's to match what AWS do now

terrycain:
#1044 - I thought it too would confuse people for a second if it doesn't look like what they're used to.


JackDanger:
This is a really great idea

JackDanger:

PR #1186, opened 2017-09-23
ssm get_parameter to honot WithDecryption

RonRothman:
This PR fixes a bug in release 1.1.13, namely that `ssm.get_parameter()` does not honor its WithDecryption parameter.

(Note that `tox` is reporting some errors elsewhere in the codebase--not in the change I just made--which seem to be environment related. I'm sorry that I haven't been able to get a clean tox build before submitting. Will continue to investigate that separately.)

### DIFF
--- a/moto/ssm/responses.py
+++ b/moto/ssm/responses.py
@@ -56,13 +56,8 @@ class SimpleSystemManagerResponse(BaseResponse):
             return json.dumps(error), dict(status=400)
 
         response = {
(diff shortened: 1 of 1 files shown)


terrycain:
Ahh my bad, good spot :smile:

JackDanger:
@RonRothman Thanks for this! Great catch. Welcome to Moto!

JackDanger:

PR #1188, opened 2017-09-25
Layer per stack

### LINKED ISSUE #1171: Moto opsworks fail to create a layer with the same name under different stack
Hi, 

I am trying to create a couple of stacks with some test layers for each stack but I am getting the exception **There is already a layer named "test_layer_1" for this stack** as soon as I create a layer with the same name under different stack. The code I am using is in /1/. It seems that in https://github.com/spulec/moto/blob/5fc87f8b2bcbce6c7f7e52d58c36ad3e7f4ead03/moto/opsworks/models.py#L425 and in https://github.com/spulec/moto/blob/5fc87f8b2bcbce6c7f7e52d58c36ad3e7f4ead03/moto/opsworks/models.py#L429 the check for layer name is done on `self.layers` dict but I think it should be done on `self.stacks[stackid].layers` dict to allow different layers in different stacks.

Thanks!
/1/
```
self.test_layers = {}
for i in range(1, 4):
    self.test_stacks.append(
        self.ops_client.create_stack(
            Name="test_stack_{0}".format(i),
            Region="us-east-1",
            ServiceRoleArn="service_arn",
            DefaultInstanceProfileArn="profile_arn"
        )
    )
    _stack_id = self.test_stacks[i-1]['StackId']
    self.test_layers[_stack_id] = {}
    for l in range(1, 3):
        self.test_layers[_stack_id].update(
            self.ops_client.create_layer(
                StackId=_stack_id,
                Type="custom",
                Name="test_layer_{0}".format(l),
                Shortname="test_layer_short_{0}".format(l),
            )
```

lleontop:
Closes #1171.

### DIFF
--- a/moto/opsworks/models.py
+++ b/moto/opsworks/models.py
@@ -422,11 +422,11 @@ class OpsWorksBackend(BaseBackend):
         stackid = kwargs['stack_id']
         if stackid not in self.stacks:
             raise ResourceNotFoundException(stackid)
-        if name in [l.name for l in self.layers.values()]:
+        if name in [l.name for l in self.stacks[stackid].layers]:
             raise ValidationException(
                 'There is already a layer named "{0}" '
                 'for this stack'.format(name))
-        if shortname in [l.shortname for l in self.layers.values()]:
+        if shortname in [l.shortname for l in self.stacks[stackid].layers]:
             raise ValidationException(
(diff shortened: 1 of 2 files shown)


JackDanger:
Great work, @lleontop! This will help thousands of people. Welcome to Moto!

JackDanger:

PR #1190, opened 2017-09-25
accepting stdin via make task

JackDanger:


JackDanger:
@toshitanian this is just a minor change I'm making to your work

PR #1193, opened 2017-09-25
Fix parsing non V4 Authorization headers

### LINKED ISSUE #1192: Question - Authorization header parsing
Am getting an IndexError here - https://github.com/spulec/moto/blob/master/moto/core/responses.py#L171-L172

After debugging im sending `AWS xxx:PDcSm14U3qk7EHxFB01rchMDsME=` which lines up with this - https://docs.aws.amazon.com/AmazonS3/latest/dev/RESTAuthentication.html

In this doc - https://docs.aws.amazon.com/AmazonS3/latest/API/sigv4-auth-using-authorization-header.html - we should only check for that region part if using a V4 signature.

terrycain:
Fixed #1192

Dont have a testcase for this one.
More info in the issue, but the gist of it is, the code relies on that the header is in the V4 format which contains a region, the other format just contains accesskey:secretkey which then results in `IndexError`s. Simple check for if were using V4 headers.

### DIFF
--- a/moto/core/responses.py
+++ b/moto/core/responses.py
@@ -167,7 +167,7 @@ class BaseResponse(_TemplateEnvironmentMixin):
         match = re.search(self.region_regex, full_url)
         if match:
             region = match.group(1)
-        elif 'Authorization' in request.headers:
+        elif 'Authorization' in request.headers and 'AWS4' in request.headers['Authorization']:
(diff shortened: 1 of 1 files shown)


JackDanger:
Seems legit.

JackDanger:

PR #1195, opened 2017-09-26
AWS Polly implementation

terrycain:
Have implemented all endpoints on Polly. Tested them but need to expand some more on SynthesizeSpeech as it has some complex parameter validation.

It parses basic PLS lexicon XML but it might choke on some proper real world examples, but until we see those we'll never know.

The SynthesizeSpeech endpoint returns binary data as is what AWS does but its just all `\x00`'s, until someone complains about the format it will do. I was thinking we could just hexlify the first n bytes of a audio clip in mp3, ogg_vorbis and pcm for use with this, which would be really neat but until its needed we might as well hold off.


terrycain:
Oh and also fixed template typo in the script that generates functions for you. Will probably PR some additions to that later, its awesome :smile:

JackDanger:
> Oh and also fixed template typo in the script that generates functions for you. Will probably PR some additions to that later, its awesome 😄

YES! I'm glad it's getting some use. Like so much of what we do it was a fantastic first pass but lacks tests and needs a little love from the community.

@toshitanian ^ Your scaffold script is helping out!

terrycain:
@toshitanian Yeah your script is great man ;-)

terrycain:
Wow flask is really picky about regex in urls :disappointed: 
If this passes tests, then its good to go

tests/test_polly/test_server.py
```diff
@@ -0,0 +1,16 @@
...
+
+'''
+Test the different server responses
+'''
+
+@mock_polly
+def test_polly_list():
+    backend = server.create_backend_app("polly")
+    test_client = backend.test_client()
+    # do test
```
JackDanger: If we add the following will the tests pass?:
```
    res = test_client.get('/')
    res.status_code.should.equal(200)
```

terrycain: Ahh I completely forgot about that file. Will add that later today.

terrycain:
Should all be good :smile:

moto/backends.py
```diff
@@ -54,9 +55,10 @@
     'iam': iam_backends,
     'moto_api': moto_api_backends,
     'instance_metadata': instance_metadata_backends,
-    'opsworks': opsworks_backends,
     'kinesis': kinesis_backends,
     'kms': kms_backends,
+    'opsworks': opsworks_backends,
```
JackDanger: And it's alphabetized 😻 !

JackDanger:
This is so satisfying. Great work, @terrycain !

JackDanger:

PR #1196, opened 2017-09-26
Added more output to scaffold.py

terrycain:
Just added line at the bottom that reminds others what files also need to be modified, pretty sure I got everything.

### DIFF
--- a/scripts/scaffold.py
+++ b/scripts/scaffold.py
@@ -383,5 +383,7 @@ def main():
     else:
         print_progress('skip inserting code', 'api protocol "{}" is not supported'.format(api_protocol), 'yellow')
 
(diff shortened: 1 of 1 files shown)


toshitanian:
@terrycain 
is adding `urls.py` and  adding the backend into "backends.py is necessary with every services?
if it is the case, i'll fix the scaffold script.

terrycain:
@toshitanian I think so, but with some effort we could probably remove most urls.py if we can programatically define the url routes in the Response object/s, and going on from that could probably autodiscover each API. But for now it would be useful I think.

toshitanian:
@terrycain I think I can find the url from botocore. will make a another PR.
You can merge this PR :)

JackDanger:

PR #1197, opened 2017-09-26
AWS Batch

terrycain:
Just a placeholder till I actually start work later. If anyone wants to take over, just comment.

TODO:
- [x] CreateComputeEnvironment
- [x] DeleteComputeEnvironment
- [x] UpdateComputeEnvironment
- [x] DescribeComputeEnvironments
- [x] CreateJobQueue
- [x] DeleteJobQueue
- [x] UpdateJobQueue
- [x] DescribeJobQueues
- [x] RegisterJobDefinition
- [x] DeregisterJobDefinition
- [x] DescribeJobDefinitions
- [x] SubmitJob
- [x] CancelJob
- [x] TerminateJob
- [x] DescribeJobs
- [x] ListJobs

CloudFormation Todo:
- [x] ComputeEnvironment
- [x] JobQueue
- [x] JobDefinition


terrycain:
I think were going to need some sort of validation class thingy, will have a think. 
Also @JackDanger am I missing something or do we not have a dict of instance spec's? As in theory CreateComputeEnvironment would spin up enough instances to meet minvcpus and for that I'd need to know what instances give what.

JackDanger:
> do we not have a dict of instance spec's? 

We do not. That'd be handy

terrycain:
Cool, will look around aws to see if there is any apis for that otherwise will just create a module with some dict

terrycain:
This is becoming a real meaty service. it makes ec2 instances, ecs clusters, goes and pokes around with security groups and ami roles. If we carry on soon we'll have MotoWebServices :wink:

terrycain:
Ooh weird, 3/4 runs pass, python2 in test_mode seems to fail

terrycain:
@JackDanger Here - https://github.com/spulec/moto/blob/master/tests/test_elb/test_elb.py#L935 we're getting some weirdness going on. Look at the test results, apparently at that line, it runs describe loadbalancers, but moto server gets asked for a v2 response, boto tries to decode a v1 response and returns nothing. Any ideas, I dont see anything I could have modified to cause that, unless a boto update has done something odd.

terrycain:
YES FINALLY, damn that was an annoying bug. Ok for some reason... randomly... sometimes flask in server mode only in python2 was putting the ELBv2 route before ELBv1 which has some version date routing thing, by including that routing function in the ELBv2 url routes, it now magically works again. Why it doesnt happen in master I dont know... but it works now

terrycain:
Added simple SubmitJob implementation, it runs docker in a thread, is a daemon thread so will die upon moto exit. Threading has some simple cleanup which should suffice for now. It runs docker containers, and actually streams output to cloudwatch.

Had to modify some cloudwatch logs as I'm sure they were not tested but we'll see what travis thinks.

JackDanger:
@terrycain Thanks for tackling the ELBv1/v2 thing. I pulled my hair out implementing that because it's the one major api change that AWS decided _not_ to change the path or subdomain for. So we have to inspect the XML payload of the request :/

This PR is still marked "WIP" but you have passing tests, is it ready to go?

moto/batch/responses.py
```diff
@@ -0,0 +1,265 @@
...
+        if val is not None:
+            return val
+        return if_none
+
+    def _get_action(self):
+        # Return element after the /v1/*
+        return urlsplit(self.uri).path.lstrip('/').split('/')[1]
+
+    # CreateComputeEnvironment
+    def createcomputeenvironment(self):
```
JackDanger: I am simply amazed that the AWS Batch team decided to allow the totally deviant naming scheme for their endpoints. CamelCase, snake_case, sure. But `POST /v1/createcomputeenvironment HTTP/1.1` ? Very confusing.

Thanks for the comments to help future developers parse these method names.

terrycain: Yeah its most annoying. But the benefit they get at least is entire teams can be isolated from one another and make apis without worrying how others are doing it... as long as they provide the spec for the client libraries everyone uses. But consistency would be nice, and then again they should just purge teh xml apis too.

terrycain:
> @terrycain Thanks for tackling the ELBv1/v2 thing. I pulled my hair out implementing that because it's the one major api change that AWS decided not to change the path or subdomain for. So we have to inspect the XML payload of the request :/
> 
> This PR is still marked "WIP" but you have passing tests, is it ready to go?

Oh god that ELB thing was making me mad, I was stumped for a day or two as the failing tests would sometime just pass for me depending on the order flask loaded the routes.

Yeah I thought I'd fix some tests, but its still wip, I'd rather finish the remaining methods as they should be easy ones, fix up the submitjob so it doesnt just use a hardcoded container spec then it should be good to go.

terrycain:
Ok the only things left on this, is actually respecting what container people want to run and not hard code a  "echo hello" container, and I want to look at how much hassle cloudformation support would be

JackDanger:
@terrycain This is my stopping by just to tell you that you've done a hell of a lot of work and this PR is super close :)

JackDanger:
Just holler when it's ready and you want a final review pass

terrycain:
Yeah, just waiting for the motivation to tackle the final part, will probably be quite easy once I sit down and do it :smile:

terrycain:
@JackDanger Have decided to leave the echo hello world container in there hardcoded for now, until someone raises an issue about it as will need to look into moto'ing whats run in the container otherwise it'll want to speak to aws.

That was the last part but I thought I'd add cloudformation support as that looks pretty simple, should be done by the end of the weekend.

terrycain:
@JackDanger Review please, find all my spelling errors :wink:

moto/batch/responses.py
```diff
@@ -0,0 +1,304 @@
...
+
+        try:
+            self.batch_backend.terminate_job(job_id, reason)
+        except AWSError as err:
+            return err.response()
+
+        return ''
+
+    # CancelJob
+    def canceljob(self):  # Theres some AWS semantics on the differences but for us they're identical ;-)
```
JackDanger: Considering this comment is it worth just having `cancelJob` call `terminateJob`?

moto/logs/models.py
```diff
@@ -120,17 +134,17 @@ def delete_log_stream(self, log_stream_name):
         del self.streams[log_stream_name]
 
     def describe_log_streams(self, descending, limit, log_group_name, log_stream_name_prefix, next_token, order_by):
-        log_streams = [stream.to_describe_dict() for name, stream in self.streams.items() if name.startswith(log_stream_name_prefix)]
+        log_streams = [(name, stream.to_describe_dict()) for name, stream in self.streams.items() if name.startswith(log_stream_name_prefix)]
 
-        def sorter(stream):
-            return stream.name if order_by == 'logStreamName' else stream.lastEventTimestamp
+        def sorter(item):
+            return item[0] if order_by == 'logStreamName' else item[1]['lastEventTimestamp']
```
JackDanger: great catch

moto/batch/responses.py
terrycain: Yeah sounds good.

moto/logs/models.py
terrycain: Stumped me for a bit, I was like surely it cant be broken

terrycain:
Other than the above, is there anything else?

tests/test_batch/test_server.py
```diff
@@ -0,0 +1,16 @@
...
+
+'''
+Test the different server responses
+'''
+
+@mock_batch
+def test_batch_list():
+    backend = server.create_backend_app("batch")
+    test_client = backend.test_client()
+    # do test
```
JackDanger: @terrycain I think this is the only thing missing from this whole amazing PR. Let's make one call to the server to test the routing and ship shippity :shipit:

JackDanger:
(apologies for the slow review - I got distracted for a bit)

terrycain:
> (apologies for the slow review - I got distracted for a bit)

No problem at all, is more me just being impatient :wink:, if the following test passes, then we should be good to go :-)

moto/batch/responses.py
```diff
@@ -0,0 +1,296 @@
...
+                print()
+        return self._json
+
+    def _get_param(self, param_name, if_none=None):
+        val = self.json.get(param_name)
+        if val is not None:
+            return val
+        return if_none
+
+    def _get_action(self):
```
toshitanian: I'm working on the mock for AWS IoT, and batch and iot both have same api-protocol, 'rest-json'. But the uri of batch is not *rest-full* at all...

JackDanger:
Woohoo! Nice work, @terrycain!

spulec:
@terrycain this was awesome!

terrycain:
Yeah was a bit laborious but is nice now we have a pretty decent batch service

PR #1201, opened 2017-09-27
Fix for regression in get_console_output()

nuwang:
Fix for issue: https://github.com/spulec/moto/pull/1152#issuecomment-332487599

### DIFF
--- a/tests/test_ec2/test_general.py
+++ b/tests/test_ec2/test_general.py
@@ -4,10 +4,11 @@ import tests.backport_assert_raises
 from nose.tools import assert_raises
 
 import boto
+import boto3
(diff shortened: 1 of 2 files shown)


tests/test_ec2/test_general.py
```diff
@@ -29,3 +29,14 @@ def test_console_output_without_instance():
     cm.exception.code.should.equal('InvalidInstanceID.NotFound')
     cm.exception.status.should.equal(400)
     cm.exception.request_id.should_not.be.none
+
+
+@mock_ec2
```
JackDanger: Lovely

JackDanger:

PR #1202, opened 2017-09-27
Instance Info

terrycain:
Created a script to grab instance info from http://www.ec2instances.info/
Creates a json file, then the ec2 backend has a constant that contains instance dict.

Realised after I did the scrape that they have a github with the json in there, but atm their json is 1.8Mb and mine is 44k :wink:

Will refactor if it breaks in the future.


scripts/get_instance_info.py
```diff
@@ -0,0 +1,150 @@
...
+    print("Finding data in table")
+    instances = data_table.find('tbody').find_all('tr')
+
+    print("Parsing data")
+    result = {}
+    for instance in instances:
+        instance_id, instance_data = Instance(instance).to_dict()
+        result[instance_id] = instance_data
+
+    root_dir = subprocess.check_output(['git', 'rev-parse', '--show-toplevel']).decode().strip()
```
JackDanger: This is handy - I haven't used git's understanding of root for this kind of thing before. I'm going to steal this from you.

JackDanger:

scripts/get_instance_info.py
terrycain: Ha np, yeah its quite useful as then you dont need to care where about the working directory or script directory is :smile:

PR #1222, opened 2017-10-01
Add more files and lines when scaffolding

toshitanian:
When adding new service:
- add `urls.py`
- add mocks declaration to `moto/__init__.py` and `moto/backends.py`

When adding new operation to a service:
- add uri to `urls.py`


JackDanger:
Great work, @toshitanian !

PR #1228, opened 2017-10-02
Adding owner id to AMI classes

snimark:
Adding owner-id/OwnerId to the AMI classes to allow the value to be specified to test filtering images based on owner.


terrycain:
@snimark are you still working on this or do you want me to have a crack at it

terrycain:
@JackDanger am fixing this, am gonna import some of the free-tier ami info into moto. At some point we want to stop people making instances with an invalid ami, for now i'll ignore that bit but that will probably be a major version change

PR #1232, opened 2017-10-02
add pep naming enforcement

thehesiod:
It appears you need to install plugins to do all the enforcements you want


thehesiod:
actually, per pycharm the first argument to `__new__` should be "mcs", I have a feeling switching to pylint will be better, no need to dumb plugins :)

thehesiod:
@JackDanger I started the work to switch to pylint but don't have enough information for what warnings/errors you want to keep.  This does catch naming conventions and wrong spacing so I think is the best way going forward.  Let me know how I can help!

thehesiod:
One of the big ones is that your global constants in the `__init__.py` files are lower-case.  I added one example where you can batch disable this warning for the file.  I don't think we want to globally disable this because we want to keep the naming checks for class members and the like.

JackDanger:
I agree with you that we should not allow global constants to be lowercase generally. How about we add exceptions to the existing `__init__.py` files and turn the check on for new code?

thehesiod:
sounds good, there are a ton of other warnings that pylint brings up as you can see from the logs, let me know what you think of each type.

bblommers:
With #4760 merged, both our source and tests directories are now covered by `pylint` - in addition to `flake8` for our source dir.
Not all pylint-checks are enabled yet, but as far as I can tell, all changes in this PR are already covered, so I'm going to go ahead and close this one.

Issue #1235, opened 2017-10-03
test_lambda.py:test_invoke_requestresponse_function test failing os x/linux since 1.1.20


dbfr3qs:
Hi there, since moto 1.1.20 the above test has started to fail on my machine (OS X 10.11) and our Ubuntu 16.04 linux build agents, both Python 2.7 and 3.5. Any test that we have which uses invoke_lambda is now erroring out. There does not appear to be a clean version of moto I can pin to because of other changes that were made that were also included in this release that my tests rely on. Would it be possible to get this addressed ASAP?

I believe this is the commit which caused it: 

https://github.com/spulec/moto/commit/9008b852995146b85296174108fc94c5f727bcb5#diff-2df06bdd95e3c78b9335da9f260a3472

Here is the stack trace:

```bash
~/D/W/moto (master↑3|…) $ nosetests tests/test_awslambda/test_lambda.py:test_invoke_requestresponse_function -s
Traceback (most recent call last):
  File "/Users/chris.keogh/Documents/Workspace/moto/moto/awslambda/models.py", line 291, in _invoke_lambda
    with _DockerDataVolumeContext(self) as data_vol:
  File "/Users/chris.keogh/Documents/Workspace/moto/moto/awslambda/models.py", line 94, in __enter__
    for vol in self._lambda_func.docker_client.volumes.list():
  File "/usr/local/lib/python2.7/site-packages/docker-2.5.1-py2.7.egg/docker/models/volumes.py", line 92, in list
    resp = self.client.api.volumes(**kwargs)
  File "/usr/local/lib/python2.7/site-packages/docker-2.5.1-py2.7.egg/docker/utils/decorators.py", line 34, in wrapper
    return f(self, *args, **kwargs)
  File "/usr/local/lib/python2.7/site-packages/docker-2.5.1-py2.7.egg/docker/api/volume.py", line 38, in volumes
    return self._result(self._get(url, params=params), True)
  File "/usr/local/lib/python2.7/site-packages/docker-2.5.1-py2.7.egg/docker/utils/decorators.py", line 46, in inner
    return f(self, *args, **kwargs)
  File "/usr/local/lib/python2.7/site-packages/docker-2.5.1-py2.7.egg/docker/api/client.py", line 189, in _get
    return self.get(url, **self._set_request_timeout(kwargs))
  File "/usr/local/lib/python2.7/site-packages/requests/sessions.py", line 521, in get
    return self.request('GET', url, **kwargs)
[...]

PR #1236, opened 2017-10-03
add lambda note

thehesiod:
as noted in https://github.com/spulec/moto/issues/1235#issuecomment-333965351 we should note that lambda mocking requires docker


JackDanger:

PR #1237, opened 2017-10-03
make lambda support opt-in

thehesiod:


thehesiod:
not sure this works yet

terrycain:
I resolved a conflict for what its worth :smile:

thehesiod:
hopefully it works, I hadn't tried it :)

terrycain:
@thehesiod if the tests pass, is this ready to go yeah?

thehesiod:
theoretically, probably should have a !SUPPORTS_LAMBDA test case to make sure it gets disabled properly.

terrycain:
In theory according to github I should be able to push to your branch. I'll see if I can finsih it off

thehesiod:
just invited you in case

terrycain:
Looking more into it, we'll need a lambda implementation that doesn't require docker that is optionally activated, as other modules will attempt to import the lambda models regardless if docker's installed.
Also AWS batch requires docker, so is there much of a point in making it optional now?

thehesiod:
this is a question for @JackDanger

JackDanger:
Considering AWS Batch requires Docker maybe we just need to go all-in. Let's all keep a lookout for ways we can ship fewer dependencies with Moto but for now lets leave the current dependency graph as is.

JackDanger:
Considering AWS Batch requires Docker maybe we just need to go all-in. Let's all keep a lookout for ways we can ship fewer dependencies with Moto but for now lets leave the current dependency graph as is.

thehesiod:
should we close this?

PR #1239, opened 2017-10-04
Use long format for EC2 instance ID

terrycain:


JackDanger:

PR #1242, opened 2017-10-06
Modify spot fleet request

drmorr0:
Hi, it's me again!  I really want to be able to test modify_spot_fleet_request operations in my code so I went ahead and added this into moto.  I'm using this code right now in my application too, so I have some additional confidence it's correct.  Would love to hear what you think.

All tests pass!


JackDanger:
Nice work! This is really great. Thanks for writing it and for the great tests. Merging!

JackDanger:

PR #1244, opened 2017-10-06
Support wildcard tag filters on SecurityGroups

radhus:
Use the `tag_filter_matches` function from utils, but could perhaps use `generic_filters` somehow. This solves the issue though!


JackDanger:
This looks great, nice work! If you have ideas for how to start replacing these individual filters with `generic_filters` like you said I'm sure the community would really benefit from that.

JackDanger:

PR #1245, opened 2017-10-07
Added overkill FilterExpression to dynamodb scan

terrycain:
Ok I got a bit carried away. DynamoDB Scan now works. 

There is a function get_filter_expression, which takes a filter expression, name substitutions and value substitutions. Removes all whitespace, tokenizes the string, then uses the shunting-yard algorithm to convert the expression into reverse polish notation which removed execution order ambiguity. Then as it steps through the RPN form, it builds the expression in Object form, as there are classes now for AND, OR BETWEEN etc.. operations etc... Its a bit messy but should handle most correct filter expressions.

... all to get FilterExpression to work.

### DIFF
--- a/moto/dynamodb2/comparisons.py
+++ b/moto/dynamodb2/comparisons.py
@@ -1,4 +1,6 @@
 from __future__ import unicode_literals
+import re
+import six
 # TODO add tests for all of these
 
 EQ_FUNCTION = lambda item_value, test_value: item_value == test_value  # flake8: noqa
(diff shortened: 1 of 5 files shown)


JackDanger:
This is pretty cool :) It's hard to read the coverage report on coveralls.io for this branch and my only reservation about merging this is that I want to make sure the logic is covered by your tests. To help with that I've just merged an experimental branch that sends data both to coveralls.io (with a newer client library) as well as codecov.io. Perhaps that'll make it easier to make sure this is super maintainable.

Great work!

terrycain:
Yeah its a bit of a beast. I have a test case that tests the end result of the logic, ie the functions in the filter expression and basic and's and ors. It doesnt test really long expressions but due to the logic if the base elements work, and the first test case works, then pretty much any longer one will work too.

Am probably missing some coverage on the if statements that guard againt rubbish data as I havent thrown any bad data though it. I think what would be best though would just be to add a try/except in the actual scan invocation in the request method and catch a clientError exception which can inform the client that its a bad expression.

Can you link me the codecov report, then I can improve that

JackDanger:
@terrycain Here's the codecov.io result for this branch (I built it off a rebase of this PR so it can compare coverage to latest master but it's effectively identical): https://codecov.io/gh/spulec/moto/pull/1247/

JackDanger:
Looks like the coverage is pretty good! The tests could use a `BETWEEN` statement and some testing of parentheses but it looks good. Take a look and either tell me to merge it or tell me to wait, I'm okay with either.

terrycain:
Thats cool, yeah I agree with that. I also got coverage working now in pycharm so I'll hunt down through some of the code paths and get them tested too + implement decent reporting (not just letting moto blow up and not respond lol)

terrycain:
Good shout on the parenthesis, there was a "small" logical error where they were taking precedence when they shouldn't of.

terrycain:
Of course I hit some real dodgy SSL bug in httpretty. Patched it with some snippet I found in a github issue... seems to have worked.

Added more tests for dynamo, that gives the +0.01% coverage lol

terrycain:
Good to merge :+1:

moto/dynamodb2/comparisons.py
```diff
@@ -39,3 +41,452 @@
...
+            try:
+                token = int(token)
+            except ValueError:
+                try:
+                    token = float(token)
+                except ValueError:
+                    pass
+            tokens2.append(token)
+
+    # Start of the Shunting-Yard algorigth. <-- Proper beast algorithm!
```
JackDanger: Just added this to my reading list. Thanks!

moto/dynamodb2/comparisons.py
```diff
@@ -39,3 +41,452 @@
...
+        return val not in ('<', '>', '=', '>=', '<=', '<>', 'BETWEEN', 'IN', 'AND', 'OR', 'NOT')
+
+    OPS = {'<': 5, '>': 5, '=': 5, '>=': 5, '<=': 5, '<>': 5, 'IN': 8, 'AND': 11, 'OR': 12, 'NOT': 10, 'BETWEEN': 9, '(': 100, ')': 100}
+
+    def shunting_yard(token_list):
+        output = []
+        op_stack = []
+
+        # Basically takes in an infix notation calculation, converts it to a reverse polish notation where there is no
+        # ambiguaty on which order operators are applied.
```
JackDanger: Not important but small typo here with 'ambiguity'

moto/dynamodb2/comparisons.py
```diff
@@ -39,3 +41,452 @@
...
+                    output.append(op_stack.pop())
+                op_stack.append(token)
+        while len(op_stack) > 0:
+            output.append(op_stack.pop())
+
+        return output
+
+    output = shunting_yard(tokens2)
+
+    # Hacky funcition to convert dynamo functions (which are represented as lists) to their Class equivelent
```
JackDanger: typo here in 'function' as well

JackDanger: Oh, and 'equivalent'. The code kicks ass so I feel terrible pointing out this stuff but since you're name is on it I want you to look good :)

terrycain:
nah no problem, me and spelling at 5 in the morning never goes well :wink:, and its a simple change so its fine

moto/packages/httpretty/core.py
```diff
@@ -103,6 +103,12 @@
 except ImportError:  # pragma: no cover
     ssl = None
 
+try:  # pragma: no cover
```
JackDanger: Damn. Nice fix. It's sad that this kind of thing is necessary.

JackDanger:
Great work, @terrycain! I love a feature that's described as 'overkill'. :D

moto/packages/httpretty/core.py
terrycain: Yeah it was a right pain

terrycain:
:smile: hopefully with some refactoring it should be useful for other expression parts of dynamo too

JackDanger:

PR #1255, opened 2017-10-10
AWS X-Ray client mock.

terrycain:
So the most common use of AWS X-Ray will be someone importing the AWS X-Ray SDK and running `patch_all()`, this will then go and infect botocore, add headers to outgoing requests, look for headers on incoming ones and after it gets the data it wants, it will send UDP packets to a "local" daemon. 

This essentially patches the X-Ray SDK so that it doesnt try and send data X-Ray daemon, and eventually it will just feed straight into our X-Ray backend so that the trace info will be there for when its queried.

Have also implemented the headers 'x-amz-crc32' and 'x-amzn-requestid' which you can then decorate requests for it to then pass on those headers. You can also decorate the `call_action` method to add crc's and requestids to all methods. I added the relevant CRCs and request IDs to the endpoints that the X-Ray SDK lists in a json whitelist file. It could quite possibly want most if not all responses to contain a request id.

Also fixed an annoying test case that was checking for dictionary keys using a list which isnt too great as its not guaranteed they'll be in the same order.

@JackDanger thoughts on this so far?

### DIFF
--- a/moto/sqs/responses.py
+++ b/moto/sqs/responses.py
@@ -2,7 +2,7 @@ from __future__ import unicode_literals
 from six.moves.urllib.parse import urlparse
 
 from moto.core.responses import BaseResponse
-from moto.core.utils import camelcase_to_underscores
+from moto.core.utils import camelcase_to_underscores, amz_crc32, amzn_request_id
 from .utils import parse_message_attributes
 from .models import sqs_backends
 from .exceptions import (
@@ -51,6 +51,8 @@ class SQSResponse(BaseResponse):
 
         return visibility_timeout
 
+    @amz_crc32  # crc last as request_id can edit XML
+    @amzn_request_id
     def call_action(self):
         status_code, headers, body = super(SQSResponse, self).call_action()
         if status_code == 404:
@@ -282,7 +284,7 @@ CREATE_QUEUE_RESPONSE = """<CreateQueueResponse>
         <VisibilityTimeout>{{ queue.visibility_timeout }}</VisibilityTimeout>
     </CreateQueueResult>
     <ResponseMetadata>
-        <RequestId>7a62c49f-347e-4fc4-9331-6e8e7a96aa73</RequestId>
+        <RequestId>{requestid}</RequestId>
     </ResponseMetadata>
 </CreateQueueResponse>"""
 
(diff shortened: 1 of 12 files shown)


moto/sqs/responses.py
```diff
@@ -282,7 +284,7 @@ def receive_message(self):
         <VisibilityTimeout>{{ queue.visibility_timeout }}</VisibilityTimeout>
     </CreateQueueResult>
     <ResponseMetadata>
-        <RequestId>7a62c49f-347e-4fc4-9331-6e8e7a96aa73</RequestId>
+        <RequestId>{{ requestid }}</RequestId>
```
JackDanger: Sneaky! I actually love this. Users won't have to pass this in but it's clearly parameterized.

moto/xray/mock_client.py
```diff
@@ -0,0 +1,83 @@
...
+from aws_xray_sdk.core.context import Context as AWSContext
+from aws_xray_sdk.core.emitters.udp_emitter import UDPEmitter
+
+
+class MockEmitter(UDPEmitter):
+    """
+    Replaces the code that sends UDP to local X-Ray daemon
+    """
+    def __init__(self, daemon_address='127.0.0.1:2000'):
+        address = os.getenv('AWS_XRAY_DAEMON_ADDRESS_YEAH_NOT_TODAY_MATE', daemon_address)
```
JackDanger: You're single-handedly increasing the Britishness of Moto and I support this ✅

moto/sqs/responses.py
terrycain: Yeah that one was a test to see if boto would accept it, and it does. Will spam the request id's on all the sqs methods to see if it likes it

JackDanger:
I like the direction this is going. I can't think of any simpler way to implement X-Ray support. Thanks for your great work, @terrycain !

terrycain:
Yeah, its a weird one X-Ray, I can see why they went the local daemon sort of approach as I can see many containers all spamming tons of segments and least that way they can collect them and upload in bulk. Is a shame they dont offer an `unpatch` method, will ask our account manager to see if he'll get me in touch with the xray devs.

terrycain:
@JackDanger I'd consider this done for now. It could do with putting the xray segments into the xray backend. Until someone needs it I don't think its worth the effort, as most people are just going to use it to keep the xray sdk happy so they can test locally.

JackDanger:

PR #1257, opened 2017-10-11
Remove newlines from create-access-key response

bpandola:
Newlines in the iam.create_access_key response template was causing the secret key to be returned as `u'\n    vr7PlfZyCVt3QHT2fpiqeKyL8lJlzYLC  \n'` instead of `u'vr7PlfZyCVt3QHT2fpiqeKyL8lJlzYLC'`.  This likely has gone unnoticed because ultimately the secret is just a string, but we are concatenating it with other data to form a URI and the newlines are obviously problematic in that scenario.  So...merging this shouldn't break anything and it will help us.  Thanks!


JackDanger:
Thanks for finding this!

JackDanger:

PR #1259, opened 2017-10-12
Fix runningTasksCount ECS container instance attribute

tomelliff:
ECS container instances have attributes of 'runningTasksCount' and 'pendingTasksCount'.
See Boto3 docs here: http://boto3.readthedocs.io/en/latest/reference/services/ecs.html#ECS.Client.describe_container_instances
REST API docs here: http://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_DescribeContainerInstances.html#API_DescribeContainerInstances_ResponseSyntax

### DIFF
--- a/moto/ecs/models.py
+++ b/moto/ecs/models.py
@@ -289,7 +289,7 @@ class ContainerInstance(BaseObject):
              'type': 'STRINGSET'}]
         self.container_instance_arn = "arn:aws:ecs:us-east-1:012345678910:container-instance/{0}".format(
             str(uuid.uuid1()))
-        self.pending_task_count = 0
+        self.pending_tasks_count = 0
         self.remaining_resources = [
             {'doubleValue': 0.0,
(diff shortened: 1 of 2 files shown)


tomelliff:
Before this change I'd expect the `describe_container_instances` call to ECS to return a container instance description with `runningTaskCount` and `pendingTaskCount` but if I print the description those attributes are completely missing. Fixing them to their proper names adds them to the returned schema.

Having not really looked at the moto codebase before is there a reason why that's the case?

tomelliff:
@JackDanger Does this need anything more from me before it can be merged? Would like to uncomment a whole block of mocked ECS tests from something I've written but they fail without these changes.

JackDanger:
@tomelliff My apologies for the delay. This fix looks great - thanks for putting it together!

tomelliff:
Cool thanks :)

Issue #1258, opened 2017-10-12
Instance.modify_attribute() not working for blockDeviceMapping


domingosnovo:
## Reporting Bugs

I am trying to set the `DeleteOnTermination` attribute of a volume attached to an instance to `True` using the boto3 method `EC2.Instance.modify_attribute()`, but it doesn't seem to be doing the trick.

I tried using the procedure described in [this link](https://github.com/spulec/moto/blob/a7049acf70afb2261c6a92c193ada51458e91cd5/moto/ec2/responses/instances.py#L147). I also tried using `EC2.Client.modify_instance_attribute()`, and neither option worked. No errors at all, just that the attribute itself wasn't updated.

I've wrote the following python snippet to reproduce this issue:

```python
from moto import mock_ec2
import boto3
from pprint import pprint


@mock_ec2
def main():
    ec2_client = boto3.resource('ec2', region_name='us-west-1')
    result = ec2_client.create_instances(ImageId='ami-12345678', MinCount=1, MaxCount=1)
    instance = result[0]
    instance.load()
    pprint(instance.block_device_mappings)
    instance.modify_attribute('blockDeviceMapping', {'/dev/sda1': True})
    instance.load()
    pprint(instance.block_device_mappings)


if __name__ == "__main__":
    main()
```

Which when executed, should set the `DeleteOnTermination` attribute of the attached EBS volume to `True`. But this is not the case, as per this bash output:

```bash
root@ip-10-132-17-214:~# ./a.py
[{'DeviceName': '/dev/sda1',
  'Ebs': {'AttachTime': datetime.datetime(2017, 10, 12, 3, 24, 23, tzinfo=tzlocal()),
          'DeleteOnTermination': False,
          'Status': 'in-use',
          'VolumeId': 'vol-8d707b2a'}}]
[{'DeviceName': '/dev/sda1',
  'Ebs': {'AttachTime': datetime.datetime(2017, 10, 12, 3, 24, 23, tzinfo=tzlocal()),
          'DeleteOnTermination': False,
          'Status': 'in-use',
          'VolumeId': 'vol-8d707b2a'}}]
root@ip-10-132-17-214:~#
```

I'm using Moto 1.1.22 and Boto3 1.4.7, which I installed using `pip install <packagename>` on a freshly launched Ubuntu 16.04.03 instance running on AWS.

terrycain:
Have pretty much solved this, technically moto was setting terminate to true but the response it was giving to boto was being interpreted as false.

[...]

PR #1260, opened 2017-10-12
Fix for #1258

terrycain:
Basically, we put `True` into the XML, not `true`, Boto obviously does some `'True=='true'` rubbish somewhere.

The only way I can see us doing this change nicer, is either by starting to use proper XML with etree, or by adding jinja2 to the deps and then you can do things like "{{ deviceobject.delete_on_termination | lower }}".

I'd go and edit the object to add a property that does the string conversion but its a boto object not one we've made. We could subclass it, but we'd need to find all the places that use it, I think the jinja2 approach would be the least amount hassle.

@JackDanger opinions on adding jinja2 to deps?


JackDanger:
I'm hesitant to add jinja2 to the deps (certainly until we clear the 'docker' dependency) but that does seem like the right call long term. If we run into an issue like this again lets consider that a sign that we should solve the underlying problem. For now this is a good fix 👍

terrycain:
Yep sounds good. And yeah its a bit of a heavy dep. Once the tests pass, the PRs ready to be merged :smile:

JackDanger:

PR #1264, opened 2017-10-15
Add attach_ and detach_instances methods to autoscaling service

captainkerk:
Hello! Long time fan, first time contributor. I wanted the detach_instances functionality for a project of mine, so I decided to try my hand at implementing it. 

Edit: I also added the `attach_instances` method, as requested in this issue: https://github.com/spulec/moto/issues/893

PR #1265, opened 2017-10-16
Add autoscaling load balancer methods

captainkerk:
This adds three methods to the autoscaling service: attach_load_balancers, describe_load_balancers, and detach_load_balancers; all of which manipulate "classic" Elastic load balancers and their association to an autoscaling group.

### DIFF
--- a/moto/autoscaling/models.py
+++ b/moto/autoscaling/models.py
@@ -539,6 +539,24 @@ class AutoScalingBackend(BaseBackend):
 
             group.tags = new_tags
 
+    def attach_load_balancers(self, group_name, load_balancer_names):
+        group = self.autoscaling_groups[group_name]
+        group.load_balancers.extend(load_balancer_names)
+        self.update_attached_elbs(group_name)
+
+    def describe_load_balancers(self, group_name):
+        return self.autoscaling_groups[group_name].load_balancers
+
+    def detach_load_balancers(self, group_name, load_balancer_names):
+        group = self.autoscaling_groups[group_name]
+        group_instance_ids = set(
+            state.instance.id for state in group.instance_states)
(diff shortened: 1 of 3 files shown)


moto/autoscaling/models.py
```diff
@@ -496,6 +496,24 @@ def create_or_update_tags(self, tags):
 
             group.tags = new_tags
 
+    def attach_load_balancers(self, group_name, load_balancer_names):
```
terrycain: Once #1255 gets merged, it wouldn't be a bad idea to see if [these](https://github.com/terrycain/moto/blob/19b75770910f2a42158044da9c69ff3783e14a89/moto/awslambda/responses.py#L36-L37) decorators would work with these responses as then you can do things like in [this](https://github.com/terrycain/moto/blob/19b75770910f2a42158044da9c69ff3783e14a89/moto/sqs/responses.py#L287) template which autogenerated a request id (useful for uniqueness when X-ray is being used).

captainkerk: Woah, that's pretty slick. I'm definitely willing to roll that in once it's merged. Looks like that should be soon?

JackDanger: I've just merged #1255 - let's make it happen :)

captainkerk: On it.

captainkerk: Done.

captainkerk: Done. I also added it to the methods I added in #1264

JackDanger:
This is so slick. Merging immediately. @terrycain I'm a big fan of your new decorators! @captainkerk I'm a big fan of you!

JackDanger:

PR #1266, opened 2017-10-17
Fix bug with update_attached_elbs

captainkerk:
So, I also wanted to add the elbv2 methods to autoscaling. I started on that and came across this weird bug where the `create_auto_scaling_group` method was erroneously adding the group's instances to all available ELBs when the group was created with no load balancers specified.

 In the test below, I create an ELB and an ASG. I expected the ELB to have an instance count of 0, but to my amazement I found that it had a count of the autoscaling group's desired size.

I tracked it down to models.py L511. On create, if no load balancers are specified, `group.load_balancers` is an empty list. the `elb_backend.describe_load_balancers` method returns all available elbs if you pass in nothing for it to match on. This is expected, however in this specific case it is the opposite of what we want. So, we should skip if group.load_balancers is empty.

### DIFF
--- a/moto/autoscaling/models.py
+++ b/moto/autoscaling/models.py
@@ -506,20 +506,21 @@ class AutoScalingBackend(BaseBackend):
 
         # skip this if group.load_balancers is empty
         # otherwise elb_backend.describe_load_balancers returns all available load balancers
-        if group.load_balancers:
-            try:
-                elbs = self.elb_backend.describe_load_balancers(
-                    names=group.load_balancers)
-            except LoadBalancerNotFoundError:
-                # ELBs can be deleted before their autoscaling group
-                return
+        if not group.load_balancers:
+            return
+        try:
+            elbs = self.elb_backend.describe_load_balancers(
+                names=group.load_balancers)
+        except LoadBalancerNotFoundError:
+            # ELBs can be deleted before their autoscaling group
+            return
 
-            for elb in elbs:
-                elb_instace_ids = set(elb.instance_ids)
-                self.elb_backend.register_instances(
-                    elb.name, group_instance_ids - elb_instace_ids)
-                self.elb_backend.deregister_instances(
-                    elb.name, elb_instace_ids - group_instance_ids)
+        for elb in elbs:
+            elb_instace_ids = set(elb.instance_ids)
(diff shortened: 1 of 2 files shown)


moto/autoscaling/models.py
```diff
@@ -504,19 +504,22 @@ def update_attached_elbs(self, group_name):
...
-
-        for elb in elbs:
-            elb_instace_ids = set(elb.instance_ids)
-            self.elb_backend.register_instances(
-                elb.name, group_instance_ids - elb_instace_ids)
-            self.elb_backend.deregister_instances(
-                elb.name, elb_instace_ids - group_instance_ids)
+        # skip this if group.load_balancers is empty
+        # otherwise elb_backend.describe_load_balancers returns all available load balancers
+        if group.load_balancers:
```
JackDanger: How about just returning here if group.load_balancers is empty? Something like:

```
if not group.load_balancers:
    return
```
That'll save a level of indentation. If you prefer it the current way I'll defer to you and merge it.

tests/test_autoscaling/test_autoscaling.py
```diff
@@ -520,6 +520,43 @@ def test_describe_load_balancers():
...
     response['LoadBalancers'][0]['LoadBalancerName'].should.equal('my-lb')
 
+@mock_autoscaling
+@mock_elb
+def test_create_elb_and_autoscaling_group_no_relationship():
+    INSTANCE_COUNT = 2
+    ELB_NAME = 'my-elb'
+
+    elb_client = boto3.client('elb', region_name='us-east-1')
+    _ = elb_client.create_load_balancer(
```
JackDanger: Most of the Moto tests don't bother assigning any return value, not even to `_`

JackDanger: (that is, when that value will not be used)

captainkerk: Thats true, I suppose I copied/pasted that from somewhere?

captainkerk:
@JackDanger All set. Thanks very much!

JackDanger:
It looks great! You're on a roll :)

JackDanger:

PR #1267, opened 2017-10-17
Correct the type of a default attribute in SNS PlatformEndpoint

nanorepublica:
The `Enabled` Attribute in the PlatformEndpoint of SNS current returns a boolean, however, the 'enabled' property is expecting a string as `.lower()` is called on the result.

This change simply changes the default from `True` to `'True'` so the property works as expected.


JackDanger:
@nanorepublica Thanks for your contribution. This is a good fix and it'll help the whole community. Welcome to Moto!

JackDanger:

pavel-puchkin:
Looks like the string actuall must be lower-cased?

PR #1268, opened 2017-10-17
requiring minimum botocore

JackDanger:
Boto and Boto3 can be a little old but Moto will throw an error if
botocoe doesn't even know about some of the services it supports.
As of this commit Polly is new enough some users are running into
exceptions.


JackDanger:
https://pastebin.com/ESrM6G6G

terrycain:
Good spot, I remember having this issue originally, but then it magically went away ;-)

PR #1270, opened 2017-10-18
Prevent attaching a load balancer multiple times

captainkerk:
Hey there, I overlooked something the other day:

While attaching load balancers, we need to filter the list by determining which are already attached. This will prevent duplicates from getting added. Right now, if you use attach_load_balancer to attach an elb, then call it again, you'll end up with a duplicate.


JackDanger:
Nice catch!

JackDanger:

PR #1274, opened 2017-10-19
Fix timezone mismatch in acm certificate created date and now date check

WhileLoop:
The certificate created time is set in UTC, but the check to see if it has expired is done using local time which causes certificates to be erroneously marked as expired.


JackDanger:
@WhileLoop Thanks for this fix! This is really helpful. Welcome to Moto!

JackDanger:

PR #1275, opened 2017-10-19
Check if subject_alt_names is not none before checking length

WhileLoop:
Calling describe_certificate was returning the following error:

   File "/usr/local/lib/python3.6/site-packages/moto/acm/responses.py", line 188, in request_certificate
    if len(subject_alt_names) > 10:
TypeError: object of type 'NoneType' has no len()

### DIFF
--- a/moto/acm/responses.py
+++ b/moto/acm/responses.py
@@ -185,7 +185,7 @@ class AWSCertificateManagerResponse(BaseResponse):
         idempotency_token = self._get_param('IdempotencyToken')
         subject_alt_names = self._get_param('SubjectAlternativeNames')
 
-        if len(subject_alt_names) > 10:
+        if subject_alt_names is not None and len(subject_alt_names) > 10:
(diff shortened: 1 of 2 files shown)


JackDanger:
Thanks for this @WhileLoop! Would you be willing to try writing a test that exercises this behavior? If we'd had one already we might have avoided this bug to begin with. Whatever your code did to trigger this edge case it would be great to have as logic in the Moto test suite. Somewhere in `tests/test_acm/test_acm.py` is the right place.

JackDanger:
If you get stuck for any reason just leave a comment here and I or somebody else will step in and help!

WhileLoop:
@JackDanger added a simple test case.

JackDanger:
Thanks @WhileLoop! You've just saved a bunch of other people from running into the same frustrating bug that you found. Nice work!

JackDanger:

PR #1281, opened 2017-10-20
Fix Error in get-logs request LogEvent instance is not JSON serializable

jorgemarcial:
When an user request get-log-events commands returns next error:

TypeError: <moto.logs.models.LogEvent instance> is not JSON serializable

Fixed bug with new list of dictionary attributes from LogEvent Object


JackDanger:
@jorgemarcial Great fix! Thanks for this

JackDanger:

PR #1283, opened 2017-10-20
Fixes #1276 SNS Subject validation

### LINKED ISSUE #1276: Support missing for SNS Publish "Subject" optional parameter
The API (and boto3) allow for an optional parameter [``Subject`` to the ``publish``](http://docs.aws.amazon.com/sns/latest/api/API_Publish.html) method on an sns client. There are some constraints on the value of that field, (<100 characters is one) which do not appear to be checked in the current ``moto``.

A simplification of the method I'd like to test with ``moto`` is at the end. If I test using a message of
```python
    msg = "nope " * 21  # > 100 characters
```
It passes with ``moto``, but throws an exception on AWS: ``An error occurred (InvalidParameter) when calling the Publish operation: Invalid parameter:
Subject``

```python
    def send_sns(self, msg):
        subject = msg.split('\n')[0]
        client = boto3.client("sns")
        topic_arn = os.environ.get('SNSARN', "")
        response = client.publish(Message=msg, Subject=subject,
                                  TopicArn=topic_arn)
```

Here are my current library versions for python2.7:
```
Package                      Version     
---------------------------- ----------- 
boto                         2.48.0      
boto3                        1.4.7       
botocore                     1.7.29      
moto                         1.1.22      
```

terrycain:


hwine:
field report -- I locally installed this PR and my moto tests now fail as expected!  \o/

I think this is a record for fastest reported -> fixed in my experience. 🥇

moto/sns/models.py
```diff
@@ -257,7 +257,7 @@ def list_subscriptions(self, topic_arn=None, next_token=None):
             return self._get_values_nexttoken(self.subscriptions, next_token)
 
     def publish(self, arn, message, subject=None):
-        if subject is not None and len(subject) > 100:
+        if subject is not None and len(subject) >= 100:
```
hwine: It would be even-better-if the supplied message was copied into the publish call, so it can round trip. Right now, any message always contains the default ``"my subject"`` from ``moto/sns/models.py``.

Since there may be more optional fields, maybe "pass all" might be a reasonable approach. E.g:
```python
    def publish(self, arn, message, subject=None, **kwargs):
...
        if subject is not None:
            kwargs['subject'] = subject
            if len(subject) >= 100:
...
        message_id = topic.publish(message, **kwargs)
```

hwine: nm -- the current functionality helps a lot -- adding roundtrip support looks more complex than I thought. It should be a separate bug after this lands, I think.

terrycain: ^^ Agreed.

JackDanger:

PR #1284, opened 2017-10-20
Add support for HealthStatus for Auto Scaling Groups

carleton-stripe:
Add support for the [`set-instance-health`](http://docs.aws.amazon.com/cli/latest/reference/autoscaling/set-instance-health.html
) aws autoscaling API.

It only sets the HealthStatus that will be returned through `describe-auto-scaling-groups`, it does not actually terminate the instances.   It also ignores the `ShouldRespectGracePeriod` parameter since that applies to the AutoScaling health check termination behavior (see the [Custom Healthcheck section](http://docs.aws.amazon.com/autoscaling/latest/userguide/healthcheck.html#as-configure-healthcheck) of the docs).  Implementing the actual health checking and automatic replacing would be more involved.

### DIFF
--- a/moto/autoscaling/models.py
+++ b/moto/autoscaling/models.py
@@ -16,9 +16,10 @@ ASG_NAME_TAG = "aws:autoscaling:groupName"
 
 
 class InstanceState(object):
-    def __init__(self, instance, lifecycle_state="InService"):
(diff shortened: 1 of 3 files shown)


JackDanger:
@carleton-stripe Great work!

JackDanger:

carleton-stripe:
Awesome, thanks!

PR #1285, opened 2017-10-21
Add ELBv2 Support  to Autoscaling Service

captainkerk:
This adds elbv2 functionality to the autoscaling service.

The `create_auto_scaling_group` now supports target group arns, and the attach, detach, and describe methods have been implemented. 

A note about the update_auto_scaling_group method:

I noticed that the [mocked method](https://github.com/spulec/moto/blob/master/moto/autoscaling/responses.py#L148)  takes a `LoadBalancerNames` argument. This does not actually conform with the boto3 spec; the way to 'update' an existing autoscaling group with a new load balancer config is to actually use the `attach_load_balancers` method. The same thing goes for the elbv2 resource.

Since this doesn't relate strictly to the elbv2 integration, I was going to open a new PR with the fix. I thought it was worth mentioning because the v1 methods work as a rough guide for the v2 methods.

### DIFF
--- a/moto/autoscaling/models.py
+++ b/moto/autoscaling/models.py
@@ -4,6 +4,7 @@ from moto.compat import OrderedDict
 from moto.core import BaseBackend, BaseModel
 from moto.ec2 import ec2_backends
 from moto.elb import elb_backends
+from moto.elbv2 import elbv2_backends
 from moto.elb.exceptions import LoadBalancerNotFoundError
 from .exceptions import (
     ResourceContentionError,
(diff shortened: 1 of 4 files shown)


JackDanger:
@captainkerk Great fix! Thanks for this. And great catch with the `LoadBalancerNames` param - a PR to fix that would be great.

captainkerk:
Done. Both PRs are ready for a review.

JackDanger:
Great work, @captainkerk !

JackDanger:

PR #1286, opened 2017-10-21
Fix 1282

terrycain:
Route 53 HostedZone should return Zone ID not Zone Name


JackDanger:

PR #1287, opened 2017-10-21
update_auto_scaling_group method incorrectly supports LoadBalancerNames

captainkerk:
the boto3 autoscaling client's `update_auto_scaling_group` method does not support modifying load balancer properties. This PR removes that parameter.


JackDanger:

PR #1289, opened 2017-10-22
DynamoDB FilterExpression NOT logic

terrycain:
Make some general improvements, thanks @batkuip for a non working example that lead to me discovering a logical issue with handling brackets, code now recurses to parse nested brackets properly.

Also after some reading looks like I was overthinking the NOT operator and it just behaves like a normal right associative operator like the `!` in C. So added NOT functionality too.


JackDanger:

PR #1291, opened 2017-10-23
Fix scaffold to support rest-json style API

toshitanian:
I fixed scaffold script and `moto/core/responses.py` to support rest-json style API in scaffold.

# changes in core module
In `moto/core/responses.py`,  I mainly changed `_get_param` method and `_dispatch`+`_get_action` method.

**In `_get_param`**
Response class has been getting parameters only from queryparameter. In the PR, it will try to get parameters in order of queryparameter, json body and path parameter.

**In `_dispatch`+`_get_action`**
previously Response class had been dispatching request to appropriate method  based on the operation name in the request. Unfortunately, rest-api request does not include operation name in itself. In this PR, when it cannot find operation name in the request, it looks up operation name in boto3 client by request method and request uri.

# changes in scaffold
some refactors + fixes to adapt to change in core module for rest-api

### DIFF
--- a/moto/core/responses.py
+++ b/moto/core/responses.py
@@ -17,6 +17,8 @@ from six.moves.urllib.parse import parse_qs, urlparse
 import xmltodict
 from pkg_resources import resource_filename
 from werkzeug.exceptions import HTTPException
+
+import boto3
 from moto.compat import OrderedDict
 from moto.core.utils import camelcase_to_underscores, method_names_from_class
 
@@ -151,7 +153,6 @@ class BaseResponse(_TemplateEnvironmentMixin):
             querystring.update(headers)
 
         querystring = _decode_dict(querystring)
-
         self.uri = full_url
         self.path = urlparse(full_url).path
         self.querystring = querystring
@@ -178,6 +179,31 @@ class BaseResponse(_TemplateEnvironmentMixin):
         self.setup_class(request, full_url, headers)
         return self.call_action()
 
+    def _get_action_from_method_and_request_uri(self, method, request_uri):
+        """basically used for `rest-json` APIs
+        You can refer to example from link below
+        https://github.com/boto/botocore/blob/develop/botocore/data/iot/2015-05-28/service-2.json
+        """
+
+        # service response class should have 'SERVICE_NAME' class member,
+        # if want to get action from method and url
+        if not hasattr(self, 'SERVICE_NAME'):
+            return None
+        service = self.SERVICE_NAME
+        conn = boto3.client(service)
+
+        # make cache if it has not exist yet
+        if not hasattr(self, 'method_urls'):
+            self.method_urls = {}
+            op_names = conn._service_model.operation_names
+            for op_name in op_names:
+                op_model = conn._service_model.operation_model(op_name)
+                k = '{}_{}'.format(op_model.http['method'], op_model.http['requestUri'])
+                self.method_urls[k] = op_model.name
+
+        given_key = '{}_{}'.format(method, request_uri)
+        return self.method_urls.get(given_key)
+
     def _get_action(self):
         action = self.querystring.get('Action', [""])[0]
         if not action:  # Some services use a header for the action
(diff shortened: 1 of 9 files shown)


toshitanian:
okay, please review this ;)

toshitanian:
@JackDanger This PR is ready. Can you review and merge this?

moto/core/responses.py
```diff
@@ -178,6 +181,58 @@ def _dispatch(self, request, full_url, headers):
...
+        return regexp
+
+    def _get_action_from_method_and_request_uri(self, method, request_uri):
+        """basically used for `rest-json` APIs
+        You can refer to example from link below
+        https://github.com/boto/botocore/blob/develop/botocore/data/iot/2015-05-28/service-2.json
+        """
+
+        # service response class should have 'SERVICE_NAME' class member,
+        # if want to get action from method and url
```
JackDanger: Small grammar change in comment: `if want to get action` -> `if you want to get the action`

moto/core/responses.py
```diff
@@ -178,6 +181,58 @@ def _dispatch(self, request, full_url, headers):
...
+        if not hasattr(self, 'SERVICE_NAME'):
+            return None
+        service = self.SERVICE_NAME
+        conn = boto3.client(service)
+
+        # make cache if it has not exist yet
```
JackDanger: I suggest changing to "make cache if it does not exist yet"

moto/core/responses.py
```diff
@@ -221,6 +278,22 @@ def _get_param(self, param_name, if_none=None):
         val = self.querystring.get(param_name)
         if val is not None:
             return val[0]
+
+        # try to get json body parameter
+        try:
+            j = json.loads(self.body)
+            # raise key error if key really does not exist
+            return j[param_name]
+        except:
```
JackDanger: This `except` catches a bit too much. How about we add clauses for `except JSONDecodeError` and `except KeyError` for the two things that could go wrong here?

```
try:
  return json.loads(self.body)[param_name]
except JSONDecodeError: pass
except KeyError: pass
```

moto/core/responses.py
```diff
@@ -221,6 +278,22 @@ def _get_param(self, param_name, if_none=None):
...
+            # do nothing if param is not found
+            pass
+        # try to get path parameter
+        if self.uri_match:
+            try:
+                return self.uri_match.group(param_name)
+            except:
```
JackDanger: How about `except IndexError:` instead?

scripts/scaffold.py
```diff
@@ -413,81 +429,55 @@ def insert_url(service, operation):
...
-
-def insert_json_codes(service, operation):
-    func_in_responses = get_function_in_responses(service, operation, 'json')
-    func_in_models = get_function_in_models(service, operation)
-
-    # edit responses.py
-    responses_path = 'moto/{}/responses.py'.format(service)
-    print_progress('inserting code', responses_path, 'green')
-    insert_code_to_class(responses_path, BaseResponse, func_in_responses)
+    if api_protocol == 'query':
```
JackDanger: Oooh, nice!

moto/core/responses.py
toshitanian: thanks, I will use Value error instead of JSONDecodeError to support both python2 and python3

toshitanian: And, made new line for each exception handling because linter warns.

toshitanian:
Thank you for reviewing for your early morning ;) I fixed the points you mentioned. @JackDanger

toshitanian:
sorry, ci is still running....

JackDanger:
Thanks, @toshitanian! Great work on this, I'm so glad we have better tools for the people working inside Moto.

The builds are a bit backed up but once this passes I'll merge it.

toshitanian:
🤔

toshitanian:
It seems that there's a bug for python2

toshitanian:
🎉

Issue #1273, opened 2017-10-19
deepcopy in describe_instances call is very slow


drmorr0:
The deepcopy in [ec2/models.py](https://github.com/spulec/moto/blob/b40c5e557e93a1cacb87228aadc6083d5ccf9ec3/moto/ec2/models.py#L793) is very slow; it takes 15-30 seconds for a `describe_instances` call when I've only created 10 instances.  I was trying to do some tests with moto and several hundred instances and it took well over a minute.

I have a branch where I replaced the deep copy with a shallow copy, and it *significantly* improved the speed of the call; I can submit a PR but I wasn't sure of the other implications of making that change.  Do you think that's the right solution, or is there a better way to resolve the problem?

JackDanger:
I've seen this behavior myself. It looks like the only reason we need to copy them is because [`get_reservations_by_instance_id`](https://github.com/spulec/moto/blob/b40c5e557e93a1cacb87228aadc6083d5ccf9ec3/moto/ec2/models.py#L763) wants to mutate the return values. What do you think about changing how that method works to create new reservation objects instead of mutating?

It looks like we'd mostly have to replace this:
```
            if matching_reservation:
                # We need to make a copy of the reservation because we have to modify the
                # instances to limit to those requested
                reservation.instances = [
                    instance for instance in reservation.instances if instance.id in instance_ids]
                reservations.append(reservation)
```
with this:
```
            if matching_reservation:
                # We need to make a copy of the reservation because we have to modify the
                # instances to limit to those requested
                new_reservation = Reservation()
                new_reservation.id = reservation.id
                new_reservation.instances = [
                    instance for instance in reservation.instances if instance.id in instance_ids]
                reservations.append(new_reservation)
```
Care to take a stab at that?

drmorr0:
Yes, I can look into it!  Probably not until next week sometime.

PR #1295, opened 2017-10-23
fix deep-copy in ec2.models.all_reservations

drmorr0:
@JackDanger Here's a potential fix for #1273 

I debated whether or not to just copy the instance list (like you suggested) but thought that seemed more brittle in the long-term than just making a shallow copy of the reservation object.  This seems to work fine in my tests, and all tests pass locally.  Let me know what you think!


JackDanger:
LGTM!

JackDanger:

PR #1296, opened 2017-10-23
[EC2] VolumeInUseError when volumes still attached.

mickeypash:
If Volumes are attached to an EC2 instance and one tries to delete them this should result in a VolumeInUse error. As seen in http://docs.aws.amazon.com/AWSEC2/latest/APIReference/errors-overview.html

>The specified Amazon EBS volume is attached to an instance. Ensure that the specified volume is in an ‘available’ state.


moto/ec2/models.py
```diff
@@ -1810,6 +1811,10 @@ def get_volume(self, volume_id):
 
     def delete_volume(self, volume_id):
         if volume_id in self.volumes:
+            volume = self.volumes[volume_id]
+            instance_id = volume.attachment.instance.id
+            if volume.attachment is not None:
```
JackDanger: This method will always raise an error because if `volume.attachment` _is_ `None` then `volume.attachment.instance` would have raised an error a line before. Maybe you want to move line 1815 inside this conditional?

moto/ec2/models.py
```diff
@@ -1810,6 +1811,10 @@ def get_volume(self, volume_id):
 
     def delete_volume(self, volume_id):
         if volume_id in self.volumes:
+            volume = self.volumes[volume_id]
+            instance_id = volume.attachment.instance.id
```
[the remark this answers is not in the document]
terrycain: wtf, i wrote this comment yesterday, apparently I need to hit comment twice

mickeypash:
@JackDanger @terrycain thank you for pointing this out! Now I just have to write a test 💃

JackDanger:
This is blocked pending tests but it's still a good addition!

terrycain:
Any update on this?

mickeypash:
Hello! I'm back! I'll try to resolve this in the next couple of days.

tests/test_ec2/test_elastic_block_store.py
```diff
@@ -53,6 +53,45 @@ def test_create_and_delete_volume():
...
+    volume.attach_data.instance_id.should.equal(instance.id)
+
+    # attempt to delete volume
+    # assert raises VolumeInUseError
+    with assert_raises(EC2ResponseError) as ex:
+        volume.delete()
+    ex.exception.error_code.should.equal("VolumeInUse")
+    ex.exception.status.should.equal(400)
+    ex.exception.message.should.equal(
+        f"Volume {volume.id} is currently attached to {instance.id}"
```
bblommers, line 80: Hi @mickeypash - we're still supporting Python2, so f-strings are not supported. Please use the string.format method for this.
Looks good to me otherwise.

mickeypash: Oh sorry, will fix that now. I checked the [Docker](https://github.com/spulec/moto/blob/master/Dockerfile#L15) for the Python version which threw me off.

PR #1297, opened 2017-10-23
Do not treat StartRecordName and StartRecordType as filters

melentye:
Do not treat StartRecordName and StartRecordType in the Route53 list_resource_record_sets as filters.

According to [AWS API reference](http://docs.aws.amazon.com/Route53/latest/APIReference/API_ListResourceRecordSets.html) and [Boto](http://boto.readthedocs.io/en/glacier/ref/route53.html#boto.route53.connection.Route53Connection.get_all_rrsets) [documentation](http://boto3.readthedocs.io/en/latest/reference/services/route53.html#Route53.Client.list_resource_record_sets) they are to begin the record listing from, not for filtering. So their current behaviour in moto is not consistent with AWS.

### DIFF
--- a/moto/route53/models.py
+++ b/moto/route53/models.py
@@ -196,14 +196,14 @@ class FakeZone(BaseModel):
         self.rrsets = [
             record_set for record_set in self.rrsets if record_set.set_identifier != set_identifier]
 
-    def get_record_sets(self, type_filter, name_filter):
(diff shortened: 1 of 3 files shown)


JackDanger:
This is really great, @melentye! Thanks for stepping in to help out and we're glad to have you here at Moto!

JackDanger:

PR #1307, opened 2017-10-25
Adding initial implementation coverage

JackDanger:
Also creating a make task for generating implementation coverage output


terrycain:
Nice! I was thinking of tackling some of that soon.

PR #1308, opened 2017-10-25
Use pkg_resources to keep the package zip-safe

frivoire:
This resolves a bug in the packaging of this package that occurs in some situation :

When `moto` is installed manually from a source distribution, in a `PYTHONDONTWRITEBYTECODE=1` environment, it ends up in the shape of a **zipfile egg** in dist-packages, as it is not marked **zip-unsafe**.
This means that importing it will result in an error, because `moto/ec2/models.py` will try to open a file in what it thinks is inside a directory, when it is actually inside a zipfile.

The package could be marked as **zip-unsafe**, but there is an even better way to solve the problem: `pkg_resources` (which is part of setuptools) that will take care (when needed) of unzipping and caching the data file, returning a valid filename which can be consumed.

### DIFF
--- a/moto/ec2/models.py
+++ b/moto/ec2/models.py
@@ -3,9 +3,9 @@ from __future__ import unicode_literals
 import copy
 import itertools
 import json
-import os
 import re
 import six
+from pkg_resources import resource_filename
 
 import boto.ec2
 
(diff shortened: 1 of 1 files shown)


JackDanger:
I didn't know about `resource_filename` - thanks for fixing this for us!

frivoire:
NB: i've just removed the (now-unused) `import os` line.

frivoire:
Any idea why the 2nd travis test is failing on `urllib2.URLError: <urlopen error [Errno 111] Connection refused>` ?

terrycain:
@frivoire I've been getting that too. Think i've solved it in my pr so might be fixed when that merges.

terrycain:
Travis has passed :smile:. It was having a bad day

Issue #1278, opened 2017-10-20
Attaching ACM certificate to ELB listener does not update the certificates InUseBy property.


WhileLoop:
`

    @mock_acm
    @mock_elb
    def test(self):

        acm_client = boto3.client('acm', region_name="us-west-2")
        elb_client = boto3.client('elb', region_name="us-west-2")

        acm_request_response = acm_client.request_certificate(
            DomainName='fake.domain.com',
            DomainValidationOptions=[
                {
                    'DomainName': 'fake.domain.com',
                    'ValidationDomain': 'domain.com'
                },
            ]
        )

        certificate_arn = acm_request_response['CertificateArn']

        elb_client.create_load_balancer(
            LoadBalancerName='test',
            Listeners=[
                {
                    'Protocol': 'https',
                    'LoadBalancerPort': 443,
                    'InstanceProtocol': 'http',
                    'InstancePort': 80,
                    'SSLCertificateId': certificate_arn
                }
            ]
        )

        response = acm_client.describe_certificate(
            CertificateArn = certificate_arn
        )

        print(response['Certificate']['InUseBy'])`

WhileLoop:
@JackDanger If you point me in the right direction I can try to get this implemented.

JackDanger:
Absolutely! Thanks for tackling this.

calling `elb_client.create_load_balancer` invokes a method of the same name in `moto/elb/responses.py`. Right here is where the `Listeners=` value is extracted: https://github.com/spulec/moto/blob/master/moto/elb/responses.py#L27 (it's assigned to 'ports').

Then, the `self.elb_backend` is an instance of `ELBBackend` as defined in `moto/elb/models.py`. The `create_load_balancer` method is here: https://github.com/spulec/moto/blob/master/moto/elb/models.py#L238

You can see that the arguments are all passed into creating a new `FakeLoadBalancer` and then the listener (called 'port') is passed into a `FakeListener` here: https://github.com/spulec/moto/blob/master/moto/elb/models.py#L88

[...]

PR #1309, opened 2017-10-25
add acm in use by logic and test

WhileLoop:
Fixes https://github.com/spulec/moto/issues/1278

@JackDanger Please advise on how to reduce code duplication and how to properly reference acm_backend.

### DIFF
--- a/moto/acm/models.py
+++ b/moto/acm/models.py
@@ -82,6 +82,7 @@ class CertBundle(BaseModel):
         self._chain = None
         self.type = cert_type  # Should really be an enum
         self.status = cert_status  # Should really be an enum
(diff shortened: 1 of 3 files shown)


WhileLoop:
@JackDanger What should happen when someone creates a ELB and references a certificate that doesn't exists? See failing tests.

JackDanger:
According to [this](https://docs.aws.amazon.com/elasticloadbalancing/2012-06-01/APIReference/API_SetLoadBalancerListenerSSLCertificate.html) it should return a 'CertificateNotfound' error. Would you mind updating the old tests to check for that error when the certificate is invalid and then add a few lines that create a valid certificate and use it?

PR #1310, opened 2017-10-26
Use current time for Cloudformation Stack CreationTime

saadgit:
Before it was using hardcoded creationtime

### DIFF
--- a/tests/test_cloudformation/test_cloudformation_stack_crud.py
+++ b/tests/test_cloudformation/test_cloudformation_stack_crud.py
@@ -2,11 +2,15 @@ from __future__ import unicode_literals
 
 import os
 import json
+import datetime
 
 import boto
 import boto.s3
 import boto.s3.key
 import boto.cloudformation
+
+from mock import Mock, patch
+
 from boto.exception import BotoServerError
 import sure  # noqa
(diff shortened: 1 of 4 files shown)


tests/test_cloudformation/test_cloudformation_stack_crud.py
```diff
@@ -243,19 +250,22 @@ def test_get_template_by_name():
 
 @mock_cloudformation_deprecated
 def test_list_stacks():
+    fixed_timestamp = datetime.datetime(2017, 10, 20, 13, 13, 13)
```
terrycain: Would using freezegun like in other tests look nicer when dealing with times?

E.g.
https://github.com/spulec/moto/blob/master/tests/test_dynamodb2/test_dynamodb_table_with_range_key.py#L66
https://github.com/spulec/moto/blob/master/tests/test_sqs/test_sqs.py#L764

terrycain:
Also if your going to be messing with time in your tests, that wont apply when in "server mode" so add something like [this](https://github.com/spulec/moto/blob/master/tests/test_sqs/test_sqs.py#L761-L762) to the respective tests

bblommers:
This is now implemented as part of #2818. Closing

PR #1312, opened 2017-10-27
validate protocol on target group creation

JackDanger:


moto/elbv2/models.py
```diff
@@ -300,6 +300,16 @@ def create_target_group(self, name, **kwargs):
...
+        if kwargs['healthcheck_protocol'] not in valid_protocols:
+            raise InvalidConditionValueError(
+                "Value {} at 'healthCheckProtocol' failed to satisfy constraint: "
+                "Member must satisfy enum value set: {}".format(kwargs['healthcheck_protocol'], valid_protocols))
+        if kwargs['protocol'] not in valid_protocols:
+            raise InvalidConditionValueError(
+                "Value {} at 'protocol' failed to satisfy constraint: "
+                "Member must satisfy enum value set: {}".format(kwargs['protocol'], valid_protocols))
+
         arn = "arn:aws:elasticloadbalancing:%s:1:targetgroup/%s/50dc6c495c0c9188" % (self.region_name, name)
```
terrycain: I know you didn't edit it, but shouldn't ARNs be made by a function as I'm assuming it would appear elsewhere too? and probably be useful for when cloudformation support is added if not already

JackDanger: Yeah, that's a great idea. I don't believe we've standardized them at all yet but we should develop a good practice for generating these.

terrycain: Yeah from what I've seen is most of the arns have functions in a `utils.py`

PR #1314, opened 2017-10-27
Add CloudFormation support to ELB V2

hltbra:
This pull request adds support to some ELBv2 resources (load balancer, target group, and listener).

I tested this implementation in an internal project and it works as expected. I found that outputs weren't properly evaluated (didn't support functions such as Fn::GetAtt). I fixed it to always evaluate output values.

### DIFF
--- a/moto/elbv2/responses.py
+++ b/moto/elbv2/responses.py
@@ -472,9 +472,14 @@ CREATE_TARGET_GROUP_TEMPLATE = """<CreateTargetGroupResponse xmlns="http://elast
         <HealthCheckTimeoutSeconds>{{ target_group.healthcheck_timeout_seconds }}</HealthCheckTimeoutSeconds>
         <HealthyThresholdCount>{{ target_group.healthy_threshold_count }}</HealthyThresholdCount>
         <UnhealthyThresholdCount>{{ target_group.unhealthy_threshold_count }}</UnhealthyThresholdCount>
+        {% if target_group.matcher %}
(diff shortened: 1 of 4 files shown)


moto/elbv2/responses.py
```diff
@@ -634,16 +640,21 @@ def _add_tags(self, resource):
         <Protocol>{{ target_group.protocol }}</Protocol>
         <Port>{{ target_group.port }}</Port>
         <VpcId>{{ target_group.vpc_id }}</VpcId>
-        <HealthCheckProtocol>{{ target_group.health_check_protocol }}</HealthCheckProtocol>
+        <HealthCheckProtocol>{{ target_group.healthcheck_protocol }}</HealthCheckProtocol>
```
JackDanger: Nice fix!

moto/elbv2/responses.py
```diff
@@ -472,9 +472,14 @@ def _add_tags(self, resource):
         <HealthCheckTimeoutSeconds>{{ target_group.healthcheck_timeout_seconds }}</HealthCheckTimeoutSeconds>
         <HealthyThresholdCount>{{ target_group.healthy_threshold_count }}</HealthyThresholdCount>
         <UnhealthyThresholdCount>{{ target_group.unhealthy_threshold_count }}</UnhealthyThresholdCount>
+        {% if target_group.matcher %}
```
JackDanger: This is much better, thanks.

PR #1320, opened 2017-11-03
ASG and LC arn fix

karthikax:
LaunchConfiguration and AutoScalingGroup returns appropriate ARNs based on the names passed instead of hard coded values.


spulec:
Thanks @karthikax and welcome to the Moto project!

PR #1326, opened 2017-11-07
Adds FilterExpression to dynamodb.query

### LINKED ISSUE #715: dynamodb.query doesn't support FilterExpression
Docs:

https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/QueryAndScan.html#FilteringResults

It looks like mock for the `FilterExpression` operator is not implemented. It simply isn't taken into account and everything is always returned.

``` python
table = session.resource('dynamodb').Table('table_name')
result = table.query(KeyConditionExpression=boto3.dynamodb.conditions.Key("key").eq(key),
                     FilterExpression=boto3.dynamodb.conditions.Attr("col").eq(1))
```

This behaves as in `FilterExpression` weren't specified.

Are there plains to add support for `FilterExpression`?

Thanks!

terrycain:
Fixes #715

### DIFF
--- a/moto/dynamodb2/models.py
+++ b/moto/dynamodb2/models.py
@@ -419,7 +419,7 @@ class Table(BaseModel):
 
     def query(self, hash_key, range_comparison, range_objs, limit,
               exclusive_start_key, scan_index_forward, projection_expression,
-              index_name=None, **filter_kwargs):
+              index_name=None, filter_expression=None, **filter_kwargs):
         results = []
         if index_name:
             all_indexes = (self.global_indexes or []) + (self.indexes or [])
@@ -501,6 +501,9 @@ class Table(BaseModel):
             results.reverse()
 
         scanned_count = len(list(self.all_items()))
+
+        if filter_expression is not None:
+            results = [item for item in results if filter_expression.expr(item)]
 
         results, last_evaluated_key = self._trim_results(results, limit,
                                                          exclusive_start_key)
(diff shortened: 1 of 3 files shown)


terrycain:
FilterExpression should be good now, I think it should work with most of what people will throw at it so should be find adding it to query. And better it die that silently be ignored.

moto/dynamodb2/models.py
```diff
@@ -707,7 +710,9 @@ def get_item(self, table_name, keys):
         return table.get_item(hash_key, range_key)
 
     def query(self, table_name, hash_key_dict, range_comparison, range_value_dicts,
-              limit, exclusive_start_key, scan_index_forward, projection_expression, index_name=None, **filter_kwargs):
+              limit, exclusive_start_key, scan_index_forward, projection_expression, index_name=None,
+              expr_names={}, expr_values={}, filter_expression=None,
```
JackDanger: Rather than having a static dictionary instance as the default for `expr_names` and `expr_values` how about just make them `None`? There's already a check for `None` for these values in `get_filter_expression`

JackDanger:
One comment but it's not actually a bug, just a vector for somebody to accidentally cause a bug in the future. Approved and please merge at will :)

moto/dynamodb2/models.py
terrycain: Ha totally agree, my ide was also complaining about the `{}`, was just toooo lazy to check `get_filter_expression`. Will fix then merge if tests pass.

PR #1329, opened 2017-11-08
SNS create_topic should only accept valid characters

### LINKED ISSUE #1328: SNS.create_topic shouldn't accept invalid input
Per the [AWS SNS documentation](http://docs.aws.amazon.com/sns/latest/api/API_CreateTopic.html) the CreateTopic shouldn't accept topic names that don't match the constraint: _Topic names must be made up of only uppercase and lowercase ASCII letters, numbers, underscores, and hyphens, and must be between 1 and 256 characters long._

Expected:
- create_topic should raise an InvalidParameterValue when an invalid topic is passed to create_topic

Results:
- no error is raised

adamgilman:
Closes #1328


adamgilman:
Fixing Travis’ whining now

terrycain:
Ha yeah flake8 gets me alot

terrycain:
The API docs also don't mention spaces, the regex will need changing
If were going on the assumption that we don't want a topic to start or end with a _/- this regex will do that and enforce the length is at least 1 and not longer than 256 too `^[a-zA-Z0-9](?:[A-Za-z0-9_-]{0,253}[a-zA-Z0-9])?$`

moto/sns/models.py
```diff
@@ -193,6 +195,9 @@ def update_sms_attributes(self, attrs):
         self.sms_attributes.update(attrs)
 
     def create_topic(self, name):
+        fails_constraints = not re.match(r'^[a-zA-Z0-9](?:[A-Za-z0-9_-]{0,253}[a-zA-Z0-9])?$', name)
+        if fails_constraints:
+            raise InvalidParameterValue("Can only include alphanumeric characters, hyphens, or underscores. 1 to 80 in length")
```
terrycain: 1 to 256 ?

adamgilman: Just copy and pasted what you gave me :)

terrycain: I meant the string in the exception

JackDanger:
@adamgilman this is just about finished, there are just a couple of comments by Terry that need to be addressed. Thanks for doing this!

adamgilman:
Sorry for the delay. Been busy at work and have it patched locally. Will
try to get a PR done today

JackDanger:
Nice! @terrycain please merge if/when you approve this.

terrycain:
Yeah looks gread :smile:

PR #1335, opened 2017-11-12
AutoScalingBackend / Implementation Coverage quirk

captainkerk:
I was curious as to why the implementation details page wasn't reporting on methods that I knew existed, so I took a look. 

For each module, the implementation script is looking up the Backend object in the services `models.py` file (e.g. for the auto scaling service the backend is called AutoScalingBackend), getting a list of methods, and comparing directly to the boto3 client. It seems that for moto itself to register that the method has been mocked (i.e. when moto is used in a test), the method contained in the `responses.py` file needs to match the boto3 client. 

It seems that both the method in `responses.py` and the method it calls in `models.py` need to be named the same? Just so the mocked method works, and the implementation script picks it up.

### DIFF
--- a/moto/autoscaling/models.py
+++ b/moto/autoscaling/models.py
@@ -185,7 +185,7 @@ class FakeAutoScalingGroup(BaseModel):
         target_group_arns = properties.get("TargetGroupARNs", [])
 
         backend = autoscaling_backends[region_name]
-        group = backend.create_autoscaling_group(
+        group = backend.create_auto_scaling_group(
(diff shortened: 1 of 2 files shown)


JackDanger:
This is great, @captainkerk! I never would have caught this. Thanks for lining this up.

captainkerk:
FWIW - I think theres a slight issue with RDS as well: https://github.com/spulec/moto/blob/master/IMPLEMENTATION_COVERAGE.md#rds---0-implemented

PR #1340, opened 2017-11-14
fix EC2 instances ebsOptimized not handled

Archa91:
Hi,

It appears that describe_instances not returns EbsOptimized field.

Here is a improvment which takes EbsOptimized into account during create_instances and correctly return the value with a describe_instances.
Default value is  False.

Regards,


terrycain:
Can we have a test case that shows ebsOptimised is on and is off. I have a feeling you'd need to lowercase `{{ instance.ebs_optimized }}` as i think boto does a pretty basic `if instance.ebs_optimized == 'true'`, so True would evaluate to false.

Archa91:
Now, is it OK for you ? Thank you for your efficiency !

terrycain:
Yup looks good. Welcome to the Moto project.

PR #1341, opened 2017-11-14
Add proper CFN attribute handling to elbv2.FakeLoadBalancer

hltbra:
When I originally implemented `elbv2.models.FakeLoadBalancer.get_cfn_attribute()` I didn't realize `moto` would raise an exception if anything tried to use undefined attributes. I mimicked the `elb.models.FakeLoadBalancer.get_cfn_attribute()` method (but focused on elbv2 load balancers).

### DIFF
--- a/moto/elbv2/models.py
+++ b/moto/elbv2/models.py
@@ -293,11 +293,32 @@ class FakeLoadBalancer(BaseModel):
         return load_balancer
 
     def get_cfn_attribute(self, attribute_name):
-        attributes = {
-            'DNSName': self.dns_name,
-            'LoadBalancerName': self.name,
-        }
-        return attributes[attribute_name]
+        '''
+        Implemented attributes:
+        * DNSName
+        * LoadBalancerName
+
+        Not implemented:
+        * CanonicalHostedZoneID
+        * LoadBalancerFullName
+        * SecurityGroups
+
+        This method is similar to models.py:FakeLoadBalancer.get_cfn_attribute()
+        '''
+        from moto.cloudformation.exceptions import UnformattedGetAttTemplateException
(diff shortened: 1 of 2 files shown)


moto/elbv2/models.py
```diff
@@ -293,11 +293,32 @@ def create_from_cloudformation_json(cls, resource_name, cloudformation_json, reg
...
-        return attributes[attribute_name]
+        '''
+        Implemented attributes:
+        * DNSName
+        * LoadBalancerName
+
+        Not implemented:
+        * CanonicalHostedZoneID
+        * LoadBalancerFullName
+        * SecurityGroups
```
terrycain: :+1: Good comment, allows future developers to quickly see what needs adding

moto/elbv2/models.py
```diff
@@ -293,11 +293,32 @@ def create_from_cloudformation_json(cls, resource_name, cloudformation_json, reg
...
+
+        This method is similar to models.py:FakeLoadBalancer.get_cfn_attribute()
+        '''
+        from moto.cloudformation.exceptions import UnformattedGetAttTemplateException
```
terrycain: Is there a reason this cant to at the top of the file?

terrycain:
Other than that, looks great.

moto/elbv2/models.py
hltbra: @terrycain: This is to work around a circular import (`moto.cloudformation` imports `elbv2.models`). The `elb.models.FakeLoadBalancer` class does the same thing, by the way (I mimicked it).

terrycain: Ahh I see, thats fine then.

PR #1343, opened 2017-11-15
Move target group default values to model FakeTargetGroup

hltbra:
My CloudFormation TargetGroup creation was not passing the default values to the model because I had assumed the FakeTargetGroup model would have the default values in its `__init__` (it didn't).

Before this commit everything that creates target groups had to handle the default values (i.e., cloudformation call & ELBV2Response call) separately from the model. The default values are the model's responsibility and this PR moves the default values to the initializer of FakeTargetGroup. This PR also fixes issues when creating `TargetGroup`s via CloudFormation without all optional parameters.

### DIFF
--- a/moto/elbv2/models.py
+++ b/moto/elbv2/models.py
@@ -52,13 +52,13 @@ class FakeTargetGroup(BaseModel):
                  vpc_id,
                  protocol,
                  port,
-                 healthcheck_protocol,
-                 healthcheck_port,
-                 healthcheck_path,
-                 healthcheck_interval_seconds,
-                 healthcheck_timeout_seconds,
-                 healthy_threshold_count,
-                 unhealthy_threshold_count,
+                 healthcheck_protocol='HTTP',
+                 healthcheck_port='traffic-port',
+                 healthcheck_path='/',
+                 healthcheck_interval_seconds='30',
+                 healthcheck_timeout_seconds='5',
+                 healthy_threshold_count='5',
+                 unhealthy_threshold_count='2',
                  matcher=None,
                  target_type=None):
         self.name = name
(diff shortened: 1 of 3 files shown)


moto/elbv2/models.py
```diff
@@ -454,28 +457,18 @@ def create_target_group(self, name, **kwargs):
                 raise DuplicateTargetGroupName()
 
         valid_protocols = ['HTTPS', 'HTTP', 'TCP']
-        if kwargs['healthcheck_protocol'] not in valid_protocols:
+        if kwargs.get('healthcheck_protocol') and kwargs['healthcheck_protocol'] not in valid_protocols:
```
hltbra: Only validate optional parameters if passed.

moto/elbv2/models.py
```diff
@@ -454,28 +457,18 @@ def create_target_group(self, name, **kwargs):
...
+        if kwargs.get('protocol') and kwargs['protocol'] not in valid_protocols:
             raise InvalidConditionValueError(
                 "Value {} at 'protocol' failed to satisfy constraint: "
                 "Member must satisfy enum value set: {}".format(kwargs['protocol'], valid_protocols))
 
-        if FakeTargetGroup.HTTP_CODE_REGEX.match(kwargs['matcher']['HttpCode']) is None:
+        if kwargs.get('matcher') and FakeTargetGroup.HTTP_CODE_REGEX.match(kwargs['matcher']['HttpCode']) is None:
             raise RESTError('InvalidParameterValue', 'HttpCode must be like 200 | 200-399 | 200,201 ...')
 
-        valid_protocols = ['HTTPS', 'HTTP', 'TCP']
```
hltbra: This block of code was duplicated.

terrycain:
Looks good, yeah it was probably me that messed up some of that a while ago.

PR #1344, opened 2017-11-16
Lambda improvements

### LINKED ISSUE #1338: lambda doesn't support get function by name and version
When calling the get_function API and specifying the version in the qualifier parameter, the call throws a ClientError('An error occurred (404) when calling the GetFunction operation: ') 

function_data = client.get_function(FunctionName="my_func", Qualifier="$LATEST")  

Looking at the source code, this occurs because the version gets concatenated to the function name.

terrycain:
Fixes #1338 

Completely revamped how lambdas are stored, now we have lambda publishing working and aliasing should be trivial to add, attempted to spam `weakref` everywhere but I have a feeling my debugger was messing with the refs so theres more boilerplate code that I would of expected.

Done the majority of this whilst sleep deprived, @JackDanger keep an eye out for the spelling mistakes ;-).

Also cleaned up the responses.py, made use of a json property and backend property, I think I inadvertently fixed some region malarkey when in server mode too.

### DIFF
--- a/moto/awslambda/models.py
+++ b/moto/awslambda/models.py
@@ -644,11 +644,11 @@ class LambdaBackend(BaseBackend):
 
     def tag_resource(self, resource, tags):
         fn = self.get_function_by_arn(resource)
-        if fn:
-            fn.tags.update(tags)
-            return True
+        if not fn:
+            return False
 
-        return False
+        fn.tags.update(tags)
+        return True
 
     def untag_resource(self, resource, tagKeys):
         fn = self.get_function_by_arn(resource)
--- a/moto/awslambda/models.py
+++ b/moto/awslambda/models.py
@@ -2,6 +2,7 @@ from __future__ import unicode_literals
 
 import base64
 from collections import defaultdict
+import copy
 import datetime
 import docker.errors
 import hashlib
@@ -17,17 +18,22 @@ import tarfile
 import calendar
 import threading
 import traceback
+import weakref
 import requests.adapters
 
 import boto.awslambda
 from moto.core import BaseBackend, BaseModel
+from moto.core.exceptions import RESTError
 from moto.core.utils import unix_time_millis
 from moto.s3.models import s3_backend
 from moto.logs.models import logs_backends
 from moto.s3.exceptions import MissingBucket, MissingKey
 from moto import settings
+from .utils import make_function_arn
 
 logger = logging.getLogger(__name__)
+
+ACCOUNT_ID = '123456789012'
 
 
 try:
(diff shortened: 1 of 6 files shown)


moto/awslambda/models.py
```diff
@@ -161,7 +167,7 @@ def replace_adapter_send(*args, **kwargs):
             'VpcConfig', {'SubnetIds': [], 'SecurityGroupIds': []})
 
         # auto-generated
```
JackDanger: Is this comment still true?

moto/awslambda/models.py
```diff
@@ -203,11 +209,15 @@ def replace_adapter_send(*args, **kwargs):
                 self.code_size = key.size
                 self.code_sha_256 = hashlib.sha256(key.value).hexdigest()
 
-        self.function_arn = 'arn:aws:lambda:{}:123456789012:function:{}'.format(
-            self.region, self.function_name)
+        self.function_arn = make_function_arn(self.region, ACCOUNT_ID, self.function_name, version)
```
JackDanger: nice.

moto/awslambda/models.py
```diff
@@ -515,23 +636,31 @@ def send_message(self, function_name, message):
...
+        self._functions[function_name][-1].invoke(json.dumps(event), {}, {})
         pass
 
     def list_tags(self, resource):
         return self.get_function_by_arn(resource).tags
 
     def tag_resource(self, resource, tags):
-        self.get_function_by_arn(resource).tags.update(tags)
+        fn = self.get_function_by_arn(resource)
+        if fn:
```
JackDanger: minor suggestion: I'd personally prefer a guard clause here that early-returns on `if not fn` so the rest of the method isn't as indented. Same as the following method.

moto/awslambda/models.py
```diff
@@ -446,41 +455,153 @@ def create_from_cloudformation_json(cls, resource_name, cloudformation_json,
         return LambdaVersion(spec)
 
 
+class LambdaStorage(object):
```
JackDanger: This is great

tests/test_awslambda/test_lambda.py
```diff
@@ -317,30 +317,25 @@ def test_get_function():
...
-            "Timeout": 3,
-            "Version": '$LATEST',
-            "VpcConfig": {
-                "SecurityGroupIds": [],
-                "SubnetIds": [],
-            }
-        },
-        'ResponseMetadata': {'HTTPStatusCode': 200},
-    })
+    result['Code']['Location'].should.equal('s3://awslambda-{0}-tasks.s3-{0}.amazonaws.com/test.zip'.format(_lambda_region))
```
JackDanger: Nice. Way less brittle

moto/awslambda/models.py
terrycain: Hmm, possibly, will investigate.

moto/awslambda/models.py
terrycain: Ahh, now we get down to personal preference :wink:. But I agree, less indentation is nice, and its an easy fix

PR #1346, opened 2017-11-16
Add default ecs attributes and format in response obj

nadlerjessie:
This PR does two things:

1) It adds the built-in attributes automatically applied by ECS. More on that in the docs, [here](http://docs.aws.amazon.com/AmazonECS/latest/developerguide/task-placement-constraints.html#attributes).
   - Note: In order to do that I had to initialize the EC2ContainerServiceBackend and ContainerInstance with `region_name`.
2) Formats the `ContainerInstance` attributes to be consistent with the `describe_container_instances` [response](http://boto3.readthedocs.io/en/latest/reference/services/ecs.html#ECS.Client.describe_container_instances).
   - Note: I didn't exactly implement it the way the docs suggest with `targetType` and `targetId` because I do not see that when I call `describe_container_instances` locally, I only see `name` and `value` (optional). Please let me know if you can get boto3 to return those values and think I should include them.

Thanks!

cc @spulec

### DIFF
--- a/moto/ecs/models.py
+++ b/moto/ecs/models.py
@@ -333,7 +333,7 @@ class ContainerInstance(BaseObject):
     @property
     def response_object(self):
         response_object = self.gen_response_object()
-        response_object['attributes'] = [self._format_attribute(name, value) for name, value in response_object['attributes'].iteritems()]
+        response_object['attributes'] = [self._format_attribute(name, value) for name, value in response_object['attributes'].items()]
         return response_object
 
     def _format_attribute(self, name, value):
--- a/moto/ecs/models.py
+++ b/moto/ecs/models.py
@@ -333,7 +333,16 @@ class ContainerInstance(BaseObject):
     @property
     def response_object(self):
         response_object = self.gen_response_object()
+        response_object['attributes'] = [self._format_attribute(name, value) for name, value in response_object['attributes'].iteritems()]
         return response_object
+
+    def _format_attribute(self, name, value):
+        formatted_attr = {
+            'name': name,
+        }
+        if value is not None:
+            formatted_attr['value'] = value
+        return formatted_attr
 
 
 class ContainerInstanceFailure(BaseObject):
--- a/moto/ecs/models.py
+++ b/moto/ecs/models.py
@@ -2,6 +2,7 @@ from __future__ import unicode_literals
 import uuid
 from datetime import datetime
 from random import random, randint
+import boto3
 
 import pytz
 from moto.core.exceptions import JsonRESTError
@@ -888,6 +889,5 @@ class EC2ContainerServiceBackend(BaseBackend):
             yield task_fam
 
 
-ecs_backends = {}
-for region, ec2_backend in ec2_backends.items():
-    ecs_backends[region] = EC2ContainerServiceBackend(region)
+available_regions = boto3.session.Session().get_available_regions("ecs")
+ecs_backends = {region: EC2ContainerServiceBackend(region) for region in available_regions}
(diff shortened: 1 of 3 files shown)


moto/ecs/models.py
```diff
@@ -868,4 +890,4 @@ def list_task_definition_families(self, family_prefix=None, status=None, max_res
 
 ecs_backends = {}
 for region, ec2_backend in ec2_backends.items():
-    ecs_backends[region] = EC2ContainerServiceBackend()
+    ecs_backends[region] = EC2ContainerServiceBackend(region)
```
terrycain: Whilst your editing this bit, can you change it so it matches how its done [here](https://github.com/spulec/moto/blob/master/moto/batch/models.py#L1042)

Mainly as I know from experience ECS isnt 100% available where EC2 is (currently Mumbai, São Paulo and I'm pretty sure China)

moto/ecs/models.py
```diff
@@ -321,14 +321,29 @@ def __init__(self, ec2_instance_id):
...
+            'ecs.ami-id': ec2_instance.image_id,
+            'ecs.availability-zone': ec2_instance.placement,
+            'ecs.instance-type': ec2_instance.instance_type,
+            'ecs.os-type': ec2_instance.platform if ec2_instance.platform == 'windows' else 'linux'  # options are windows and linux, linux is default
+        }
 
     @property
     def response_object(self):
         response_object = self.gen_response_object()
+        response_object['attributes'] = [self._format_attribute(name, value) for name, value in response_object['attributes'].iteritems()]
```
terrycain: Just use `.items()`, there's not going to be 100's of items so you gain very little from the iterator, plus loose py3 compatibility.

moto/ecs/models.py
nadlerjessie: Will do!

nadlerjessie:
@terrycain Do you know what is causing the `Exception: Patch operation "add" not implemented` and `AttributeError: 'dict' object has no attribute '__dict__'` in the py3 + server mode test case? I haven't worked with the server mode and am a bit confused about the exceptions.

tests/test_ecs/test_ecs_boto3.py
```diff
@@ -1757,6 +1764,111 @@ def test_list_task_definition_families():
...
+    )
+
+    response['containerInstance'][
+        'ec2InstanceId'].should.equal(test_instance.id)
+    full_arn = response['containerInstance']['containerInstanceArn']
+    container_instance_id = full_arn.rsplit('/', 1)[-1]
+
+    default_attributes = response['containerInstance']['attributes']
+    assert len(default_attributes) == 4
+    assert sorted(default_attributes) == sorted([
```
terrycain: Cant sort dictionaries like this. you need to add `key=lambda item: item['name']` as an argument to the sorted function so it knows what to sort on, do it for both calls to sorted.

terrycain:
You can ignore that patch bit. The error is further down, have left a comment inline.

tests/test_ecs/test_ecs_boto3.py
nadlerjessie: I can do that but this is a list of dictionaries. Even if it isn't sorting the way I expect, is that causing the errors I mentioned?  I'm having trouble understanding the connection between how that assertion error would fail and cause the Exception/AttributeError described above.

terrycain: The exception caused above appears because moto in server mode repurposes its Flask internals for pretending to be an API gateway, and a test is running a PATCH request, which is not supported, therefore flask returns a 500like error code and prints a stack trace, can be ignored.

Skip to line 9281 in the build output.

nadlerjessie: I actually missed that, thanks for pointing it out and thanks for the review!

terrycain:
Great job :+1:

PR #1349, opened 2017-11-20
Set default value to elbv2 FakeTargetGroup

toshitanian:
Setting `None` to property of FakeTargergroup in ElbV2 class breaks response of create_target_group sometimes.
So, set aws's default values to them

The error status is as below.
```
======================================================================
ERROR: test_elbv2.test_create_target_group_without_non_required_parameters
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/usr/local/lib/python3.6/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/app/moto/core/models.py", line 71, in wrapper
    result = func(*args, **kwargs)
  File "/app/moto/core/models.py", line 71, in wrapper
    result = func(*args, **kwargs)
  File "/app/tests/test_elbv2/test_elbv2.py", line 453, in test_create_target_group_without_non_required_parameters
    Matcher={'HttpCode': '200'})
  File "/usr/local/lib/python3.6/site-packages/botocore/client.py", line 312, in _api_call
    return self._make_api_call(operation_name, kwargs)
  File "/usr/local/lib/python3.6/site-packages/botocore/client.py", line 592, in _make_api_call
    operation_model, request_dict)
  File "/usr/local/lib/python3.6/site-packages/botocore/endpoint.py", line 141, in make_request
    return self._send_request(request_dict, operation_model)
  File "/usr/local/lib/python3.6/site-packages/botocore/endpoint.py", line 168, in _send_request
    request, operation_model, attempts)
  File "/usr/local/lib/python3.6/site-packages/botocore/endpoint.py", line 233, in _get_response
    response_dict, operation_model.output_shape)
  File "/usr/local/lib/python3.6/site-packages/botocore/parsers.py", line 212, in parse
    parsed = self._do_parse(response, shape)
  File "/usr/local/lib/python3.6/site-packages/botocore/parsers.py", line 462, in _do_parse
    parsed = self._parse_shape(shape, start)
  File "/usr/local/lib/python3.6/site-packages/botocore/parsers.py", line 261, in _parse_shape
    return handler(shape, node)
  File "/usr/local/lib/python3.6/site-packages/botocore/parsers.py", line 330, in _handle_structure
    member_shape, member_node)
  File "/usr/local/lib/python3.6/site-packages/botocore/parsers.py", line 261, in _parse_shape
    return handler(shape, node)
  File "/usr/local/lib/python3.6/site-packages/botocore/parsers.py", line 314, in _handle_list
[...]

PR #1351, opened 2017-11-20
tweak elbv2 FakeTargetgroups's default value

toshitanian:
@terrycain  
Small fix for [earlier PR](https://github.com/spulec/moto/pull/1349).

As `FakeTargetgroup`'s default values are defined as default value of method parameter (modified to be so in #1343), aws's default parameters are now defined in two places.

So, I set `None` as default values in method parameter and set instance member's value to aws's default value in initialization phase if passed parameters are 'None'.
When service is called directly from `elbv2` client, not from cfn, these optional parameter are passed as `None`. So we have to check if passed parameters are None and set to default values if some of them are None.

### DIFF
--- a/moto/elbv2/models.py
+++ b/moto/elbv2/models.py
@@ -52,13 +52,13 @@ class FakeTargetGroup(BaseModel):
                  vpc_id,
                  protocol,
                  port,
-                 healthcheck_protocol='HTTP',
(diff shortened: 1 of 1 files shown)


terrycain:
So it all works now right?

toshitanian:
Yes, it looks good to me.

PR #1355, opened 2017-11-20
Updating CONTRIBUTING with release instructions

JackDanger:


toshitanian:
This will help new comers to contribute to this project :)

scripts/bump_version
```diff
@@ -0,0 +1,22 @@
...
+main() {
+  local version=$1
+  if [[ -z "${version}" ]]; then
+    echo "USAGE: $0 1.3.2"
+    echo "Provide a new version number as an argument to bump the version"
+    echo -n "Current:"
+    grep version= setup.py
+    return 1
+  fi
+  sed -i '' "s/version=.*$/version='${version}',/g" setup.py
```
terrycain: Would make sense to update https://github.com/spulec/moto/blob/master/moto/__init__.py#L6

Also the python package bumpversion does all of this for you ;-)

JackDanger: I didn't even see that other version number! What an oversight.

> Also the python package bumpversion does all of this for you ;-)

TIL! Thanks :)

PR #1359, opened 2017-11-22
fix KeyError in delete_alias in the KmsBackend.

xiaket:
If there're several aliases in the backend, previously we will bump into an unexpected KeyError here.

I haven't created a unit test for this little change. If the devs are inclined to add one, I'd be happy to do it.


terrycain:
Yeah a test would be nice if thats alright

terrycain:
Nice :smile:

PR #1366, opened 2017-11-30
modify_db_instance should be able to rename DB instances

syskill:
Fix the RDS and RDS2 backends to recognize the NewDBInstanceIdentifier argument and update their models accordingly.


syskill:
Ugh. Back to the drawing board.

PR #1374, opened 2017-12-04
fix cloudwatch bug

thehesiod:
introduced  by my last change as `Dimension.member` is not required..also add some needed unittests


thehesiod:
hrm, some unrelated infra failed, will have to look into this later

terrycain:
LOL death to boto, I second that!

Ok looks good.

PR #1375, opened 2017-12-05
Fix prebuilt images usability.

### LINKED ISSUE #1372: new PendingDeprecationWarning: Could not find AMI
Upgrading moto to latest version 1.25

Following snippet was used to create an imageId in moto:
```
        instance = r.create_instances(ImageId='foo', MinCount=1, MaxCount=1)[0]
        c.create_image(InstanceId=instance.id, Name="foo", Description="bar")
        c.terminate_instances(InstanceIds=[instance.id])
```
Now this snippet fails with a deprecation warning:

```
exceptions.PendingDeprecationWarning: Could not find AMI with image-id:foo, in the near future this will cause an error
```
We forbid all deprecation warnings in buildbot code, so we would like to have a new way of doing it which would not trigger the warning.

tardyp:
- add a better error message when user uses an unknown ami
- add image_location in the default fake images

Fixes #1372


terrycain:
Excellent, thanks for fixing that :smile: Small bit of oversight on my part.

PR #1377, opened 2017-12-05
Add cloudformation update from s3 support

william-richard:
Updating a cloudformation stack from a template body works, but not from an s3 template url.  This adds support for updating a stack from the s3 url.

In the process, I ended up finding a bug where the `dummy_update_template_json` in tests/test_cloudformation/test_cloudformation_stack_crud_boto3.py was pointed at the wrong template, so I fixed that and needed to add support for cloudformation deleting an ec2 instances to get the tests to pass afterwards.

### DIFF
--- a/tests/test_cloudformation/test_cloudformation_stack_crud_boto3.py
+++ b/tests/test_cloudformation/test_cloudformation_stack_crud_boto3.py
@@ -92,7 +92,6 @@ Resources:
           Value: !Ref TagName
 """
 
-
 dummy_update_template = {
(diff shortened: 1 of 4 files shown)


tests/test_cloudformation/test_cloudformation_stack_crud_boto3.py
```diff
@@ -1,18 +1,13 @@
 from __future__ import unicode_literals
 
+import json
+
 import boto3
-import boto
```
terrycain: Yes death to `boto`

terrycain:
Yup looks great :D

PR #1378, opened 2017-12-05
Make test_amis not executable, so nose runs it

william-richard:
In trying to debug changes to the ami mock introduced in 1.1.25,
I noticed that the ami tests were not running.  Turns out that nose does not
run test files that are executable.

http://nose.readthedocs.io/en/latest/finding_tests.html

The ami test file was the only test file I could find that had the executable bit set.

This seems like an unusual edge case, but should we add a test to travis that looks for executable test files?


william-richard:
It is expected that travis will fail on this PR - these tests have been broken for a while.

I need to go all the way back to tag 0.4.31 to get 0 failed tests.

There are 6 failures on this branch, and 3 on 1.1.24 so I'm pretty sure that the changes to the ami logic introduced is 1.1.25 introduced some test failures.

william-richard:
That said, I'm finding that I'm a bit over my head in trying to get these tests fixed. I'm just not positive how boto is calling out to moto, and when/how errors should be raised.  I'll keep working on it for a bit, but I'm not sure how much time I can commit to trying to fix this.

terrycain:
Awesome work, looks good to me. @JackDanger second opinion as the test's been off for so long.

william-richard:
*bump*

william-richard:
*bump again* @terrycain @JackDanger - it seems like there's other work going on with the files I touch here, so it would be great if you could merge this so those changes are getting the expected test coverage.  I even saw some changes to the ami tests.

william-richard:
*bump again* - #1398 introduced tests that did not pass, which was not noticed because of the issue this PR is trying to fix.  I have fixed those tests.  Is there something blocking this PR from being merged?

william-richard:
*bump*

JackDanger:
Thanks for this great work and for bumping it so we can see - apologies for the delay. This is a great addition. Thanks, @william-richard!

william-richard:
No worries - happy to help 😄

PR #1381, opened 2017-12-07
cast MaxRecords to int on  describe_db_instances

koxudaxi:
I couldn't call describe_db_instance with MaxRecords in rds2

boto3 expect MaxRecords argument as Integer, but rds2 use the variable as string
http://boto3.readthedocs.io/en/latest/reference/services/rds.html#RDS.Client.describe_db_instances

### DIFF
--- a/moto/rds2/responses.py
+++ b/moto/rds2/responses.py
@@ -123,7 +123,7 @@ class RDS2Response(BaseResponse):
             start = all_ids.index(marker) + 1
         else:
             start = 0
-        page_size = self._get_param('MaxRecords', 50)  # the default is 100, but using 50 to make testing easier
+        page_size = int(self._get_param('MaxRecords', 50))  # the default is 100, but using 50 to make testing easier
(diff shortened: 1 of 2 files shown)


bpandola:
A couple suggestions:

1) Use the built-in moto helper `self._get_int_param('MaxRecords', 50))` instead of `int(self._get_param('MaxRecords', 50))`.

2) There is already a `test_get_database_paginated` test case in `tests/test_rds2/test_rds2.py`, but it was missing a specific test around the `MaxRecords` parameter.  Adding the following lines to the end of that test case will cover the issue you encountered:
```
resp3 = conn.describe_db_instances(MaxRecords=100)
resp3["DBInstances"].should.have.length_of(51)
```

koxudaxi:
@bpandola thank you for good 
 suggestions.
I will fix and commit for 2 or 3 days.

spulec:
Thanks!

PR #1382, opened 2017-12-07
Path is an optional property of instance profile cloudformation resource

william-richard:
Based on http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-iam-instanceprofile.html

### DIFF
--- a/moto/iam/models.py
+++ b/moto/iam/models.py
@@ -119,11 +119,10 @@ class Role(BaseModel):
     def create_from_cloudformation_json(cls, resource_name, cloudformation_json, region_name):
         properties = cloudformation_json['Properties']
 
-        path = properties['Path'] if 'Path' in properties else '/'
         role = iam_backend.create_role(
             role_name=resource_name,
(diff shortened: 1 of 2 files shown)


moto/iam/models.py
```diff
@@ -119,10 +119,11 @@ def __init__(self, role_id, name, assume_role_policy_document, path):
     def create_from_cloudformation_json(cls, resource_name, cloudformation_json, region_name):
         properties = cloudformation_json['Properties']
 
+        path = properties['Path'] if 'Path' in properties else '/'
         role = iam_backend.create_role(
             role_name=resource_name,
             assume_role_policy_document=properties['AssumeRolePolicyDocument'],
-            path=properties['Path'],
```
terrycain: Replace with `properties.get('Path', '/')`

william-richard: Doi yeah that's way smarter.

terrycain:
:D

PR #1383, opened 2017-12-08
Fix co-operative inheritance for ec2 model

c3st7n:
Small change to fix co-operative inheritance model for ec2 model.

I was working on a private fork for work and spent a lot of time trying to get my new features to work before I realised that NatGatewayBackend wasn't calling super() and breaking the chain.

Hopefully this will save someone some time in the future.

### DIFF
--- a/moto/ec2/models.py
+++ b/moto/ec2/models.py
@@ -3722,6 +3722,7 @@ class NatGateway(object):
 class NatGatewayBackend(object):
     def __init__(self):
         self.nat_gateways = {}
+        super(NatGatewayBackend, self).__init__()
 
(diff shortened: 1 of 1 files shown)


terrycain:
Can we get a test case on the nat gateway, as we should of been aware of this

c3st7n:
Might be being ignorant here, but there is already a test suite for NatGateway which I assume is passing. My issue came because I added another backend and it's constructor was not called because I added it after NatGatewayBackend.

I can't think of a quick way to test this specific use case.

terrycain:
Ahh good point

PR #1384, opened 2017-12-08
Fix linting errors

c3st7n:
Use latest version of flake8 (not sure exactly why this needs pinning - newer versions are catching real issues) and fix the violations it mentions.

The bare exceptions will catch things such as keyboard interrupts which is bad behaviour.


c3st7n:
Closing as unrelated change was included by mistake

PR #1385, opened 2017-12-08
Bump flake8 version and handle new lint errors

c3st7n:
Use latest version of flake8 (not sure exactly why this needs pinning - newer versions are catching real issues) and fix the violations it mentions.

The bare exceptions will catch things such as keyboard interrupts which is bad behaviour.


terrycain:
Yeah the only downside of unpinning it would be say they added another check, then next time any build ran, it would start erroring. We should just connect pyup.io

terrycain:
Great work though :D

PR #1389, opened 2017-12-10
Add create change set endpoint

alexmorken:


moto/cloudformation/responses.py
```diff
@@ -77,6 +77,50 @@ def create_stack(self):
...
+        stack_name = self._get_param('StackName')
+        change_set_name = self._get_param('ChangeSetName')
+        stack_body = self._get_param('TemplateBody')
+        template_url = self._get_param('TemplateURL')
+        role_arn = self._get_param('RoleARN')
+        update_or_create = self._get_param('ChangeSetType', 'CREATE')
+        parameters_list = self._get_list_prefix("Parameters.member")
+        tags = dict((item['key'], item['value'])
+                    for item in self._get_list_prefix("Tags.member"))
+        # Hack dict-comprehension
```
terrycain: There a reason why we cant use regular dictionary comprehensions?

moto/cloudformation/responses.py
```diff
@@ -250,6 +294,17 @@ def list_exports(self):
 </UpdateStackResponse>
 """
 
+CREATE_CHANGE_SET_RESPONSE_TEMPLATE = """<CreateStackResponse>
+  <CreateChangeSetResult>
+    <Id>{{change_set_id}}</Id>
+    <StackId>{{ stack_id }}</StackId>
+  </CreateChangeSetResult>
+  <ResponseMetadata>
+    <RequestId>b9b4b068-3a41-11e5-94eb-example</RequestId>
```
terrycain: Look [here](https://github.com/spulec/moto/blob/81f3cbb54880bdf68fd6f52b9aa30ea25a4b7bd9/moto/cloudwatch/responses.py#L18) and [here](https://github.com/spulec/moto/blob/81f3cbb54880bdf68fd6f52b9aa30ea25a4b7bd9/moto/cloudwatch/responses.py#L195) for returning a proper request ID. If your up to it, see if the other methods can be done too

moto/cloudformation/utils.py
```diff
@@ -10,6 +10,11 @@ def generate_stack_id(stack_name):
     return "arn:aws:cloudformation:us-east-1:123456789:stack/{0}/{1}".format(stack_name, random_id)
 
 
+def generate_changeset_id(changeset_name):
+    random_id = uuid.uuid4()
+    return 'arn:aws:cloudformation:us-east-1:123456789:changeSet/{0}/{1}'.format(changeset_name, random_id)
```
terrycain: Cloudwatch should have region specific ARNs right?

terrycain:
Great contribution, welcome to Moto. PR looks good, but I think theres a few places where we could be a bit more flexible.

moto/cloudformation/responses.py
alexmorken: I was just following convention in the file.  I would be happy to change it to a real dict comprehension.

moto/cloudformation/utils.py
alexmorken: Yes - I was again just copying convention used in `generate_stack_id` above.  Trying to be predictable.

moto/cloudformation/responses.py
terrycain: Yeah swap it out, we dont need to support py26 anymore and py27 supports it if I remember correctly.

moto/cloudformation/utils.py
terrycain: Fair enough, if your up for fixing all of them to use regions, by all means ;-)

moto/cloudformation/responses.py
alexmorken: Cool - thanks - will take a look at it!

alexmorken:
I will start working on execute_change_set and should hopefully have a PR up for that in a few.

moto/cloudformation/responses.py
alexmorken: Yea - that's what I was thinking - probably some 2.6 cruft.

alexmorken:
Everything should addressed per your feedback.

alexmorken:
I made another small change. I set the status of the stack to `REVIEW_IN_PROGRESS` if `create_change_set` created the stack initially.

alexmorken:
Ok - everything is looking good at this point and also a bit more true to how AWS acts, too.

terrycain:
:D Looks great

PR #1390, opened 2017-12-10
Fixes S3 duplicate bucket

### LINKED ISSUE #1371: no error or warning when creating s3 bucket that already exists in standalone server
Python 3.6.3
boto3==1.4.8
moto==1.1.25

I know I should be using the decorators and not the standalone server, but this still seems like a bug. After starting the standalone server: 

```
moto_server s3
```

I run this code:

```python
import boto3
boto3.setup_default_session()
s3 = boto3.client(service_name="s3", region_name="us-west-2", endpoint_url="http://localhost:5000")
result = s3.create_bucket(Bucket='mybucket')
print(result)
result = s3.create_bucket(Bucket='mybucket')
print(result)
```
I expect that last line to return a stack trace with the message `BucketAlreadyOwnedByYou: An error occurred (BucketAlreadyOwnedByYou) when calling the CreateBucket operation: Your previous request to create the named bucket succeeded and you already own it.` That's what I get in actual S3 when trying to create a bucket I have already created.

But instead it succeeds without comment just like the first call to `create_bucket`.

This is what it outputs:

```
{'ResponseMetadata': {'HTTPStatusCode': 200, 'HTTPHeaders': {'content-type': 'text/html; charset=utf-8', 'content-length': '174', 'server': 'Werkzeug/0.12.2 Python/3.6.3', 'date': 'Mon, 04 Dec 2017 04:24:04 GMT'}, 'RetryAttempts': 0}}
{'ResponseMetadata': {'HTTPStatusCode': 200, 'HTTPHeaders': {'content-type': 'text/html; charset=utf-8', 'content-length': '174', 'server': 'Werkzeug/0.12.2 Python/3.6.3', 'date': 'Mon, 04 Dec 2017 04:24:04 GMT'}, 'RetryAttempts': 0}}
```

If I upload a file to the bucket after the first bucket creation, that file is still there after the second creation. So it's not overwriting the original bucket, it's just not telling me that bucket already exists.

I'm trying to test code that relies on the correct behavior, so...

terrycain:
Fixes #1371

Currently tests are failing, will fix later.


tests/test_s3/test_s3.py
```diff
@@ -1943,6 +1943,17 @@ def test_get_stream_gzipped():
     assert res == payload
 
 
+@mock_s3
+def test_duplicated_buckets():
+    s3 = boto3.client('s3', region_name='eu-west-1')
+    s3.create_bucket(Bucket='mybucket')
+
+    try:
+        s3.create_bucket(Bucket='mybucket')
```
c3st7n: To actually get the bucket to be created in the non-default region, you need to set the location constraint:
 s3.create_bucket(Bucket='mybucket', CreateBucketConfiguration={'LocationConstraint': 'eu-west-1'})

bblommers:
Superceded by #4257

Issue #1396, opened 2017-12-14
images.filter does not work for Owners='self'


afgane:
boto3 supports filtering of images owned by the current user as follows (https://stackoverflow.com/questions/40213023/boto3-get-ec2-images-owned-by-me): 
```
boto3conn.images.filter(Owners=['self'])
``` 

However, that does not work with moto:
```
import boto3
s = boto3.session.Session(region_name='us-east-1', aws_access_key_id='akey', aws_secret_access_key='skey')
rsrc = s.resource('ec2', region_name='us-east-1', use_ssl=True, verify=True, endpoint_url='http://localhost:5000')
inst = rsrc.create_instances(ImageId='ami-f4cf1d8d', InstanceType='t2.large', MinCount=1, MaxCount=1)[0]
inst.create_image(Name='ea-ci')
list(rsrc.images.filter(Owners=['self']))  # returns []
```

(affects https://github.com/spulec/moto/issues/1357)

PR #1398, opened 2017-12-16
Added support for filtering AMIs by self

nuwang:
Closes: https://github.com/spulec/moto/issues/1396


terrycain:
Looks good :+1:

PR #1400, opened 2017-12-17
mock_xray_client: don't swallow function's return value

### LINKED ISSUE #1399: mock_xray_client cannot be used as a context manager
PR #1255 added support for `aws_xray_sdk` which is great.
But there is a problem with it: `moto.mock_xray_client` is *only* a function decorator, and unlike all other `mock_*` methods it cannot be used as a context manager or directly with `start()`...`stop()`.
As a result, it is not possible to write a `py.test` fixture which would add support for mocking `xray_client`.

Also, `mock_xray_client` does not return the result of the function it decorates. Given it is meant to be used to decorate test functions it is most likely not a big issue, but I think it is still worth fixing.

I will prepare a PR for the return value issue soon.
Also I am thinking about refactoring `mock_xray_client` to base it on the existing infrastructure (`BaseBackend`, `base_decorator`) but am not yet enough familiar with `moto` internals to be sure which would be the best way to implement it.

Installed version: `moto-ext==1.1.25`
The problem seemingly persists in current `master` branch.

MarSoft:
fixes #1399
this won't help with fixtures though

### DIFF
--- a/moto/xray/mock_client.py
+++ b/moto/xray/mock_client.py
@@ -51,7 +51,7 @@ def mock_xray_client(f):
         aws_xray_sdk.core.xray_recorder._emitter = MockEmitter()
 
         try:
-            f(*args, **kwargs)
+            return f(*args, **kwargs)
         finally:
 
             if old_xray_context_var is None:


JackDanger:
Great work, @MarSoft! And welcome to Moto!

PR #1406, opened 2017-12-20
Add batch_get_image support for ECR

mbjerkness:


moto/ecr/models.py
```diff
@@ -245,6 +256,38 @@ def put_image(self, repository_name, image_manifest, image_tag):
...
+            response = set()
+            for image_id in image_ids:
+                found = False
+                for image in repository.images:
+                    if (('imageDigest' in image_id and image.get_image_digest() == image_id['imageDigest']) or
+                            ('imageTag' in image_id and image.image_tag == image_id['imageTag'])):
+                        found = True
+                        response.add(image)
+                if not found:
+                    image_id_representation = "{imageDigest:'%s', imageTag:'%s'}" % (
```
JackDanger: What do you think about using `json.dumps` here like in ecr/responses.py for crafting the JSON string? It might be easier to follow for future developers (I didn't immediately realize this was JSON at first)

mbjerkness:
@JackDanger I updated the code per your comment. You comment actually highlighted a larger issue that I added some tests for. The code wasn't responding with the right errors in some cases. It should be now.

JackDanger:
Nice! Thanks, @mbjerkness

PR #1407, opened 2017-12-21
FakeStack update to support both JSON and YAML templates

nimbusscale:
Change needed to allow update to a FakeStack to accept both JSON and YAML templates.


nimbusscale:
Will validate tests locally and append PR when all tests pass.

JackDanger:
Nice work, @nimbusscale! This is great. Welcome to Moto :)

Issue #1288, opened 2017-10-22
Instance created by Autoscaling does not have subnet


mishraawake:
When we create instance through autoscaling, the subnet of instance is not specified, It should be derived by autoscaling property "VPCZoneIdentifier"

captainkerk:
Yeah, looks like we do ignore that parameter when launching new instances. Also, it looks like I've mistakenly set VPCZoneIdentifier to a VPC's ID, rather than a subnet in a few tests.

I'll try to fix this in the next few days.

captainkerk:
Sorry about the delay, I have an update:

Turns out the actual autoscaling spec requires either Availability Zones or VPC Zone Identifier or both to be specified on ASG create, which we do not enforce.

The ideal way to fix this would be to enforce that requirement, rather than just tacking on creating an instance with a subnet if specified. It's been a bit more complicated to tack that on, and most of our tests don't create groups with AZs or VPC Zone Identifiers. 

I'll have a pull request soon.

captainkerk:
Sorry @spulec, can you re-open this issue? I haven't quite fixed it, #1413 was just the first part. I'll have another PR with the remaining changes shortly.

spulec:
Ah, got it. Thanks

spulec:
Fixed with https://github.com/spulec/moto/pull/1985

PR #1413, opened 2017-12-27
AWS API raises an exception if both AZ and VPCZoneIdentifier params a…

captainkerk:
In trying to fix https://github.com/spulec/moto/issues/1288, I realized that the AWS API requires either `AvailabilityZones` or `VPCZoneIdentifier`.

I think the mock_autoscaling service should also have this requirement, as boto calls without either fail.

### DIFF
--- a/tests/test_autoscaling/test_autoscaling.py
+++ b/tests/test_autoscaling/test_autoscaling.py
@@ -11,13 +11,13 @@ import sure  # noqa
 from moto import mock_autoscaling, mock_ec2_deprecated, mock_elb_deprecated, mock_elb, mock_autoscaling_deprecated, mock_ec2
 from tests.helpers import requires_boto_gte
 
-from utils import setup_networking
+from utils import setup_networking, setup_networking_deprecated
 
 
 @mock_autoscaling_deprecated
 @mock_elb_deprecated
 def test_create_autoscaling_group():
-    mocked_networking = setup_networking()
+    mocked_networking = setup_networking_deprecated()
     elb_conn = boto.ec2.elb.connect_to_region('us-east-1')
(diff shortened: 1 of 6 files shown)


tests/test_autoscaling/test_autoscaling.py
```diff
@@ -105,7 +111,7 @@ def test_create_autoscaling_groups_defaults():
     # Defaults
     list(group.availability_zones).should.equal([])
     group.desired_capacity.should.equal(2)
-    group.vpc_zone_identifier.should.equal('')
+    group.vpc_zone_identifier.should.equal(mocked_networking['subnet1'])
```
JackDanger: Nice!

captainkerk:
Oops. Tests were passing locally, I'll investigate the travis failures.

captainkerk:
Ok, the test suite is passing again. Only a few tests of the server suite were failing, the same tests using the running with server=false were passing.

In these last two commits, I created a `setup_networking_deprecated()` method to do the same thing as setup_networking, but use the older boto calls. I didn't think it was a problem with the way I was wrapping the `setup_networking()` with a mock_ec2, but when the test calling `setup_networking()` was wrapped in a mock_autoscaling_deprecated decorator, I found that the boto3 calls contained in `setup_networking()` actually went out to the AWS API. It seems like 'deprecated' and 'not' decorators cannot be intermingled? I.e. If you use the deprecated variant first, the other kind is ignored?

Is that expected?

captainkerk:
Heyy @JackDanger sorry to be a bother, anything I could do to help move this along? 

I know I'm touching pretty much every test, but it seemed unavoidable.

spulec:
Thanks!

PR #1416, opened 2017-12-28
describe_regions: handle region-names parameter

goraxe:
I have been testing terraform against moto, this PR helps fix the terraform construct

```
data "aws_region" {
   current = true
}
```

That code uses the RegionNames parameter to limit the return to just the current region


JackDanger:
Excellent. Thanks, @goraxe!

PR #1417, opened 2017-12-28
Cloudformation vpc add tags

goraxe:
I found when creating a vpc via cloudformation tags where not being applied.  This fixes that.


JackDanger:
Nice fix, @goraxe! This'll help every developer who uses cloudformation, we're grateful for your contribution. Welcome to Moto!

PR #1418, opened 2018-01-02
SNS: do not duplicate subscriptions

stillwaiting:
Hi.

This is apparently is a bug. If I try to create 2 subscriptions between a topic and an endpoint, Amazon will create only a single one and on the 2nd call will return the ARN of the 1st one.

Though, I couldn't any description of it in the docs (event [stackoverflow](https://stackoverflow.com/questions/48016591/amazon-sns-multiple-subscriptions-of-the-same-endpoint) didn't help.) But the experiments show that this is the behavior they have.

### DIFF
--- a/moto/sns/models.py
+++ b/moto/sns/models.py
@@ -247,10 +247,20 @@ class SNSBackend(BaseBackend):
         setattr(topic, attribute_name, attribute_value)
 
     def subscribe(self, topic_arn, endpoint, protocol):
+        # AWS doesn't create duplicates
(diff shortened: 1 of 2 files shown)


bgvozdev:
img

bgvozdev:
Dear maintainers, this issue keeps biting us in the arse. Any chance this could be reviewed/merged soonish?

spulec:
Thanks for the contribution and welcome to the community!

PR #1422, opened 2018-01-08
Change name of 'state' attribute of 'FakeAlarm' CloudWatch model to '…

godfried:
…state_value'. This ensures that the 'StateValue' returned by 'describe_alarms' is correct.

The 'DESCRIBE_ALARMS_TEMPLATE' response template references a 'state_value' attribute on the 'FakeAlarm' model which does not exist; it is named 'state'.
This commit updates the attribute to be called 'state_value', in-line with the naming convention used elsewhere.

### DIFF
--- a/moto/cloudwatch/models.py
+++ b/moto/cloudwatch/models.py
@@ -74,18 +74,18 @@ class FakeAlarm(BaseModel):
 
         self.state_reason = ''
         self.state_reason_data = '{}'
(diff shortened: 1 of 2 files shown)


godfried:
Any chance of getting this looked at?

spulec:
Thanks and welcome to the Moto community!

godfried:
Thanks :)

PR #1425, opened 2018-01-11
Implement Fn::GetAZs function in CloudFormation

whummer:
Implement `Fn::GetAZs` function in CloudFormation. This PR also adds a simple test to cover the new functionality.


spulec:
Thanks!

PR #1428, opened 2018-01-14
Add logs exceptions

toshitanian:
As logs module has not raises boto fashion exceptions, we could not test the resource coflict. I therefore added exceptions.


spulec:
Thank you!

PR #1430, opened 2018-01-14
Url decode x-amz-copy-source for S3 copyObject requests

ciprianaradulescu:
As per [https://docs.aws.amazon.com/AmazonS3/latest/API/RESTObjectCOPY.html](https://docs.aws.amazon.com/AmazonS3/latest/API/RESTObjectCOPY.html) the x-amz-copy-source header should be url encoded when sent to the s3 API.
Currently, if the header is url encoded the backend throws an error. I've url decoded the header value before parsing it any further.

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,0 +1,7 @@
+[bumpversion]
+current_version = 1.2.0
+
+[bumpversion:file:setup.py]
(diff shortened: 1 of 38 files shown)


spulec:
Thanks and welcome to the project!

PR #1442, opened 2018-01-21
importing sure in tests that require it

JackDanger:


dbfr3qs:
Hey @JackDanger, tests for this (and for every PR submitted as of roughly a week ago) will fail due to an update in botocore. I've added a fix in #1432

JackDanger:
@dbfr3qs Nice work! I spent some time trying to track that error down and came up empty. Thanks for fixing it for everybody!

dbfr3qs:
It's worth merging this with the current version of master and running the build again.

Issue #1365, opened 2017-11-29
createdDate should be int(time.time()) for apigateway


raymondhardy:
## Reporting Bugs

When using local stack that relies on moto to generate api data the response is not the same as aws.

Moto results for create-rest-api

{
    "id": "823788A-Z5A-Z1",
    "name": "My First API",
    "description": "This is my first API",
    "createdDate": "2017-11-29T22:59:03.320Z"
}

AWS results

aws apigateway create-rest-api --name 'My First API' --description 'This is my first API'
{
    "id": "yd8wlf5aw1",
    "name": "My First API",
    "description": "This is my first API",
    "createdDate": 1511983669
}

This is causing issues in validation of the createdDate value.

It should be updated to use int(time.time()) instead of current datetime object

terrycain:
Feel free to create a PR

raymondhardynike:
https://github.com/spulec/moto/pull/1457

spulec:
Closed with #1457

PR #1443, opened 2018-01-23
API Gateway‎ createdDate #1365

fdfk:


dbfr3qs:
It's worth merging this with the current version of master and running the build again.

PR #1448, opened 2018-01-26
Change `'image'` for `'volume'` when getting tags

grahamlyons:
The AWS docs say that: "Currently, the resource types that support tagging on creation are instance and volume." Calling `create_volume` and passing `image` as the resource type in tag specifications causes an `InvalidParameterValue` error.


dbfr3qs:
@grahamlyons nice! Thanks for fixing my mistake 👍

grahamlyons:
No worries, @dbfr3qs.

spulec:
Thanks!

PR #1450, opened 2018-01-26
Add Fn::GetAZs

RasmusWernerLarsen:
Implement Fn::GetAZs cloudformation function.


spulec:
I'm going to close this in favor of https://github.com/spulec/moto/pull/1425 due to additional tests there. If we can find a generic way (using boto/boto3?) to add the specific zones here on top of that PR, that would be great.

PR #1451, opened 2018-01-28
Redshift: Add Cross Region Snapshot Functionality

captainkerk:
This adds three methods to the boto3 redshift client, dealing with cross-region snapshot copies.


spulec:
Awesome! thanks

PR #1452, opened 2018-01-28
Redshift: Add Copy Grant Functionality

captainkerk:
This builds off of https://github.com/spulec/moto/pull/1451. That PR should also probably be reviewed/merged first.

To copy KMS-enabled snapshots cross region, you need to create a copy grant allowing the redshift service to use a KMS key in the destination region. This PR is adding that functionality. 

As you can see in the tests, there's no validation around the given kms key when creating these snapshot grants. It didn't look too bad to wire that up, however that mixed with what I have here was rather large, so I'll add the validation later.


spulec:
Thank you!

PR #1455, opened 2018-01-29
Snapshot search bug

grahamlyons:
When I tried to search for a long list of snapshots by ID I got the following error:
```
...
    moto/ec2/models.py line 1957 in describe_snapshots
      unknown_ids = set(snapshot_ids) - set(matches)
   TypeError: unhashable type: 'dict'
```

This fixes that and preserves the original correct behaviour.


spulec:
thanks!

Issue #1454, opened 2018-01-29
Don't create volumes for AMIs


grahamlyons:
These lines in the EC2 model create a volume for each AMI: https://github.com/spulec/moto/blob/4520cd930ff265834689a3d945a172b375133be5/moto/ec2/models.py#L1088-L1091

I don't think this behaviour is correct.

When you create an AMI yourself you don't get an EBS volume, although you do get a snapshot; when you run an instance _then_ an EBS volume is created:
```python
>>> import boto3
>>> ec2 = boto3.client('ec2', region_name='eu-west-1')
>>>  v = ec2.describe_volumes()
>>> len(v['Volumes'])
0
>>> i = ec2.run_instances(ImageId='ami-1e749f67', InstanceType='t2.nano', MaxCount=1, MinCount=1)
>>>  v = ec2.describe_volumes()
>>> len(v['Volumes'])
1
>>> s = ec2.describe_snapshots(OwnerIds=['self'])
>>> len(s['Snapshots'])
0
>>>  ec2.create_image(InstanceId=i['Instances'][0]['InstanceId'], Name='moto-test')
>>> s = ec2.describe_snapshots(OwnerIds=['self'])
>>> len(s['Snapshots'])
1
```

Moto has a list of public AMI data which is used populate some `Ami` objects in each region's backend. This currently results in the creation of snapshots and volumes, of which I think the latter is incorrect, and the volumes are created without an availability zone, which is definitely wrong.

Having added a filter for volumes by availability zone (in here https://github.com/spulec/moto/blob/4520cd930ff265834689a3d945a172b375133be5/moto/ec2/models.py#L1813) this issue surfaced because the ones created by the AMI data threw an exception when trying to access their `zone.name` (`zone` is `None`).

Each snapshot has a volume but in the case of public snapshots or where the volume has been deleted that is not accessible.

I propose that the `Ami` class deletes the volume used to create the block device snapshot once it's done with it. It should also pass through the `owner_id` so that it matches the one passed to the AMI.

spulec:
Closed with https://github.com/spulec/moto/pull/1456

PR #1456, opened 2018-01-29
Don't create volumes for AMIs

grahamlyons:
This addresses #1454, and also adds the `owner_id` to the `Snapshot` object.

The test is actually dependent on the fix in #1455 so this pull request is likely to be failing until that gets merged and I rebase master onto this branch.


JackDanger:
Thanks for the fix,. @grahamlyons! Apologies for the huge delay.

PR #1457, opened 2018-01-29
fixes apigateway timestamp to match aws

raymondhardynike:
This is a PR for the issue i opened a while back
https://github.com/spulec/moto/issues/1365

When using local stack that relies on moto to generate api data the response is not the same as aws.

Moto results for create-rest-api

{
"id": "823788A-Z5A-Z1",
"name": "My First API",
"description": "This is my first API",
"createdDate": "2017-11-29T22:59:03.320Z"
}

AWS results

aws apigateway create-rest-api --name 'My First API' --description 'This is my first API'
{
"id": "yd8wlf5aw1",
"name": "My First API",
"description": "This is my first API",
"createdDate": 1511983669
}

This is causing issues in validation of the createdDate value.

It should be updated to use int(time.time()) instead of current datetime object

### DIFF
--- a/moto/apigateway/models.py
+++ b/moto/apigateway/models.py
@@ -1,12 +1,11 @@
 from __future__ import absolute_import
 from __future__ import unicode_literals
 
-import datetime
 import requests
+import time
 
 from moto.packages.responses import responses
 from moto.core import BaseBackend, BaseModel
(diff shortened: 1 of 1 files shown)


spulec:
Thanks!

PR #1459, opened 2018-01-30
Fix the S3 HEAD response body

okomestudio:
This PR is a minor fix for the S3 bucket HEAD response to return an empty body when the bucket does not exist, so that `botocore` would not complain about being unable to parse the body into XML. The real AWS API returns an empty body, so this is the correct mock response.

### DIFF
--- a/moto/s3/responses.py
+++ b/moto/s3/responses.py
@@ -172,7 +172,7 @@ class ResponseObject(_TemplateEnvironmentMixin):
             # HEAD (which the real API responds with), and instead
             # raises NoSuchBucket, leading to inconsistency in
             # error response between real and mocked responses.
-            return 404, {}, "Not Found"
+            return 404, {}, ""
         return 200, {}, ""
 
     def _bucket_response_get(self, bucket_name, querystring, headers):


okomestudio:
This is really a minor issue, preventing an error like this:

```
DEBUG:botocore.parsers:Exception caught when parsing error response body:
  File "/home/taro/.virtualenvs/moto/local/lib/python3.4/site-packages/botocore/parsers.py", line 385, in _parse_xml_string_to_dom
    parser.feed(xml_string)
  File "<string>", line None
xml.etree.ElementTree.ParseError: syntax error: line 1, column 0

  File "/home/taro/.virtualenvs/moto/local/lib/python3.4/site-packages/botocore/parsers.py", line 750, in _do_error_parse
    return self._parse_error_from_body(response)
  File "/home/taro/.virtualenvs/moto/local/lib/python3.4/site-packages/botocore/parsers.py", line 772, in _parse_error_from_body
    root = self._parse_xml_string_to_dom(xml_contents)
  File "/home/taro/.virtualenvs/moto/local/lib/python3.4/site-packages/botocore/parsers.py", line 390, in _parse_xml_string_to_dom
    "invalid XML received:\n%s" % (e, xml_string))
botocore.parsers.ResponseParserError: Unable to parse response (syntax error: line 1, column 0), invalid XML received:
b'Not Found'
```

from appearing when investigating (other) issues in logging message.

spulec:
Thanks!

PR #1460, opened 2018-01-31
Support both virtual-hosted and path-style URLs for region name parsing

okomestudio:
This PR adds a little more robust helper function for parsing a region name from an S3 bucket URL. In particular, it explicitly supports both virtual-hosted and path-style URLs, and also allows a possibility (which the real API via `botocore` appears to sometimes use) where region appears in the form `s3.<region>` as opposed to `s3-<region>` (which is mentioned by the AWS API documentation).


spulec:
Thank you!

PR #1464, opened 2018-02-02
Make putparameter behaviour consistent with real endpoint

ramox-:
This patch makes SSM PutParameter behaviour consistent with the real endpoint in SSM.

It will return a response containing a 'Version'-key on successful calls and can now also return a ParameterAlreadyExists error if you attempt to Put a parameter that already exists (and overwrite is set to False).

Also reworked the tests to account for this functionality.


JackDanger:
This is extremely helpful, @ramox! Thanks so much for the fix and welcome to Moto!
Apologies for the huge delay, we're looking to improve our process around code reviews and releases.

PR #1467, opened 2018-02-09
XML escaping in EMR responses

ashb:
Some of our steps include a literal `&` and this caused the EMR list_steps call to fail due to invalid xml.

This adds a fix to the template and tests.

(Sorry, this was edited via Github web-ui so is done as two commits. I can squash if desired.)


ashb:
Test failures are now just in the (unchanged) lambda module

JackDanger:
Thanks for finding this edge case, @ashb! I'm sure any of us could have stumbled on it and you've just saved us from a big headache. Welcome to Moto!

PR #1468, opened 2018-02-09
Add in filtering of EBS volumes on availability-zone

keirbadger:
Requires https://github.com/spulec/moto/pull/1456 before merge


JackDanger:
This is super helpful, @keirbadger! Thanks for the fix and welcome to Moto!

Issue #1469, opened 2018-02-09
filter_expression and projection_expression are evaluated in the wrong order in dynamodb2/models.py


eastlondoner:
E.g. If you project only item "Id" but you want to filter on "Name" then Moto will return nothing because it does the projection first (removing the Name field) then evaluates the filter.

eastlondoner:
PR here https://github.com/spulec/moto/pull/1470

PR #1470, opened 2018-02-09
Put projection after filter so that filters can act on fields that ar…

eastlondoner:
…e not projected


eastlondoner:
Issue here: https://github.com/spulec/moto/issues/1469

JackDanger:
Nice fix,. @eastlondoner! Welcome to Moto!

Issue #1472, opened 2018-02-09
Downgrade Jinja2 requirement to 2.7.3


acrewdson:
As a result of issue https://github.com/spulec/moto/issues/728, the Jinja2 requirement was bumped to `>=2.8`, even though the missing feature in question, `lstrip_blocks`, was added in Jinja 2.7 ('Added finer grained support for stripping whitespace on the left side of blocks', from  http://jinja.pocoo.org/docs/2.10/changelog/#version-2-7).

I know this might be an unusual request, but I believe that pushing the `Jinja2` requirement in `setup.py` back down to `>=2.7.3` would not break anything, and it would be really helpful to projects that are still on that old version of Jinja2 (to avoid dependency resolution conflicts).

If you'd be willing to consider this, I can make a PR. Thank you.

JackDanger:
I think that’s a very disciplined approach, @acrewdson
A PR would be appreciated!

acrewdson:
@JackDanger sounds good, will do 👍

Issue #728, opened 2016-10-11
TypeError: __init__() got an unexpected keyword argument 'lstrip_blocks'


anentropic:
I have decorated my tests with `@mock_s3`

I get the following error when running:

```
Traceback (most recent call last):
  File "/Users/me/.pyenv/versions/2.7.11/lib/python2.7/unittest/case.py", line 329, in run
    testMethod()
  File "/myproject/VENV/lib/python2.7/site-packages/moto/core/models.py", line 71, in wrapper
    result = func(*args, **kwargs)
  File "/myproject/proj/../proj/tests/integration/engine/management/commands/test_archive_deleted_chat_messages.py", line 95, in test_archive_exception_log_and_continue
    self.cmd.handle(write=True, noinput=True)
  File "/myproject/proj/engine/management/commands/__init__.py", line 335, in handle
    super(LiveUpdateCommand, self).handle(*args, **kwargs)
  File "/myproject/proj/engine/management/commands/__init__.py", line 165, in handle
    self._prepare_args(*args, **kwargs)
  File "/myproject/proj/engine/management/commands/archive_deleted_chat_messages.py", line 219, in _prepare_args
    self.bucket = conn.create_bucket(self.bucket_name)
  File "/myproject/VENV/lib/python2.7/site-packages/boto/s3/connection.py", line 616, in create_bucket
    data=data)
  File "/myproject/VENV/lib/python2.7/site-packages/boto/s3/connection.py", line 668, in make_request
    retry_handler=retry_handler
  File "/myproject/VENV/lib/python2.7/site-packages/boto/connection.py", line 1071, in make_request
    retry_handler=retry_handler)
  File "/myproject/VENV/lib/python2.7/site-packages/boto/connection.py", line 944, in _mexe
    response = connection.getresponse()
  File "/Users/me/.pyenv/versions/2.7.11/lib/python2.7/httplib.py", line 1133, in getresponse
    response = self.response_class(*args, **kwds)
  File "/myproject/VENV/lib/python2.7/site-packages/boto/connection.py", line 390, in __init__
    http_client.HTTPResponse.__init__(self, *args, **kwargs)
  File "/Users/me/.pyenv/versions/2.7.11/lib/python2.7/httplib.py", line 390, in __init__
    self.fp = sock.makefile('rb', 0)
  File "/myproject/VENV/lib/python2.7/site-packages/httpretty/core.py", line 332, in makefile
    self._entry.fill_filekind(self.fd)
[...]

PR #1474, opened 2018-02-09
Allow slightly older Jinja2 dep for project compatability

acrewdson:
(see also https://github.com/spulec/moto/issues/1472)

In commit 4157abe8de154b7ff00c79d6427c0d774889da3f, the Jinja2 dep in `setup.py` was changed from being unpinned to requiring version 2.8 or newer. That change was made in response to issue https://github.com/spulec/moto/issues/728, which pointed out that the use of `lstrip_blocks` needed to be supported. Support for that option was added in Jinja 2.7, so it seems like the best option here (unless 2.8 is truly required by moto) would be to allow Jinja 2.7.3 and newer -- preventing dependency resolution conflicts in Python projects that pin Jinja2 2.7.3 but also need to use moto.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -6,7 +6,7 @@ import sys
 
 
 install_requires = [
-    "Jinja2>=2.8",
+    "Jinja2>=2.7.3",
     "boto>=2.36.0",
     "boto3>=1.2.1",
     "botocore>=1.7.12",


JackDanger:
Great idea. Thanks, @acrewdson! And welcome to Moto!

acrewdson:
@JackDanger thanks for merging!

PR #1476, opened 2018-02-13
Use the TemplateBody by default it exist

cliveli:
serverless cli use the PUT + TemplateBody + TemplateURL to upload cloudformation template

This fix try to get the TemplateBody from the request first, if not exist, fetch the one from s3 bucket

And it fix the issue reported to localstack here
[https://github.com/localstack/localstack/issues/580](https://github.com/localstack/localstack/issues/580)


spulec:
Thanks!

Issue #1477, opened 2018-02-16
Implement basic OpsWorks App and Deployments mocks 


lleontop:
Moto is missing some basic mocks for OpsWorks create_app, describe_apps, create_deployment, describe_deployments. I am planning to implement these mocks.

lleontop:
The basic App endpoints are done in PR #1481 but deployments are a bit tricker. My main thought/problem with deployments is how we will expose fields like CompletedAt and Duration, Status and maybe IamUserArn(we can hardcode this to a user like the account number). If we want the end user to have control over success/failure/timeout statuses and plan his tests accordingly then we should expose this somehow. 

Also, the different commands in stack/app level are a bit complicated but should be manageable.

Anybody has any ideas/suggestions on this?

spulec:
Going to close based on previous work that is done. Feel free to open additional issues for specific endpoints.

PR #1481, opened 2018-02-19
Add opsworks app mocks

lleontop:
Add the mocks of OpsWork create_app and describe_apps calls. This is part of #1477


JackDanger:
This is hugely good. Thanks for the great work @lleontop!

PR #1482, opened 2018-02-19
Change lambda backend to support docker-py dependency changes

lleontop:
There are breaking changes in docker-py 3.0.0 as depicted [here](https://github.com/docker/docker-py/releases). We were using 2.7.0 and on Feb 1st Travis started picking up the 3.x.x version so the builds are failing. We can either limit the version of docker dependency in setup.py to <3.0.0 or support the breaking changes and upgrade docker dependency. I think the latter would be preferable since the changes are relatively small and lambda tests are green.

### DIFF
--- a/moto/awslambda/models.py
+++ b/moto/awslambda/models.py
@@ -104,7 +104,7 @@ class _DockerDataVolumeContext:
 
             # It doesn't exist so we need to create it
             self._vol_ref.volume = self._lambda_func.docker_client.volumes.create(self._lambda_func.code_sha_256)
-            container = self._lambda_func.docker_client.containers.run('alpine', 'sleep 100', volumes={self.name: '/tmp/data'}, detach=True)
+            container = self._lambda_func.docker_client.containers.run('alpine', 'sleep 100', volumes={self.name: {'bind': '/tmp/data', 'mode': 'rw'}}, detach=True)
             try:
                 tar_bytes = zip2tar(self._lambda_func.code_bytes)
                 container.put_archive('/tmp/data', tar_bytes)
@@ -309,7 +309,7 @@ class LambdaFunction(BaseModel):
                 finally:
                     if container:
                         try:
-                            exit_code = container.wait(timeout=300)
+                            exit_code = container.wait(timeout=300)['StatusCode']
                         except requests.exceptions.ReadTimeout:
                             exit_code = -1
                             container.stop()


lleontop:
Hi @JackDanger any chance you can have a look at this. This should unblock broken builds in PRs.
Thanks!

spulec:
Thank you!

2rs2ts:
Speifically, we get this kind of error:

```
  File "/home/admin/.virtualenvs/stars/local/lib/python2.7/site-packages/moto/awslambda/models.py", line 313, in _invoke_lambda
    except requests.exceptions.ReadTimeout:
  File "/home/admin/.virtualenvs/stars/local/lib/python2.7/site-packages/moto/awslambda/models.py", line 121, in __exit__
    self._vol_ref.volume.remove()
  File "/home/admin/.virtualenvs/stars/local/lib/python2.7/site-packages/docker/models/volumes.py", line 25, in remove
    return self.client.api.remove_volume(self.id, force=force)
  File "/home/admin/.virtualenvs/stars/local/lib/python2.7/site-packages/docker/utils/decorators.py", line 34, in wrapper
    return f(self, *args, **kwargs)
  File "/home/admin/.virtualenvs/stars/local/lib/python2.7/site-packages/docker/api/volume.py", line 165, in remove_volume
    self._raise_for_status(resp)
  File "/home/admin/.virtualenvs/stars/local/lib/python2.7/site-packages/docker/api/client.py", line 222, in _raise_for_status
    raise create_api_error_from_http_exception(e)
  File "/home/admin/.virtualenvs/stars/local/lib/python2.7/site-packages/docker/errors.py", line 31, in create_api_error_from_http_exception
    raise cls(e, response=response, explanation=explanation)
APIError: 409 Client Error: Conflict ("unable to remove volume: remove 5bf0f7dc5817523de86ada3ec32c62394d4c8405766e8954ec10647aaa215be2: volume is in use - [c89fdf6db178730df1dc945542ba91691e1d3004c772ba6763381887976cb869]")
```

spulec:
@2rs2ts is it a big deal to upgrade `docker-py`? I'm trying to figure out how important it is for us to make this backwards-compatible.

Obviously in hindsight, we should try to never break compatibility like this on minor versions, but what is done is done.

2rs2ts:
The tool I had this problem on isn't getting new features, it just didn't have its deps pinned so I've pinned them and now it's stable, and I can just stay on an old version of moto... until AWS changes something :)

Upgrading would be a yak shave for us, which I know is more our problem than yours, but if you can fix backwards compatibility in a day's worth of work or less I'd really appreciate that. If not then ok, maybe just call it out in the documentation or if there's some way you can have moto's setup conflict with docker-py < 3 then that's even better.

spulec:
@2rs2ts I made a change with https://github.com/spulec/moto/commit/cf71532fa9c5f281efd3eca01ed776e7c0d4ac26 that should work. Let me know if it doesn't.

2rs2ts:
@spulec sorry it took me so long to get back to you on this, but it did in fact work. I'm looking forward to the next release!

Issue #1483, opened 2018-02-19
Add SSM send_command support


derwolfe:
Add basic support for boto3's SSM `send_command` function as detailed at https://boto3.readthedocs.io/en/latest/reference/services/ssm.html#SSM.Client.send_command.

As of right now, it doesn't seem like this mock should require any real persistence unless users would want to be able to alter the state of jobs or test timeout logic.

spulec:
Closed with #1484

PR #1484, opened 2018-02-19
Add support for SSM send_command

derwolfe:
Addresses https://github.com/spulec/moto/issues/1483.

This adds very basic support for `send_command` by simply returning a successful send command with the parameters the user has sent.


spulec:
Thanks and welcome to Moto!

Issue #1487, opened 2018-02-22
`redshift.respones.py` does not provide `IamRoles` key in the response.


andharris:
Redshift now allows `IamRoles` param so that specific roles can assume the cluster role and this is supported in `boto` but not able to be mocked in `moto`. 

I tried adding 
```
@mock_redshift
def test_create_cluster_with_iam_roles():
    iam_role = 'arn:aws:iam:::role/my-iam-role'
    client = boto3.client('redshift', region_name='us-east-1')
    cluster_id = 'my_cluster'
    client.create_cluster(
        ClusterIdentifier=cluster_id,
        NodeType="dw.hs1.xlarge",
        MasterUsername="username",
        MasterUserPassword="password",
        IamRoles=[iam_role],
    )
    response = client.describe_clusters(ClusterIdentifier=cluster_id)
    cluster = response['Clusters'][0]
    iam_roles = [role['IamRoleArn'] for role in cluster['IamRoles']]
    list(iam_roles).should.equal([iam_role.arn])
```

but get the following error
```
ERROR: test_redshift.test_create_cluster_with_iam_roles
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/venv/lib/python3.6/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/moto/moto/core/models.py", line 71, in wrapper
    result = func(*args, **kwargs)
  File "/moto/tests/test_redshift/test_redshift.py", line 311, in test_create_cluster_with_iam_roles
    iam_roles = [role['IamRoleArn'] for role in cluster['IamRoles']]
KeyError: 'IamRoles'
```

I've attempted a PR but the tests are still failing. Any input/guidance would be awesome and I'm happy to continue working on the PR to support this. 
https://github.com/spulec/moto/pull/1488

bpandola:
Make sure you update the `to_json()` method in the `Cluster` class in `redshift/models.py` to include the list of IAM roles.

andharris:
The PR is passing the tests now.

spulec:
Closed with #1488

PR #1488, opened 2018-02-22
add iam roles to redshift

andharris:
Attempting to add support for iam roles to the redshift via https://github.com/spulec/moto/issues/1487

### DIFF
--- a/moto/redshift/models.py
+++ b/moto/redshift/models.py
@@ -67,7 +67,7 @@ class Cluster(TaggableResourceMixin, BaseModel):
                  preferred_maintenance_window, cluster_parameter_group_name,
                  automated_snapshot_retention_period, port, cluster_version,
                  allow_version_upgrade, number_of_nodes, publicly_accessible,
-                 encrypted, region_name, tags=None):
+                 encrypted, region_name, tags=None, iam_roles=None):
         super(Cluster, self).__init__(region_name, tags)
         self.redshift_backend = redshift_backend
         self.cluster_identifier = cluster_identifier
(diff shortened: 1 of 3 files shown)


moto/redshift/models.py
```diff
@@ -67,7 +67,7 @@ def __init__(self, redshift_backend, cluster_identifier, node_type, master_usern
                  preferred_maintenance_window, cluster_parameter_group_name,
                  automated_snapshot_retention_period, port, cluster_version,
                  allow_version_upgrade, number_of_nodes, publicly_accessible,
-                 encrypted, region_name, tags=None):
+                 encrypted, region_name, tags=None, iam_roles_arn=[]):
```
spulec: Using mutable default arguments can be dangerous; see [here](http://docs.python-guide.org/en/latest/writing/gotchas/#mutable-default-arguments). Probably better to default to None and then set it to an empty list if None after.

moto/redshift/models.py
```diff
@@ -351,14 +357,15 @@ class Snapshot(TaggableResourceMixin, BaseModel):
 
     resource_type = 'snapshot'
 
-    def __init__(self, cluster, snapshot_identifier, region_name, tags=None):
+    def __init__(self, cluster, snapshot_identifier, region_name, tags=None, iam_roles_arn=[]):
```
spulec: Same thing here

spulec:
Two small comments, but otherwise looks great. Excited to get this in.

moto/redshift/models.py
andharris: Ah yes, this definitely needs to be avoided. I've updated the default to `None`.

spulec:
Awesome. Thanks for this and welcome to Moto!

andharris:
Thank you. This is a great project and happy to be able to make a small contribution

Issue #1026, opened 2017-07-27
@mock_s3 permanently breaks requests library


samfrances:
At first look, I thought this was a duplicate of https://github.com/spulec/moto/issues/303.

However, on further inspection, the problem is only present with the `mock_s3` decorator, not with the `mock_s3_deprecated` decorator. Therefore, if my inspection of the code base is right, this looks like a problem related to the `responses` library or the use of it in `moto`, whereas https://github.com/spulec/moto/issues/303 concerned the HTTPretty library.

Anyway, the meat of the issue:

After using `@mock_s3` in django tests, other test classes that use `LiveserverTestCase` and the `requests` library stop working, and fail with the error: ` ConnectionError: Connection refused`.

This is not an inherent problem with `responses` itself, as the following test demonstrates:

```
python -c "from moto.packages.responses.responses import mock; import requests; mock.start(); mock.stop(); print requests.get('http://google.com')"
<Response [200]>
```

However, if `mock.stop()` is omitted, you see the same error as in the django test:

```
python -c "from moto.packages.responses.responses import mock; import requests; mock.start(); print requests.get('http://google.com')"
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/home/REDACTED/env/local/lib/python2.7/site-packages/requests/api.py", line 72, in get
    return request('get', url, params=params, **kwargs)
  File "/home/REDACTED/env/local/lib/python2.7/site-packages/requests/api.py", line 58, in request
    return session.request(method=method, url=url, **kwargs)
  File "/home/REDACTED/env/local/lib/python2.7/site-packages/requests/sessions.py", line 502, in request
    resp = self.send(prep, **send_kwargs)
  File "/home/REDACTED/env/local/lib/python2.7/site-packages/requests/sessions.py", line 612, in send
    r = adapter.send(request, **kwargs)
  File "/home/REDACTED/env/local/lib/python2.7/site-packages/moto/packages/responses/responses.py", line 302, in unbound_on_send
    return self._on_request(adapter, request, *a, **kwargs)
  File "/home/REDACTED/env/local/lib/python2.7/site-packages/moto/packages/responses/responses.py", line 244, in _on_request
    raise response
requests.exceptions.ConnectionError: Connection refused: GET http://google.com/
```
This suggests to me that, one way or another, moto is failing to call `mock.stop()`.

moto version 1.0.1
Django version 1.11.3

samfrances:
[...]

PR #1489, opened 2018-02-23
According to what I said on spulec/moto#1026 Add a unittest and attem…

hoangelos:
This isn't the whole fix, but it does expose the problem a bit better.  Using HTTPretty, it checks to see if it's enabled already before it enables it.  There doesn't seem to be a way for the MockResponse object to do that easily without adding new code to the fork, or by introspecting class attributes.  This makes my test pass, but breaks many other tests.

### DIFF
--- a/moto/core/models.py
+++ b/moto/core/models.py
@@ -130,7 +130,8 @@ class ResponsesMockAWS(BaseMockAWS):
         responses.reset()
 
     def enable_patching(self):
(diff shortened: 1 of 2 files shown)


bblommers:
As far as I can tell, the original problem has already been solved, so I'll close this.
Do let us know if there are still issues with pass throughs.

PR #1498, opened 2018-03-05
Fix handling of PageSize for ELB/ELBv2

### LINKED ISSUE #1497: ELB (and v1 and v2) describe_load_balancers should correctly support PageSize
If I call `response = elbv2_client.describe_load_balancers(PageSize=5)`, it raises

```
self = <moto.elbv2.responses.ELBV2Response object at 0x10c9284a8>

    @amzn_request_id
    def describe_load_balancers(self):
        arns = self._get_multi_param("LoadBalancerArns.member")
        names = self._get_multi_param("Names.member")
        all_load_balancers = list(self.elbv2_backend.describe_load_balancers(arns, names))
        marker = self._get_param('Marker')
        all_names = [balancer.name for balancer in all_load_balancers]
        if marker:
            start = all_names.index(marker) + 1
        else:
            start = 0
        page_size = self._get_param('PageSize', 50)  # the default is 400, but using 50 to make testing easier
>       load_balancers_resp = all_load_balancers[start:start + page_size]
E       TypeError: unsupported operand type(s) for +: 'int' and 'str'
```

When I call this method without specifying PageSize, it works correctly.

It looks like it should be using _get_int_param('PageSize', 50) instead of _get_param.  I'll submit a patch with this fix shortly.

dgolombek:
Resolves #1497

### DIFF
--- a/moto/elb/responses.py
+++ b/moto/elb/responses.py
@@ -61,7 +61,7 @@ class ELBResponse(BaseResponse):
             start = all_names.index(marker) + 1
         else:
             start = 0
-        page_size = self._get_param('PageSize', 50)  # the default is 400, but using 50 to make testing easier
+        page_size = self._get_int_param('PageSize', 50)  # the default is 400, but using 50 to make testing easier
         load_balancers_resp = all_load_balancers[start:start + page_size]
         next_marker = None
         if len(all_load_balancers) > start + page_size:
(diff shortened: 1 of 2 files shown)


spulec:
Thanks and welcome to the Moto community!

dgolombek:
Thanks!

PR #1501, opened 2018-03-06
ELBv2.create_listener links TargetGroup to LB

### LINKED ISSUE #1500: ELBv2.create_listener should associate TargetGroup with LB
If you create a TargetGroup, associate it with a LB via create_listener, and then call describe_target_groups specifying BOTH the load balancer ARN and the target group name,  no target group is returned.  

```python
response = conn.create_target_group(
        Name='a-target',
        Protocol='HTTP',
        Port=8080,
        VpcId=vpc.id)
target_group = response.get('TargetGroups')[0]
conn.create_listener(
        LoadBalancerArn=load_balancer_arn,
        Protocol='HTTP',
        Port=80,
        DefaultActions=[{'Type': 'forward', 'TargetGroupArn': target_group.get('TargetGroupArn')}])
response = conn.describe_target_groups(LoadBalancerArn=load_balancer_arn,
                                                                      Names=['a-target'])
````

The bug is in `create_listener`, which is not adding the load balancer to the target_group's list of load_balancer_arns. I'll submit a PR for this shortly.

dgolombek:
In order to search target_groups by LB, we need this link in place.

Resolves #1500

### DIFF
--- a/moto/elbv2/models.py
+++ b/moto/elbv2/models.py
@@ -486,6 +486,10 @@ class ELBv2Backend(BaseBackend):
         arn = load_balancer_arn.replace(':loadbalancer/', ':listener/') + "/%s%s" % (port, id(self))
         listener = FakeListener(load_balancer_arn, arn, protocol, port, ssl_policy, certificate, default_actions)
         balancer.listeners[listener.arn] = listener
+        for action in default_actions:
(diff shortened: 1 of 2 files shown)


spulec:
Looks great, thanks!

PR #1504, opened 2018-03-08
add UsePreviousValue support for parameters when updating a CloudFormation stack

RealSalmon:
This adds support for the UsePreviousValue field of the CloudFormation Parameter data type when updating a CloudFormation stack as described at https://docs.aws.amazon.com/AWSCloudFormation/latest/APIReference/API_Parameter.html

I hope this is helpful. I have been using moto _alot_ lately, and I'm glad to be able to contribute in some small way!


JackDanger:
@RealSalmon This is hugely helpful! Glad Moto has been useful to you. This contribution is much appreciated and we're glad to have you here at Moto!

Issue #1370, opened 2017-12-01
Implement suspend_processes in AutoScaling service


LondonAppDev:
I am using moto `1.1.25` and when I try and mock `client.suspend_processes()` I get:

```
NotImplementedError: The suspend_processes action has not been implemented
```

I searched issues for `suspend_processes` and had no hits so assuming this feature hasn't been requested yet?

terrycain:
Can I get a simple test case for this, then I'll go and add it

LondonAppDev:
Sure, here would be an example test case:
```
@mock_autoscaling
def test_asg():
    client = boto3.client('autoscaling')
    client.create_launch_configuration(
        LaunchConfigurationName='lc',
    )
    client.create_auto_scaling_group(
        LaunchConfigurationName='lc',
        AutoScalingGroupName='test-asg',
        MinSize=1,
        MaxSize=1,
    )

    # Testing something that calls the below...
    client.suspend_processes(
        AutoScalingGroupName='test-asg',
        ScalingProcesses=['Launch']
    )

    res = client.describe_auto_scaling_groups(
        AutoScalingGroupNames=['test-asg']
    )
    launch_suspended = False
    for proc in res['AutoScalingGroups'][0]['SuspendedProcesses']:
        if proc.get('ProcessName') == 'Launch':
            launch_suspended = True

    assert launch_suspended is True
```

jimjshields:
I opened a PR for this (https://github.com/spulec/moto/pull/1509) that has your test case passing — I haven't used boto for this, so unsure if there are other test cases necessary. Let me know if you have any feedback, thanks!

PR #1509, opened 2018-03-09
Add suspend processes to AutoScaling service

jimjshields:
This is to address https://github.com/spulec/moto/issues/1370. Let me know if you have any feedback, happy to add more tests if necessary!

* Implement `suspend_processes` in `autoscaling.models` and `autoscaling.responses`
* Use suggested test from https://github.com/spulec/moto/issues/1370


spulec:
This is great! Thanks for your help and welcome to Moto!

PR #1511, opened 2018-03-11
Enable Extended CIDR Associations on VPC

rocky4570:
August 29, 2017 AWS announced the ability to expand their VPCs by adding additional CIDR ranges (up to 5 at this time), as well as added support for IPV6. Note: Boto which is effectively out of date now doesnt support extended cidrs in the API and has to be done via Boto3! Error codes from actual runs on AWS.

This is not fully tested with all VPC networking as someone added docker to lambda which makes it bit more awkward for developers running Windows 7 or non dockerable Windows 10 environments. So will see how the unittests work here. 

Tested for Python 2.7.14 and 3.6.4.

### DIFF
--- a/moto/ec2/exceptions.py
+++ b/moto/ec2/exceptions.py
@@ -400,12 +400,14 @@ class FilterNotImplementedError(MotoNotImplementedError):
             "The filter '{0}' for {1}".format(
                 filter_name, method_name))
 
+
 class CidrLimitExceeded(EC2ClientError):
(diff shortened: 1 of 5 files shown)


JackDanger:
Thanks for this @rocky4570! Apologies for the difficulty running tests locally - we did add a Docker dependency for running Lambda tests.
It's amazing to see boto get left behind by boto3. Thanks for keeping us up to date with AWS!

PR #1516, opened 2018-03-17
Implemented S3 get/put_notification_configuration

### LINKED ISSUE #973: S3 Notification to Lambda Configuration
Hi,
It would be great if I could use `moto` to simulate S3 notifications to an AWS Lambda. An e.g. use case is every time an object is created in S3, invoke a Lambda. My Deployment scripts have code which normally looks like this which sets up the notification configuration

```python
conf = .. my aws keys etc..
lambda_client = client(conf, 'lambda', region_name=region_name)
s3 = resource(conf, 's3', region_name=region_name)
lambda_client.add_permission(**s3_permission_params)
bucket_notification = s3.BucketNotification(bucket_name)
bucket_notification.put(
        NotificationConfiguration={'LambdaFunctionConfigurations': [
            lambda_configuration_params,
 ]})
```

Currently, is it supported that `moto` will automatically invoke the lambda if I use `mock_s3` and put an object in the in-memory s3 that `moto` creates? I searched the repo a bit but I couldn't find any. I just wanted to make sure. 

Great library btw, thanks for the contribution.

Regards,
Rajiv

mikegrima:
Implements `get_bucket_notification_configuration` and `put_bucket_notification_configuration`.

Supports SNS, SQS, and Lambda. I don't check for the existence of the Queues, Topics, or Functions -- I assume that is already there.

This does check that the event names are correct, that the notification target is in the same region as the bucket, and that the ARN is "correct".

closes #973

### DIFF
--- a/moto/s3/exceptions.py
+++ b/moto/s3/exceptions.py
@@ -138,3 +138,33 @@ class CrossLocationLoggingProhibitted(S3ClientError):
             "CrossLocationLoggingProhibitted",
             "Cross S3 location logging not allowed."
         )
+
+
+class InvalidNotificationARN(S3ClientError):
+    code = 400
+
+    def __init__(self, *args, **kwargs):
+        super(InvalidNotificationARN, self).__init__(
+            "InvalidArgument",
+            "The ARN is not well formed",
+            *args, **kwargs)
+
+
+class InvalidNotificationDestination(S3ClientError):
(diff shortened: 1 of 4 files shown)


JackDanger:
This is great @mikegrima! Thanks for adding this (and for the great comments in your code).

PR #1520, opened 2018-03-21
Improve SQS Compatibility with AWS

iainb:
* Return correct error code when fetching a queue that does not exist
* Queue creation and set_attributes uses the same code path
  - ensure bool/int values are cast correctly
* RedrivePolicy is handled properly with set_attributes
  - _setup_dlq is called
  - is json decoded, so that returned RedrivePolicy is not json encoded twice
* As per AWS not all attributes are returned when they are not set, for  example RedrivePolicy, fifoQueue, Policy, Kms*
* WaitTimeSeconds is not a queue attribute switch to ReceiveMessageWaitTimeSeconds

### DIFF
--- a/moto/sqs/models.py
+++ b/moto/sqs/models.py
@@ -152,63 +152,85 @@ class Message(BaseModel):
 
 
 class Queue(BaseModel):
-    camelcase_attributes = ['ApproximateNumberOfMessages',
-                            'ApproximateNumberOfMessagesDelayed',
-                            'ApproximateNumberOfMessagesNotVisible',
-                            'ContentBasedDeduplication',
-                            'CreatedTimestamp',
-                            'DelaySeconds',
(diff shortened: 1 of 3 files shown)


moto/sqs/models.py
```diff
@@ -152,64 +152,86 @@ def delayed(self):
...
+                       'ApproximateNumberOfMessagesNotVisible',
+                       'CreatedTimestamp',
+                       'DelaySeconds',
+                       'LastModifiedTimestamp',
+                       'MaximumMessageSize',
+                       'MessageRetentionPeriod',
+                       'QueueArn',
+                       'ReceiveMessageWaitTimeSeconds',
+                       'VisibilityTimeout']
+    fifo_attributes = ['FifoQueue',
```
JackDanger: I love the way you broke this out 👍

PR #1521, opened 2018-03-21
Implement basic SNS message filtering

iainb:
This adds support for exact string value matching along with AND/OR logic as described here:
    
https://docs.aws.amazon.com/sns/latest/dg/message-filtering.html
   
It does not provide support for:
  - Anything-but string matching
  - Prefix string matching
  - Numeric Value Matching

The above filter policies (if configured) will not match messages.


JackDanger:
This is really helpful, @ianb. Thanks so much! Merging now.

PR #1535, opened 2018-03-26
Added Filtering support for S3 lifecycle

### LINKED ISSUE #1479: ExpiredObjectDeleteMarker not supported in bucket lifecycle rules
I can do a put against a mocked BucketLifecycleConfiguration with ExpiredObjectDeleteMarker set in the Expiration, but then subsequent accessor to rule doesn't show.  

Looks like the LifecycleRule in models.py doesn't have include attributes to model this flag.
https://github.com/spulec/moto/blob/770281aef2132de8b4d8abb7ec3325a3038b6f09/moto/s3/models.py#L312

######
  mocked_s3 = boto3.resource('s3', region_name='us-east-1')
  mocked_s3.create_bucket(Bucket='mybucket')
  bucket_lifecycle_configuration = mocked_s3.BucketLifecycleConfiguration('mybucket')
  bucket_lifecycle_configuration.put(LifecycleConfiguration={
    'Rules': [
      {
        'Expiration': {
          'Days': 123,
          'ExpiredObjectDeleteMarker': True
        },
        'ID': 'id1',
        'Status': 'Enabled'
      }
    ]
  })

  bucket_lifecycle_configuration2 = mocked_s3.BucketLifecycleConfiguration('mybucket')
  assert bucket_lifecycle_configuration2.rules[0]['Expiration']['ExpiredObjectDeleteMarker'] == True

### LINKED ISSUE #1533: S3 Bucket Lifecycle Needs Filter Support
The `put_bucket_lifecycle_configuration()` indicates that the top level `Prefix` field is now deprecated http://boto3.readthedocs.io/en/latest/reference/services/s3.html#S3.Client.put_bucket_lifecycle_configuration.

Instead, there are now filter rules where the prefix can be added.

The `Filter` rules will need to be implemented (currently missing).

~I don't have a ton of time to work on this, but wanted to track it here in case anyone else wanted to take a stab at it. (I'm going to continue using `Prefix` until I'm no longer required to :))~

Screw it; I'll do it.

mikegrima:
Added Filtering support for S3 lifecycle and also added `ExpiredObjectDeleteMarker`.

Note: The `Filter` feature is pretty new, so I bumped up the `boto3` and `botocore` versions to the latest. @JackDanger Please let me know if this is a problem.

closes #1533
closes #1479

### DIFF
--- a/moto/s3/models.py
+++ b/moto/s3/models.py
@@ -415,7 +415,7 @@ class FakeBucket(BaseModel):
             # Pull out the filter:
             lc_filter = None
             if rule.get("Filter"):
-                # Can't have both filter and prefix (Need to check for the presence of the Key):
+                # Can't have both `Filter` and `Prefix` (need to check for the presence of the key):
                 try:
                     if rule["Prefix"] or not rule["Prefix"]:
                         raise MalformedXML()
--- a/moto/s3/models.py
+++ b/moto/s3/models.py
@@ -15,7 +15,7 @@ from bisect import insort
 from moto.core import BaseBackend, BaseModel
 from moto.core.utils import iso_8601_datetime_with_milliseconds, rfc_1123_datetime
 from .exceptions import BucketAlreadyExists, MissingBucket, InvalidPart, EntityTooSmall, MissingKey, \
-    InvalidNotificationDestination
+    InvalidNotificationDestination, MalformedXML
 from .utils import clean_key_name, _VersionedKeyStore
 
 UPLOAD_ID_BYTES = 43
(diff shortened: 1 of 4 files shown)


mikegrima:
Added more tests, but coverage decreased for some reason... 🤔

mikegrima:
Don't merge in yet -- found a bug where the top-level `Prefix` is always being set.

mikegrima:
Should be fixed now.

JackDanger:
The boto/botocore upgrades should be good - we've recently released a version so folks can pin to that for a while if this changes causes them upgrade pain.

JackDanger:
Nice work, @mikegrima! I'll merge when the build finishes.

PR #1537, opened 2018-03-26
If Properies isn't set, cloudformation will just use default values

william-richard:
The default ClusterName was already taken care of for ECS clusters, but if the `Properties` key wasn't provide, it throws a KeyError.

This is a pretty rare situation, since almost every cloudformation resource has at least one required property.


spulec:
Ah, good to know. Thanks for the fix!

PR #1544, opened 2018-03-29
fix bug with Kinesis ResourceInUse exception

dbjellan:


JackDanger:
Nice work! Thanks for the great contribution and welcome to Moto!

PR #1549, opened 2018-04-04
Fix AttributeError in filter_log_events

pj2:
An AttributeError would be thrown if the `interleaved` parameter was
passed.


spulec:
Awesome. Thanks and welcome to Moto!

PR #1550, opened 2018-04-04
Allow tagging snapshots on creation

tomelliff:
Largely copying what was done for volume creation in https://github.com/spulec/moto/pull/1432


spulec:
Awesome. Thank you and welcome to Moto!

PR #1552, opened 2018-04-04
ECS CPU, memory hard limits and host ports are all optional.

william-richard:
http://boto3.readthedocs.io/en/latest/reference/services/ecs.html#ECS.Client.register_task_definition


spulec:
Thanks!

PR #1553, opened 2018-04-05
Unvendor responses, move back to upstream

spulec:
Trying to move us back upstream.

Should help with #1026


moto/core/models.py
```diff
@@ -124,31 +124,90 @@ def disable_patching(self):
...
+            return False
+
 
+botocore_mock = responses.RequestsMock(assert_all_requests_are_fired=False, target='botocore.vendored.requests.adapters.HTTPAdapter.send')
+
+
+class ResponsesMockAWS(BaseMockAWS):
     def reset(self):
+        botocore_mock.reset()
         responses.reset()
```
lsutic: I think we will have problems if multiple clients use `responses`, some of them with a different lifecycle for the responses default mock? I think the `ResponsesMockAWS` should have its own instance of a `RequestsMock`, and use that. The `requests` quickstart teaches people to use the default mock, and if another library like `moto` starts resetting it, then there will be issues.

moto/core/models.py
```diff
@@ -124,31 +124,90 @@ def disable_patching(self):
...
+        else:
+            return False
+
 
+botocore_mock = responses.RequestsMock(assert_all_requests_are_fired=False, target='botocore.vendored.requests.adapters.HTTPAdapter.send')
+
+
+class ResponsesMockAWS(BaseMockAWS):
     def reset(self):
+        botocore_mock.reset()
```
lsutic: What happens if we have multiple `ResponsesMockAWS` instances? The `RequestsMock` does not have an internal usage counter, so calling `reset` or `disable_patching` on any `ResponsesMockAWS` instance will reset or stop the `botocore_mock` for all.

Consider:

```
with mock_s3():
    with mock_ec2():
        ...
    with mock_ec2():
        ...
```
If I understand the workings of `moto` right, once leaving the scope of the first `mock_ec2` the `botocore_mock` is stopped.

moto/core/models.py
spulec: Hmmmm.... I think we will have problems either way, right? `requests` will only be pointed at one or the other. Agree that resetting the default isn't great either though.

moto/core/models.py
spulec: `reset` should only get called on the first decorator due to this block:

https://github.com/spulec/moto/blob/master/moto/core/models.py#L36-L37

I confirmed with your example code that it will only get called in the `mock_s3` one.

moto/core/models.py
lsutic: You're right. I thought `responses` mocks were chained, but unless you explicitly specify a passthrough prefix, they don't. (And I'm not even sure the passthrough will chain properly.)

moto/core/models.py
lsutic: You're right. My bad.

jskebo:
On a fresh install from master (`pip install git+git://github.com/spulec/moto#egg=moto`) I'm getting this error on import:

```python-traceback
  File "./manage.py", line 13, in <module>
    execute_from_command_line(sys.argv)
  File "/Users/jeff/.virtualenv/test_moto/lib/python2.7/site-packages/django/core/management/__init__.py", line 364, in execute_from_command_line
    utility.execute()
  File "/Users/jeff/.virtualenv/test_moto/lib/python2.7/site-packages/django/core/management/__init__.py", line 356, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/Users/jeff/.virtualenv/test_moto/lib/python2.7/site-packages/django/core/management/commands/test.py", line 29, in run_from_argv
    super(Command, self).run_from_argv(argv)
  File "/Users/jeff/.virtualenv/test_moto/lib/python2.7/site-packages/django/core/management/base.py", line 283, in run_from_argv
    self.execute(*args, **cmd_options)
  File "/Users/jeff/.virtualenv/test_moto/lib/python2.7/site-packages/django/core/management/base.py", line 330, in execute
    output = self.handle(*args, **options)
  File "/Users/jeff/.virtualenv/test_moto/lib/python2.7/site-packages/django/core/management/commands/test.py", line 62, in handle
    failures = test_runner.run_tests(test_labels)
  File "/Users/jeff/.virtualenv/test_moto/lib/python2.7/site-packages/django/test/runner.py", line 600, in run_tests
    suite = self.build_suite(test_labels, extra_tests)
  File "/Users/jeff/.virtualenv/test_moto/lib/python2.7/site-packages/django/test/runner.py", line 484, in build_suite
    tests = self.test_loader.loadTestsFromName(label)
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/unittest/loader.py", line 91, in loadTestsFromName
    module = __import__('.'.join(parts_copy))
  File "/Users/jeff/code/test_moto/tests/views/test_community_views.py", line 10, in <module>
    from moto import mock_s3
  File "/Users/jeff/.virtualenv/test_moto/lib/python2.7/site-packages/moto/__init__.py", line 8, in <module>
    from .acm import mock_acm  # flake8: noqa
  File "/Users/jeff/.virtualenv/test_moto/lib/python2.7/site-packages/moto/acm/__init__.py", line 2, in <module>
    from .models import acm_backends
  File "/Users/jeff/.virtualenv/test_moto/lib/python2.7/site-packages/moto/acm/models.py", line 6, in <module>
  File "/Users/jeff/.virtualenv/test_moto/lib/python2.7/site-packages/moto/core/__init__.py", line 2, in <module>
    from .models import BaseModel, BaseBackend, moto_api_backend  # flake8: noqa
  File "/Users/jeff/.virtualenv/test_moto/lib/python2.7/site-packages/moto/core/models.py", line 174, in <module>
    botocore_mock = responses.RequestsMock(assert_all_requests_are_fired=False, target='botocore.vendored.requests.adapters.HTTPAdapter.send')
TypeError: __init__() got an unexpected keyword argument 'target'
```

Previously, this error did not occur. Think this pull request and the error are related?

spulec:
Yeah, can you check your version of responses? I'm guessing we need to require a higher version than what you are using.

PR #1556, opened 2018-04-07
Fix/lambda backend

avara1986:
- Fix error when use this code:
```
client = boto3.client('lambda')
client.get_policy([...])
```
 and moto rise:
```
moto/awslambda/responses.py", line 109, in _get_policy
    lambda_backend = self.get_lambda_backend(full_url)
Exception: 'LambdaResponse' object has no attribute 'get_lambda_backend'
```

- fix shadows built-in name in `moto/awslambda/responses.py`


JackDanger:
@avara1986 Great fix! Thanks for this. And welcome to Moto!

PR #1558, opened 2018-04-09
Fix/cloud watch fix put metric data with timestamp bug

syunwei:
Cloud Watch put metric data method was using string in timestamp parameter.
It caused error when try to get metric statistic data with timestamp.
Fix put metric data method and change response of get metric statistic data.

### DIFF
--- a/moto/cloudwatch/models.py
+++ b/moto/cloudwatch/models.py
@@ -1,12 +1,13 @@
 
 import json
-from moto.core.utils import iso_8601_datetime_with_milliseconds
+from moto.core.utils import iso_8601_datetime_without_milliseconds
 from moto.core import BaseBackend, BaseModel
 from moto.core.exceptions import RESTError
 import boto.ec2.cloudwatch
 from datetime import datetime, timedelta
 from dateutil.tz import tzutc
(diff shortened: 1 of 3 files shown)


spulec:
Hmmm, looks like Travis is having some issues

jakul:
I've had this issue today too:

`ValueError: time data '2018-01-10T00:00:00Z' does not match format '%Y-%m-%dT%H:%M:%S.%fZ'`, due to

```
.tox/app/lib/python3.6/site-packages/moto/cloudwatch/models.py:235: in put_metric_data
    timestamp = datetime.strptime(timestamp, '%Y-%m-%dT%H:%M:%S.%fZ')
```

lodo1995:
I had the same issue as @jakul . I build the timestamp with

    datetime.utcfromtimestamp(some_integer_number_of_seconds)

I think I might be able to avoid the error by doing

    datetime.utcfromtimestamp(some_integer_number_of_seconds + 0.1)

This change would be ignored by AWS (which even with high resolution only uses whole seconds) and would make moto happy. But I don't like "polluting" my code just to make tests pass.

bblommers:
Closing - implemented as part of #2801

PR #1563, opened 2018-04-11
Correctly generate resource name for target groups when using cloudformation

william-richard:
They need to have less than 32 character names, so when you don't specify a name
cloudformation generates a name that is less than 32 characters.


spulec:
Looks like one of the tests is failing on Travis

william-richard:
I'm not able to reliably reproduce this error, either locally or in travis.  I made the test a bit less strict, and I think that should fix the problems.

spulec:
This looks great; thanks!

PR #1572, opened 2018-04-17
Fixing create_key_and_certificate boolean parameter

anozaki:
create_key_and_certificate's setAsActive was getting pulled in as a string. Updating code to pull the parameter in as boolean.


JackDanger:
Nice fix, @anozaki! This will help lots of people. Welcome to Moto!

PR #1576, opened 2018-04-18
SNS support for all boto3 regions.

### LINKED ISSUE #1575: SNS does not support new regions
SNS appears to be basing its available regions off of boto2:
```
sns_backends = {}
for region in boto.sns.regions():
    sns_backends[region.name] = SNSBackend(region.name)
```
We will need to move this over to boto3 to support the new regional endpoints.

I'm going to submit a PR for this change.

mikegrima:
Closes #1575 .

### DIFF
--- a/moto/sns/models.py
+++ b/moto/sns/models.py
@@ -4,10 +4,11 @@ import datetime
 import uuid
 import json
 
-import boto.sns
 import requests
 import six
 import re
+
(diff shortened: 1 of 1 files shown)


mikegrima:
Confirmed that this works internally for me.

spulec:
Awesome; thanks!

Issue #1578, opened 2018-04-18
SQS messages don't contain MessageGroupId


DHager:
My use-case involves a simple FIFO queue, where the MessageGroupId isn't only an instruction to SQS, but also contains some data that the consumer is also interested in accessing later. (Note that this is *not* the same kind of "attribute" as the user-land ones using `MessageAttributes`)

### Reproduction script

```python
from moto import mock_sqs
import boto3

with mock_sqs():
    queue = boto3.resource('sqs').create_queue(
        QueueName="test_queue",
        Attributes={
            'FifoQueue ': 'true',
            'ContentBasedDeduplication': 'false'
        }
    )

    queue.send_message(
        MessageBody="{}",
        MessageDeduplicationId="DDID-1",
        MessageGroupId="GID-1",
    )

    messages = queue.receive_messages(
        AttributeNames=['MessageGroupId'],
        MaxNumberOfMessages=10,
        VisibilityTimeout=10,
        WaitTimeSeconds=0,
    )

    msg = messages[0]
    mgid = msg.attributes['MessageGroupId']
    assert "GID-1" == mgid
    
```

Fails with:

```
Traceback (most recent call last):
  File "test.py", line 27, in <module>
    mgid = msg.attributes['MessageGroupId']
KeyError: 'MessageGroupId'
```

### Next steps 

I plan to start a PR later which will at least allow the data to flow from producer to consumer.  

I am not familiar with the moto codebase, but my initial scan of the code suggests changes would be necessary to:

* Inside `moto.sqs.models.SQSBackend#send_message` pull the data out with `self._get_param('MessageGroupId')`
* Store the value as a property of `moto.sqs.models.Message`
* Put in an XML `<Attribute>` block in `moto.sqs.responses.RECEIVE_MESSAGE_RESPONSE` that emits the value from the mode.

I expect the main issues will involve the extent to which the attribute should only be allowed or appear on certain queue-types.

### Out of scope

Will not implement actually using the `MessageGroupId` for selectively ordering messages -- that's a much bigger problem and may be related to work needed for #1438.

spulec:
Closed with #1580

azhadsyed:
I noticed in "out of scope":

> Will not implement actually using the MessageGroupId for selectively ordering messages

Do you know if this is being tracked? I didn't see it but just checking.

bblommers:
[...]

PR #1580, opened 2018-04-19
Fix missing MessageGroupId and MessageDeduplicationId

DHager:
Solves issue detailed in #1578, where some system-level attributes on a message don't survive the mock queuing.

Does **not** change or enhance FIFO queue behavior for true deduplication or groups.

### DIFF
--- a/moto/sqs/models.py
+++ b/moto/sqs/models.py
@@ -38,6 +38,8 @@ class Message(BaseModel):
         self.sent_timestamp = None
         self.approximate_first_receive_timestamp = None
         self.approximate_receive_count = 0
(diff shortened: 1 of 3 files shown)


spulec:
This is great! Thanks for the contribution and welcome to Moto!

DHager:
A little more experimentation on this, and technically the Attribute should only be present if the receiver indicated they wanted it with `AttributeNames=['MessageGroupId']`. (Not to be confused with a different argument, `MessageAttributeNames`.) 

So right now it's... over-expressing, which shouldn't *break* anything, but it diverges a bit from the non-mocked behavior.

PR #1574, opened 2018-04-18
Include SNS message attributes with message body when delivering to SQS.

wblackconv:
Aligning with behaviour described here:
https://docs.aws.amazon.com/sns/latest/dg/SNSMessageAttributes.html

PR #1582, opened 2018-04-19
Add tests for message attribute validation in SNS

wblackconv:
Fixes regression in test coverage from #1574 

Fixes up bug in return value of moto.sns.responses.SNSResponse._parse_message_attributes due to accidental recycling of a variable.


spulec:
Hmm, looks like Travis is complaining on a string vs bytes issue.

JackDanger:
Nice work, @wblackconv!

Issue #1393, opened 2017-12-11
SQS Queue creation with RedrivePolicy fails due to dict vs. string definition


philvarner:
Moto 1.1.25

The proximate issue is that the error `TypeError: the JSON object must be str, bytes or bytearray, not 'dict'` is thrown when trying to load the redrive policy in:
```
moto/sqs/models.py:214: in _setup_dlq
    self.redrive_policy = json.loads(policy_json)
```

This is because with properly constructed Redrive Policy, the variable `policy_json` is a dictionary rather than a string. 

The test for this passes, but incorrect uses:

 'RedrivePolicy': json.dumps({'deadLetterTargetArn': queue_arn1, 'maxReceiveCount': 2})

which constructs the code for RedrivePolicy as a string rather than a nested object, which is incorrect, as the json.loads of the CloudFormation template JSON will result in an in-memory dictionary, and thereby triggering this bug. 

The CloudFormation template I triggered this with is at the bottom of the issue.  This has been successfully executed in real CloudFormation, so I know it is valid. 

This fix should be straightforward in simply changing the parsing from string to assuming a dictionary (or possibly handling both cases, even though I believe the string case is not valid)

Another option would be to generate the JSON for testing using Troposphere, as I imagine most users are using it in conjunction with Boto3. That code would look something like: 

```

    dlq = t.add_resource(
        Queue(
            'DeadLetterQueue',
            QueueName=DLQQueueName,
            MaximumMessageSize=262144,  # 256KiB
            MessageRetentionPeriod=1209600,  # 14 days
            VisibilityTimeout=30
        )
    )
t.add_resource(
        Queue(
            'PrimaryQueue',
...
  RedrivePolicy=RedrivePolicy(
                deadLetterTargetArn=GetAtt(
                    dlq.title,
                    'Arn'),
                maxReceiveCount=10
            ),
...
)
```

Stack:
```
mycode/queue_forklift.py:168: in create_or_update_queues
    stack_result = client.create_stack(**params)
../../.local/share/virtualenvs/surveygizmo-3-y45dWN/lib/python3.6/site-packages/botocore/client.py:314: in _api_call
    return self._make_api_call(operation_name, kwargs)
../../.local/share/virtualenvs/surveygizmo-3-y45dWN/lib/python3.6/site-packages/botocore/client.py:599: in _make_api_call
    operation_model, request_dict)
[...]

PR #1586, opened 2018-04-20
Fix Cloudformation not accepting non-JSON SQS redrive policy

DHager:
Adds a unit-test for issue #1393 and fixes it. 

Note that this is focused towards the CloudFormation use-case, and not the SQS client or resource APIs. Those other ways of creating queues may still have their own validation requirements, which may-or-may-not still need to require JSON-only inputs.


spulec:
Thanks!

PR #1519, opened 2018-03-21
Tighten upper bound on python-dateutil to match botocore

iainb:
Due to:

https://github.com/boto/botocore/commit/90d7692702be1a423af15e0f49b58365f2a400f2

The following is seen when trying to run `pip3 --no-cache-dir install ".[server]"` for the `Dockerfile` in this repo.

```python
Traceback (most recent call last):
  File "/usr/lib/python3.6/site-packages/pkg_resources/__init__.py", line 574, in _build_master
    ws.require(__requires__)
  File "/usr/lib/python3.6/site-packages/pkg_resources/__init__.py", line 892, in require
    needed = self.resolve(parse_requirements(requirements))
  File "/usr/lib/python3.6/site-packages/pkg_resources/__init__.py", line 783, in resolve
    raise VersionConflict(dist, req).with_context(dependent_req)
pkg_resources.ContextualVersionConflict: (python-dateutil 2.7.0 (/usr/lib/python3.6/site-packages), Requirement.parse('python-dateutil<2.7.0,>=2.1'), {'botocore'})
```

PR #1589, opened 2018-04-21
Allow dateutil below 3.0.0

singingwolfboy:
The `dateutil` dependency was tightened in #1519, in order to match `botocore`. However, it looks like `botocore` is working on loosening that dependency for Python 2.7 and above: https://github.com/boto/botocore/pull/1433. Therefore it makes sense that this project should _also_ loosen its dependencies, and since we don't support Python 2.6 or below, we can do so more simply.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -18,7 +18,7 @@ install_requires = [
     "werkzeug",
     "pyaml",
     "pytz",
-    "python-dateutil<2.7.0,>=2.1",
+    "python-dateutil<3.0.0,>=2.1",
     "mock",
     "docker>=2.5.1",
     "jsondiff==1.1.1",


spulec:
Looks good; thanks!

kevgliss:
Did this change get reverted?

https://github.com/spulec/moto/blob/1.3.3/setup.py#L21 looks like it's still requiring `<2.7.0`.

kevgliss:
Correction, can we get a release with this change in it? It looks like it didn't make it out with 1.3.3.

forestmonster:
@spulec Thanks -- do you anticipate another release to bring this in?

spulec:
Yes, we'll try to get a release out in the next few days.

forestmonster:
The commit message was a little premature of me, issue hasn't been fixed yet

avatarworf:
Disappointing this missed the 1.3.4 release.

Issue #1554, opened 2018-04-06
NotImplementedError: The describe_log_groups action has not been implemented


koweblomke:
Hi. 

I tried to mock the Logs describe_log_groups(), but got NotImplementedError
I use the @mock_logs decorator

spulec:
Added as a feature request

spulec:
Closed with https://github.com/spulec/moto/pull/1593#issuecomment-393008847

Issue #1514, opened 2018-03-16
Adding the CloudFormation StackSet endpoints


svdgraaf:
I want to start implementing the CloudFormation StackSet endpoints, as I need those for a project. Is there anything I need to keep in mind when doing this?

spulec:
Nothing in particular. I'll add this issue as a feature request.

We do generally like to have high test coverage so that would be appreciated.

corrjo:
@svdgraaf have you made any progress toward adding stackset support? Let me know if I can be of any help.

corrjo:
@svdgraaf No worries, I'll see if I can carve out some time and add them :)

corrjo:
Finally found some time to do this #2015

spulec:
Closed with https://github.com/spulec/moto/pull/2015

PR #1593, opened 2018-04-24
Implement describe_log_groups() method for CloudWatchLogs

fujimotos:
### What's this patch?

This patch teaches `LogsResponse` class how to handle the DescribeLogGroups
request, so that we can mock out the `boto.describe_log_groups()` call.

With this change in place, we can write as below:

```python
@mock_logs
def test_log_group():
    conn = boto3.client('logs', 'us-west-2')

    some_method_to_init_log_groups()

    resp = conn.describe_log_groups(logGroupNamePrefix='myapp')
    assert ...
```

This should be fairly useful for a number of programs which handles
CloudWatchLogs.

### Note

The basic support this patch adds should solve #1554.

### DIFF
--- a/moto/logs/models.py
+++ b/moto/logs/models.py
@@ -132,6 +132,9 @@ class LogGroup:
     def __init__(self, region, name, tags):
         self.name = name
         self.region = region
+        self.arn = "arn:aws:logs:{region}:1:log-group:{log_group}".format(
+            region=region, log_group=name)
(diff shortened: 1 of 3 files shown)


spulec:
This looks great! Welcome to Moto!

fujimotos:
Thanks for merging!

BTW I believe the feature request #1514 has been resolved by this patch,
so it can be closed too.

PR #1595, opened 2018-04-24
SQS add missing validation to ReceiveMessage

iainb:
Motivation - I noticed that it was possible to fetch more than 10 messages from an SQS queue with moto, where as AWS enforces a limit of between 1 and 10. 

Also updated WaitTimeSeconds validation while I was near by.


JackDanger:
Nice validation add, @iainb!

PR #1600, opened 2018-04-25
Extend ssm backend

### LINKED ISSUE #1602: SSM backend doesn't support the ListCommands API call.
moto version: 1.3.3
botocore version: 1.10.4

In addition, the SendCommand call still requires InstanceIds even with this fix:
https://github.com/spulec/moto/commit/94ba2e68bdca09a48114723eedb40d38a75b6ba6.

I've made a pull request to address both these issues: #1600.

mliudev:
Extends the SSM backend with the ListCommands endpoint. Resolves #1602.

### DIFF
--- a/moto/ssm/models.py
+++ b/moto/ssm/models.py
@@ -64,10 +64,22 @@ MAX_TIMEOUT_SECONDS = 3600
 
 class Command(BaseModel):
     def __init__(self, comment='', document_name='', timeout_seconds=MAX_TIMEOUT_SECONDS,
-            instance_ids=[], max_concurrency='', max_errors='',
-            notification_config={}, output_s3_bucket_name='',
-            output_s3_key_prefix='', output_s3_region='', parameters={},
-            service_role_arn='', targets=[]):
+            instance_ids=None, max_concurrency='', max_errors='',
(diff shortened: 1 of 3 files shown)


moto/ssm/models.py
```diff
@@ -58,11 +59,74 @@ def describe_response_object(self, decrypt=False):
         return r
 
 
+MAX_TIMEOUT_SECONDS = 3600
+
+
+class Command(BaseModel):
+    def __init__(self, comment='', document_name='', timeout_seconds=MAX_TIMEOUT_SECONDS,
+            instance_ids=[], max_concurrency='', max_errors='',
```
spulec: We should try to avoid [default mutable arguments](http://docs.python-guide.org/en/latest/writing/gotchas/#mutable-default-arguments)

mliudev: Nice! I didn't know about this at all. Great catch. I'll fix this right up. I've learned something new today.

mliudev:
@spulec Made the requested changes. Not sure why coverage went down though.

spulec:
Awesome. Thanks for all your work on this and welcome to Moto!

Issue #1606, opened 2018-05-02
Batch jobs do not report log streams


theothertom:
In my tests, I am attempting to mock getting the log stream for a failed AWS Batch job. I have code like the following:
```python 
   client = boto3.client('batch')
    compute_environment_arn = create_mock_compute_environment()
    queue_arn = create_mock_job_queue(
        compute_environment_arn=compute_environment_arn)
    job_arn = create_mock_job_definition()
    response = client.submit_job(
        jobName='fake-job',
        jobQueue=queue_arn,
        jobDefinition=job_arn,
    )
```

Once the mocked job finishes, I'm expecting to be able to get the `logStreamName` property of the job.

It appears that the property is not available.

PR #1607, opened 2018-05-02
1606 Add additional fields to Batch job status endpoint

theothertom:
Adds what I needed to fix https://github.com/spulec/moto/issues/1606

Note: this doesn't add all the data that could be returned from AWS for Batch job status, just a few extra fields.


spulec:
Thanks for this and welcome to Moto!

PR #1609, opened 2018-05-03
adding reallocate address support eip

sepulworld:
Usage:

from boto3 docs:
```
response = client.allocate_address(
    Domain='vpc'|'standard',
    #Adding support for address 
    Address='string', 
    DryRun=True|False
)
```
```
Address (string) -- [EC2-VPC] The Elastic IP address to recover. <-- adding support for this.
```
We will just assume the address is available for recovery.


sepulworld:
I'm going to close this and create an issue to discuss best way to approach this change: https://github.com/spulec/moto/issues/1632

Issue #1610, opened 2018-05-03
CopyObject broken with non VersionId query string params


thehesiod:
aiobotocore has a test where it's doing a copy from with `Key=a+b/notVersionId=bar`.  This ends up creating a key with that literally that key name, and apparently this works, this is broken in moto because in `_key_response_put` it breaks the url into parts, but doesn't put back non-versionId query params, so you get an assert with a null key

thehesiod:
so the issue was when you did something like copy_object(Bucket='foo', Key='bar?versionID=1', CopySource=dict(Bucket='foo', Key='bar?versionID=1', VersionId=1).  You need to parse the unquoted string first

PR #1611, opened 2018-05-03
fix for issue 1610

thehesiod:
fixes https://github.com/spulec/moto/issues/1610 and issue when removing versioned keys


spulec:
Thanks!

Issue #1615, opened 2018-05-03
CloudWatch metrics: Support statistic-values in put-metric-data


barryoneill:
When putting statistic data (and not single values), moto returns a 500 error.  

*Version: 1.3.3*

- when `--value` is being supplied, everything is fine.  
- when `--value` is not present (e.g. when using `--statistic-values` instead - see the example in [Amazon's Docs](https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/PublishMetrics.html)), a 500 error occurs.

### to reproduce

In one terminal, leave moto running cloudwatch:

```
$ moto_server cloudwatch
 * Running on http://127.0.0.1:5000/ (Press CTRL+C to quit)
```

In another, issue the following commands to see the difference (again, following the example in [Amazon's Docs])(https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/PublishMetrics.html):

- Success Case

```
$ aws --endpoint=http://localhost:5000 cloudwatch put-metric-data --metric-name RequestLatency --namespace GetStarted --timestamp 2016-10-14T20:30:00Z --value 235 --unit Milliseconds

// server logs:
127.0.0.1 - - [03/May/2018 19:22:10] "POST / HTTP/1.1" 200 -
```

- failure case

```
$ aws --endpoint=http://localhost:5000 cloudwatch put-metric-data --metric-name RequestLatency --namespace GetStarted --timestamp 2016-10-14T21:30:00Z --statistic-values Sum=577,Minimum=65,Maximum=189,SampleCount=5 --unit Milliseconds

Unable to parse response (syntax error: line 1, column 54), invalid XML received:
b'<!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 3.2 Final//EN">\n<title>500 Internal Server Error</title>\n<h1>Internal Server Error</h1>\n<p>The server encountered an internal error and was unable to complete your request.  Either the server is overloaded or there is an error in the application.</p>\n'

// server logs:
127.0.0.1 - - [03/May/2018 19:22:13] "POST / HTTP/1.1" 500 -
Error on request:
Traceback (most recent call last):
  File "/home/vm/.local/lib/python2.7/site-packages/werkzeug/serving.py", line 270, in run_wsgi
    execute(self.server.app)
  File "/home/vm/.local/lib/python2.7/site-packages/werkzeug/serving.py", line 258, in execute
    application_iter = app(environ, start_response)
[...]

PR #1616, opened 2018-05-03
Issue 1615 - missing Value should not kill put_metric_data

barryoneill:
While this doesn't implement statistics, it should stop `put_metric_data` from erroring on a `KeyValue` when the `Value` attribute is missing.  

Issue: https://github.com/spulec/moto/issues/1615


spulec:
Thanks!

PR #1617, opened 2018-05-04
Implement some of cognito-idp

jbergknoff-10e:
Hi, this PR implements parts of [Cognito Identity Provider](https://docs.aws.amazon.com/cognito-user-identity-pools/latest/APIReference/Welcome.html). The API is huge, so this PR just covers the endpoints that are used in the application I'm working on.

This PR pulls in [python-jose](https://pypi.org/project/python-jose/) for the sake of signing JWTs for the tokens generated by the fake Cognito. This doesn't require any additional OS packages to be installed.

There's one important test missing, which will verify that the access/id tokens returned by the fake Cognito are cryptographically correct. I'll add that test ASAP.


jbergknoff-rival:
The Travis failures seem to be unrelated to the PR, something about not being able to download packages from PyPi. I can't find an option to retry the build without pushing something.

PR #1618, opened 2018-05-04
Accept paths to user-controlled SSL cert

jbergknoff-rival:
This PR makes it so the user can control the SSL certificate used by moto_server. This is particularly useful to whitelist the certificate when writing blackbox tests: for example, being able to pass in an `AWS_CA_BUNDLE` for boto, or `REQUESTS_CA_BUNDLE` for requests, which points at your self-signed certificate.

The existing behavior is unchanged: run with an auto-generated SSL certificate by doing
```
$ moto_server --ssl
```

Now you can alternately specify the certificate and private key files:
```
$ moto_server --ssl-key key.pem --ssl-cert cert.pem
```

### DIFF
--- a/moto/server.py
+++ b/moto/server.py
@@ -186,9 +186,18 @@ def main(argv=sys.argv[1:]):
     parser.add_argument(
         '-s', '--ssl',
         action='store_true',
-        help='Enable SSL encrypted connection (use https://... URL)',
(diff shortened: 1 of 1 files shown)


spulec:
This is great, thanks. Welcome to Moto!

PR #1620, opened 2018-05-04
Allow dashes, underscores in SNS topic creates for first/last characters

### LINKED ISSUE #1619: Moto CreateTopic for SNS uses incorrect regex
AWS allows me to create an SNS topic with leading underscores (`__app-ses-bounces`).

Moto (1.3.3) does not.

The cause is the regex incorrectly insists that the topic name may not begin or end with underscores, which makes no sense given that AWS documentation states:

> Topic names are limited to 256 characters. Alphanumeric characters plus hyphens (-) and underscores (_) are allowed. Topic names must be unique within an AWS account.

That's all the restrictions in AWS SNS provides. I have personally tested topics that starts and ends with `-` and `_`.

According to http://boto3.readthedocs.io/en/latest/reference/services/sns.html#SNS.Client.create_topic one needs at least one character.

benjolitz:
This change updates the regex to simply `^[a-zA-Z0-9_-]{1,256}$` as AWS does not have prohibitions against starting/ending the topic name with underscores and/or dashes.

Fixes #1619

### DIFF
--- a/moto/sns/models.py
+++ b/moto/sns/models.py
@@ -240,7 +240,7 @@ class SNSBackend(BaseBackend):
         self.sms_attributes.update(attrs)
 
     def create_topic(self, name):
-        fails_constraints = not re.match(r'^[a-zA-Z0-9\_\-]{0,256}$', name)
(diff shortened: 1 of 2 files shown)


spulec:
Awesome; thanks!

PR #1622, opened 2018-05-05
Support optional Source, parse from header

### LINKED ISSUE #1621: SES support does not parse Source or Message['from'] for email address
The `From` header on MIME Messages can be either `email@domain` or `Name <email@domain>`. For my tests, I have a `From` header of `=?utf-8?q?Name?= <email@domain.com>` which is a valid format style for production emails.

Furthermore, AWS SES allows for either a `Source` (which overrides any Message['From']) or a Message['From'] to be specified.

benjolitz:
The Email ``from`` header is either formatted as ``name <address>`` or ``address``.

This commit will use `parseaddr` to extract a ``(name, address)`` tuple, which we will use the ``address`` to check if it's verified.

Also support the case where ``Source`` is omitted (which AWS requires the ``from`` header to be set).

Fixes #1621

### DIFF
--- a/moto/ses/models.py
+++ b/moto/ses/models.py
@@ -1,6 +1,7 @@
 from __future__ import unicode_literals
 
 import email
+from email.utils import parseaddr
 
 from moto.core import BaseBackend, BaseModel
 from .exceptions import MessageRejectedError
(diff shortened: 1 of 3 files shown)


spulec:
Thanks for this and welcome to Moto!

Issue #1632, opened 2018-05-09
Allocate_address provide specific IP address to recover


sepulworld:
First off, thanks for putting together moto. Great project! 

I would like to help contribute by adding support for EIP allocate_address with a given public address. 

Per Boto3 docs under https://boto3.readthedocs.io/en/latest/reference/services/ec2.html#EC2.Client.allocate_address

Address (string) -- [EC2-VPC] The Elastic IP address to recover. 

AWS offers the ability to recover a specified public IP address. I would like to be able to test certain behavior when recovering a specific IP.

Basically, I believe we just need to bypass the current allocate_address behavior if the ```Address``` parameter is passed in and perhaps we should validate it is an IP address. Looking for advice on where to start on this. I tried a few things but haven't found the right solution yet.

sepulworld:
Test describing what I'm looking to add to moto

```
@mock_ec2
def test_specific_eip_allocate_vpc():
    """Allocate VPC EIP with specific public address"""
    service = boto3.resource('ec2', region_name='us-west-1')
    client = boto3.client('ec2', region_name='us-west-1')

    vpc = client.allocate_address(Domain="vpc", Address="127.38.43.222")
    vpc['Domain'].should.be.equal("vpc")
    vpc['PublicIp'].should.be.equal("127.38.43.222")
    logging.debug("vpc alloc_id:".format(vpc['AllocationId']))
```

PR #1636, opened 2018-05-11
adding Address reallocate capability for EIP

sepulworld:
AWS offers the ability to recover a specified public IP address. I would like to be able to test certain behavior when recovering a specific EIP public IP.

Available in boto3 for EIP address_allocate. 

Reference to issue: https://github.com/spulec/moto/issues/1632


sepulworld:
Thanks! Curious, when do you typically do releases? Monthly?

PR #1639, opened 2018-05-15
Bugfix: describe_event_bus()['Policy'] should be JSON string, not object

cpitchford:
boto3.client('events')['Policy'] must be a string as per:
http://boto3.readthedocs.io/en/latest/reference/services/events.html#CloudWatchEvents.Client.describe_event_bus

Adding json.dumps() around the policy value ensures we do not return an unexpected dict

import boto3
from moto import mock_events

real_events = boto3.client('events')
real_bus_policy_json = real_events.describe_event_bus()['Policy']

mock = mock_events()
mock.start()

moto_events = boto3.client('events')
mock_bus_policy_json = moto_events.describe_event_bus()['Policy']

print("Real bus policy returned as %{0}".format(str(type(real_bus_policy_json))))
print("Mock bus policy returned as %{0}".format(str(type(mock_bus_policy_json))))

### DIFF
--- a/moto/events/models.py
+++ b/moto/events/models.py
@@ -1,5 +1,6 @@
 import os
 import re
+import json
 
 from moto.core.exceptions import JsonRESTError
 from moto.core import BaseBackend, BaseModel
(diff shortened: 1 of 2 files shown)


spulec:
Thanks and welcome to Moto!

PR #1641, opened 2018-05-15
Bugfix events put permission action filter

cpitchford:
The Action parameter for events.put_permission must be fully qualified as per:

http://boto3.readthedocs.io/en/latest/reference/services/events.html#CloudWatchEvents.Client.put_permission

It expects 'PutEvents' where as it should expect 'events:PutEvents'

This corrects this behaviour


spulec:
Thanks!

Issue #1588, opened 2018-04-20
Cannot create Lambda function in Localstack via Serverless


albegali:
`moto` may not build the ARN correctly on the CloudFormation output XML.
Exemple: (see: temyers/serverless-localstack#24 (comment))

```
<DescribeStacksResponse>
  <DescribeStacksResult>
   <Stacks>
     <member>
       <Outputs>
         <member>
           <OutputKey>HelloLambdaFunctionQualifiedArn</OutputKey>
           <OutputValue><moto.awslambda.models.LambdaVersion object at 0x7f82f8363090></OutputValue>
         </member>
         <member>
           <OutputKey>ServiceEndpoint</OutputKey>
           <OutputValue>https://<moto.apigateway.models.RestAPI object at 0x7f82f8363c10>.execute-api.us-east-1.amazonaws.com/local</OutputValue>
         </member>
       </Outputs>
     </member>
   </Stacks>
 </DescribeStacksResult>
</DescribeStacksResponse>
```

(some members were suppressed on this sample above)

I guess that `<moto.awslambda.models.LambdaVersion object at 0x7f82f8363090>` should had been coerced to its ARN, and <moto.apigateway.models.RestAPI object at 0x7f82f8363c10> to its URL.

(refs: https://github.com/localstack/localstack/issues/711)

johnbarry343:
@alanjds Did you end up making a fork for this?

alanjds:
No. Forked only localstack/localstack#759. The described patch here was not properly "coded" anywhere but on my virtualenv on my machine.

alanjds:
@johnbarry343 Done: PR https://github.com/spulec/moto/pull/1643

spulec:
Fixed with https://github.com/spulec/moto/pull/1643

alanjds:
Just noticed: Maybe `moto.apigateway.models.RestAPI` needs a new `__repr__` too:

```xml
<member>
   <OutputKey>ServiceEndpoint</OutputKey>
   <OutputValue>https://<moto.apigateway.models.RestAPI object at 0x7f82f8363c10>.execute-api.us-east-1.amazonaws.com/local</OutputValue>
</member>
```
@albegali: Should reopen?

spulec:
I just pushed https://github.com/spulec/moto/commit/7951eeddc365130c9606718dc665ad4f944d4687 which should fix that. Let me know if you are still seeing issues.

alanjds:
Thanks @spulec

PR #1643, opened 2018-05-15
LambdaVersion providing ARN on __repr__

alanjds:
Just a dirty patch

### DIFF
--- a/moto/awslambda/models.py
+++ b/moto/awslambda/models.py
@@ -445,6 +445,9 @@ class LambdaVersion(BaseModel):
     def __init__(self, spec):
         self.version = spec['Version']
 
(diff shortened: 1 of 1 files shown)


wyvern8:
@alanjds any progress on this one?

alanjds:
Well, it fixes the problem pointed on #1588, but I had not pursued a full test.

And to be completely transparent, we had settled to use real AWS Lambdas on our CI for now. Then I am out of priority to fix localstack for now.

spulec:
Thanks!

beau6183:
When will this dirty patch make it into a release? Seems like  localstack/localstack#649 is being held up by it, unless I'm missing something...

PR #1656, opened 2018-05-30
Add MessageGroupId support to SQS queues (fixes #1655)

### LINKED ISSUE #1655: SQS backend does not honor MessageGroupId
The SQS backend currently does not support the message attribute `MessageGroupId` (`Message.queue_id`).

The attribute basically allows creating sub-queues within a fifo-queue which are separated by the message group. SQS makes sure to have at most one concurrent consumer receiving a message for each group, guaranteeing that all messages in the group are consumed in order. Once a message is either deleted or returned to the queue, the next one with equal group id can be consumed.

A possible test case might look like this

```python3
@mock_sqs
def test_receive_messages_with_message_group_id():
    sqs = boto3.resource('sqs', region_name='us-east-1')
    queue = sqs.create_queue(QueueName="blah")
    queue.send_message(
        MessageBody="message-1",
        MessageGroupId="group"
    )
    queue.send_message(
        MessageBody="message-2",
        MessageGroupId="group"
    )

    messages = queue.receive_messages()
    messages.should.have.length_of(1)
    message = messages[0]

    # previous message is still active
    # no other worker may fetch from the same group
    messages = queue.receive_messages(WaitTimeSeconds=0)
    messages.should.have.length_of(0)

    # message is now processed
    message.delete()

    # message group is available again
    messages = queue.receive_messages()
    messages.should.have.length_of(1)

```

https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/using-messagegroupid-property.html

Birne94:
#1655 

> The SQS backend currently does not support the message attribute MessageGroupId (Message.queue_id).
> 
> The attribute basically allows creating sub-queues within a fifo-queue which are separated by the message group. SQS makes sure to have at most one concurrent consumer receiving a message for each group, guaranteeing that all messages in the group are consumed in order. Once a message is either deleted or returned to the queue, the next one with equal group id can be consumed.
> 

In this PR the queue takes note of pending tasks and their message groups. If a new message is fetched from the queue and a pending message with the same group already exists, it is skipped.

Additionally, I have moved the infinite-loop check in `SQSBackend::receive_messages` below the message loop so I could account for messages filtered due to their group. This does not change any existing behavior.

### DIFF
--- a/moto/sqs/models.py
+++ b/moto/sqs/models.py
@@ -180,6 +180,7 @@ class Queue(BaseModel):
         self.permissions = {}
 
         self._messages = []
+        self._pending_messages = set()
 
         now = unix_time()
         self.created_timestamp = now
@@ -208,6 +209,16 @@ class Queue(BaseModel):
         # Check some conditions
         if self.fifo_queue and not self.name.endswith('.fifo'):
             raise MessageAttributesInvalid('Queue name must end in .fifo for FIFO queues')
+
+    @property
+    def pending_messages(self):
+        return self._pending_messages
+
+    @property
+    def pending_message_groups(self):
+        return set(message.group_id
+                   for message in self._pending_messages
+                   if message.group_id is not None)
 
     def _set_attributes(self, attributes, now=None):
         if not now:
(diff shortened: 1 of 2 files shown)


spulec:
This looks great! Thanks for the contribution.

Issue #1657, opened 2018-05-30
Fn::ImportValue in CF mock does not handle missing exports gracefully


tmwong2003:
When creating a CloudFormation stack that imports a nonexistent export, `moto` throws an exception that seems unrelated to the issue at hand. For example, given the following CF template that imports the nonexistent export `SimpleSecurityGroup`:

```
Resources:
  SimpleInstance:
    Type: AWS::EC2::Instance
    Properties:
        ImageId: ami-03cf127a
        InstanceType: t2.micro
        SecurityGroups: !Split [',', !ImportValue SimpleSecurityGroup]
```

`moto` raises the following exception:

```
Traceback (most recent call last):
# ...
  File "/Users/theodorew/moto/moto/cloudformation/parsing.py", line 206, in clean_json
    values = [x.value for x in resources_map.cross_stack_resources.values() if x.name == cleaned_val]
AttributeError: 'list' object has no attribute 'values'
```

The underlying issues are two-fold:

1. `FakeStack` in `moto.cloudformation.models` sets the default value of the `cross_stack_resources` attribute to an empty list instead of an empty dict.
1. When parsing the `Fn::ImportValue` intrinsic function, `moto.cloudformation.parsing.clean_json()` simply drops out of the `if` block responsible for processing `Fn::ImportValue`, instead of raising an exception or returning some kind of empty value.

The sample code to trigger this issue is as follows:

```
import boto3
from moto import mock_cloudformation

EC2_STACK_NAME = 'simple-ec2-stack'
EC2_TEMPLATE = """
Resources:
  SimpleInstance:
    Type: AWS::EC2::Instance
    Properties:
        ImageId: ami-03cf127a
        InstanceType: t2.micro
        SecurityGroups: !Split [',', !ImportValue SimpleSecurityGroup]
"""


with mock_cloudformation():
    client = boto3.client('cloudformation')
    client.create_stack(StackName=EC2_STACK_NAME, TemplateBody=EC2_TEMPLATE)
```

spulec:
Fixed with https://github.com/spulec/moto/pull/1658

PR #1658, opened 2018-05-30
Fixed CF creation to trap imports of non-existent values

tmwong2003:
This PR addresses the issues raised in #1657:

1. If a `FakeStack` has no cross-stack resources (i.e., those created by previous `create_stack` operations), it sets `self.cross_stack_resources` to an empty dict instead of an empty list.
2. If a CF template tries to import a non-existent export value, `create_stack` will raise a `ExportNotFound` exception.

This PR also adds a test when creating a stack that imports (existing) export values, and a test when creating a stack that import nonexistent export values.

### DIFF
--- a/moto/cloudformation/exceptions.py
+++ b/moto/cloudformation/exceptions.py
@@ -33,6 +33,18 @@ class MissingParameterError(BadRequest):
         )
 
 
+class ExportNotFound(BadRequest):
+    """Exception to raise if a template tries to import a non-existent export"""
+
+    def __init__(self, export_name):
+        template = Template(ERROR_RESPONSE)
+        super(ExportNotFound, self).__init__()
+        self.description = template.render(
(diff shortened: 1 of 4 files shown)


spulec:
Thanks for this! It looks like there are some tests failing on Travis, but I will be happy to merge if we can get those working.

tmwong2003:
Hi there - thank you - I fixed the issue and Travis testing now passes. The issue there was that I had a default region configured in my environment, so I didn't catch the missing `region_name` parameter in the `boto3.client()` calls.

spulec:
Excellent! Thanks for your contribution and welcome to Moto!

tmwong2003:
You’re welcome!

-- 
Theodore Wong <tmw@tmwong.org>
http://www.tmwong.org

PR #1669, opened 2018-06-04
Extend support for ECS

### LINKED ISSUE #1668: ECS/create_service should support Load Balancers
## Feature description

As [the documentation says](http://boto3.readthedocs.io/en/latest/reference/services/ecs.html#ECS.Client.create_service), 

> you can optionally run your service behind a load balancer

The function in `boto3` accepts a parameter called `loadBalancers`, containing an array of dicts.

## Current implementation

As of the current version of moto (1.3.3), when I call `create_service()`:

- the parameter `loadBalancers`  is ignored
- the response of this call always contain an empty array for the key `"loadBalancers"`

I'm going to create a PR to support this feature ASAP, together with a new test scenario.

fferrara:
Add support for creating ECS services with Load Balancer and fixes #1668

### DIFF
--- a/moto/ecs/models.py
+++ b/moto/ecs/models.py
@@ -177,7 +177,7 @@ class Task(BaseObject):
 
 class Service(BaseObject):
 
-    def __init__(self, cluster, service_name, task_definition, desired_count):
+    def __init__(self, cluster, service_name, task_definition, desired_count, load_balancers=None):
         self.cluster_arn = cluster.arn
         self.arn = 'arn:aws:ecs:us-east-1:012345678910:service/{0}'.format(
             service_name)
@@ -199,7 +199,7 @@ class Service(BaseObject):
                 'updatedAt': datetime.now(pytz.utc),
             }
         ]
-        self.load_balancers = []
+        self.load_balancers = load_balancers if load_balancers is not None else []
         self.pending_count = 0
 
     @property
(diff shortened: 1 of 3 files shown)


tests/test_ecs/test_ecs_boto3.py
```diff
@@ -2015,3 +2015,62 @@ def _fetch_container_instance_resources(container_instance_description):
     registered_resources['PORTS'] = \
     [x['stringSetValue'] for x in registered_resources_list if x['name'] == 'PORTS'][0]
     return remaining_resources, registered_resources
+
+
+@mock_ecs
+def test_create_service_load_balancing():
+    client = boto3.client('ecs', region_name='us-east-1')
+    _ = client.create_cluster(
```
terrycain: More stylistc, but no need to assign to _ if its not going to be used.

moto/ecs/models.py
```diff
@@ -652,18 +652,21 @@ def stop_task(self, cluster_str, task_str, reason):
...
         if cluster_name in self.clusters:
             cluster = self.clusters[cluster_name]
         else:
             raise Exception("{0} is not a cluster".format(cluster_name))
         task_definition = self.describe_task_definition(task_definition_str)
         desired_count = desired_count if desired_count is not None else 0
+        load_balancers = load_balancers if load_balancers is not None else []
+
         service = Service(cluster, service_name,
                           task_definition, desired_count)
```
terrycain: Would it not make more sense to add load_balancers=None to the constructor of `Service` and then if provided extend `self.load_balancers`?

terrycain:
@JackDanger @spulec When specifing a load balancer when creating an ecs service, do we want to validate that loadbalancer arn? Means more mock setup, but would also catch specifying the wrong lb

tests/test_ecs/test_ecs_boto3.py
fferrara: Completely right!

moto/ecs/models.py
fferrara: Initially, my solution was as you said. Agree that makes more sense, changing back

spulec:
Yes, ideally we should add validation for arn. I don't think that is enough to block getting these great contributions merged in though.

Thanks for the work @fferrara and welcome to Moto!

Issue #1674, opened 2018-06-06
Some tests not being run in travis


william-richard:
I noticed that pytest is being imported in the dynamodb tests, but this is not being caught by travis!

https://github.com/spulec/moto/blob/master/tests/test_dynamodb2/test_dynamodb.py#L9

I'm not sure why isn't travis running these tests, or what other tests it might be missing. When I run `make test` locally, the `test_dynamodb2` directory runs, and fails because of an import error.

Is this a problem with nose? or how we're using it? or something else?

spulec:
Hmm, maybe Travis has it installed by default?

william-richard:
I don't think that's the issue.  I went through the travis output, and I could not find any mention of the`test_dynamodb2` in the section of the output that lists when each test is run.  That's why I think tests are not being run, rather than pytest being installed already in travis (though it does look like pytest might be installed already).

spulec:
It looks like nose only print the module name, not the folder name so we will see `test_dyanmodb.` for both the dynamo and dynamo2 tests.

You can confirm by seeing two different `test_list_tables_layer_1` here: https://api.travis-ci.org/v3/job/441495243/log.txt

Going to close, but feel free to reopen if I'm missing something.

PR #1675, opened 2018-06-06
Make dynamodb table indexes a list

william-richard:
When you create a global index via the table resource `update` command, it was getting stored as a `dict_values` object, but most of the codebase expected a list.

I added a test and fixed the problem.

I also fixed an invalid import of pytest - as reported in #1674, travis isn't running the `test_dynamodb2` tests, so this import error was not caught.  I'm not sure why travis isn't running those tests, nor do I know what other tests it is missing.


william-richard:
Any advice on how to debug these travis failures? I don't see anything obvious in the output

spulec:
Looks to be passing now; not positive what the issue was.

Thanks for the help!

PR #1687, opened 2018-06-15
Fix ECR and bug fixes

### LINKED ISSUE #1673: Bug: Implementation Coverage script not working as expected
For a project of mine I needed to use the _iot-data_ service. Looking into [`IMPLEMENTATION_COVERAGE.md`](https://github.com/spulec/moto/blob/master/IMPLEMENTATION_COVERAGE.md) it looked like the service wasn't implemented yet. After doing a bit of searching I found #1303, where the _iot-data_ service was added to moto.

https://github.com/spulec/moto/blob/0f816566df3aefc20b9054e4dc9f7ca7373bc6f6/IMPLEMENTATION_COVERAGE.md#L2527-L2532

Being interested why this isn't reflected in `IMPLEMENTATION_COVERAGE.md` I dug a bit deeper. I found out that coverage information is not only missing for _iot-data_, but also for _cognito-identity_, and _cognito-idp_. The all have in common that their moto attribute represation is without the "-" i.e. iotdata, cognitoidentity, cognitoidp. If this "renaming pattern" can be seen as set (default) adding following line between these two lines:
https://github.com/spulec/moto/blob/4198abfd24a3aab49a8b83640ca265870e2688d1/scripts/implementation_coverage.py#L8-L9

`service_name = service_name.replace("-", "") if "-" in service_name else service_name`

This will result in having correct coverage info on _iot-data_, _cognito-identity_, as well as _cognito-idp_ services.

Should I make a pull request for this?

### LINKED ISSUE #1684: ECR `describe_image` returns None values
### Actual
`describe_image` returns `imageTags` even when no tags were set. This will result in returning `[ None ]`

   ```
   {
   ...
      "imageTags": [ None ]
   ...
   }
   ```

### Expected
 `describe_image` returns `imageTags` only when they exist, otherwise they are just left out of the response, which is the way AWS behaves on this method!
Generally I believe that `None` values should be left out of the response.

### LINKED ISSUE #1685: ECR `describe_image` should return list of imageTags
### Actual
Creating images `put_image` with the same `image_manifest` and different tags results in different images being created and returned through `describe_image`:
   ```
   {
    "imageDetails": [
        {
            "imageSizeInBytes": 55600,
            "imageDigest": "sha256:abcdefg",
            "imageTags": [
                "v1"
            ],
            "registryId": "1234567890",
            "repositoryName": "test",
            "imagePushedAt": 1505819275.0
        },
        {
            "imageSizeInBytes": 55600,
            "imageDigest": "sha256:abcdefg",
            "imageTags": [
                "latest"
            ],
            "registryId": "1234567890",
            "repositoryName": "test",
            "imagePushedAt": 1508933240.0
        }
    ]
   }
   ```

### Expected
Creating and updating images with `put_image` with the same `image_manifest` should have the same `image_digest` and should be returned with a list of the corresponding tags through `describe_image` i.e.:
   ```
   {
    "imageDetails": [
        {
            "imageSizeInBytes": 55600,
            "imageDigest": "sha256:abcdefg",
            "imageTags": [
                "v1",
                "latest"
            ],
            "registryId": "1234567890",
            "repositoryName": "test",
            "imagePushedAt": 1505819275.0
        }
    ]
   }
   ```

sthuber90:
Fixes #1673 
Fixes #1684 
Fixes #1685

### DIFF
--- a/IMPLEMENTATION_COVERAGE.md
+++ b/IMPLEMENTATION_COVERAGE.md
@@ -814,17 +814,17 @@
 - [ ] update_team_member
 - [ ] update_user_profile
 
-## cognito-identity - 0% implemented
-- [ ] create_identity_pool
+## cognito-identity - 22% implemented
+- [X] create_identity_pool
 - [ ] delete_identities
 - [ ] delete_identity_pool
 - [ ] describe_identity
 - [ ] describe_identity_pool
-- [ ] get_credentials_for_identity
-- [ ] get_id
+- [X] get_credentials_for_identity
+- [X] get_id
 - [ ] get_identity_pool_roles
 - [ ] get_open_id_token
-- [ ] get_open_id_token_for_developer_identity
+- [X] get_open_id_token_for_developer_identity
 - [ ] list_identities
 - [ ] list_identity_pools
 - [ ] lookup_developer_identity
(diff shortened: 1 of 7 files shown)


sthuber90:
Sound good to me. Let me know if I can be of any help

spulec:
Looks good; thanks!

PR #1692, opened 2018-06-21
BugFix: Exception in ListObjectVersions with a Prefix that Returns an Empty Result Set

sanjeevsuresh:
… otherwise an exception is raised when trying to iterate over an empty list


spulec:
Thanks and welcome to Moto!

Issue #1599, opened 2018-04-25
Support Route53 alias records


2rs2ts:
Please add support for alias records. You can see in the route53/responses.py that there is code reading upserted records for their AliasTarget and setting them to be the record set's `ResourceRecords` list:

https://github.com/spulec/moto/blob/f18259d49e5e66929a793ec8105d32ccfcc70d21/moto/route53/responses.py#L135-L137

I believe this is wrong; you should maintain the alias as an alias. I think normally you can't create an alias and also give it `ResourceRecords` (for example, [the CloudFormation documentation](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-route53-recordset.html#cfn-route53-recordset-resourcerecords) says not to specify it if you specify an `AliasTarget`.)

spulec:
Thanks for opening; I'm going to set this as a feature request.

elliottmb:
I've just run into this issue when writing tests for an application that manipulates alias targets. 

What I expect to see
```
{
    "AliasTarget": {
        "HostedZoneId": "FAKEZONE", 
        "EvaluateTargetHealth": false, 
        "DNSName": "other.example.com."
    }, 
    "Type": "A", 
    "Name": "test.example.com."
}
```

and what moto is responding with:

```
{
    'Name': 'test.example.com',
    'Type': 'A',
    'ResourceRecords': [
        {
            'Value': 'other.example.com.'
        }
    ]
}
```

This has meant that part of my code can't be tested using moto.

I can open an PR to try to fix this, if that would be the best route.

terrycain:
@elliottmb Sure open up a PR and I'll review it.

bblommers:
As far as I can tell this was fixed by #2082, so I'll close this. Please let us know if there's still anything wrong/outstanding though.

PR #1694, opened 2018-06-22
WIP: Support Route53 alias records

elliottmb:
Work in Progress MR for #1599 

Also looks like there is a use before assignment that was occurring in `moto/route53/responses.py` on line 137 before my changes. As the RecordSet class handles the empty case cleanly, and the `elif` case is not correct, I simplified the assignment. 

I haven't added tests, so marked it with WIP. I did have a little proof of concept script based on the problem I'd run into, which is

```
import boto3
import moto

mroute = moto.mock_route53()
mroute.start()

client = boto3.client('route53')

hosted_zone_response = client.create_hosted_zone(
	Name='example.com',
	CallerReference='testing'
)
hosted_zone_id = hosted_zone_response.get('HostedZone', {}).get('Id')

create_response = client.change_resource_record_sets(
	HostedZoneId=hosted_zone_id,
	ChangeBatch={
		'Changes': [
			{
				'Action': 'UPSERT',
				'ResourceRecordSet': {
					'Name': 'test.example.com',
					'Type': 'A',
					'AliasTarget': {
						'DNSName': 'other.example.com',
						'HostedZoneId': hosted_zone_id,
						'EvaluateTargetHealth': False
					}
				}
			}
		]
	}
)

print(client.list_resource_record_sets(HostedZoneId=hosted_zone_id))
```

Before these changes, the response would match that detailed in my comment on #1599. It now returns correctly.

### DIFF
--- a/moto/route53/models.py
+++ b/moto/route53/models.py
@@ -85,6 +85,7 @@ class RecordSet(BaseModel):
         self.health_check = kwargs.get('HealthCheckId')
         self.hosted_zone_name = kwargs.get('HostedZoneName')
         self.hosted_zone_id = kwargs.get('HostedZoneId')
+        self.alias_target = kwargs.get('AliasTarget')
 
     @classmethod
     def create_from_cloudformation_json(cls, resource_name, cloudformation_json, region_name):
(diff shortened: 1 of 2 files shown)


markchalloner:
@spulec, @elliottmb. 

If you agree, please close in favor of #2082 which fixes an issue and adds tests.

elliottmb:
Closing in favor of #2082

PR #1696, opened 2018-06-24
Add instance type validator

alecrajeev:
This pull request attempts to solve the problem of launching an invalid instance type.

A beautiful soup script from http://www.ec2instances.info pulls of the available instance types and puts them into a hash table that is a 2d numpy array. This is then stored as a csv file in ec2/resources/instance_types_hash.csv.

Each row of the hash table contains several instance types. To get to the row a small hash function is used that calculates the corresponding row based on the instance family. This way only a few instance family's instance types have to be traversed when validating an instance type.

This hash table is then imported from the csv file as a 2d numpy array to ec2/models.py. There is an additional inner function to add_instances called invalid_instance_type that will check if an instance is invalid.

If the instance type is invalid, then it will throw an InvalidParameterValueError.

If the instance_type is None, then it will use the default value of m1.small that is also the default in other parts of moto.

I added the python module [numpy](https://github.com/numpy/numpy) as a required module because this pr uses it heavily. It a highly useful library for manipulating numerical data and has many potential uses.


alecrajeev:
@terrycain any thoughts on this pr?

terrycain:
My only concern is adding numpy to the requirements as some people will consider it a heavy dependency, @JackDanger thoughts?

alecrajeev:
Yeah I can see that. Using numpy here is a bit of over-doing it in terms of storing the data when there are only around 100 instance types.

My plan is to eventually add reserved instances support. This will require the describe_reserved_instances_offerings. These offering ids are a bunch of csv data that has to be stored and process (one offering id for every region, instance type, tenancy, offering class etc.). Numpy is the best way to import and parse the large csv files. I've pretty much finished with the describe_reserved_instances_offerings part here https://github.com/alecrajeev/moto/pull/5/files.

So while numpy is really necessary at this stage, I think it will be very helpful in the near future. Numpy also allows you to do Vectorized operations, which can speed up many programs.

terrycain:
Yup totally agree. With that im happy to merge this @JackDanger ?

PR #1699, opened 2018-06-25
ECR: list_images should return RepositoryNotFoundException when repository is not found

### LINKED ISSUE #1698: ECR `list_images` should return exception when repository does not exist
### Actual
`moto` [will return](https://docs.aws.amazon.com/AmazonECR/latest/APIReference/API_ListImages.html) an empty list when the ECR repository passed to `list_images` does not exist.

### Expected
The ECR client returns a `RepositoryNotFoundException` exception when the specified repository does not exist in the registry.

See AWS [docs](https://docs.aws.amazon.com/AmazonECR/latest/APIReference/API_ListImages.html).

chezclem:
Fixes #1698

### DIFF
--- a/moto/ecr/models.py
+++ b/moto/ecr/models.py
@@ -200,17 +200,22 @@ class ECRBackend(BaseBackend):
         """
         maxResults and filtering not implemented
         """
-        images = []
-        for repository in self.repositories.values():
-            if repository_name:
-                if repository.name != repository_name:
-                    continue
+        repository = None
+        found = False
+        if repository_name in self.repositories:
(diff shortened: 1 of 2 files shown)


chezclem:
@spulec I'm not very familiar with the way `coveralls` calculates the coverage stats. The `list_images` function is fully covered by tests (unless I'm missing something). Happy to increase the coverage of another function if necessary.
Let me know what you think.
And thank you for open-sourcing `moto`!

spulec:
This looks great! Thanks for the help and welcome to Moto!

PR #1702, opened 2018-06-28
Updated index.rst to fix overflow

RobRoseKnows:
The EC2 endpoint status was overflowing and creating a scroll bar on my screen. It was bugging me so I fixed it via the GitHub web interface. Will test to ensure it builds correctly when I get home from work, but it does render correctly on GitHub's preview.


spulec:
Thanks for this. Do you think we should update all of the services?

RobRoseKnows:
Not sure what you mean, were there others that were overflowing or are the docs as a whole out of date?

Issue #1704, opened 2018-07-03
Unable to set subscribe attributes for SNS with subscribe function


MichaelBellBOM:
From:
https://boto3.readthedocs.io/en/latest/reference/services/sns.html#SNS.Client.subscribe
There is an optional Attributes keyword parameter that allows the setting of subscription attributes.

This feature is missing from moto and is preventing this from being mocked.

To reproduce mock the subscribe call for sns and then the get_subscription_attributes, to see that the attributes are not set.

spulec:
Fixed with #1705

PR #1705, opened 2018-07-03
Allow attributes to be set with subscribe command

MichaelBellBOM:
https://github.com/spulec/moto/issues/1704

I have run into an issue with mock the subscribe functionality from moto not recording the subscription attributes.  I have made an attempt to fix the issue myself, hence this pull request.


terrycain:
Looks good, good job :)

Issue #1513, opened 2018-03-15
S3 multipart upload completion error


JonWhitehead:
I have been trying to run the Multipart upload example against Moto from:
https://docs.aws.amazon.com/AmazonS3/latest/dev/llJavaUploadFile.html
Using the class UploadObjectMPULowLevelAPI.

When calling completeMultipartUpload, I get an error response:

com.amazonaws.services.s3.model.AmazonS3Exception: One or more of the specified parts could not be found. The part might not have been uploaded, or the specified entity tag might not have matched the part's entity tag. (Service: Amazon S3; Status Code: 400; Error Code: InvalidPart; Request ID: null; S3 Extended Request ID: null), S3 Extended Request ID: null

I believe the error is being thrown from the below code:

s3/models.py

    def complete(self, body):
        decode_hex = codecs.getdecoder("hex_codec")
        total = bytearray()
        md5s = bytearray()

        last = None
        count = 0
        for pn, etag in body:
            part = self.parts.get(pn)
            if part is None or part.etag != etag:
                raise InvalidPart()
                

The part.etag appears to contain a string with additional quotes e.g '"7319d066c5e41e4c25f3fc3cef366adb"' They are being removed on line 175, however I think they are causing the prior equality check on line 171 to fail.

The addtional quotes are introduced in FakeKey, line 97: return '"{0}"'.format(self._etag)
https://github.com/spulec/moto/blob/master/moto/s3/models.py#L97

Removing the outer quotes appears to fix the issue e.g return "{0}".format(self._etag)
I'm not sure whether this is an appropriate fix, it would be great it someone could take a look.
Thanks in advance.
Jon

bwidmeier:
I am also running into this issue. Anyone have any thoughts on this?

avizana:
I removed the outer quotes as well and everything is working just fine for me.

galvinograd:
+1

spulec:
Can one of your add a minimum failing test case? We have multiple tests that do multipart uploads so it would be helpful to understand what is different.

avizana:
I just used the multipart upload example from AWS site:

        AmazonS3 s3Client = AmazonS3ClientBuilder.standard()
                .withEndpointConfiguration(new AwsClientBuilder.EndpointConfiguration("some-kind-of-endpoint"))
                .withCredentials(new AWSStaticCredentialsProvider(new BasicAWSCredentials("user", "pass")))
                .disableChunkedEncoding()
[...]

PR #1707, opened 2018-07-03
Added support for multipart upload completion with unquoted etags

Whamied:
This pull request is to address the issue encountered in #1513 

My team encountered this issue while working with a library that built upon the Java AWS SDK. There was no eTag modification being done in the library, but we saw that on the completion step, our XML body had etags without quotes. 

Live S3 supports multipart upload completion in both cases, whether etags are quoted or not.

As @spulec requested, we added a minimum failing test case that now works with our proposed change.

@bwidmeier's comments about the Java SDK vs other SDKs reflect what we've seen. We understand the intention of moto is to mock out for boto specifically. But, with popular projects such as localstack developing against moto, it would be beneficial to work with as many SDK's as possible.

### DIFF
--- a/moto/s3/models.py
+++ b/moto/s3/models.py
@@ -175,11 +175,14 @@ class FakeMultipart(BaseModel):
         count = 0
         for pn, etag in body:
             part = self.parts.get(pn)
-            if part is None or part.etag != etag:
+            part_etag = None
+            if part is not None:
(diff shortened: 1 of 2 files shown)


spulec:
Moto is a misnomer now; I want it to work with all SDKs :)

This is good work. Thanks for helping us out and welcome to Moto!

PR #1709, opened 2018-07-06
dynamodb2 ComparisonOperator should treat NULL/NOT_NULL correctly

garyd203:
The [AWS documentation ](https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_Condition.html) says that a ComparisonOperator of NULL means the attribute should not exist, whereas NOT_NULL means that the attribute should exist. It explicitly says that an attribute with a
value of NULL is considered to exist, which contradicts our previous implementation. This affects both put_item and update_item in dynamodb2.

Note that I only updated dynamodb2, not dynamodb, because I wasn't sure if that was expected to have different behaviour. Let m  know if I should copy this across to dynamodb also.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,4 +14,5 @@ build/
 python_env
 .ropeproject/
 .pytest_cache/
+venv/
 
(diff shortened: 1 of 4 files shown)


terrycain:
Yeah that looks sane :)

PR #1713, opened 2018-07-10
APIGateway - Generate API key value when no value provided

fewstera:
On AWS when creating an API key if a value is not provided then AWS generates a value for you.

Example:

`aws apigateway create-api-key --name "Test"`

Returns:

```
{
    "name": "Test",
    "enabled": true,
    "value": "dsasDDsn7qweQWE2r327lq44nLRNq0AsdAd",
    "stageKeys": [],
    "lastUpdatedDate": 1531220553,
    "createdDate": 1531220553,
    "id": "twvchui76ac"
}
```

On moto the value was only generated when the `--generate-distinct-id` flag was passed, which is a bug.

This PR fixed this.

### DIFF
--- a/moto/apigateway/models.py
+++ b/moto/apigateway/models.py
@@ -300,11 +300,7 @@ class ApiKey(BaseModel, dict):
                  generateDistinctId=False, value=None, stageKeys=None, customerId=None):
         super(ApiKey, self).__init__()
         self['id'] = create_id()
-        if generateDistinctId:
-            # Best guess of what AWS does internally
-            self['value'] = ''.join(random.sample(string.ascii_letters + string.digits, 40))
-        else:
-            self['value'] = value
+        self['value'] = value if value else ''.join(random.sample(string.ascii_letters + string.digits, 40))
         self['name'] = name
         self['customerId'] = customerId
         self['description'] = description
--- a/tests/test_apigateway/test_apigateway.py
+++ b/tests/test_apigateway/test_apigateway.py
@@ -981,7 +981,7 @@ def test_api_keys():
     apikey['value'].should.equal(apikey_value)
 
     apikey_name = 'TESTKEY2'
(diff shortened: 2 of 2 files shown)


moto/apigateway/models.py
```diff
@@ -297,14 +297,10 @@ def _apply_operation_to_variables(self, op):
 class ApiKey(BaseModel, dict):
 
     def __init__(self, name=None, description=None, enabled=True,
-                 generateDistinctId=False, value=None, stageKeys=None, customerId=None):
```
terrycain: If someone passes in JSON with generateDistinctId, it will be given as an argument to the ApiKey class, and as it wont be a named parameter, it will fail.

moto/apigateway/models.py
```diff
@@ -297,14 +297,10 @@ def _apply_operation_to_variables(self, op):
...
+                 value=None, stageKeys=None, customerId=None):
         super(ApiKey, self).__init__()
         self['id'] = create_id()
-        if generateDistinctId:
-            # Best guess of what AWS does internally
-            self['value'] = ''.join(random.sample(string.ascii_letters + string.digits, 40))
-        else:
-            self['value'] = value
```
terrycain:
```
        if generateDistinctId or value is None:
            # Best guess of what AWS does internally
            self['value'] = ''.join(random.sample(string.ascii_letters + string.digits, 40))
        else:
            self['value'] = value
```

moto/apigateway/models.py
terrycain: Regardless, if its specified when the request is made, that payload will be expanded as for the class arguments and it wont exist and then an exception will be raised.

fewstera: Thank you, I've updated the PR to fix this.

fewstera:
@terrycain thank you for the review. I've re-added `generateDistinctId ` and all checks are passing now.

PR #1721, opened 2018-07-16
Added SecretsManager get_random_password mock

sepulworld:
Added SecretsManager get_random_password mock with tests

Ref: https://boto3.readthedocs.io/en/latest/reference/services/secretsmanager.html#SecretsManager.Client.get_random_password


spulec:
This is awesome. Thanks for your continued work!

Issue #1722, opened 2018-07-17
Customer Managed IAM Policy ARN Should Include Account Number 


nmische:
In the current moto implementation of IAM policies, both customer managed policies and AWS managed policies use `aws` in the policy ARN, however customer managed policies should include the AWS account number in the ARN.

spulec:
Closed with #1723

PR #1723, opened 2018-07-17
Adding account id to ManagedPolicy ARN

nmische:
Addresses #1722.


spulec:
Good catch; thanks and welcome to Moto!

PR #1724, opened 2018-07-18
feature: add parameters back to Message models

r1b:
These were removed in #560 but I think we should keep them around. I work with a test suite that makes liberal use of snapshots & I would like to use this style of testing with moto. It is useful to check that you sent the right thing to the right people.


spulec:
Looks great; thanks and welcome to Moto!

PR #1737, opened 2018-07-25
Skip freeze_time tests in server mode

garyd203:
Skip tests that use `freeze_time` when testing server mode. This is because freeze_time is unable to affect the time on the external server process, which makes the tests fragile and liable to incorrectly fail on a time comparison

### DIFF
--- a/tests/test_sqs/test_sqs.py
+++ b/tests/test_sqs/test_sqs.py
@@ -20,7 +20,6 @@ from moto import settings, mock_sqs, mock_sqs_deprecated
 from tests.helpers import requires_boto_gte, skip_in_server_mode
 import tests.backport_assert_raises  # noqa
 from nose.tools import assert_raises
-from nose import SkipTest
 
 
 @mock_sqs
@@ -840,9 +839,6 @@ def test_delete_message_after_visibility_timeout():
 @skip_in_server_mode
 @mock_sqs
 def test_batch_change_message_visibility():
-    if settings.TEST_SERVER_MODE:
-        raise SkipTest('Cant manipulate time in server mode')
-
     with freeze_time("2015-01-01 12:00:00"):
         sqs = boto3.client('sqs', region_name='us-east-1')
         resp = sqs.create_queue(
@@ -971,9 +967,6 @@ def test_create_fifo_queue_with_dlq():
 @skip_in_server_mode
 @mock_sqs
 def test_queue_with_dlq():
-    if settings.TEST_SERVER_MODE:
-        raise SkipTest('Cant manipulate time in server mode')
-
     sqs = boto3.client('sqs', region_name='us-east-1')
 
     with freeze_time("2015-01-01 12:00:00"):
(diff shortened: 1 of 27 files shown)


tests/test_sqs/test_sqs.py
```diff
@@ -837,9 +837,10 @@ def test_delete_message_after_visibility_timeout():
     assert new_queue.count() == 0
 
 
+@skip_in_server_mode
 @mock_sqs
 def test_batch_change_message_visibility():
-    if os.environ.get('TEST_SERVER_MODE', 'false').lower() == 'true':
+    if settings.TEST_SERVER_MODE:
```
terrycain: If the function has been decorated with skip_test then you dont need the if statement right?

tests/test_sqs/test_sqs.py
```diff
@@ -967,9 +968,10 @@ def test_create_fifo_queue_with_dlq():
         )
 
 
+@skip_in_server_mode
 @mock_sqs
 def test_queue_with_dlq():
-    if os.environ.get('TEST_SERVER_MODE', 'false').lower() == 'true':
+    if settings.TEST_SERVER_MODE:
```
terrycain: As above

tests/test_sqs/test_sqs.py
```diff
@@ -1153,9 +1155,10 @@ def test_receive_messages_with_message_group_id_on_requeue():
     messages[0].message_id.should.equal(message.message_id)
 
 
+@skip_in_server_mode
 @mock_sqs
 def test_receive_messages_with_message_group_id_on_visibility_timeout():
-    if os.environ.get('TEST_SERVER_MODE', 'false').lower() == 'true':
+    if settings.TEST_SERVER_MODE:
```
terrycain: As above

terrycain:
Other than those changes, this looks good. @JackDanger 2nd opinion

garyd203:
Thanks @terrycain, I have stripped out those duplicate checks now

garyd203:
hi @terrycain, is there anything else we need to do to get this change accepted? Ta!

garyd203:
Hi @terrycain Are we able to merge this PR now?

garyd203:
Hi @spulec and @terrycain Is there anything else I can do to get this patch accepted? If it is no longer relevant I am happy to abandon it.

Thanks heaps (and thanks for continuing to maintain moto).

spulec:
I'm a little worried about this. There are legitimate differences between server-mode and non-server mode and I'm worried about the loss of coverage here.

Is there another approach? Setting the time to freeze as an environment variable or something?

bpandola:
There have been a lot of updates to the codebase since this was opened, including a move from `nosetests` (upon which this new `@skip_in_server_mode` decorator had been based) to `pytest`.  Most of the tests that use `@freeze_time` have had explicit `TEST_SERVER_MODE` checks added over the years, which obviates the need to try to bring this PR up-to-date and fix the long list of now-conflicting files.

garyd203:
Thanks @bpandola 👍

PR #1739, opened 2018-07-27
bugfix: support name-addr mailbox form

### LINKED ISSUE #1738: [SES] Does not properly verify mailbox with display name
https://tools.ietf.org/html/rfc2822.html#section-3.4 defines two forms of valid mailbox:

* `foo@example.com`
* `"Foo Bar" <foo@example.com>`

SES supports both of these forms. Per https://github.com/spulec/moto/blob/master/moto/ses/models.py#L55, only the first form is supported by moto.

r1b:
Closes #1738. It is not clear in the docs but you can use this form with both `SendEmail` and `SendRawEmail`.

### DIFF
--- a/moto/ses/models.py
+++ b/moto/ses/models.py
@@ -49,7 +49,8 @@ class SESBackend(BaseBackend):
         self.sent_messages = []
         self.sent_message_count = 0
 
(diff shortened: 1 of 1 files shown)


spulec:
Thanks!

Issue #1744, opened 2018-08-02
Dynamodb2 update_item does not raise validation exception for attribute values set to empty strings


cm-iwata:
This is very similar to this issue

[Dynamodb2 put_item does not raise validation exception for attribute values set to empty strings](https://github.com/spulec/moto/issues/1129)

This example run with unittest illustrates the problem

```
import unittest
from unittest import mock

import boto3
import botocore
from moto import mock_dynamodb2

class TestDynamoEmptyString(unittest.TestCase):

  def test_donot_use_moto(self):
    dynamo = boto3.resource('dynamodb')
    table = dynamo.Table('test_table')
    table.put_item(Item={
      'id': 1,
      'col1': 'val1'
    })

    with self.assertRaises(botocore.exceptions.ClientError):
      table.update_item(
        Key={
          'id': 1
        },
        UpdateExpression='SET col1 = :col1',
        ExpressionAttributeValues={
          ':col1': ''
        }
      )
  @mock_dynamodb2
  def test_use_moto(self):
    dynamo = boto3.resource('dynamodb')
    dynamo.create_table(
      TableName='test_table',
      KeySchema=[
          {
              'AttributeName': 'id',
              'KeyType': 'HASH'
          },
      ],
      AttributeDefinitions=[
          {
              'AttributeName': 'id',
              'AttributeType': 'N'
          },
      ],
      ProvisionedThroughput={
          'ReadCapacityUnits': 5,
          'WriteCapacityUnits': 5
      }
    )
    table = dynamo.Table('test_table')
    table.put_item(Item={
      'id': 1,
      'col1': 'val1'
    })

    with self.assertRaises(botocore.exceptions.ClientError):
      table.update_item(
        Key={
          'id': 1
        },
        UpdateExpression='SET col1 = :col1',
        ExpressionAttributeValues={
          ':col1': ''
        }
      )
```

I expect both test will pass
but second test case test_use_moto failed

```
/Users/xxxxx/.anyenv/envs/pyenv/versions/3.6.5/lib/python3.6/unittest/case.py:605: ResourceWarning: unclosed <ssl.SSLSocket fd=7, family=AddressFamily.AF_INET, type=SocketKind.SOCK_STREAM, proto=6, laddr=('xxx.xxx.xxx.xxx', 52086), raddr=('xxx.xxx.xxx.xxx', 443)>
  testMethod()
.F
======================================================================
FAIL: test_use_moto (test.TestDynamoEmptyString)
[...]

PR #1745, opened 2018-08-02
BugFix dynamodb2 update_item with empty string Fixes#1744

cm-iwata:
Dynamodb2 update_item does not raise validation exception for attribute values set to empty strings

This pull requests fixes issue #1744
and raises ClientError, when using the update_item  method with empty string on DynamoDB2


spulec:
Thanks!

PR #1746, opened 2018-08-02
- Updated ResourceAlreadyExistsException(LogsClientError) to latest msg

Brett55:
- Updated ResourceAlreadyExistsException(LogsClientError) to latest Boto3 error msg when using CWL Logs


Brett55:
I don't see how to increase coverage when the previous exception msg was not tested anywhere

spulec:
Thanks!

Issue #1561, opened 2018-04-10
Feature request: AWS Glue Mock


OElesin:
I would like to ask if the maintainers are adding AWS Glue mock. If not all but some of the core APIs.

Thanks in advance

spulec:
As a general rule, new endpoints get added when people that need them have some extra time to add them to Moto.

Right now I don't have a particular use for this so I can't promise anything for timing.

If someone does have time, pull requests are much appreciated (even small one just to get a service started). If you have some example code you want to work, preferably with some example data, that can also speed up the process.

chayac:
Has any work been done on this? I'm interested in contributing, but wanted to make sure someone else isn't already ahead of me.

TheDooner64:
@chayac @OElesin I just opened a PR #1750 with some initial work on the AWS Glue mock. I am focused on the Glue Data Catalog, so not sure if that aligns with your own use cases

spulec:
Going to close this for now given #1750. Feel free to open feature requests for specific endpoints.

PR #1750, opened 2018-08-05
Scaffolding for AWS Glue Data Catalog

TheDooner64:
I started implementing the AWS Glue mock mentioned in issue #1561. My primary use case is the Glue Data Catalog -- so far I implemented basic database and table interactions. This felt like a good point to get some feedback. I'm happy to work on this w/ others that are interested in mocking AWS Glue

cc @spulec


spulec:
This looks great! It looks like the server mode tests are failing because we need to add the glue backend to the [backends](https://github.com/spulec/moto/blob/master/moto/backends.py).

If you can add a line there for that, I think the tests will pass and we will be good to merge.

TheDooner64:
Ah, got it, done. Thanks, @spulec !

spulec:
Thanks for the great work and welcome to Moto!

PR #1751, opened 2018-08-06
Add support for unregistered instances in ELB DescribeInstancesHealth.

### LINKED ISSUE #1748: DescribeInstanceHealth should return State: Unknown for unregistered instances.
The test below fails with `AssertionError: assert 'InService' == 'Unknown'`.
I will provide a PR if you agree that this is something that should be fixed :) 

```python3
@mock_ec2
@mock_elb
def test_failing_test_case():
    elb = boto3.client()
    ec2 = boto3.client()
    # Create an instance (we will not use it)
    instance_id = ec2.run_instances(MinCount=1, MaxCount=1)['Instances'][0]['InstanceId']
    # Create a classic load balancer, without any registered instances
    lb_name = "my_load_balancer"
    elb.create_load_balancer(
        Listeners=[{
            'InstancePort': 80,
            'LoadBalancerPort': 8080,
            'Protocol': 'HTTP'
        }],
        LoadBalancerName=lb_name,
    )
    # The instance was never registered in the load balancer...
    response = elb.describe_instance_health(LoadBalancerName=lb_name,
                                            Instances=[{'InstanceId': instance_id}])
    # and therefore should not be 'InService'.
    assert response['InstanceStates'][0]['State'] == "Unknown"
```

Edit: I made the PR anyway.

Neki:
Fixes #1748.

When an instance is not registered in an ELB, its health status should be "Unknown", not "InService".

### DIFF
--- a/moto/elb/responses.py
+++ b/moto/elb/responses.py
@@ -259,12 +259,22 @@ class ELBResponse(BaseResponse):
 
     def describe_instance_health(self):
         load_balancer_name = self._get_param('LoadBalancerName')
-        instance_ids = [list(param.values())[0] for param in self._get_list_prefix('Instances.member')]
-        if len(instance_ids) == 0:
-            instance_ids = self.elb_backend.get_load_balancer(
-                load_balancer_name).instance_ids
+        provided_instance_ids = [
+            list(param.values())[0]
+            for param in self._get_list_prefix('Instances.member')
+        ]
+        registered_instances_id = self.elb_backend.get_load_balancer(
+            load_balancer_name).instance_ids
+        if len(provided_instance_ids) == 0:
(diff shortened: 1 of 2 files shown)


Neki:
Is there anything I can do to help get this merged? I'm not sure whether the decrease in test coverage is meaningful (I did add tests that cover the new features), but if that's a blocking issue I can invest some time in it :slightly_smiling_face:

spulec:
Looks great; thanks!

Neki:
I've been bitten by this today, so for future reference: when an instance was never registered in the load balancer, the DescribeInstanceHealth API actually fails with an error 400 (status "InvalidInstance"). However, when an instance is registered then deregistered, this PR seems to conform to the AWS API.

Issue #1760, opened 2018-08-08
Running SNS invoke with govcloud results in a KeyError


drmarkpowell:
I have encountered a similar problem to #1613 and #1614 only using the SNS service rather than AWSLambda. When the region is us-gov-west-1 for Govcloud, there is a KeyError at runtime.

I have created a pull request to add the region to the Moto regions list for SQS and SNS following the example from #1613 and #1614.

PR #1761, opened 2018-08-08
added us-gov-west-1 Govcloud regions for SNS and SQS

drmarkpowell:


drmarkpowell:
This PR is to address issue #1760

spulec:
It looks like the linter is unhappy. If you can fix that, will be happy to merge!

bblommers:
This has already been added in #2657, so I'll close this PR.

Issue #1762, opened 2018-08-09
TypeError receiving messages from SQS queue with ReceiveMessageWaitTimeSeconds attribute set


nate:
When an SQS queue is created with the ReceiveMessageWaitTimeSeconds attribute, calling the ReceiveMessage action triggers a TypeError.

Test to reproduce the issue:

```python
sqs = boto3.resource('sqs', region_name='us-east-1')

queue = sqs.create_queue(
    QueueName='test-queue',
    Attributes={
        'ReceiveMessageWaitTimeSeconds': '2',
    }
)

queue.receive_messages()
```

I have already created fix for this and this is the stack trace for the above code when the test is run without the patch in place:

```
$ nosetests -s -m 'test_receive_message_for_queue_with_receive_message_wait_time_seconds_set' tests/test_sqs/test_sqs.py
Coverage.py warning: --include is ignored because --source is set (include-ignored)
E
======================================================================
ERROR: test_sqs.test_receive_message_for_queue_with_receive_message_wait_time_seconds_set
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/nsutton/.asdf/installs/python/3.6.6/lib/python3.6/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/Users/nsutton/Code/moto/moto/core/models.py", line 71, in wrapper
    result = func(*args, **kwargs)
  File "/Users/nsutton/Code/moto/tests/test_sqs/test_sqs.py", line 1213, in test_receive_message_for_queue_with_receive_message_wait_time_seconds_set
    queue.receive_messages()
  File "/Users/nsutton/.asdf/installs/python/3.6.6/lib/python3.6/site-packages/boto3/resources/factory.py", line 520, in do_action
    response = action(self, *args, **kwargs)
  File "/Users/nsutton/.asdf/installs/python/3.6.6/lib/python3.6/site-packages/boto3/resources/action.py", line 83, in __call__
    response = getattr(parent.meta.client, operation_name)(**params)
  File "/Users/nsutton/.asdf/installs/python/3.6.6/lib/python3.6/site-packages/botocore/client.py", line 314, in _api_call
    return self._make_api_call(operation_name, kwargs)
  File "/Users/nsutton/.asdf/installs/python/3.6.6/lib/python3.6/site-packages/botocore/client.py", line 599, in _make_api_call
    operation_model, request_dict)
[...]

PR #1763, opened 2018-08-09
Force receive_message_wait_time_seconds to be int

nate:
Fixes https://github.com/spulec/moto/issues/1762

When a queue is created with the ReceiveMessageWaitTimeSeconds
attribute the value is never converted to an integer. When the
ReceiveMessage action is called it tries to compare the string
ReceiveMessageWaitTimeSeconds with the min and max wait times which
raises a TypeError.

The solution is to convert this value to an integer before comparing.


nate:
How can coverage be decreased when I added a test and the test covers my change??

spulec:
Thanks!

Issue #1764, opened 2018-08-09
Duplicate SQS queue creation errors when attributes are the same


nate:
When performing a CreateQueue for a queue name with identical attributes, the response should be 200 with the correct queue information. Instead, depending on the attributes that are set, it errors.

Code to reproduce:

```python
sqs = boto3.client('sqs', region_name='us-east-1')

dlq_url = sqs.create_queue(QueueName='test-queue-dlq')['QueueUrl']
dlq_arn = sqs.get_queue_attributes(QueueUrl=dlq_url)['Attributes']['QueueArn']

attributes = {
    'DelaySeconds': '900',
    'MaximumMessageSize': '262144',
    'MessageRetentionPeriod': '1209600',
    'ReceiveMessageWaitTimeSeconds': '20',
    'RedrivePolicy': '{"deadLetterTargetArn": "%s", "maxReceiveCount": 100}' % (dlq_arn),
    'VisibilityTimeout': '43200'
}

sqs.create_queue(
    QueueName='test-queue',
    Attributes=attributes
)

sqs.create_queue(
    QueueName='test-queue',
    Attributes=attributes
)
```

This results in the following stack trace:

```
$ nosetests -s -m 'test_create_queue_with_same_attributes' tests/test_sqs/test_sqs.py
Coverage.py warning: --include is ignored because --source is set (include-ignored)
E
======================================================================
ERROR: test_sqs.test_create_queue_with_same_attributes
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/nsutton/.asdf/installs/python/3.6.6/lib/python3.6/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/Users/nsutton/Code/moto/moto/core/models.py", line 71, in wrapper
    result = func(*args, **kwargs)
  File "/Users/nsutton/Code/moto/tests/test_sqs/test_sqs.py", line 66, in test_create_queue_with_same_attributes
    Attributes=attributes
  File "/Users/nsutton/.asdf/installs/python/3.6.6/lib/python3.6/site-packages/botocore/client.py", line 314, in _api_call
    return self._make_api_call(operation_name, kwargs)
  File "/Users/nsutton/.asdf/installs/python/3.6.6/lib/python3.6/site-packages/botocore/client.py", line 612, in _make_api_call
    raise error_class(parsed_response, operation_name)
[...]

PR #1765, opened 2018-08-09
Don't error on double create queue with same attrs

nate:
Fixes https://github.com/spulec/moto/issues/1764

Creating a queue a second time with the same attributes should not raise an error. Since the Queue casts the values and applies defaults it's best to create a new queue and compare all the attributes.


spulec:
Looks great, thanks. Welcome to Moto!

PR #1772, opened 2018-08-10
Issue 1770: Deal with the friendly name properly

NeilRoberts:
- Save friendly name in create_secret.
- Reference the saved friendly name in responses that have "Name" field.
- Verify the received secret_id matches the current value. Don't just
test for an empty string.
- Add test for mismatched secret_id.


spulec:
Thanks for you work and welcome to Moto!

PR #1774, opened 2018-08-11
nextSequenceToken that return by cloudwatch logs putLogEvents should be string, not int.

chrisLeeTW:
according to aws api docs (https://docs.aws.amazon.com/AmazonCloudWatchLogs/latest/APIReference/API_PutLogEvents.html#CWL-PutLogEvents-request-sequenceToken), nextSequenceToken's type should be string, not int.

### DIFF
--- a/moto/logs/models.py
+++ b/moto/logs/models.py
@@ -86,7 +86,7 @@ class LogStream:
         self.events += [LogEvent(self.lastIngestionTime, log_event) for log_event in log_events]
         self.uploadSequenceToken += 1
 
-        return self.uploadSequenceToken
+        return '{:056d}'.format(self.uploadSequenceToken)
 
     def get_log_events(self, log_group_name, log_stream_name, start_time, end_time, limit, next_token, start_from_head):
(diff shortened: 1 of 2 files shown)


moto/logs/models.py
```diff
@@ -86,7 +86,7 @@ def put_log_events(self, log_group_name, log_stream_name, log_events, sequence_t
         self.events += [LogEvent(self.lastIngestionTime, log_event) for log_event in log_events]
         self.uploadSequenceToken += 1
 
-        return self.uploadSequenceToken
+        return "SEQ-{:010d}".format(self.uploadSequenceToken)
```
sfdb: Hey why did you add "SEQ-" here to the string? This is not what I'm getting back from the API, also in the doc page you linked it just shows as:

{
  "nextSequenceToken": "49536701251539826331025683274032969384950891766572122113"
}

chrisLeeTW: thanks for correction, fixed already.

spulec:
Thanks!

PR #1779, opened 2018-08-14
Fix the `StatusCode` returned by lambda invoke

repinel:
According to the AWS documentation:
```
The HTTP status code will be in the 200 range for successful request.
For the RequestResponse invocation type this status code will be 200.
For the Event invocation type this status code will be 202.
For the DryRun invocation type the status code will be 204.
```


repinel:
Close in favor of #2704

PR #1789, opened 2018-08-21
Add `ClusterCreateTime` to Redshift response

### LINKED ISSUE #1778: Redshift does not have "ClusterCreateTime" when "describe_clusters"
```
  File "/path/to/lib/python2.7/site-packages/moto/core/models.py", line 71, in wrapper
  File "test_redshift_cluster.py", line 29, in test_check_cluster_status
    self.assertEqual(self.rs._check_cluster_status(self.source_cluster), 'available')
  File "/path/to/redshift_cluster.py", line 83, in _check_cluster_status
    status = sorted(clusters, key=lambda k: k['ClusterCreateTime'], reverse=True)[0]['ClusterStatus']
  File "/path/to/redshift_cluster.py", line 83, in <lambda>
    status = sorted(clusters, key=lambda k: k['ClusterCreateTime'], reverse=True)[0]['ClusterStatus']
KeyError: 'ClusterCreateTime'
```
From the docs of [boto3](https://boto3.readthedocs.io/en/latest/reference/services/redshift.html#Redshift.Client.describe_clusters), it's shown that `ClusterCreateTime` should be present, but it's not the case when mocking the instance/cluster using moto.

How to reproduce:

function code:
```
import boto3
def quick_describe(cluster_name):
    client = boto3.client('redshift')
    response = client.describe_clusters(ClusterIdentifier=cluster_name)
    return response
```

test code:
```
from moto import mock_redshift

from some_path import quick_describe

@mock_redshift
class TestRedshift(unittest.TestCase):
    def test_cluster_status(self):
        client = boto3.client('redshift')
        response = client.create_cluster(
            ClusterIdentifier='unittest',
            NodeType='ds2.xlarge',
            MasterUsername='TestUser',
            MasterUserPassword='TestPassword')
       response = quick_describe('unittest')
       print(response)
```
And there's no `ClusterCreateTime`

bpandola:
Fixes #1778

### DIFF
--- a/moto/redshift/models.py
+++ b/moto/redshift/models.py
@@ -78,6 +78,7 @@ class Cluster(TaggableResourceMixin, BaseModel):
         super(Cluster, self).__init__(region_name, tags)
         self.redshift_backend = redshift_backend
         self.cluster_identifier = cluster_identifier
+        self.create_time = iso_8601_datetime_with_milliseconds(datetime.datetime.now())
         self.status = 'available'
         self.node_type = node_type
         self.master_username = master_username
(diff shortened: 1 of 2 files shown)


spulec:
Thanks!

PR #1792, opened 2018-08-24
More accurate Filter and Transition rules for S3 Lifecycle Policy

entropicdrifter:
Changes S3 Lifecycle Rule construction to be more accurate to the Boto3 documentation.
Specifically:
"A Filter must have exactly one of Prefix, Tag, or And specified."
and
"Transitions (list) --"
https://boto3.readthedocs.io/en/latest/reference/services/s3.html#S3.Client.put_bucket_lifecycle_configuration

In addition, adds validation for some multiple-transition scenario constraints, as per the official S3 docs:
https://docs.aws.amazon.com/AmazonS3/latest/dev/lifecycle-transition-general-considerations.html#lifecycle-general-considerations-transition-sc

### DIFF
--- a/moto/s3/models.py
+++ b/moto/s3/models.py
@@ -331,20 +331,25 @@ class LifecycleAndFilter(BaseModel):
         self.prefix = prefix or ''
         self.tags = tags
 
+class LifecycleTransition(BaseModel):
+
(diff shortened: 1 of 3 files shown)


spulec:
Code seems reasonable; if we can get Travis to pass, I'll be happy to look into merging.

Issue #1793, opened 2018-08-25
Test cases written in moto using Latest version of boto3 fails


krishnamee2004:
Test cases written in moto makes actual AWS API calls to botocore instead of mocking them. This happens with the latest version of boto3 (1.8.*). It used to work fine without issues with 1.7.* versions.

**Sample code to reproduce error**

```
import boto3
import json
from moto import mock_s3

@mock_s3
def test_mock_s3():
    client = boto3.client('s3', region_name='us-east-1')
    client.create_bucket(Bucket='testbucket')
    response = client.list_buckets()
    print json.dumps(response, default=str)

if __name__ == "__main__":
    test_mock_s3()
```

**Expected result**
Method should return the `ListBuckets` response. It should look something like:
```
{"Owner": {"DisplayName": "webfile", "ID": "bcaf1ffd86f41161ca5fb16fd081034f"}, "Buckets": [{"CreationDate": "2006-02-03 16:45:09+00:00", "Name": "testbucket"}], "ResponseMetadata": {"RetryAttempts": 0, "HTTPStatusCode": 200, "HTTPHeaders": {"Content-Type": "text/plain"}}}
```

**Actual error**
botocore.errorfactory.BucketAlreadyExists: An error occurred (BucketAlreadyExists) when calling the CreateBucket operation: The requested bucket name is not available. The bucket namespace is shared by all users of the system. Please select a different name and try again.

** Full stack trace**
```
Traceback (most recent call last):
  File "testcases.py", line 14, in <module>
    test_mock_s3()
  File "/private/tmp/virtualenv2/lib/python2.7/site-packages/moto/core/models.py", line 71, in wrapper
    result = func(*args, **kwargs)
  File "testcases.py", line 8, in test_mock_s3
    client.create_bucket(Bucket='testbucket')
  File "/private/tmp/virtualenv2/lib/python2.7/site-packages/botocore/client.py", line 314, in _api_call
    return self._make_api_call(operation_name, kwargs)
  File "/private/tmp/virtualenv2/lib/python2.7/site-packages/botocore/client.py", line 612, in _make_api_call
    raise error_class(parsed_response, operation_name)
botocore.errorfactory.BucketAlreadyExists: An error occurred (BucketAlreadyExists) when calling the CreateBucket operation: The requested bucket name is not available. The bucket namespace is shared by all users of the system. Please select a different name and try again.
```

**Library versions**
```
moto : 1.3.4
boto3 : 1.8.1 - fails
[...]

PR #1794, opened 2018-08-27
RFC: Change setup.py to require older version of botocore with compatible implementation

garyd203:
Mitigate #1793 by restricting maximum version for botocore.

botocore v1.11.0 [changed it's internal implementation](https://github.com/boto/botocore/pull/1495) so that it now uses a different library for HTTP requests. This means that moto's mocking will not work, and test code will inadvertently call the live AWS service.

As an interim solution to reduce the impact of this breakage, we restrict the "required" (ie. recommended) version of botocore so that users will be less likely to use an incompatible version, and will receive a pip warning when they do.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -9,7 +9,7 @@ install_requires = [
     "Jinja2>=2.7.3",
     "boto>=2.36.0",
     "boto3>=1.6.16",
-    "botocore>=1.9.16",
+    "botocore>=1.9.16,<1.11",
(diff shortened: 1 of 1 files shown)


garyd203:
Note to other readers - also read the discussion in #1793

spulec:
Thanks @garyd203 for the help with all of this. Going to cut a release now.

PR #1804, opened 2018-09-01
Fix small issue in README example

gardenunez:
While I was checking moto library I found a small typo in the first example. This PR fix it.


spulec:
I don't think this is a typo. We are making the string a byte-string. Let me know if I'm reading it wrong though.

terrycain:
The `.decode("utf-8")` would return a normal string unless im reading it wrong. Think we're good on this one

PR #1808, opened 2018-09-05
Upgrade xray sdk version

luyun-aa:


yan12125:
Now here's a good news - the latest aws-xray-sdk is allowed since https://github.com/spulec/moto/commit/ed861ecae1039a048a6350a4ff832ef094cdf2c2. I've tried to run moto with aws-xray-sdk 2.2.0 and it works fine.

spulec:
This should be fixed with https://github.com/spulec/moto/pull/1948

PR #1810, opened 2018-09-05
Fix #1809: skip patching staticmethods

### LINKED ISSUE #1809: Decorating a class breaks staticmethods
If a test class is decorated `@moto.mock_s3`:

> moto/core/models.py:71: in wrapper
```python
```

causes the instance to be passed to `func` even if it's `@staticmethod` decorated.

I have not tested with the other mocks, but that seems to be a generic wrapper, so I assume all are affected.

OJFord:


OJFord:
Not sure why the build before the fix looks like it passed on Travis: it fails (as expected) for me locally.

spulec:
Thanks!

Issue #1813, opened 2018-09-06
botocore version conflict between awscli, boto3, and moto


matthew-craig-ncino:
Performing a side by side pip install of awscli, boto3, botocore, and moto results in an incompatible conflict.

To Reproduce run this in a terminal with a fresh machine:
`pip install boto3 awscli moto botocore`

Actual results yeild: 
`moto 1.3.5 has requirement botocore<1.11,>=1.9.16, but you'll have botocore 1.11.8 which is incompatible.`
Performing a `pip list` confirms that the moto library is not installed after the above command.

Expected results:
Successful install.

I am attempting to use the following versions:
- awscli-1.16.8 
- boto3-1.8.8 
- botocore-1.11.8 
- moto-1.3.5

I have tried this on the following versions of python:
- Python 3.7.0
- Python 2.7.15
- pip 18.0 (for both python2 and python3)

The source of the problem is:

awscli requires: botocore>1.11.8
- https://github.com/aws/aws-cli/blob/4515c962ab8d186e9e76262b2367c845b14d4a64/setup.py#L26

boto3 requires botocore>1.11.8
- https://github.com/boto/boto3/blob/b2e203799aa2dd18bca33ba095934bc75bf12ff0/setup.py#L17

moto requires: botocore>=1.9.16,<1.11
- https://github.com/spulec/moto/blob/e8c65d3d85ba62e857c7448691568ad4b02f4a46/setup.py#L12

This creates a grid lock that prevents moto from being installed.

I am running on Mac OSX High Sierra. Working with the latest versions of

matthew-craig-ncino:
I attempted to bump the moto requirement to anything less than 1.12 but this caused numerous NoCredential Exceptions in unit testing.  After an hour of looking I was not able pinpoint the cause.  At about botocore version 1.11.4(~) they did make some changes to auth and signatures, but I was not familiar enough with moto/botocore to resolve. I may look at when I have more time, but if anyone knows the projects better it would be nice to have additional input.  

I also had a question for devs that I didn't see in the documentation, How do the unit tests run locally? I was able to follow the travis setup to the point of running `make test` but got errors about `nosetests not defined`.  I use nose in my own testing so but I'm not sure what I'm missing on this.

Thanks.

nikolay-saskovets:
We faced with slighty the same issue.
For us the rollback to `moto==1.3.4` is a workable workaround.

[...]

PR #1814, opened 2018-09-06
Issue #1813 Updating botocore dependency version

matthew-craig-ncino:
- Updated version of botocore to allow awscli and boto3 to work with
moto


matthew-craig-ncino:
Looks like they may have changed some auth methods in 1.11.4 ? Looking.

PR #1821, opened 2018-09-11
Fix #1820: Add support for multiple delete markers on an s3 object

### LINKED ISSUE #1820: Multiple delete markers on a single s3 object breaks when listing bucket object versions
In S3 it is possible to delete an object multiple times, this adds multiple delete markers to the object versions. Moto does not support listing all object versions in a bucket if there is an object that has multiple delete markers. 

It breaks at [moto/s3/responses.py L1276](https://github.com/spulec/moto/blob/master/moto/s3/responses.py#L1276) when it tries to access `marker.key.name` since the `key` attribute of the second delete marker is a `FakeDeleteMarker` which does not have the `name` attribute. The same goes for [L1279](https://github.com/spulec/moto/blob/master/moto/s3/responses.py#L1279) with the `last_modified_ISO8601` attribute.

Some code to reproduce:
```
    bucket_name = 'mybucket'
    key = u'key-with-versions-and-unicode-ó'
    s3.create_bucket(Bucket=bucket_name)
    s3.put_bucket_versioning(
        Bucket=bucket_name,
        VersioningConfiguration={
            'Status': 'Enabled'
        }
    )
    items = (six.b('v1'), six.b('v2'))
    for body in items:
        s3.put_object(
            Bucket=bucket_name,
            Key=key,
            Body=body
        )
     # Delete the object twice to add multiple delete markers
    s3.delete_object(Bucket=bucket_name, Key=key)
    s3.delete_object(Bucket=bucket_name, Key=key)
    # List object versions
    response = s3.list_object_versions(Bucket=bucket_name)
```

Jtupiter:
This commit adds a `name` and `last_modified` so it does not break when listing object versions if an object has more than one delete marker. 

Setting `last_modified` on the `FakeDeleteMarker` will allow list object versions to display the actual last_modified of the `DeleteMarker` rather than just displaying the DeleteMarker's key's `last_modified`.

### DIFF
--- a/moto/s3/models.py
+++ b/moto/s3/models.py
@@ -27,7 +27,13 @@ class FakeDeleteMarker(BaseModel):
 
     def __init__(self, key):
         self.key = key
+        self.name = key.name
+        self.last_modified = datetime.datetime.utcnow()
         self._version_id = key.version_id + 1
+
+    @property
+    def last_modified_ISO8601(self):
+        return iso_8601_datetime_with_milliseconds(self.last_modified)
 
     @property
     def version_id(self):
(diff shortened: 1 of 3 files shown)


spulec:
This looks great; thanks!

PR #1825, opened 2018-09-14
bumping botocore and boto3

JackDanger:


softinio:
@spulec Any ETA on this PR being merged soon? Thanks!

yan12125:
New botocore and boto3 break moto (https://github.com/spulec/moto/issues/1793), so it's unlikely that this patch is merged as-is.

PR #1846, opened 2018-10-01
Filter event log ids should be strings

william-richard:
Based on the boto docs, eventId should be returned as a string.
https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/logs.html#CloudWatchLogs.Client.filter_log_events


spulec:
Thanks!

Issue #1598, opened 2018-04-25
Moto 1.3 causes tests to fail when `assert_all_requests_are_fired = True`


jakul:
Moto v1.3 causes my tests to fail if I set `responses.mock.assert_all_requests_are_fired = True`.

I set `responses.mock.assert_all_requests_are_fired = True` to make sure that my application is behaving as expected, but since it doesn't make all the requests to all the `moto` urls then my test fails.

## How to recreate
```
craig@amogato2:/tmp
$ virtualenv /tmp/moto-error
New python executable in /tmp/moto-error/bin/python
Installing setuptools, pip, wheel...done.

craig@amogato2:/tmp
$ source /tmp/moto-error/bin/activate

[moto-error] craig@amogato2:/tmp
$ pip install pytest-responses moto

... output removed ...

[moto-error] craig@amogato2:/tmp
$ echo -e "import responses\nfrom moto import *\nresponses.mock.assert_all_requests_are_fired = True\n@mock_dynamodb2\ndef test_pass():\n    pass" > test.py

[moto-error] craig@amogato2:/tmp
$ pytest test.py
======================================== test session starts ========================================
platform darwin -- Python 2.7.10, pytest-3.5.1, py-1.5.3, pluggy-0.6.0
rootdir: /private/tmp, inifile:
plugins: responses-0.3.0
collected 1 item

test.py F

============================================= FAILURES ==============================================
_____________________________________________ test_pass _____________________________________________

args = (), kwargs = {}, result = None

    def wrapper(*args, **kwargs):
        self.start(reset=reset)
        try:
            result = func(*args, **kwargs)
        finally:
>           self.stop()

/tmp/moto-error/lib/python2.7/site-packages/moto/core/models.py:73:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
/tmp/moto-error/lib/python2.7/site-packages/moto/core/models.py:65: in stop
    self.disable_patching()
/tmp/moto-error/lib/python2.7/site-packages/moto/core/models.py:220: in disable_patching
    responses_mock.stop()
[...]

PR #1847, opened 2018-10-02
Support the latest version of botocore

joguSD:
As discussed in #1793 in a recent version of `botocore`, the vendored version of `requests` was dropped in favor of a direct dependency on `urllib3`. This broke the `MockAWS` implementation based on `responses`. As of `botocore` version 1.12.13 the [before-send](https://botocore.amazonaws.com/v1/documentation/api/latest/topics/events.html#before-send) event has been added which allows for the HTTP layer in `botocore` to be bypassed by registering a handler to this event that will introspect the request and return the appropriate mocked response.

The event handler implementation is largely the same as the logic used in the `responses` module but boiled down to the functionality required by `moto`. This event handler is registered globally at the import time of `moto` and all `botocore` clients created while `moto` is imported will be able to be mocked.

Unfortunately, the `before-send` event *cannot* support all of the functionality that mocking out `requests` could. Specifically, it cannot support mocking the instance metadata credential resolver. This means that credentials need to be provided to `botocore` in some manner now for the mocks to function, which is why I've set the credential environment variables to be present for all tests. Currently the tests that fail are either expecting instance metadata to be mocked out when accessed via `requests`, or are otherwise using `requests`. However, this PR should support the majority of `moto` use cases.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,6 +23,8 @@ matrix:
       sudo: true
 before_install:
   - export BOTO_CONFIG=/dev/null
+  - export AWS_SECRET_ACCESS_KEY=foobar_secret
+  - export AWS_ACCESS_KEY_ID=foobar_key
 install:
   # We build moto first so the docker container doesn't try to compile it as well, also note we don't use
   # -d for docker run so the logs show up in travis
@@ -32,8 +34,6 @@ install:
 
     if [ "$TEST_SERVER_MODE" = "true" ]; then
       docker run --rm -t --name motoserver -e TEST_SERVER_MODE=true -e AWS_SECRET_ACCESS_KEY=server_secret -e AWS_ACCESS_KEY_ID=server_key -v `pwd`:/moto -p 5000:5000 -v /var/run/docker.sock:/var/run/docker.sock python:${TRAVIS_PYTHON_VERSION}-stretch /moto/travis_moto_server.sh &
-      export AWS_SECRET_ACCESS_KEY=foobar_secret
-      export AWS_ACCESS_KEY_ID=foobar_key
     fi
     travis_retry pip install boto==2.45.0
     travis_retry pip install boto3
--- a/moto/core/models.py
+++ b/moto/core/models.py
@@ -295,8 +295,10 @@ class BotocoreStubber(object):
 
         return response
 
+
 botocore_stubber = BotocoreStubber()
 BUILTIN_HANDLERS.append(('before-send', botocore_stubber))
+
 
 class BotocoreEventMockAWS(BaseMockAWS):
     def reset(self):


spulec:
This is looking great; thanks for all of your work on this.

If we can add https://github.com/joguSD/moto/pull/1 on top of this, I think we are in a good place to merge and hopefully push past all of this.

If I get some time in the next couple of days, I'll try to get the final parts of this done. If anyone else gets some time before that, feel free to jump in.

joguSD:
@spulec Do you think continuing to patch `requests` is the correct approach to get all the tests working again? `botocore` no longer uses `requests` anywhere so it's really just changing how the current stubber implementation works just to get the tests that use `requests` directly to pass. In practice, `requests` will never be used by `botocore` so I guess it comes down to a question of whether or not you view:
```python
@mock_s3
def main():
    requests.get('s3.url') # should be mocked
```
as a feature of `moto` and not just a byproduct of the implementation.

spulec:
I think that is a nice feature of Moto (at least that is how we have made it work previously) so I am in favor of that.

jakul:
I am -1 on still using `responses` in moto, because it interferes with my use of `responses` as an application developer. I raised an issue about this previously https://github.com/spulec/moto/issues/1598 for which I don't currently have a reasonable workaround (other than still using `moto` 1.2). Removing `responses` completely from Moto will definitely make my life simpler!

lhufnagel:
Yeah, actually I agree to that

spulec:
I suppose we could have a flag to allow the user to control if we setup `responses` mocking.

e271828-:
Look forward to seeing this work land in a release. May be worth updating the README at the top to warn people in the interim. Also -1 on keeping `responses` around.

joguSD:
@spulec Any updates on this? I think in the short-term just getting moto working with newer versions of botocore is important, and the decision to add configuration for `responses` or removing it can be a separate issue.

spulec:
Agreed that we can discuss if `responses` should be removed later, since we can't remove it now and break existing behavior.

I don't have a lot of time right now, but if someone has time to create a PR that works with all of our existing tests, I'll be happy to merge and cut a new release.

lhufnagel:
@spulec can you have a look at this https://github.com/spulec/moto/pull/1907 ? Thx!

spulec:
@lhufnagel the code looks good!

Can someone else here give the branch in #1907 a test? Due to the nature of this change, would be great to have another person or two test it.

jvtm:
I was able to run our internal project's test cases with:

* Boto versions from `pip freeze`:

        boto==2.49.0
        boto3==1.9.34
        botocore==1.12.34

* #1907 checkout in `PYTHONPATH`
* `BOTO_CONFIG=/dev/null AWS_SECRET_ACCESS_KEY=bogus AWS_ACCESS_KEY_ID=bogus PYTHONPATH=$HOME/checkout/moto pytest ...`
* tests use `mock_s3()` and `mock_sqs()` in few different ways (fixtures for pytest, setup/teardown for traditional unittest)
* verified manually that I don't have working creds in this environment, and even run the relevant parts of the test set without network :)

:+1:

rouge8:
#1907 passes on one of my Python projects, using cloudformation, ec2, elb, iam, s3, rds, route53, and sqs mocks.

PR #1848, opened 2018-10-02
Mock more of the Glue Data Catalog APIs

ashb:
This expands on #1750  and adds some more APIs (much easier to have a base to work on! As pointed out in #1561)

This adds some of the missing Get/Update/Create APIs relating to the Glue data catalog -- but not yet all of them, and improves some of the error handling/edge cases when tables or DBs don't exist etc.

### DIFF
--- a/moto/glue/exceptions.py
+++ b/moto/glue/exceptions.py
@@ -6,19 +6,56 @@ class GlueClientError(JsonRESTError):
     code = 400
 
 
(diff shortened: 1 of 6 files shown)


ashb:
100% coverage on the moto/glue/* files now. I'm not sure why coverals is reporting a drop in coverage -  unless it's a bit noisy some times? 

PTAL when you have a spare moment @spulec

spulec:
This is great; thanks for your help!

PR #1854, opened 2018-10-06
base64.encodestring() is deprecated

asherf:
https://docs.python.org/3/library/base64.html#base64.encodestring

>base64.encodestring(s)
>Deprecated alias of encodebytes().
>
>Deprecated since version 3.1.


spulec:
We still support Python 2.7 so I don't think this is going to work.

I'm going to close, but let me know if I'm misunderstanding.

PR #1859, opened 2018-10-08
Fixes #1857 Ensures a UserPool Id starts like {region}_

### LINKED ISSUE #1857: cognito-idp UserPool id format does not match AWS format
The format for a Cognito UserPool Id produced by the the cognito-idp mock does not produce ids in the expected format for a Cognito UserPool - The ids produced are uuids. The format for the Id field of a UserPool is documented as 

>  Id
> 
>     The ID of the user pool.
> 
>     Type: String
> 
>     Length Constraints: Minimum length of 1. Maximum length of 55.
> 
>     Pattern: [\w-]+_[0-9a-zA-Z]+
> 
>     Required: No

https://docs.aws.amazon.com/cognito-user-identity-pools/latest/APIReference/API_UserPoolType.html

So a uuid isn't a valid representation of an Id

This can be reproduced by 

```
import moto
import boto3

create_pool_kwargs = {
            "PoolName": "test_pool",
            "Schema": [
                {
                    "Name": "email",
                    "AttributeDataType": "String",
                    "Required": True,
                    "Mutable": True,
                },
                {
                    "Name": "tenant_id",
                    "AttributeDataType": "String",
                    "Mutable": False,
                },
            ],
            "AdminCreateUserConfig": {
                "AllowAdminCreateUserOnly": True,
                "UnusedAccountValidityDays": 1,
            },
        }


def set_up_tear_down_user_pool():
    cognito_idp = boto3.client('cognito-idp')
    pool = cognito_idp.create_user_pool(**create_pool_kwargs)
    pool_id = pool['UserPool']['Id']
    print(pool_id)
    cognito_idp.delete_user_pool(UserPoolId=pool_id)


# with moto
with moto.mock_cognitoidp() as mock_cognito:
    set_up_tear_down_user_pool()

# without
set_up_tear_down_user_pool()
```

Produces:

```
eb9ef17e-acea-4a95-8440-7ee79dd1f172
eu-west-1_qtdBQSSL4
```

The general expectation is that the pool_id is in the format "{region}_{id}". I usually use the region part when attempting to determine, from a pool id, the region that pool is available in. 

I'm using the package installed via pip and python mocks.

```
moto==1.3.4
botocore==1.10.52
boto3==1.7.3
```

georgealton:
To resolve #1857 this commit changes the format of the id string generated for a new user pool. It prefixes the id with the region, followed by an _ then the uuid as a hex. This matches the AWS id format for user pools, validating against the pattern declared in the API documentation (https://docs.aws.amazon.com/cognito-user-identity-pools/latest/APIReference/API_UserPoolType.html)

### DIFF
--- a/moto/cognitoidp/models.py
+++ b/moto/cognitoidp/models.py
@@ -24,7 +24,7 @@ class CognitoIdpUserPool(BaseModel):
 
     def __init__(self, region, name, extended_config):
         self.region = region
-        self.id = str(uuid.uuid4())
+        self.id = "{}_{}".format(self.region, str(uuid.uuid4().hex))
         self.name = name
         self.status = None
         self.extended_config = extended_config or {}
--- a/tests/test_cognitoidp/test_cognitoidp.py
+++ b/tests/test_cognitoidp/test_cognitoidp.py
@@ -24,6 +24,7 @@ def test_create_user_pool():
     )
 
     result["UserPool"]["Id"].should_not.be.none
+    result["UserPool"]["Id"].should.match(r'[\w-]+_[0-9a-zA-Z]+')
     result["UserPool"]["Name"].should.equal(name)
     result["UserPool"]["LambdaConfig"]["PreSignUp"].should.equal(value)
 


spulec:
Thanks and welcome to Moto!

PR #1869, opened 2018-10-11
Allow spaces to if_not_exists, attribute_exists and attribute_not_exists

Surgo:
# TL;DR

- PynamoDB generate `if_not_exists`, `attribute_exists` and `attribute_not_exists`  with [extra spaces](https://github.com/pynamodb/PynamoDB/blob/3.3.1/pynamodb/expressions/operand.py#L173)
- DynamoDB (and [DynamoDB local](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/DynamoDBLocal.html)) works fine but moto does not work `if_not_exists` with extra spaces.


spulec:
Thanks!

PR #1872, opened 2018-10-12
Fix wrong type if exists

Surgo:
- `cast_value` s are always `"S"` (string type) value
- Because `cast_value` does not in `expression_attribute_values`


spulec:
Thanks!

PR #1879, opened 2018-10-17
Fixes for IAM Groups

mikegrima:
Fixed a few bugs for IAM Groups.


mikegrima:
Confirmed that this is working with my unit tests.

PR #1884, opened 2018-10-17
remove the marker since this is truncated

willbengtson:
remove the marker since this is truncated.

The original PR I left the marker in which can break some code depending on how pagination is implemented.


spulec:
Thanks!

PR #1733, opened 2018-07-24
Python 3.7 support

yan12125:
Fixes #1706

`.travis.yml` is so complicated as Python 3.7 is not fully supported on Travis CI. See https://github.com/travis-ci/travis-ci/issues/9815

PR #1886, opened 2018-10-19
Add Python 3.7 support, remove Python 3.3 support

jonafato:
As of #1733, Python 3.7 is supported, so reflect that in the Trove
classifiers.

As of 2017-09-29, Python 3.3 is end-of-life and no longer receives
updates of any kind (including security fixes), so remove it from the
list of supported versions.


jonafato:
@spulec I mistakenly removed something required for 2.7 support. I've added that back in and rebased against master, so I expect everything to pass now.

mikegrima:
@spulec Any issues merging this? We're using 3.7 internally with without moto issues.

dargueta:
While you're at it here, you could add `python_requires` to the `setuptools.setup()` call to make this package unavailable to unsupported Python versions. It won't even show up in `pip search`. I don't think the classifiers are obeyed by `pip`.

Unfortunately because you have to skip over 3.0-3.3, it's a little ugly:

```py
setup(
    ...
    python_requires='>=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*'
)
```

terrycain:
I second adding python_requires,

spulec:
Thanks!

PR #1888, opened 2018-10-22
Fix boto core requests

electroniceagle:
Latest version of boto3 (1.8) causes moto test cases to make actual AWS
calls to botocore instead of mocking them. This is fixed in this PR.
See: https://github.com/spulec/moto/issues/1793


electroniceagle:
Sorry, I totally did a hub pull-request in the wrong directory.

PR #1889, opened 2018-10-22
Without double quotes, pip would not install

adamdavis40208:
On my system without quotes:

```
pip install moto[server]
zsh: no matches found: moto[server] 
```

with quotes:

```
pip install "moto[server]"
Requirement already satisfied: moto[server] [..] 
```


spulec:
Thanks!

PR #1897, opened 2018-10-24
Adds limiting/pagination to cognitoidp list_* functions

### LINKED ISSUE #1896: cognito idp list endpoints don't currently support pagination
When calling cognitoidp functions such as list_user_pools and list_users, should return a limited set of results. The result set that is returned to the caller when using cognito does implement the limiting and pagination mechanics 

Therefore result produced by cognito will be much larger than the expected response from cognito when calling these endpoints, and might lead to implementation issues arising from a user expecting more results from cognito. 

eg
```
import uuid

import moto
import boto3

create_pool_kwargs = {
            "PoolName": "test_pool",
            "Schema": [
                {
                    "Name": "email",
                    "AttributeDataType": "String",
                    "Required": True,
                    "Mutable": True,
                },
                {
                    "Name": "tenant_id",
                    "AttributeDataType": "String",
                    "Mutable": False,
                },
            ],
            "AdminCreateUserConfig": {
                "AllowAdminCreateUserOnly": True,
                "UnusedAccountValidityDays": 1,
            },
        }


def set_up_tear_down():
    cognito_idp = boto3.client('cognito-idp')
    pool = cognito_idp.create_user_pool(**create_pool_kwargs)
    pool_id = pool['UserPool']['Id']
    for i in range(10):
        cognito_idp.admin_create_user(UserPoolId=pool_id,
                                      Username=str(uuid.uuid4()))
    max_results = 1
    resp = cognito_idp.list_users(UserPoolId=pool_id, Limit=max_results)
    print(f"Number of users {len(resp['Users'])} should be {max_results}")
    print(resp.get("PaginationToken", "NotImplemented"))
    cognito_idp.delete_user_pool(UserPoolId=pool_id)


# with moto
with moto.mock_cognitoidp() as mock_cognito:
    print("With moto")
    set_up_tear_down()

# without
print("With AWS:")
set_up_tear_down()
```

produces
```
With moto
Number of users 10 should be 1
NotImplemented
With AWS:
Number of users 1 should be 1
CAISlAIIARLtAQgsdE8iLCJuZXh0S2V5IjoiQUFBQUFBQUFCRmxBQVFFQkpnRzFxUU9xUkN1R2wyOGRyK2pncUhyRGpvbWVSVVl5bnRGZVdxdkJCdlZsYm1ZN01tWXlPVFE0T0RNdFl6VmtNeTAwT1RjM0xUZzBNR1l0Wm1Jd05ghjghjghjwcmV2aW91c1JlcXVlc3RUaW1lIjoxNTQwMzkxNDA0MTIyfRogtES8CLAwq7CVpp/smYQpJbb2xgU8K7m1JMMCkLfabyo=
```

georgealton:
resolves #1896 

PR adds in a decorator for wrapping current methods that return non paginated responses in cognito (might be useful elsewhere?) 

It converts the returned iterable from a backend call into a list and then slices from the pagination token, through the limit/maxitems -- default page sizes can be configured in the decorator creation, to handle many calls not requiring the maxitems / limit to be explicitly set by the caller.

### DIFF
--- a/moto/cognitoidp/models.py
+++ b/moto/cognitoidp/models.py
@@ -2,6 +2,7 @@ from __future__ import unicode_literals
 
 import datetime
 import functools
+import itertools
 import json
 import os
 import time
@@ -43,16 +44,12 @@ def paginate(limit, start_arg="next_token", limit_arg="max_results"):
     def outer_wrapper(func):
         @functools.wraps(func)
         def wrapper(*args, **kwargs):
-            # setup
             start = int(default_start if kwargs.get(start_arg) is None else kwargs[start_arg])
-            stop = int(limit if kwargs.get(limit_arg) is None else kwargs[limit_arg])
-            end = start + stop
-            # call
+            lim = int(limit if kwargs.get(limit_arg) is None else kwargs[limit_arg])
+            stop = start + lim
             result = func(*args, **kwargs)
-            # modify
-            results = list(result)
-            limited_results = results[start: end]
-            next_token = end if end < len(results) else None
+            limited_results = list(itertools.islice(result, start, stop))
+            next_token = stop if stop < len(result) else None
             return limited_results, next_token
         return wrapper
     return outer_wrapper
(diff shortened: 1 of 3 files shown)


spulec:
Thanks!

Issue #1893, opened 2018-10-23
SecretsManager can only store one secret


sirech:
The implementation of SecretsManager can only store one secret. If you create a second secret, the first one will be replaced. This means that any test or app that you want to use can only store one secret.

**Note**: I haven't done Python in a while but I could try to create a PR for this. My idea would be to replace the implementation that stores everything in instance variables with a hash that stores secrets based on the `secret_id`

sepulworld:
Hi @sirech! How are you writing the test? Moto should keep track of multiple secrets created in same test function (covered by the decorator `@mock_secretsmanager`)

Like so...

https://github.com/spulec/moto/blob/314d10431143a77307274da62700e42dfc432e80/tests/test_secretsmanager/test_secretsmanager.py#L37-L53

sepulworld:
Oh, I think I understand what you are looking for. You are looking to generate multiple secrets and use the (list_secrets)[https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/secretsmanager.html#SecretsManager.Client.list_secrets] API to retreive (not covered in moto yet). .Yeah, I think your idea of using a dictionary for instance vars makes sense.

sepulworld:
I like your idea though, switch out variable storage to dict like what is happening in Dynamodb2

https://github.com/spulec/moto/blob/master/moto/dynamodb2/models.py#L643
and it gets added to here: https://github.com/spulec/moto/blob/master/moto/dynamodb2/models.py#L655. Do something similar with implementing list_secrets in secretsmanager as you mentioned, good call 👍

sepulworld:
I can help add this week/weekend @sirech.

sirech:
actually @sepulworld is missing the condition that would make the test fail. If you create `test-secret` and then `test-secret-2` (in that order), and then use `get_secret` to retrieve `test-secret`, it will fail, as the only secret that is stored at that point is the second one. That is what I want to address. My use case is testing an initialization script that requires access to 2 secrets to boot.

sepulworld:
Makes sense. Yes, get_secret and list_secrets need to be implemented. I have a branch I'm working on. Stand by for review

sirech:
I did a rough implementation. Sadly I'm having a bit of trouble running the tests, so I cannot do any proper TDD. In any case, my approach is pretty simple. Here is the diff:

```diff
diff --git a/moto/secretsmanager/models.py b/moto/secretsmanager/models.py
[...]

PR #1898, opened 2018-10-24
Refactor to store multiple scrects, use uuid, tags support

sepulworld:
@sirech thanks for the ideas and help here! See: https://github.com/spulec/moto/issues/1893 

* Refactor for multiple secrets @sirech
* Use UUID replace hardcoded value
* Use random string in ARN id value
* Fix datetime stamp
* Adding support for Tags https://github.com/spulec/moto/issues/2004
* Add tests

### DIFF
--- a/moto/secretsmanager/models.py
+++ b/moto/secretsmanager/models.py
@@ -2,6 +2,7 @@ from __future__ import unicode_literals
 
 import time
 import json
+import uuid
 
 import boto3
 
@@ -18,10 +19,6 @@ class SecretsManager(BaseModel):
 
     def __init__(self, region_name, **kwargs):
         self.region = region_name
-        self.secret_id = kwargs.get('secret_id', '')
-        self.version_id = kwargs.get('version_id', '')
-        self.version_stage = kwargs.get('version_stage', '')
-        self.secret_string = ''
 
 
 class SecretsManagerBackend(BaseBackend):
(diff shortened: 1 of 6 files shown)


moto/secretsmanager/models.py
```diff
@@ -29,51 +30,57 @@ class SecretsManagerBackend(BaseBackend):
...
+        self.secrets = {}
 
     def reset(self):
         region_name = self.region
         self.__dict__ = {}
         self.__init__(region_name)
 
     def _is_valid_identifier(self, identifier):
-        return identifier in (self.name, self.secret_id)
+        if identifier in self.secrets:
```
sirech: would `return identifier in self.secrets` work?

moto/secretsmanager/models.py
```diff
@@ -29,51 +30,57 @@ class SecretsManagerBackend(BaseBackend):
...
         })
 
         return response
 
     def create_secret(self, name, secret_string, **kwargs):
 
-        self.secret_string = secret_string
-        self.secret_id = name
-        self.name = name
+        self.generated_version_id = str(uuid.uuid4())
```
sirech: why is this stored as an instance variable? Shouldn't it be just a regular variable?

sirech:
looks good to me!

moto/secretsmanager/models.py
sepulworld: Good catch! Switching to a regular variable

moto/secretsmanager/models.py
sepulworld: Yes, much cleaner! ty

sepulworld:
@spulec this is good to go! Looks like the travis test failed due to an intermittent test failure with x-ray test suite.

damacus:
Confirmed this also fixes the secrets http endpoint not returning the correct datetime.

sirech:
I was wondering, is this one ready to be merged? I would really like to use it :)

spulec:
Thanks for the help!

TopherGopher:
@spulec I saw this was merged in a bit ago (awesome). What is your timeline for releases?

djmrr:
@TopherGopher this is rad, also curious about a release timeline :)

PR #1903, opened 2018-10-26
Remove microsleep

alsuren:


spulec:
Looks like Travis is failing

bpandola:
This PR had gotten pretty stale, but replacing the `sleep()` call with `threading.Condition` is the Right Thing to Do here.  I brought this up-to-date with the latest on `master`, fixed a bug in the original implementation (which caused the Travis failure mentioned by @spulec), and added an explicit test case.

alsuren:
Oh wow. I wrote this at a company that was in the middle of porting from rabbitmq to sqs, and needed to have something for testing in CI. We spotted the sleep because the tests got slower when we switched the SQS feature on, and the sleep was showing up when we ran it under a sampling profiler. I then forgot about it, because I wasn't even supposed to be on the team doing the port.

I don't think I know anyone who still works at this company. Hopefully they've not still got things pointed at that branch, because I just deleted it ;-).

PR #1905, opened 2018-10-26
Allow installing newer versions of python-jose

illia-v:


spulec:
Closed in favor of https://github.com/spulec/moto/pull/1927

PR #1907, opened 2018-10-29
Get all tests working with latest botocore 

lhufnagel:
This is a rebased version of https://github.com/spulec/moto/pull/1847 including [this commit](https://github.com/joguSD/moto/pull/1/commits/86671950058453e56c0f73c64a32d28796d77f25)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,6 +23,8 @@ matrix:
       sudo: true
 before_install:
   - export BOTO_CONFIG=/dev/null
+  - export AWS_SECRET_ACCESS_KEY=foobar_secret
+  - export AWS_ACCESS_KEY_ID=foobar_key
 install:
   # We build moto first so the docker container doesn't try to compile it as well, also note we don't use
   # -d for docker run so the logs show up in travis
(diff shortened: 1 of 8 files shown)


lhufnagel:
Hm, looks like `policy()` in `LambdaResponses` is not tested in server mode..

jrideout:
I tried upgrading to 1.3.7 and found some sort of issue with the monkey patching that occurs only when other tests (without moto) are run first. I suspect something is getting imported that prevents the patching. When I run just test files with moto imported, they all pass. When I run my whole test suite, or even just two files (one without moto, but that still imports some of our project that has boto3 imports), provided the test without moto runs first, the moto tests will fail with IAM permission issues.

summermon:
I am still seeing error:
botocore.exceptions.ClientError: An error occurred (InvalidAccessKeyId) when calling the CreateBucket operation: The AWS Access Key Id you provided does not exist in our records.

Version I am using:
boto==2.49.0
boto3==1.9.38
botocore==1.12.38
moto==1.3.7

PR #1801, opened 2018-08-29
Fix botocore/boto3 dependency version mismatch

meneal:
Pinning boto3 to this version would remove the mismatch between
boto3 and the botocore dependency in setup.py.

Closes #1800

PR #1910, opened 2018-10-29
Pin boto3/botocore to latest working versions for dev.

markchalloner:
When installing dev requirements boto changes mean [real resources are created on AWS](https://github.com/spulec/moto/issues/1793) This removes the separate requirements in requirements-dev.txt upgrading boto from the versions pinned by #1801 in setup.py.


markchalloner:
Unnecessary due to #1907.

PR #1912, opened 2018-10-29
Be able to change user attributes in cognito-idp.

garyd203:
Update existing partial implementation to store attributes as a Python dict, and only convert to a list-of-name-value-pairs when required

### DIFF
--- a/IMPLEMENTATION_COVERAGE.md
+++ b/IMPLEMENTATION_COVERAGE.md
@@ -827,13 +827,13 @@
 - [ ] unlink_identity
 - [ ] update_identity_pool
 
-## cognito-idp - 34% implemented
+## cognito-idp - 37% implemented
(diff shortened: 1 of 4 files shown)


coyotecote:
Will this PR get merged to master anytime soon? I need the "admin_update_user_attributes" API. Thanks

garyd203:
I have rebased this PR to be against current git master branch, in case that helps it be accepted

mikegrima:
@coyotecote Have you tested if this PR works for you locally?

bblommers:
The `admin_update_user_attributes` was merged with #2185, and `admin_delete_user_attributes` was merged with #4487, so the features of this PR are all implemented now as far as I can tell.

Will close this, but let me know if I've missed anything.

garyd203:
Thanks @bblommers :-)

PR #1899, opened 2018-10-25
Add support for ReturnValues in dynamodb2 put_item

Ben-Wu:
This is a fix for issue https://github.com/spulec/moto/issues/1404

`put_item` currently always returns the new item but the return value should be either `NONE` or `ALL_OLD`.  This also applies to other actions like `update_item` but I wanted to look at one for now.

Here's a notebook showing the behaviour using a real dynamodb and a mock: https://gist.github.com/Ben-Wu/0a0d09cbe2c8f365ef78e5960525c115

PR #1919, opened 2018-11-01
Make CreatedDate an integer as per AWS API spec

damacus:
[AWS SecretsManager API documentation](https://docs.aws.amazon.com/secretsmanager/latest/apireference/API_GetSecretValue.html#SecretsManager-GetSecretValue-response-CreatedDate).

```json
{
   "ARN": "string",
   "CreatedDate": number,
   "Name": "string",
   "SecretBinary": blob,
   "SecretString": "string",
   "VersionId": "string",
   "VersionStages": [ "string" ]
}
```

In particular the value should be a number rather than an ISO-8601 date format.

How we're testing this currently using the Boto3 library means we're not able to confirm the correct return type as we use Boto3 to get a client. 

Boto3 appears to convert any response to it's own format.

### DIFF
--- a/moto/secretsmanager/models.py
+++ b/moto/secretsmanager/models.py
@@ -31,7 +31,7 @@ class SecretsManagerBackend(BaseBackend):
         self.region = region_name
         self.secret_id = kwargs.get('secret_id', '')
         self.name = kwargs.get('name', '')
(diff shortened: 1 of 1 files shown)


damacus:
Closing as it looks like #1899 fixes this 🎉

PR #1920, opened 2018-11-01
Support IAM Credential Report in boto3

pall-valmundsson:
Lowercase XML element names in API responses seem to cause issues for boto3.


spulec:
Thanks!

PR #1927, opened 2018-11-06
Update to jose 3.x

justin8:
I'm working with an environment that is using jose 3.x for things, and it appears that forcing versions >3.0.0 was a legacy from before 3.x came out.

It appears that the usage of jose in moto works just fine with jose 3.0.1


spulec:
Thanks!

PR #1932, opened 2018-11-07
Fix IAM role name when created from CloudFormation JSON

luispollo:
I noticed that, when creating mocked IAM roles using the boto CloudFormation client, and subsequently trying to find that same role using the IAM client, that failed. The reason was that the logical resource name from the CloudFormation template was being used for the role name, not the `RoleName` property.

This fixes that.


spulec:
Hmm, it looks like Travis is failing.

luispollo:
@spulec I noticed, but didn’t have a chance to fix the tests. Will take a look after the holidays.

spulec:
Solved in https://github.com/spulec/moto/pull/2802

Issue #1923, opened 2018-11-05
create_image - possibility to mangle/set creation date (testing purposes)


Constantin07:
I'm working on testing a cleanup script for AMI images.
For deleting images it looks for image count as well as creation date. e.g.

- keep last 3 images no matter how old
- keep images newer than 1 day

In test script I create new images (around 10) but I cannot test it properly because creation date is almost the same  and the deletion is skipped based on the second criteria above.

```
ami-4db63882 Name: project 1541429211 Description: Created from i-1776d2bedafa7591a Creation date: 2018-11-05T14:46:51.000Z
ami-96214ce5 Name: projec 1541429211 Description: Created from i-1776d2bedafa7591a Creation date: 2018-11-05T14:46:51.000Z
ami-4a994a27 Name: project 1541429211 Description: Created from i-1776d2bedafa7591a Creation date: 2018-11-05T14:46:51.000Z
ami-38fcdfc0 Name: project 1541429211 Description: Created from i-1776d2bedafa7591a Creation date: 2018-11-05T14:46:51.000Z
ami-ce1a7575 Name: project 1541429211 Description: Created from i-1776d2bedafa7591a Creation date: 2018-11-05T14:46:51.000Z
```

Is there any way to set/mangle the creation date of images ?

spulec:
Closed as similar to https://github.com/spulec/moto/pull/1936

PR #1936, opened 2018-11-08
#1923 Allow to create images with creation date in past (for testing purposes)

Constantin07:
This PR is to address issue reported in #1923.

Reason: In order to test some clean up scripts I need to be able to create in test fixture some AMIs (properly named, tagged, etc.) but with creation date in the past.

The test script checks the datetime of AMI as a criteria what to keep and what to delete.

### DIFF
--- a/moto/ec2/models.py
+++ b/moto/ec2/models.py
@@ -1150,6 +1150,7 @@ class AmiBackend(object):
 
         ami = Ami(self, ami_id, instance=instance, source_ami=None,
                   name=name, description=description,
+                  creation_date=None,
                   owner_id=context.get_current_user() if context else '111122223333')
         self.amis[ami_id] = ami
         return ami


spulec:
It isn't clear to me what this changes. Won't it still set the current time as the creation date in the AMI instantiation?

Constantin07:
the idea was to be able to pass `creation_date` value in the past instead of using default (current time).

spulec:
Maybe I'm missing it, but it seems like this change still won't allow you to pass a `creation_date`. Where would you be passing it?

spulec:
Some example code of how you are planning on using this would be very helpful.

queue-tip:
I don't think this PR is necessary. the Ami class in ect/models.py already sets `creation_date` to the current time. For testing, you could import `freeze_time` from freezegun and create images at any point in time. This will only work when not in server_mode.

spulec:
I tend to agree with @queue-tip . I'm going to close this out, but let me know if I am missing something.

Issue #1937, opened 2018-11-08
Mock dynamodb2 update_item doesn't return attributes on new items


kgutwin:
Calling `update_item` to create a new item does not return any attributes, but the same call when the item already exists does return attributes. This is with the latest moto (1.3.7).

    >>> from moto import mock_dynamodb2
    >>> md = mock_dynamodb2()
    >>> md.start()
    >>> import boto3
    >>> conn = boto3.client('dynamodb')
    >>> conn.create_table(
    ...     TableName='test-moto',
    ...     AttributeDefinitions=[{'AttributeName': 'id', 'AttributeType': 'S'}],
    ...     KeySchema=[{'AttributeName': 'id', 'KeyType': 'HASH'}],
    ...     ProvisionedThroughput={'ReadCapacityUnits': 1, 'WriteCapacityUnits': 1})
    ...
    >>> r = conn.update_item(
    ...     TableName='test-moto',
    ...     Key={'id': {'S': 'foo'}},
    ...     AttributeUpdates={'col1': {'Value': {'S': 'val1'}, 'Action': 'PUT'}},
    ...     ReturnValues='ALL_NEW')
    ...
    >>> r['Attributes']
    {}
    >>> r = conn.update_item(
    ...     TableName='test-moto',
    ...     Key={'id': {'S': 'foo'}},
    ...     AttributeUpdates={'col1': {'Value': {'S': 'val2'}, 'Action': 'PUT'}},
    ...     ReturnValues='ALL_NEW')
    ...
    >>> r['Attributes']
    {'id': {'S': 'foo'}, 'col1': {'S': 'val2'}}

spulec:
Fixed with https://github.com/spulec/moto/pull/1938

PR #1938, opened 2018-11-08
Correct behavior of ReturnValues parameter to put_item and update_item

kgutwin:
Resolve issue #1937 by correctly returning attributes as specified by the ReturnValues parameter to the DynamoDB `put_item` and `update_item` methods.


spulec:
Thanks!

Issue #380, opened 2015-07-22
feature: resource state transition


galvinograd:
It could be useful and more realistic if there where some sort state transition.
I think it's best if it would be implemented based on the times job flow has been pulled, for example, after 5 calls to describe_job_flow a specific job flow will transition from STARTING to WAITING.

What do you think?

spulec:
We used to have something like this for EC2, but it ended up causing confusion among people.

One way for you to do this is to use some private Moto APIs:

```python
from moto.emr import emr_backends

emr_backends['us-east-1'].job_flows[job_id].state = "WAITING"
```

and then you can change it as needed.

I do wish we could find a better solution than this, but I don't want to have to add additional APIs or documentation to allow it. The simplicity of how the project interacts with the AWS APIs right now is nice.

galvinograd:
I agree that unit tests should expect a simple and consistent behavior.

Although I think that in contrast to mocking, moto_server should reflect as accurately as possible AWS usage as it's mostly used for integration tests where the time axis plays an important role.
For example, a typical boto code would probably look like this: call a function, sleep, assert completed, repeat.

I think architecturally appending workers at moto_server startup that updates the backends' values solves that kind of logic separation.

What do you think? (I would be glad to implement it :)

spulec:
If we were going to do this, I would want it to be opt in (`@mock_emr(slow_mode=True)`). I would also prefer using the resource creation/update time to determine the status to avoid having to add extra processes/threads. For example, ec2 instances can say they are available 30 seconds after they are created.

Thoughts?

donato:
I'm going through a similar problem.

I have a helper which launches an EMR, does bootstrapping and then waits for the clusters state to go to "WAITING"

Any tips for testing this?

spulec:
@donato you should be able to do something similar to the code I posted above.

maxwellchandler:
Im having trouble using the 
`from moto.emr import emr_backends

emr_backends['us-east-1'].job_flows[job_id].state = "WAITING"`

suggestion in 2020. It doesnt seem like the moto backends have a job_flows attribute anymore. Is there a new way to update a the state of a job?

maxwellchandler:
If anyone is still interested the best way to do this now is something like: `athena_backends['us-east-2'].executions.get(query_id).status = "SUCCEEDED"`

dchudz:
> For example, ec2 instances can say they are available 30 seconds after they are created.

[...]

Issue #192, opened 2014-09-09
Realistic timers


tarekziade:
Creating an instance or any of these operations, can take quite some time.

It would be great if moto had a mode where it tries to make the API calls as "slow" as AWS.

spulec:
Two ideas from #1940:

- Utilize [Boto events](https://botocore.amazonaws.com/v1/documentation/api/latest/topics/events.html)
- Use an approach similar to what Freezegun with the concept of [ticking forward time](https://github.com/spulec/freezegun#manual-ticks)

Either of these approaches would need to work with server-based mode.

bblommers:
Moto 3.1.8 now contains a state manager, that would allow you to artificially slow down how fast Moto moves through states, like how long it takes to get from `initializing` to `ready`, etc.

The general documentation can be found here:
http://docs.getmoto.org/en/latest/docs/configuration/state_transition/index.html

Note that while the framework is in place, it still has to be implemented for every model separately. The documentation contains a list of models that currently support this.

Given the age of this ticket, I'm going to close this now. Please open a new issue if you have any questions around this, or if you'd like to see support for specific models.

PR #1940, opened 2018-11-09
add support for ECS desiredCount mocking

thehesiod:
This adds a 10s delay between setting desiredCount, and it being reflected in runningCount


thehesiod:
let me know what you guys think of the pendingCount changes in the unittests. Also the tests can be sped up by mocking of threading.Timer...let me know if you guys want me do that

terrycain:
You could mock threading.Timer but that wouldn't work when moto is running in server mode would it?

thehesiod:
Perhaps tests only run outside server mode?

spulec:
Should we have a more general idea how we want to solve https://github.com/spulec/moto/issues/380 and https://github.com/spulec/moto/issues/192? It seems like this issue will come up for many services.

thehesiod:
I think the general solution I think could build on this, by defining the timeout periods, etc

spulec:
As a first step, it would be great to agree on our philosophy for this. I think we will need it for the docs anyway.

How does moto think of state changes? Are they immediate or delayed? If delayed, can the user choose a timeframe for the change or can they make it immediate?

Freezegun has this concept of [ticking forward time](https://github.com/spulec/freezegun#manual-ticks). Should we be providing something similar for state?

Apologies for blocking this very helpful work on a broader conversation, but I think these questions are fundamental and this will be very helpful across all of Moto.

My initial thoughts: I like state changes not being immediate by default, but the ability for users to globally configure how long a state change should take (including the option of setting that delay at zero). Alternatively, I also like the idea of a user being able to tick forward state like Freezegun.

cc @JackDanger @mikegrima in case they might have opinions too

mikegrima:
I think I would need more context. 

However, I am generally against adding threading, timing delays, and other complex logic to moto. I agree with @spulec in this comment: https://github.com/spulec/moto/issues/380#issuecomment-123932483 -- if you have a need to do more special logic, then mock out the backend interface for your testing -- which is generally simple to do with pytest fixtures. In fact, I generally do with with IAM, which has a constant AWS account ID that is used to formulate the ARN -- I simply mock this value to the AWS account ID I am expecting.

It's very difficult to predict how users will expect their tests to operate, and I would prefer if we simply represent the data structure that AWS would be expected to return. Of course, you are then free to make whatever complex mocking logic you would like on your end to adjust the backend data for the test that you are looking to run.

castrapel:
What is the specific use case for this timeout?

If it's a common one, I'd feel comfortable if the default behavior is simply returning the expected response (current behavior), but allowing configurable options for enhanced behavior doesn't hurt.

thehesiod:
it sounds like for my simple test perhaps to avoid the timeouts I'd have to do the whole work on actually supporting bringing some form of simple docker-based ECS clusters up/down.  That would solve my particular use case.

The timeouts in this particular case are to support the different task states: `pending, desired, and running` counts to ensure I test these three codepaths.

thehesiod:
Freezegun idea is interesting, for me this would need to be supported in server mode. Also the botocore events model is interesting, where you could get called back to dynamically set the timeouts.

spulec:
Yeah, I like the freezegun idea. We could add an endpoint for moto-core for server mode similar to the [reset one](https://github.com/spulec/moto/blob/master/moto/core/urls.py#L13).

I'm not sure what you mean with the botocore events model. Can you explain that more?

thehesiod:
See https://botocore.amazonaws.com/v1/documentation/api/latest/topics/events.html. Basically you can register for various events to either observe or modify behavior

thehesiod:
So perhaps could create events like `before-transition.ecs.task.mode` or something

spulec:
It seems like a thread-based approach probably isn't our best option, so going to close this for now and copy the ideas here over to #192.

Issue #1924, opened 2018-11-05
NoCredentialsError if no credentials defined


t-chaik:
Testing 1.3.7 and the fix for #1793:

```
boto                     2.49.0
boto3                    1.9.37
botocore                 1.12.37
moto                     1.3.7
```

Trying to run the [context manager example](https://github.com/spulec/moto#context-manager):

```python
Python 3.6.7 (default, Oct 21 2018, 08:08:16) 
[GCC 8.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import boto3
>>> from moto import mock_s3
>>> with mock_s3():
...     conn = boto3.resource('s3', region_name='us-east-1')
...     conn.create_bucket(Bucket='mybucket')
... 

Traceback (most recent call last):
  File "<stdin>", line 3, in <module>
  File "/home/martinblanchard/Work/env/buildgrid/lib/python3.6/site-packages/boto3/resources/factory.py", line 520, in do_action
    response = action(self, *args, **kwargs)
  File "/home/martinblanchard/Work/env/buildgrid/lib/python3.6/site-packages/boto3/resources/action.py", line 83, in __call__
    response = getattr(parent.meta.client, operation_name)(**params)
  File "/home/martinblanchard/Work/env/buildgrid/lib/python3.6/site-packages/botocore/client.py", line 320, in _api_call
    return self._make_api_call(operation_name, kwargs)
  File "/home/martinblanchard/Work/env/buildgrid/lib/python3.6/site-packages/botocore/client.py", line 610, in _make_api_call
    operation_model, request_dict)
  File "/home/martinblanchard/Work/env/buildgrid/lib/python3.6/site-packages/botocore/endpoint.py", line 102, in make_request
    return self._send_request(request_dict, operation_model)
  File "/home/martinblanchard/Work/env/buildgrid/lib/python3.6/site-packages/botocore/endpoint.py", line 132, in _send_request
    request = self.create_request(request_dict, operation_model)
  File "/home/martinblanchard/Work/env/buildgrid/lib/python3.6/site-packages/botocore/endpoint.py", line 116, in create_request
    operation_name=operation_model.name)
  File "/home/martinblanchard/Work/env/buildgrid/lib/python3.6/site-packages/botocore/hooks.py", line 356, in emit
    return self._emitter.emit(aliased_event_name, **kwargs)
[...]

PR #1952, opened 2018-11-16
Mock AWS credentials

lhufnagel:
This should allow Moto-users to keep using existing tests without having to export dummy credentials from v1.3.7 onward. See also https://github.com/spulec/moto/issues/1924

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,8 +23,6 @@ matrix:
       sudo: true
 before_install:
   - export BOTO_CONFIG=/dev/null
-  - export AWS_SECRET_ACCESS_KEY=foobar_secret
-  - export AWS_ACCESS_KEY_ID=foobar_key
 install:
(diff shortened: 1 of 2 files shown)


jakul:
@spulec Can this PR be merged?

kgutwin:
While this isn't at all my use case, wouldn't this change affect users who mix-and-match Moto mocks and real AWS calls in the same test run? I know it uses `os.environ.setdefault` to avoid overwriting environment variables, but this would disrupt any case when valid credentials are present in `~/.aws` or instance metadata.

spulec:
Thanks!

yan12125:
Theses changes appear to break a test suite written by another developer: https://github.com/spulec/moto/issues/2058

s0undt3ch:
> Theses changes appear to break a test suite written by another developer: #2058
Same here.
Don't make these changes globally. If its for tests, do it in the tests...

PR #1956, opened 2018-11-19
Check for stack name in use when creating stack

### LINKED ISSUE #1954: CloudFormation allows for creating stacks with duplicate names
Using the CloudFormation moto service allows for creating stacks with the same name - the docs state that names must be unique in a region:

> The name must be unique in the region in which you are creating the stack.
https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/cloudformation.html#CloudFormation.Client.create_stack

Version: moto-1.3.7

I would expect that the API return an `AlreadyExistsException` when trying to do this for real.

grahamlyons:
The AWS API requires that stack names are unique within a region.
Without this logic moto allows stacks with duplicate names to be created, albeit with differing IDs.

Fixes #1954

### DIFF
--- a/moto/cloudformation/models.py
+++ b/moto/cloudformation/models.py
@@ -7,6 +7,7 @@ import uuid
 import boto.cloudformation
 from moto.compat import OrderedDict
 from moto.core import BaseBackend, BaseModel
+from botocore.exceptions import ClientError
 
 from .parsing import ResourceMap, OutputMap
 from .utils import (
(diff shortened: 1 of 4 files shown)


spulec:

moto/cloudformation/models.py
```diff
@@ -150,6 +151,15 @@ def __init__(self):
         self.change_sets = OrderedDict()
 
     def create_stack(self, name, template, parameters, region_name, notification_arns=None, tags=None, role_arn=None, create_change_set=False):
+        for _, stack in self.stacks.items():
```
terrycain: fyi, you could have done `for stack in self.stacks.values():`

bblommers:
This has now implemented in #3072, very much inspired by your work here.
Closing this

PR #1957, opened 2018-11-19
Fix the instantiation of `ClientError`

grahamlyons:
The previous code would throw an exception if it was executed. The `ClientError` class takes a `dict` as the first argument so passing a `str` gives an error like this:

```
...
  File "/home/foo/top-secret-work/moto/moto/redshift/models.py", line 478, in enable_snapshot_copy
    'SnapshotCopyGrantName is required for Snapshot Copy '
  File "/usr/local/lib/python3.7/site-packages/botocore/exceptions.py", line 396, in __init__
    error = error_response.get('Error', {})
AttributeError: 'str' object has no attribute 'get'
```

### DIFF
--- a/moto/redshift/models.py
+++ b/moto/redshift/models.py
@@ -476,10 +476,9 @@ class RedshiftBackend(BaseBackend):
                 raise ClientError(
                     {'Error': {
                         'Message': 'SnapshotCopyGrantName is required for '
-                            'Snapshot Copy on KMS encrypted clusters.',
+                                   'Snapshot Copy on KMS encrypted clusters.',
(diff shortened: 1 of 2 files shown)


spulec:

bpandola:
@grahamlyons Your identification of a bug is correct, but your fix is not.  Instead of raising a `ClientError` in `models.py`, you can just do this:
```
raise InvalidParameterValueError(
    'SnapshotCopyGrantName is required for Snapshot Copy on KMS encrypted clusters.'
)
```

bblommers:
This has now been fixed by #3485, so I'll close this one.

PR #1958, opened 2018-11-19
Add functionalities for SAML Providers

Joeskyyy:
Adds ability to do CRUD mocking for IAM SAML provider types :D

### DIFF
--- a/moto/iam/models.py
+++ b/moto/iam/models.py
@@ -50,6 +50,16 @@ class Policy(BaseModel):
         self.update_datetime = datetime.now(pytz.utc)
 
 
+class SAMLProvider(BaseModel):
+    def __init__(self, name, saml_metadata_document=None):
+        self.name = name
+        self.saml_metadata_document = saml_metadata_document
+
+    @property
+    def arn(self):
+        return "arn:aws:iam::{0}:saml-provider/{1}".format(ACCOUNT_ID, self.name)
+
+
 class PolicyVersion(object):
 
     def __init__(self,
@@ -427,6 +437,7 @@ class IAMBackend(BaseBackend):
         self.credential_report = None
         self.managed_policies = self._init_managed_policies()
         self.account_aliases = []
+        self.saml_providers = {}
         super(IAMBackend, self).__init__()
 
     def _init_managed_policies(self):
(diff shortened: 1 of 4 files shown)


moto/iam/models.py
```diff
@@ -937,5 +948,33 @@ def get_account_authorization_details(self, filter):
...
+        return saml_provider
+
+    def update_saml_provider(self, saml_provider_arn, saml_metadata_document):
+        saml_provider = self.get_saml_provider(saml_provider_arn)
+        saml_provider.saml_metadata_document = saml_metadata_document
+        return saml_provider
+
+    def delete_saml_provider(self, saml_provider_arn):
+        try:
+            for saml_provider in list(self.list_saml_providers()):
```
Joeskyyy: This was done because it made more sense (chicken and egg) in my opinion to store the objects by their `Name` when they get created. By virtue of that, we can't query on the ARN for the object intrinsically.

I had to cast this as a `list()` so that way it makes a copy, otherwise we get a race condition:

```
RuntimeError: dictionary changed size during iteration
```

tests/test_iam/test_iam.py
```diff
@@ -756,5 +758,49 @@ def test_get_account_authorization_details():
...
+        SAMLMetadataDocument='a' * 1024
+    )
+    response['SAMLProviderArn'].should.equal("arn:aws:iam::123456789012:saml-provider/TestSAMLProvider")
 
+@mock_iam()
+def test_get_saml_provider():
+    conn = boto3.client('iam', region_name='us-east-1')
+    saml_provider_create = conn.create_saml_provider(
+        Name="TestSAMLProvider",
+        SAMLMetadataDocument='a' * 1024
```
Joeskyyy: This was done because you have to have a minimum of 1k characters in your SAML document. I had originally wanted to pass in a real SAML document, but then it started getting super crowded, or reading from the OS for a SAML file. Additionally, there appears to be some weird issues with having XML in XML when the botocore parser is invoked from Moto during a `make test` 🤷‍♀️ happy to dig further but not sure how much value we truly get from it.

spulec:
Looks great; thanks!

PR #1964, opened 2018-11-22
Check bucket name length at CreateBucket

jbmchuck:
Check that s3 bucket names follow the documented length restriction:

'Bucket names must be at least 3 and no more than 63 characters long.'

See https://docs.aws.amazon.com/AmazonS3/latest/dev/BucketRestrictions.html


spulec:
Thanks!

PR #1965, opened 2018-11-22
Added double to to_str/from_str, returned by emr create_cluster

andrewmccall:
When I tried to use @mock_emr and create a cluster some of the values from the spec for InstanceFleets are double.


spulec:
Thanks!

PR #1967, opened 2018-11-26
Feature/update record file when delete

sasaki-tomoya:


spulec:

PR #1969, opened 2018-11-27
Return the deleted stacks in the list stacks call

### LINKED ISSUE #1968: Listing CloudFormation stacks should include deleted stacks
Using the AWS API directly returns stacks with a status `DELETE_COMPLETE`; using moto those stacks are not returned.

The existing unit tests actually test for this incorrect behaviour:
https://github.com/spulec/moto/blob/4a275cc/tests/test_cloudformation/test_cloudformation_stack_crud_boto3.py#L516-L518

The tests should be updated and the `list_stacks` method on the backend should include the `self.deleted_stacks` data (https://github.com/spulec/moto/blob/4a275cc/moto/cloudformation/models.py#L225-L226)

grahamlyons:
This matches the behaviour of the AWS API.

Fixes #1968

### DIFF
--- a/moto/cloudformation/models.py
+++ b/moto/cloudformation/models.py
@@ -223,7 +223,11 @@ class CloudFormationBackend(BaseBackend):
             return list(stacks)
 
     def list_stacks(self):
-        return self.stacks.values()
+        return [
+            v for v in self.stacks.values()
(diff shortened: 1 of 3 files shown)


spulec:
Thanks!

PR #1977, opened 2018-11-28
Strip parenthesis in the KeyConditionExpression

vadym-serdiuk:
The "bloop" package uses parenthesis in the KeyConditionExpression, so the "query" method returns nothing due to the wrong parsing of the parameters.


spulec:
Thanks!

PR #1985, opened 2018-12-04
Add instance AZ support in autoscaling

### LINKED ISSUE #1984: moto's autoscaling instances do not have availability zones
This issue relates to #1288.

Currently, when creating instances for auto scaling groups (ASGs), we do not provide the instances with an availability zone (or placement). Additionally, when calls are made to describe the instances, the region `us-east-1e` is hard coded into the return templates.

kawaiwanyelp:
This issue refers to/fixes #1984.

### Fix
- I have made changes to ASG initialization that properly sets and updates availability zones (both direct setting, and if there are VPC identifiers that need to be taken into account).
- I have made changes to propagate one of an ASG's availability zone to its instances when creating them.
- I have made changes to the return templates for an ASG's describe methods (`describe_auto_scaling_groups`, `describe_auto_scaling_instances`) so that instead of returning a hardcoded AZ (`us-east-1e`), a 'real' AZ is returned.
- I have written tests for the above changes.

### DIFF
--- a/moto/autoscaling/models.py
+++ b/moto/autoscaling/models.py
@@ -1,4 +1,7 @@
 from __future__ import unicode_literals
+
+import random
+
 from boto.ec2.blockdevicemapping import BlockDeviceType, BlockDeviceMapping
 from moto.compat import OrderedDict
 from moto.core import BaseBackend, BaseModel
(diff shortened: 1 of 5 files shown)


spulec:

kawaiwanyelp:
@spulec I've committed a fix but Travis doesn't seem to be running. Do you happen to know why?

drmorr0:
@spulec we can close this PR in favor of #2030.  Thanks!

PR #1992, opened 2018-12-07
Replace pyaml with PyYAML dependency

### LINKED ISSUE #1991: Replace pyaml dependency with PyYAML
There is a dependency on pyaml in setup.py:

https://github.com/spulec/moto/blob/master/setup.py#L18

I think that this is intended to be PyYAML (which pyaml depends on), and I do not see any usages of pyaml itself in this codebase.

pyaml uses WTFPL (https://github.com/mk-fg/pretty-yaml/blob/master/COPYING) which is not approved by the OSI (https://opensource.org/minutes20090304)

TrevorEdwards:
[fix #1991]

Note: I tried simply removing the pyaml dependency and saw travis tests fail, but replacing it with PyYAML allows them to continue to pass.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -15,7 +15,7 @@ install_requires = [
     "xmltodict",
     "six>1.9",
     "werkzeug",
-    "pyaml",
+    "PyYAML",
     "pytz",
     "python-dateutil<3.0.0,>=2.1",
     "python-jose<3.0.0",


TrevorEdwards:
I think the build failure was a flake. See https://github.com/TrevorEdwards/moto/pull/2

spulec:
Thanks!

Issue #1994, opened 2018-12-10
Lambda CloudFormation - No support for environment variables


fivepapertigers:
Found my way here from [localstack/localstack](http://github.com/localstack/localstack) via [serverless-localstack](https://github.com/temyers/serverless-localstack). When deploying a new `AWS::Lambda::Function` resource, it seems that the [Environment map](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-lambda-function.html#cfn-lambda-function-environment) is not observed.

I chased it down, and it appears the issue is that `Environment` is not included in the optional configuration: https://github.com/spulec/moto/blob/ed861ecae1039a048a6350a4ff832ef094cdf2c2/moto/awslambda/models.py#L389

I would be happy to open a PR with a fix.

PR #2005, opened 2018-12-29
Lambda Cloudformation

fivepapertigers:
Adds support for the Environment variable map syntax in the `AWS::Lambda::Function` CloudFormation resource. See issue #1994.


fivepapertigers:
👍

Issue #2002, opened 2018-12-20
create_subnet AvailableIpAddressCount always 251


hibbert:
create_subnet AvailableIpAddressCount always 251
```
response = client.create_vpc(
    CidrBlock='10.0.0.0/16'
 )
vpc_id = response["Vpc"]["VpcId"]

# Create small subnet in VPC
small_subnet = client.create_subnet(CidrBlock='10.0.1.0/27', VpcId=vpc_id)
print(small_subnet)
```

Result

```
{'Subnet': {'AvailabilityZone': 'us-east-1a', 'AvailableIpAddressCount': 251, 'CidrBlock': '10.0.1.0/27', 'State': 'pending', 'SubnetId': 'subnet-f77b31ee', 'VpcId': 'vpc-65214079', 'Tags': []}, 'ResponseMetadata': {'RequestId': '7a62c49f-347e-4fc4-9331-6e8eEXAMPLE', 'HTTPStatusCode': 200, 'HTTPHeaders': {'server': 'amazon.com'}, 'RetryAttempts': 0}}
```

hibbert:
I have also tried running instances in these subnets and the AvailableIpAddressCount doesn't seem to decrease

PR #2009, opened 2019-01-05
Displays the correct number of available IP addresses for a subnet

alidonmez:
Fixes issue #2002 

create_subnet and describe_subnets displays the correct number of
available IP addresses. The module currently has '251' hardcoded.
This fix shows available IPs minus 5, which are reserved in each
subnet by AWS.

https://docs.aws.amazon.com/vpc/latest/userguide/VPC_Subnets.html#vpc-sizing-ipv4


terrycain:
The available ip address count needs to take into account the number of eni's attached to that subnet

bblommers:
Closing this, as it has been superceded by #2550. 
Thanks again for doing all the leg work, @alidonmez!

PR #2016, opened 2019-01-11
Enable SES feedback via SNS

adriangalera:
Enable SES feedback (Delivery, Bounce, Complaint) via SNS topic.
- SES implements the AWS SES API method set_identity_notification_topic that will save the topic configuration

- `send_email` and `send_raw_email` check the SNS Topic configuration

- The SES feedback messages are static and are extracted from https://docs.aws.amazon.com/ses/latest/DeveloperGuide/notification-contents.html


spulec:
Thanks!

PR #2018, opened 2019-01-15
1987 support transfer acceleration

### LINKED ISSUE #1987: Feature - S3 bucket to support transfer acceleration settings
Implement
* `get_bucket_accelerate_configuration`
* `put_bucket_accelerate_configuration`

Bonus
* If the client connection is opened with `botocore.config.Config(s3={'use_accelerate_endpoint': True}`, then `generate_presigned_url` should return an url with the transfer accelerated endpoint

--> I'll try to give it a go this month

gbataille:
Fixes #1987 

I think I have all the odd cases. let me know what you think


gbataille:
Hello,

This was opened in Jan, any reason why it's not moving? Do you need anything more from me?

gbataille:
FYI, with respect to the coverage decrease... this is due to an except clause I'm not going into. The problem is that `botocore` will not let me create a "malformed" query that would go there.

Still since `moto` can be used with other things than `boto`, I think this error handling should stay.

Therefore I can
- leave it like that (other bucket parameters are not testing this path either)
- mock the moto response but I feel like that'll defeat most of the purpose and I can't see it done in the repo so far
- don't use boto to do the AWS request --> a lot more effort to write the test for not enough value in my opinion.

Let me know.

terrycain:
Hey nice job, i'll merge later today

PR #2019, opened 2019-01-15
Correctly returning ami launch index

acsbendi:
The `run-instances` command returned a constant 0 in the `AmiLaunchIndex` field for every created instance even if multiple instances were created in the same command (specifying the `--count` option with a value other than 1). This PR fixes this, the launch indexes are returned properly.


mikegrima:
@acsbendi Please provide a description for the PR and I will merge it in.

acsbendi:
@mikegrima Provided a short description :)

PR #2023, opened 2019-01-17
add implemented kinesis method describe_stream_summary

vir-mir:
Dear moto team.

I added implementation of the `describe_stream_summary` method for kinesis.


terrycain:
Nice one

PR #2024, opened 2019-01-17
Hide CloudFormation pararamters with NoEcho

### LINKED ISSUE #2021: CloudFormation parameters marked with NoEcho are being echoed
## How to reproduce the issue

```python
import json
from pprint import pprint

import boto3
import moto


def create_stack_and_print_params():
    cf = boto3.resource("cloudformation")

    # Create a stack.
    cf.create_stack(
        StackName="MyCoolStack",
        TemplateBody=json.dumps(
            {
                "Resources": {
                    "MyCoolResource": {
                        "Type": "AWS::SSM::Parameter",
                        "Properties": {
                            "Name": " MyCoolParameter",
                            "Value": "Hey, this is a cool parameter!",
                        },
                    }
                },
                "Parameters": {
                    "JustAString": {"Type": "String"},
                    "NoEchoString": {"Type": "String", "NoEcho": True},
                },
            }
        ),
        Parameters=[
            {"ParameterKey": "JustAString", "ParameterValue": "This is just a string"},
            {
                "ParameterKey": "NoEchoString",
                "ParameterValue": "This should not be echoed",
            },
        ],
    )

    # Get parameters from stack.
    stack = cf.Stack("MyCoolStack")
    pprint(stack.parameters)

    # Delete stack
    stack.delete()


if __name__ == "__main__":
    print("Without moto:")
    create_stack_and_print_params()

    with moto.mock_cloudformation():
        print("With moto:")
        create_stack_and_print_params()
```

Note that the resource here isn't important. It's only present because CloudFormation needs to have at least one resource.

## What I expect to happen

The value of the parameter with NoEcho should be replaced with "****".

```
[{'ParameterKey': 'NoEchoString', 'ParameterValue': '****'},
 {'ParameterKey': 'JustAString', 'ParameterValue': 'This is just a string'}]
```

## What actually happens

The value isn't replaced.

```
[{'ParameterKey': 'JustAString', 'ParameterValue': 'This is just a string'},
 {'ParameterKey': 'NoEchoString',
  'ParameterValue': 'This should not be echoed'}]
```

## Versions and installation method

Using moto 1.3.7, boto3 1.9.79 and botocore 1.12.79 installed with pipenv.

craiga:
Fixes #2021.


craiga:
Looks like tests are failing due to a timeout. Unfortunately I can't trigger another build without adding another commit (AFAIK).

ralewis85:
@craiga if you commit amend with no changes and force push to your branch it will trigger another build.

craiga:
@ralewis85 That's a **great** tip. Thanks so much!

terrycain:
Nice thank

PR #2028, opened 2019-01-21
Network ACL fixes

acsbendi:
I fixed the following smaller inaccuracies/differences from AWS:

- When creating a subnet in a non-default VPC, it was associated with a network ACL from a different VPC than the one the subnet was created in. This was caused by the first (default) VPC's default network ACL being associated with all new subnets regardless of what VPC it was created in.
- When a default network ACL gets created, a number (4) of default routes are automatically added to it. Refer to [the official documentation](https://docs.aws.amazon.com/vpc/latest/userguide/vpc-network-acls.html#default-network-acl).


mikegrima:
@acsbendi Can you provide a summary to the changes being made?

acsbendi:
Yes, although the commit messages describe them pretty well.

PR #2030, opened 2019-01-23
Autoscaling instance azs

### LINKED ISSUE #1984: moto's autoscaling instances do not have availability zones

drmorr0:
This is picking up @kawaiwanyelp's #1985, we can close the other one.  The change in that PR said:

> This issue refers to/fixes #1984.

> * I have made changes to ASG initialization that properly sets and updates availability zones (both direct setting, and if there are VPC identifiers that need to be taken into account).
> * I have made changes to propagate one of an ASG's availability zone to its instances when creating them.
> * I have made changes to the return templates for an ASG's describe methods (describe_auto_scaling_groups, describe_auto_scaling_instances) so that instead of returning a hardcoded AZ (us-east-1e), a 'real' AZ is returned.
> * I have written tests for the above changes.

In addition to the above changes, I'm going through and fixing tests that were broken by the change.  All the broken tests appear to have been relying on a "default" region being set, but we now need a region to be set explicitly.

### DIFF
--- a/moto/autoscaling/models.py
+++ b/moto/autoscaling/models.py
@@ -1,4 +1,7 @@
 from __future__ import unicode_literals
+
+import random
+
 from boto.ec2.blockdevicemapping import BlockDeviceType, BlockDeviceMapping
 from moto.compat import OrderedDict
 from moto.core import BaseBackend, BaseModel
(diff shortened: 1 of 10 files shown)


drmorr0:
@spulec Any thoughts on this PR?

terrycain:
Good job, i'll merge later

Issue #2035, opened 2019-01-25
Latest botocore changed ECR endpoints


william-richard:
botocore 1.12.85 has removed the old ecr endpoints (`ecr.<region>.amazonaws.com`), and replaced them with new endpoints (by the looks of [this part of the changes](https://github.com/boto/botocore/commit/b5fa8a51393e24fbc7337a357a2d51a0f750c627#diff-9dfab05d4ba739e097a193e8b5fa61caR111), the new endpoints are `api.ecr.<region>.amazonaws.com`).

Moto does not yet seem to mock those calls (https://github.com/spulec/moto/blob/master/moto/ecr/urls.py#L5) so it isn't correctly mocking ECR.

I haven't had time to figure out what other changes come with the new ecr api endpoint, so this may or may not be a simple drop-in replacement.

william-richard:
Does [this line](https://github.com/boto/botocore/commit/4c09604380d0de572cb452f2600d3ca87f1fe49e#diff-0d9c16e152c7f468f96dc760cf3a1774R5) mean that the endpoints were just renamed, with no other api changes?

PR #2036, opened 2019-01-25
ECR now uses api.ecr as its endpoint.

william-richard:
This changed in botocore 1.12.85. See https://github.com/boto/botocore/commit/b5fa8a51393e24fbc7337a357a2d51a0f750c627

This fixes https://github.com/spulec/moto/issues/2035


piercefreeman:
@mikegrima @spulec Is it possible to bump a version of the package to pick up on this addition?

PR #1827, opened 2018-09-14
Fix #1826, Add support for IoT register_certificate

ajhodges:
This adds support for importing a certificate by specifying the certificate PEM. Fix for #1826

PR #2037, opened 2019-01-27
Support CustomAmiId in EMR

jmsanders:
This change adds support for passing `CustomAmiId` in EMR's `run_job_flow` and returning `CustomAmiId` in EMR's `describe_cluster`.

https://docs.aws.amazon.com/emr/latest/ManagementGuide/emr-custom-ami.html

I only added tests to `test_emr_boto3.py` since `test_emr.py` appears to be deprecated. If you'd like, I can add tests there as well.

### DIFF
--- a/moto/emr/models.py
+++ b/moto/emr/models.py
@@ -97,7 +97,8 @@ class FakeCluster(BaseModel):
                  visible_to_all_users='false',
                  release_label=None,
                  requested_ami_version=None,
-                 running_ami_version=None):
+                 running_ami_version=None,
+                 custom_ami_id=None):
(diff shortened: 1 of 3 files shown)


jmsanders:
Can somebody point me in the right direction re: why the coverage dropped so much? I see there are some other PRs that have been merged with a similarly large drop in coverage:

https://github.com/spulec/moto/pull/1825
https://github.com/spulec/moto/pull/1827
https://github.com/spulec/moto/pull/1847

terrycain:
The drop in converage is because half the travis builds succeeded but half did not I think...
Let me trigger travis again to see if its any happier.

PR #2041, opened 2019-02-01
IAM Role Tagging support

mikegrima:
IAM Role tagging support.


mikegrima:
@spulec @JackDanger Should probably have someone other than me review and merge :)

mikegrima:
Updated with some fixes for `get_account_authorization_details()` so that it has less pre-canned responses and proper results.

PR #2053, opened 2019-02-08
Updates to create_subnet and describe_subnets responses

acsbendi:
The structure of the responses to the `create_subnet` and the `describe_subnets` commands have changed recently, as testified by the official documentation:

- [`create_subnet`](https://docs.aws.amazon.com/cli/latest/reference/ec2/create-subnet.html#output)
- [`describe_subnets`](https://docs.aws.amazon.com/cli/latest/reference/ec2/describe-subnets.html#output)

I modified the responses returned by moto to precisely mock the original ones. In the process, I also collected all the existing availability zones, because the previous data was outdated and inaccurate.

Although this is not documented, the `Tags` field is only included in the response if it has at least one element, thus the `create_subnet` response never has that field.


terrycain:
Nice job, great work. Fix conflicts and i'll merge

acsbendi:
@terrycain Thank you, fixed now

PR #2056, opened 2019-02-09
Bug fix for default network ACL entries

acsbendi:
Default network ACL entries were saved with an integer rule number instead of a string, so they could not be deleted. This fix is for my [previous PR.](https://github.com/spulec/moto/pull/2028)
This PR also includes validating the rule number, if a network ACL entry exists with the same rule number and egress value, an error is raised.


terrycain:
Looks good fix merge conflicts and i'll have a look

acsbendi:
@terrycain Done:)

PR #2061, opened 2019-02-16
Update kms, iam and lambda to work with terraform

rubelw:
These are just a few of the many PR to come to get moto working with kitchen-terraform.  In this pull request, I'm added a couple functions for lambda and kms - with associated tests.


rubelw:
Couldn't get it to pass tests without adding some additional tests for various packages because my packages have 100% coverage.

moto/iam/responses.py
```diff
@@ -655,6 +673,34 @@ def untag_role(self):
         return template.render()
 
 
+LIST_ENTITIES_FOR_POLICY_TEMPLATE = """<ListEntitiesForPolicyResponse>
+ <ListEntitiesForPolicyResult>
+ <PolicyRoles>
```
mikegrima: Please update this template so that it returns a dynamic list consistent with the AWS API.

rubelw: I has been updated to return dynamic data.

PR #2062, opened 2019-02-16
Fixing S3 copy_key error when using unicode.

domtes:


domtes:
Solves #2062

mikegrima:
TIL you can have emojis in S3 key names.

PR #2082, opened 2019-03-01
Fix route53 alias response.

markchalloner:
This commit
* includes the work by @elliotmb in #1694,
* removes the AliasTarget.DNSName copy into a RecordSet.Value,
* fixes and adds tests.


spulec:
Thanks!

Issue #2079, opened 2019-03-01
Certificate Authority support in AWS IoT


sahilkoladiya:
IoT is missing support for following APIs.

list_things, register_ca_certificate, register_certificate, describe_ca_certificate, list_certificates_by_ca, delete_ca_certificate, update_ca_certificate, get_registration_code

Without this APIs, it's hard to mock a service via moto when someone gets device certificate signed by their own certificate authority vs relying in AWS IoT default CA.

spulec:
Thanks; setting as a feature request.

PR #2089, opened 2019-03-02
IoT is missing support for public key infrastructure features #2079

sahilkoladiya:
list_things, register_ca_certificate, register_certificate, describe_ca_certificate, list_certificates_by_ca, delete_ca_certificate, update_ca_certificate, get_registration_code

Without this APIs, it's hard to mock a service via moto when someone gets device certificate signed by their own certificate authority vs relying in AWS IoT default CA.

Implemented above APIs.

Resolving #2079


mikegrima:
It looks like this was replaced with #2089

PR #2091, opened 2019-03-02
Fix sorting of log groups

whummer:
Fix sorting of log groups - use `creationTime` as the sort key.

python3 doesn't support sorting of lists that contain dicts:
```
$ python3 -c 'sorted([{}, {}])'
  File "<string>", line 1, in <module>
TypeError: '<' not supported between instances of 'dict' and 'dict'
```

Only a minor change - would be great if we could get this merged @mikegrima @spulec . Thanks!

### DIFF
--- a/moto/logs/models.py
+++ b/moto/logs/models.py
@@ -242,7 +242,8 @@ class LogsBackend(BaseBackend):
         if next_token is None:
             next_token = 0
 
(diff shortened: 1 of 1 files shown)


mikegrima:
LGTM -- merging on Monday.

PR #2055, opened 2019-02-09
Fix S3 backend operations with VersionId

thehesiod:

PR #2100, opened 2019-03-08
[S3] Support null version ids for buckets with versioning disabled

gilbsgilbs:
Currently, if versioning is disabled or suspended on a bucket, moto will push a new version of the object instead of replacing the old one. This is very different from what happens in real life and from what is documented on AWS docs. Instead, when the versioning is deactivated or suspended on a bucket, they will use what they call the null version id, which is just like a normal version id, except it's the string `null`.

Less relevant, but moto actually creates predictable, incremental `VersionId`s for versioned objects. According to my tests, AWS uses a random string as `VersionId`. That's kind of logical, a counter wouldn't be a great fit for them as it would require locks and ugly stuff. However, I think using an incremental id might lead to some confusion to moto users because they might wrongly expect the ids to also be incremental in real life.

**What this PR does:**

- Uses an UUID instead of a counter for `VersionId`s to better fit with what AWS actually does and avoid bad expectations when working with VersionIds in moto.
- Stops duplicating objects when versioning is disabled in a bucket.
- Uses the null version id when versioning is disabled.
- Finds latest object by last modification date instead of relying on the `VersionId`

**Some things I had to fix:**

- Some tests did expect the `VersionId` to be a counter. I had to fix them.
- Fixed dead code and wrong assertion in a test case.

Let me know if I can improve anything.

### DIFF
--- a/moto/s3/models.py
+++ b/moto/s3/models.py
@@ -10,6 +10,7 @@ import random
 import string
 import tempfile
 import sys
+import uuid
 
 import six
 
@@ -35,7 +36,7 @@ class FakeDeleteMarker(BaseModel):
         self.key = key
         self.name = key.name
         self.last_modified = datetime.datetime.utcnow()
-        self._version_id = key.version_id + 1
+        self._version_id = str(uuid.uuid4())
 
     @property
     def last_modified_ISO8601(self):
@@ -115,15 +116,16 @@ class FakeKey(BaseModel):
         self.last_modified = datetime.datetime.utcnow()
         self._etag = None  # must recalculate etag
         if self._is_versioned:
-            self._version_id += 1
+            self._version_id = str(uuid.uuid4())
         else:
-            self._is_versioned = 0
+            self._version_id = None
 
     def restore(self, days):
         self._expiry = datetime.datetime.utcnow() + datetime.timedelta(days)
 
-    def increment_version(self):
-        self._version_id += 1
+    def refresh_version(self):
+        self._version_id = str(uuid.uuid4())
+        self.last_modified = datetime.datetime.utcnow()
 
     @property
     def etag(self):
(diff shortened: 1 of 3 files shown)


moto/s3/models.py
```diff
@@ -115,15 +116,14 @@ def append_to_value(self, value):
         self.last_modified = datetime.datetime.utcnow()
         self._etag = None  # must recalculate etag
         if self._is_versioned:
-            self._version_id += 1
-        else:
-            self._is_versioned = 0
```
gilbsgilbs: Note that this was very likely to be a bug. I think the original author wanted to write `self._version_id = 0`.

tests/test_s3/test_s3.py
```diff
@@ -1507,9 +1523,6 @@ def test_boto3_copy_object_with_versioning():
     obj1_version = client.get_object(Bucket='blah', Key='test1')['VersionId']
     obj2_version = client.get_object(Bucket='blah', Key='test2')['VersionId']
 
-    # Versions should be the same
-    obj1_version.should.equal(obj2_version)
```
gilbsgilbs: I'm not sure what the exact intent of the original author was here, but I don't think it's any relevant to the test case. The fact that this assertion was true was just a side-effect of the previous versionid implementation, but had no meaning AFAICT.

mikegrima: I would assume it may have to do with differences between the `list` v1 and v2 APIs? Though I'm just guessing.

mikegrima: Was this test supposed to test the `copy_object` API: https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3.html#S3.Client.copy_object ?

It's not clear what this test is trying to do.  Please feel free to fix it.

mikegrima:
If that one test should be re-done, please fix it. Otherwise, it looks good.

tests/test_s3/test_s3.py
gilbsgilbs: > It's not clear what this test is trying to do. Please feel free to fix it.
> Was this test supposed to test the `copy_object` API

It tests that `copy_object` function actually updates the object version (see below). This is still true and I did not change this behavior. Apart from this specific assertion, the test case looks fine to me. I don't think there's something else to fix.

gilbsgilbs:
@mikegrima Thanks for the prompt review. As I responded, I don't think there's something to fix. Removing suspicious assertion is fine.

gilbsgilbs:
Thanks @mikegrima . That was fast actually 🤯 .

thehesiod:
weird, I had a PR for this since Feb 8: https://github.com/spulec/moto/pull/2055

thehesiod:
in fact, this PR does not fix all the issues, ex: copy-part <sigh>

gilbsgilbs:
@thehesiod I saw your PR before doing mine and I came to the conclusion that it didn't fix the same things. I expected to change (and as far as I know I changed) the only behaviors mentioned in description and nothing else. Therefore I didn't touch the copy-part thing because I have no idea how it behaves with the versionId. I'm just pretty sure it doesn't use integers as versionId and it uses a null versionId when versioning is disabled. But that's all I have changed.

It would help should your PR provided some description and tests relating what it changes. Let's continue the discussion there if you need help or something.

PR #2106, opened 2019-03-12
Add support for iam update_user

### LINKED ISSUE #2105: Add support for IAM update_user
The IAM update_user call isn't currently implemented. I've created a fork with an implementation (and test), so creating a quick issue here to tie the PR to.

andytumelty:
This covers both the NewPath and NewUserName parameters for update_user, but
without regex validation for these values.

I'm in two minds about adding validation. It would be fairly trivial to add validation, but that currently isn't in place on create_user either. Happy to add this If people prefer it. I also didn't fancy adding two new exception classes for the validation errors either (side point: should all validation exceptions be one exception class?)

Fixes #2105

### DIFF
--- a/IMPLEMENTATION_COVERAGE.md
+++ b/IMPLEMENTATION_COVERAGE.md
@@ -2334,7 +2334,7 @@
 - [ ] update_service_specific_credential
 - [X] update_signing_certificate
 - [ ] update_ssh_public_key
-- [ ] update_user
+- [X] update_user
 - [X] upload_server_certificate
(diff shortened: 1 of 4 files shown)


mikegrima:
@andytumelty As far as the `ValidationError` classes, I personally try to reflect what the API returns as much as possible. I will often create separate exception classes if the output is unique to the API call to better simulate what AWS would return.

andytumelty:
Good point @mikegrima  -- I'll bear that in mind, thanks 👍

PR #2112, opened 2019-03-14
Add glue.delete_table endpoint, for allowing tables to be deleted

dkuntz2:
As the title says, this just adds the `delete_table` call to glue.

### DIFF
--- a/moto/glue/models.py
+++ b/moto/glue/models.py
@@ -56,6 +56,14 @@ class GlueBackend(BaseBackend):
         database = self.get_database(database_name)
         return [table for table_name, table in database.tables.items()]
 
+    def delete_table(self, database_name, table_name):
+        database = self.get_database(database_name)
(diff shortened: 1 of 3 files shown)


dkuntz2:
Looking at the errors in travis, I'm not sure how to fix them, as they appear to all be related to cloudformation and yaml parsing. I'm willing to take a stab at fixing them.

From just looking at pyyaml's version history, it looks like they released a new version yesterday which may be causing those issues. Pinning to `3.13` would probably fix the errors, but I don't know if that's something you'd want to do.

dkuntz2:
As a note, pinning pyyaml to 3.13 fixes the cloudformation errors. The remaining two are related to KMS, and appear to be FakeDatetime using UTC, and Datetime using localtime.

terrycain:
Looks good

Issue #2126, opened 2019-03-25
Unit tests are broken


mikegrima:
The following versions of boto:
```
boto==2.49.0
boto3==1.9.120
botocore==1.12.120
```
... seem to be breaking the following tests:

- `test_cloudformation`
- `test_kms`
- `test_s3`

CloudFormation is breaking with:
```
yaml.constructor.ConstructorError: could not determine a constructor for the tag '!Ref'
  in "<unicode string>", line 22, column 18:
              Value: !Ref TagName
                     ^
```

KMS is breaking on:
```
line 663, in test_schedule_key_deletion
    assert response['DeletionDate'] == datetime(2015, 1, 31, 12, 0, tzinfo=tzlocal())
AssertionError
```

S3 is breaking on server tests.

mikegrima:
It looks like the S3 tests are failing because the the `authenticated_client` doesn't seem to be setting a proper `Content-Length` header 🤔:
```
        if not request.headers.get('Content-Length'):  # <--- This is getting tripped
            return 411, {}, "Content-Length required"  
```

mikegrima:
It looks like PyYAML 5.1 broke the CloudFormation tests. Pinning to 3.1 should resolve, but I would rather make it 5.1 compatible.

nijave:
Yep, 3.13 has a CVE against it where loading yaml can execute arbitrary code

kyledecot:
Semi-related–https://github.com/spulec/moto/pull/2047

kgutwin:
The CloudFormation error can be resolved on PyYAML 5.1 with the fix in yaml/pyyaml#266, or you can wait for the release of PyYAML 5.2 which should automatically resolve the issue.

mikegrima:
Part of the issue is that it appears that boto's code is raising the exception. So unless we alter the YAML, we'll need to wait until 5.2 (or a newer version of boto that implements the fixes).

mikegrima:
Related: https://github.com/boto/boto3/issues/1468

kgutwin:
I looked at the CI output and it looks like it isn't boto code ultimately causing the exception, but rather Moto parsing the CloudFormation template in `moto.cloudformation.models`:

https://github.com/spulec/moto/blob/master/moto/cloudformation/models.py#L85-90

Try adding `Loader=yaml.Loader` to the `yaml.load()` call in line 88; that's the fix/workaround.

kgutwin:
[...]

PR #2127, opened 2019-03-25
Fixing broken tests #2126

mikegrima:
- KMS
- S3


mikegrima:
Will work on the CloudFormation tests later.

kgutwin:
The fix in yaml/pyyaml#266 should be useful for fixing the CloudFormation tests.

PR #2128, opened 2019-03-25
Add get_cfn_attribute support for ECS Cluster and Service

nadlerjessie:
This adds `get_cfn_attribute` methods for ECS clusters and services.

Supported `Fn::GetAtt` attributes are in the [docs](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/intrinsic-function-reference-getatt.html), a screen grab of the ECS ones are below

Screen Shot 2019-03-25 at 7 17 23 PM

EDIT: I also fixed the `get_cfn_response` return for iam instance profiles. I'm happy to rename this branch if you'd prefer as it's no longer just ECS outputs!

### DIFF
--- a/moto/ecs/models.py
+++ b/moto/ecs/models.py
@@ -94,6 +94,12 @@ class Cluster(BaseObject):
             # no-op when nothing changed between old and new resources
             return original_resource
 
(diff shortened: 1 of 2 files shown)


spulec:
Thanks!

PR #2145, opened 2019-04-05
Feature: Secrets Manager delete-secret

### LINKED ISSUE #2144: Support Secrets Manager delete-secret
Implement support for `aws secretsmanager delete-secret`.

chriskilding:
Add support for `aws secretsmanager delete-secret` (fixes #2144)

## Questions

- [x] Soft deletion (with recovery window, default 30 days) - can we support this in Moto?


chriskilding:
To avoid users getting the impression that soft deletion with a default 30 day recovery period is supported (which I think is tricky in a mock), the implementation currently demands that the user sets the `ForceDeleteWithoutRecovery` flag to `true` - and fails with `InvalidParameterException` if they do not.

I could use some direction here. Is this the right approach? Or should the implementation just do an immediate deletion and not care whether the user specified the force delete flag? Or is a different approach better?

chroche:
Yes please we need this for our testing.

mikegrima:
moto should try to reflect the behavior of AWS and not make assumptions that fields be included or not included for testing purposes only.

You should be able to implement a soft deletion with an additional flag on the field and corresponding logic.

chriskilding:
Ok, I've now added soft deletion support.

mikegrima:
Very nice!

Issue #2148, opened 2019-04-08
Moto EC2 Tags Not responding the same as aws.


TBurnip:
## Summary of the problem
An Instance can be described like this `instances = client.describe_instances()['Reservations'][0]['Instances']`.
This should respond with a dictionary. One of the items in the dictionary, when there are tags, will be a list of tags. When there are no tags in the list this item will not exist in the dictionary.

### AWS Response
This section below is where the tag key should be. It just isn't present because it doesn't have any tags.
```python
{ ...
  'SourceDestCheck': True,
  'VirtualizationType': 'hvm',
  'CpuOptions': {
    'CoreCount': 1,
    'ThreadsPerCore': 1
  },
  'CapacityReservationSpecification': {
    'CapacityReservationPreference': 'open'
  },
  'HibernationOptions': {
    'Configured': False
  }
}
```

### Moto Response
```python
[{
  'AmiLaunchIndex': 0,
  'ImageId': 'ami-1234abcd',
  'InstanceId': 'i-107f0d2d536b93149',
  'InstanceType': 'm1.small',
  'KernelId': 'None',
  'KeyName': 'None',
  'LaunchTime': datetime.datetime(2019, 4, 8, 14, 29, 7, tzinfo = tzutc()),
  'Monitoring': {
    'State': 'disabled'
  },
  'Placement': {
    'AvailabilityZone': 'us-west-1a',
    'GroupName': '',
    'Tenancy': 'default'
  },
  'PrivateDnsName': 'ip-10-242-127-254.us-west-1.compute.internal',
  'PrivateIpAddress': '10.242.127.254',
  'ProductCodes': [],
  'PublicDnsName': 'None',
  'State': {
    'Code': 80,
    'Name': 'stopped'
  },
  'StateTransitionReason': 'User initiated (2019-04-08 14:29:07 UTC)',
  'Architecture': 'x86_64',
  'BlockDeviceMappings': [{
    'DeviceName': '/dev/sda1',
    'Ebs': {
      'AttachTime': datetime.datetime(2019, 4, 8, 14, 29, 7, tzinfo = tzutc()),
      'DeleteOnTermination': False,
      'Status': 'in-use',
      'VolumeId': 'vol-f3bde407'
    }
  }],
  'ClientToken': 'ABCDE1234567890123',
  'EbsOptimized': False,
  'Hypervisor': 'xen',
  'NetworkInterfaces': [{
    'Attachment': {
[...]

PR #2153, opened 2019-04-15
A fix for tags not mimicking AWS

TBurnip:
This is a very simple fix for the tags problem metioned in #2148 be aware a similar problem may occur in other places. It is also not easy to find what it should act like aside from probing boto3 this means a similar problem will no doubt have to be fixed at somepoint in the future.

The fix is increadibly simple all it does is remove tagSet from the response when there are no tags to be contained within it. This means it acts the same as AWS and by virtue boto3.

### DIFF
--- a/moto/ec2/responses/instances.py
+++ b/moto/ec2/responses/instances.py
@@ -450,6 +450,7 @@ EC2_DESCRIBE_INSTANCES = """<DescribeInstancesResponse xmlns="http://ec2.amazona
                     </blockDeviceMapping>
                     <virtualizationType>{{ instance.virtualization_type }}</virtualizationType>
                     <clientToken>ABCDE1234567890123</clientToken>
+                    {% if instance.get_tags() != [] %}
                     <tagSet>
                       {% for tag in instance.get_tags() %}
                         <item>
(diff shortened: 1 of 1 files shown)


moto/ec2/responses/instances.py
```diff
@@ -450,6 +450,7 @@ def _security_grp_instance_attribute_handler(self):
                     </blockDeviceMapping>
                     <virtualizationType>{{ instance.virtualization_type }}</virtualizationType>
                     <clientToken>ABCDE1234567890123</clientToken>
+                    {% if instance.get_tags() != [] %}
```
mikegrima: Please make this `if not instance.get_tags()` instead.

TBurnip: I have updated the repository to fix this.

mikegrima: Sorry! I'm a moron! It should be `if instance.get_tags()` and then list the tags.

TBurnip: fixed it again. in all fairness it slipped passed me aswell

mikegrima:
For some reason, the last test is failing.  Can you try rebasing off of master and seeing if that fixes it?

TBurnip:
What do you mean rebasing?

mikegrima:
@TBurnip rebase your branch with the current master branch. You would then need to push override your current branch to update the PR: https://help.github.com/en/articles/using-git-rebase-on-the-command-line.

TBurnip:
I have now rebased it and it seems to be passing the checks.

PR #2160, opened 2019-04-18
Add test to verify create_role path defaults to /

copland:
When testing that a role ARN is created as expected, if path is not passed to the create_role call then the ARN is returned with None for the path. 

```
iam_client = boto3.client('iam')
iam_client.create_role(
    RoleName="test-role",
    AssumeRolePolicyDocument=json.dumps(policy_document),
    Description="test-description")
resp = iam_client.get_role(RoleName="test-role")
print(resp['Role']['Arn'])
```

results in `arn:aws:iam::123456789012:roleNonetest-role` being printed.

### DIFF
--- a/tests/test_iam/test_iam.py
+++ b/tests/test_iam/test_iam.py
@@ -1287,4 +1287,9 @@ def test_list_entities_for_policy():
     assert response['PolicyRoles'] == [{'RoleName': 'my-role'}]
 
 
+@mock_iam()
+def test_create_role_no_path():
+    conn = boto3.client('iam', region_name='us-east-1')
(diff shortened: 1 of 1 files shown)


mikegrima:
Can you add a unit test for this?

copland:
After doing some more digging it looks like this was fixed in a commit after 1.3.7 was released. I can reproduce it if I add my test on 1.3.7 but it passes without changes required on master. I pushed up the unit test to prevent regressions but it appears no changes are required.

copland:
I also updated the name of the PR to reflect that.

PR #2167, opened 2019-04-23
BUG FIX: Use cast_value when comparing DynamoType

johnlpuc163:
When returning results in `query` method, results are sorted by `range_key`.
```
results.sort(key=lambda item: item.range_key)
```

However, when comparing `range_key`, it uses `DynamoType.value` instead of `DynamoType.cast_value`. It causes errors when `DynamoType` is a number.
```
>>> '10' > '2'
False
```

### DIFF
--- a/moto/dynamodb2/models.py
+++ b/moto/dynamodb2/models.py
@@ -45,16 +45,16 @@ class DynamoType(object):
         )
 
     def __lt__(self, other):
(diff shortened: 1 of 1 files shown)


mikegrima:
@johnlpuc163 Are there any tests to verify that this is working properly?

PR #2168, opened 2019-04-23
Update instances.py

Rs15z:
This didn't seem quite right.
Pending status can be preceded only by the first launch or stopped state, according to https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-instance-lifecycle.html


terrycain:
What does this solve?

mickeypash:
It looks, correct to me based on the state diagram.

bpandola:
Superseded by #5528

PR #2169, opened 2019-04-24
Uniform IAM datetime ISO 8601 handling

pall-valmundsson:
`str(datetime.utcnow())` returns a timestamp that's not of the same format as the AWS SDK uses, in short it's missing the `T` between the date and the time.

This causes issues for e.g. Terraform and probably other AWS Go SDK users. There seems to be some differences between endpoints whether they return milliseconds or not, the AWS API docs were reviewed and the decision whether to return timestamps with milliseconds or not based on the example response documented.

As the timstamps are generated for uniqueness rather than being hardcoded and then directly cast to a UTC (Z) formed timestamp pytz was removed as timezone correctness is probably not important.

### DIFF
--- a/moto/iam/models.py
+++ b/moto/iam/models.py
@@ -8,9 +8,8 @@ import re
 from cryptography import x509
 from cryptography.hazmat.backends import default_backend
 
-import pytz
 from moto.core import BaseBackend, BaseModel
-from moto.core.utils import iso_8601_datetime_without_milliseconds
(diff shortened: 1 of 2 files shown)


pall-valmundsson:
Further testing shows that using pytz uses `+00:00` instead of `Z` and the standard Go way of parsing ISO 8601 can't handle that either.

terrycain:
Resolve conflicts, then looks good to me, nice catch

pall-valmundsson:
@terrycain conflicts resolved

PR #2170, opened 2019-04-25
Handle UnicodeEncodeError when parsing querystring

garrettheel:
It seems like we always call `parse_qs` on the request body, which may raise a `UnicodeEncodeError` for some bodies. I observed this happening with polylines in particular when using this with DynamoDB.

This just ignores those errors, as if it happens then we wouldn't usually look at the querystring anyway.


garrettheel:
@mikegrima would you mind taking a look?

terrycain:
Looks ok to me, @mikegrima 2nd opinion

PR #2180, opened 2019-05-02
Adds Support for filtering on schedulingStrategy in ECS#list_services

kyledecot:
https://docs.aws.amazon.com/sdk-for-ruby/v3/api/Aws/ECS/Client.html#list_services-instance_method


terrycain:
Looks good :+1:

PR #2186, opened 2019-05-07
Updated delete_cluster() for redshift

redspart:
When running:
```python
client.delete_cluster(ClusterIdentifier="identifier", SkipFinalClusterSnapshot=False, FinalClusterSnapshotIdentifier="Identifier-snapshot")
```

It would delete the cluster but not create the final snapshot. I have extended the functionality of `delete_cluster` such that a final snapshot will be generated. 

[boto3's delete_cluster reference](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/redshift.html?highlight=redshift#Redshift.Client.delete_cluster)

### DIFF
--- a/tests/test_redshift/test_redshift.py
+++ b/tests/test_redshift/test_redshift.py
@@ -9,7 +9,8 @@ from boto.redshift.exceptions import (
     ClusterParameterGroupNotFound,
     ClusterSecurityGroupNotFound,
     ClusterSubnetGroupNotFound,
-    InvalidSubnet,
+    InvalidParameterCombinationFault,
+    InvalidSubnet
 )
 from botocore.exceptions import (
     ClientError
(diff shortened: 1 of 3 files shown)


tests/test_redshift/test_redshift.py
```diff
@@ -339,7 +339,7 @@ def test_create_cluster_with_vpc_security_groups_boto3():
 
 @mock_redshift
 def test_create_cluster_with_iam_roles():
-    iam_roles_arn = ['arn:aws:iam:::role/my-iam-role',]
+    iam_roles_arn = ['arn:aws:iam:::role/my-iam-role', ]
```
terrycain: Why add a space when you could remove the `,`

redspart: Think pycharms or something was being silly. Need me to update it?

terrycain: Nah it'll do for now, was merged yesterday

PR #2029, opened 2019-01-22
Add support for redirect actions on ELBv2 listeners

william-richard:

PR #2200, opened 2019-05-14
Add support for redirect and fixed response listener types

joshlartz:
I built on #2029 to also add support for fixed response listeners as well as adding some additional tests.


moto/elbv2/utils.py
```diff
@@ -1,8 +1,15 @@
+import re
+
+
 def make_arn_for_load_balancer(account_id, name, region_name):
-    return "arn:aws:elasticloadbalancing:{}:{}:loadbalancer/{}/50dc6c495c0c9188".format(
+    return "arn:aws:elasticloadbalancing:{}:{}:loadbalancer/app/{}/50dc6c495c0c9188".format(
```
terrycain: What happens if its a network loadbalancer? should be /net/ right?

joshlartz: That was my initial assumption as well, and I was planning on accepting a parameter to specify which ARN you needed, but it turns out network load balancers aren't actually supported in moto yet. (See line 67 in moto/elbv2/models.py)
I figured I would leave that for a PR specific to implementing network load balancers rather than asserting an opinion on what parameters to use for it up front. 
I can put in a skeleton to allow for that if you think it is valuable at this point.

terrycain: Yeah add in a skeleton just so whoever next appears will have less to change

bblommers:
There have been so many changes since this PR was started, creating a new branch will be easier than trying to merge this. So I'll close this one.
Thank you for your effort on this though @joshlartz!

Issue #2201, opened 2019-05-15
Batch.Client.describe_jobs(...) fails for not-yet-started jobs


mztechins:
## How to reproduce

Pseudocode:
```python
boto3.client('batch', region_name=...)
resp = batch_client.submit_job(...)
job_id = resp['jobId']
resp = batch_client.describe_jobs(jobs=[job_id])
```
or just un the [tests.test_batch.test_batch.test_submit_job](https://github.com/spulec/moto/blob/8cb4db1896521af99ff295b028f604c6ab599401/tests/test_batch/test_batch.py#L652) test.
OS Name:	Microsoft Windows 10 Pro

## Expected behavior

resp holds a valid job jescription for given job_id.

## Actual results

It crashes with:
```
Error
Traceback (most recent call last):
  File "C:\Program Files\Python36\lib\unittest\case.py", line 59, in testPartExecutor
    yield
  File "C:\Program Files\Python36\lib\unittest\case.py", line 605, in run
    testMethod()
  File "E:\opensource\moto\venv\lib\site-packages\nose\case.py", line 198, in runTest
    self.test(*self.arg)
  File "E:\opensource\moto\moto\core\models.py", line 80, in wrapper
    result = func(*args, **kwargs)
  File "E:\opensource\moto\moto\core\models.py", line 80, in wrapper
    result = func(*args, **kwargs)
  File "E:\opensource\moto\moto\core\models.py", line 80, in wrapper
    result = func(*args, **kwargs)
  [Previous line repeated 2 more times]
  File "E:\opensource\moto\tests\test_batch\test_batch.py", line 700, in test_submit_job
    resp = batch_client.describe_jobs(jobs=[job_id])
  File "E:\opensource\moto\venv\lib\site-packages\botocore\client.py", line 357, in _api_call
    return self._make_api_call(operation_name, kwargs)
  File "E:\opensource\moto\venv\lib\site-packages\botocore\client.py", line 648, in _make_api_call
    operation_model, request_dict, request_context)
  File "E:\opensource\moto\venv\lib\site-packages\botocore\client.py", line 667, in _make_request
    return self._endpoint.make_request(operation_model, request_dict)
  File "E:\opensource\moto\venv\lib\site-packages\botocore\endpoint.py", line 102, in make_request
    return self._send_request(request_dict, operation_model)
  File "E:\opensource\moto\venv\lib\site-packages\botocore\endpoint.py", line 137, in _send_request
    success_response, exception):
[...]

PR #2202, opened 2019-05-15
Bug batch describe job fail 1.3.8

mztechins:
It fixes the issue #2201.
I have tested the actual AWS behavior with boto3 and it turned out it doesn't return 'startedAt' from describe_jobs for not-yet-started jobs. This pull request makes moto to reflect that behavior.


mztechins:
Referenced the wrong branch by mistake. There is a new pull request for the right branch.

PR #2203, opened 2019-05-15
Bug batch describe job fail

mztechins:
It fixes the issue #2201.
I have tested the actual AWS behavior with boto3 and it turned out it doesn't return 'startedAt' from describe_jobs for not-yet-started jobs. This pull request makes moto to reflect that behavior.

### DIFF
--- a/moto/batch/models.py
+++ b/moto/batch/models.py
@@ -289,10 +289,11 @@ class Job(threading.Thread, BaseModel):
             'jobId': self.job_id,
             'jobName': self.job_name,
             'jobQueue': self.job_queue.arn,
-            'startedAt': datetime2int(self.job_started_at),
             'status': self.job_state,
             'dependsOn': []
         }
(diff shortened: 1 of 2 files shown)


terrycain:
Fix the tests, other than that it looks ok

mztechins:
Thank you for the response.

I’ve looked at the coverage reports before and after the fix and it seems that the decreased coverage is caused by mktime not raising exception anymore (which is what the fix is all about). Since mktime doesn’t throw now (in tests: test_list_jobs and test_terminate_job, decorated as @expected_failure), certain exception-catching constructs are not executed anymore, which effectively decreases the code coverage.

Should we try to invent some test case to execute those except blocks and thus bring the coverage up again? If not, then we can just neglect that.

Thanks

terrycain:
The coverage looks worse than it is when some but not all of the travis builds pass. Ignore coverage until all builds pass.

spulec:
Added with https://github.com/spulec/moto/pull/2808

Issue #1870, opened 2018-10-11
XML parsing error when writing to S3 via Pyspark


dslaw:
I'm trying to mock writes to s3, using the moto standalone server. Writing to s3 with pyspark produces what looks like an error in the AWS JDK. The stacktrace indicates that it can't parse the returned XML from list objects operation. I can write to a real s3 location fine, though.

Details: python 2.7.15, moto 1.3.6, pyspark 2.3.1, all installed through conda. AWS JDK version is 1.7.4, which seems to be a bit outdated.

Here's an example:

```python
from pyspark.sql import SparkSession
import boto3
import os
import pandas as pd

os.environ["PYSPARK_SUBMIT_ARGS"] = (
    '--packages "org.apache.hadoop:hadoop-aws:2.7.3" pyspark-shell'
)
spark = SparkSession.builder.getOrCreate()

# Setup spark to use s3, and point it to the moto server.
hadoop_conf = spark.sparkContext._jsc.hadoopConfiguration()
hadoop_conf.set("fs.s3.impl", "org.apache.hadoop.fs.s3a.S3AFileSystem")
hadoop_conf.set("fs.s3a.access.key", "mock")
hadoop_conf.set("fs.s3a.secret.key", "mock")
hadoop_conf.set("fs.s3a.endpoint", "http://127.0.0.1:5000")

conn = boto3.resource("s3", endpoint_url="http://127.0.0.1:5000")
conn.create_bucket(Bucket="bucket")

data = pd.DataFrame({"a": [1, 2, 3], "b": [4, 5, 6]})
df = spark.createDataFrame(data)
df.write.csv("s3://bucket/test/out", header=True)
```

```
Py4JJavaError: An error occurred while calling o41.csv.
: com.amazonaws.AmazonClientException: Unable to unmarshall response (Failed to parse XML document with handler class com.amazonaws.services.s3.model.transform.XmlResponsesSaxParser$ListBucketHandler). Response Code: 200, Response Text: OK
        at com.amazonaws.http.AmazonHttpClient.handleResponse(AmazonHttpClient.java:738)
        at com.amazonaws.http.AmazonHttpClient.executeHelper(AmazonHttpClient.java:399)
        at com.amazonaws.http.AmazonHttpClient.execute(AmazonHttpClient.java:232)
        at com.amazonaws.services.s3.AmazonS3Client.invoke(AmazonS3Client.java:3528)
[...]

PR #2205, opened 2019-05-16
Fix for #1870 thanks to dslaw

lyle-nel:
Fix for #1870 

I think modifying the jinja templates wont get rid of the problem since there is nothing ensuring that those are always whitespace free.

I also don't know what the tests(if any) should look like since this is a quirk that is only apparent when interfacing with Moto through Pyspark. Pyspark tests are notoriously slow and bloated and noisy.

### DIFF
--- a/moto/core/responses.py
+++ b/moto/core/responses.py
@@ -98,7 +98,7 @@ class _TemplateEnvironmentMixin(object):
         template_id = id(source)
         if not self.contains_template(template_id):
             # Real response body of AWS APIs dont have whitespace between tags
-            collapsed = "".join(line.strip() for line in source.split("\n"))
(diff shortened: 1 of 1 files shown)


lyle-nel:
I will fix the tests later, let me know if I need to close this for the time being.

terrycain:
looks OK to me so just deal with the tests

PMende:
Is there anything I can do to help resolve this issue? Tests are still failing? Any thoughts on what the underlying cause is?

spulec:
Thanks for opening; closing in favor of https://github.com/spulec/moto/pull/2523

lyle-nel:
Thanks a lot. I had a hard time getting all the tests to pass.

PR #2209, opened 2019-05-20
Fix base64 deprecated methods in kinesis.utils

sergejs-katusenoks:
Using right encode and decode methods according to python version.
There was an attempt in #1854 before, but python 2 were not covered.


terrycain:
Fix what travis is complaining about and then your good to go

PR #2213, opened 2019-05-22
#2212 add support for delete on update_with_attribute_updates

### LINKED ISSUE #2212: DynamoDB2 Does not support DELETE on update_with_attribute_updates
If you use UpdateAttributes instead of UpdateExpressions, and try to DELETE from a set, it doesn't work and doesn't raise an exception. PR incoming.

ZoidBB:
Fixes #2212. Raises exceptions properly when functions are attempted that don't have support yet, to prevent confusion.


ZoidBB:
The new code seems to be passing tests fine -- broken tests are inherited from master...

terrycain:
Nice good job, and yeah the tests are because of a later version of docker lib

PR #2222, opened 2019-05-28
SecretsManager secret value binary support

p1c2u:
Adds binary values (SecretBinary parameter) support for the following operations in SecretsManager:
* [create_secret](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/secretsmanager.html#SecretsManager.Client.create_secret)
* [get_secret_value](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/secretsmanager.html#SecretsManager.Client.get_secret_value)


terrycain:
Nice job

PR #2236, opened 2019-06-11
Add glue.delete_partition and glue.batch_delete_partition

dkuntz2:
Added more glue endpoints, also updated the implementation coverage list with the covered glue endpoints.


spulec:
Looks great; thanks!

Issue #2230, opened 2019-06-03
Support VersionIdsToStages field in SecretsManager.DescribeSecret


sam-clotfelter:
describe_secret does not return VersionIdsToStages.

I'm using moto master branch on a local docker container.

klesniewski:
Not having VersionIdsToStages renders moto unusable with [aws-secretsmanager-caching-java](https://github.com/aws/aws-secretsmanager-caching-java) library (see [intro](https://aws.amazon.com/blogs/security/use-aws-secrets-manager-client-side-caching-libraries-to-improve-the-availability-and-latency-of-using-your-secrets/). The caching library seeks for secret version at specific stage (by default `"AWSCURRENT"`) and if it doesn't have this map or none of the versions are marked with the stage, then it does not resolve the secret.

Current options:
* Not use moto/localstack for integration testing
* Not use caching library

bblommers:
This has been implemented by #3029 , and should be available as of moto >= 1.3.15.dev882

sbwhitney:
> This has been implemented by #3029 , and should be available as of moto >= 1.3.15.dev882

Hi, thanks for adding the feature. Do you intend to post an updated version to pypi? 1.3.14 is available when pip installing.

PR #2237, opened 2019-06-11
In SecretsManager.DescribeSecret, include VersionIdsToStages field

sam-clotfelter:
Implements this issue.
https://github.com/spulec/moto/issues/2230

### DIFF
--- a/moto/secretsmanager/models.py
+++ b/moto/secretsmanager/models.py
@@ -74,7 +74,7 @@ class SecretsManagerBackend(BaseBackend):
             "ARN": secret_arn(self.region, secret['secret_id']),
             "Name": secret['name'],
             "VersionId": secret_version['version_id'],
-            "VersionStages": secret_version['version_stages'],
+            "VersionStages": list(secret_version['version_stages']),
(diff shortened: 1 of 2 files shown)


terrycain:
Add some tests to verify this behaviour

sam-clotfelter:
Tests have been added. Sorry about that.

I found and fixed a bug from when I implemented support for multiple version secrets. AWSCURRENT was not being set correctly when put_secret_value was called on an empty secret and a custom VersionStage was passed to it.

While fixing this, I changed the data structure. version_stages has been changed from a list to a set.

spulec:
Looks like everything is good other than the linter. If we can get that cleaned up, happy to merge!

rolfarrestam:
This PR is exactly what I'm looking for, to be able to run local tests of my secrets rotation lambdas.
When will the extra linebreak be removed, so the merge can be made?

spulec:
@rolfarrestam if you want to copy these changes into a new PR with the newline removed, I'll be happy to merge.

bblommers:
Will close this, as the functionality has now been implemented in #3029.

Issue #1834, opened 2018-09-20
dynamodb2 UpdateExpression "REMOVE" not working for nested attributes.


avinashbitra:
table.update_item(...
            UpdateExpression='REMOVE attra.#attrb',
            ExpressionAttributeNames={
                "#attrb": 'value'
            }
        )

Expected:  attra.attrb removed from my item
Result: attra.attrb not removed and no error.

CODE: dynamodb2/models.py line 147

                if action == "REMOVE":
                    self.attrs.pop(value, None) -- value = "atta.attr.b"

spulec:
Thanks; setting as a feature request.

BradyShober:
This was driving me crazy trying to figure out why it wasn't working in my test

ZoidBB:
I'll be submitting a PR for this shortly.

PR #2238, opened 2019-06-12
#1834 Fix DynamoDB UpdateExpression support for REMOVE on nested maps

ZoidBB:
This adds nested map support to the REMOVE operation on dynamodb2, fixing #1834


ZoidBB:
Yikes, I've got a bug. I'll have it fixed tonight. Don't worry about reviewing yet.

mfogel:
@ZoidBB what was the bug you ran into?

Issue #2246, opened 2019-06-17
Fix bug in ECR batch_delete_image


queue-tip:
An ECR image should be deleted when its last tag is deleted. If it's not deleted, then subsequent calls to describe_image return an invalid response.

spulec:
Closed with #2247

PR #2247, opened 2019-06-17
Delete ECR image when it has no tags

queue-tip:
This should fix the bug identified in #2246 

This is a bug fix to my initial work when implementing
batch_delete_image. Deleting the last tag for a given image should
delete the image from the backend.

I also cleaned up the tests previously-added in the initial
implementation.


spulec:
thanks!

PR #2250, opened 2019-06-18
Raising MalformedXML exception when using boto3 client and s3.delete_objects()

szczeles:
This PR fixes the S3 behavior when the `delete_objects()` (via boto3) is called without parameters. According to the AWS API usage, it should raise an exception:
```
>>> import boto3
>>> boto3.client('s3').delete_objects(Bucket='bucket-name-here', Delete={'Objects': []})
  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/site-packages/botocore/client.py", line 324, in _api_call
  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/site-packages/botocore/client.py", line 622, in _make_api_call
botocore.exceptions.ClientError: An error occurred (MalformedXML) when calling the DeleteObjects operation: The XML you provided was not well-formed or did not validate against our published schema
```

### DIFF
--- a/moto/s3/responses.py
+++ b/moto/s3/responses.py
@@ -566,6 +566,8 @@ class ResponseObject(_TemplateEnvironmentMixin):
         keys = minidom.parseString(body).getElementsByTagName('Key')
         deleted_names = []
         error_names = []
+        if len(keys) == 0:
+            raise MalformedXML()
 
         for k in keys:
             key_name = k.firstChild.nodeValue
--- a/tests/test_s3/test_s3.py
+++ b/tests/test_s3/test_s3.py
@@ -648,6 +648,7 @@ def test_delete_keys_with_invalid():
     Key(bucket=bucket, name='file3').set_contents_from_string('abc')
     Key(bucket=bucket, name='file4').set_contents_from_string('abc')
 
+    # non-existing key case
     result = bucket.delete_keys(['abc', 'file3'])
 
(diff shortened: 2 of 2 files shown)


spulec:
Thanks for opening.

It looks like the new test is failing in server mode for some reason. I'm not sure why off the top of my head. If we can get that fixed, I'll be happy to merge.

szczeles:
Hi @spulec! Yeah, I saw the CI failure, I tried to replicate this behaviour locally running the server but no luck yet. I will try this again today and let you know.

szczeles:
Hi @spulec, I tried to make the test pass but no luck yet... On local mode (when I follow the content of the travis file on my laptop) everything works OK, no matter if I run only s3 tests or all tests. 

Could you please have a look?

spulec:
Ah, I think I found the issue. It looks like we are using `boto` and the `mock_s3_deprecated` for most of the test, but then using `boto3` in the new code. `boto3` is designed to be used with `mock_s3`. We can either break this out into a separate test with `boto3` and `mock_s3`, change the addition to use `boto`, or upgrade the entire test to use `boto3`.

szczeles:
You were 100% right, Steve, thank you! I decided to split test into two, because boto and boto3 behave differently on the tested case (boto returns empty response, boto3 raises exception)

spulec:
thanks!

PR #2251, opened 2019-06-18
Fixed random_suffix()

acsbendi:
The bug: `random.choice` cannot interpret `'A-Z'`, it requires all the individual characters in the parameter list.
This should resolve the sometimes failing `test_create_target_groups_through_cloudformation` test case.


spulec:
thank you!

PR #2253, opened 2019-06-18
Implement secretsmanager.update_secret_version_stage

sam-clotfelter:
Implements update_secret_version_stage. There is no related issue. Also includes the same fix to put_secret_value from https://github.com/spulec/moto/pull/2237 .

One change to internal data structure. version_stages is changed from a list to a set.

### DIFF
--- a/IMPLEMENTATION_COVERAGE.md
+++ b/IMPLEMENTATION_COVERAGE.md
@@ -3658,7 +3658,7 @@
 - [ ] put_attributes
 - [ ] select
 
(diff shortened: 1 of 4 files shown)


spulec:
Code looks good. It looks like Travis is only complaining about the linter so happy to merge if we can get that cleaned up.

bblommers:
Going to close this, as the functionality has been implemented by #3912 and is available as of moto >= 2.0.7.
Let me know if I've missed anything.

Issue #1491, opened 2018-02-26
dynamodb scan projection expressions do not limit items returned


ssolari:
In a table, "MyTable", with multiple attributes, "attribute1", "attribute2", etc.   The following command should return a list of dictionaries having a single key in each->"attribute1".   However, each dictionary in the return result includes all attributes. 
```
response = dynamodb.scan(TableName="MyTable", ProjectionExpression="attribute1", Select='SPECIFIC_ATTRIBUTES')
```

IVIURRAY:
having the same issue here - I might take a look

chris-erickson:
Seems like this also affects `get_item`: https://github.com/spulec/moto/blob/3a5d857a60c3a2d140ed2c8adfe8dcaf71a4cac8/moto/dynamodb2/models.py#L596

Similar functionality was added to `query` https://github.com/spulec/moto/pull/1176 so perhaps this isn't a huge lift for someone?

spulec:
Thanks for opening; setting this as a feature request to implement in scans too.

bblommers:
Closing this, as query/scan/get_item now all support the projection expression

PR #2265, opened 2019-06-26
ProjectionExpression works with table.scan()

IVIURRAY:
Ability to use ProjectionExpression with dynamodb.scan().

ProjectionExpression is not implemented for table.scan() but is for table.query() - took the majority of my inspiration for this from the query function.

#1491


spulec:
Looks great; thanks!

PR #2136, opened 2019-04-01
Dynamodb expressions parser

mjstevens777:
This PR improves condition expression support for dynamodb2. It should hopefully fix #1819, #1587, and #1711.

The essential changes are:
- Update the parsing code for filter expressions to an OOP design that is hopefully more readable and reliable
- Add some test cases for condition expressions that were previously not implemented
- Support nested attributes in List and Map types.
- Switch `Expected` to use the code in `comparisons`
- Switch `ConditionExpression` to use code in `comparisons` as well

The remaining TODO items are
- [x] Finish adding test cases
- [ ] Finish documenting
- [x] Update uses of `super()` to support python 2
- [ ] DRY up the code for `DynamoType` and `AttributeValue`

PR #2266, opened 2019-06-26
Improve DynamoDB condition expression support

garrettheel:
Picking up @mjstevens777's work in https://github.com/spulec/moto/pull/2136.

This adds proper support for parsing and evaluating condition expressions.


tests/test_dynamodb2/test_dynamodb.py
```diff
@@ -1641,6 +1718,89 @@ def test_condition_expressions():
...
+    client.put_item(
+        TableName='test',
+        Item={
+            'forum_name': {'S': 'foo'},
+            'ttl': {'N': 'bar'},
+        }
+    )
+
+
+    def update_if_attr_doesnt_exist():
```
garrettheel: @killthrush I think this captures your comment from #2136 but let me know if not!

killthrush: Thanks for picking this up @garrettheel!  This fixes one issue, however in the interim I've collected a few more test cases related to conditional checks that still fail with this commit.  I'll see if I can capture them too by cloning your branch.

killthrush: Hmm, it looks like some of these failures are unrelated to moto.  I'll keep testing, but feel free to move ahead with this PR - I can always file an issue after the fact.  Thank you!

killthrush: OK, I *did* find some test failures related to `list_append` not working properly, but I suspect that is a different issue judging from the commit history.

garrettheel:
@terrycain I'd really appreciate if you could take a look and let me know what this needs to be merged. thanks!

garrettheel:
@spulec would you mind taking a look?

spulec:
Based on coveralls, it looks like this added a lot of new code that isn't covered by tests. Can you check into that?

mjstevens777:
Sorry, it looks like I left some dead code in `moto/dynamodb2/responses.py`. Lines 35-93.

garrettheel:
Removed the existing code that was made irrelevant by this change!

spulec:
Awesome; thanks for all of your work!

alexholtz:
I could be mistaken, but from what I can tell the way this was implemented it seems like the API-style conditions such as `attribute_not_exists('attr1')` is supported, but the SDK style conditions such as: `Attr('attr1').not_exists()` are not.

alexholtz:
I could be mistaken, but it seems like the way this was implemented it supports the API-style ConditionExpressions: `ConditionExpression=attribute_not_exists('attr_name')` but not the SDK-style conditions: `ConditionExpression=Attr('attr_name').not_exists()`.

PR #2268, opened 2019-06-29
IAM policy version fixes

### LINKED ISSUE #2008: IAM create_policy bug
According to the the AWS documentation, when we call the create_policy operation, it should create a a policy version with a version identifier of v1 and set v1 as the policy's default version. However, moto set the v1 policy default as False. 

Sample policy that was created:
```
{'Versions': [{'Document': {'Version': '2012-10-17', 'Statement': [{'Effect': 'Allow', 'Action': 'logs:CreateLogGroup', 'Resource': '*'}, {'Effect': 'Allow', 'Action': ['dynamodb:DeleteItem', 'dynamodb:GetItem', 'dynamodb:PutItem'], 'Resource': '*'}]}, 'VersionId': 'v1', 'IsDefaultVersion': False, 'IsTruncated': False, 'ResponseMetadata': {'RequestId': '20f7279f-99ee-11e1-a4c3-27EXAMPLE804', 'HTTPStatusCode': 200, 'HTTPHeaders': {'server': 'amazon.com'}, 'RetryAttempts': 0}}```

acsbendi:
Changes:

- `IsDefaultVersion` is always correctly returned. There is always only one default version. Fixes #2008.
- It is not possible to create more than 5 policy versions for the same policy. The returned error matches that of the AWS API.
- Added checking `IsDefaultVersion` to tests and created new tests too.


spulec:
Thanks!

PR #2269, opened 2019-07-01
IAM policy document validation

### LINKED ISSUE #1435: IAM: create_policy does not complain about invalid JSON in policy document
Versions:

```
$ pip freeze|grep oto
boto==2.48.0
boto3==1.5.14
botocore==1.8.28
moto==1.2.0
```

Note that the python code in #1434 contains invalid json (the final comma on the third line from the bottom of the JSON should not be there). Calling AWS with this json fails, but moto lets me create this policy. Should it at least try to convert the json to a python object to see if it's valid?

Thanks.

acsbendi:
I implemented the basic validation of IAM policy documents. The implementation is not complete, for example, I only validate a few conditions (those starting with `Date`) thoroughly, but all the basic cases are covered. 

I added a lot of test cases for both valid and invalid policy documents. I also verified the results (responses returned by AWS) for every test case.

Fixes #1435.

### DIFF
--- a/moto/iam/exceptions.py
+++ b/moto/iam/exceptions.py
@@ -34,6 +34,14 @@ class MalformedCertificate(RESTError):
             'MalformedCertificate', 'Certificate {cert} is malformed'.format(cert=cert))
 
 
+class MalformedPolicyDocument(RESTError):
+    code = 400
+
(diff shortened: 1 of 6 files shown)


spulec:
This is amazing; thanks!

PR #2280, opened 2019-07-04
Create access key fix

acsbendi:
Minor fixes in the CreateAccessKey operation, such as CreateDate returned in the response, and using a format for secret access key and access key ID that matches the format currently used by AWS.


spulec:
Awesome; thanks!

PR #2281, opened 2019-07-05
Describe instance attribute fix

acsbendi:
Minor fixes in the DescribeInstanceAttribute operation:

- the correct error is raised for invalid attributes
- no error for any valid attribute
- security group IDs returned correctly (instead of the object representing a SecurityGroup)
- value is not returned if it's `None`


spulec:
Looks great; thanks!

PR #2283, opened 2019-07-07
Example of using a moto server with Scala

myazinn:
This is an example of a stand-alone server configuration using Scala.
The application logic is the same as for a [Java example](https://github.com/spulec/moto/blob/master/other_langs/sqsSample.java_langs/sqsSample.java). It also uses AWS SDK for Java. However, the Java example uses deprecated API, so the implementation for Scala is slightly different.


spulec:
Thanks!

PR #2286, opened 2019-07-08
AssumeRole returns randomly generated credentials.

acsbendi:
Previously, the operation AssumeRole returned fixed hardcoded credentials. This PR fixes this by using randomly generated credentials, which are in the same format as the ones from AWS.

### DIFF
--- a/moto/sts/models.py
+++ b/moto/sts/models.py
@@ -2,6 +2,7 @@ from __future__ import unicode_literals
 import datetime
 from moto.core import BaseBackend, BaseModel
 from moto.core.utils import iso_8601_datetime_with_milliseconds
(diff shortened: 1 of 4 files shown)


spulec:
Thanks!

acsbendi:
@gabrielfalcao We use moto to mock AWS in general, as closely as possible. Sorry for causing inconvenience though. I think it's not necessary to assert the whole Credentials `dict`, maybe you could just check for the length and other properties that are invariant for real AWS responses too. I recommended a similar way of testing as in [here](https://github.com/spulec/moto/blob/master/tests/test_sts/test_sts.py).

gabrielfalcao:
@acsbendi thanks for the quick response. I'll update my tests according to the way you recommended.

PR #2294, opened 2019-07-11
Cleanup responses mocking. Closes #1567

### LINKED ISSUE #1567: Should Moto mock external requests?
While moto is active, should requests to other sites (ex `http://www.google.com`) still work?

I'm leaning toward yes. At the same time, I think it is important that any amazon service that is not implemented yet throw an error (see #1254).

I believe this means we should add a passthrough for `*` , but add a fake error responses for `*amazonaws.com*`

Thoughts?

spulec:
This unblocks requests to other websites with requests while Moto
is activated. It also adds a wildcard for AWS services to still
ensure no accidental requests are made for unmocked services

### DIFF
--- a/moto/core/models.py
+++ b/moto/core/models.py
@@ -193,53 +193,8 @@ class CallbackResponse(responses.CallbackResponse):
 
 botocore_mock = responses.RequestsMock(assert_all_requests_are_fired=False, target='botocore.vendored.requests.adapters.HTTPAdapter.send')
 responses_mock = responses._default_mock
-
-
-class ResponsesMockAWS(BaseMockAWS):
-    def reset(self):
-        botocore_mock.reset()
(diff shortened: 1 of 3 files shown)


mikegrima:
LGTM -- I'll let you merge it @spulec :)

moto/core/models.py
```diff
@@ -193,53 +193,8 @@ def _url_matches(self, url, other, match_querystring=False):
...
-            botocore_mock.stop()
-        except RuntimeError:
-            pass
-
-        try:
-            responses_mock.stop()
-        except RuntimeError:
-            pass
+# Add passthrough to allow any other requests to work
+responses_mock.add_passthru("http")
```
e271828-: Worth adding a comment that it applies to https as well, since the test is simply "startswith"

PR #2295, opened 2019-07-11
dynamodb2: Defer evaluation of the OR RHS in condition expr

garrettheel:
The test case should hopefully explain why this is an issue. I was doing some testing with more complicated condition expressions and came across this


garrettheel:
@spulec the coverage should effectively be the same, just one less line considered

spulec:
Thank you!

Issue #2183, opened 2019-05-06
creating rds with option group 


tallyl:
Seems to be ignoring when creating an rds with an option group - seems to be ignoring the option group and it isnt displayed in the response 
or when using describe 
code example : 

@mock_ec2
    def __subnet_list(self):
        ec2 = boto3.client('ec2', region_name='us-east-1')
        subnet_list = []
        for subnet in ec2.describe_subnets()["Subnets"]:
            # print(subnet["SubnetId"])
            subnet_list.append(subnet["SubnetId"])
            return subnet_list

    @mock_rds2
    def __setup(self):
        subnet_list = self.__subnet_list()
        rds = boto3.client('rds', region_name='us-east-1')
        response = rds.create_db_parameter_group(DBParameterGroupName="pg", DBParameterGroupFamily="oracle-se2-12.2", Description="Test")
        response = rds.create_option_group(OptionGroupName="og", EngineName="oracle-se", MajorEngineVersion="11.2", OptionGroupDescription="Test")
    
        response = rds.create_db_subnet_group(DBSubnetGroupName="dbsg", SubnetIds=subnet_list, DBSubnetGroupDescription="Text")
        response = rds.create_db_instance(DBName="TestDB", DBInstanceIdentifier="TestDB", DBInstanceClass="db.t2.micro", OptionGroupName="og", Engine="oracle-se2", MultiAZ=False,
                                          StorageType="gp2", AllocatedStorage=10, LicenseModel="bring-your-own-license", MasterUsername="oracle", MasterUserPassword="oracle",
                                          Port=1521, DBSubnetGroupName="dbsg")
                                          # DBParameterGroupName="tally", DBSubnetGroupName="subnet")

response : 

[...]

PR #2296, opened 2019-07-11
Add support for OptionGroupName in create_db_instance

bkovacki:
Add support for OptionGroupName parameter in create_db_instance. Resolves: #2183


spulec:
Looks great; thank you!

Issue #2249, opened 2019-06-17
RDS: VpcSecurityGroups are not set/updated


goodspark:
```python
import moto
import boto3

with moto.mock_rds2():
  rds = boto3.client("rds", region_name="us-west-2")
  rds.create_db_instance(
    DBName="mytest",
    DBInstanceIdentifier="mytest",
    DBInstanceClass="m5.large",
    AllocatedStorage=64,
    Engine="mysql",
  )

  rds.modify_db_instance(
    DBInstanceIdentifier="mytest",
    VpcSecurityGroupIds=["blah"],
    ApplyImmediately=True,
  )

  resp = rds.describe_db_instances(DBInstanceIdentifier="mytest")
  assert resp["DBInstances"][0]["VpcSecurityGroups"] == ["blah"]
```

Python 3.6.5
boto3==1.9.170
moto==1.3.8

The same happens when if I were to pass the security groups to the `create_db_instance` call.

I've even tried going as far as creating the VPC and security groups in moto (excluded because the result is the same).

spulec:
Thanks for opening. It looks like we haven't implemented this yet. I'm going to set this as a feature request.

PR #2299, opened 2019-07-13
Add support for VpcSecurityGroups set/update in RDS

bkovacki:
Add support for VpcSecurityGroups set/update in RDS. Resolves: #2249


spulec:
Thank you!

PR #2300, opened 2019-07-13
Basic IAM authentication, authorization

### LINKED ISSUE #1131: Add IAM User/key configuration
This tool is amazing, thanks for it.

Is there any way to configure users/keys so I can test authentication is denied using wrong key par?

### LINKED ISSUE #1863: Determine required IAM permissions
Howdy,

So something I've been thinking about recently is a good way to estimate the IAM permissions that a given application would need in order to function, Doing code analysis seems super difficult, and would be language-specific.

Is this something that would make sense to extend from moto_server, or roll into it?

I have been looking through the code a little bit, but I am generally not familiar with moto, it just seemed to me that there might be a way to have moto_server track which actions are taken, have an understanding of the permissions required to do that thing, and then provide a report, or even a generated policy document that would make all of those actions possible against real AWS endpoints.

Does that seem plausible? Any hints on the best place to start? Am I a crazy person?

acsbendi:
For an internal project, I implemented basic IAM authentication and authorization in moto. This includes the following:

- returning an error when the request is signed with invalid credentials (either `AccessKeyId` does not exist, or `Signature` does not match the calculated one)
- returning an error when the IAM principle making the request does not have the required permission

It is a very basic implementation, with a lot of incomplete areas, such as

- only IAM users and assumed roles are supported as principals
- currently the `Resource` and `Condition` parts of the policies are not checked at all
- the service-specific error messages are only supported for IAM, S3 and EC2
- since the `Resource` element inside IAM policies is not supported, the returned error messages does not contain this information either

To make the changes non-intrusive, I introduced an environment variable `INITIAL_NO_AUTH_ACTION_COUNT`. If this is set, moto will skip performing any authentication as many times as the variable's value, and only starts authenticating requests afterwards. If it is not set, it defaults to infinity, thus moto will never perform any authentication at all.

For now, please let me know if the concept (and the code) is okay, and if so, I'll create tests and add a section about it to the `README.md`.

Resolves #1131, resolves #1863.


spulec:
This is a great feature and some nice cleanup! Thank you.

I agree with this approach of using an environment variable for now. We can discuss later if we should default it on in a future major version. Can you move `INITIAL_NO_AUTH_ACTION_COUNT` to the settings file?

As you mention, if we can get some tests added and an update to the docs, I will be happy to merge!

acsbendi:
@spulec Everything should be ready for this to be merged now.

spulec:
Thanks again for this; it is very cool!

acsbendi:
You're welcome!

PR #2301, opened 2019-07-13
Update supported python versions, add badages.

asherf:
Python 3.4 is EOL, also add python 3.7 to tox.ini.
Screenshot 2019-07-13 13 17 40

https://github.com/asherf/moto/blob/versions/README.md


spulec:
thanks!

Issue #2302, opened 2019-07-14
API Gateway missing validator endpoints


jakzo:
A few API Gateway endpoints are not currently supported which prevents validation.

In my case this caused Pulumi Crosswalk used with LocalStack to fail because it uses the `restapi:put` endpoint to add an OpenAPI spec to the generated API. It will probably be calling the unimplemented `requestvalidator:create`, `requestvalidator:delete` and `requestvalidator:update` endpoints later.

Related to #1962 

List of OpenAPI validator implementations: https://openapi.tools/#data-validators
AWS endpoint docs: https://docs.aws.amazon.com/apigateway/api-reference/link-relation/restapi-put/

spulec:
Thanks; marking as a feature request.

rajendersaini:
@bblommers  -  https://github.com/spulec/moto/pull/4309  - please suggest how to fix one test case. I don't know why the response is not being in case of get_request_validators()

rajendersaini:
@spulec  - this can be closed? see the MR - #4309

bblommers:
Thanks for the reminder @rajendersaini 

This is now part of moto >= 2.2.8.dev22, so I'll close it @jakzo . Please let us know if we've missed anything.

PR #2303, opened 2019-07-14
Implement basic support for restapi:put

jakzo:
Issue: #2302 

Although this endpoint accepts an OpenAPI spec, this PR will not add validation to created gateway endpoints.


jakzo:
Closing for now because the request validator endpoints will also need to be implemented to solve the issue.

Issue #1786, opened 2018-08-17
filter 'description' for DescribeNetworkInterfaces has not been implemented


mcaulifn:
When running describe_network_interface and filtering on 'description', I'm getting the following error:

```
ec2_client.describe_network_interfaces(Filters=[{'Name':'description','Values':[eni_desc]}])

self = <moto.ec2.models.NetworkInterface object at 0x0000021E6CB099E8>, filter_name = 'description', method_name = 'DescribeNetworkInterfaces'

    def get_filter_value(self, filter_name, method_name=None):
        tags = self.get_tags()

        if filter_name.startswith('tag:'):
            tagname = filter_name.replace('tag:', '', 1)
            for tag in tags:
                if tag['key'] == tagname:
                    return tag['value']

            return ''
        elif filter_name == 'tag-key':
            return [tag['key'] for tag in tags]
        elif filter_name == 'tag-value':
            return [tag['value'] for tag in tags]
        else:
>           raise FilterNotImplementedError(filter_name, method_name)
moto.ec2.exceptions.FilterNotImplementedError: The filter 'description' for DescribeNetworkInterfaces has not been implemented in Moto yet.
```

moto version 1.3.4 installed via pipenv
Using python mocks

spulec:
Thanks; setting as a feature request.

PR #2306, opened 2019-07-14
Extend EC2 DescribeNetworkInterface filter support

bkovacki:
Extend EC2 DescribeNetworkInterface filter support

* add description property to EC2 NetworkInterface
* extend DescribeNetworkInterfaces filter support with description,
subnet-id, private-ip-address attributes

Resolves: #1786


spulec:
Thank you!

Issue #2310, opened 2019-07-16
cognito-identity: NotImplementedError: The get_open_id_token action has not been implemented


caguado:
## Description
The current master does not implement the [get_open_id_token](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/cognito-identity.html#CognitoIdentity.Client.get_open_id_token) endpoint of cognito-identity.

spulec:
Closed with https://github.com/spulec/moto/pull/2311

PR #2311, opened 2019-07-16
Implement get_open_id_token

caguado:
Introduce the CognitoIdentity's GetOpenIDToken endpoint to retrieve a
JWT tuple from Cognito's Identity Pool for a given IdentityId.

This implements feature #2310


caguado:
Looking at the error message I'm not sure why tests fail on `moto_server` (never used it). I'm rerunning tests with in my local env to try to reproduce.

caguado:
I found an issue on the handling of unsigned requests inside `moto/server.py`, adding an additional commit to address it.

spulec:
Looks great; thanks!

Issue #2312, opened 2019-07-16
sts: NotImplementedError: The assume_role_with_web_identity action has not been implemented


caguado:
## Description
The current master does not implement the [assume_role_with_web_identity](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/sts.html#STS.Client.assume_role_with_web_identity) endpoint of STS.

spulec:
Closed with https://github.com/spulec/moto/pull/2313

PR #2313, opened 2019-07-16
Implement assume_role_with_web_identity

caguado:
The AssumeRoleWithWebIdentity is a similar endpoint to STS's AssumeRole
where the authentication element is a JWT id_token from a configured OP.
This commit implements the functionality and relies on the same result
generated for the regular AssumeRole.

This implements feature #2312


spulec:
Thank you!

Issue #2257, opened 2019-06-21
ASG update to MinSize and MaxSize does not update ASG instances.


pasipa2:
Moto version: moto==1.3.8
Python Version: Python 3.7.3

Boto:
boto==2.49.0
boto3==1.9.169
botocore==1.12.169

Installed using Pip in virtualenv provided by Miniconda
Using Python mocks

Example 
===============
```
with moto.mock_autoscaling():
    name = 'test'
    size = 1
    region = 'eu-west-2'
    launch_config = '{}-launch-config'.format(name)
    client = boto3.client('autoscaling', region_name=region)
    ami_id = 'ami-1f12c066'
    asg_name = '{}-az1'.format(name)
    avail_zone = '{}a'.format(region) 

    client.create_launch_configuration(
        LaunchConfigurationName=launch_config,
        ImageId=ami_id,
        KeyName='pytest',
        SecurityGroups=[
            'sg-pytest',
        ],
        InstanceType='t2.medium'
    )
    client.create_auto_scaling_group(
        AutoScalingGroupName=asg_name,
        LaunchConfigurationName=launch_config,
        MaxSize=size,
        MinSize=size,
        AvailabilityZones=[
            avail_zone,
        ],
        LoadBalancerNames=[
            asg_name,
        ]
    )
    client.update_auto_scaling_group(
        AutoScalingGroupName=asg_name,
        MinSize=0,
        MaxSize=0,
    )
    response = client.describe_auto_scaling_groups(
            AutoScalingGroupNames=[
                asg_name,
            ]
        )['AutoScalingGroups'][0]

print(response)
```

Produces

```
{'AutoScalingGroupName': 'test-az1',
 'AutoScalingGroupARN': 'arn:aws:autoscaling:us-east-1:803981987763:autoScalingGroup:ca861182-c8f9-4ca7-b1eb-cd35505f5ebb:autoScalingGroupName/test-az1',
 'LaunchConfigurationName': 'test-launch-config',
 'MinSize': 0,
 'MaxSize': 0,
 'DesiredCapacity': 1,
 'DefaultCooldown': 300,
 'AvailabilityZones': ['eu-west-2a'],
 'LoadBalancerNames': ['test-az1'],
 'TargetGroupARNs': [],
 'HealthCheckType': 'EC2',
 'HealthCheckGracePeriod': 300,
 'Instances': [{'InstanceId': 'i-2d9b95392223ac103',
   'AvailabilityZone': 'us-east-1e',
[...]

PR #2315, opened 2019-07-16
Handle change of ASG desired capacity on min and max size update

bkovacki:
A change in UpdateAutoScalingGroup:
* if a value for MinSize is specified without specifying a value for
DesiredCapacity, and the new MinSize is larger than the current size of
the group, set the group's DesiredCapacity to the new MinSize value
* if a value for MaxSize is specified without specifying a value for
DesiredCapacity, and the new MaxSize is smaller than the current size of
the group, set the group's DesiredCapacity to the new MaxSize value

Resolves: #2257


spulec:
Looks great; thanks!

PR #2316, opened 2019-07-16
Fixed restriction on incoming requests

### LINKED ISSUE #1855: Add trusted hosts / networks to Server mode
Currently (Java) AWS API can only work with Moto from the local network (if it starts with "192.168.X.X") or if the request is coming from the same host as the server, that's very limiting because i want to run Moto on a container in an external network.

Any chance to add trusted networks or hosts via config file or CLI args?

I'd appreciate any info if this had been addressed or i'm using it wrong

trickstyler:
Suggested in order to fix #1855


spulec:
Closing in favor of https://github.com/spulec/moto/pull/2330

PR #2319, opened 2019-07-17
Implemented creating Auto Scaling group from instance

acsbendi:
Implemented creating Auto Scaling group from instance and added tests for it.

Without this PR, when `CreateAutoScalingGroup` was called without a `LaunchConfigurationName` parameter, and the `InstanceId` was specified instead, moto returned an internal server error (500).


spulec:
thank you!

PR #2320, opened 2019-07-17
Add glue.batch_get_partition endpoint

dkuntz2:
Unexpectedly, the actual response from AWS when attempting to `batch_get_partition` is to just ignore missing partitions in the call, and not error or place them in another key.


spulec:
thank you!

PR #2321, opened 2019-07-17
added support to UPSERT records that are not simple routing policy such as weighted records

sblumin:
Implemented UPSERT for `change_resource_records()` in Route 53 if the record is not a simple routing policy i.e. weighted or latency based.
Without this PR, moto only compares the record name and type and will always change the first record in the list. Now, upsert also looks at the `set_identifier` to change the correct record.


spulec:
thank you!

PR #2324, opened 2019-07-18
ELBv2 DescribeTargetHealth returns correct response for stopped instance

acsbendi:
When an EC2 instance is registered as a target in an ELBv2 target group, and it is stopped, the following should be its health status:
```json
{
     "State": "unused",
     "Reason": "Target.InvalidState",
     "Description": "Target is in the stopped state"
}
```


spulec:
thank you!

PR #2325, opened 2019-07-19
Auto Scaling groups recreate terminated instances

acsbendi:
When an EC2 instance that is attached to an Auto Scaling group is terminated through the `TerminateInstances` API operation, the Auto Scaling group should automatically replace this terminated instance.


spulec:
Thanks!

PR #2307, opened 2019-07-15
moto: s3: support partNumber for head_object

efiop:
To support it, we need to keep multipart info in the key itself when
completing multipart upload.

Fixes #2154

Signed-off-by: Ruslan Kuprieiev <ruslan@iterative.ai>

PR #2327, opened 2019-07-20
tests: multipart: use REDUCED_PART_SIZE and variable part size

efiop:
There is no reason to use 5M chunks, especially with the reduced_part_size
decorator. Also made part_size to be variable to add extra layer of
testing to make sure that moto is handling that scenario correctly.

Related to #2307

Signed-off-by: Ruslan Kuprieiev <ruslan@iterative.ai>


spulec:
thank you!

Issue #2328, opened 2019-07-21
@mock_dynamodb2 update_item with AttributeUpdates doesn't work for list items


swen128:
Similar to https://github.com/spulec/moto/issues/1823

When I try to update a mocked DynamoDB item using the `update_item` method with the `Attribute Updates` argument, it fails with `TypeError: unhashable type: 'dict'`.

## Minimal code to reproduce the problem

```python
import boto3
from moto import mock_dynamodb2


@mock_dynamodb2
def test_update_item_with_list():
    dynamodb = boto3.resource('dynamodb', region_name='us-east-1')

    dynamodb.create_table(
        TableName='Table',
        KeySchema=[
            {
                'AttributeName': 'key',
                'KeyType': 'HASH'
            }
        ],
        AttributeDefinitions=[
            {
                'AttributeName': 'key',
                'AttributeType': 'S'
            },
        ],
        ProvisionedThroughput={
            'ReadCapacityUnits': 1,
            'WriteCapacityUnits': 1
        }
    )
    table = dynamodb.Table('Table')
    table.update_item(
        Key={'key': 'the-key'},
        AttributeUpdates={
            'list': {'Value': [1, 2], 'Action': 'PUT'}
        }
    )


if __name__ == '__main__':
    test_update_item_with_list()
```

## Traceback

```
Traceback (most recent call last):
  File "tests/test_moto.py", line 41, in <module>
    test_update_item_with_list()
  File "{my_project}/venv/lib/python3.7/site-packages/moto/core/models.py", line 80, in wrapper
    result = func(*args, **kwargs)
  File "tests/test_moto.py", line 37, in test_update_item_with_list
    return table.get_item(Key={'key': 'the-key'})
  File "{my_project}/venv/lib/python3.7/site-packages/boto3/resources/factory.py", line 520, in do_action
    response = action(self, *args, **kwargs)
  File "{my_project}/venv/lib/python3.7/site-packages/boto3/resources/action.py", line 83, in __call__
    response = getattr(parent.meta.client, operation_name)(**params)
  File "{my_project}/venv/lib/python3.7/site-packages/botocore/client.py", line 357, in _api_call
    return self._make_api_call(operation_name, kwargs)
  File "{my_project}/venv/lib/python3.7/site-packages/botocore/client.py", line 655, in _make_api_call
    model=operation_model, context=request_context
  File "{my_project}/venv/lib/python3.7/site-packages/botocore/hooks.py", line 356, in emit
[...]

PR #2329, opened 2019-07-21
Fix DynamoDB update_item to deal with list-type attributes

swen128:
Fix issue https://github.com/spulec/moto/issues/2328


spulec:
Looks great; thanks!

PR #2334, opened 2019-07-23
Use the same default Owner ID between EC2 models

dkuntz2:
When attempting to use Moto to do something similar to this:

``` python
@mock_ec2
def test_copy_image_example():
  conn = boto3.client("ec2")
  copy_resp = conn.copy_image(SourceImageId="one of the default ami ids from resources")
  describe_resp = conn.describe_images(Owners=["self"])
  describe_resp["Images"].should.have.length_of(1)
```

The test ended up failing, and `describe_images` responded with an empty list.

To me, the expected result would be that any image created with `copy_image` would be given the same OwnerId as images found when using `Owners=["self"]`.

The first commit does just that, by creating a common `OWNER_ID` used by all the ec2 models.

The second commit adds a test case for the above example code to ensure future changes don't accidentally result in `copy_image` using a different owner id than the one used for `self`.

There may be additional test cases that could be useful to ensure the owner ids are the same between calls using the same resource types, and I would be more than glad to write those tests, but I'm not sure if they'd be helpful, or even what would need to be tested.

### DIFF
--- a/moto/ec2/models.py
+++ b/moto/ec2/models.py
@@ -142,6 +142,8 @@ AMIS = json.load(
          __name__, 'resources/amis.json'), 'r')
 )
 
+OWNER_ID = "111122223333"
+
 
 def utc_date_and_time():
     return datetime.utcnow().strftime('%Y-%m-%dT%H:%M:%S.000Z')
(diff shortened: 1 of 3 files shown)


spulec:
Looks great; thanks!

PR #2337, opened 2019-07-26
Use a dict instead of a list for storing task definition revisions

mvantellingen:
Before when a task definition revision was de-registered all revisions after that changed their revision id. This doesn't match the way it is handled in AWS. Using a dict and manually increment the revision id solves that.


spulec:
Looks great; thank you!

PR #2345, opened 2019-07-30
Fix timezone problem in test create cluster boto3

### LINKED ISSUE #2344: test_create_cluster_boto3 fails with timezones east of 0000
I get the following failure when running moto tests using `make test` on latest master:
```
F
======================================================================
FAIL: test_redshift.test_create_cluster_boto3
----------------------------------------------------------------------
  File "/home/iiro/.local/share/virtualenvs/moto-5xIDD5E0/lib/python3.6/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/home/iiro/Private/dev/moto/moto/core/models.py", line 84, in wrapper
  File "/home/iiro/Private/dev/moto/tests/test_redshift/test_redshift.py", line 38, in test_create_cluster_boto3
    create_time.should.be.lower_than(datetime.datetime.now(create_time.tzinfo))
  File "/home/iiro/.local/share/virtualenvs/moto-5xIDD5E0/lib/python3.6/site-packages/sure/__init__.py", line 387, in wrapper
    raise AssertionError(e)
AssertionError: expected `2019-07-30 13:54:17.705000+00:00` to be lower than `2019-07-30 10:54:17.707495+00:00`
```
It seems to be due to timezones. `Cluster.__init__` takes current time in the local timezone and adds `Z` to the end to indicate UTC. In the test the current time is taken using the timezone from `ClusterCreateTime`.

The error can be recreated by changing the local timezone to UTC+N. The test passes with timeszones UTC and UTC-N.

iisulop:
Fixes #2344.

### DIFF
--- a/moto/redshift/models.py
+++ b/moto/redshift/models.py
@@ -78,7 +78,7 @@ class Cluster(TaggableResourceMixin, BaseModel):
         super(Cluster, self).__init__(region_name, tags)
         self.redshift_backend = redshift_backend
         self.cluster_identifier = cluster_identifier
-        self.create_time = iso_8601_datetime_with_milliseconds(datetime.datetime.now())
+        self.create_time = iso_8601_datetime_with_milliseconds(datetime.datetime.utcnow())
         self.status = 'available'
         self.node_type = node_type
         self.master_username = master_username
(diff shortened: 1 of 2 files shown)


spulec:
Thank you!

Issue #1728, opened 2018-07-23
Dependencies overhead


slafs:
By default moto installs a lot of dependencies: https://github.com/spulec/moto/blob/ba1ceee95f289334b68330555598e80e581622ef/setup.py#L8-L28
Currently the dependency graph (created via `pipenv graph`) for moto has 84(!) positions with 43(!) unique packages. 

Those can introduce a substantial overhead when moto is used only for some AWS services. E.g. we use moto only to mock S3 service where e.g. `aws-xray-sdk` or `docker` aren't used, but we still need to install them.

My idea of cleaning this up would be to keep only core dependencies inside the `install_requires` list and extract the rest of them and organize inside the [`extras_require` mapping](https://setuptools.readthedocs.io/en/latest/setuptools.html#declaring-extras-optional-features-with-their-own-dependencies). Then users could install moto only for the services they need e.g. with `pip install moto[s3]` or `pip install moto[xray]`.

Would you be willing to accept a PR with such solution?

spulec:
This seems interesting.

I'm not very familiar with `extras_require`. If someone wants to install everything, is there a way to do that?

slafs:
Well, you could have a special entry e.g. `all` and then you could `pip install moto[all]`. I think the more important question is: Why would want to have that, beside being backward compatible? Is there someone that would actually want to mock out _all_ of the services that moto provides in one project?

spulec:
Got it.

As an example, at my work we use dozens of services in our Moto tests so would probably prefer to include everything than listing out a dozen+ specific dependencies.

mredaelli:
This would come in very handy for me too. We have another package that depends still on `docker-py`, so installing `moto` breaks everything due to conflict.

So, would a PR be useful?

ameier38:
This would also be useful to me as I mostly use for mocking out S3. Airflow has an [example of `extras_require`](https://github.com/apache/airflow/blob/master/setup.py#L378)

spulec:
Yes, a PR would be welcome.

slafs:
I took a stab on resolving this. I've created a Draft PR - see https://github.com/spulec/moto/pull/2346. Comments are welcome.

PR #2346, opened 2019-07-30
Fix dependencies overhead

slafs:
This is the initial approach of fixing #1728 i.e. resolve the dependencies overhead and allow for installing minimal set of dependencies for a given service (e.g. when using only `s3` backend, install dependencies used by `moto.s3` module only).

Currently this is backwards incompatible (!!!)
and probably fails on some tests.

Before I start any efforts on backwards compat, I'd like to get some guidance on how to progress with this, first.

In order to reduce required dependencies we need to change the way we're accessing mock decorators (in `moto/__init__.py`) and the way we import backends (in `moto/backends.py`).
This might need an API change, so instead of doing `from moto import mock_s3` we'd need to go for `from moto.s3 import mock_s3`.

BTW This PR reduces the time needed for `import moto` from 3.5 seconds to 0.5s (on my machine).

Things to discuss/decide:
- [ ] is this approach even feasible for @spulec 
- [ ] should we use the new `_backends` variable or how we should avoid using it?
- [ ] do we still need `botocore.awsrequest` and `ConnectionCls` fix in `moto/__init__.py` (why do we need it anyway)?
- [ ] do we want to include ALL services in `extras_require` (in `setup.py`)?
- [ ] can we trim down more dependencies (i.e. remove any from `install_requires` still, e.g. werkzeug or Jinja2)?


edisongustavo:
Hello @slafs , if there is something I can do to help get this PR merged, I'm happy to help.

edisongustavo:
> This might need an API change, so instead of doing from moto import mock_s3 we'd need to go for from moto.s3 import mock_s3.

What about a potential workaround to this be, on `moto/__init__.py`, to avoid breaking the API:

```python
def mock_s3():
    from . import s3
    return s3.mock_s3()
```

slafs:
> Hello @slafs , if there is something I can do to help get this PR merged, I'm happy to help.

Hey @edisongustavo. Thanks for the offer. At this point I think I'd like some input from someone responsible for this codebase, so we get at least an initial acceptance for this work. I guess I'd like some help on keeping this PR fresh (i.e. rebasing on top of `spulec:master`).

> What about a potential workaround to this be, on moto/__init__.py, to avoid breaking the API:

Good idea! I'll try to incorporate this here. Then it should be easier to keep the PR up-to-date as well.

slafs:
Closing in favor of #2697 and #2676

PR #2352, opened 2019-08-04
Convert fields createdAt and updatedAt of ECS deployment to timestamp

wndhydrnt:
The AWS SDK for Go fails to parse the response of the `ecs.DescribeServices()` function because it expects the values of the fields [`createdAt`][1] and [`updatedAt`][2] to be timestamps, not ISO 8601.

This change converts the value of each field to a timestamp instead of ISO.

[1]: https://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_Deployment.html#ECS-Type-Deployment-createdAt
[2]: https://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_Deployment.html#ECS-Type-Deployment-updatedAt

### DIFF
--- a/moto/ecs/models.py
+++ b/moto/ecs/models.py
@@ -8,6 +8,7 @@ import boto3
 import pytz
 from moto.core.exceptions import JsonRESTError
 from moto.core import BaseBackend, BaseModel
+from moto.core.utils import unix_time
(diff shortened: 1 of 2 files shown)


spulec:
thanks!

PR #2354, opened 2019-08-06
Resolve invalid escape sequence

aksagrimada:
When run not as a decorator dynamodb2 displays an invalid escape sequence error in 3.6/7

/moto/dynamodb2/responses.py:603: DeprecationWarning: invalid escape sequence \s
  '\s*([=\+-])\s*', '\\1', update_expression)


mikegrima:
LGTM

Issue #1976, opened 2018-11-28
search job definition by name only in batch submit job method 


dedominicisfa:
In the real aws, you can call the boto3 batch.submit_job method with a job definition name, without including the entire arn, nor the revision. If you include only the name, the latest revision is assumed. IN moto, in the BatchBackend class, the job definition name is searched by arn, or by name and revision. It should also be searched just by name, as the real aws does.

spulec:
Thanks; setting this as a feature request.

PR #2359, opened 2019-08-06
Add option to call batch submit_job with job definition name only

bkovacki:
* Add option to call batch submit_job with job definition name only
* Fix bug which causes register_job_definition not to increment job
revision number after a second revision

Resolves: #1976


mikegrima:
LGTM

PR #2360, opened 2019-08-06
[Resolves #2355] - create_organization(): add master account, default…

### LINKED ISSUE #2355: Organizations: create_organization fails to add master account and default policy to mock organization
the mock organization does not contain the master account or the default policy.  After running
create_organization, the list_accounts and list_policies methods each return an empty list.

ashleygould:
… policy

Model: OrganizationsBackend
Method: create_organization

create_organization now creates master account, root ou, and a
default service control policy objects and adds them to the
OrganizationsBackend object.  the policy is attached to both
the master account and the root ou.  any subsiquently created
accounts or OU also have the default policy attached.

### DIFF
--- a/moto/organizations/models.py
+++ b/moto/organizations/models.py
@@ -2,6 +2,7 @@ from __future__ import unicode_literals
 
 import datetime
 import re
+import json
 
 from moto.core import BaseBackend, BaseModel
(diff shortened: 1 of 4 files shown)


mikegrima:
LGTM

Issue #2259, opened 2019-06-25
Implement pagination for Cloudwatch logs


StevenReitsma:
moto == 1.3.8
boto3 == 1.9.175
botocore == 1.12.175

I'm using `mock_logs` to test my CloudWatch logging. When using `get_log_events` (https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/logs.html#CloudWatchLogs.Client.get_log_events) I expect the response to contain three keys: `events`, `nextBackwardToken` and `nextForwardToken`. If I'm *not* mocking, I indeed get these three keys in the response of `conn.get_log_events`. However, when mocking, I only get the `events` key in my response dictionary. The `nextForwardToken` and `nextBackwardToken` are omitted.

I modified the `test_put_logs` test to show the problem:

```python
@mock_logs
def test_put_logs():
    conn = boto3.client('logs', 'us-west-2')
    log_group_name = 'dummy'
    log_stream_name = 'stream'
    conn.create_log_group(logGroupName=log_group_name)
    conn.create_log_stream(
        logGroupName=log_group_name,
        logStreamName=log_stream_name
    )
    messages = [
        {'timestamp': 0, 'message': 'hello'},
        {'timestamp': 0, 'message': 'world'}
    ]
    putRes = conn.put_log_events(
        logGroupName=log_group_name,
        logStreamName=log_stream_name,
        logEvents=messages
    )
    res = conn.get_log_events(
        logGroupName=log_group_name,
        logStreamName=log_stream_name
    )
    nextSequenceToken = putRes['nextSequenceToken']
    assert isinstance(nextSequenceToken, six.string_types) == True
    assert len(nextSequenceToken) == 56

    assert 'events' in res
    assert 'nextForwardToken' in res
    assert 'nextBackwardToken' in res

    events = res['events']
    events.should.have.length_of(2)
```

This test fails on `assert 'nextForwardToken' in res`

spulec:
Thanks; setting this issue as a feature request to implement proper pagination in cloudwatch logs.

PR #2361, opened 2019-08-07
Implement pagination support for GetLogEvents

bkovacki:
* Add nextForwardToken and nextBackwardToken to every GetLogEvents response
* Handle end of stream by returning the same token as passed in

Resolves: #2259


mikegrima:
LGTM

PR #2369, opened 2019-08-14
Add basic endpoints for EC2 Launch Templates

dkuntz2:
Specifically, this adds the CreateLaunchTemplate, CreateLaunchTemplateVersion, DescribeLaunchTemplates, and DescribeLaunchTemplateVersions endpoints.

I'm not super sure that I matched the code conventions, and if I didn't please let me know and I'll fix them. This is the first time I've tried to add an entirely new resource type, and I think I eventually got there, but it still feels rough.

One other note is that the travis tests for my fork failed on the python2.7 `SERVER_MODE=true` set, and I'm not sure why, though it does appear to have succeeded in not server mode python2.7, and server mode for python3. Not sure if it'll continue with this PR or if it's an intermittent thing, and if it does continue I have no idea how to fix it. [Link to failing tests](https://travis-ci.org/dkuntz2/moto/jobs/572040376).

### DIFF
--- a/moto/ec2/responses/launch_templates.py
+++ b/moto/ec2/responses/launch_templates.py
@@ -1,4 +1,3 @@
-import json
 import six
 import uuid
 from moto.core.responses import BaseResponse
@@ -18,6 +17,7 @@ def xml_root(name):
     ElementTree.SubElement(root, "requestId").text = request_id
 
     return root
+
 
 def xml_serialize(tree, key, value):
     if key:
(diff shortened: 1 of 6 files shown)


dkuntz2:
Ah I apparently just failed to notice that it failed in general for python2.7, not only server mode. Looking into fixing.

moto/ec2/responses/launch_templates.py
```diff
@@ -0,0 +1,247 @@
...
+from moto.core.responses import BaseResponse
+from moto.ec2.models import OWNER_ID
+from moto.ec2.exceptions import FilterNotImplementedError
+from moto.ec2.utils import filters_from_querystring
+
+from xml.etree import ElementTree
+from xml.dom import minidom
+
+
+def xml_root(name):
```
mikegrima: Curious why the XML parsing functions are needed. If I recall, most of the others don't have these.

dkuntz2: The other EC2 mocks that I looked at all use string templates which are just xml blobs waiting to be filled in. I thought generating the xml rather than filling in string templates was safer and, when it came to more complex objects, easier to read.

mikegrima: Sounds good.

Only thing I would ask is to test for the lack of coverage here: https://coveralls.io/builds/25143402/source?filename=moto%2Fec2%2Fresponses%2Flaunch_templates.py#L27

dkuntz2: 👍

mikegrima:
LGTM

PR #2375, opened 2019-08-20
Fix KeyCount in s3.list_objects_v2

vuryleo:
Base on AWS [document](https://docs.aws.amazon.com/AmazonS3/latest/API/v2-RESTBucketGET.html#v2-RESTBucketGET-responses) and real behavior, the `KeyCount` field in `list_objects_v2` should be sum of Keys and CommonPrefixes, though it's called "Key"Count.

### DIFF
--- a/moto/s3/responses.py
+++ b/moto/s3/responses.py
@@ -463,10 +463,13 @@ class ResponseObject(_TemplateEnvironmentMixin, ActionAuthenticatorMixin):
         else:
             result_folders, is_truncated, next_continuation_token = self._truncate_result(result_folders, max_keys)
 
+        key_count = len(result_keys) + len(result_folders)
+        
         return template.render(
             bucket=bucket,
             prefix=prefix or '',
             delimiter=delimiter,
+            key_count=key_count,
             result_keys=result_keys,
             result_folders=result_folders,
             fetch_owner=fetch_owner,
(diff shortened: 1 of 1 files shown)


vuryleo:
Yes, indeed, the document 'claims' it should be count of 'keys'. However, it can be figured out via [aws document](https://docs.aws.amazon.com/AmazonS3/latest/API/v2-RESTBucketGET.html). In the examples like
```xml
<ListBucketResult xmlns="http://s3.amazonaws.com/doc/2006-03-01/">
  <Name>example-bucket</Name>
  <Prefix>photos/2006/</Prefix>
  <KeyCount>3</KeyCount>
  <MaxKeys>1000</MaxKeys>
  <Delimiter>/</Delimiter>
  <IsTruncated>false</IsTruncated>
  <Contents>
    <Key>photos/2006/</Key>
    <LastModified>2016-04-30T23:51:29.000Z</LastModified>
    <ETag>&quot;d41d8cd98f00b204e9800998ecf8427e&quot;</ETag>
    <Size>0</Size>
    <StorageClass>STANDARD</StorageClass>
  </Contents>

  <CommonPrefixes>
    <Prefix>photos/2006/February/</Prefix>
  </CommonPrefixes>
  <CommonPrefixes>
    <Prefix>photos/2006/January/</Prefix>
  </CommonPrefixes>
</ListBucketResult>
```
it's very clear that `KeyCount` sums length of `Contents` and `CommonPrefixes`. This behavior can be verified via
```python
from boto3 import client
c = client(service_name='s3')
resp = c.list_objects_v2(Bucket='oss-playground')  # It's a bucket I created for this issue.
print(resp)
```
as well.

Besides, the `MaxKeys` limits the sum of CommonPrefixes and Contents other than each. If this change got accepted I will propose that one.

mikegrima:
Sounds good -- and thank you for confirming!

PR #2376, opened 2019-08-21
And event source mapping endpoints and SQS trigger support

rwestergren:


moto/awslambda/models.py
```diff
@@ -429,24 +431,39 @@ def _create_zipfile_from_plaintext_code(code):
...
+        self.function_arn = spec['FunctionArn']
         self.event_source_arn = spec['EventSourceArn']
-        self.starting_position = spec['StartingPosition']
-
+        self.uuid = str(uuid.uuid4())
+        self.last_modified = time.mktime(datetime.datetime.utcnow().timetuple())
         # optional
-        self.batch_size = spec.get('BatchSize', 100)
+        self.starting_position = spec.get('StartingPosition', 'TRIM_HORIZON')
+        self.batch_size = spec.get('BatchSize', 10)  # TODO: Add source type-specific defaults
```
mikegrima: BatchSize is currently set to the following https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/lambda.html:

> Amazon Kinesis - Default 100. Max 10,000.
Amazon DynamoDB Streams - Default 100. Max 1,000.
Amazon Simple Queue Service - Default 10. Max 10.

Otherwise, this field should be blank (add a check if it's Kinesis, DynamoDB, or SQS)

rwestergren:
Made those changes

moto/awslambda/models.py
```diff
@@ -429,24 +431,57 @@ def _create_zipfile_from_plaintext_code(code):
...
-        self.batch_size = spec.get('BatchSize', 100)
+        self.starting_position = spec.get('StartingPosition', 'TRIM_HORIZON')
         self.enabled = spec.get('Enabled', True)
         self.starting_position_timestamp = spec.get('StartingPositionTimestamp',
                                                     None)
 
+    def get_configuration(self):
+        return {
+            'UUID': self.uuid,
+            'BatchSize': self.batch_size,
```
mikegrima: Can you confirm if `BatchSize` is returned for non SQS/Kinesis/DynamoDB events?

rwestergren: Do you have an example event type you're talking about? Per docs, Kinesis/SQS/DynamoDB are the only valid event source types for Lambda: https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/lambda.html#Lambda.Client.create_event_source_mapping

mikegrima: Oh sorry -- you are right! I was getting confused by how the AWS console presents this (since you can have S3, SNS, Cloudwatch Events, etc.) invoking lambdas.

In that case, should we just assume that if it's not DynamoDB/SQS/Kinesis that it's invalid?

mikegrima:
@rwestergren Is this good to merge?

rwestergren:
I think we're good to go - thanks!

PR #2377, opened 2019-08-21
GetCallerIdentity returns real data based on the access key used

acsbendi:
Previously, GetCallerIdentity always returned the same hardcoded values.
With this PR, moto will try to find the IAM principal whose access key was used to sign the request and return its data. The currently supported principals are IAM users and assumed roles.
Another minor change included in this PR is that GetCallerIdentity is always allowed by moto, just like by AWS.


acsbendi:
There's a failing test where `@freeze_time` does not work for some reason. It only happens in server mode and when used along with `boto3` (rather than `boto`). 
Should I just delete the check so the tests will pass?

mikegrima:
I think @spulec has some context on the Freeze time issue.

spulec:
Two options come to mind. The first is to not run that test in server mode like [this](https://github.com/spulec/moto/blob/master/tests/test_awslambda/test_lambda.py#L132-L172) but the opposite. Obviously, that isn't ideal.

The other option would be to add a flag to the moto server `--freeze-time 2019-01-01` or something like that.

If you want to go with the first option for now until we have a better solution, I would be okay with that.

acsbendi:
@spulec Thanks, I think I'll just skip it for server mode then.

mikegrima:
@acsbendi is this now ready to merge?

acsbendi:
@mikegrima Yes, thank you

Issue #2366, opened 2019-08-13
Secretsmanager put_secret_value still does not support SecretBinary


kevinmbt:
The functions for put_secret_value don't have fields for secret_binary in them. 

Offending methods: https://github.com/spulec/moto/blob/master/moto/secretsmanager/models.py#L157, https://github.com/spulec/moto/blob/master/moto/secretsmanager/responses.py#L72

spulec:
Thanks; setting as a feature request.

spulec:
Closed with #2385

PR #2385, opened 2019-08-23
Secretsmanager put_secret_value still does not support SecretBinary

bblommers:
Fix for https://github.com/spulec/moto/issues/2366


spulec:
Thank you!

PR #2387, opened 2019-08-23
issues-2386 make comparing exists and new queues only by static attrs

oman36:
I got list of static attributes from list of **Attributes** in docs here https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/sqs.html#SQS.Client.create_queue


spulec:
Thank you!

PR #2391, opened 2019-08-25
Add exact Number, exact String.Array and attribute key matching to SN…

gruebel:
…S subscription filter policy and validate filter policy

If this PR is good to go, I would also like to add in a follow up PR anything-but, prefix and numeric matching.


spulec:
Looks great; thank you!

PR #2393, opened 2019-08-26
fix #2392 Add validation for device shadow version

### LINKED ISSUE #2392: iot_data.update_thing_shadow should raise ConflictException, when specified shadow version  is not same as backend
## summary
Tthis document says,
>When you supply a version with an update request, the service rejects the request with an HTTP 409 conflict response code if the current version of the state document does not match the version supplied.

https://docs.aws.amazon.com/iot/latest/developerguide/using-device-shadows.html 	

But, moto will accept update request ,that has different shadow version

## moto and boto version
boto               2.49.0
boto3              1.9.215
moto               1.3.10

cm-iwata:
This PR add validation for iot-data update_thing_shadow(),fixing #2392

### DIFF
--- a/moto/iotdata/exceptions.py
+++ b/moto/iotdata/exceptions.py
@@ -21,3 +21,11 @@ class InvalidRequestException(IoTDataPlaneClientError):
         super(InvalidRequestException, self).__init__(
             "InvalidRequestException", message
         )
+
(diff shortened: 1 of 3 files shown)


spulec:
Thank you!

PR #2394, opened 2019-08-26
Change creationTime and lastIngestionTime type in LogStream and LogGroup to Long

dezkoat:
As documented here:
https://docs.aws.amazon.com/AmazonCloudWatchLogs/latest/APIReference/API_LogGroup.html
https://docs.aws.amazon.com/AmazonCloudWatchLogs/latest/APIReference/API_LogStream.html

The type for `creationTime` and `lastIngestionTime` should be Long.


spulec:
Thanks!

Issue #1220, opened 2017-09-29
Enable linter to catch PEP issues


thehesiod:
There are a lot of annoying things from PEP, but some good ones like standardize spacing and naming conventions.  Perhaps can enable a linter with a subset of PEP?  I know some projects use flake. This will help avoid people having to catch these issues by hand, which is very error prone.

terrycain:
Travis does run make lint at the start of all test runs. flake8 is catching some stuff.

thehesiod:
We should enable spacing and naming conventions I'm guessing to reduce the work for the reviewers as it seems those are things you want

spulec:
I have a PR to add Black to this repo. Feedback welcome

https://github.com/spulec/moto/pull/2290

thehesiod:
cool, that is quite the amount of string quoting normalization :) I'm not sure the value that particular normalization for the first iteration given it's not meant to catch bugs and adds a ton of noise to the first commit (and marginal value IMHO).  I think for some of the checks you need to install the specific flake8 plugin.  Here are some that I was thinking of for one of the repos I'm working on:

Perhaps it's worth an iterative approach?

```
flake8-return
flake8-comprehensions
flake8-builtins
flake8-eradicate
flake8-mutable
flake8-SQL ?
flake8-deprecated ?
flake8-type-annotations
flake8-string-format
flake8-mypy
pep8-naming
```

spulec:
Yeah, I don't have particular thoughts on string quoting. Happy to rerun black with that fix disabled.

Any of those seem reasonable to me to add.

bblommers:
I think we can close this, now that _black_ is enforced

PR #2397, opened 2019-08-28
Making kms:Encrypt/Decrypt/GenerateDataKey more real

mattsb42-aws:
## Background

There are a variety of properties that the `kms:Encrypt`, `kms:Decrypt`, and `kms:GenerateDataKey/WithoutPlaintext` methods in `moto` have been lacking in order to properly model the properties of the KMS API. Some of the most important of these are honoring encryption context, only allowing decryption by the master key that encrypted a ciphertextblob (and finding that master key based on a blob), and identifying the decrypting master key on decrypt. A less important, but valuable, property is that the ciphertextblob should be reasonably small and opaque.

I was discussing these issues with a colleague a while ago, and it occurred to us that the simplest way to achieve all of these would be to just actually do the encryption. So I did.

## Words of Warning

I do want to note a few very important things here for the record:
1. The ciphertextblobs that this creates are not compatible with the actual KMS APIs.
1. The format has nothing to do with the actual KMS ciphertextblob format and is just the first format that came to mind. I would recommend it be treated the same way as the actual KMS ciphertextblob format: as a black box that is understood by the thing that wrote it but that could change at any time.
1. I am making an assumption that the ciphertextblobs created by these methods will never live longer than the lifetime of a test execution. If that is an invalid assumption, then a slightly more complex, and versioned, format should be defined.

## Details

Ok, warning labels out of the way, what's actually going on here?

* When a `Key` (CMK) is created in a `KmsBackend`, it now gets a key material.
* When `kms:Encrypt` is called, the plaintext is actually encrypted using the CMK's key material, using AES-GCM and applying the serialized encryption context as AAD.
    * The returned ciphertextblob contains the key ID, IV, GCM tag, and ciphertext.
* When `kms:Decrypt` is called with a valid ciphertextblob, the key ID is extracted from the blob and used to locate the correct CMK.
    * Just like with the actual KMS APIs, you MUST provide the same encryption context pairs as you provided on encrypt or decryption will fail.
* `kms:GenerateDataKey` behaves as you would expect, composing with `Encrypt`.

Also, new methods implemented are:
* `kms:ReEncrypt`
* `kms:GenerateRandom`

Aside from authorization, these should all now have as close as we can get to the actual properties of the relevant KMS APIs.

## Future thoughts

As food for thought, these changes should now make it fairly straightforward to implement BYOK/key-imports.

### black

It looks like I must have accidentally run `black` over the `test_kms/test_kms.py` file. I would go back and revert the changes it made, but I see in #1220 and related that you're working on migrating to `black` anyway. Let me know if the changes it made are an issue and I'll do the git surgery necessary to roll them back.

### key ID validation

I have another set of changes that centralize and slightly revamp how key ID validation is done across all of the KMS response object methods. Given the scope of changes[1], I figure those are probably better to leave them for a separate PR once this one is processed.

[1] https://github.com/mattsb42-aws/moto/compare/kms...mattsb42-aws:key_id_validation

### DIFF
--- a/IMPLEMENTATION_COVERAGE.md
+++ b/IMPLEMENTATION_COVERAGE.md
@@ -3801,14 +3801,14 @@
 - [ ] update_stream
 
 ## kms
-41% implemented
+54% implemented
 - [X] cancel_key_deletion
 - [ ] connect_custom_key_store
 - [ ] create_alias
 - [ ] create_custom_key_store
 - [ ] create_grant
 - [X] create_key
-- [ ] decrypt
+- [X] decrypt
 - [X] delete_alias
 - [ ] delete_custom_key_store
 - [ ] delete_imported_key_material
@@ -3819,10 +3819,10 @@
 - [ ] disconnect_custom_key_store
 - [X] enable_key
 - [X] enable_key_rotation
-- [ ] encrypt
+- [X] encrypt
 - [X] generate_data_key
-- [ ] generate_data_key_without_plaintext
-- [ ] generate_random
+- [X] generate_data_key_without_plaintext
+- [X] generate_random
 - [X] get_key_policy
 - [X] get_key_rotation_status
 - [ ] get_parameters_for_import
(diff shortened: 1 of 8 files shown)


spulec:
This is great! Thank you very much for the contribution.

PR #2398, opened 2019-08-28
Allows leading // for mocked s3 paths (fix for #1637)

### LINKED ISSUE #1637: mock_s3 gives "ConnectionError: Connection refused..." error when Key used in put_object call contains leading forward slash
This is very specific situation, however the behavior of moto in this situation doesn't match what happens when you use the actual boto3 connecting to AWS.

The situation is if you have a Key for a S3 object that you are putting into a S3 bucket that has a leading forward slash ("/").

So if your Key is "/abc" and you try putting an object into a S3 bucket with the key as follows:

```
def do_something():
  my_bucket = boto3.resource('s3', region_name='us-west-2').Bucket('my_bucket')
  my_bucket.put_object(Key='/abc', Body='ABCD')
```

If you run this snippet of code (along with required imports and so on), then an object with the key "/abc" will be placed in the bucket with the content specified.

When you view this on the AWS Console, it will seem a bit strange because at the "root" level of the S3 bucket, you'll find a "folder" with no name. Upon clicking this no-named folder, you'll see an object named "abc" with the content specified. So this is the expected behavior when boto3 is being used without moto.

When moto is used in a similar situation, lets say as follows:

```
@moto_s3
def do_something_moto()
  conn = botot3.resource('s3', region_name='us-west-2')
  conn.create_bucket(Bucket='my_bucket')
  bucket.put_object(Key='/abc', Body='ABCD')
```

If you run this code, you will get an error with a traceback ending with:

`ConnectionError: Connection refused: PUT https://foobar.s3.us-west-2.amazonaws.com//abc`

If you remove the leading forward slash from the Key in the put_object call:

`bucket.put_object(Key='abc', Body='ABCD')`

Then this code will run without an error. As pointed out above when using the boto3 without moto, a Key with a leading forward slash is a valid key and should not cause an error. This is really an edge case, but I think moto should be made to handle this situation correctly.

As for versions of boto3 and moto being used:
boto3: 1.7.19
moto: 1.3.3

RDProjekt:
Allows request paths to start with multiple slashes for S3. This PR fixes #1637 for mocked requests.

Unfortunately, similar problem exists for server mode, but it's caused by an error in Werkzeug.

### DIFF
--- a/moto/s3/urls.py
+++ b/moto/s3/urls.py
@@ -15,4 +15,6 @@ url_paths = {
     '{0}/(?P<key_or_bucket_name>[^/]+)/?$': S3ResponseInstance.ambiguous_response,
     # path-based bucket + key
     '{0}/(?P<bucket_name_path>[^/]+)/(?P<key_name>.+)': S3ResponseInstance.key_response,
+    # subdomain bucket + key with empty first part of path
+    '{0}//(?P<key_name>.*)$': S3ResponseInstance.key_response,
 }
--- a/moto/server.py
+++ b/moto/server.py
@@ -174,10 +174,11 @@ def create_backend_app(service):
     backend_app.url_map.converters['regex'] = RegexConverter
     backend = list(BACKENDS[service].values())[0]
     for url_path, handler in backend.flask_paths.items():
+        view_func = convert_flask_to_httpretty_response(handler)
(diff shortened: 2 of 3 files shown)


spulec:
Looks great; thank you!

PR #2399, opened 2019-08-28
Don't error out on route53.list_tags_for_resource when resource has no tags

dkuntz2:
Currently, when you call `boto3.client('route53').list_tags_for_resource()` on a resource without any tags, it results in a jinja templating error (because the backend returns `None` when the template expects a dict).

This fixes the backend to return an empty dict when no resources are found, so that the call works as expected.


spulec:
Thanks!

PR #2400, opened 2019-08-30
Terminated EC2 instances are deregistered from ELBv2 target groups.

acsbendi:
Previously, when querying the target healths for an ELBv2 target group (DescribeTargetHealth), it showed terminated instance to be healthy. Now when an instance is terminated, it is also completely removed from any ELBv2 target group it has been registered with.


spulec:
Looks great; thank you!

PR #2401, opened 2019-08-31
add enhanced vpc routing option to redshift moto

rocky4570:
Enhanced VPC Routing was added sometime ago and now add for unit testing.

Also EnhancedVpcRouting is only available when mock_redshift not mock_redshift_deprecated (boto3 not boto).


spulec:
Thank you!

PR #2403, opened 2019-09-01
Adds Cognito Identity describe-identity-pool

wesselvdv:
This allows it to work with Terraform.


spulec:
Looks great; thank you!

PR #2409, opened 2019-09-04
Step Functions - basic method implementation

bblommers:
Hi,
A first attempt at the implementation for Step Functions. 

The start_execution does not do anything at the moment - don't know whether it's achievable/recommendable to actually execute the underlying activities/lambdas, considering it's supposed to be a mock?

Right now it just creates a static Execution object that is 'Running', until you manually stop() it.

In my (limited) experience, the State Machine was set up as part of the infrastructure.
Subsequent method calls would only call start_execution, so the test setup would look like this:

```python
def before():
  create_state_machine(definition:'NOOP')

@mock_stepfunctions
def test():
  method_under_test()
  assert state_machine was executed

def method_under_test():
  client.start_execution(state_machine)
```

This doesn't test the actual definition of course. Ideally we at least validate the provided JSON - but that would require some more work.

LMKWYT


bblommers:
@spulec If you have some time, would appreciate some guidance on why the tests are failing in Travis. Works fine locally, in both server modes..

spulec:
It looks like the issues are coming from stepfunctions/models.py:L162 (`identity = sts.get_caller_identity()`). Instead of creating a boto client, you can try importing the STS backend directly.

bblommers:
Thanks @spulec - that did the trick

spulec:
Looks great; thanks!

bblommers:
FYI @kilsbo and @estherthetester
start_execution is effectively a no-op at the moment, but it's a start!

PR #2411, opened 2019-09-05
sqs physical resource id is queue url in cfn list-stack-resources 

### LINKED ISSUE #2407: SQS PhysicalResourceId should always be its QueueUrl
According to the [AWS CFN docs for SQS](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-sqs-queues.html#aws-properties-sqs-queues-return-values-ref), the physical resource ID should be its queue URL.

>Ref
>
>When you pass the logical ID of this resource to the intrinsic Ref function, Ref returns the queue URL. For example:
>
>{ "Ref": "https://sqs.us-east-2.amazonaws.com/123456789012/aa4-MyQueue-Z5NOSZO2PZE9" }

However, `moto` currently uses the queue name as the physical resource ID:

https://github.com/spulec/moto/blob/3a5d857a60c3a2d140ed2c8adfe8dcaf71a4cac8/moto/sqs/models.py#L322-L324

maxenglander:
The physical resource id of AWS::SQS::Queue should be the queue URL [according to the AWS CFN docs for SQS](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-sqs-queues.html#aws-properties-sqs-queues-return-values-ref) but is currently the queue name in Moto.

This PR makes the necessary change for the `cloudformation:ListStackResources` action. It should be easy to extend the same approach in this PR to other CFN actions.

Fixes #2407

### DIFF
--- a/moto/cloudformation/responses.py
+++ b/moto/cloudformation/responses.py
@@ -226,9 +226,16 @@ class CloudFormationResponse(BaseResponse):
         stack_name_or_id = self._get_param('StackName')
         resources = self.cloudformation_backend.list_stack_resources(
             stack_name_or_id)
+        request_url = urlparse(self.uri)
+        resources_with_request_url = []
+        for resource in resources:
+            if hasattr(resource, 'with_request_url'):
+                resources_with_request_url.append(resource.with_request_url(request_url))
+            else:
+                resources_with_request_url.append(resource)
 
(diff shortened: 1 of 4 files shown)


maxenglander:
This actually doesn't work as intended -- it will always return the CloudFormation endpoint as the base of the queue URL, when what is wanted is the SQS endpoint.

Issue #10, opened 2013-04-09
Demonstrate how to mock describe_instances for existing instances


boosh:
Please can you demonstrate how to mock a call to get_all_instances() so that it will return a preconfigured set of instances. Here's the code I want to test:

```
conn = boto.ec2.connect_to_region(
        self.region, aws_access_key_id=self.access_key,
        aws_secret_access_key=self.secret_key)
reservations = conn.get_all_instances()
```

How can I mock the response to the 'reservations' variable?

Thanks

PR #2412, opened 2019-09-09
Prevent overlapping expr name prefixes from corrupting projection expr

joolean:
This patch handles the case when ProjectionExpression looks like "#1, ..., #10" - the previous code used `replace`, which would make the resulting projection into "foo, ..., foo0".


spulec:
Thank you!

PR #2415, opened 2019-09-11
Fix multiple IAM Policy Statement creation with empty sid

brunog3:
When I executed terraform with multiple statements it generate an request with empty `Sid` attribute on each statement.
This problem was originally created in localstack ([Issue 1556](https://github.com/localstack/localstack/issues/1556)), but the related validation was applied by moto.
The same request were sent to AWS and were successful.


spulec:
Thank you!

PR #2417, opened 2019-09-12
MaxKeys limits the sum of folders and keys

Gapex:
As mentioned in https://github.com/spulec/moto/pull/2375

<blockquote>
Besides, the <code>MaxKeys</code> limits the sum of CommonPrefixes and Contents other than each. If this change got accepted I will propose that one.
</blockquote>

### DIFF
--- a/moto/s3/responses.py
+++ b/moto/s3/responses.py
@@ -2,6 +2,8 @@ from __future__ import unicode_literals
 
 import re
 
+from collections import namedtuple
+
 import six
 
 from moto.core.utils import str_to_rfc_1123_datetime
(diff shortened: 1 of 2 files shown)


moto/s3/responses.py
```diff
@@ -458,10 +458,10 @@ def _handle_list_objects_v2(self, bucket_name, querystring):
             else:
                 result_folders = self._get_results_from_token(result_folders, limit)
 
-        if not delimiter:
-            result_keys, is_truncated, next_continuation_token = self._truncate_result(result_keys, max_keys)
-        else:
-            result_folders, is_truncated, next_continuation_token = self._truncate_result(result_folders, max_keys)
+        all_keys = [(key, True) for key in result_keys] + [(folder, False) for folder in result_folders]
```
vuryleo: make (key, True) as a NamedTuple looks better

Gapex:
a doubt about result_keys and result_folders: why each key is of type FakeKey, but folder is string? What about uniforming the both type?

spulec:
Looks great; thank you!

PR #2420, opened 2019-09-13
Normalize KMS key ID validation

mattsb42-aws:
Following up from #2397, this:
1. Normalizes how key IDs are validated for KMS calls.
2. Adds key ID validation to all KMS calls that were missing key ID validation.

I also made sure that each method accepts the same types of key IDs (ex: aliases or no aliases) as the actual KMS APIs.


spulec:
Looks great; thank you!

PR #2422, opened 2019-09-16
fix: raises ValueError on conditional and operator

kiyonori-matsumoto:
if lhs evaluates to false, rhs must be ignored, but rhs was evaluated then ValueError is occurred.

I fixed this issue same as OR operator.


spulec:
Thank you!

PR #2428, opened 2019-09-23
fix tail message problem in get_log_events

mangin:


spulec:
Looks great; thank you!

PR #2430, opened 2019-09-24
WIP Fix for #2431 delete objects fails with url encoded key

jessevogt:
Issue: https://github.com/spulec/moto/issues/2431

~~Open to suggestions on a better approach than just calling clean_key_name multiple times.~~

Updated solution in the PR is to undo the the key cleaning in the delete keys method.

Open to any other suggestions.

update 2019-09-24:
Something is still not quite right - I am using moto to mock an s3 back for use some spark integration tests. The write is still failing and I haven't been able to full figure out what is different when the key is url encoded. I'll work on getting some tcpdumps of a working flow and non-working flow for further analysis.


spulec:
Looks great; thanks!

PR #2436, opened 2019-09-29
Fix copying S3 objects with question marks in keys

yan12125:
PR #2062 fixes copy for unicode keys. However, the change breaks keys
with question marks. Here is an alternative fix that works for both
emojis and question marks.

I notice this when running the test suite of aiobotocore. One of the
tests [1] is broken with moto >= 1.3.8.

@domtes Could you check if your tests are still green after this change?

[1] https://github.com/aio-libs/aiobotocore/blob/0.10.3/tests/test_basic_s3.py#L314-L329

### DIFF
--- a/moto/s3/responses.py
+++ b/moto/s3/responses.py
@@ -913,11 +913,11 @@ class ResponseObject(_TemplateEnvironmentMixin, ActionAuthenticatorMixin):
             # Copy key
             # you can have a quoted ?version=abc with a version Id, so work on
             # we need to parse the unquoted string first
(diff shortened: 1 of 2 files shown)


yan12125:
Oh, the original fix is actually for Python 2. Let me check if there is a universal fix...

yan12125:
Fortunately a universal fix is not difficult 😃

mikegrima:
LGTM

thehesiod:
hah, just ran into this while attempting to bump aiobotocore reqs, thanks

Issue #2441, opened 2019-10-02
'attributes' is empty for SNS subscriptions


Yaronn44:
When SNS `subscribe` function is call, a SNS `subscription` is created but some informations are missing.

With a real `subscription` creation, the field `attributes` of every subscription look something like this :
```
{
  'Owner': '123456789012', 
  'RawMessageDelivery': 'false', 
  'TopicArn': 'arn:aws:sns:eu-central-1:123456789012:topic_xxxx', 
  'Endpoint': 'https://test.com', 
  'EffectiveDeliveryPolicy': '{"healthyRetryPolicy":{"minDelayTarget":20,"maxDelayTarget":20,"numRetries":3,"numMaxDelayRetries":0,"numNoDelayRetries":0,"numMinDelayRetries":0,"backoffFunction":"linear"},"sicklyRetryPolicy":null,"throttlePolicy":null,"guaranteed":false}', 
  'Protocol': 'https', 
  'PendingConfirmation': 'false', 
  'ConfirmationWasAuthenticated': 'false', 
  'SubscriptionArn': 'arn:aws:sns:eu-central-1:123456789012:topic_xxxx:11111111-2222-3333-4444-5555555555'
}
```

With mocked subscription, this field `subscription` is totally empty. 
In my case, I need at least to have the `PendingConfirmation` (I discovered I can directly check the subscription's arn instead of checking this but whatever), `TopicArn` and `Endpoint` fields to be present in the `attributes` dict.

I'm going to make a PR for having the :  'PendingConfirmation', 'Endpoint', 'TopicArn', 'Protocol and 'SubscriptionArn' in the `attributes` dict. They can be retrieve and set easily so it's at least a good start.

PR #2442, opened 2019-10-02
Issue #2441 Add informations in SNS subscriptions `attributes`

Yaronn44:
As I explained in the issue #2441 the field `attributes` of SNS subscription is empty when they are created with moto. 
This PR is only here to have mocked subscriptions that look a bit more like real subscriptions.

### DIFF
--- a/moto/sns/models.py
+++ b/moto/sns/models.py
@@ -340,9 +340,9 @@ class SNSBackend(BaseBackend):
         topic = self.get_topic(topic_arn)
         subscription = Subscription(topic, endpoint, protocol)
         attributes = {
-            'PendingConfirmation' : 'false',
-            'Endpoint' : endpoint,
-            'TopicArn' : topic_arn,
+            'PendingConfirmation': 'false',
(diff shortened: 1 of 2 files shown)


Yaronn44:
When trying to run `make` i'm having a problem with libxml2 that's not installed and with multiples tries I didn't succeed to install it properly... 
I'm working on windows 10 with python 3.7
I already spend some time to try to make it work and I currently have a lot of work to do, so I'll try again later

gruebel:
@Yaronn44 I'm also working on windows 10 and installed it manually. You can find the precompiled package here https://www.lfd.uci.edu/~gohlke/pythonlibs/#lxml. Just install the newest version and you should be good to go.

Yaronn44:
I already did this but it didn't work. The problem was because of the `requirements-dev.txt` that overrode my version of lxml... I solved it on my computer by changing the version in the `requirements-dev.txt` .

Yaronn44:
No problem ! Just did a commit, tell me if you need more change about this PR.

PR #2443, opened 2019-10-03
Adding support for querying AWS Config

mikegrima:
PR to add experimental support for listing and querying AWS Config for resource types that have the feature implemented.

The benefit of this PR is to develop unit tests against a data set that AWS Config would likely return based on the existing state in Moto.

At this time, I am developing the feature for S3 as I have an immediate need for it. This PR only contains the ability to list S3 buckets in AWS Config (both aggregated and non-aggregated). A follow-up PR will contain a feature to return the object from the Config API.


mikegrima:
Going to send a follow-up PR for querying actual config data.

PR #2448, opened 2019-10-03
Stepfunctions minor improvements

jthorniley:
Some updates for the recently added stepfunctions utility.

* The ARN format for IAM seems to be wrong, there should be a double colon after ``iam``: ``arn:aws:iam::<account>...``
* I spotted a probable copy-paste mistake in an exception message, where it reported "Invalid Role ARN" where it meant "Invalid State Machine ARN"
* ``start_execution()`` can accept a ``name`` argument in which case stepfunctions will use that instead of a random UUID as the execution ID


mikegrima:
LGTM

PR #2449, opened 2019-10-03
Events and ecs multi region

nadlerjessie:
This PR does two things:
1) Pass the `region_name` to ECS models to use when creating and describing resources. Previously `us-east-1` was hard coded in the ARNs
2) Add multi region support for Cloudwatch Events

Please let me know if there are any questions, thanks!


spulec:
thank you!

PR #2450, opened 2019-10-03
Add support for tagging and untagging ECS services

jimjshields:
Adds support for:
* Passing `tags` to `create_service`
* Getting tags for an ECS service from `list_tags_for_resource`
* Tagging an ECS service with `tag_resource`
* Untagging an ECS service with `untag_resource`


spulec:
thanks!

PR #2453, opened 2019-10-05
Add ParameterFilters to SSM describe_parameters

### LINKED ISSUE #2013: SSM describe_parameters does not support ParameterFilters
### How to reproduce the issue

```python
from pprint import pprint

import boto3
import moto


def create_params_describe_params():
    ssm = boto3.client('ssm')

    ssm.put_parameter(Name='aaabbb', Value='As and Bs',
                      Type='String', Overwrite=True)
    ssm.put_parameter(Name='aaaccc', Value='As and Cs',
                      Type='String', Overwrite=True)
    ssm.put_parameter(Name='bbbccc', Value='Bs and Cs',
                      Type='String', Overwrite=True)

    response = ssm.describe_parameters(ParameterFilters=[
        {'Key': 'Name',
         'Option': 'BeginsWith',
         'Values': ['aaa']}])
    return [p['Name'] for p in response['Parameters']]


if __name__ == '__main__':
    print("Without moto:")
    pprint(create_params_describe_params())

    with moto.mock_ssm():
        print("With moto:")
        print(create_params_describe_params())
```

Output:

```
Without moto:
['aaabbb', 'aaaccc']
With moto:
['aaabbb', 'aaaccc', 'bbbccc']
```

### What I expect to happen

The name "bbbccc" shouldn't be returned as it does not start with "aaa".

### What does happen

"bbbccc" is returned.

Note that I originally thought this was a duplicate of #1517, but I now realise that it's a separate issue.

gruebel:
Fixes #2013
Additionally added validation for ParameterFilters and cleaned up some of the tests.

I think there is still a lot to do, but this should be enough to fix the issue and I didn't want to blow up this PR.

### DIFF
--- a/moto/ssm/exceptions.py
+++ b/moto/ssm/exceptions.py
@@ -1,0 +1,34 @@
+from __future__ import unicode_literals
+from moto.core.exceptions import JsonRESTError
+
+
+class InvalidFilterKey(JsonRESTError):
+    code = 400
+
+    def __init__(self, message):
+        super(InvalidFilterKey, self).__init__(
+            "InvalidFilterKey", message)
+
+
+class InvalidFilterOption(JsonRESTError):
+    code = 400
+
+    def __init__(self, message):
+        super(InvalidFilterOption, self).__init__(
+            "InvalidFilterOption", message)
+
+
+class InvalidFilterValue(JsonRESTError):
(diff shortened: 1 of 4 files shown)


mikegrima:
LGTM

Issue #815, opened 2017-01-24
Support for DynamoDB Streams?


jong-eatsa:
I'm wondering if you have any plans to support DynamoDB Streams

spulec:
No plans now as it isn't something I use, but pull requests are very much welcome if someone has time.

terrycain:
@jong-eatsa if you have some example code you want to work, preferably with some example data, I'll have a look

bblommers:
@jong-eatsa @spulec This should be implemented now!

spulec:
Closed with #2457

PR #2457, opened 2019-10-07
Feature: Add option for DynamoDB stream to kick off lambda

bblommers:
Currently, only option for EventSourceMappings is to kick off a Lambda function after receiving an SQS message
This PR adds the option to kick off a Lambda function after updating a DynamoDB table

### DIFF
--- a/moto/awslambda/models.py
+++ b/moto/awslambda/models.py
@@ -694,18 +694,15 @@ class LambdaBackend(BaseBackend):
                     queue.lambda_event_source_mappings[esm.function_arn] = esm
 
                     return esm
-        try:
-            stream = json.loads(dynamodbstreams_backends[self.region_name].describe_stream(spec['EventSourceArn']))
-            spec.update({'FunctionArn': func.function_arn})
-            esm = EventSourceMapping(spec)
-            self._event_source_mappings[esm.uuid] = esm
-            table_name = stream['StreamDescription']['TableName']
-            table = dynamodb_backends2[self.region_name].get_table(table_name)
-            table.lambda_event_source_mappings[esm.function_arn] = esm
-
-            return esm
-        except Exception:
-            pass  # No DynamoDB stream exists
+        for stream in json.loads(dynamodbstreams_backends[self.region_name].list_streams())['Streams']:
+            if stream['StreamArn'] == spec['EventSourceArn']:
+                spec.update({'FunctionArn': func.function_arn})
+                esm = EventSourceMapping(spec)
+                self._event_source_mappings[esm.uuid] = esm
+                table_name = stream['TableName']
+                table = dynamodb_backends2[self.region_name].get_table(table_name)
+                table.lambda_event_source_mappings[esm.function_arn] = esm
+                return esm
         raise RESTError('ResourceNotFoundException', 'Invalid EventSourceArn')
 
     def publish_function(self, function_name):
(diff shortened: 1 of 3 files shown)


bblommers:
Suppose this fixes https://github.com/spulec/moto/issues/815?

moto/awslambda/models.py
```diff
@@ -692,6 +694,18 @@ def create_event_source_mapping(self, spec):
...
+            stream = json.loads(dynamodbstreams_backends[self.region_name].describe_stream(spec['EventSourceArn']))
+            spec.update({'FunctionArn': func.function_arn})
+            esm = EventSourceMapping(spec)
+            self._event_source_mappings[esm.uuid] = esm
+            table_name = stream['StreamDescription']['TableName']
+            table = dynamodb_backends2[self.region_name].get_table(table_name)
+            table.lambda_event_source_mappings[esm.function_arn] = esm
+
+            return esm
+        except Exception:
```
spulec: Can we make this more specific to the exceptions that can be thrown here?

moto/awslambda/models.py
```diff
@@ -811,6 +825,19 @@ def send_sns_message(self, function_name, message, subject=None, qualifier=None)
...
         func.invoke(json.dumps(event), {}, {})
 
+    def send_dynamodb_items(self, function_arn, items, source):
+        event = {'Records': [
+            {
+                'eventID': item.to_json()['eventID'],
+                'eventName': 'INSERT',
+                'eventVersion': item.to_json()['eventVersion'],
+                'eventSource': item.to_json()['eventSource'],
+                'awsRegion': 'us-east-1',
```
spulec: Should this be `self.region_name`?

moto/dynamodb2/models.py
```diff
@@ -409,6 +409,15 @@ def add(self, old, new):
         seq = len(self.items) + self.starting_sequence_number
         self.items.append(
             StreamRecord(self.table, t, event_name, old, new, seq))
+        result = None
+        from moto.awslambda import lambda_backends
```
spulec: Any reason to import this inline vs at the module level?

tests/test_awslambda/test_lambda.py
```diff
@@ -1027,6 +1027,54 @@ def test_invoke_function_from_sqs():
...
+        if not log_streams:
+            time.sleep(1)
+            continue
+
+        assert len(log_streams) == 1
+        result = logs_conn.get_log_events(logGroupName='/aws/lambda/testFunction', logStreamName=log_streams[0]['logStreamName'])
+        for event in result.get('events'):
+            if event['message'] == 'get_test_zip_file3 success':
+                return
+        time.sleep(1)
```
spulec: What is the sleeping for? Is there a way to make this test without needing them?

moto/awslambda/models.py
bblommers: Good spot - I've updated it

moto/awslambda/models.py
bblommers: Rewritten this so that it doesn't throw an exception at all

moto/dynamodb2/models.py
bblommers: I've followed the style in the sqs/models.py, where a lambda function is kicked off whenever an SQS message is received. There the lambda_backends  is also imported locally.
I'm ambivalent where the import happens really - it shouldn't be too resource-heavy.

tests/test_awslambda/test_lambda.py
bblommers: Again following the SQS example, in the test just above. I guess the Lambda is invoked async when this is run in server mode, so we'd really have to wait for the Lambda to finish executing/writing logs

moto/dynamodb2/models.py
spulec: Ah, no. This is right. Sorry about that.

tests/test_awslambda/test_lambda.py
spulec: Ah, okay. That is a bit annoying, but I can understand it and don't have a better suggestion off the top of my head.

spulec:
This is great; thanks for helping!

PR #2468, opened 2019-10-10
Remove dead code

bblommers:
Removed some small methods that were never called.

The RDS model was deprecated in favour of RDS2


spulec:
Thank you!

PR #2471, opened 2019-10-10
Implement support for lambda policies

JackDanger:


JackDanger:
@mikegrima Thanks for asking about the tests - I was wondering why the old tests didn't fail and it led me to two methods that didn't have the `test_*` prefix and therefore weren't executing in the test suite. Fixed! This code is now covered by tests.

PR #2474, opened 2019-10-12
Added Exception to describe_container_instances to match aws api

### LINKED ISSUE #2473: containerInstances argument to ECS DescribeContainerInstances cannot be empty
I looked and couldn't find this, so I'm reporting it here, but I'm happy to look into the solution and do a PR for it.

# Reproduction Steps

```
import boto3
ecs_client = boto3.client('ecs')
cluster_name = 'test-cluster'
ecs_client.create_cluster(clusterName=cluster_name)
print(ecs_client.describe_container_instances(cluster=cluster_name,containerInstances=[]))
```
# What I expect to happen

If I run the same code against the actual AWS SDK, I get the following exception:
```
InvalidParameterException: An error occurred (InvalidParameterException) when calling the DescribeContainerInstances operation: Container instance cannot be empty.
```

# What actually happens

If I run this with a `mock_ecs` like the following

```
import boto3

from moto import mock_ecs

with mock_ecs():
    ecs_client = boto3.client('ecs')
    cluster_name = 'test-cluster'
    ecs_client.create_cluster(clusterName=cluster_name)
    print(ecs_client.describe_container_instances(cluster=cluster_name,containerInstances=[]))
```
I get:

```
{'containerInstances': [], 'failures': [], 'ResponseMetadata': {'HTTPStatusCode': 200, 'HTTPHeaders': {'server': 'amazon.com'}, 'RetryAttempts': 0}}
```
Although this is a nicer answer, it isn't consistent with the AWS API, and therefore hides edge cases.

# Version

I tested this against the latest commit at the time of reporting: `31198c75b8011f067e8ebdc97c604f555d2e7c04`

jrstarke:
Fixes #2473

### DIFF
--- a/moto/ecs/models.py
+++ b/moto/ecs/models.py
@@ -790,6 +790,8 @@ class EC2ContainerServiceBackend(BaseBackend):
         cluster_name = cluster_str.split('/')[-1]
         if cluster_name not in self.clusters:
             raise Exception("{0} is not a cluster".format(cluster_name))
+        if not list_container_instance_ids:
+            raise JsonRESTError('InvalidParameterException', 'Container instance cannot be empty')
         failures = []
         container_instance_objects = []
         for container_instance_id in list_container_instance_ids:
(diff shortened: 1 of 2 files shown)


mikegrima:
LGTM

PR #2475, opened 2019-10-12
Add sns tags

gruebel:
I added following methods to SNS

- list_tags_for_resource
- tag_resource
- untag_resource


mikegrima:
Nice job!

PR #2478, opened 2019-10-13
Added AWS Config query batching capabilities

mikegrima:
- Support for aggregated and non-aggregated batching
- Currently only implemented for S3


spulec:
LGTM

PR #2480, opened 2019-10-15
fix(rds2): handle create_db_instance when AllocatedStorage is not specified

hakamadare:
In all of the tests of `create_db_instance()`, the `AllocatedStorage` parameter is provided.  The [RDS API reference](https://docs.aws.amazon.com/AmazonRDS/latest/APIReference/API_CreateDBInstance.html) says this parameter is optional; however, when none is provided, moto returns an obscure error message:

```
self = <botocore.parsers.QueryParser object at 0x113745890>, shape = <Shape(Integer)>, text = 'None'

    @_text_content
    def _handle_integer(self, shape, text):
>       return int(text)
E       ValueError: invalid literal for int() with base 10: 'None'

/usr/local/Cellar/pyenv/1.2.13_1/versions/3.7.4/envs/rds_encrypt/lib/python3.7/site-packages/botocore/parsers.py:466: ValueError
```

This PR adds default values that correspond to the current default API behaviors.

### DIFF
--- a/moto/rds2/models.py
+++ b/moto/rds2/models.py
@@ -66,7 +66,7 @@ class Database(BaseModel):
             self.auto_minor_version_upgrade = True
         self.allocated_storage = kwargs.get('allocated_storage')
         if self.allocated_storage is None:
-            self.allocated_storage = Database.default_allocated_storage(engine=self.engine,storage_type=self.storage_type)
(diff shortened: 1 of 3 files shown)


mikegrima:
LGTM

PR #2481, opened 2019-10-16
Fix getsecretmananger error

aacampbell:
Secretsmanager will raise a ResourceNotFoundException with a slightly different message when a secret exists but has not SecretString or SecretBinary. Current behaviour of moto is to successfully return a response dictionary with no SecretString or SecretBinary, which does not happen with boto3.

Existing uses of ResourceNotFoundException will still give the same message, and the newly added one will give the observed, different message. A slight punctuation mistake was fixed in the original message.

### DIFF
--- a/moto/secretsmanager/models.py
+++ b/moto/secretsmanager/models.py
@@ -1,3 +1,4 @@
+# -*- coding: utf-8 -*-
 from __future__ import unicode_literals
 
 import time
--- a/moto/secretsmanager/models.py
+++ b/moto/secretsmanager/models.py
@@ -88,7 +88,7 @@ class SecretsManagerBackend(BaseBackend):
 
         if 'secret_string' not in secret_version and 'secret_binary' not in secret_version:
             raise ResourceNotFoundException(
-                u"Secrets Manager can’t find the specified secret value for staging label: %s" %
+                u"Secrets Manager can\u2019t find the specified secret value for staging label: %s" %
                 (version_stage or u"AWSCURRENT")
             )
 
--- a/moto/secretsmanager/models.py
+++ b/moto/secretsmanager/models.py
@@ -11,6 +11,7 @@ import boto3
 from moto.core import BaseBackend, BaseModel
 from .exceptions import (
     ResourceNotFoundException,
+    SecretNotFoundException,
     InvalidParameterException,
     ResourceExistsException,
     InvalidRequestException,
@@ -47,7 +48,7 @@ class SecretsManagerBackend(BaseBackend):
     def get_secret_value(self, secret_id, version_id, version_stage):
 
         if not self._is_valid_identifier(secret_id):
-            raise ResourceNotFoundException(u"Secrets Manager can’t find the specified secret.")
+            raise SecretNotFoundException()
 
         if not version_id and version_stage:
             # set version_id to match version_stage
@@ -57,7 +58,7 @@ class SecretsManagerBackend(BaseBackend):
                     version_id = ver_id
                     break
             if not version_id:
-                raise ResourceNotFoundException(u"Secrets Manager can’t find the specified secret.")
+                raise SecretNotFoundException()
 
         # TODO check this part
         if 'deleted_date' in self.secrets[secret_id]:
@@ -176,7 +177,7 @@ class SecretsManagerBackend(BaseBackend):
 
     def describe_secret(self, secret_id):
         if not self._is_valid_identifier(secret_id):
-            raise ResourceNotFoundException(u"Secrets Manager can’t find the specified secret.")
+            raise SecretNotFoundException()
 
         secret = self.secrets[secret_id]
 
@@ -205,7 +206,7 @@ class SecretsManagerBackend(BaseBackend):
         rotation_days = 'AutomaticallyAfterDays'
 
         if not self._is_valid_identifier(secret_id):
-            raise ResourceNotFoundException(u"Secrets Manager can’t find the specified secret.")
+            raise SecretNotFoundException()
 
         if 'deleted_date' in self.secrets[secret_id]:
             raise InvalidRequestException(
@@ -347,7 +348,7 @@ class SecretsManagerBackend(BaseBackend):
     def delete_secret(self, secret_id, recovery_window_in_days, force_delete_without_recovery):
 
         if not self._is_valid_identifier(secret_id):
-            raise ResourceNotFoundException(u"Secrets Manager can’t find the specified secret.")
+            raise SecretNotFoundException()
 
         if 'deleted_date' in self.secrets[secret_id]:
             raise InvalidRequestException(
@@ -377,7 +378,7 @@ class SecretsManagerBackend(BaseBackend):
             secret = self.secrets.get(secret_id, None)
 
         if not secret:
-            raise ResourceNotFoundException(u"Secrets Manager can’t find the specified secret.")
+            raise SecretNotFoundException()
 
         arn = secret_arn(self.region, secret['secret_id'])
         name = secret['name']
@@ -387,7 +388,7 @@ class SecretsManagerBackend(BaseBackend):
     def restore_secret(self, secret_id):
 
         if not self._is_valid_identifier(secret_id):
-            raise ResourceNotFoundException(u"Secrets Manager can’t find the specified secret.")
+            raise SecretNotFoundException()
 
         self.secrets[secret_id].pop('deleted_date', None)
 
--- a/moto/secretsmanager/models.py
+++ b/moto/secretsmanager/models.py
@@ -46,7 +46,7 @@ class SecretsManagerBackend(BaseBackend):
     def get_secret_value(self, secret_id, version_id, version_stage):
 
         if not self._is_valid_identifier(secret_id):
-            raise ResourceNotFoundException()
+            raise ResourceNotFoundException("Secrets Manager can't find the specified secret")
 
         if not version_id and version_stage:
             # set version_id to match version_stage
@@ -56,7 +56,7 @@ class SecretsManagerBackend(BaseBackend):
                     version_id = ver_id
                     break
             if not version_id:
-                raise ResourceNotFoundException()
+                raise ResourceNotFoundException("Secrets Manager can't find the specified secret")
 
         # TODO check this part
         if 'deleted_date' in self.secrets[secret_id]:
@@ -83,6 +83,12 @@ class SecretsManagerBackend(BaseBackend):
 
         if 'secret_binary' in secret_version:
             response_data["SecretBinary"] = secret_version['secret_binary']
+
+        if 'secret_string' not in secret_version and 'secret_binary' not in secret_version:
+            raise ResourceNotFoundException(
+                "Secrets Manager can’t find the specified secret value for staging label: %s" %
+                (version_stage or "AWSCURRENT")
+            )
 
         response = json.dumps(response_data)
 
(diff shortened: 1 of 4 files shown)


aacampbell:
Oh wow, even botocore itself can't handle the punctuation under python2:

```
In [4]: client.get_secret_value(SecretId='bogus')
---------------------------------------------------------------------------
ResourceNotFoundException                 Traceback (most recent call last)
<ipython-input-4-0c21453728ff> in <module>()
----> 1 client.get_secret_value(SecretId='bogus')

/home/acampbell/.local/lib/python2.7/site-packages/botocore/client.pyc in _api_call(self, *args, **kwargs)
    355                     "%s() only accepts keyword arguments." % py_operation_name)
    356             # The "self" in this scope is referring to the BaseClient.
--> 357             return self._make_api_call(operation_name, kwargs)
    358 
    359         _api_call.__name__ = str(py_operation_name)

/home/acampbell/.local/lib/python2.7/site-packages/botocore/client.pyc in _make_api_call(self, operation_name, api_params)
    659             error_code = parsed_response.get("Error", {}).get("Code")
    660             error_class = self.exceptions.from_code(error_code)
--> 661             raise error_class(parsed_response, operation_name)
    662         else:
    663             return parsed_response

<type 'str'>: (<type 'exceptions.UnicodeEncodeError'>, UnicodeEncodeError('ascii', u'An error occurred (ResourceNotFoundException) when calling the GetSecretValue operation: Secrets Manager can\u2019t find the specified secret.', 108, 109, 'ordinal not in range(128)'))

In [5]: 
```

aacampbell:
Thanks for the suggestion, but I don't think that solves the actual problem.

The problem is that AWS themselves actually send that U+2019 in the response error message as part of a unicode string. It is actually incorrect for moto to give a message without it. This is no problem for python3, where `str()` can handle unicode. This is a problem for python2 where calling `str()` on a unicode string gives you a `UnicodeDecodeError`. Exceptions under python3 no longer have a `.message` field, and it is actually deprecated for python2.7, so the cross-compatible way of getting an exception's message is by calling `str()` on it. This happens quite commonly in tests, where one might call `unittest.assertRaisesRegex` or something similar. It also happens when tracebacks are printed, as shown in my above snippet - what happens is that it tries to print the type of the exception and then its message, but the `UnicodeDecodeError` happens during the printing of the traceback! Not good.

For clarity, I'll show that the tracebacks actually look different under the standard python2 REPL and under ipython, which lends credence to the argument that the error occurs during the printing of the traceback:
```
$ python2
Python 2.7.15+ (default, Oct  7 2019, 17:39:04) 
[GCC 7.4.0] on linux2
>>> import boto3
>>> client = boto3.client('secretsmanager')
>>> client.get_secret_value(SecretId='bogus')
  File "/home/acampbell/.local/lib/python2.7/site-packages/botocore/client.py", line 357, in _api_call
  File "/home/acampbell/.local/lib/python2.7/site-packages/botocore/client.py", line 661, in _make_api_call
botocore.errorfactory.ResourceNotFoundException: <exception str() failed>
>>> exit()
$ ipython2
Python 2.7.15+ (default, Oct  7 2019, 17:39:04) 
Type "copyright", "credits" or "license" for more information.

IPython 5.8.0 -- An enhanced Interactive Python.
?         -> Introduction and overview of IPython's features.
%quickref -> Quick reference.
help      -> Python's own help system.
object?   -> Details about 'object', use 'object??' for extra details.

In [1]: import boto3

In [2]: client = boto3.client('secretsmanager')

In [3]: client.get_secret_value(SecretId='bogus')
---------------------------------------------------------------------------
ResourceNotFoundException                 Traceback (most recent call last)
<ipython-input-3-0c21453728ff> in <module>()
----> 1 client.get_secret_value(SecretId='bogus')

/home/acampbell/.local/lib/python2.7/site-packages/botocore/client.pyc in _api_call(self, *args, **kwargs)
    355                     "%s() only accepts keyword arguments." % py_operation_name)
    356             # The "self" in this scope is referring to the BaseClient.
--> 357             return self._make_api_call(operation_name, kwargs)
    358 
    359         _api_call.__name__ = str(py_operation_name)

/home/acampbell/.local/lib/python2.7/site-packages/botocore/client.pyc in _make_api_call(self, operation_name, api_params)
    659             error_code = parsed_response.get("Error", {}).get("Code")
    660             error_class = self.exceptions.from_code(error_code)
--> 661             raise error_class(parsed_response, operation_name)
    662         else:
    663             return parsed_response

<type 'str'>: (<type 'exceptions.UnicodeEncodeError'>, UnicodeEncodeError('ascii', u'An error occurred (ResourceNotFoundException) when calling the GetSecretValue operation: Secrets Manager can\u2019t find the specified secret.', 108, 109, 'ordinal not in range(128)'))

In [4]:
```

This is actually a botocore issue, but it seems appropriate that moto actually replicate this behaviour in order to be representative.

mikegrima:
Oof! Someone at AWS needs to turn off smart quotes :)

Can you update the exception such that if it's running in Python 2, it takes the exception string and converts it to a unicode string?

mikegrima:
Make the string in the exception as follows:
```
message = u"botocore.errorfactory.ResourceNotFoundException: An error occurred (ResourceNotFoundException) when calling the GetSecretValue operation: Secrets Manager can\u2019t find the specified secret."
```
I like having the explicit unicode declaration in there. And keep it in that one location.

aacampbell:
There is actually more than one message that can be associated with that exception. The first is the u"Secrets Manager can’t find the specified secret." and the second I've seen is the one I added in the initial commits, which is u"Secrets Manager can’t find the specified secret value for staging label: AWSCURRENT".

If you look at the order of commits, I first add in that second error case (and its unusual punctuation) and then later fix the original string to also include the U+2019. So it looks like we do need to have a configurable message for that exception type.

mikegrima:
Could you make a separate Exception class unique to this error message?

aacampbell:
That would be incorrect, as it is `boto3`'s behaviour to give the same class of exception but with a different message - it was an odd anomaly that in `moto` this exception had a baked-in message. See the follow snippet:
```
$ python3
Python 3.6.8 (default, Oct  7 2019, 12:59:55) 
[GCC 8.3.0] on linux
>>> import boto3
>>> client = boto3.client('secretsmanager')
>>> secret_name = '/test/bogus-secret'
>>> client.get_secret_value(SecretId=secret_name)
  File "/home/acampbell/.local/lib/python3.6/site-packages/botocore/client.py", line 357, in _api_call
  File "/home/acampbell/.local/lib/python3.6/site-packages/botocore/client.py", line 661, in _make_api_call
botocore.errorfactory.ResourceNotFoundException: An error occurred (ResourceNotFoundException) when calling the GetSecretValue operation: Secrets Manager can’t find the specified secret.
>>> import sys
>>> first_err = sys.last_value
>>> client.create_secret(Name=secret_name)
{'ARN': 'arn:aws:secretsmanager:ap-southeast-2:my-aws-id:secret:/test/bogus-secret-bxKNB4', 'Name': '/test/bogus-secret', 'ResponseMetadata': {'RequestId': '69467484-141b-49cf-8927-c0bab16b4228', 'HTTPStatusCode': 200, 'HTTPHeaders': {'date': 'Thu, 17 Oct 2019 01:11:45 GMT', 'content-type': 'application/x-amz-json-1.1', 'content-length': '121', 'connection': 'keep-alive', 'x-amzn-requestid': '69467484-141b-49cf-8927-c0bab16b4228'}, 'RetryAttempts': 0}}
>>> client.get_secret_value(SecretId=secret_name)
  File "/home/acampbell/.local/lib/python3.6/site-packages/botocore/client.py", line 357, in _api_call
  File "/home/acampbell/.local/lib/python3.6/site-packages/botocore/client.py", line 661, in _make_api_call
botocore.errorfactory.ResourceNotFoundException: An error occurred (ResourceNotFoundException) when calling the GetSecretValue operation: Secrets Manager can’t find the specified secret value for staging label: AWSCURRENT
>>> second_err = sys.last_value
>>> type(first_err) == type(second_err)
True
>>> str(first_err) == type(second_err)
False
>>> 
```

If the root of your concern is the repetition of the literals then I can add a commit that uses a variable.

mikegrima:
Hi @aacampbell : by `make a separate Exception` I was referring to adding a new class in the `exceptions.py` that looks exactly like the other but tailored to this exception, such as:

```
class SecretNotFoundException(SecretsManagerClientError):
    def __init__(self):
        self.code = 404
        super(SecretNotFoundException, self).__init__(
            "ResourceNotFoundException",
             message = u"Secrets Manager can\u2019t find the specified secret."
        )
```

In fact, just copy and paste that code in and then replace the instances of the `raise ResourceNotFoundException` with `SecretNotFoundException` and you should be good to go.

aacampbell:
I don't understand your reasoning for requesting that change - it makes moto unrepresentative of the behaviour of boto3.

There is no such thing as a `SecretNotFoundException` and no such thing will be raised by boto3 - this can be seen by looking at the documentation for `GetSecretValue`: https://docs.aws.amazon.com/secretsmanager/latest/apireference/API_GetSecretValue.html#API_GetSecretValue_Errors . The correct behaviour is for different `ResourceNotFoundException` instances to be raised with differing messages, which is the behaviour I've introduced. In fact, now that I look at the AWS docs, I see that moto has always been raising with an incorrect http status code (404, docs say 400).

Again, if your reasoning for requesting your change is because you do not want the string literals to be repeated, then I can trivially refactor to use a variable.

mikegrima:
> I don't understand your reasoning for requesting that change - it makes moto unrepresentative of the behaviour of boto3.
No. The change I suggested is only happening in the moto logic. It will return to the client a `ResourceNotFoundException` (see the code again), namely:

```
        super(SecretNotFoundException, self).__init__(
            "ResourceNotFoundException", # <---- This is the name of the exception that the client will receive.
             message = u"Secrets Manager can\u2019t find the specified secret."
        )
```

aacampbell:
Interesting! The system of raising exception appears to be more complicated than I realised. Must be related to how the `.response` dictionary appears in the resultant exceptions without appearing in the hierarchy. I'll make the changes you've requested.

moto/secretsmanager/models.py
```diff
@@ -87,7 +88,7 @@ def get_secret_value(self, secret_id, version_id, version_stage):
 
         if 'secret_string' not in secret_version and 'secret_binary' not in secret_version:
             raise ResourceNotFoundException(
```
mikegrima: This should just raise a `SecretNotFoundException()`.

aacampbell: No it shouldn't it gives a different message, as I've demonstrated above. This was the whole reason I made this PR.

mikegrima: Oh I'm sorry... I mis read.

In this case........... Can you make _another_ class similar to the above :)?  I mostly want to keep the special unicode exceptions in 1 place.

aacampbell: Out of interest, why use specialised classes rather than a variable passed to the same class? Wouldn't it be better to more directly mirror the AWS interface? The other exceptions for `secretsmanager` take the messages as arguments.

mikegrima: As far as the AWS interface goes... how that's implemented on AWS's end is unknown to us. Moto will still return the proper data back to the client.

From moto code maintainability perspective, we would prefer to have very specific exceptions so it's easy to find where in the code base there are issues. In this case, because there are odd things going on with unicode characters, I would like to keep that centralized as much as possible and not have odd permutations of it.

In most cases, where users are passing in custom exception messages, I typically don't care and will often prefer it for code reuse. However, because of the oddities with the unicode characters, I would like to have this centralized and isolated as much as possible.

aacampbell: Got it! Changes made.

PR #2484, opened 2019-10-16
Avoid exception occur in `moto.s3.models.S3Backend.get_key`

koshigoe:
The variable `key` may `None`.
But, its attribute `multipart` will be accessed even though `key` is `None`.

https://github.com/spulec/moto/blob/4da9ec134693afbbcf05d8ae284a140d103ba422/moto/s3/models.py#L1110-L1111

#2307


mikegrima:
LGTM

PR #2485, opened 2019-10-17
Add tagging to s3 versioned objects

Htermotto:
Adding the ability to specify the version id of an S3 key when tagging an object.

Issue: https://github.com/spulec/moto/issues/2486


mikegrima:
Very nice!

PR #2488, opened 2019-10-17
Add sqs error handling

### LINKED ISSUE #1799: Moto should raise an error when SQS send_message_batch pass an empty entries
When doing `client.send_message_batch(QueueUrl=queue_url, Entries=[])`, boto will raise an error.

```
  File "<stdin>", line 3, in <module>
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/botocore/client.py", line 314, in _api_call
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/botocore/client.py", line 612, in _make_api_call
botocore.errorfactory.EmptyBatchRequest: An error occurred (AWS.SimpleQueueService.EmptyBatchRequest) when calling the SendMessageBatch operation: There should be at least one SendMessageBatchRequestEntry in the request.
```

moto should raise the same error, rather than to be silent on that.

Environment:

```
$ python --version
Python 2.7.11
$ python -c "import boto3; print(boto3.__version__)"
1.4.7
$ python -c "import moto; print(moto.__version__)"
1.3.1
```

gruebel:
I had to rewrite `sqs.send_message_batch` a bit to be able to handle most errors.

Also fixes #1799

Over time I will also check other functions in the SQS implementation and update the coverage file while doing so.

### DIFF
--- a/moto/sqs/responses.py
+++ b/moto/sqs/responses.py
@@ -10,7 +10,6 @@ from .models import sqs_backends
 from .exceptions import (
     MessageAttributesInvalid,
     MessageNotInflight,
-    QueueDoesNotExist,
     ReceiptHandleIsInvalid,
 )
 
@@ -90,11 +89,7 @@ class SQSResponse(BaseResponse):
         request_url = urlparse(self.uri)
         queue_name = self._get_param("QueueName")
 
-        try:
-            queue = self.sqs_backend.get_queue(queue_name)
-        except QueueDoesNotExist as e:
-            return self._error('AWS.SimpleQueueService.NonExistentQueue',
-                               e.description)
+        queue = self.sqs_backend.get_queue(queue_name)
 
         if queue:
             template = self.response_template(GET_QUEUE_URL_RESPONSE)
@@ -175,11 +170,8 @@ class SQSResponse(BaseResponse):
 
     def get_queue_attributes(self):
         queue_name = self._get_queue_name()
-        try:
-            queue = self.sqs_backend.get_queue(queue_name)
-        except QueueDoesNotExist as e:
-            return self._error('AWS.SimpleQueueService.NonExistentQueue',
-                               e.description)
+
+        queue = self.sqs_backend.get_queue(queue_name)
 
         template = self.response_template(GET_QUEUE_ATTRIBUTES_RESPONSE)
         return template.render(queue=queue)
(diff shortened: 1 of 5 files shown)


moto/sqs/responses.py
```diff
@@ -397,6 +439,20 @@ def tag_queue(self):
         queue_name = self._get_queue_name()
         tags = self._get_map_prefix('Tag', key_end='.Key', value_end='.Value')
 
+        try:
+            self.sqs_backend.get_queue(queue_name)
+        except QueueDoesNotExist as e:
+            return self._error('AWS.SimpleQueueService.NonExistentQueue',
+                               e.description)
+
+        if len(tags) == 0:
```
mikegrima: Make this `if not len(tags):`

moto/sqs/responses.py
```diff
@@ -242,25 +242,59 @@ def send_message_batch(self):
 
         queue_name = self._get_queue_name()
 
+        try:
+            self.sqs_backend.get_queue(queue_name)
+        except QueueDoesNotExist as e:
```
mikegrima: Is this needed? If it raises a `QueueDoesNotExist`, that should have all the info required to format the exception to the client.

moto/sqs/responses.py
```diff
@@ -397,6 +439,20 @@ def tag_queue(self):
         queue_name = self._get_queue_name()
         tags = self._get_map_prefix('Tag', key_end='.Key', value_end='.Value')
 
+        try:
+            self.sqs_backend.get_queue(queue_name)
+        except QueueDoesNotExist as e:
```
mikegrima: Ditto on if this is needed.

moto/sqs/responses.py
```diff
@@ -418,7 +474,11 @@ def untag_queue(self):
     def list_queue_tags(self):
         queue_name = self._get_queue_name()
 
-        queue = self.sqs_backend.get_queue(queue_name)
+        try:
+            queue = self.sqs_backend.get_queue(queue_name)
+        except QueueDoesNotExist as e:
```
mikegrima: Ditto.

moto/sqs/responses.py
gruebel: The problem is the general implementation of the `QueueDoesNotExist` exception. It results in an 503 error while running the moto in server mode. I would need to change it to a `RESTError` then the handling would be right, I think.

moto/sqs/responses.py
gruebel: Done

moto/sqs/responses.py
mikegrima: Thank You -- that makes sense.

PR #2493, opened 2019-10-18
add NotImplementedError when filter_pattern is given to filter_log_events

### LINKED ISSUE #2419: Logs filter-log-events ignores filterPattern argument.
When a `filterPattern` is given to `filter_log_events`  the filter is ignored.

e.g.

```
        malformed = {"message": "hello", "timestamp": 100}

        logs_client.put_log_events(
            logGroupName=log_group,
            logStreamName=log_streams[0],
            logEvents=[malformed],
        )

       res = logs_client.filter_log_events(
            logGroupName=log_group,
            filterPattern='{$.my_json_field = "foo"}',
            startTime=0,
            endTime=12345667,
        )
 
      # Fails
       assert len(res["events"]) == 0 
```

On 1.3.9 locally (installed with pip), but the problem exists on master.

Here's the specific LOC where it just ignores `filter_pattern`: https://github.com/spulec/moto/blob/master/moto/logs/models.py#L126

I think it's probably reasonable to support basic key equality for json filters, as well as simple matching on text. Anything beyond that might get tricky, but you can throw an error in that case instead of silently ignoring.

At the very least, raising NotImplemented() when filter_pattern is non-empty would be desirable.

SelenaHaettasch:
fix #2419

### DIFF
--- a/moto/logs/models.py
+++ b/moto/logs/models.py
@@ -125,6 +125,9 @@ class LogStream:
         return events_page, get_paging_token_from_index(back_index, True), get_paging_token_from_index(next_index)
 
     def filter_log_events(self, log_group_name, log_stream_names, start_time, end_time, limit, next_token, filter_pattern, interleaved):
+        if filter_pattern:
+            raise NotImplementedError('filter_pattern is not yet implemented')
+
         def filter_func(event):
             if start_time and event.timestamp < start_time:
                 return False
--- a/tests/test_logs/test_logs.py
+++ b/tests/test_logs/test_logs.py
@@ -1,10 +1,12 @@
 import boto3
+import os
 import sure  # noqa
 import six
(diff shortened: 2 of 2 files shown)


spulec:
Thanks for opening!

It looks like the tests aren't passing in server-mode. The easiest solution for now is probably to put your new tests behind an if statement and exclude it from the server tests. Something like [this](https://github.com/spulec/moto/blob/master/tests/test_awslambda/test_lambda.py#L141-L181).

JackDanger:
@SelenaHaettasch nice work! Thanks for getting the tests passing. Your contribution here will help anyone in the world who's trying to test their log-filtering code. This is really helpful!

JackDanger:
Oh, and congrats on your first contribution. Welcome to Moto! 😃

PR #2497, opened 2019-10-21
Refactor DynamoDB update expressions

bblommers:
Hi, I've done some refactoring to the if/else/for hell in the DynamoDB update-method. (Think I'm allowed to call it that.. considering I've contributed to it as well :) )

Major changes:

- Internal representation is now nested DynamoTypes.
  Before, only the toplevel was a DynamoType:
```
DynamoType('M', {'foo': 'bar', 'foo2': ['bar2']
```
   and is now
```
DynamoType('M', {'foo': DynamoType('S': 'bar'), 'foo2': DynamoType('L', [DynamoType('S', 'bar2')])})
 ```

- Updating nested items is now the responsibility of the (DynamoType) item itself, rather than inside the update-method
- Thanks to the new DynamoType structure, updates uses recursion rather than a for-loop.

IMHO, this is a lot more readable/understandable now.

LMKWYT - open to suggestions

### DIFF
--- a/moto/dynamodb2/models.py
+++ b/moto/dynamodb2/models.py
@@ -40,8 +40,62 @@ class DynamoType(object):
     """
 
     def __init__(self, type_as_dict):
-        self.type = list(type_as_dict)[0]
-        self.value = list(type_as_dict.values())[0]
+        if type(type_as_dict) == DynamoType:
+            self.type = type_as_dict.type
+            self.value = type_as_dict.value
+        else:
+            self.type = list(type_as_dict)[0]
+            self.value = list(type_as_dict.values())[0]
+        if self.is_list():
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: moto/dynamodb2/models.py [hunk 1]
```diff
@@ -34,14 +34,76 @@ def bytesize(val):
     return len(str(val).encode('utf-8'))
 
 
+def attribute_is_list(attr):
+    """
+    Checks if attribute denotes a list, and returns the regular expression if so
```
JackDanger:
```suggestion
    Checks if attribute denotes a list, and returns the name of the list and the given list index if so
```

JackDanger:
I've left one suggested edit to documentation if you'd like to make a follow-up PR. But it's no reason to block the merge here.

PR #2499, opened 2019-10-21
Add Documentation URL

ghickman:
This adds a [Documentation URL](https://packaging.python.org/guides/distributing-packages-using-setuptools/#project-urls), which will display in the left-hand nav of the projects PyPI page, allowing users arriving there to get to the documentation slightly faster.


JackDanger:
Nice work, @ghickman! Thanks for the contribution and welcome to Moto!

PR #2502, opened 2019-10-22
Feature - Lambda Delete by arn

bblommers:
Existing tests only tested whether Lambdas could be deleted when passing in a name.
According to the docs, Lambdas can also be deleted by passing in an ARN: https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/lambda.html#Lambda.Client.delete_function

The list of mock-URLs had to be adjusted to account for the fact we're passing a HTML-safe ARN in the URL, which can contain ampersands

### DIFF
--- a/moto/awslambda/models.py
+++ b/moto/awslambda/models.py
@@ -645,8 +645,10 @@ class LambdaStorage(object):
         self._arns[fn.function_arn] = fn
         return fn
 
-    def del_function(self, name, qualifier=None):
-        if name in self._functions:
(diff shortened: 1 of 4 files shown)


bblommers:
Looks like a test failed that's unrelated to my changes (test_kms.test_generate_random(1, {})) - will look into why/how that happened

JackDanger:
I re-ran the failing test and is passed. Looks like we've got a flaky test in KMS. That's unrelated to this PR so I'll merge it. Nice work on this!

Issue #2503, opened 2019-10-22
Support deletion conflicts for IAM roles and users


csmurton:
Moto does not currently validate whether IAM roles or users have constraints on them which would otherwise cause deletion to fail, e.g. instance profiles, managed and inline policies.

The result of this is that `delete_role` and `delete_user` calls currently succeed as long as the entity has been created, regardless of whether or not there is a constraint which in the real world would prevent the action from completing.

This has been demonstrated below with the following versions:

boto3: 1.9.221
botocore: 1.12.221
moto: 1.3.10

```
@mock_iam
def test_iam_create_delete_role():
    iam = boto3.client('iam', region_name='us-east-1')
    iam.create_role(
        RoleName="test",
        AssumeRolePolicyDocument='''
        {
            "Version": "2012-10-17",
            "Statement": [
                {
                    "Action": "sts:AssumeRole",
                    "Principals": {
                        "Service": "ec2.amazonaws.com"
                    }
                }
            ]
        }
        '''
    )
    iam.put_role_policy(
        RoleName="test",
        PolicyName="testpolicy",
        PolicyDocument='''
        {
            "Version": "2012-10-17",
            "Statement": [
                {
                    "Effect": "Allow",
                    "Action": "s3:ListBucket",
                    "Resource": "arn:aws:s3:::my-bucket"
                }
            ]
        }
        '''
    )
    # Should be caught because of DeleteConflictException, but currently isn't
    with assert_raises(iam.exceptions.DeleteConflictException):
        response = iam.delete_role(RoleName="test")
```

I've opened #2504 which should cover all of the necessary validation to delete_user and delete_role operations and have included updates to the existing tests accordingly.

Please let me know if there are any other supporting changes required :)

spulec:
Closed with https://github.com/spulec/moto/pull/2504

PR #2504, opened 2019-10-22
Conflict detection for IAM delete_role and delete_user

csmurton:
Add validation for delete_role and delete_user to ensure that:

- Any managed policies have been detached
- Any inline policies have been deleted
- The IAM entity has been removed from instance profiles (roles only)

prior to allowing successful deletion. This is being added to match the current behaviour in AWS.

Issue: #2503


mikegrima:
Very nice!

PR #2508, opened 2019-10-23
Refactor dynamodb update documentation

bblommers:
@JackDanger Appreciate your merge yesterday! Good suggestion on how to fix the documentation as well - classic case of updating the method itself, without updating the corresponding docs :)


JackDanger:
awesome :D

PR #2515, opened 2019-10-24
Detach principal from thing when it is deleted

### LINKED ISSUE #2514: IoT Delete Thing should remove the attachment to the principals
# Bug
When deleting a thing in IoT that has some principals attached to it, those attachment go away automatically. This is not the case in moto, so if you try to delete the certificate you will get an error.

# What should happen
In the `delete_thing` function of the backend, you should iterate over the structure `self.principal_things` to remove all that contain the `thing_name`.

# Code to reproduce
```python
client = boto3.client("iot")
client.create_thing(thingName="mything")
certificate = client.create_keys_and_certificate(setAsActive=True)
client.attach_thing_principal(thingName="mything", principal=certificate["certificateArn"])

client.delete_thing(thingName="mything")
client.update_certificate(certificateId=certificate["certificateId"], newStatus="INACTIVE")
client.delete_certificate(certificateId=certificate["certificateId"])
```

EDIT 1: I am working on the fix, I will submit a PR in the next hour

Sytten:
This fixes #2514

### DIFF
--- a/moto/iot/models.py
+++ b/moto/iot/models.py
@@ -331,6 +331,12 @@ class IoTBackend(BaseBackend):
 
         # can raise ResourceNotFoundError
         thing = self.describe_thing(thing_name)
+
+        # detach all principals
+        for k in list(self.principal_things.keys()):
+            if k[1] == thing_name:
+                del self.principal_things[k]
+
         del self.things[thing.arn]
 
     def delete_thing_type(self, thing_type_name):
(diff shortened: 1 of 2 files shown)


Sytten:
I can, but it will throw on the last call if the mutations did not occur anyway.

mikegrima:
I see so if the code was not there then it would fail anyway?

Sytten:
Yes, but I added a check. Its a good practice.

PR #2517, opened 2019-10-25
Fix sns.add_permission & remove_permission

gruebel:
The whole permission handling was wrongly implemented. If you add or remove a permission it changes the access policy of the topic. I also changed the default policy to the correct one and create it with the correct region and account id. Additionally I added more tests and some extra error handling.


mikegrima:
Looks great!

PR #2290, opened 2019-07-10
Run black linter and enforce going forward

spulec:

PR #2518, opened 2019-10-27
Run black linter and enforce going forward

asherf:
Basically a working copy of https://github.com/spulec/moto/pull/2290


Makefile
```diff
@@ -13,9 +13,9 @@ init:
 	@pip install -r requirements-dev.txt
 
 lint:
-	flake8 moto
+	black --check moto/ tests/
 
-test: lint
```
asherf: because we run it via travis and only in python 3.6 and up (black doesn't support older python versions.)

asherf:
@spulec @mikegrima friendly ping... ;-)

spulec:
Thanks for the work on this; it looks great!

My one concern is that this will result in a lot of people making PRs, having tests pass locally, and then Travis fails because of linting. Removing `lint` from `make test` also makes this worse. Any ideas here to give people feedback locally?

asherf:
> Thanks for the work on this; it looks great!
> 
> My one concern is that this will result in a lot of people making PRs, having tests pass locally, and then Travis fails because of linting. Removing `lint` from `make test` also makes this worse. Any ideas here to give people feedback locally?

@spulec how about leaving the test target as is and adding a ‘test-only’ target that will get called from travis.yaml ?

spulec:
That seems like a good plan to me.

Apologies for the merge conflicts that have resulted from me merging some other PRs.

asherf:
@spulec updated.

spulec:
Awesome. Thanks for all of the work!

PR #2519, opened 2019-10-27
Python 3.8 support

yan12125:
Enable tests on Python 3.8 and fix them.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,7 @@ python:
 - 2.7
 - 3.6
 - 3.7
+- 3.8
 env:
 - TEST_SERVER_MODE=false
 - TEST_SERVER_MODE=true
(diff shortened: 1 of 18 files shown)


bblommers:
Exploratory question @mikegrima: would it make more sense to drop support for 2.7? 
- It'lll be EOL in two months
- Based on this survey, much more people use 3.6 than 2.7: https://www.jetbrains.com/research/python-developers-survey-2018/

spulec:
Let's punt on dropping anything for now. I'd like to see if we can get some more data on usage first.

spulec:
This looks great; thank you!

baolsen:
The "make lint" checks only run when Python == 3.7. 
Should this be changed to include checks for 3.8?

PR #2523, opened 2019-10-30
Remove newlines from XML responses

PMende:
This is a fix for [#1870].

This is a second attempt at resolving the issues with producing an XML consistent with what is produced from AWS (i.e., no spaces/new lines between tags). Another attempt (https://github.com/spulec/moto/pull/2205) is currently failing in tests.

This attempt uses precompiled regex patterns as class attributes of the `_TemplateEnvironmentMixin` to remove trailing spaces and newlines after a `">"`, and preceding newlines/spaces before a `"<"`. This *explicitly* wasn't done with a single regex to ensure that even things like `"...<ID>\n      12345\n    </ID>"` are properly collapsed.

### DIFF
--- a/moto/core/responses.py
+++ b/moto/core/responses.py
@@ -83,6 +83,8 @@ class DynamicDictLoader(DictLoader):
 
 
 class _TemplateEnvironmentMixin(object):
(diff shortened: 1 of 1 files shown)


spulec:
This looks great; thanks!

PR #2525, opened 2019-11-01
Add iam account password policy

### LINKED ISSUE #2362: NotImplementedError: update_account_password_policy
I am mocking account management processes using moto and I'm getting the following error:

`The update_account_password_policy action has not been implemented.` 

The code i'm running looks like this: 
``` 
iam_client = get_aws_session("iam","create_account_lambda","OrganizationAccountAccessRole",account,region)
    
###### update the password policy

    
response = iam_client.update_account_password_policy(
        MinimumPasswordLength=100,
        RequireSymbols=True,
        RequireNumbers=True,
        RequireUppercaseCharacters=True,
        RequireLowercaseCharacters=True,
        AllowUsersToChangePassword=True,
        MaxPasswordAge=90,
        PasswordReusePrevention=24,
        HardExpiry=False
)
```

I'm using moto version 1.3.13. 
I can see that in the implementation coverage that it has not yet been implemented. Any idea if or when it will be?

gruebel:
Fixes #2362

### DIFF
--- a/IMPLEMENTATION_COVERAGE.md
+++ b/IMPLEMENTATION_COVERAGE.md
@@ -3387,7 +3387,7 @@
 - [X] untag_role
 - [ ] untag_user
 - [X] update_access_key
-- [ ] update_account_password_policy
+- [X] update_account_password_policy
 - [ ] update_assume_role_policy
 - [ ] update_group
 - [X] update_login_profile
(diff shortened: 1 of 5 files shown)


spulec:
Awesome; thanks!

PR #2530, opened 2019-11-04
Fix deprecations in travis.yaml

asherf:
https://docs.travis-ci.com/user/reference/overview/#deprecated-virtualization-environments
https://blog.travis-ci.com/2019-08-27-deployment-tooling-dpl-v2-preview-release

> skip_cleanup is now deprecated, and cleanup is false by default. The default used to be true, so you had to opt out using skip_cleanup … and has been used a lot. Cleaning up the working directory from any left over build artifacts only made sense for few providers, so we have changed this default.

Screenshot 2019-11-03 20 12 43

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,5 @@
 dist: xenial
 language: python
-sudo: false
 services:
 - docker
 python:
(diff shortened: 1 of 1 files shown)


spulec:
Thank you!

PR #2533, opened 2019-11-04
Feature - Lambda get by arn and update/get function environment

ianyon:
I added to the lambda get_function the functionality to also accept an arn: [docs](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/lambda.html#Lambda.Client.get_function)
Added test for this case

Also I repaired the the get_function and update_function_configuration because get_function didn't returned the environment variables of the function and the [update_function_configuration](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/lambda.html#Lambda.Client.update_function_configuration) didn't set the environment in the function.
Updated tests to check this


spulec:
Looks great! Thanks and welcome to Moto!

PR #2534, opened 2019-11-04
Add ssm `get_parameter_history` support

william-richard:
Basically, this PR changes how SSM parameters are stored in the backend to keep all previous versions, and then allow access to those versions via a `get_parameter_history` request.  All other endpoints should work exactly the same, grabbing the latest version of the parameter.


spulec:
Thanks!

Issue #2322, opened 2019-07-18
Add support for EventBridge


dmulter:
Please add support for both partner and customer API endpoints for the new [Amazon EventBridge](https://aws.amazon.com/eventbridge/) service. See the [EventBridge boto3 API](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/events.html?highlight=eventbridge) docs for all methods.

spulec:
Setting this as a feature request; thanks!

spulec:
Going to call this closed with https://github.com/spulec/moto/pull/2536

Feel free to open other issues with more specific endpoints.

PR #2536, opened 2019-11-04
Add events event bus

gruebel:
I had to redo the whole permissions handling in events to support Event Buses. I left out the implementation of `NextToken` & `Limit` in `list_event_buses()`, because the PR is already big enough and it doesn't add up any meaningful functionality at the moment.

Adds support for custom event buses, but not for partner ones, because per default an AWS account doesn't have any partner event buses available for usage. Any advice or idea on how to support partner event buses? Add a new decorator `@mock_event_bus(partner="datadog")` or extend existing one `@mock_events(partner_event_bus="datadog")`?

Partially solves #2322

### DIFF
--- a/IMPLEMENTATION_COVERAGE.md
+++ b/IMPLEMENTATION_COVERAGE.md
@@ -2713,9 +2713,9 @@
 - [ ] upgrade_elasticsearch_domain
 
 ## events
-48% implemented
+51% implemented
 - [ ] activate_event_source
-- [ ] create_event_bus
+- [X] create_event_bus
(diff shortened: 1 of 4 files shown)


spulec:
Looks like there is an issue with the Py2.7 tests. If we can get those passing, will be happy to merge!

I'm not super familiar with event buses so I may not be the best person to handle this. Instead of passing a parameter to the decorator, would it be better for the user to add an ActivateEventSource call to their test?

spulec:
Thanks for your help!

PR #2542, opened 2019-11-05
Added Exception for create_policy when policy exists

ianyon:
IAM create_policy should throw an exception if the policy exists: [link](https://docs.aws.amazon.com/IAM/latest/APIReference/API_CreatePolicy.html#API_CreatePolicy_Errors)

Also updated two tests that broke because of this.

### DIFF
--- a/moto/iam/models.py
+++ b/moto/iam/models.py
@@ -735,7 +735,9 @@ class IAMBackend(BaseBackend):
         )
         if policy.arn in self.managed_policies:
             raise EntityAlreadyExists(
-                f"A policy called {policy_name} already exists. Duplicate names are not allowed."
+                "A policy called {} already exists. Duplicate names are not allowed.".format(
+                    policy_name
+                )
             )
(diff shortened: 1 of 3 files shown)


moto/iam/models.py
```diff
@@ -733,6 +733,10 @@ def create_policy(self, description, path, policy_document, policy_name):
         policy = ManagedPolicy(
             policy_name, description=description, document=policy_document, path=path
         )
+        if policy.arn in self.managed_policies:
+            raise EntityAlreadyExists(
+                f"A policy called {policy_name} already exists. Duplicate names are not allowed."
```
gruebel: f-strings were introduced in Python 3.6 for the first time and moto also supports Python 2.7. Please change this line to
`"A policy called {} already exists. Duplicate names are not allowed.".format(policy_name)`

ianyon: Fixed!

PR #2547, opened 2019-11-07
AWS Lambda: Add Role validation when creating functions

### LINKED ISSUE #2546: AWS Lambda - Missing validation for roles when creating functions
When creating a Lambda function, there is no validation whether the role passed in actually exists/is valid.
Every test is set up like this:
```
conn.create_function(
    ....
    Role="test-iam-role",
    ....
```
, circumventing the role-behaviour entirely.

PR to add basic validation will follow shortly

bblommers:
Fixes #2546

Adds three validation steps for the IAM role when creating a Lambda function:
 - Does the role follow the appropriate regex
 - Does the role have the current account ID?
 - Does the role exist?


moto/awslambda/models.py
```diff
@@ -667,6 +669,28 @@ def put_function(self, fn):
         :param fn: Function
         :type fn: LambdaFunction
         """
+        pattern = r"arn:(aws[a-zA-Z-]*)?:iam::(\d{12}):role/?[a-zA-Z_0-9+=,.@\-_/]+"
+        valid_role = re.match(pattern, fn.role)
+        if valid_role:
+            account = valid_role.group(2)
+            if account != ACCOUNT_ID:
+                raise ValueError(
```
mikegrima: Please add the specific exceptions to the corresponding `exceptions.py` file to implement one of the moto exception classes.

moto/awslambda/responses.py
```diff
@@ -214,11 +214,7 @@ def _create_function(self, request, full_url, headers):
         try:
```
mikegrima: You can remove the `try...except` if you use one of the moto exception classes.

tests/test_cloudformation/test_cloudformation_stack_crud.py
```diff
@@ -591,3 +593,12 @@ def test_create_stack_kinesis():
     stack = conn.describe_stacks()[0]
     resources = stack.list_resources()
     assert len(resources) == 1
+
+
+def get_role_name():
+    with mock_iam_deprecated():
```
mikegrima: Don't use the deprecated decorator. Update this code to use boto3.

mikegrima:
It looks like the response codes need to be set for the Exceptions.

bblommers:
Hi @mikegrima  - the tests failed before because Flask didn't know how to deal with ClientErrors.
A lot of tests only verify whether a error has been thrown - not whether it's the right kind of error. That's because, before, Flask would encounter a ClientError and throw a 500 - Server error message. The actual error message would be lost.

The try-catch I've introduced fixes this - it will catch any errors, and return them in the same format as AWS does.

Tests are still failing for IOT - it's complaining about an invalid regex. Not sure why! Will continue to  look into it.

bblommers:
Hi @mikegrima - I had to revert back to the deprecated IAM, as the test was failing.
It looks like the deprecated CF doesn't talk to the server (when in server mode), but does everything in memory - so when IAM does create the role on the server, it cant be found by the in-memory CF.

PR #2552, opened 2019-11-11
IAM - Validate duplicate role names

bblommers:
Adds a check when creating a role, to make sure a role with that name doesn't already exist.

Now matches AWS behaviour.

Edit: Added the same validation when creating an IAM policy

### DIFF
--- a/moto/iam/models.py
+++ b/moto/iam/models.py
@@ -898,6 +898,10 @@ class IAMBackend(BaseBackend):
                     permissions_boundary
                 ),
             )
+        if [role for role in self.get_roles() if role.name == role_name]:
+            raise EntityAlreadyExists(
+                "Role with name {0} already exists.".format(role_name)
+            )
 
         clean_tags = self._tag_verification(tags)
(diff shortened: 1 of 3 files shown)


moto/iam/models.py
```diff
@@ -892,6 +898,10 @@ def create_role(
                     permissions_boundary
                 ),
             )
+        if [role for role in self.get_roles() if role.name == role_name]:
```
mikegrima: `self.roles` is a `dict` indexed by role name. Please rely on that instead of getting a list of roles that need to be O(n) searched.

bblommers: _self.roles_ is indexed by role_id! Which is a randomly generated id, so that wouldn't work unfortunately.
The _get_role(role_name)_-function has the same conundrum - it has to search the entire list

bblommers: @mikegrima

mikegrima: 😄Well that's unfortunate.

PR #2554, opened 2019-11-11
Match AWS's change from smart-quote to ASCII quote

aacampbell:
Sometime since I made the PR to represent the smart-quotes in the error messages sent by AWS secretsmanager AWS appear to have changed them to regular old quotes. This PR matches that change in order to make moto representative of actual behaviour once again.


aacampbell:
Can we get this merged quickly? Now that 1.3.14 is out the behaviour is incorrect.

PR #2404, opened 2019-09-02
Adds basic implementations for vpc-classic-link and vpc-classic-link-dns-support

wesselvdv:
Implements the following routes:

* describe-vpc-classic-link
* enable-vpc-classic-link
* disable-vpc-classic-link
* describe-vpc-classic-link-dns-support
* enable-vpc-classic-link-dns-support
* disable-vpc-classic-link-dns-support

PR #2555, opened 2019-11-12
Adds basic implementations for vpc-classic-link and vpc-classic-link-dns-support - Redo

catherinetcai:
Update to https://github.com/spulec/moto/pull/2404, didn't want to step over his branch

### DIFF
--- a/moto/ec2/models.py
+++ b/moto/ec2/models.py
@@ -2433,9 +2433,9 @@ class VPC(TaggedEC2Resource):
         self.dhcp_options = None
         self.state = "available"
         self.instance_tenancy = instance_tenancy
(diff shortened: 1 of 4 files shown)


moto/ec2/models.py
```diff
@@ -2435,6 +2435,7 @@ def __init__(
         self.instance_tenancy = instance_tenancy
         self.is_default = "true" if is_default else "false"
         self.enable_dns_support = "true"
+        self.classic_link_enabled = "false"
```
mikegrima: Strings and not booleans?

catherinetcai: I think this is just following in the style of the other attributes on this class.

PR #2573, opened 2019-11-16
Add the ssh_public_key methods that were missing from IAM

wr0ngway:
Needed these for some testing that I was doing, so just tried to follow the patterns that were there for the similar access_key capabilities.

I didn't do any validation of keys, or conversion (PEM vs SSH encoded) for the get method as I'd probably have to pull in some crypto lib to do so, and wasn't sure what the preference would be there

### DIFF
--- a/moto/iam/models.py
+++ b/moto/iam/models.py
@@ -1,5 +1,6 @@
 from __future__ import unicode_literals
 import base64
+import hashlib
 import os
 import random
 import string
@@ -475,6 +476,27 @@ class AccessKey(BaseModel):
         raise UnformattedGetAttTemplateException()
 
 
+class SshPublicKey(BaseModel):
+    def __init__(self, user_name, ssh_public_key_body):
+        self.user_name = user_name
+        self.ssh_public_key_body = ssh_public_key_body
+        self.ssh_public_key_id = "APKA" + random_access_key()
+        self.fingerprint = hashlib.md5(ssh_public_key_body.encode()).hexdigest()
+        self.status = "Active"
+        self.upload_date = datetime.utcnow()
+
+    @property
+    def uploaded_iso_8601(self):
+        return iso_8601_datetime_without_milliseconds(self.upload_date)
+
+    def get_cfn_attribute(self, attribute_name):
+        from moto.cloudformation.exceptions import UnformattedGetAttTemplateException
+
+        if attribute_name == "PublicKeyBody":
+            return self.public_key_body
+        raise UnformattedGetAttTemplateException()
+
+
 class Group(BaseModel):
     def __init__(self, name, path="/"):
         self.name = name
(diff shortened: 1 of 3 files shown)


moto/iam/models.py
```diff
@@ -475,6 +476,27 @@ def get_cfn_attribute(self, attribute_name):
...
+        self.upload_date = datetime.utcnow()
+
+    @property
+    def uploaded_iso_8601(self):
+        return iso_8601_datetime_without_milliseconds(self.upload_date)
+
+    def get_cfn_attribute(self, attribute_name):
+        from moto.cloudformation.exceptions import UnformattedGetAttTemplateException
+
+        if attribute_name == "PublicKeyBody":
```
wr0ngway: Wasn't sure what I should do here, so just made a guess

mikegrima: I guess the main question is why is this raising a CloudFormation exception?

wr0ngway: No idea - I don't use CF, and I'm not sure how moto is using it, or what/where the get_cfn_attribute 
 is used, and what it is supposed to deliver...?  Any pointers as to where to look?

Looking at the CF docs for [IAM::AccessKey](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-iam-accesskey.html), I see the attribute it is defining, but don't see [any CF docs for the ssh public key in IAM](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/AWS_IAM.html).  Should I just remove this entire method?

mikegrima: Was this code not added by your Pull Request?

wr0ngway: It was, but was uncertain as to what it was for, so made a guess and called it out in my comment above.  I probably should have been more explicit in asking for help there :)

I'm thinking I should just remove it since CF doesn't have support for ssh public keys either.

Should I also change the superclass of SshPublicKey to be object and not BaseModel?  I'm not sure what subclassing from BaseModel implies.  The aws ssh public key is always associated with a user, and doesn't have a resource/model in AWS api, only client methods.

mikegrima: `BaseModel` is what all moto objects use. It has a number of attributes and functions used by the backend.

Delete this function and I will merge it in.

wr0ngway: Ok, done, thanks!

mikegrima: The tests are failing due to a reason un-related to your PR. 

I am going to investigate. I will merge this in after I get that resolved.

mikegrima:
Can you do me a favor and rebase from the latest master?

1. Add the upstream: `git add upstream git@github.com:spulec/moto.git`
2. Fetch: `git fetch --all`
3. Rebase: `git rebase upstream/master`
4. Force push your branch `git push origin add_iam_ssh_public_key_methods --force`

This should help address the broken tests.

wr0ngway:
> Can you do me a favor and rebase from the latest master?

Done

PR #2576, opened 2019-11-17
Add organizations tags

### LINKED ISSUE #2479: Add support  for organizations API to tag_resource and list_tag_for_resources
Hi
Thank you for creating this amazing library.

Currently I'm working creating tools for managing aws organizations and account(organizations api), unfortunately not all functions is covered yet [IMPLEMENTATION_COVERAGE](https://github.com/spulec/moto/blob/master/IMPLEMENTATION_COVERAGE.md),
I would like to mockout these functions:
- `tag_resource`
- `list_tag_for_resources`

I'm currently working on this issue in my local  and will create PR for this feature, might need your help and guidance.

gruebel:
Fixes #2479

### DIFF
--- a/moto/organizations/exceptions.py
+++ b/moto/organizations/exceptions.py
@@ -1,15 +1,12 @@
 from __future__ import unicode_literals
-
-import json
-from werkzeug.exceptions import BadRequest
+from moto.core.exceptions import JsonRESTError
 
 
-class InvalidInputError(BadRequest):
+class InvalidInputException(JsonRESTError):
+    code = 400
+
     def __init__(self):
-        super(InvalidInputError, self).__init__()
-        self.description = json.dumps(
-            {
-                "message": "You provided a value that does not match the required pattern.",
-                "__type": "InvalidInputException",
-            }
+        super(InvalidInputException, self).__init__(
+            "InvalidInputException",
+            "You provided a value that does not match the required pattern.",
         )
--- a/moto/organizations/exceptions.py
+++ b/moto/organizations/exceptions.py
@@ -1,0 +1,15 @@
+from __future__ import unicode_literals
+
+import json
+from werkzeug.exceptions import BadRequest
+
+
+class InvalidInputError(BadRequest):
+    def __init__(self):
+        super(InvalidInputError, self).__init__()
+        self.description = json.dumps(
+            {
+                "message": "You provided a value that does not match the required pattern.",
+                "__type": "InvalidInputException",
+            }
+        )
--- a/moto/organizations/models.py
+++ b/moto/organizations/models.py
@@ -455,5 +455,17 @@ class OrganizationsBackend(BaseBackend):
         new_tags = {tag["Key"]: tag["Value"] for tag in kwargs["Tags"]}
         account.tags.update(new_tags)
 
+    def list_tags_for_resource(self, **kwargs):
+        account = next((a for a in self.accounts if a.id == kwargs["ResourceId"]), None)
+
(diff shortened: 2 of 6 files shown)


moto/organizations/models.py
```diff
@@ -442,5 +443,41 @@ def list_targets_for_policy(self, **kwargs):
         ]
         return dict(Targets=objects)
 
+    def tag_resource(self, **kwargs):
+        account = next((a for a in self.accounts if a.id == kwargs["ResourceId"]), None)
+
+        if account is None:
+            raise RESTError(
```
mikegrima: Please define this as a class in `exceptions.py`.

moto/organizations/models.py
```diff
@@ -442,5 +443,41 @@ def list_targets_for_policy(self, **kwargs):
...
+            )
+
+        new_tags = {tag["Key"]: tag["Value"] for tag in kwargs["Tags"]}
+        account.tags.update(new_tags)
+
+    def list_tags_for_resource(self, **kwargs):
+        account = next((a for a in self.accounts if a.id == kwargs["ResourceId"]), None)
+
+        if account is None:
+            raise RESTError(
```
mikegrima: Ditto.

moto/organizations/models.py
```diff
@@ -442,5 +443,41 @@ def list_targets_for_policy(self, **kwargs):
...
+            )
+
+        tags = [{"Key": key, "Value": value} for key, value in account.tags.items()]
+        return dict(Tags=tags)
+
+    def untag_resource(self, **kwargs):
+        account = next((a for a in self.accounts if a.id == kwargs["ResourceId"]), None)
+
+        if account is None:
+            raise RESTError(
```
mikegrima: This also.

moto/organizations/exceptions.py
```diff
@@ -0,0 +1,15 @@
+from __future__ import unicode_literals
+
+import json
+from werkzeug.exceptions import BadRequest
+
+
+class InvalidInputError(BadRequest):
```
mikegrima: This should be subclassed from the moto `JsonRESTError`.

Please see https://github.com/spulec/moto/blob/master/moto/config/exceptions.py#L2 for an example.

tests/test_organizations/test_organizations_boto3.py
```diff
@@ -671,9 +671,9 @@ def test_list_tags_for_resource_errors():
         client.list_tags_for_resource(ResourceId="000000000000")
     ex = e.exception
     ex.operation_name.should.equal("ListTagsForResource")
-    ex.response["Error"]["Code"].should.equal("400")
-    ex.response["Error"]["Message"].should.contain("InvalidInputException")
-    ex.response["Error"]["Message"].should.contain(
+    ex.response["ResponseMetadata"]["HTTPStatusCode"].should.equal(400)
```
mikegrima: The old code will work when using the moto exception classes.

moto/organizations/exceptions.py
gruebel: ah ok, I just made it similar to the Athena one, but no problem to change it.

mikegrima: Oh! I would imagine that the athena one is broken then (since it wouldn't return the proper exception back to boto).

gruebel:
I also had to fix failing tests for DynamoDB, because they added a new required parameter, which is related to streams.

mikegrima:
Sorry -- totally forgot about this -- merging.

PR #2602, opened 2019-11-26
Add familyPrefix option to ecs:ListTaskDefinitions

atareshawty:
AWS defines this option as:
```
--family-prefix (string)
  The full family name with which to filter the ListTaskDefinitions results.
  Specifying a familyPrefix limits the listed task definitions to task definition revisions that belong to that family.
```

This option behaves differently than ecs:ListTaskDefinitionFamilies. Instead of doing a comparison like `startswith`, it does a full string comparison by matching the entire task definition family to the prefix. For example, let's say there exists a task definition with the family `super-cool-task-def`.

ListTaskDefinitionFamilies would look like this:

```
aws ecs list-task-definition-families --family-prefix super-cool
{
    "families": [
      "super-cool-task-def"
    ]
}
```

ListTaskDefinitions would look like this:

```
aws ecs list-task-definitions --family-prefix super-cool
{
    "taskDefinitionArns": []
}
```

### DIFF
--- a/tests/test_ecs/test_ecs_boto3.py
+++ b/tests/test_ecs/test_ecs_boto3.py
@@ -172,6 +172,69 @@ def test_list_task_definitions():
 
 
 @mock_ecs
+def test_list_task_definitions_with_family_prefix():
+    client = boto3.client("ecs", region_name="us-east-1")
+    _ = client.register_task_definition(
+        family="test_ecs_task_a",
+        containerDefinitions=[
+            {
+                "name": "hello_world",
+                "image": "docker/hello-world:latest",
+                "cpu": 1024,
+                "memory": 400,
+                "essential": True,
+                "environment": [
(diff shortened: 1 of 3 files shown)


tests/test_ecs/test_ecs_boto3.py
```diff
@@ -1756,7 +1819,7 @@ def test_update_task_definition_family_through_cloudformation_should_trigger_a_r
     cfn_conn.update_stack(StackName="test_stack", TemplateBody=template2_json)
 
     ecs_conn = boto3.client("ecs", region_name="us-west-1")
-    resp = ecs_conn.list_task_definitions(familyPrefix="testTaskDefinition")
+    resp = ecs_conn.list_task_definitions(familyPrefix="testTaskDefinition2")
```
atareshawty: Adding the filter broke this test due since it filters now. I needed to change the prefix value here since it does a full string match on the family

atareshawty:
Thank you!

Issue #2537, opened 2019-11-04
The second call to `apigateway.create_api_key` with the same key name should respond with a  `ConflictError`


rawrgulmuffins:
# Simple Problem Statement
The second add of a api-gateway key should raise a ConflictException on a call to`create_api_key` to match the actual API Gateway behavior

# Expectation
The second add of a api-gateway key should raise a ConflictException on a call to`create_api_key` to match the actual API Gateway behavior

# Versions
Python==3.7.3
moto==1.3.13

> Tell us whether you're using standalone server mode or the Python mocks

Python Mocks

# Problem Statement Expanded

If you create a lambda with the following code
```
def example_add_client_key():
    try:
        key_name = f'static_test_example'
        xak = xak_encode(key_name)
        response = CLIENT_API.create_api_key(
             name=name,
             description=f'auto-generated customer api key',
             enabled=True,
             generateDistinctId=True,
             value=xak)
    except botocore.exceptions.ClientError as err:
         ecode = err.response['Error']['Code']
         elif ecode in ['ConflictException']:
             # The x-api-key already exists.
             # Grab it and return.
             print("Hit ConflictException")

def handler(event: dict, context: dict) -> dict:
    example_add_client_key()
```

And then call the API gateway endpoint with this lambda twice in a row you'll hit the conflict exception

![conflict_error_on_api_keys_with_same_name](https://user-images.githubusercontent.com/1521575/68156247-50ab5780-ff00-11e9-9f70-d497c8995809.png)

# Mock Test Case
If you create a test case with

```
@mock_apigateway
@pytest.mark.unit
def test_retrieve_api_key_by_value_on_create_conflict():
    gateway_client = boto3.client(
        'apigateway',
        region_name=os.environ['SECRET_REGION'])
    key_name = 'test-key'
    xak = xak_encode(key_name)

    response = gateway_client.create_api_key(
        name=key_name,
        description=f'auto-generated customer api key',
        enabled=True,
        generateDistinctId=True,
        value=xak)

    # Add second key that'll cause a ConflictError
    response = gateway_client.create_api_key(
        name=key_name,
        description=f'auto-generated customer api key',
        enabled=True,
        generateDistinctId=True,
        value=xak + "a")
    print(response)
    # Returns
[...]

PR #2606, opened 2019-11-27
api gateway handle duplicate api key values

jzivanov:
FIX handling duplicated api keys values on api gateway

Closes: #2537


rawrgulmuffins:
Thank you for the fix @jzivanov.

PR #2610, opened 2019-11-30
Add missing attributes to SNS subscription

### LINKED ISSUE #2441: 'attributes' is empty for SNS subscriptions

gruebel:
Adds missing attributes to the subscription and I also updated the default EffectiveDeliveryPolicy.

I got one question regarding the code in [moto/sns/response.py](https://github.com/spulec/moto/blob/3d5d477/moto/sns/responses.py#L101). Most of the methods have the part `if self.request_json:` in it. Is it some kind of legacy code or what is it used for?

Fixes #2441

### DIFF
--- a/moto/sns/models.py
+++ b/moto/sns/models.py
@@ -432,11 +432,18 @@ class SNSBackend(BaseBackend):
         subscription = Subscription(topic, endpoint, protocol)
         attributes = {
             "PendingConfirmation": "false",
+            "ConfirmationWasAuthenticated": "true",
             "Endpoint": endpoint,
             "TopicArn": topic_arn,
             "Protocol": protocol,
             "SubscriptionArn": subscription.arn,
+            "Owner": DEFAULT_ACCOUNT_ID,
+            "RawMessageDelivery": "false",
         }
+
+        if protocol in ["http", "https"]:
+            attributes["EffectiveDeliveryPolicy"] = topic.effective_delivery_policy
+
         subscription.attributes = attributes
         self.subscriptions[subscription.arn] = subscription
         return subscription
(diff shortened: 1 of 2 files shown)


mikegrima:
I'm honestly not sure about the `self.request_join` code. What happens if you remove it, do the tests fail?
The PR looks good as is, so it's ready to merge when you want.

mikegrima:
@gruebel This OK to merge RE: comment above.

gruebel:
@mikegrima sure go ahead. I will make a dedicated PR for removing the `self.request_json` parts. When I ran the tests they were all successful, so let's try it out. My guess it is some legacy part to create json output.

PR #2611, opened 2019-12-01
Add missing information in SSM parameters

Sytten:
We were missing some new information returned by SSM.
A typical message now looks like:
```
{
    "Parameter": {
        "Name": "test",
        "Type": "String",
        "Value": "test",
        "Version": 1,
        "LastModifiedDate": 1575156149.591,
        "ARN": "arn:aws:ssm:us-east-1:1234567890:parameter/test"
    }
}
```

### DIFF
--- a/moto/ssm/utils.py
+++ b/moto/ssm/utils.py
@@ -1,0 +1,4 @@
+def parameter_arn(region, parameter_name):
+    if parameter_name[0] == "/":
+        parameter_name = parameter_name[1:]
+    return "arn:aws:ssm:{0}:1234567890:parameter/{1}".format(region, parameter_name)
--- a/moto/ssm/utils.py
+++ b/moto/ssm/utils.py
@@ -1,4 +1,9 @@
+ACCOUNT_ID = "1234567890"
+
+
 def parameter_arn(region, parameter_name):
     if parameter_name[0] == "/":
         parameter_name = parameter_name[1:]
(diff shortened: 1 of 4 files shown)


moto/ssm/utils.py
```diff
@@ -0,0 +1,4 @@
+def parameter_arn(region, parameter_name):
+    if parameter_name[0] == "/":
+        parameter_name = parameter_name[1:]
+    return "arn:aws:ssm:{0}:1234567890:parameter/{1}".format(region, parameter_name)
```
mikegrima: Please pull out the account ID into a separate constant at the top of the file. This makes it easy to mock out tests where account ID's matter.

Sytten: Hum this is not an util for tests though. I based this code on `secret_arn` of the secrets manager.

mikegrima: You can use it in your tests by doing something along the lines of:
`mock.patch("moto.ssm.models.ACCOUNT_ID", "222222222222")`
^^ This would then return ARNs with the account ID portion set to `222222222222`.

Sytten: Well it is fixed, I have not idea why the tests fail though. Not related to this PR.

Sytten: Make sense! It is fixed, but the tests fail for some reason not related to this PR.

mikegrima: I'll take a look at what happened with they xray tests.

Issue #2618, opened 2019-12-09
Mock of boto3.client("ec2").describe_instance_status Filters Not Working As Expected


rcuza:
## Summary
When I mock out an AWS account with mock ec2 instances in different states (e.g. 2 in "running" and 1 in "stopped), I do not get the expected results when I run `describe_instance_status` using the `Filters` attribute. 

I am using `moto` version '1.0.1', `boto3` version '1.7.9', `botocore` version '1.10.9', `boto` version '2.49.0', and python 3.6.8. Everything was installed via `pip`.

## How to Reproduce the Issue
```
import moto

mock_aws_account = moto.mock_ec2()
mock_aws_account.start()

import boto3

# Set up mock "AWS Account"
c = boto3.client("ec2", region_name="us-west-2")
c.run_instances(ImageId="ami-faab1234", MinCount=3, MaxCount=3)

# stop the first one
first_instance_id = c.describe_instances()["Reservations"][0]["Instances"][0][
    "InstanceId"
]
c.stop_instances(InstanceIds=[first_instance_id])


# Count the number of instances (all, running, stopped)
# via boto3.client().describe_instances()

all_instances_via_describe_instances = c.describe_instances()["Reservations"][0][
    "Instances"
]

all_states_via_describe_instances = [
    i["State"]["Name"] for i in all_instances_via_describe_instances
]

count_describe_instances = {
    "all": len(all_states_via_describe_instances),
    "running": all_states_via_describe_instances.count("running"),
    "stopped": all_states_via_describe_instances.count("stopped"),
}


# Count the number of instances (all, running, stopped)
# via boto3.client().describe_instance_status()

# Count not using `Filters`

all_instances_via_describe_instance_status_no_filter = c.describe_instance_status(
    IncludeAllInstances=True
)["InstanceStatuses"]

all_states_via_describe_instance_status_no_filter = [
    i["InstanceState"]["Name"]
    for i in all_instances_via_describe_instance_status_no_filter
]

count_describe_instance_status_no_filter = {
    "all": len(all_states_via_describe_instance_status_no_filter),
    "running": all_states_via_describe_instance_status_no_filter.count("running"),
    "stopped": all_states_via_describe_instance_status_no_filter.count("stopped"),
}

# Count using `Filters`
filter_for = {
    "running_and_stopped": [
        {"Name": "instance-state-name", "Values": ["running", "stopped"]}
    ],
    "running": [{"Name": "instance-state-name", "Values": ["running"]}],
[...]

PR #2619, opened 2019-12-09
Create Not Implemented Exception

rcuza:
This change is a band aid for #2618. It is supposed to raise an exception if someone attempts to use `Filters` with `describe_instance_status`. Once I understand the `moto` codebase, I will try to make a PR that implements `Filters`.


spulec:
Thanks; this code was integrated with the changes in https://github.com/spulec/moto/pull/2764

PR #2624, opened 2019-12-10
Import ABC from collections.abc for Python 3.9 compatibility.

tirkarthi:
Importing ABC from collections module has been deprecated since Python 3.4 and removed in Python 3.9. I have made the import fix in compat.py to reuse it across other modules . Reference : https://github.com/python/cpython/pull/10596


mikegrima:
Please make the lint fixes.

tirkarthi:
Thanks, fixed the linting errors and CI is green.

PR #2630, opened 2019-12-11
Add missing default value for moto.ec2.models.VPC.classic_link_dns_su…

### LINKED ISSUE #2629: Bad response model for EC2 DescribeVpcClassicLinkDnsSupport
The response to https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_DescribeVpcClassicLinkDnsSupport.html returns an empty string on the wire instead of `true` or `false`:

```
<DescribeVpcClassicLinkDnsSupportResponse xmlns="http://ec2.amazonaws.com/doc/2016-11-15/"><requestId>7a62c442-3484-4f42-9342-6942EXAMPLE</requestId><vpcs><item><vpcId>vpc-aa9e54e7</vpcId><classicLinkDnsSupported></classicLinkDnsSupported></item></vpcs></DescribeVpcClassicLinkDnsSupportResponse>
```

kislyuk:
…pported

Fixes #2629

### DIFF
--- a/moto/ec2/models.py
+++ b/moto/ec2/models.py
@@ -2449,6 +2449,7 @@ class VPC(TaggedEC2Resource):
         self.is_default = "true" if is_default else "false"
         self.enable_dns_support = "true"
         self.classic_link_enabled = "false"
+        self.classic_link_dns_supported = "false"
         # This attribute is set to 'true' only for default VPCs
         # or VPCs created using the wizard of the VPC console
         self.enable_dns_hostnames = "true" if is_default else "false"


kislyuk:
Can you show me how to do that? The test at https://github.com/spulec/moto/blob/d86852bfa0fd79c62b8a5931d12d2c8a25b47833/tests/test_ec2/test_vpcs.py#L797-L804 seems to already assert this, but the problem is that the response on the wire contains an empty string instead of `false`. From what I can gather, it looks like moto (or boto3?) does type coercion loosely enough that the empty string gets coerced into False, but the Go SDK (for example) does not.

mikegrima:
Sounds good -- as long as an existing test is confirming the change, then all is good.

PR #2640, opened 2019-12-15
CodePipeline - implement CRUD endpoints

gruebel:
I also updated the implementation coverage and due to the reInvent there are a lot of changes.

I actually wanted to implement the the tag endpoints, but I think it is better to do it in a second PR :)


mikegrima:
Please rebase off master. I made some fixes to the tests.

Issue #2545, opened 2019-11-07
Mocking the AccountId


fabioasdias:
As far as I could tell, the accountid used in moto is hard-coded in iam.models.

I would very much like to be able to test stuff across several custom accounts (so I can have unified code that behaves differently based on where it is). Is this on the roadmap? Am I missing an existing way of doing this?

This is related to #27 in spirit.

mikegrima:
Unfortunately, this is really hard to solve, as there is effectively 1 dataset for each region, and adding account abstraction would be a big endeavor.

You might be able to fake things by mocking out the account number.

fabioasdias:
Easier said than done, a quick search on the source codes revealed that '123456789012' is hardcoded in about 282 places in 57 files. So mocking moto.iam.models.ACCOUNT_ID (which would be the reasonable place for it, if a central configuration setting isn't adopted - which would be handy for the multiple account support) doesn't affect much.

fabioasdias:
(Matter of fact, shouldn't the AccountId be a string, to allow for leading zeros?)

PR #2641, opened 2019-12-16
Added references to moto.iam.models.ACCOUNT_ID instead of hardcoded id

fabioasdias:
This adds support to mocking the AccountID by replacing all hardcoded '123456789012' that referenced account ids with imports from moto.iam.models.

Related issue #2545

### DIFF
--- a/moto/acm/models.py
+++ b/moto/acm/models.py
@@ -13,7 +13,7 @@ import cryptography.hazmat.primitives.asymmetric.rsa
 from cryptography.hazmat.primitives import serialization, hashes
 from cryptography.hazmat.backends import default_backend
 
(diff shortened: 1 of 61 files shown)


fabioasdias:
I'll fix the linting issue soonish, but I'll hold until master is building to try a new commit....

fabioasdias:
It's easy enough to do, especially since all imports I used are exactly the same. And I was thinking about allowing this to be an environment variable, like MOTO_ACCOUNT_ID. Thoughts?

mikegrima:
Env var support should be fine. But not in `moto.iam`.

fabioasdias:
Done (ish), I'll add a test for this on my backlog (with a new PR)

PR #2648, opened 2019-12-20
Implement List user tags

asherf:


gruebel:
The tests in server mode add more information to the HTTPHeaders, that's the reason for the failures.
Best would be to change the tests to use the `sure` package and ignore the `ResponseMetadata`
```
response = conn.list_user_tags(UserName="jackie-chiles")
response["Tags"].should.equal([{"Key": "Sue-Allen", "Value": "Oh-Henry"}])
response["IsTruncated"].should_not.be.ok
```

PR #2650, opened 2019-12-20
Fix issue when working with dyanmodb stream and complex types.

asherf:
I think this was introduced in https://github.com/spulec/moto/pull/2605 (not a 100% sure thought).
I know this logic works properly in the v1.3.13 and it is busted in v1.3.14.

I have written a test to repo the issue, here is the relevant part of the call stack from the test failure.
```
File "/Users/asher/projects/moto/moto/dynamodb2/models.py", line 620, in add
  self.items.append(StreamRecord(self.table, t, event_name, old, new, seq))
File "/Users/asher/projects/moto/moto/dynamodb2/models.py", line 589, in __init__
  self.record["dynamodb"]["SizeBytes"] = len(json.dumps(self.record["dynamodb"]))
File "/usr/local/Cellar/python/3.7.4_1/Frameworks/Python.framework/Versions/3.7/lib/python3.7/json/__init__.py", line 231, in dumps
  return _default_encoder.encode(obj)
File "/usr/local/Cellar/python/3.7.4_1/Frameworks/Python.framework/Versions/3.7/lib/python3.7/json/encoder.py", line 199, in encode
  chunks = self.iterencode(o, _one_shot=True)
File "/usr/local/Cellar/python/3.7.4_1/Frameworks/Python.framework/Versions/3.7/lib/python3.7/json/encoder.py", line 257, in iterencode
  return _iterencode(o, 0)
File "/usr/local/Cellar/python/3.7.4_1/Frameworks/Python.framework/Versions/3.7/lib/python3.7/json/encoder.py", line 179, in default
  raise TypeError(f'Object of type {o.__class__.__name__} '
TypeError: Object of type DynamoType is not JSON serializable
```

### DIFF
--- a/moto/dynamodb2/models.py
+++ b/moto/dynamodb2/models.py
@@ -586,7 +586,9 @@ class StreamRecord(BaseModel):
             self.record["dynamodb"]["OldImage"] = old_a
 
         # This is a substantial overestimate but it's the easiest to do now
-        self.record["dynamodb"]["SizeBytes"] = len(dynamo_json_dump(self.record["dynamodb"]))
+        self.record["dynamodb"]["SizeBytes"] = len(
+            dynamo_json_dump(self.record["dynamodb"])
+        )
 
     def to_json(self):
         return self.record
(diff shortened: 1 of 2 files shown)


asherf:
@mikegrima @spulec friendly ping ;-)

bblommers:
Hi @asherf - you're probably right, that it was introduced by #2605. Thanks for providing a fix.

Would it make sense to add an assert to the test, checking that the stream item now has the correct size?
That probably makes the test more readable, as it's more clear what the original issue was.

PR #2661, opened 2019-12-27
Add IsTruncated to Route53.list_resource_record_sets

### LINKED ISSUE #2660: [bug] Route53.list_resource_record_sets not returning `IsTruncated`
`boto3` documentation specifies that `list_resource_record_sets` should return:
```python
'IsTruncated': True|False,
```
to indicate if results were truncated.

AWS API always returns this field, whether records are truncated or not. We rely on this field to know if we need to fetch more results, and using `@mock_route53` around this code fails due to missing field.

https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/route53.html#Route53.Client.list_resource_record_sets

deanrock:
We should return `IsTruncated: False`  on `list_resource_record_sets`.

I've tested this, and code relying on this flag works as expected with this PR.

Fixes #2660

### DIFF
--- a/moto/route53/responses.py
+++ b/moto/route53/responses.py
@@ -271,6 +271,7 @@ LIST_RRSET_RESPONSE = """<ListResourceRecordSetsResponse xmlns="https://route53.
       {{ record_set.to_xml() }}
    {% endfor %}
    </ResourceRecordSets>
+   <IsTruncated>false</IsTruncated>
 </ListResourceRecordSetsResponse>"""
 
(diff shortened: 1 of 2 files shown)


mikegrima:
Typically AWS doesn't include this flag unless it's `True`. Is this one of those odd AWS inconsistencies? :)

deanrock:
I guess it depends on the specific service.

According to [service JSON](https://github.com/boto/botocore/blob/develop/botocore/data/route53/2013-04-01/service-2.json#L3368), it's a required field for every route53 `Response` object that supports it.

bblommers:
Thanks for providing a fix @deanrock - merging

PR #2664, opened 2020-01-01
2663 | Add support for ADMIN_USER_PASSWORD_AUTH by using the same flo…

franz-see:
…w as ADMIN_NO_SRP_AUTH


bblommers:
Hi @franz-see - can you add some tests for this use case (i.e.: copy existing ones) ? Cheers

bblommers:
Implemented by #3547

PR #2667, opened 2020-01-03
Made UserName parameter optional for IAM access key functions

### LINKED ISSUE #2666: IAM Operations Require UserName Field, Should be Optional
When calling some IAM functions, Moto requires the `UserName` parameters, even though it is optional (per the [AWS CLI docs](https://docs.aws.amazon.com/cli/latest/reference/iam/index.html)).

Using Moto 1.3.14, calls to all of the following functions return an error when omitting the `UserName` parameter:
[ ] create_access_key
[ ] update_access_key
[ ] list_access_keys
[ ] delete_access_key

To reproduce the issue:

1. Install the `moto[server]` and `awscli-plugin-endpoint` packages via `pip` 
1. Start Moto in standalone server mode
1. Configure the AWS CLI to use Moto's server address as the API endpoint
1. Call the `create-access-key` function from the AWS CLI without a `--user-name` argument. 

```
# aws iam create-access-key

An error occurred (NoSuchEntity) when calling the CreateAccessKey operation: User None not found
```

owenfarrell:
This update adds conditional logic to support IAM access key operations that do not include the (optional) `UserName` parameter.

This change uses the existing `get_access_key_last_used` function to retrieve an IAM user name based on IAM access key.

Fixes #2666

### DIFF
--- a/tests/test_iam/test_iam.py
+++ b/tests/test_iam/test_iam.py
@@ -785,13 +785,26 @@ def test_delete_login_profile():
     conn.delete_login_profile("my-user")
 
 
-@mock_iam()
+@mock_iam
 def test_create_access_key():
     conn = boto3.client("iam", region_name="us-east-1")
     with assert_raises(ClientError):
         conn.create_access_key(UserName="my-user")
     conn.create_user(UserName="my-user")
     access_key = conn.create_access_key(UserName="my-user")["AccessKey"]
+    (
+        datetime.utcnow() - access_key["CreateDate"].replace(tzinfo=None)
+    ).seconds.should.be.within(0, 10)
+    access_key["AccessKeyId"].should.have.length_of(20)
+    access_key["SecretAccessKey"].should.have.length_of(40)
+    assert access_key["AccessKeyId"].startswith("AKIA")
+    conn = boto3.client(
+        "iam",
+        region_name="us-east-1",
+        aws_access_key_id=access_key["AccessKeyId"],
+        aws_secret_access_key=access_key["SecretAccessKey"],
+    )
+    access_key = conn.create_access_key()["AccessKey"]
     (
         datetime.utcnow() - access_key["CreateDate"].replace(tzinfo=None)
     ).seconds.should.be.within(0, 10)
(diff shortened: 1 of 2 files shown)


owenfarrell:
Could someone point me to a working example of where the `get_current_user()` method is called from a response class?

My approach relies on the ability to determine the user based on the context of the request, but when running my unit tests, I always get the fallback (hardcoded) value.

> The Access Key with id 111122223333 cannot be found

bblommers:
Hi @owenfarrell - IAM is not my strongest area, but some things I found:
get_current_user is used in IAM#get_user: https://github.com/spulec/moto/blob/master/moto/iam/responses.py#L451
Which has a couple of tests already: https://github.com/spulec/moto/blob/master/tests/test_iam/test_iam.py#L717

However: it looks like, if UserName is not supplied, it tries to retrieve the access-key from the header/querystring. 
It doesn't look like either is supplied in your test, so I would expecting it to fail. (It should fail with a UserNotFoundException or something similar, instead of ```The Access Key with id 111122223333 cannot be found```, but the fact that it fails is correct IMO

owenfarrell:
@bblommers - Thanks for feedback.

It took me a while to wrap my head around how to explicitly set the access-key header/querystring, but I realized that creating a fresh client (with an explicit access key) for the purposes of my test is a simple and effective approach.

From my perspective, this addresses the issue I raised. Let me know if there are any adjustments you want me to make to the PR.

tests/test_iam/test_iam.py
```diff
@@ -798,41 +798,58 @@ def test_create_access_key():
...
+    access_key = conn.create_access_key()["AccessKey"]
+    (
+        datetime.utcnow() - access_key["CreateDate"].replace(tzinfo=None)
+    ).seconds.should.be.within(0, 10)
+    access_key["AccessKeyId"].should.have.length_of(20)
+    access_key["SecretAccessKey"].should.have.length_of(40)
+    assert access_key["AccessKeyId"].startswith("AKIA")
 
 
-@mock_iam_deprecated()
```
bblommers: Both @mock_iam and mock_iam_deprecated will need to keep working with these changes - can you keep the existing test case, and simply add the new one?

tests/test_iam/test_iam.py
```diff
@@ -798,41 +798,58 @@ def test_create_access_key():
...
+    )
+    response = conn.list_access_keys()
+    assert_equals(
+        sorted(response["AccessKeyMetadata"][0].keys()),
+        sorted(["Status", "CreateDate", "UserName", "AccessKeyId"]),
     )
 
 
-@mock_iam_deprecated()
+@mock_iam
```
bblommers: Same here

owenfarrell:
@bblommers - Done. Re-adding the decorators didn't impact the test results (just waiting on Travis).

owenfarrell:
@bblommers - Got it. I misunderstood the intent of the decorator used in those tests.

There was 1 name collision for testing the deletion of an access key. I followed the pattern used by the delete user test and renamed the existing test method to add a `_deprecated` suffix.

bblommers:
Thank you @owenfarrell! LGTM

PR #2669, opened 2020-01-05
AWS Lambda - Return response only, and skip log/print statements

bblommers:


BenjenJones:
when is this scheduled to be released?
this issue breaks my unit tests, for some time now.

thanks a lot!

PR #2673, opened 2020-01-05
moto/issues/2672 | Modified 'token_use' to return 'id' for an id toke…

franz-see:
…n, and 'access' for an access token

### DIFF
--- a/moto/cognitoidp/models.py
+++ b/moto/cognitoidp/models.py
@@ -108,7 +108,9 @@ class CognitoIdpUserPool(BaseModel):
 
         return user_pool_json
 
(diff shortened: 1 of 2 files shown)


gruebel:
Can you please run the command `black --check moto/ tests/` locally to format your files and then the build will become green.

franz-see:
Thanks @gruebel . i've now pushed another commit fixing the formatting. This is now the output of `black --check moto/ tests/`

```
$ black --check moto tests
All done! ✨ 🍰 ✨
583 files would be left unchanged.
```

bblommers:
LGTM

PR #2676, opened 2020-01-07
Fix #1728 by introducing per-service extras_require

### LINKED ISSUE #1728: Dependencies overhead

slafs:
This is the initial approach of fixing #1728 i.e. resolve the dependencies overhead and allow for installing minimal set of dependencies for a given service (e.g. when using only `s3` backend, install dependencies used by `moto.s3` module only).

Things to discuss/decide:
- [ ] is this approach even feasible for @spulec 
- [ ] do we still need `botocore.awsrequest` and `ConnectionCls` fix in `moto/__init__.py` (why do we need it anyway)?
- [ ] do we want to include ALL services in `extras_require` (in `setup.py`) i.e. even those that don't need any additional requirements (future-proof)?
- [ ] can we trim down more dependencies (i.e. remove any from `install_requires` still, e.g. werkzeug or Jinja2)?

### DIFF
--- a/moto/rds/models.py
+++ b/moto/rds/models.py
@@ -3,7 +3,6 @@ from __future__ import unicode_literals
 import boto.rds
 from jinja2 import Template
 
-from moto.cloudformation.exceptions import UnformattedGetAttTemplateException
 from moto.core import BaseBackend, BaseModel
 from moto.core.utils import get_random_hex
 from moto.ec2.models import ec2_backends
@@ -16,6 +15,10 @@ class Database(BaseModel):
             return self.address
         elif attribute_name == "Endpoint.Port":
             return self.port
+
+        # Local import to avoid circular dependency with cloudformation.parsing
+        from moto.cloudformation.exceptions import UnformattedGetAttTemplateException
+
         raise UnformattedGetAttTemplateException()
 
     @classmethod
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,2 +1,2 @@
 #  Please add requirements to setup.py
--e .
+-e .[all]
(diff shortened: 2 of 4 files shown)


slafs:
@spulec I'd love some feedback/guidance on this approach.

spulec:
Thanks for opening this! I think this approach makes sense.

For questions 2 and 3, I think the answer is no.

RE trim down more dependencies: probably. If you want to do that, happy to accept it, but also think we could merge with something like this.

davidszotten:
@slafs are you still interested? are you planning to update this patch (at least to resolve the conflicts)? (if not i might take a stab)

spulec:
Closing in favor of https://github.com/spulec/moto/pull/3281

Thanks for the help!

PR #2677, opened 2020-01-07
fix: small fixes to get scripts/scaffold.py working

ptdel:
* updated the way some prompt_toolkit resources are imported
* changed regex to match double-quotes in `BACKENDS` of `backends.py`

most of the requirements for the scaffold script aren't in requirements.txt or anything, so I assumed there wasn't any venv typically used to run it.  Was able to generate scaffolding for rekognition to test.

thanks for the awesome mocking library, hope to contribute to coverage now that this works for me

### DIFF
--- a/scripts/scaffold.py
+++ b/scripts/scaffold.py
@@ -20,8 +20,8 @@ import jinja2
 from prompt_toolkit import (
     prompt
 )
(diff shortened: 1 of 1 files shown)


spulec:
Thanks Patrick! Welcome to Moto!

PR #2684, opened 2020-01-10
s3: implement ListParts parameters

### LINKED ISSUE #2680: S3 ListParts API does not respect MaxParts
Hello,

I *think* there may be an issue with the `MaxParts` argument of `s3.list_parts()`. This pseudo-code:
```python
s3.list_parts(
    Bucket=bucket,
    Key=key,
    UploadId=multiPartUploadId,
    PartNumberMarker=0,
    MaxParts=2,
)
```

Should result in only 2 parts.

For the next snippets, I uploaded 4 parts of a file and tried to get uploaded parts by batch of 2.

Enabling DEBUG logging shows:
```
endpoint       D Sending http request: <AWSPreparedRequest stream_output=False, method=GET, url=URL?uploadId=UID&part-number-marker=0&max-parts=2, headers={...}>
```

But the answer is not what I was waiting. In fact, the `MaxParts` had no effect:
```xml
<?xml version="1.0" encoding="UTF-8"?>
<ListPartsResult xmlns="http://s3.amazonaws.com/doc/2006-03-01/">
    <PartNumberMarker>1</PartNumberMarker>
    <NextPartNumberMarker>4</NextPartNumberMarker>
    <MaxParts>4</MaxParts>
    <IsTruncated>false</IsTruncated>
<ListPartsResult>
```

The same code in a real environment works as expected.

BoboTiG:
Fixes #2680.

I added minimal tests, I hope it is correct as-is.

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -58,3 +58,4 @@ Moto is written by Steve Pulec with contributions from:
 * [Craig Anderson](https://github.com/craiga)
 * [Robert Lewis](https://github.com/ralewis85)
 * [Kyle Jones](https://github.com/Kerl1310)
+* [Mickaël Schoentgen](https://github.com/BoboTiG)
--- a/IMPLEMENTATION_COVERAGE.md
+++ b/IMPLEMENTATION_COVERAGE.md
@@ -6225,7 +6225,7 @@
 - [ ] list_object_versions
 - [X] list_objects
 - [X] list_objects_v2
-- [ ] list_parts
+- [X] list_parts
 - [X] put_bucket_accelerate_configuration
 - [X] put_bucket_acl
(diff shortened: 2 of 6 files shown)


BoboTiG:
I don't get why `test_multipart_upload_out_of_order()` fails with my changes. Does someone have a clue?

gruebel:
You changed the behavior of `s3.models.list_parts()`, which results in a different ordering of the multiparts. You have to change this line
https://github.com/spulec/moto/blob/4063f13bb4d36a19e36c0f476bdb58cc3faa9b93/moto/s3/models.py#L301
to:
```
for part_id in self.partlist:
    part = self.parts[part_id]
```

bblommers:
Closing this, as it has now been implemented by #3658

PR #2700, opened 2020-01-22
awslambda CreateFunction: return State=Active

kislyuk:
Terraform aws provider asserts on this field.

See https://docs.aws.amazon.com/lambda/latest/dg/API_CreateFunction.html#SSS-CreateFunction-response-State


bryanlalexander:
@kislyuk We've got a fix for this in #2703 along with a few other fixes we ran into while trying to make moto work with terraform.

kislyuk:
Closing since #2703 was merged

PR #2701, opened 2020-01-23
awslambda: explicitly specify json-file log driver

kislyuk:
This is analogous to #2635.


kislyuk:
@bblommers can you please take a look at this and merge it? It's blocking me from using moto master in my application (which uses a different Docker log driver by default).

PR #2702, opened 2020-01-23
awslambda: return actual result in sync invoke

kislyuk:
Return actual output of the Lambda instead of echoing the input.

### DIFF
--- a/moto/awslambda/models.py
+++ b/moto/awslambda/models.py
@@ -472,7 +472,7 @@ class LambdaFunction(BaseModel):
             payload["result"] = response_headers["x-amz-log-result"]
             result = res.encode("utf-8")
         else:
-            result = json.dumps(payload)
+            result = res
         if errored:
             response_headers["x-amz-function-error"] = "Handled"
 
--- a/tests/test_awslambda/test_lambda.py
+++ b/tests/test_awslambda/test_lambda.py
@@ -148,7 +148,7 @@ def test_invoke_event_function():
         FunctionName="testFunction", InvocationType="Event", Payload=json.dumps(in_data)
     )
     success_result["StatusCode"].should.equal(202)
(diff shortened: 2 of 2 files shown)


bblommers:
Hi @kislyuk, the Boto3 docs tell me that this is the expected response for the invoke-function:
```
{
    'StatusCode': 123,
    'FunctionError': 'string',
    'LogResult': 'string',
    'Payload': StreamingBody(),
    'ExecutedVersion': 'string'
}
```

kislyuk:
@bblommers can you clarify precisely what you mean?

The code in moto erroneously returns the **request payload** instead of the **response payload**. You can verify the correct behavior in the live Lambda API. The docs at https://docs.aws.amazon.com/lambda/latest/dg/API_Invoke.html use the word **Payload** for both the request and the response payload, but clarify the distinction in the description of each.

I will update the failing test.

kislyuk:
Correction, the code in moto erroneously returns **an empty dict** instead of the response payload. So there is unequivocally a bug here.

bblommers:
Hi @kislyuk, my response was based on these docs:
https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/lambda.html#Lambda.Client.invoke
Looks like there's an inconsistency in the documentation.

I haven't tested it against the live API, but I'll take your word for it if you say it returns the response payload only. :) 

Agreed that there's definitely a bug!

kislyuk:
@bblommers if you're asking why moto.awslambda.models.LambdaFunction.invoke() does not return an object that looks like the one you posted, the answer is that invoke() has a special model. The fields other than payload that you see in the response are rendered by the client from response headers, and payload is handled specially.

See also:
https://github.com/spulec/moto/blob/8f05d4aa3d7d2d8b67012bb13446ae8ea9d53e8b/moto/awslambda/responses.py#L165-L177

bblommers:
Understood! LGTM then

PR #2703, opened 2020-01-23
Fixes awslambda policy management

bryanlalexander:
- adds support for multiple policy statements
- adds support for RevisionId parameter
- adds support for lambda/RemovePermission
- adds State parameter to lambda/GetFunctionConfiguration (always returns Active)

Most of these changes were required in order to more closely mimic the AWS API and resolve issues when using moto[server] with terraform.

### DIFF
--- a/moto/awslambda/policy.py
+++ b/moto/awslambda/policy.py
@@ -1,0 +1,145 @@
+from __future__ import unicode_literals
+
+import json
+import uuid
+
+from six import string_types
+
+from moto.awslambda.exceptions import PreconditionFailedException
+
+
+class Policy:
+    def __init__(self, parent):
+        self.revision = str(uuid.uuid4())
+        self.statements = []
+        self.parent = parent
+
+    def __repr__(self):
+        return json.dumps(self.get_policy())
+
+    def wire_format(self):
+        return json.dumps(
+            {
+                "Policy": json.dumps(
+                    {
+                        "Version": "2012-10-17",
+                        "Id": "default",
+                        "Statement": self.statements,
+                    }
+                ),
+                "RevisionId": self.revision,
+            }
+        )
+
+    def get_policy(self):
+        return {
+            "Policy": {
+                "Version": "2012-10-17",
+                "Id": "default",
+                "Statement": self.statements,
+            },
+            "RevisionId": self.revision,
+        }
+
+    # adds the raw JSON statement to the policy
+    def add_statement(self, raw):
+        policy = json.loads(raw, object_hook=self.decode_policy)
+        if len(policy.revision) > 0 and self.revision != policy.revision:
+            raise PreconditionFailedException(
+                "The RevisionId provided does not match the latest RevisionId"
+                " for the Lambda function or alias. Call the GetFunction or the GetAlias API to retrieve"
+                " the latest RevisionId for your resource."
+            )
+        self.statements.append(policy.statements[0])
+        self.revision = str(uuid.uuid4())
+
+    # removes the statement that matches 'sid' from the policy
+    def del_statement(self, sid, revision=""):
+        if len(revision) > 0 and self.revision != revision:
+            raise PreconditionFailedException(
+                "The RevisionId provided does not match the latest RevisionId"
+                " for the Lambda function or alias. Call the GetFunction or the GetAlias API to retrieve"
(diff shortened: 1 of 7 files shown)


brady-gsa:
@bblommers @spulec @JackDanger

Can you review this one as well?

bryanlalexander:
Removed [WIP] from title, PR is ready for initial review

moto/awslambda/policy.py
```diff
@@ -0,0 +1,145 @@
...
+        self.parent = parent
+
+    def __repr__(self):
+        return json.dumps(self.get_policy())
+
+    def wire_format(self):
+        return json.dumps(
+            {
+                "Policy": json.dumps(
+                    {
```
bblommers: Can we re-use the get-policy method here?
It looks like the JSON is exactly the same

tests/test_awslambda/test_policy.py
```diff
@@ -0,0 +1,104 @@
...
+from moto.awslambda.policy import Policy
+
+
+class MockLambdaFunction:
+    def __init__(self, arn):
+        self.function_arn = arn
+        self.policy = None
+
+
+class TC:
```
bblommers: Can you rework the tests to be more inline with the rest of Moto?
Typically we just have separate methods per test case, i.e.:

def test_add_statement():
    # setup
    # add statement
    # assert ...

moto/awslambda/policy.py
bryanlalexander: wire_format returns JSON and the `Policy` must be escaped JSON so it's somewhat different, however, I agree I was looking at that the other day. We'll take a look and see if we can refactor it a bit.

tests/test_awslambda/test_policy.py
bryanlalexander: We'll get it swapped to one execution of all possible policy options instead of a table driven format.

bryanlalexander:
@bblommers @brady-gsa had the cycles and got those fixed, if you wouldn't mind taking another look.

bblommers:
Looks mostly good to me @bryanlalexander. 

One minor niggle: Can you remove the TestPolicy-class? 
The test-method itself can be made global - as long as the name starts with _test_, Nosetests will recognize it.

brady-gsa:
@bblommers we use assertions in our tests so if we remove the class we'd do something like the below. Do you still want us to remove it?
```
tc = unittest.TestCase('__init__')
tc.assertEquals(a, b)
```
We also had trouble running individual tests with nosetest when there was no test class. Can you give a little background why it was decided not to use test classes?

bblommers:
I don't know why this structure was decided, that was well before my time. I'm just trying to keep things consistent :) 

Other tests typically use _sure_ for assertions:
```
import sure # noqa
a.should.equal(b)
```
E.g.: https://github.com/spulec/moto/blob/master/tests/test_awslambda/test_lambda.py#L185

brady-gsa:
@bblommers we can look into using sure. Do you know how to run a single test with nosetest when there is no test class?

bblommers:
@brady-gsa  Out the top of my head you can simply use the file as a command line argument:
```
nosetests -v tests/test_tests/test_awslambda/test_policy.py
nosetests -v tests/test_tests/test_awslambda/test_policy.py:test_method_name
```
Might have to run this in a virtualenv

brady-gsa:
> @brady-gsa Out the top of my head you can simply use the file as a command line argument:
> 
> ```
> nosetests -v tests/test_tests/test_awslambda/test_policy.py
> nosetests -v tests/test_tests/test_awslambda/test_policy.py:test_method_name
> ```
> 
> Might have to run this in a virtualenv

Thanks @bblommers, specifying the file worked fine - I thought we had tried that. I updated the tests to use sure.

One thing that's a little strange about sure is that PyCharm will throw errors such as the below. Do you know of a solution for that?

> "unresolved attribute should for class dict"

bblommers:
```
unresolved attribute should for class dict
```
That probably means that PyCharm hasn't indexed the sure-package, or doesn't know it's part of the project.
```
unused import statement sure
```
That's expected - on import it will do some magic to add the 'should'-attribute to dicts, but it's not referred anywhere after that.
Most tests will have _# noqa_ after the import to ignore that warning.

Thank you for your PR!

PR #2704, opened 2020-01-24
Fix AWS Lambda status code on invoke

### LINKED ISSUE #2642: Invoke Lambda InvocationType RequestResponse not returning 200 HTTPStatusCode
AWS docs stating syncronous ResquestResponse should return a 200
https://docs.aws.amazon.com/lambda/latest/dg/API_Invoke.html#API_Invoke_ResponseSyntax

https://github.com/spulec/moto/blob/8e2e3cec5a5b31ad4b81c24e087b4df6f1a2b279/tests/test_awslambda/test_lambda.py#L117

#2623 is focusing on the return value but also alluded to this discrepancy.

bblommers:
Fixes #2642 

Supercedes https://github.com/spulec/moto/pull/1779

Work done and credits to @repinel - I just fixed the test


repinel:
Thanks, @bblommers 😄

tests/test_awslambda/test_lambda.py
```diff
@@ -151,6 +151,34 @@ def test_invoke_event_function():
     json.loads(success_result["Payload"].read().decode("utf-8")).should.equal({})
 
 
+@mock_lambda
+def test_invoke_dryrun_function():
+    conn = boto3.client("lambda", "us-west-2")
```
Kerl1310: Should we be declaring this in a constructor or up around line 32 so that it can be re-used? Also don't like that the region is hardcoded (as it is in several places in this file) when it's being declared and initialized as `_lambda_region" up top

bblommers: Fair point for the _lambda_region. Didn't realize that it was already declared up top. I've changed it in this test, and have changed it everywhere else in a separate PR

For the constructor - it has to be created inside the mock_lambda decorator for the mock to work, as far as I'm aware, so it's not as easy as making it a global variable

spulec:
Thanks!

kislyuk:
@bblommers I'm running into an issue following this change. Are you sure the `"X-Amz-Invocation-Type"` header is required? Per my reading of the docs, it is not required and the default value is Request-Response.

PR #2716, opened 2020-02-01
Fixed UnboundLocalError in dynamodb2.query when no filtering arguments are passed

krigar1184:


bblommers:
Hi @krigar1184 - welcome to moto, and thanks for your PR!
The code looks good to me - happy to merge after the tests pass.

I think they fail because our DynamoDB-module handles errors slightly different, and we never raise Exceptions explicitly. Have a look at other exceptions in the query-method how they are handled:
https://github.com/spulec/moto/blob/master/moto/dynamodb2/responses.py#L530

krigar1184:
Hi @bblommers, thanks for your help! I fixed error handling as you had asked.

PR #2717, opened 2020-02-02
S3 Feature: Add IllegalLocationConstraintException

bblommers:
When creating an S3 bucket in the non-default region, the LocationConstraint must be specified.
This PR validates and enforces this.

All tests have been updated to add either the default region, or a LocationConstraint


spulec:
Thanks!

PR #2724, opened 2020-02-04
Handle map or list parameters in sns processing

jayudey-vertex:
When utilizing sns the `__type_of_message__` can receive either a list or a map.

map:
- https://github.com/spulec/moto/blob/master/moto/ses/responses.py#L58
- https://github.com/spulec/moto/blob/master/moto/ses/responses.py#L79

list:
- https://github.com/spulec/moto/blob/master/moto/ses/responses.py#L104

### DIFF
--- a/moto/ses/models.py
+++ b/moto/ses/models.py
@@ -148,11 +148,15 @@ class SESBackend(BaseBackend):
     def __type_of_message__(self, destinations):
         """Checks the destination for any special address that could indicate delivery,
         complaint or bounce like in SES simulator"""
-        alladdress = (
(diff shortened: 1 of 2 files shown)


jayudey-vertex:
if you pop my first commit, you should see the new test fail as expected

```
File "/Users/jay.udey/moto/moto/ses/models.py", line 157, in __type_of_message__
    + destinations.get("BccAddresses", [])
AttributeError: 'list' object has no attribute 'get'

```

jayudey-vertex:
This library has been super helpful, thank you!

PR #2668, opened 2020-01-04
Add support for using arns in lambda requests

ianyon:
Lambda requests can use the function name or function arn, so I modified the parsing to parse for arn expressions now that there are functions in moto to search for name or arn

PR #2725, opened 2020-02-05
Fixed issue with Lambda invoke via ARN

mikegrima:
- Fixed an issue where Lambda invokes via an ARN was hitting real AWS.


bblommers:
Related to #2668 , which has similar the same fix for more Lambda actions, but is lacking tests

PR #2733, opened 2020-02-06
Added 'x-amzn-ErrorType' in the return header for missing lambda function

griffint61:
This is so that lambda client get_function() method will raise the same exception that AWS does for a non-existent function.


griffint61:
@bblommers Tests are passing.

bblommers:
Thanks @griffint61!

PR #2738, opened 2020-02-08
EC2 AMI Creation bugfix - Fix AMI owner

### LINKED ISSUE #2732: Potential bug? create image in ec2 mocking, but the owner id of this imaged returned error
Here is my code to mock creating AMI in ec2:

```
@mock_ec2_deprecated
@mock_iam_deprecated
def test_images(self):
    conn = boto.ec2.connect_to_region(CLOUD_REGION,
                                      aws_access_key_id=access_key,
                                      aws_secret_access_key=secret_key)
    reservation = conn.run_instances("ami-1234abcd")
    instance = reservation.instances[0]
    image_id = conn.create_image(instance.id, "test-ami", "this is a test ami")
    amis = conn.get_all_images()
    for ami in amis:
         print ami.id, ami.owner_id
```

Then i found that owner id of image i just created return access key instead of account id.
Am I doing something wrong?

bblommers:
Fixes #2732 

Ensures that the AMI owner always points to the current Account ID
Before, when supplying a valid AccessKey with the request, it would have the AccessKey as the AMI Owner

Validated against EC2 itself


spulec:
Thanks!

PR #2742, opened 2020-02-10
MessageAction for cognito admin_create_user is now handled

### LINKED ISSUE #2741: Cognito admin_create_user resend flow broken
PR #2357 broke the resend flow in admin_create_user.
From APIReference:

>  MessageAction
>     Set to "RESEND" to resend the invitation message to a user that already exists and reset the expiration limit on the user's account. Set to "SUPPRESS" to suppress sending the message. Only one value can be specified.
> 
>     Type: String
>     Valid Values: RESEND | SUPPRESS
>     Required: No

https://docs.aws.amazon.com/cognito-user-identity-pools/latest/APIReference/API_AdminCreateUser.html

As a solution If the MessageAction iis given the reset value, the user should exist in the system and if not UserNotFoundException must be raised to match AWS behaviour

xnegativx:
Fixes #2741

### DIFF
--- a/moto/cognitoidp/models.py
+++ b/moto/cognitoidp/models.py
@@ -564,7 +564,9 @@ class CognitoIdpBackend(BaseBackend):
         user.groups.discard(group)
 
     # User
-    def admin_create_user(self, user_pool_id, username, message_action, temporary_password, attributes):
+    def admin_create_user(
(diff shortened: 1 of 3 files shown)


spulec:
Code looks good. Please fix the linting that failed on Travis and we'll be happy to merge!

PR #2751, opened 2020-02-17
Implemented S3 Account-level public access block.

mikegrima:
- Also added AWS Config listing and fetching support


mikegrima:
This also fixes the broken AWS Lambda test.

mikegrima:
I see that tests are failing because `coverage` was upgraded to 5.0. For some reason, my pinning to version `4.5.4` isn't working so I am going to investigate why.

mikegrima:
Found the issue -- updating.

spulec:
Looks great! Thanks for all of this.

PR #2754, opened 2020-02-17
Allow ports in k8s service urls for s3 mock

gtmanfred:
If there is a port in the host for the request, then this if statement
is not tripped.


spulec:
Thanks!

PR #2755, opened 2020-02-18
DynamoDB - Send item to DDB Stream on update, not just on create

bblommers:
Created DynamoDB items were already send to DynamoDB streams.
This PR also adds supports updates to be send through


spulec:
Looks great; thanks!

PR #2756, opened 2020-02-18
Support default task-priority and Lambda-role in SWF workflow registration

EpicWink:
SWF workflow type now keeps track of the default task-priority and default AWS Lambda role, set at workflow registration.

This requires the test to use `boto3` and `mock_swf` (instead of the existing import of `boto` and `mock_swf_deprecated`), as `boto` doesn't support these parameters in workflow registration.

A future PR will be needed to support decision task priority.


spulec:
Looks great; thank you!

PR #2761, opened 2020-02-20
EMR - Return start time of first step

### LINKED ISSUE #1427: EMR list_step responses don't contain start date in output
There is a bug in the LIST_STEP_TEMPLATE and other EMR templates around the start date for steps: the conditional looks at `ready_datetime` but then outputs `start_datetime`.

This from https://github.com/spulec/moto/blob/896f040fca975865c1d441d90016446c72db3ed4/moto/emr/responses.py#L794-L796

```
            {% if step.ready_datetime is not none %}
            <StartDateTime>{{ step.start_datetime.isoformat() }}</StartDateTime>
            {% endif %}
```

Nothing in the code looks at `step.ready_datetime` that I could see (use create, start, and end), so I'm guessing the conditional is wrong. This exists in a few other templates too.

bblommers:
Fixes #1427 

'Starts' the first step of a cluster, and sets the start time of that step.

### DIFF
--- a/moto/emr/models.py
+++ b/moto/emr/models.py
@@ -86,6 +86,9 @@ class FakeStep(BaseModel):
         self.start_datetime = None
         self.state = state
 
+    def start(self):
+        self.start_datetime = datetime.now(pytz.utc)
+
 
 class FakeCluster(BaseModel):
     def __init__(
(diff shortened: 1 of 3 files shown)


spulec:
Thanks!

ashb:
Good work @bblommers !

PR #2763, opened 2020-02-24
DynamoDB Streams: Bugfix when sending multiple items

bblommers:
Items processed by Lambda were not removed from the stream afterwards, due to the missing return-statement.

Also refactored the log-logic to improve the debugability


spulec:
Thanks!

PR #2767, opened 2020-02-24
Feature: Add test mode support for Config PutEvaluations

### LINKED ISSUE #2765: Config missing support for PutEvaluations with TestMode
Would be nice to test code for custom Config rules with moto if it supported Config's PutEvaluations with TestMode.

Per https://docs.aws.amazon.com/config/latest/APIReference/API_PutEvaluations.html

> TestMode
> Use this parameter to specify a test run for PutEvaluations. You can verify whether your AWS Lambda function will deliver evaluation results to AWS Config. No updates occur to your existing evaluations, and evaluation results are not sent to AWS Config.
> 
> Note
> When TestMode is true, PutEvaluations doesn't require a valid value for the ResultToken parameter, but the value cannot be null.
> 
> Type: Boolean
> 
> Required: No

jrbeilke:
Resolves #2765

### DIFF
--- a/moto/config/exceptions.py
+++ b/moto/config/exceptions.py
@@ -366,3 +366,13 @@ class TooManyResourceKeys(JsonRESTError):
         message = str(message)
 
         super(TooManyResourceKeys, self).__init__("ValidationException", message)
+
+
+class InvalidResultTokenException(JsonRESTError):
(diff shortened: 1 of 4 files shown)


spulec:
Looks great; thanks!

PR #2769, opened 2020-02-25
Return empty task-token on no-task

EpicWink:
To match the [SWF documentation](https://docs.aws.amazon.com/amazonswf/latest/apireference/API_PollForActivityTask.html), an empty task is one where the task-token is the empty string, rather than being a non-existent key


bblommers:
Looks like this has some merge conflicts after merging your other PR  - happy to merge once those are resolved, @EpicWink!

PR #2770, opened 2020-02-26
Changes behavior of delete_parameter when parameter doesn't exist

oanise93:
**Reference Issues/PRs**
N/A

**What does this implement/fix? Explain your changes.**
Currently, the `delete_parameter` function for the ssm client will respond with a `dict` containing a key of` InvalidParameter` which has a value of a list containing the parameter name that was requested to be deleted when a parameter with said name doesn't exist which doesn't match the behavior of `boto3`. 

With this merge request, `delete_parameter` not responds with a 400 error and ParameterNotFound exception is raised. 

**Any other comments?**
N/A

### DIFF
--- a/tests/test_ssm/test_ssm_boto3.py
+++ b/tests/test_ssm/test_ssm_boto3.py
@@ -34,14 +34,12 @@ def test_delete_parameter():
 def test_delete_nonexistent_parameter():
     client = boto3.client("ssm", region_name="us-east-1")
 
-    try:
+    with assert_raises(ClientError) as ex:
         client.delete_parameter(Name="test_noexist")
-        raise RuntimeError("Should of failed")
-    except botocore.exceptions.ClientError as err:
-        err.operation_name.should.equal("DeleteParameter")
-        err.response["Error"]["Message"].should.equal(
-            "Parameter test_noexist not found."
-        )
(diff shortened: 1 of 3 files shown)


tests/test_ssm/test_ssm_boto3.py
```diff
@@ -30,6 +30,20 @@ def test_delete_parameter():
     len(response["Parameters"]).should.equal(0)
 
 
+@mock_ssm
+def test_delete_nonexistent_parameter():
+    client = boto3.client("ssm", region_name="us-east-1")
+
+    try:
```
bblommers: Please use the _assert_raises_ paradigm for tests like this.
You can find a good example here:
https://github.com/spulec/moto/blob/master/tests/test_apigateway/test_apigateway.py#L180

oanise93: Thanks @bblommers for that feedback. I'll incorporate that now.

oanise93:
It looks like the test that are failing for Python 2.7 TEST_SERVER_MODE=false are unrelated to this PR.

bblommers:
Agreed @oanise93 - the test looks unrelated. I'll look into that later, and once that's fixed I'm happy to get this merged

PR #2772, opened 2020-02-27
Bugfix - When mocking URLs, always return the first match

### LINKED ISSUE #2567: Second request to presigned URL fails (1.3.14, after c7147b06)
Hello.

Looks like second request to generated presigned URL (using `requests`) fails.

I was able to find with `git bisect` that the offending change was  c7147b06 in #2294.

It looks like `responses` has a bit of weird logic: if there are multiple matches to the given URL, then the first one is popped out from the internal list. So, with requests to presigned URLs, the first (valid) regular expression is popped out, and then second call fails.

Not sure if this affects other intercepted requests (also in other backends than S3).

Responses code: https://github.com/getsentry/responses/blob/c8a1733b099f4b3c1bb3db3b42710c7cf8168142/responses.py#L580-L587

This patch to existing tests reproduces the issue:

```diff
--- a/tests/test_s3/test_s3.py
+++ b/tests/test_s3/test_s3.py
@@ -1017,8 +1017,9 @@ def test_s3_object_in_public_bucket():
     presigned_url = boto3.client("s3").generate_presigned_url(
         "get_object", params, ExpiresIn=900
     )
-    response = requests.get(presigned_url)
-    assert response.status_code == 200
+    for i in range(1, 10):
+        response = requests.get(presigned_url)
+        assert response.status_code == 200, "Failed on req number {}".format(i)

```

Output:
```
FAIL: tests.test_s3.test_s3.test_s3_object_in_public_bucket
AssertionError: Failed on req number 2
    <Response [400]> = <module 'requests' from ...>.get('https://test-bucket.s3.amazonaws.com/file.txt?AWSAccessKeyId=foobar_key&Signature=Gry%2F
OmBZWksaCaoNGVgRazZVxiI%3D&Expires=1573860668')
```

Spotted this from private project test where we do create multiple presigned `PUT` to same bucket but with different key. Second upload using `requests.put()` fails (using session object actually). This used to work with `moto==1.3.13`.

Versions in use:
```
boto3==1.10.16
botocore==1.13.16
requests==2.22.0
responses==0.10.6
moto==1.3.14         # (or latest git revision)
```

Note: recently botocore switched fully away from vendored requests, but that doesn't seem to affect the result.

bblommers:
Fixes #2567 

Modifies behaviour of the URL matcher to only/always return the first match
Default behaviour is to return subsequent matches for subsequent requests, which leads to https://github.com/spulec/moto/issues/2567
What happens there is:
 - First request matches on the appropriate S3 URL
 - Same request, executed again, will be matched on the subsequent match, which happens to be the catch-all, not-yet-implemented, callback
  
Fix: Always return the first match

Original code/behaviour that I'm overriding can be found here:
https://github.com/getsentry/responses/blob/master/responses.py#L612


spulec:
Awesome; thanks!

Issue #2774, opened 2020-02-28
Meta - build fails for new PRs


bblommers:
Raising a bug here as a source of documentation

As of a few days ago, PRs are failing in Python 2.7:
```
FAIL: test_auth.test_access_denied_with_not_allowing_policy
raise AssertionError(AssertionError(u"given\nX = 'AuthFailure'\n    and\nY = 'AccessDenied'\nX is 'AuthFailure' whereas Y is 'AccessDenied'",))
```

This looks to be an issue since botocore 1.15.8. Manually fixing requirements.txt to 1.15.7 makes the tests pass
[Changes from botocore 1.15.7 > 1.15.8](https://github.com/boto/botocore/commit/484969ee851d8c5907f5284bdb1651c49e641553)

The underlying reason that the test fails, is that the signature no longer matches for ~~this one specific test~~ tests that use the run_instances method. The error is raised here: 
https://github.com/spulec/moto/blob/master/moto/core/access_control.py#L207

Will continue to investigate

bblommers:
Observed change that might break the signature verification:
The body of the run_instances request now contains 'ClientToken' as of 1.15.8

bblommers:
Ran into the issue that a few dependencies dropped support for Python 2, so running a new venv fails atm.
The attached WIP fixes these dependencies

bblommers:
The test that currently fails verifies that:
- a policy exists that grants permission for action A
- The user executes action B
- An AccessDenied is thrown as the policy does not grant permission for B

The failure only appears with:
- Python 2.7
- Auth enabled
- The run_instances action

Slightly cheeky solution:
We can rewrite the test to verify the same conditions using a different action, where the failure does not appear?

We'll still have to fix the underlying issue, but it would unblock the pipeline.

WDYT @spulec ?

spulec:
I'm fine with the cheeky solution for now to unblock the pipeline.

I hadn't realized Mock and others were dropping Py2 support. It may be time for us to start thinking about that as well.

I'm hoping to get some more time tonight, but definitely will have some more time this weekend to help clean a few things up.

bblommers:
Thanks @spulec. The PR with the cheeky solution is ready for review

bblommers:
Looks like it's been fixed again in botocore 1.15.13 - they decided to rollback the idempotencyToken that broke it in the first place:
[...]

PR #2776, opened 2020-02-29
Python 2 dependencies + Test fix

bblommers:
A few dependencies have dropped support for Python2 - this PR fixes these deps to the last known versions where Python 2 still works

Used as a WIP for #2774


spulec:
Thanks!

PR #2740, opened 2020-02-09
DynamoDB update_item: Allow list_append and if_not_exists-functions in one expression

bblommers:
Fixes #2580

PR #2780, opened 2020-03-05
fixing error when _get_default return value that already exists, on _get_appended_list

heyderpd:
Hello,

I found a error on use **if_not_exists** combined with **list_append**.
I know in merge https://github.com/spulec/moto/pull/2740 they tried to correct it.

But now raise a new error, is `Validation Exception` because in this line 454.
If the function **_get_default** dont return the default value will crash on `expression_attribute_values[key]`

This error occurs because if the data in db has the key.
The **_get_default** will not return an key of `expression_attribute_values`. 
In this case the **_get_default** return an `DynamoType` with data value in the db.


heyderpd:
Hello @bblommers 
Right!
I did the test that covers this new case.
Hope this helps.

PR #2787, opened 2020-03-09
SSM: Added support for label_parameter_version and getting labels on …

driskillb43:
This adds support for boto3's label_parameter_version and viewing the labels when calling get_parameter_history.

### DIFF
--- a/moto/ssm/exceptions.py
+++ b/moto/ssm/exceptions.py
@@ -22,6 +22,23 @@ class InvalidFilterValue(JsonRESTError):
     def __init__(self, message):
         super(InvalidFilterValue, self).__init__("InvalidFilterValue", message)
 
+class ParameterNotFound(JsonRESTError):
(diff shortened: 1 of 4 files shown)


driskillb43:
I am not sure I understand why the build is failing: 

Makefile:16: recipe for target 'lint' failed
make: *** [lint] Error 1
The command "if [[ $TRAVIS_PYTHON_VERSION == "3.7" ]]; then make lint; fi" failed and exited with 2 during .

This seems to be outside of anything I changed. Any ideas?

bblommers:
@driskillb43 That just means that the formatting isn't quite right. You can run ```black moto tests``` to fix it, and ```make lint``` to verify

driskillb43:
> @driskillb43 That just means that the formatting isn't quite right. You can run `black moto tests` to fix it, and `make lint` to verify

Thank you!

Issue #657, opened 2016-07-03
boto3 s3.Object().put() always injects text/plain; charset=utf-8' into ContentType


tednology:
Python 3.5.2, Ubuntu 14.04, moto 0.4.23, boto3 1.3.1.

It seems like moto is injecting 'text/plain; charset=utf-8' into the content type of all files when I read them / upload them as below.

``` Python
    def test_direct_s3(self):
        bucket_name = '...'
        local_path = 'core/bucket_template/index.html'
        s3_path = 'testzooga/index.html'
        s3 = boto3.resource('s3', verify=False)

        with open(local_path, 'rb') as _file:
            content_type = mimetypes.guess_type(local_path)
            s3.Object(bucket_name, s3_path).put(
                ContentType=content_type[0],
                Body=_file,
                ACL='public-read'
            )

            print(s3.Object(bucket_name, s3_path).content_type)
            # print(s3.Object(bucket_name, s3_path).get())

        assert 1 == 0

    @mock_s3
    def test_mocked_s3(self):
        my_bucket = s3.Bucket('my-cool-bucket')
        my_bucket.create()
        local_path = 'core/bucket_template/index.html'
        s3_path = 'testzooga/index.html'
        s3 = boto3.resource('s3', verify=False)

        with open(local_path, 'rb') as _file:
            content_type = mimetypes.guess_type(local_path)
            s3.Object(my_bucket.name, s3_path).put(
                ContentType=content_type[0],
                Body=_file,
                ACL='public-read'
            )

            print(s3.Object(my_bucket.name, s3_path).content_type)
            # print(s3.Object(my_bucket.name, s3_path).get())

        assert 1 == 0
```

Will print:
`text/html` during the first test as it writes to the un-mocked S3
`text/plain; charset=utf-8, text/html` during the second test when writing to the mocked S3

spulec:
I don't believe we currently have much work done around content types, but PRs welcome.

edward-smith:
Thanks for the response Steve; currently doing an exploratory port of an app from Python 2 to 3 and this cropped up in our test suite.  

If/when I get a chance I will do some digging to see if I can submit a PR.

taion:
I'm seeing this as well. It only happens with Python 3, not with Python 2.

thanizebra:
i am seeing this in both Python 3 and Python 2

spulec:
Confirmed fixed with https://github.com/spulec/moto/pull/2795

PR #2795, opened 2020-03-10
S3 - Verify content type is set/returned as appropriate

bblommers:
Verifies #657 is fixed, by adding a test case that verifies the content type of an S3 object is no longer returned as 'text/html' by default


spulec:
Thank you!

PR #2799, opened 2020-03-10
Fix UPDATED_NEW return values differences

### LINKED ISSUE #2798: DynamoDB : UPDATED_NEW return value is not equivalent to the real one
Hello,

While using the tool I found out that the return value UPDATED_NEW when doing an UpdateItem queries are not the same than what we retrieve from the database with real calls.

The rules around what to returns are a bit blurry, here is what I have:
* For lists, if items are added / removed return the whole list
* For lists, if items are modified return only the modified items
* As a general rule, we only return items that changed

Do not hesitate to tell me if I missed rules and see the PR for my fix attempt.

Thank you

ImFlog:
This should fix #2798
The general idea was to play all the UpdateItem queries on a real DynamoDb and dump the value in the tests.
I then added a method that recursively checks for changes in the new item. It works but as stated in the issue I may miss some corner cases.

I don't do a lot of python so this may not be the cleanest code you've reviewed, i'll be glad to improve it following advices 😉

### DIFF
--- a/moto/dynamodb2/models.py
+++ b/moto/dynamodb2/models.py
@@ -146,6 +146,9 @@ class DynamoType(object):
     def __eq__(self, other):
         return self.type == other.type and self.value == other.value
 
+    def __ne__(self, other):
+        return self.type != other.type or self.value != other.value
+
     def __lt__(self, other):
         return self.cast_value < other.cast_value
 
--- a/moto/dynamodb2/responses.py
+++ b/moto/dynamodb2/responses.py
@@ -823,7 +823,9 @@ class DynamoHandler(BaseResponse):
                     return changed
                 else:
                     return [
-                        self._build_updated_new_attributes(original[index], changed[index])
+                        self._build_updated_new_attributes(
+                            original[index], changed[index]
+                        )
                         for index in range(len(changed))
                     ]
             elif changed != original:
--- a/moto/dynamodb2/responses.py
+++ b/moto/dynamodb2/responses.py
@@ -1,8 +1,11 @@
 from __future__ import unicode_literals
-import itertools
+
+import copy
 import json
-import six
 import re
+
+import itertools
+import six
 
 from moto.core.responses import BaseResponse
 from moto.core.utils import camelcase_to_underscores, amzn_request_id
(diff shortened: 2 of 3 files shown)


ImFlog:
Hi @bblommers, I just added some new tests on the existing update_item ones. Also I saw an existing test that does exactly what you asked for :
```python
def test_update_return_attributes():
    dynamodb = boto3.client("dynamodb", region_name="us-east-1")

    dynamodb.create_table(
        TableName="moto-test",
        KeySchema=[{"AttributeName": "id", "KeyType": "HASH"}],
        AttributeDefinitions=[{"AttributeName": "id", "AttributeType": "S"}],
        ProvisionedThroughput={"ReadCapacityUnits": 1, "WriteCapacityUnits": 1},
    )

    def update(col, to, rv):
        return dynamodb.update_item(
            TableName="moto-test",
            Key={"id": {"S": "foo"}},
            AttributeUpdates={col: {"Value": {"S": to}, "Action": "PUT"}},
            ReturnValues=rv,
        )

    r = update("col1", "val1", "ALL_NEW")
    assert r["Attributes"] == {"id": {"S": "foo"}, "col1": {"S": "val1"}}

    r = update("col1", "val2", "ALL_OLD")
    assert r["Attributes"] == {"id": {"S": "foo"}, "col1": {"S": "val1"}}

    r = update("col2", "val3", "UPDATED_NEW")
    assert r["Attributes"] == {"col2": {"S": "val3"}}

    r = update("col2", "val4", "UPDATED_OLD")
    assert r["Attributes"] == {"col2": {"S": "val3"}}

    r = update("col1", "val5", "NONE")
    assert r["Attributes"] == {}

    with assert_raises(ClientError) as ex:
        r = update("col1", "val6", "WRONG")
```

So the upsert is covered and overall I think that the majority of the cases are.
I'll be glad to improve it if someone spot a difference during the tool usage :)

ImFlog:
The tests don't pass on python 2.7, I found a really strange behavior while debugging the issue, with two `DynamoType` I end up with 
```python
original == changed # True
original != changed # True
```
And that's why it doesn't pass...

Any idea about what could cause this ?

bblommers:
List ordering in Python2 and 3 is usually different - make sure both lists are sorted before comparing them

ImFlog:
I finally understand what happened, this is because of the fact that __ne__ need to be redefined when __eq__ is redefined (https://docs.python.org/2.7/reference/datamodel.html#object.__ne__) in Python2.

bblommers:
Good find. Thank you!

PR #2802, opened 2020-03-11
Cloudformation - Set IAM role name

### LINKED ISSUE #2746: Roles created in cloudformation not created with proper name
When creating a role through cloudformation, the role name ends up being incorrect. The pattern seems to be, with parameters, role_name as the role name and stack_name as the stack name

stack_name-role_name-some_random_id

however, in AWS, 

The role name would just be role_name

bblommers:
Sets the IAM rolename if specified, otherwise randomly generates one
Docs: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-iam-role.html

Supercedes #1932 
Fixes #2746

### DIFF
--- a/moto/iam/models.py
+++ b/moto/iam/models.py
@@ -331,7 +331,9 @@ class Role(BaseModel):
         cls, resource_name, cloudformation_json, region_name
     ):
         properties = cloudformation_json["Properties"]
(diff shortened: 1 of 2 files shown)


spulec:
Looks great!

PR #2803, opened 2020-03-11
Use access key create date for credential report

zscholl:
This was using the user create date for all keys so the credential report would not distinguish between access keys created at different times.


zscholl:
I'll add some tests for these changes, I need to do some manual testing on my end first

zscholl:
@bblommers I've added some tests to validate the functionality. This would be a bit better if usage of access keys was somehow tracked by moto, but since it isn't I have to use the backend object to modify usage dates.

Let me know your thoughts

zscholl:
> @bblommers I've added some tests to validate the functionality. This would be a bit better if usage of access keys was somehow tracked by moto, but since it isn't I have to use the backend object to modify usage dates.
> 
> Let me know your thoughts

Turns out using the backend doesn't work when testing in server mode so I was unable to write a test when setting the last_used date to something other than `None`

bblommers:
You could check whether the test is run in server mode, and only make those changes/assertions in non-server mode?
Similar to this: https://github.com/spulec/moto/blob/e00af2f73cb7d27c3755f18b2161b9acbd8ca8aa/tests/test_core/test_context_manager.py#L11

PR #2804, opened 2020-03-11
Loosen jinja2 requirement.

asherf:
This allows repos consuming moto to use the latest jinaj2 2.11.x patched version (currently 2.11.1) w/o breaking moto's python 2 support.
See https://github.com/spulec/moto/pull/2776


asherf:
@bblommers

PR #2805, opened 2020-03-11
Fix a misleading error message

hipplej:
AWSEvents.DescribeRule throws an error that references a rule named "test" rather than the specified rule name when a rule with the specified name does not exist. It has been fixed to reference the specified rule name.

### DIFF
--- a/moto/events/responses.py
+++ b/moto/events/responses.py
@@ -62,7 +62,9 @@ class EventsHandler(BaseResponse):
         rule = self.events_backend.describe_rule(name)
 
         if not rule:
-            return self.error("ResourceNotFoundException", "Rule test does not exist.")
+            return self.error(
+                "ResourceNotFoundException", "Rule " + name + " does not exist."
+            )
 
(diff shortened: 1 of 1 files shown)


moto/events/responses.py
```diff
@@ -62,7 +62,7 @@ def describe_rule(self):
         rule = self.events_backend.describe_rule(name)
 
         if not rule:
-            return self.error("ResourceNotFoundException", "Rule test does not exist.")
+            return self.error("ResourceNotFoundException", "Rule " + name + " does not exist.")
```
asherf: `return self.error("ResourceNotFoundException", "Rule {} does not exist.".format(name))`
is more pythonic.

hipplej: I agree, but I felt it was better to keep consistency with the rest of the file.

hipplej:
> Thanks @hipplej!

Happy to pitch in where I can, this project has been super helpful! Thanks to all the contributors making it happen!

PR #2017, opened 2019-01-13
add suport for dynamodb transact_get_items

cm-iwata:
This pull request add suport for DynamoDB transact_get_items

PR #2809, opened 2020-03-12
Feature - DynamoDB: transact_get_items

bblommers:
Implementation by @cm-iwata - I've just taken his PR and rebased it

Supercedes #2017 

Removed an error condition that's not specified in the docs, and updated the max items to 25 (again as per the docs)


spulec:
Thanks!

PR #2818, opened 2020-03-17
Cloudformation - Stack creation time

### LINKED ISSUE #2773: Improvement: Cloudformation responses always have a hardcoded `CreationDate`
Whilst using Localstack (which uses this framework) whenever responses are returned through the CLI the creationdate fields are always a fixed value in the past. eg. the LIST_STACKS_RESPONSE (in moto/cloudformation/models.py) contains the `CreationDate` value `2011-05-23T15:47:44Z`.

Expected behaviour;
A proper timestamp is generated and passed to the response (for instance {{ stack.creation_time }} ). 

Note: Localstack has fixed this issue by replacing the response body as seen in [this pull request](https://github.com/localstack/localstack/pull/2103/files) but a fix in the underlying framework would be the best way to go

bblommers:
Fixes #2773 

Most of the work taken from the PR from @saadgit : #1310

### DIFF
--- a/moto/cloudformation/models.py
+++ b/moto/cloudformation/models.py
@@ -8,6 +8,7 @@ from boto3 import Session
 
 from moto.compat import OrderedDict
 from moto.core import BaseBackend, BaseModel
+from moto.core.utils import iso_8601_datetime_without_milliseconds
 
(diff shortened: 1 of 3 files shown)


spulec:
Thanks!

PR #2825, opened 2020-03-20
EC2 - Add option to add volumes using CloudFormation

### LINKED ISSUE #2699: Volume Size in Describe Instances when using Cloud Formation
The volume size being returned in describe instances is not accurate. I created the ec2 using cloud formation, along with an ebs volume that is mounted on it. The volume I created was specified to be 100 mb but the one when using describe volume is of size 8 mb. To be clear this works fine with boto3.client('ec2').create_volume but not cloud formation

bblommers:
Fixes #2699 

The bug described the behaviour that the device size was not stored, but block devices were not persisted at all using CloudFormation.
A default volume with size 8 was always created, which was what the user was seeing - in reality, this default volume should only be created if no volumes have been specified.

### DIFF
--- a/moto/ec2/models.py
+++ b/moto/ec2/models.py
@@ -556,6 +556,10 @@ class Instance(TaggedEC2Resource, BotoInstance):
             # worst case we'll get IP address exaustion... rarely
             pass
 
+    def add_block_device(self, size, device_path):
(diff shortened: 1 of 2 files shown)


spulec:
Thank you!

PR #2828, opened 2020-03-21
DynamoDB - Implement 1MB item limit

### LINKED ISSUE #2042: Moto doesn't implement default limits on dynamoDB return
According to the [AWS docs](https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_Scan.html) a DynamoDB scan will return data up to the maximum data size limit (1MB) or to the limit if it's set. Currently Moto doesn't do this and returns everything or the limit if set. There are two primary solutions for adding this feature.
1) A quick and dirty fix is to set a default limit. 
2) The correct way is to check the size and return up to 1MB but this can get quite complicated but doable. We could check the sizes of the items as they are being inserted and store the size with the item somehow or check the sizes of the items within the trim method and once we reach the 1 MB limit return.

I have done 1 already but have a few questions and still need to write tests. I could probably do the second option but it would take a big longer.

[Here](https://github.com/shade34321/moto/blob/implement_scan_limits/moto/dynamodb2/models.py#L674-L676) is my rudimentary implementation of the first option. It sets a default value of 1K for the limit and if you try and set the limit to something greater than that it will set it back to 1K. This will mock AWS's behavior of returning a max of "1 MB" regardless of limit. If you're ok with this I can add the tests and put up the PR for this.

I did have some questions around what some of these other scan functions are doing specifically the two below.
https://github.com/shade34321/moto/blob/implement_scan_limits/moto/dynamodb2/models.py#L895-L910
https://github.com/shade34321/moto/blob/implement_scan_limits/moto/dynamodb2/responses.py#L487-L537 

Thanks!

### LINKED ISSUE #2782: DynamoDB query does not cap the size of the results to 1MB
When querying a DynamoDB table, as per the [docs](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Query.Pagination.html), the size of the response should always be capped to 1MB. moto does not appear to do this.

```Python

import sys

import boto3
from moto.dynamodb2 import mock_dynamodb2
from boto3.dynamodb.conditions import Key


# Create table
with mock_dynamodb2():
    my_session = boto3.session.Session()
    my_region = my_session.region_name
    boto3_dynamodb_resource = boto3.resource('dynamodb', my_region)

    table_name = 'populated-mock-table'
    table = boto3_dynamodb_resource.create_table(
        TableName=table_name,
        KeySchema=[
            {
                'AttributeName': 'partition_key',
                'KeyType': 'HASH'
            },
            {
                'AttributeName': 'sort_key',
                'KeyType': 'SORT'
            }
        ],
        AttributeDefinitions=[
            {
                'AttributeName': 'partition_key',
                'AttributeType': 'S'
            },
            {
                'AttributeName': 'sort_key',
                'AttributeType': 'S'
            }
        ],
        ProvisionedThroughput={
            'ReadCapacityUnits': 5,
            'WriteCapacityUnits': 5
        }
    )

    # Populate the table
    num_items = 50000
    items = [{
        'partition_key': f'partition_key_value',
        'sort_key': f'sort_key_value_{i}',
    } for i in range(num_items)]
    with table.batch_writer() as batch:
        for item in items:
            batch.put_item(Item=item)

    response = table.query(
        KeyConditionExpression = Key('partition_key').eq('partition_key_value')
    )
    print(f'Size of items in MB: {sum([sys.getsizeof(item) for item in response["Items"]]) / 1000000}')
```

Expected: The response has a maximum of 1MB of items, and a LastEvaluatedKey parameter to signify the end of the page.

Actual: All items matching the query are returned.

bblommers:
Fixes #2782 
Fixes #2042


bblommers:
TravisCI was down for maintenance when I submitted this, so there was a delay in running the tests.
GitHub doesn't show it (I guess because of the delay), but Travis is happy: https://travis-ci.org/github/spulec/moto/builds/665153537

moto/dynamodb2/models.py
```diff
@@ -285,6 +285,9 @@ def __init__(self, hash_key, hash_key_type, range_key, range_key_type, attrs):
     def __repr__(self):
         return "Item: {0}".format(self.to_json())
 
+    def size(self):
+        return sum([bytesize(key) + value.size() for key, value in self.attrs.items()])
```
pvbouwel, line 289: By having the square brackets you will first materialize a list with the sizes of the keys and values of each item in `self.attrs`. By using round brackets instead you would have a generator which has lazy initialization. Python allows for dropping round brackets if they are immediately nested inside a function so could just drop square brackets.

This is more of a nit since there are limited amount of items in `self.attrs`
```suggestion
        return sum(bytesize(key) + value.size() for key, value in self.attrs.items())
```

moto/dynamodb2/models.py
```diff
@@ -1123,6 +1126,14 @@ def _trim_results(self, results, limit, exclusive_start_key, scanned_index=None)
                     break
 
         last_evaluated_key = None
+        size_limit = 1000000  # DynamoDB has a 1MB size limit
+        item_size = sum([res.size() for res in results])
```
pvbouwel, line 1130: Same here as previous nit; can drop the square brackets.
```suggestion
        item_size = sum(res.size() for res in results)
```

moto/dynamodb2/models.py
```diff
@@ -1123,6 +1126,14 @@ def _trim_results(self, results, limit, exclusive_start_key, scanned_index=None)
                     break
 
         last_evaluated_key = None
+        size_limit = 1000000  # DynamoDB has a 1MB size limit
```
pvbouwel, line 1129: How do we know it is 1000*1000 and not 1024*1024? Was there official documentation or a test done against real DDB?

bblommers: I've just done some testing against a real DDB. Querying the same data gives me 16913 items back in one go, with 60 bytes per item.
16913 * 60 = 1014780 - i.e., some weird number in between 1000000 and 10241024.

I've noticed a similar discrepancy when implementing the 400KB item limit. DDB will happily accept an item with size 402KB, but it will only start complaining around 404/405KB?

Not sure why, but it looks like they have some soft limits where you can go slightly over, but not by much.

tests/test_dynamodb2/test_dynamodb.py
```diff
@@ -4132,3 +4132,42 @@ def test_gsi_verify_negative_number_order():
...
+    ]
+    with table.batch_writer() as batch:
+        for item in items:
+            batch.put_item(Item=item)
+
+    response = table.query(
+        KeyConditionExpression=Key("partition_key").eq("partition_key_val")
+    )
+    # We shouldn't get everything back - the total result set is well over 1MB
+    assert response["Count"] < len(items)
```
gmcrocetti, line 4171: Not quite sure about project's standards so please, consider it as a nitpick.
```suggestion
     len(items).should.be.greater_than(response["Count"])
```

moto/dynamodb2/models.py
```diff
@@ -285,6 +285,9 @@ def __init__(self, hash_key, hash_key_type, range_key, range_key_type, attrs):
     def __repr__(self):
         return "Item: {0}".format(self.to_json())
 
+    def size(self):
+        return sum(bytesize(key) + value.size() for key, value in self.attrs.items())
```
gmcrocetti, line 289: Is it good or a standard practice in this project to comment with a link to the oficial docs ?

moto/dynamodb2/models.py
gmcrocetti: Hi guys ! :). I'm impressed with this behavior 😮 . @bblommers , I would like to test this behavior by myself due some personal/job related projects. Did you try this with your AWS account or with the dockerized version of DDB ?

bblommers:
@gmcrocetti I've tested this with my own AWS account - ran the same test (just add a timer to make sure the table is created) and commented out the mock-decorator

pvbouwel:
With the changes it looks good to me.  I don't know many cases in which people rely on getting exactly the same limit so the `1000000` comment was more of a nit. <speculation mode>I suppose that according to your testing it is a little stricter than `1024*1024` that AWS might have aimed for `1000*1000`  as documentation. So for example they might have had a reason to stick below 1 MB but they also required some reservations and due to that went to document `1000*1000` without strictly enforcing it.</speculation mode>

I suppose as well having it a little bit stricter than the actual is no big concern if it follows the documentation. So if @gmcrocetti has the same findings all looks good to me.

tests/test_dynamodb2/test_dynamodb.py
```diff
@@ -4132,3 +4132,41 @@ def test_gsi_verify_negative_number_order():
...
+    ]
+    with table.batch_writer() as batch:
+        for item in items:
+            batch.put_item(Item=item)
+
+    response = table.query(
+        KeyConditionExpression=Key("partition_key").eq("partition_key_val")
+    )
+    # We shouldn't get everything back - the total result set is well over 1MB
+     len(items).should.be.greater_than(response["Count"])
```
gmcrocetti, line 4171: Looks like there's an extra space here. Think it was my fault with the last suggestion, sorry.

tests/test_dynamodb2/test_dynamodb.py
```diff
@@ -4132,3 +4132,41 @@ def test_gsi_verify_negative_number_order():
...
+        BillingMode="PAY_PER_REQUEST",
+    )
+
+    # Populate the table
+    items = [
+        {
+            "partition_key": "partition_key_val",  # size=30
+            "sort_key": "sort_key_value____" + str(i),  # size=30
+        }
+        for i in range(10000, 29999)
```
gmcrocetti, line 4161: @bblommers , what do you think about decreasing the range in favor of increasing the size of the item, considering the maximum size of partition and sort keys, of course.

gmcrocetti:
> With the changes it looks good to me. I don't know many cases in which people rely on getting exactly the same limit so the `1000000` comment was more of a nit. I suppose that according to your testing it is a little stricter than `1024*1024` that AWS might have aimed for `1000*1000` as documentation. So for example they might have had a reason to stick below 1 MB but they also required some reservations and due to that went to document `1000*1000` without strictly enforcing it.</speculation mode>
> 
> I suppose as well having it a little bit stricter than the actual is no big concern if it follows the documentation. So if @gmcrocetti has the same findings all looks good to me.

Yeap, same findings here. <speculation> Maybe something related with the storage of partition and sorting keys ? </speculation>
I agree, there's no problem for the end user to reduce or increase this number by a margin of some kbytes.

tests/test_dynamodb2/test_dynamodb.py
bblommers: Either will work - I don't have a particular preference

If you're worried about performance: against the real DDB, it took me maybe 10 seconds to write all items, thanks to the batch-writer

spulec:
Thanks!

PR #2812, opened 2020-03-13
Fix apigateway issues with terraform v0.12.21

duongpv7:
Some fixes:
* moto.apigateway.responses.APIGatewayResponse.resource_methods
 -> Add support for "DELETE" method
  (This is an issue when deleting an apigateway method)
* moto.apigateway.responses.APIGatewayResponse.integrations
-> Get "httpMethod" value from method resource, this "httpMethod" was not included in body request

PR #2832, opened 2020-03-22
Minor fixes in API Gateway for Terraform compatibility

whummer:
Minor fixes in API Gateway for Terraform compatibility. Supersedes #2812 . /cc @bblommers 

* Add method to delete API Gateway methods
* Fix setting of API Gateway `integration_http_method` by reading `httpMethod` request body parameter

### DIFF
--- a/moto/apigateway/responses.py
+++ b/moto/apigateway/responses.py
@@ -142,11 +142,11 @@ class APIGatewayResponse(BaseResponse):
         resource_id = url_path_parts[4]
         method_type = url_path_parts[6]
 
-        if self.method == "GET":
+        if self.method == 'GET':
             method = self.backend.get_method(function_id, resource_id, method_type)
             return 200, {}, json.dumps(method)
-        elif self.method == "PUT":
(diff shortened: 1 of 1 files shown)


moto/apigateway/responses.py
```diff
@@ -328,31 +335,38 @@ def integrations(self, request, full_url, headers):
...
                     function_id, resource_id, method_type
                 )
             elif self.method == "PUT":
                 integration_type = self._get_param("type")
                 uri = self._get_param("uri")
-                integration_http_method = self._get_param("httpMethod")
-                creds = self._get_param("credentials")
+                credentials = self._get_param("credentials")
                 request_templates = self._get_param("requestTemplates")
+                method = self.backend.get_method(function_id, resource_id, method_type)
```
bblommers: If I'm reading this correctly, 'method' doesn't do anything - is this just to verify that the method exists? If so, please add a comment

bblommers:
This was all part of #4082, as far as I know, so I'll close this. Let me know if I've missed anything

Issue #2831, opened 2020-03-21
Add Cloudformation support for Events


gmcrocetti:
During a serverless deploy (cloudformation) with a schedule (Events::Rule) the following log was raised: `No Moto CloudFormation support for AWS::Events::Rule`

It's an easy problem to reproduce: use cloudformation to create an Event Rule.
I'm using the latest version of moto (from `moto-ext`).

gmcrocetti:
Hello @bblommers , thanks for your fast reply. I was working in a solution and it looks like our `Rule` model requires a `Name` property. The [docs](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-events-rule.html) doesn't require one as it can be automatically generated. If that's all true, are you okay with a single PR containing those two features (we still need to open a new issue) ?

bblommers:
PR's are always welcome, @gmcrocetti! An implementation with some features missing is better than no implementation :)

bblommers:
Closed with #2833

gmcrocetti:
Hi @bblommers. I was trying this new feature today (after moto-ext release) and looks like the implementation in #2833 is missing a feature:

<img width="941" alt="image" src="https://user-images.githubusercontent.com/24530683/77978045-72eea900-72d7-11ea-9195-3ea0d7674181.png">

I'm able to solve this new problem but right now I'm not very confident if we need something else. Are we missing something else besides `get_cfn_attribute` ?

bblommers:
Hi @gmcrocetti - your example works for me, but I guess there's another CF resource that tries to get the ARN of ```ScheduleSendEmailEventsRuleSchedule1```.
Implementing ```get_cfn_attribute``` on the Rule-object definitely fixes that - looks like the ARN is the only attribute that has to be implemented ([according to the docs](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-events-rule.html))

As far as I can see, the only other property that's missing from  the implementation is ```physical_resource_id```. It's a non-mandatory property, but good practice to implement anyway. We can just use the name of the Rule, similar to how this is implemented in [AWS Lambda](https://github.com/spulec/moto/blob/master/moto/awslambda/models.py#L256).

bblommers:
Smallest test that that I can reproduce this with:
```
@mock_cloudformation
@mock_events
@mock_ec2
def test_stack_events_create_rule_with_target():
    events_template = {
[...]

PR #2833, opened 2020-03-22
Cloudformation support for Events::Rule

gmcrocetti:
@bblommers , related with our ealier discussion in #2831.

I'm not quite sure about the cloudformation tests. Any tips to improve or create new ones ?
I'm also not quite sure about using `resource_name` during stack creation but I think it's enough for now.

### DIFF
--- a/moto/events/models.py
+++ b/moto/events/models.py
@@ -250,10 +250,10 @@ class EventsBackend(BaseBackend):
         return return_obj
 
     def put_rule(self, name, **kwargs):
-        rule = Rule(name, self.region_name, **kwargs)
-        self.rules[rule.name] = rule
-        self.rules_order.append(rule.name)
-        return rule.arn
+        new_rule = Rule(name, self.region_name, **kwargs)
+        self.rules[new_rule.name] = new_rule
+        self.rules_order.append(new_rule.name)
+        return new_rule
 
     def put_targets(self, name, targets):
         rule = self.rules.get(name)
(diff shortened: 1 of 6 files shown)


tests/test_events/test_events.py
```diff
@@ -1,15 +1,16 @@
-from moto.events.models import EventsBackend
-from moto.events import mock_events
 import json
 import random
 import unittest
 
 import boto3
+import sure  # noqa
```
gmcrocetti: Tests were failing due to the lack of this import.

moto/events/models.py
```diff
@@ -55,6 +55,24 @@ def remove_targets(self, ids):
             if index is not None:
                 self.targets.pop(index)
 
+    @classmethod
+    def create_from_cloudformation_json(
+        cls, resource_name, cloudformation_json, region_name
+    ):
+        properties = cloudformation_json["Properties"]
+        event_backend = events_backends[region_name]
+        event_name = properties.get("Name") or resource_name
```
bblommers: Change this to ```properties.get("Name", resource_name)``` please - same in the delete. Slightly more readable IMO

gmcrocetti: I used this "safer" approach to handle the case where `{"Name": ""}`. Is it possible ?

bblommers: That is a very valid point! Didn't think about that possibility. Leave it as is then

gmcrocetti:
> Hi @gmcrocetti, can you add two more tests?
> 
>     * Check what happens if you do not provide a rule name in the CF template
> 
>     * Check whether you can delete a rule via CF. So create a test that:
> 
> 
>     1. Creates a rule via CF
> 
>     2. Deletes a rule via CF
> 
>     3. Verify that no rules exist in the system

Added the required tests. Hope to hear from you soon. Thanks :)

PR #2840, opened 2020-03-25
Extend implementation coverage

bblommers:
Changes:
 - Renames ```@mock_ec2_instance_connect``` to ```@mock_ec2instanceconnect``` (in-line with other services with spaces/dashes)
 - Moves/renames some methods in S3 and SWF in their respective backends to ensure the implementation-coverage script picks up all implemented methods
 - Change the styling to add each service in it's own collapsible section. This makes it slightly more readable IMO: https://github.com/bblommers/moto/blob/extend_impl_cov_doc/IMPLEMENTATION_COVERAGE.md

Let me know what you think. Happy to amend styling (or get rid of it completely, if you prefer the old style!)


bblommers:
@spulec I need to rebase this - will probably start from scratch due to the number of changes that have gone into master since then.

Is the suggested layout with dropdowns acceptable for you?
If not I'll just create a PR with just the other changes

spulec:
Yes, I like the new layout!

PR #2854, opened 2020-03-30
[WIP] Feature: S3 - Set metadata embedded in presigned url

### LINKED ISSUE #2104: Metadata not available when using pre-signed urls
Hi all! 👋

This is a better formulation (hopefully) than #2040. What I do:
- generate a pre-signed url adding Metadata when doing it
- upload data via PUT to that generated url
- retrieve the data just uploaded

Expected: I should be able to see the metadata.
Actual result: No metadata is present.

To reproduce, run this script:

```python
# https://github.com/spulec/moto/issues/2040
import boto3
from moto import mock_s3
import requests


@mock_s3
def test_generate_presigned_url():
    conn = boto3.resource('s3', region_name='us-east-1')
    conn.create_bucket(Bucket='mybucket')

    # Create a pre-signed url with some metadata.
    url = s3.generate_presigned_url(ClientMethod="put_object", Params={
        'Bucket': 'mybucket', 'Key': 'mykey', 'Metadata': {'venue': "123"}})

    # Use the url to upload data.
    requests.put(url, data={'Bucket': 'mybucket',
                            'Key': 'mykey', 'Body': 'myvalue'})

    # Retrieve metadata
    return s3.head_object(Bucket='mybucket', Key='mykey').get('Metadata')


# To show the bug, use test_my_model_head_object() to fill metadata. For the working example, use test_my_model_put_object().
if __name__ == '__main__':
    metadata = test_generate_presigned_url()
    assert 'venue' in metadata
    assert metadata['venue'] == '123'
```

- moto version: 1.3.7

I'll be back with a patch if time permits.

bblommers:
Fixes #2104 

The metadata is part of the querystring when using presigned URLs, so we can simply parse it the same way we parse headers

TODO: Extend the signature to include header-names, and verify that incoming request contains those headers. See comment below

### DIFF
--- a/moto/s3/responses.py
+++ b/moto/s3/responses.py
@@ -1224,6 +1224,7 @@ class ResponseObject(_TemplateEnvironmentMixin, ActionAuthenticatorMixin):
             )
             request.streaming = True
             metadata = metadata_from_headers(request.headers)
+            metadata.update(metadata_from_headers(query))
             new_key.set_metadata(metadata)
             new_key.set_acl(acl)
             new_key.website_redirect_location = request.headers.get(
@@ -1965,6 +1966,7 @@ S3_OBJECT_COPY_RESPONSE = """\
     <ETag>{{ key.etag }}</ETag>
     <LastModified>{{ key.last_modified_ISO8601 }}</LastModified>
 </CopyObjectResult>"""
+
 
 S3_MULTIPART_INITIATE_RESPONSE = """<?xml version="1.0" encoding="UTF-8"?>
 <InitiateMultipartUploadResult xmlns="http://s3.amazonaws.com/doc/2006-03-01/">
--- a/moto/s3/utils.py
+++ b/moto/s3/utils.py
@@ -75,7 +75,11 @@ def metadata_from_headers(headers):
                 # Check for special metadata that doesn't start with x-amz-meta
                 meta_key = header
             if meta_key:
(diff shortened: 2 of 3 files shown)


jvtm:
AFAIK `Metadata` passed to `get_presigned_url()` just affects the generated signature, and adds `x-amz-meta-foo` (where `foo` is the key you set) to `X-Amz-SignedHeaders` query parameter of the URL.

Then, the _client using the URL_ is supposed to fill in that header, and S3 verifies that the signature is intact.

Generating the URL does not assign anything on the server side, and also the values are NOT part of the generated URL. You can verify this by generating a presigned URL against real S3 and checking what it holds.

At least that's how s3v4 signatures seem to work.

This patch would fail that assumption, essentially not allowing moto to be used for negative tests (client failing to add the header).

Note: if adding `ContentMD5` to presigned url creation parameters (in boto3/botocore), then client is supposed to send `Content-MD5: <base64 encoded md5 hash>` header.

Just my few cents. Would love to see moto's presigned URL logic being improved further.

bblommers:
Hi @jvtm, thanks for the explanation! I wasn't aware that the signature also should include the header-names.

Will mark this as WIP - hopefully I'll have some time soon to implement this behaviour.

If anyone else wants to have a go, that's more than welcome as well obviously :)

bblommers:
Closing in favor of #3121

PR #2694, opened 2020-01-15
adds tagging support for cloudwatch events service

bryanlalexander:
The CloudWatch events service lacks tagging support. This causes Terraform to fail when executing against a running moto server.

This pull request updates the following:
- Adds `TaggingService` (a generic implementation of tagging for easing the adoption of tagging support across moto)
- Implements tagging for CloudWatch events using `TaggingService`
- removes duplicate `test_list_rules` in `test_events.py`

PR #2857, opened 2020-04-01
S3 - Extend generic TaggingService to s3

bblommers:
A generic tagging service was introduced in #2694 . 
This PR changes the S3 service to use this tagging service to store tags, instead of the custom implementation it used until now.

Long-term solution (IMO) is to convert all services to use the same tagging service - potentially having a TaggableResource super class so that every model gets tag-behaviour OOTB?


spulec:
Good call; thanks!

PR #2861, opened 2020-04-03
Lambda - Add test to verify remove_permission functionality

### LINKED ISSUE #883: missing features from awslambda mock
In the process of writing tests for some lambda function management code, I've noticed that a number of features are not implemented. Interestingly, there are no exceptions raised by `moto`. Instead, the calls are pushed through to AWS, and time out due to insufficient permissions (thank goodness!). These calls include:

* awslambda.add_permission
* awslambda.remove_permission
* awslambda.update_function_code
* awslambda.update_function_configuration

These were problematic enough for my testing that I didn't go any deeper. A quick look at the code suggests that there are other functions that would be missing as well.

Rounding out the implementation for awslambda is one thing, but the fact that some calls are caught and others are passed through to lambda proper is another. If a user has default credentials set up, they may end up making costly calls to the real service while developing tests. Any chance of a quick solution to that secondary problem?

bblommers:
Fixes #883

All missing features that were mentioned in the ticket (add_permission, remove_permission, update_function_code, update_function_configuration) have already been implemented.

The methods have been renamed so that it's picked up the by implementation coverage script.

The remove_permission-functionality didn't have a test yet, so I added one

### DIFF
--- a/moto/awslambda/models.py
+++ b/moto/awslambda/models.py
@@ -1006,11 +1006,11 @@ class LambdaBackend(BaseBackend):
             return True
         return False
 
-    def add_policy_statement(self, function_name, raw):
+    def add_permission(self, function_name, raw):
         fn = self.get_function(function_name)
         fn.policy.add_statement(raw)
 
-    def del_policy_statement(self, function_name, sid, revision=""):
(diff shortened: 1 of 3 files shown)


spulec:
Looks good; thanks!

Issue #2862, opened 2020-04-03
Can't find targets with instance tags in SSM send_command()


Zullybur:
**moto**: 1.3.15.dev0
**boto**: 2.49.0
**boto3**: 1.12.35
**botocore**: 1.15.35

When using `SSM:send_command` with the `Targets` argument, no targets are found. Additionally when you try to run `SSM:get_command_invocation` on said command using an instance that should have been found (but wasn't in this case), it looks like boto3 is trying to contact AWS instead of being mocked by moto.

I'd expect the following test to pass:
```python
@mock_ec2
@mock_ssm
def test_get_command_invocations_by_instance_tag():
    ec2_client = boto3.client("ec2", region_name="us-east-1")
    tag_specifications = [{
        "ResourceType": "instance",
        "Tags": [{
            "Key": "Name",
            "Value": "test-tag"
        }]
    }]
    ec2_response = ec2_client.run_instances(ImageId="ami-1234abcd", MaxCount=1, MinCount=1,
                                            TagSpecifications=tag_specifications)
    instance_id = ec2_response["Instances"][0]["InstanceId"]

    ssm_client = boto3.client("ssm", region_name="us-east-1")
    ssm_response = ssm_client.send_command(DocumentName="AWS-RunShellScript",
                                           Targets=[{"Key": "tag:Name", "Values": ["test-tag"]}])
    target_count = ssm_response["Command"]["TargetCount"]
    assert target_count == 1

    command_id = ssm_response["Command"]["CommandId"]
    ssm_invocation = ssm_client.get_command_invocation(CommandId=command_id, InstanceId=instance_id)
    status = ssm_invocation["Status"]
    assert status == "Success"
```

Instead I get a huge amount of logging output that fills my terminal buffer, which makes me think maybe boto3 is not being mocked properly somewhere in this workflow e.g.:
```
botocore.loaders: DEBUG: Loading JSON file: /Users/****/.local/share/virtualenvs/moto-b58KUgnk/lib/python3.7/site-packages/botocore/data/endpoints.json
botocore.hooks: DEBUG: Event choose-service-name: calling handler <function handle_service_name_alias at 0x10666ab90>
botocore.loaders: DEBUG: Loading JSON file: /Users/****/.local/share/virtualenvs/moto-b58KUgnk/lib/python3.7/site-packages/botocore/data/xray/2016-04-12/service-2.json
[...]

PR #2863, opened 2020-04-03
[WIP] - SSM - Add test for send_command

Zullybur:
Test for issue #2862 demonstrating ssm.send_command using the Targets
parameter has issues, specifically not finding the targets and not using
the mock when trying to run ssm.get_command_invocation against the command id
and the instance id that should have been found.


spulec:
Changed added with https://github.com/spulec/moto/pull/3443

PR #2864, opened 2020-04-04
CognitoIdentity - Fix Format of Identity ID

### LINKED ISSUE #2800: Cognito Identity identity_id format should be in format `REGION:GUID`
Hi

Running tests using moto revealed issue with the _identity id_ gotten from `cognitoidentity`. In reality aws _identity_ids_ should be `REGION:GUID`.[0] However moto creates an _identity id_ in format `"{0}:{1}".format(region, get_random_hex(length=19))`. [1] This causes issues when being strict with what the format is expected to be.

[0] https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/cognito-identity.html#CognitoIdentity.Client.get_id
[1] https://github.com/spulec/moto/blob/master/moto/cognitoidentity/utils.py

bblommers:
Fixes #2800

### DIFF
--- a/moto/cognitoidentity/utils.py
+++ b/moto/cognitoidentity/utils.py
@@ -1,5 +1,5 @@
-from moto.core.utils import get_random_hex
+from uuid import uuid4
 
 
 def get_random_identity_id(region):
-    return "{0}:{1}".format(region, get_random_hex(length=19))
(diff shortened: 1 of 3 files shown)


spulec:
Thanks!

PR #2866, opened 2020-04-05
Fix duplicated elements in S3 put_object / delete_object response XML structure

dholroyd:
I think most tools just look at the response headers for PutObject, but where the response body is consumed, I suspect that the top-level `<PutObjectResponse>` should not be duplicated.

Same issue for `<DeleteObjectResponse>` nearby, so I altered that too.

### DIFF
--- a/moto/s3/responses.py
+++ b/moto/s3/responses.py
@@ -1868,18 +1868,16 @@ S3_DELETE_KEYS_RESPONSE = """<?xml version="1.0" encoding="UTF-8"?>
 {% endfor %}
 </DeleteResult>"""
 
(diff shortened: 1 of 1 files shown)


dholroyd:
On further inspection, [the botocore service definition](https://github.com/boto/botocore/blob/cbd95b0db2c6b03f35d2db79c94c9fa40bd6ad43/botocore/data/s3/2006-03-01/service-2.json) seems to imply headers-only response for both these operations (no XML body at all) :thinking:

bblommers:
Hi @dholroyd, thanks for providing a PR for this.
You're right on the response - I just tested it, and S3 seems to respond with a headers-only request (no body at all). Will merge this PR for now, as it is an improvement.

PR #2871, opened 2020-04-06
S3 - PutObject and DeleteObject should not have a response body

### LINKED ISSUE #2867: S3 - PutObject/DeleteObject should have no response body
Both the PutObject- and DeleteObject-methods currently have a response body, but they should only return headers.

Originally raised in #2866

bblommers:
Fixes #2867


bblommers:
Closing - superceded by #2875

PR #2878, opened 2020-04-08
Adding functionalities aws apigateway create domain name,get domain names,get domain name

usmangani1:
Adding functionalities aws apigateway create domain name,get domain names,get domain name.

### DIFF
--- a/moto/apigateway/exceptions.py
+++ b/moto/apigateway/exceptions.py
@@ -120,6 +120,16 @@ class ApiKeyAlreadyExists(RESTError):
             "ConflictException", "API Key already exists"
         )
 
+
+class InvalidDomainName(BadRequestException):
+    code = 404
+
+    def __init__(self):
+        super(InvalidDomainName, self).__init__(
+            "BadRequestException", "No Domain Name specified"
+        )
+
+
 class DomainNameNotFound(RESTError):
     code = 404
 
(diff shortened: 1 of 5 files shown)


usmangani1:
> Adding functionalities aws apigateway create domain name,get domain names,get domain name.

@bblommers  I have added another new functionality.Can you do me a favour here as well in case if linting creates any issues ?

usmangani1:
> Hey @usmangani1, I've pushed some linting changes.
> 
> Would you mind fleshing out the tests a bit more?
> Would be good to see:
> 
> * More asserts on the create_domain_name() response
> * A separate test for the get_domain_name(), with a few more asserts
> * A separate test for the get_domain_names(), where we also check what happens if no domain_names have been created yet
> * A test where the DomainNameNotFound is thrown
> * A test where get_domain_name() has a non-existing domain-name as a parameter. Should that throw an error?

Sure,That should not be a problem.I will add more tests.

usmangani1:
> Hey @usmangani1, I've pushed some linting changes.
> 
> Would you mind fleshing out the tests a bit more?
> Would be good to see:
> 
> * More asserts on the create_domain_name() response
> * A separate test for the get_domain_name(), with a few more asserts
> * A separate test for the get_domain_names(), where we also check what happens if no domain_names have been created yet
> * A test where the DomainNameNotFound is thrown
> * A test where get_domain_name() has a non-existing domain-name as a parameter. Should that throw an error?

@bblommers  Done.Please let me know if you fell if something is missing

Issue #2806, opened 2020-03-11
 DynamoDB2: update set incorrectly defaults to string when using attribute


pvbouwel:
This is a bug for moto==1.3.14 .

When using an update expression and setting an attribute `att1` to the value of another attribute `att2` moto will set the attribute to a string equal to the name `att2`.  Whereas AWS DynamoDB will make sure the value is passed on.

Reproduction via tests:
```
import boto3
from moto import mock_dynamodb2

table_name = 'Thread'
forum_name = 'ForumName'
table_definition = {
        "AttributeDefinitions": [
            {
                'AttributeName': forum_name,
                'AttributeType': 'S'
            }
        ],
        "TableName": table_name,
        "KeySchema": [
            {
                'AttributeName': forum_name,
                'KeyType': 'HASH'
            }
        ],
        "BillingMode": 'PAY_PER_REQUEST',
    }
subject1_value = 'my_subject1'
subject2_value = 'my_subject2'
subject1_attr_name = 'subject1'
subject2_attr_name = 'subject2'


def create_load_update_get_delete(deleteTable=False):
    session = boto3.Session(profile_name='default')
    ddb_client = session.client('dynamodb')
    try:
        ddb_client.create_table(**table_definition)
        waiter = ddb_client.get_waiter('table_exists')
        waiter.wait(TableName=table_name)
    except ddb_client.exceptions.ResourceInUseException:
        print("Table exists not creating.")

    ddb_client.put_item(
        TableName=table_name,
        Item={
            forum_name: {'S': 'forum1'},
            subject1_attr_name: {'S': subject1_value},
            subject2_attr_name: {'S': subject2_value}
        }
    )
    ddb_client.update_item(
        TableName=table_name,
        Key={forum_name: {'S': 'forum1'}},
        UpdateExpression=f'SET {subject1_attr_name} = {subject2_attr_name}'
    )
    response = ddb_client.get_item(
        TableName=table_name,
        Key={forum_name: {'S': 'forum1'}}
    )
    items = response.get('Item', {})
    if deleteTable:
        ddb_client.delete_table(TableName=table_name)
    return items


@mock_dynamodb2
def test_mocked():
    # This tests fails since it would actually be equal to subject2_attr_name
    updated_item = create_load_update_get_delete()
    updated_subject1_value = updated_item[subject1_attr_name]['S']
    assert updated_subject1_value == subject2_value


def test_aws():
    # This tests fails since it would actually be equal to subject2_attr_name
[...]

PR #2885, opened 2020-04-11
DDB structured parsing

pvbouwel:
Improve DDB expressions support.

Currently DDB expressions are mostly handled in an adhoc manner. This pull request is the start of going towards a structured approach. In order to action expressions the following steps will be taken:

 1) Expression gets parsed into a tokenlist (tokenized)
 2) Tokenlist get transformed to expression tree (AST) -> This commit
 3) The AST gets validated (full semantic correctness)
 4) AST gets processed to perform the update

This pull request provides 1 & 2 for the update expression. Code for each point resides in a commit of its own to ease review.  Publishing this PR as review of this part can already be done and at this stage it will already provide benefit as it will be able to throw Validation Exceptions that now silently pass and will be raised when running against AWS DynamoDB.

This approach should both work for the UpdateExpression and the ConditionExpression.  The ConditionExpression is already more structured and closer to this approach but can be converted further.

Going for the structured approach should allow easier bug fixing and adding of enhancements since each code part has clearly identified responsibilities. In particular I started this work to tackle https://github.com/spulec/moto/issues/2806 .


pvbouwel:
I will add user facing tests.  It seems like my second commit broke Py27 compatibility. I wanted to check what precisely broke but when I go to job log in travis CI I always seem to get `There was an error while trying to fetch the log.`.  Not sure if that is the right way to retrieve why that build failed. Is there a way for me to retrieve the errors of that build?

bblommers:
No, that should be the right way! Not sure why it's not showing anything.
You might have to run it locally.

pvbouwel:
The extra commits add the validation of the AST.  It caused some tests to fail. Checked against AWS DynamoDB and the test were invalid.  I've changed them to pass against AWS that is in the separate commit e6b51a2. I had to add a small commit as well that makes sure setup.py packages reserved_keywords.txt in order to make sure this file is available as a data file.

Only step remaining to switch fully to the structured parsing is implementing the executor which will take the validated AST and executes all the actions of the UpdateExpression. When enabling that it will be possible to perform code cleanup.

bblommers:
Thanks @pvbouwel. Is this ready to merge now?

pvbouwel:
This by itself should be good to merge. It will already improve error messaging a lot imho. I think it's still a bit too forgiving for some updates but that is something that I will test and tackle when looking at using the structured parsing for actually performing the updates (ETA next weekend). But merging should not be blocked by the implementation of the executor.

setup.py
```diff
@@ -100,4 +100,5 @@ def get_version():
     project_urls={
         "Documentation": "http://docs.getmoto.org/en/latest/",
     },
+    data_files=[('', ['moto/dynamodb2/parsing/reserved_keywords.txt'])],
```
kislyuk, line 103: This incorrectly causes this file to be installed into the root of the Python package installation directory.

kislyuk:
After merging this PR, I now get the following error when installing moto:

`ERROR: Could not install packages due to an EnvironmentError: [Errno 13] Permission denied: '/usr/local/reserved_keywords.txt'`

This is a misconfiguration of setuptools as I highlighted in https://github.com/spulec/moto/pull/2885#discussion_r415150276. The `data_files` content needs to be updated to install the file in the correct package-specific location. @pvbouwel can you please take a look?

setup.py
pvbouwel: Can you check whether https://github.com/spulec/moto/pull/2925 works for your environment?

PR #2889, opened 2020-04-12
Allow reuse of DynamoType from models.py

pvbouwel:
By having models.py as one big file it causes to
easily create circular dependencies. With the current
setup it is not possible to re-use DynamoType. This
refactor moves it out to its own file while trying to
keep the structure as much as it is.

Creating a separate pull request given that it is a refactor change without functional change.

At this time models contains a lot of code in one file due to which it is easy to get circular dependencies. For example; for adding Validation to the tree-based parsing https://github.com/spulec/moto/pull/2885#issuecomment-612452300 I want to re-use the DynamoType class but I would need to import it from models.py which is the component that would already be creating the parser.  This can be solved by splitting out model.py.

There are multiple options here with at extreme ends:
 - Only split out the minimum to have DynamoType isolated (this CR)
 - Move out the maximum currently possible. Also move out Item, StreamRecord, StreamShard, DynamoJsonEncoder

 I have taken the the minimal approach to avoid conflicts with existing pull requests as much as possible but I am open to any suggestion that would unblock me from having this import cycle.

### DIFF
--- a/moto/dynamodb2/__init__.py
+++ b/moto/dynamodb2/__init__.py
@@ -1,5 +1,5 @@
 from __future__ import unicode_literals
-from .models import dynamodb_backends as dynamodb_backends2
+from moto.dynamodb2.models import dynamodb_backends as dynamodb_backends2
 from ..core.models import base_decorator, deprecated_base_decorator
 
 dynamodb_backend2 = dynamodb_backends2["us-east-1"]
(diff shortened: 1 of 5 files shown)


bblommers:
```the minimal approach to avoid conflicts```
That's my preference as well - hopefully this unblocks you for now!

PR #2890, opened 2020-04-12
fix incorrect response tag for DescribeListeners

bennyhat:
I can write a test for this if really desired. I didn't write one for this, since this API requires an LB or Listener as a parameter, it's quite a bit of setup, for not much gain.

### DIFF
--- a/moto/elbv2/responses.py
+++ b/moto/elbv2/responses.py
@@ -987,7 +987,7 @@ DESCRIBE_TARGET_GROUP_ATTRIBUTES_TEMPLATE = """<DescribeTargetGroupAttributesRes
   </ResponseMetadata>
 </DescribeTargetGroupAttributesResponse>"""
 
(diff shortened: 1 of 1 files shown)


bennyhat:
closed and re-opened to re-kick travis

bennyhat:
> A corresponding test would be good, @bennyhat! I appreciate that it's quite some work, but it would be good to ensure this works correctly (and keeps working correctly!)

Will do!

bblommers:
Closed and reopened this to kick off the CI, to verify it still passes.

I just verified: there already are tests for this feature, that pass, so boto3 does not care about the exact value. So I'll just merge it.

If we really want to properly test this, we'd need to start setting up tests using the Java/.NET SDK's, who are more strict in parsing the response.

bblommers (approved):

PR #2894, opened 2020-04-16
Add the STS call assume_role_with_saml

microe:
My company uses SAML to get access to our accounts on AWS. So we wrote a module that would use the STS call assume_role_with_saml to retrieve certificates. I went to test this new module we wrote using moto and found out that the endpoint for this had not been done yet. This is my take on adding this endpoint.

I used the responses I received to create the templates for the SAML assertion and the response from the endpoint. The SAML assertion is large and I am open to getting it into the test in some other way, but the test follows the pattern of the STS calls. All of the digests and hashes are faked, please let me know if higher fidelity of hashes are need.

Thanks for making a super useful Python module. I hope this helps make it better!

### DIFF
--- a/moto/server.py
+++ b/moto/server.py
@@ -139,12 +139,16 @@ class DomainDispatcherApplication(object):
     def get_action_from_body(self, environ):
         body = None
         try:
-            request_body_size = int(environ.get("CONTENT_LENGTH", 0))
-            if "wsgi.input" in environ:
+            # AWS requests use querystrings as the body (Action=x&Data=y&...)
+            simple_form = environ["CONTENT_TYPE"].startswith(
+                "application/x-www-form-urlencoded"
+            )
+            request_body_size = int(environ["CONTENT_LENGTH"])
+            if simple_form and request_body_size:
                 body = environ["wsgi.input"].read(request_body_size).decode("utf-8")
-                body_dict = dict(x.split("=") for x in str(body).split("&"))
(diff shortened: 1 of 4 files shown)


microe:
When I run make test-only on either python 2.7 or python 3.8 the new test passes, but a bunch of the tests it doesn't pass. Is there something I have to do to hook this call up to the moto server?

bblommers:
Looks like all the tests fail in server mode. 

To reproduce this locally: 

1. Run motoserver with  your local changes
2. Enable server mode in settings.py
3. Run the test again

Running motoserver locally can be done like this:
```
python setup.py sdist
pip install dist/moto-1.3.15.dev0.tar.gz
moto_server
```
Keep in mind that you need to rebuild/restart the motoserver after making any changes

bblommers:
@spulec Would you mind reviewing this as well? I think this is the best way of dealing with unsigned requests, but would be good to get a second pair of eyes

moto/server.py
```diff
@@ -139,12 +139,16 @@ def get_application(self, environ):
     def get_action_from_body(self, environ):
         body = None
         try:
-            request_body_size = int(environ.get("CONTENT_LENGTH", 0))
-            if "wsgi.input" in environ:
+            # AWS requests use querystrings as the body (Action=x&Data=y&...)
```
microe, line 142: I'll squash this if you want. Let me know.

spulec:
Yeah, I think this makes sense.

Thanks for the contribution @microe !

PR #2897, opened 2020-04-16
Adds initial support for secretsmanager update_secret

### LINKED ISSUE #2895: Support for secretsmanager update_secret
I would like to explore the possibility of adding update_secret support to the secretsmanager endpoint and am filing this issue per the contributor guidelines.

As of moto 1.3.14 support for update_secret is not yet implemented. The following is raised when the secretsmanager client is used in my personal code

```
E.   NotImplementedError: The update_secret action has not been implemented
.tox/unit/lib/python3.7/site-packages/moto/core/responses.py:393: NotImplementedError
```

I would like to take a stab at adding this so that I can use moto to test in-house code that calls this method.

I'm new to the moto process though, so want to have a maintainer +1 this work before I put any serious effort into it. Preliminary toying with a fix seems relatively straightforward.

caphrim007:
Fixes #2895 

The support in this patch is preliminary and may or may not be feature complete.
It provides the basic support for update_secret so that future work can build on it as needed.

### DIFF
--- a/moto/secretsmanager/models.py
+++ b/moto/secretsmanager/models.py
@@ -107,6 +107,34 @@ class SecretsManagerBackend(BaseBackend):
 
         return response
 
+    def update_secret(
+        self, secret_id, secret_string=None, secret_binary=None, **kwargs
+    ):
+
+        # error if secret does not exist
+        if secret_id not in self.secrets.keys():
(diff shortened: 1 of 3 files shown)


caphrim007:
@bblommers ready for review

Issue #2006, opened 2018-12-31
How to set RDS to stopped state?


yeukhon:
##  Question

I am trying to run RDS's stop_db_instance. But I am getting:
```
InvalidDBClusterStateFault: An error occurred (InvalidDBClusterStateFault) when calling the StopDBInstance operation: Invalid DB type, when trying to perform StopDBInstance on infra-managere. See AWS RDS documentation on rds.stop_db_instance
```

How do I mock the state to be 'stopped'?

Thank you.

spulec:
Can you please add the code you are using to set this up?

thisiswei:
great; running into this error today

```python
from moto import (
    mock_ec2, mock_autoscaling,
    mock_rds2, mock_cloudformation, mock_route53,
...
)
....

class cf_test_cm(object):
    """
    Same idea as `cf_test` decorator, but this is a context manager.
    This was created because `pytest.mark.parametrize` does not work well with other decorators
    """

    def __init__(self):
        self.ec2 = mock_ec2()
        self.autoscaling = mock_autoscaling()
        self.rds = mock_rds2()
        self.route53 = mock_route53()
        self.cloudformation = mock_cloudformation()
        self.s3 = mock_s3()
        self.sts = mock_sts()
        self.ecs = mock_ecs()
        self.events = mock_events()
        self.ecr = mock_ecr()
        self.elb = mock_elb()
        self.elbv2 = mock_elbv2()
        self.cloudwatch = mock_cloudwatch()
        self.iam = mock_iam()
        self.cf_fixture = {}

    def __getitem__(self, attr):
        return self.cf_fixture[attr]

    def __enter__(self):
        self.ec2.start()
        self.autoscaling.start()
        self.rds.start()
        self.route53.start()
        self.cloudformation.start()
        self.s3.start()
        self.sts.start()
        self.ecs.start()
        self.events.start()
        self.ecr.start()
        self.elb.start()
        self.elbv2.start()
        self.cloudwatch.start()
        self.iam.start()

        self.cf_fixture = create_cf_fixture()

        return self

    def __exit__(self, *args, **kw):
        self.ec2.stop()
        self.autoscaling.stop()
        self.rds.stop()
        self.route53.stop()
        self.cloudformation.stop()
        self.s3.stop()
        self.sts.stop()
        self.ecs.stop()
        self.events.stop()
        self.ecr.stop()
        self.elb.stop()
        self.elbv2.stop()
        self.cloudwatch.stop()
[...]

PR #2900, opened 2020-04-16
Bugfix: Allow stop_db_instance for compatible engines (refs #2006)

hltbra:
From the [RDS documentation](https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/USER_StopInstance.html#USER_StopInstance.Limitations):
> You can stop and start a DB instance whether it is configured for a single Availability Zone or for Multi-AZ, for database engines that support Multi-AZ deployments. You can't stop an Amazon RDS for SQL Server DB instance in a Multi-AZ configuration.

---

Related to https://github.com/spulec/moto/issues/2006


spulec:
Good catch; thank you!

PR #2909, opened 2020-04-20
S3 - Allow keys to start with leading slashes

### LINKED ISSUE #2759: S3 will not return '/' in CommonPrefixes when there is.
## Reporting Bugs

# Description
When there is some Key with consecutive slashes `'/'`, `list_objects_v2` with `Delimiter='/'` will not include the right answer in `CommonPrefixes`

# Code to reproduce
```python
import boto3
from moto import mock_s3

@mock_s3
def list_slashes():
    c = boto3.client('s3')
    c.create_bucket(Bucket='bucket')
    c.put_object(Bucket='bucket', Key='///')

    resp = c.list_objects_v2(Bucket='bucket', Prefix='', Delimiter='/')
    print(resp)

    resp = c.list_objects_v2(Bucket='bucket', Prefix='/', Delimiter='/')
    print(resp)

list_slashes()
```

# Expected
```python
{
  ...,
 'CommonPrefixes': [{'Prefix': '/'}],
 'KeyCount': 1,
  ...,
}
{
  ...,
 'CommonPrefixes': [{'Prefix': '//'}],
 'KeyCount': 1,
  ...,
}
```

# Reality
```python
{
   ...,
   'Contents': [{'Key': '', ..., }],
   'KeyCount': 1,
   ...
}
{
   ...,
   'KeyCount': 0,
   ...,
}
```

sample output from real aws S3
```python
import boto3
c = boto3.client('s3')
c.put_object(Bucket='oss-playground', Key='///') 
print(c.list_objects_v2(Bucket='oss-playground', Prefix='//', Delimiter='/'))

```
Output
```python
{
 ...,
 'IsTruncated': False,
 'Name': 'oss-playground',
 'Prefix': '//',
 'Delimiter': '/',
 'CommonPrefixes': [{'Prefix': '///'}],
 'KeyCount': 1}
```

oliparcol:
Fixes #2759 

S3 allows objects keys to start with slashes (`/`); all leading slashes were previously incorrectly removed from key in `parse_key_name` instead of only removing the first one.

`moto.s3.urls` was also updated to clearly include the leading slash in the `key_name` group.

### DIFF
--- a/moto/s3/responses.py
+++ b/moto/s3/responses.py
@@ -135,7 +135,7 @@ ACTION_MAP = {
 
 def parse_key_name(pth):
     # strip the first '/' left by urlparse
-    return pth[1:] if pth.startswith('/') else pth
+    return pth[1:] if pth.startswith("/") else pth
 
 
 def is_delete_keys(request, path, bucket_name):
--- a/moto/s3/responses.py
+++ b/moto/s3/responses.py
@@ -134,7 +134,8 @@ ACTION_MAP = {
 
 
 def parse_key_name(pth):
-    return pth.lstrip("/")
+    # strip the first '/' left by urlparse
+    return pth[1:] if pth.startswith('/') else pth
 
 
 def is_delete_keys(request, path, bucket_name):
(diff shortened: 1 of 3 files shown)


oliparcol:
oops... sorry. I will have a look.

oliparcol:
Formatting fixed in d852f7dd, thanks

spulec:
Looks great; thanks!

PR #2914, opened 2020-04-21
Add get_databases method to glue and update IMPLEMENTATION_COVERAGE.md

MarcosBernal:
### Overview
- Add get_databases method to glue moto client
- Update IMPLEMENTATION_COVERAGE.md with methods that were implemented previously

It only adds the simplified functionality. As in the rest of the glue methods, this method assumes no optional parameters are provided (CatalogId, NextToken, MaxResults)

### Test
Test can be checked with:
`nosetests --tests tests/test_glue/test_datacatalog.py`

### DIFF
--- a/IMPLEMENTATION_COVERAGE.md
+++ b/IMPLEMENTATION_COVERAGE.md
@@ -3351,11 +3351,11 @@
 - [ ] update_listener
 
 ## glue
-4% implemented
-- [ ] batch_create_partition
+14/123 = 11% implemented
+- [X] batch_create_partition
 - [ ] batch_delete_connection
-- [ ] batch_delete_partition
-- [ ] batch_delete_table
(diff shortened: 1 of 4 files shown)


MarcosBernal:
I would suggest to add the command `black --diff moto/ tests/` either in the CONTRIBUTING.md file or in the Makefile.

IMPLEMENTATION_COVERAGE.md
```diff
@@ -3351,11 +3351,11 @@
 - [ ] update_listener
 
 ## glue
-4% implemented
-- [ ] batch_create_partition
+14/123 = 11% implemented
```
bblommers, line 3354:
```suggestion
11% implemented
```

Just stick with the %-notation

PR #2155, opened 2019-04-16
set default status for s3 posts and add success_action_redirect support

gtmanfred:
and add support for success_action_redirect.

Based on https://docs.aws.amazon.com/AmazonS3/latest/API/RESTObjectPOST.html

Posts should default to `204`.

Also, add support for setting a `success_action_redirect` on

PR #2916, opened 2020-04-22
Make all CallbackResponse requests into a Werkzeug Request

gtmanfred:
The "request" object in CallbackResponse is the PreparedRequest send by
whatever client is used to contact the mocked moto service. This can end
up with unparsed bodies, as we added for processing presigned post
requests in #2155. This will make sure that all of the requests comming
in from mocked functions also get processed by werkzeug as if it was
running a live server.


gtmanfred:
I was thinking about this last night, and decided to take a stab at making it available of all functions this morning.

PR #2919, opened 2020-04-23
Handle VersionId in S3:delete_objects

### LINKED ISSUE #2918: S3 delete_objects ignores VersionId field
**Moto version:** 1.3.14, installed with pip (the bug is present on master)

When calling delete_objects on a versioned bucket, the `VersionId` field is ignored, and a regular deletion is performed instead of deleting the specific version.
This is contrary to what happens when doing it with `delete_object`

```python
client = boto3.client("s3", region_name=DEFAULT_REGION_NAME)
client.delete_objects(
        Bucket="blah", Delete={"Objects": [{"Key": "test1", "VersionId": "XX"}]}
) # Does not delete the specific version
client.delete_object(Bucket="blah", Key="test1", VersionId"="XX") # Works as expected
```

I'm preparing a PR with a fix.

(thanks for the awesome lib !)

antoinewdg-mt:
Fixes #2918

### DIFF
--- a/moto/s3/responses.py
+++ b/moto/s3/responses.py
@@ -839,21 +839,22 @@ class ResponseObject(_TemplateEnvironmentMixin, ActionAuthenticatorMixin):
 
     def _bucket_response_delete_keys(self, request, body, bucket_name):
         template = self.response_template(S3_DELETE_KEYS_RESPONSE)
+        body_dict = xmltodict.parse(body)
 
-        objects = minidom.parseString(body).getElementsByTagName("Object")
-
-        deleted_objects = []
-        error_names = []
+        objects = body_dict["Delete"].get("Object", [])
+        if not isinstance(objects, list):
+            # We expect a list of objects, but when there is a single <Object> node xmltodict does not
+            # return a list.
+            objects = [objects]
         if len(objects) == 0:
             raise MalformedXML()
 
+        deleted_objects = []
+        error_names = []
+
         for object_ in objects:
-            key_name = object_.getElementsByTagName("Key")[0].firstChild.nodeValue
-            version_id_node = object_.getElementsByTagName("VersionId")
-            if version_id_node:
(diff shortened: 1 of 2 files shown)


moto/s3/responses.py
```diff
@@ -849,26 +849,33 @@ def _get_path(request):
...
 
-        for k in keys:
-            key_name = k.firstChild.nodeValue
+        for object_ in objects:
+            key_name = object_.getElementsByTagName("Key")[0].firstChild.nodeValue
+            version_id_node = object_.getElementsByTagName("VersionId")
+            if version_id_node:
+                version_id = version_id_node[0].firstChild.nodeValue
+            else:
+                version_id = None
```
antoinewdg-mt, line 865: Not really an expert on manipulating XML DOM, there may be a more elegant way to do this

gmcrocetti: Hi @antoinewdg-mt, great to see you here :). Did you take a look at [etree](https://docs.python.org/3.8/library/xml.etree.elementtree.html) ? It think it's possible to create a xpath lookup for your problem. Hope it helps :)

bblommers: In other parts of moto we tend to use xmltodict:
```xmltodict.parse(body)["Object"]["Key"]["VersionId"]```

Let's use that here as well - it's quite clean, and it makes it obvious what the expected input is

antoinewdg-mt:
Updated to use `xmltodict`.
The XML validation feels a bit lacking, but the wasn't much before the PR. Is it something that would be useful, or are we only receiving XML that has only been generated by the lib that we can trust to be valid ?

bblommers:
Thanks @antoinewdg-mt. The XML is generated by boto, so that should be fine.

antoinewdg-mt:
Thanks for the super fast feedback cycle !

PR #2921, opened 2020-04-24
Support OptInStatus for EC2 describe_region calls

getglad:
The `describe_regions` call now includes an `OptInStatus` key in the return.

Examples:
- https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/ec2.html#EC2.Client.describe_regions
- https://docs.aws.amazon.com/cli/latest/reference/ec2/describe-regions.html

Added this capability given these constraints:
- Regions can have a status of `opt-in-not-required`, `not-opted-in`, or `opted-in`
- Regions introduced by AWS after 3/20/19 must be enabled (ie, status of `not-opted-in`, or `opted-in`)
   - Source: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-regions-availability-zones.html (under "Available Regions" section)
   - Far as I know, this should be `me-south-1` and `ap-east-1`
- Regions before 3/20/19 should have status of `opt-in-not-required`

I am not exactly sure what China or Gov regions should return, but I set it to `opt-in-not-required`.

### DIFF
--- a/moto/ec2/models.py
+++ b/moto/ec2/models.py
@@ -1504,21 +1504,30 @@ class Zone(object):
 
 
 class RegionsAndZonesBackend(object):
-    regions_not_enabled_by_default = [
-        'ap-east-1',
-        'me-south-1'
-    ]
+    regions_not_enabled_by_default = ["ap-east-1", "me-south-1"]
 
     regions = []
     for region in Session().get_available_regions("ec2"):
         if region in regions_not_enabled_by_default:
-            regions.append(Region(region, "ec2.{}.amazonaws.com".format(region), "not-opted-in"))
+            regions.append(
+                Region(region, "ec2.{}.amazonaws.com".format(region), "not-opted-in")
+            )
         else:
-            regions.append(Region(region, "ec2.{}.amazonaws.com".format(region), "opt-in-not-required"))
+            regions.append(
+                Region(
+                    region, "ec2.{}.amazonaws.com".format(region), "opt-in-not-required"
+                )
+            )
     for region in Session().get_available_regions("ec2", partition_name="aws-us-gov"):
-        regions.append(Region(region, "ec2.{}.amazonaws.com".format(region), "opt-in-not-required"))
+        regions.append(
+            Region(region, "ec2.{}.amazonaws.com".format(region), "opt-in-not-required")
+        )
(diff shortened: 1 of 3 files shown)


bblommers:
Hi @getglad, based on this blog post it looks like all new regions will be not-opted-in:
https://aws.amazon.com/blogs/security/setting-permissions-to-enable-accounts-for-upcoming-aws-regions/

This is consistent with my personal account, where the new Capetown region ```af-south-1``` is disabled by default - so I think the table is wrong.

I've created a PR at AWS (https://github.com/awsdocs/amazon-ec2-user-guide/pull/108) to verify this.

Related issue in boto to expose the Opt-In-Status: https://github.com/boto/boto3/issues/2022

getglad:
@bblommers makes sense. So I guess there are two questions that stem from that:

1. Should I swap how I've done this, so that there is a static set of `opt-in-not-required` vs a static set of `not-opted-in`? More typing for me short term, but long term would eliminate a need to open a PR for every new region. I would imagine the regions will roll in slowly, but still... something I could do here.

2. Given my implementation, you would never get a `opted-in` result. Should we consider a method to simulate enabling a region? It's not clear to me if this can actually be done programmatically (https://docs.aws.amazon.com/awsaccountbilling/latest/aboutv2/manage-account-payment.html#disable-region), so I don't know what the best way to get to that functionality would be. I could see figuring out how we want to at least do `opt-in-not-required` vs `not-opted-in` for now, and opening and issue to see if we can get a proposal rolling.

Thoughts?

bblommers:
Hey @getglad, the PR at AWS was just merged, so we're correct that the Capetown should also be opt-in-required.

1. Yes, I think it makes sense to have a static set of `opt-in-not-required`, so we don't have to be on top of it.
2. As far as I know it's not possible to programmatically disable/enable regions, so lets not worry about that now. If/when people need that functionality we can always think of how to expand this

getglad:
@bblommers done

spulec:
Looks great, thanks!

getglad:
@spulec or @bblommers - just wondering when/if there is a plan to tag a new release for pypi so that this would be available?

bblommers:
Hi @getglad, dev releases are done automatically after every PR merge: https://pypi.org/project/moto/#history

getglad:
@bblommers thanks!

PR #2922, opened 2020-04-25
Add policy to API Gateway 

### LINKED ISSUE #2721: API Gateway missing support for resource policies on REST APIs
API Gateway supports adding a resource policy to REST APIs in order to control access to the API, but moto is missing support for the `policy` parameter in the API Gateway models/responses:
https://docs.aws.amazon.com/apigateway/api-reference/resource/rest-api/#policy

Would also be nice if the syntax of the resource policy was validated using the existing IAM policy validation logic in moto (although it doesn't fully support resource-based policies currently).

johnwc:
We needed to be able to unit test lambda code that checks the existence of policy, added policy to ApiGateway model and response.

Fixes #2721

### DIFF
--- a/moto/apigateway/models.py
+++ b/moto/apigateway/models.py
@@ -461,6 +461,7 @@ class RestAPI(BaseModel):
         self.description = description
         self.create_date = int(time.time())
         self.api_key_source = kwargs.get("api_key_source") or "HEADER"
+        self.policy = kwargs.get("policy") or None
         self.endpoint_configuration = kwargs.get("endpoint_configuration") or {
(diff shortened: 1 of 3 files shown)


JackDanger:
@johnwc thanks for this addition! This is helpful and much appreciated. Feel free to contribute more of the missing parts as you see fit. Welcome to Moto :)

johnwc:
@JackDanger I came across needing VPCEndpoints today and mocked it up just to get moving on my project. I will try to move that over within the next week. I just replicated how VPCs where being managed.

What is the time frame for changes like this to be published?

PR #2924, opened 2020-04-26
Improve DDB expressions support4: Execution using AST

pvbouwel:
Part of structured approach for UpdateExpressions:
 1) Expression gets parsed into a tokenlist (tokenized)
 2) Tokenlist get transformed to expression tree (AST)
 3) The AST gets validated (full semantic correctness)
 4) AST gets processed to perform the update -> this commit

This commit uses the AST to execute the UpdateExpression.
All the existing tests pass. The only tests that have been
updated are in test_dynamodb_table_with_range_key.py because
they wrongly allow adding a set to a path that doesn't exist.
This has been alligend to correspond to the behavior of AWS
DynamoDB.

This commit will resolve https://github.com/spulec/moto/issues/2806
Multiple tests have been implemented that verify this.

This would have conflicts with https://github.com/spulec/moto/pull/2896 but this solution also implements the required functionality and thus the tests pass. I can copy over the tests from there or that pull request can be reduced to its test cases.

### DIFF
--- a/moto/dynamodb2/exceptions.py
+++ b/moto/dynamodb2/exceptions.py
@@ -39,6 +39,17 @@ class AttributeDoesNotExist(MockValidationException):
         super(AttributeDoesNotExist, self).__init__(self.attr_does_not_exist_msg)
 
 
+class ProvidedKeyDoesNotExist(MockValidationException):
+    provided_key_does_not_exist_msg = (
+        "The provided key element does not match the schema"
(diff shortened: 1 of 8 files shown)


spulec:
This looks great; thanks for all of the work!

bblommers:
I'll reduce and rebase #2896 @pvbouwel - thanks for your contribution!

PR #2925, opened 2020-04-26
Place reserved_keywords.txt not in root.

pvbouwel:
Do not place reserved_keywords.txt in the root
as this can give environment errors.

This was raised in: https://github.com/spulec/moto/pull/2885#discussion_r415150276

From testing in virtual_env it avoids placing the one in the root and there will only be one in the parsing module directory.

@kislyuk could you verify whether this works for your setup as well?

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -3,5 +3,6 @@ include requirements.txt requirements-dev.txt tox.ini
 include moto/ec2/resources/instance_types.json
 include moto/ec2/resources/amis.json
 include moto/cognitoidp/resources/*.json
+include moto/dynamodb2/parsing/reserved_keywords.txt
 recursive-include moto/templates *
 recursive-include tests *
--- a/setup.py
+++ b/setup.py
@@ -101,5 +101,4 @@ setup(
     project_urls={
         "Documentation": "http://docs.getmoto.org/en/latest/",
     },
-    data_files=[('', ['moto/dynamodb2/parsing/reserved_keywords.txt'])],
 )


kislyuk:
@pvbouwel no, I still get the following:
```
ERROR: Could not install packages due to an EnvironmentError: [Errno 13] Permission denied: '/usr/local/moto'
Consider using the `--user` option or check the permissions.
```
I believe you need to add your file to https://github.com/spulec/moto/blob/master/MANIFEST.in instead of using `data_files`. See this comment: https://github.com/jupyter/jupyter-packaging/issues/11#issuecomment-295262893. (There are lots of scattered tidbits of information online but the gist of it is that `data_files` is for files that the package (for some extraordinary reason) wants to install system-wide, while `package_data` is for files attached to the package; to use `package_data`, you set `include_package_data=True` in setup.py and EITHER set `package_data` or add your files to `MANIFEST.in`; since moto already has a `MANIFEST.in` and sets `include_package_data=True`, the correct solution is to just add to `MANIFEST.in`).

pvbouwel:
> @pvbouwel no, I still get the following:
> 
> ```
> ERROR: Could not install packages due to an EnvironmentError: [Errno 13] Permission denied: '/usr/local/moto'
> Consider using the `--user` option or check the permissions.
> ```
> 
> I believe you need to add your file to https://github.com/spulec/moto/blob/master/MANIFEST.in instead of using `data_files`. See this comment: [jupyter/jupyter-packaging#11 (comment)](https://github.com/jupyter/jupyter-packaging/issues/11#issuecomment-295262893). (There are lots of scattered tidbits of information online but the gist of it is that `data_files` is for files that the package (for some extraordinary reason) wants to install system-wide, while `package_data` is for files attached to the package; to use `package_data`, you set `include_package_data=True` in setup.py and EITHER set `package_data` or add your files to `MANIFEST.in`; since moto already has a `MANIFEST.in` and sets `include_package_data=True`, the correct solution is to just add to `MANIFEST.in`).

Thanks for educating me. I didn't know about the difference but after the explanation it does make sense to prefer using package_data over data_files!  Removed the data_files entry and added it to the manifest.

kislyuk:
@pvbouwel OK, I ran `pip install git+https://github.com/pvbouwel/moto@0bd586e` and it no longer triggers this error for me:
```
Successfully installed moto-1.3.15.dev634
```

bblommers:
I learned something new today as well! Thanks for the PR @pvbouwel, and thanks for helping out @kislyuk!

PR #2929, opened 2020-04-27
Fixed a regression with CloudWatch

mikegrima:
Fixed a regression in CloudWatch.

Also fixes an unmocking issue related to a recent CloudWatch change.
@bblommers FYI ^^


mikegrima:
@bblommers I needed to disable some of the S3 Cloudwatch functionality you wrote in a previous PR to resolve the unmocking regression.

I added some comments in the S3 model to reflect what needs to be done (TL;DR - need to wrap the dictionary with setter and getter functions vs. direct import).

bblommers:
Thanks for letting me know @mikegrima, and thanks for fixing it!
Do you have an example test/script that demonstrates how it breaks things?

Issue #2826, opened 2020-03-20
implementation of ec2 vpc endpoint - are there plans?


111andre111:
Are there any plans to implement vpc endpoint creation at some time?
Regarding this document 
https://github.com/spulec/moto/blob/master/IMPLEMENTATION_COVERAGE.md#ec2
and looking in the source code under @ mock-ec2 decorator I cannot see any equivalent up to now.
It would be nice to to have that feature.

Thank you

johnwc:
I've actually did this, as we needed testing for endpoints as well. I'm hoping to get the addition submitted within the next week.

bblommers:
Already in place @johnwc :)

111andre111:
Perfect, thank you indeed.

PR #2930, opened 2020-04-28
EC2 enhancement  create vpc end point

usmangani1:
Enhancement create vpc end point

https://github.com/localstack/localstack/issues/1450

### DIFF
--- a/moto/ec2/models.py
+++ b/moto/ec2/models.py
@@ -104,6 +104,7 @@ from .utils import (
     random_internet_gateway_id,
     random_ip,
     random_ipv6_cidr,
+    randor_ipv4_cidr,
     random_launch_template_id,
     random_nat_gateway_id,
     random_key_pair,
@@ -112,6 +113,8 @@ from .utils import (
     random_reservation_id,
     random_route_table_id,
     generate_route_id,
+    generate_vpc_end_point_id,
+    create_dns_entries,
     split_route_id,
     random_security_group_id,
     random_snapshot_id,
@@ -2735,6 +2738,7 @@ class VPCBackend(object):
 
     def __init__(self):
         self.vpcs = {}
+        self.vpc_end_points = {}
         self.vpc_refs[self.__class__].add(weakref.ref(self))
         super(VPCBackend, self).__init__()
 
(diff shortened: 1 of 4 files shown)


usmangani1:
> Enhancement create vpc end point
> 
> [localstack/localstack#1450](https://github.com/localstack/localstack/issues/1450)

@bblommers  Can you please review this and help me get it merged to master

moto/ec2/models.py
```diff
@@ -2877,6 +2881,66 @@ def associate_vpc_cidr_block(
...
+                             tag_specifications=None,
+                             private_dns_enabled=None
+                            ):
+
+        vpc_endpoint_id = generate_vpc_end_point_id(vpc_id)
+
+        #validates if vpc is present or not.
+        self.get_vpc(vpc_id)
+
+        if type == "interface" or "Interface ":
```
bblommers, line 2904: This is evaluated as ```(type == "interface") or ("Interface")```
The string `Interface` will always evaluate to True, regardless of the outcome of `type == interface` - so this statement doesn't do what you think it does
```suggestion
        if type and type.lower() == "interface":
```

tests/test_ec2/test_route_tables.py
```diff
@@ -618,3 +618,39 @@ def test_describe_route_tables_with_nat_gateway():
     nat_gw_routes[0]["DestinationCidrBlock"].should.equal("0.0.0.0/0")
     nat_gw_routes[0]["NatGatewayId"].should.equal(nat_gw_id)
     nat_gw_routes[0]["State"].should.equal("active")
+
```
bblommers, line 621: Can you add some test cases for the different VPC types ('Interface'|'Gateway'|None).

Also add some asserts that either DNS entries are created, or routes - but not both.

bblommers:
Fixes #2826

111andre111:
@bblommers Any idea of when will be the next release cycle, because this a quite huge effort here?

bblommers:
@111andre111 Every PR creates a new dev-release, available here: https://pypi.org/project/moto/#history
If I'm not mistaken, this feature should be available in moto >= 1.3.15.dev776

111andre111:
Cool, perfect, that's good to know.

PR #2932, opened 2020-04-29
Always call update ELBs for ASGs

DenverJ:
Fixed a bug where the attached ELBs were not correctly updated when an ASG instance is put in standby.


DenverJ:
I think there was a problem with the pipeline here but I also think I was a bit hasty and need to do some more tests so I'll close and re-raise this.

PR #2933, opened 2020-04-29
relax version pins in setup.py for non-Py2 users

### LINKED ISSUE #2836: Moto's dependencies are locked to specific versions, makes it hard to install
As of https://github.com/spulec/moto/commit/00134d2df37bb4dcd5f447ef951d383bfec0903c some of the dependencies in `setup.py` are specified to exact (and outdated) versions. This makes it hard to install moto.

For example `more-itertools==5.0.0` is now over two years old, they're now on version 8.2... I am using version 8.2 in my project. I can no longer update moto to the latest on master in my project without doing some hackage to force my package manager (poetry) to ignore the dependency on _exactly_ 5.0.0

More details here python-poetry/poetry#697

Would moving python 2 support to a separate branch be a possibility? Or doing a version check on major python version in setup.py so there can effectively be a `install_requires_py2` and a `install_requires_py3` ?

mpenkov:
Fix #2836

### DIFF
--- a/moto/s3/models.py
+++ b/moto/s3/models.py
@@ -1370,7 +1370,7 @@ class S3Backend(BaseBackend):
             raise MissingKey(key_name)
         self.tagger.delete_all_tags_for_resource(key.arn)
         self.tagger.tag_resource(
-            key.arn, [{"Key": key, "Value": value} for key, value in tags.items()],
+            key.arn, [{"Key": k, "Value": v} for (k, v) in tags.items()],
         )
         return key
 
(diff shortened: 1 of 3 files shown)


mpenkov:
The CI checks seem to be failing, but I suspect it's for reasons unrelated to this PR, because they fail for the master HEAD as well.

andersk:
Conditional requirements like this need to be properly expressed in the metadata using environment markers, not in code, because the metadata is only computed from `setup.py` once with one particular version of Python. (#3618)

PR #2697, opened 2020-01-17
Import time reduction

Procrat:
This speeds up import time from ~5.4s to 0.2s on my machine.

PR #2934, opened 2020-04-29
Revert "Import time reduction"

asherf:
Breaks importing backend.
See 
https://github.com/spulec/moto/pull/2697#issuecomment-621300975
Reverts spulec/moto#2697

### DIFF
--- a/moto/__init__.py
+++ b/moto/__init__.py
@@ -1,114 +1,62 @@
 from __future__ import unicode_literals
 
-import importlib
-
-
-def lazy_load(module_name, element):
-    def f(*args, **kwargs):
-        module = importlib.import_module(module_name, "moto")
-        return getattr(module, element)(*args, **kwargs)
-
-    return f
-
-
-mock_acm = lazy_load(".acm", "mock_acm")
-mock_apigateway = lazy_load(".apigateway", "mock_apigateway")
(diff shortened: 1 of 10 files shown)


asherf:
Screen Shot 2020-04-29 at 8 55 37 AM

this revert fixes the issue.

asherf:
FWIW, since this library is mostly used in tests, I don't think load time is such a big issue.
The perf issue can be somewhat mitigated by eagerly loading moto and triggering expensive import in test suite initialization.

The added complexity and risk of doing local imports is not worth it in this case.
just me 2c.

bblommers:
@asherf I've created this PR to fix the problem #2938, which has now been merged.
Would you mind pulling master and verifying that the problem is now fixed?

bblommers:
> 
> 
> FWIW, since this library is mostly used in tests, I don't think load time is such a big issue.


I agree it's not a big issue @asherf  - definitely classifies as a nice to have. Quick feedback is always good though, and I personally quite like the fact I don't have to wait 3/4 secs for a single unit test.

As far as local imports go - I'd like to think that our test coverage is good enough to catch most issues. (And yes, I realize the irony of that statement, considering the context of this conversation.. :) )

It did highlight the fact that there is space for structural improvements! Still thinking of a way to remove the local imports here, and remove the cyclical dependency altogether.

PR #2938, opened 2020-04-30
Circular import error

bblommers:
Breaks importing backend.
See 
https://github.com/spulec/moto/pull/2697#issuecomment-621300975
Reverts spulec/moto#2697

@spulec Would you mind reviewing this?


spulec:
Thanks for the quick fix!

PR #2939, opened 2020-04-30
Update Dockerfile

redyvon:
use python:3.7-slim image to avoid packages building steps for alpine and have a smaller image size


redyvon:
also the current alpine:3.6 is segfaulting after starting it as an "sns"  service and doing a basic curl on url

mikegrima:
@spulec any issues?

spulec:
Looks good to me; thanks!

Issue #2943, opened 2020-05-01
placementConstraint is not stored in task definition


rifqifatih:
Hi,
I'm trying to get the `placementConstraints` set on ecs `register_task_definition`. I assume this is not yet implemented because when I do `describe_task_definition`, it didn't return the `placementConstraints` that I described before.

I'm not sure if this will complicate the mocking behavior, but returning the assigned value should be ok, I guess.

Reproduce:
```
ecs_client.register_task_definition(family='family',
                                            containerDefinitions=[{'image': 'image',
                                                                   'name': 'name'}],
                                            placementConstraints=[{
                                                "type": "memberOf",
                                                "expression": "something"
                                            }])

taskdef = ecs_client.describe_task_definition(taskDefinition='family:1')
taskdef.get("placementConstraints")
```
Moto version: 1.3.14

spulec:
Closed with https://github.com/spulec/moto/pull/2948

PR #2948, opened 2020-05-02
ECS: add placement constraints

rifqifatih:
I'm trying to add additional information `placementConstraints` into task definition mentioned in #2943 . Any advice would be appreciated.


spulec:
This looks great; thanks for your contribution and welcome to Moto!

In case you aren't already aware, Moto does a pre-release on every commit to master so you should be able to use this shortly if helpful: https://github.com/spulec/moto#releases

Issue #2956, opened 2020-05-03
Add managedblockchain network functions


jpbelleau:
Add managedblockchain - start with network functions

jpbelleau:
@bblommers - should I add another PR to this issue to continue adding functionaltiy or close and open a new issue?

bblommers:
Hey @jpbelleau - if you're adding new functionality, you can just create a standalone PR. 
It doesn't have to be tied to a specific issue.

Will close this - if you happen to run into problems with any of the services, feel free to open a new issue

PR #2957, opened 2020-05-03
Add managedblockchain network functions #2956

jpbelleau:
Add managedblockchain network functions #2956


moto/managedblockchain/models.py
```diff
@@ -0,0 +1,151 @@
...
+        self.region_name = region_name
+
+    def reset(self):
+        region_name = self.region_name
+        self.__dict__ = {}
+        self.__init__(region_name)
+
+    def create_network(
+        self,
+        json_body,
```
bblommers, line 93: Can we make the function-arguments explicit here? The models-class shouldn't need to know the JSON structure - that's the responsibility of the Response-class
```create_network(self,, name, framework, ..)```

File: tests/test_managedblockchain/test_managedblockchain_networks.py [hunk 2]
```diff
@@ -0,0 +1,53 @@
+from __future__ import unicode_literals
+
+import boto3
+import sure  # noqa
+
+from moto import mock_managedblockchain
+
+
+@mock_managedblockchain
+def test_create_network():
```
bblommers, line 10: Would you mind adding a few negative tests as well? I.e., test_invalid_framework, test_invalid_framework_version, etc.
Ideally also a test to verify what happens if `get_network` is called with an unknown NetworkId

jpbelleau:
Thanks!  Will be working on these later this week.

moto/managedblockchain/responses.py
```diff
@@ -0,0 +1,92 @@
...
+    def _network_response_post(self, json_body, querystring, headers):
+        name = json_body["Name"]
+        framework = json_body["Framework"]
+        frameworkversion = json_body["FrameworkVersion"]
+        frameworkconfiguration = json_body["FrameworkConfiguration"]
+        voting_policy = json_body["VotingPolicy"]
+        member_configuration = json_body["MemberConfiguration"]
+
+        # Optional
+        description = None
```
bblommers, line 58: You can simplify this:
`description = json_body.get("Description",  None)`

tests/test_managedblockchain/test_managedblockchain_networks.py
jpbelleau: Updated

moto/managedblockchain/responses.py
jpbelleau: Updated

moto/managedblockchain/models.py
```diff
@@ -0,0 +1,176 @@
...
+        self,
+        name,
+        framework,
+        frameworkversion,
+        frameworkconfiguration,
+        voting_policy,
+        member_configuration,
+        description=None,
+    ):
+        self.name = name
```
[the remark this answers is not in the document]
jpbelleau: Member configuration will be used when the member functions are added.  Creating a network will also create the first member of the network.  The converse is the same, a network is removed when the last member is removed.  Wanted to start small with the core network functions to learn the ropes and thank you for guiding me.

PR #2962, opened 2020-05-05
Implement parentGroup, namePrefixFilter and recursive for list_thing_groups()

### LINKED ISSUE #2961: iot list_thing_groups is incomplete
Current implementation of list_thing_groups doesn't take into account any function inputs. These are the inputs listed in boto3.
```
response = client.list_thing_groups(
    nextToken='string',
    maxResults=123,
    parentGroup='string',
    namePrefixFilter='string',
    recursive=True|False
)
```
This is the current implementation in moto, returning all the groups always:
```
def list_thing_groups(self, parent_group, name_prefix_filter, recursive):
    thing_groups = self.thing_groups.values()
    return thing_groups
```

It must also take into account `ResourceNotFoundException`

I'll be looking into this right now. 
I need at least an implementation for parentGroup for my project.

Chagui-:
Closes #2961

### DIFF
--- a/moto/iot/models.py
+++ b/moto/iot/models.py
@@ -857,8 +857,30 @@ class IoTBackend(BaseBackend):
         del self.thing_groups[thing_group.arn]
 
     def list_thing_groups(self, parent_group, name_prefix_filter, recursive):
-        thing_groups = self.thing_groups.values()
-        return thing_groups
+        if recursive is None:
+            recursive = True
+        if name_prefix_filter is None:
+            name_prefix_filter = ""
+        if parent_group and parent_group not in [
+            _.thing_group_name for _ in self.thing_groups.values()
+        ]:
+            raise ResourceNotFoundException()
+        thing_groups = [
+            _ for _ in self.thing_groups.values() if _.parent_group_name == parent_group
+        ]
+        if recursive:
+            for g in thing_groups:
+                thing_groups.extend(
+                    self.list_thing_groups(
+                        parent_group=g.thing_group_name,
+                        name_prefix_filter=None,
+                        recursive=False,
+                    )
+                )
+        # thing_groups = groups_to_process.values()
+        return [
+            _ for _ in thing_groups if _.thing_group_name.startswith(name_prefix_filter)
+        ]
 
     def update_thing_group(
         self, thing_group_name, thing_group_properties, expected_version
--- a/tests/test_iot/test_iot.py
+++ b/tests/test_iot/test_iot.py
@@ -757,6 +757,140 @@ def test_delete_principal_thing():
 
 
 @mock_iot
+def test_list_thing_groups():
+    client = boto3.client("iot", region_name="ap-northeast-1")
+    group_name_1a = "my-group-name-1a"
+    group_name_1b = "my-group-name-1b"
+    group_name_2a = "my-group-name-2a"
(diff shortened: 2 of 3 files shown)


tests/test_iot/test_iot.py
```diff
@@ -756,6 +756,140 @@ def test_delete_principal_thing():
...
+    resp = client.list_thing_groups(parentGroup=group_name_1a)
+    resp.should.have.key("thingGroups")
+    resp["thingGroups"].should.have.length_of(6)
+    resp = client.list_thing_groups(parentGroup=group_name_2a)
+    resp.should.have.key("thingGroups")
+    resp["thingGroups"].should.have.length_of(2)
+    resp = client.list_thing_groups(parentGroup=group_name_1b)
+    resp.should.have.key("thingGroups")
+    resp["thingGroups"].should.have.length_of(0)
+    try:
```
bblommers, line 842: Please use the `with assert_raises(ClientException) as e`, here, similar to https://github.com/spulec/moto/blob/111d0519f5bcde1fa915c3f3aefbae1c70b6c745/tests/test_config/test_config.py#L31

tests/test_iot/test_iot.py
```diff
@@ -756,6 +756,140 @@ def test_delete_principal_thing():
     client.delete_certificate(certificateId=cert_id)
 
 
+@mock_iot
+def test_list_thing_groups():
```
bblommers, line 760: Can you split out the various test cases in their own methods? The setup is all the same, so that can be extracted into it's own method as well.

Would make things a bit more readable

moto/iot/models.py
```diff
@@ -857,8 +857,30 @@ def delete_thing_group(self, thing_group_name, expected_version):
...
+        if recursive is None:
+            recursive = True
+        if name_prefix_filter is None:
+            name_prefix_filter = ""
+        if parent_group and parent_group not in [
+            _.thing_group_name for _ in self.thing_groups.values()
+        ]:
+            raise ResourceNotFoundException()
+        thing_groups = [
+            _ for _ in self.thing_groups.values() if _.parent_group_name == parent_group
```
bblommers, line 869: The `parent_group` argument is not mandatory - what happens if it's None? This if-statement reads like `thing_groups` would become an empty list - when it should return everything if parent_group is not supplied.

Chagui-: Its actually allowed for it to be None. When `parent_group` is None `thing_groups` would be the list of groups which have no parents. That is, the full list of groups on the first level of the tree. Now, as `recursive` is true by default, each of those groups return the data for the childs recursively, returning all the groups in the end. 
But don't take my word for it, its actually on a test case 😃

bblommers: Ah, gotcha. I saw the test case, but couldn't quite figure out how it worked :)

tests/test_iot/test_iot.py
Chagui-: I'll make a thing_group_tree_generator so its more flexible

tests/test_iot/test_iot.py
Chagui-: Done

tests/test_iot/test_iot.py
Chagui-: Apart form the generator, I ended up creating a test class to group all those tests together.

PR #2963, opened 2020-05-06
Fix ssm.get_parameters missing validation

### LINKED ISSUE #2883: Mocked SSM client can return more than 10 values from  API_GetParameters
Mocked ssm client can return more than 10 parameters which is an API limit.

https://docs.aws.amazon.com/systems-manager/latest/APIReference/API_GetParameter.html

This probably got overlooked because this is not mentioned in boto3 documentation!

https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/ssm.html#SSM.Client.get_parameters

Versions used
```
boto3==1.12.34
botocore==1.15.34
moto==1.3.14
```

## How to reproduce:

```python
import moto

@moto.mock_ssm
def test_resolve_12_ssm_parameters():

    from string import ascii_lowercase
    from random import choice

    input_ssm_parameters = {letter: (f"/{letter}/{letter}", "value") for letter in ascii_lowercase[:12]}
    ssm_client = boto3.client("ssm")
    param_types = ["String", "SecureString"]

    for key, (name, value) in input_ssm_parameters.items():
        ssm_client.put_parameter(Name=name, Value=value, Type=choice(param_types))

    param_names = {key: name for key, (name, value) in input_ssm_parameters.items()}
    response = ssm_client.get_parameters(Names=list(param_names.values()))

    assert len(response['Parameters']) == 12
```

## Expected 
The code above should fail with exception
```
botocore.exceptions.ClientError: An error occurred (ValidationException) when calling the GetParameters operation: 1 validation error detected: Value '[<list of parameter names>]' at 'names' failed to satisfy constraint: Member must have length less than or equal to 10.
```

## Actual
The fuction `test_resolve_12_ssm_parameters()` can pass.

gruebel:
Fixes #2883

### DIFF
--- a/moto/ssm/models.py
+++ b/moto/ssm/models.py
@@ -514,6 +514,16 @@ class SimpleSystemManagerBackend(BaseBackend):
 
     def get_parameters(self, names, with_decryption):
         result = []
+
+        if len(names) > 10:
+            raise ValidationException(
+                "1 validation error detected: "
+                "Value '[{}]' at 'names' failed to satisfy constraint: "
+                "Member must have length less than or equal to 10.".format(
+                    ", ".join(names)
+                )
+            )
+
         for name in names:
             if name in self._parameters:
                 result.append(self.get_parameter(name, with_decryption))
--- a/tests/test_ssm/test_ssm_boto3.py
+++ b/tests/test_ssm/test_ssm_boto3.py
@@ -1,5 +1,7 @@
 from __future__ import unicode_literals
 
+import string
+
 import boto3
 import botocore.exceptions
 import sure  # noqa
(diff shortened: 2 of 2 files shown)


bblommers:
Just noticed there are some conflicts - happy to merge after they are resolved

gruebel:
@bblommers I have no idea what happened, but you already merged my commits, so there was no conflict for me to be resolved. And somehow I closed the PR after I rebased it on the master. Can you also close the related issue #2883

bblommers:
@gruebel Ah, that makes sense actually. Sort of. I saw a Github unicorn when I tried to merge the first time, so I guess it only finished part of the merge.

PR #2966, opened 2020-05-06
SES get send statistics response modification

usmangani1:
@bblommers  Can you please merge this.A small change in response of get send statistics.


gruebel:
Hi @usmangani1, you have to format your code once. just run following command in the root folder
```
black moto/ tests/
```

PR #2970, opened 2020-05-06
Fix the online status in OpsWorks

rigaspapas:
When an instance is running, OpsWorks reports its status as "online" [1], while EC2 reports it as "running". Until now, moto copied the EC2 instance's status as is. This commit introduces the converts the running status to online when returned by OpsWorks.

[1]: https://docs.aws.amazon.com/cli/latest/reference/opsworks/describe-instances.html


rigaspapas:
Hi @bblommers, I extended the existing tests so the instance status is also checked. I force-pushed to my branch.

PR #2973, opened 2020-05-07
Lambda - Return LogResult only if requested

### LINKED ISSUE #2886: moto lambda response Payload is the log content, not function response data
See all details in https://github.com/aio-libs/aiobotocore/issues/793
- https://github.com/aio-libs/aiobotocore/pull/796 demonstrates the problem

Although I don't know how to reconcile that with the moto unit tests, i.e.
- https://github.com/spulec/moto/blob/master/tests/test_awslambda/test_lambda.py#L95-L125
- this appears to distinguish between the logs and payload

bblommers:
Fixes #2886 

Before: LogResult was always returned, and contained the Lambda response

Now: LogResult is only returned when LogType=Tail, and contains the actual logs

### DIFF
--- a/moto/awslambda/models.py
+++ b/moto/awslambda/models.py
@@ -379,6 +379,7 @@ class LambdaFunction(BaseModel):
             event = dict()
         if context is None:
             context = {}
+        output = None
 
         try:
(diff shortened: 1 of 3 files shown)


spulec:
Thanks!

PR #2974, opened 2020-05-07
Adding missing Thing Group ARN to FakeThingGroup

amir-hadi:
The `thingGroupArn` should be returned for Thing Groups, according to `boto3` [docs](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/iot.html#IoT.Client.create_thing_group).


bblommers:
Superceded by #3634 - will close this one

PR #2982, opened 2020-05-10
Add CloudWatch logs subscription filters

### LINKED ISSUE #2955: Feature: Logs: put_subscription_filter
moto version: v1.3.14

I try to deploy a lambda with log group and permissions and a dynamoDb table to moto with terraform.

I get 2 errors:

`NotImplementedError: The put_subscription_filter action has not been implemented`
and
`AttributeError: 'DynamoHandler' object has no attribute 'describe_continuous_backups'`

Additional it looks like I can't create a global secondary index.

gruebel:
Adds also log event transmission to specified destinations. At the moment only AWS Lambda is implemented. I had to move the `lambda_backends` import to the function level, because circular dependency issues.

I hope I correctly connected CloudWatch Logs with Lambda, if not please some feedback on which parts should be changed/fixed.

Fixes #2955

### DIFF
--- a/moto/awslambda/models.py
+++ b/moto/awslambda/models.py
@@ -5,6 +5,8 @@ import time
 from collections import defaultdict
 import copy
 import datetime
+from gzip import GzipFile
+
 import docker
 import docker.errors
 import hashlib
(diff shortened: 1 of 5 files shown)


moto/awslambda/models.py
```diff
@@ -983,6 +985,28 @@ def send_dynamodb_items(self, function_arn, items, source):
         func = self._lambdas.get_arn(function_arn)
         return func.invoke(json.dumps(event), {}, {})
 
+    def send_log_event(
+        self, function_arn, filter_name, log_group_name, log_stream_name, log_events
+    ):
+        data = {
+            "messageType": "DATA_MESSAGE",
+            "owner": "123456789012",
```
bblommers, line 993: Can  you use the ACCOUNT_ID from moto.core.models.py  here?

gruebel: Thanks, good point! Fixed and committed.

PR #2983, opened 2020-05-10
Managedblockchain member additions

jpbelleau:
Most functions complete - will send the rest in another PR

### DIFF
--- a/moto/managedblockchain/utils.py
+++ b/moto/managedblockchain/utils.py
@@ -79,9 +79,17 @@ def member_name_exist_in_network(members, networkid, membername):
     return membernamexists
 
 
-def number_of_members_in_network(members, networkid):
+def number_of_members_in_network(members, networkid, member_status=None):
     return len(
-        [membid for membid in members if members.get(membid).network_id == networkid]
+        [
+            membid
+            for membid in members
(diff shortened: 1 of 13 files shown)


moto/managedblockchain/utils.py
```diff
@@ -23,7 +28,73 @@ def get_network_id():
...
+    return membernamexists
+
+
+def number_of_members_in_network(members, networkid):
+    return len(
+        [membid for membid in members if members.get(membid).network_id == networkid]
+    )
+
+
+def admin_password_ok(password):
```
bblommers, line 89: Doesn't look like the return value means anything - lets just return a boolean here.
I guess the comment was more a reminder to yourself?

jpbelleau: Yeah, I thought it would be more complex with different exceptions for the cases - turns out it is not.  Will update and send an update later on.  Have a couple other additions once I skip a test during server mode tests (freeze_time and expired proposals).

jpbelleau: Updated

jpbelleau:
Finished member, proposal, and network functionality - no more additions to this PR.

jpbelleau:
Really fixed - needed future division

PR #2989, opened 2020-05-12
Improve Implementation coverage

bblommers:
Supercedes #2840 

Changes:

-  Renames @mock_ec2_instance_connect to @mock_ec2instanceconnect (in-line with other services with spaces/dashes)
-  Moves/renames some methods in DynamoDB to it's backend to ensure the implementation-coverage script picks up all implemented methods
-  Change the styling to add each service in it's own collapsible section. This makes it slightly more readable IMO: https://github.com/bblommers/moto/blob/improve_coverage/IMPLEMENTATION_COVERAGE.md

TODO: 
These services need the same fix, and move some logic towards the model, as the script isnt aware of all implemented methods at the moment:
 - S3
 - XRay

And maybe others as well. Will probably create a separate PR for them.

FYI @spulec


spulec:
Looks good!

PR #2994, opened 2020-05-15
Cloud formation "depends_on" #2845 Add depends on and update name type mapping

### LINKED ISSUE #2845: Cloud formation "depends_on"
When your cloud formation parser is lazily iterating through the provided resources, I don't think it's taking the depends_on variable into account, which is causing resources to be created in the wrong order, and ends up failing a cloud formation script. The example that I came up with was creating an autoscaling group before the launch configuration was created, even though the autoscaling group has a depends on flag for the launch configuration

zbrookle:
This pull request will fix creating CloudFormation scripts that specify dependencies using DependsOn, and also adds in updated names for resources that are nameable because many were missing and some were incorrect. Note that there is also one undocumented one that I added and there may be more of these undocumented ones in the future because it seems that this webpage https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-name.html
isn't completely up to date
Fixes #2845


zbrookle:
@bblommers Sure! I'll add those in

PR #2996, opened 2020-05-16
Athena - Start/Stop query execution, get_work_group

bblommers:
Initial (basic) implementation of some Athena-methods.


spulec:
Nice; thanks!

Issue #2997, opened 2020-05-16
Add mock describe_network_acls


anardy:
Hi,

Thank you for creating this library. I'm writting my tests and I need a `describe_network_acls`, but according to [IMPLEMENTATION_COVERAGE](https://github.com/spulec/moto/blob/master/IMPLEMENTATION_COVERAGE.md) these are not yet implemented. The [CONTRIBUTING](https://github.com/spulec/moto/blob/master/CONTRIBUTING.md) document asks me to create an issue first.

I'm trying to write this method, but will probably need some guidance along the way. As soon as possible, I will send PR referencing this issue.

bblommers:
Hi @anardy, if I'm not mistaken, this method [is already implemented](https://github.com/spulec/moto/blob/master/moto/ec2/responses/network_acls.py#L83) - the coverage document might just be out of date.

There is no test for it yet, though, so if you want to make a PR to add a test case and update the coverage doc that's very welcome! :)

anardy:
Thank you for reply and I'm sorry for the confusion. I thought that the method implemention should also modify the `models.py` file.

Follow below the test I wrote and I would like to know if it's correct.

```
@mock_ec2
def test_describe_network_acls():
    conn = boto3.client("ec2", region_name="us-west-2")

    vpc = conn.create_vpc(CidrBlock="10.0.0.0/16")
    vpc_id = vpc["Vpc"]["VpcId"]

    network_acl = conn.create_network_acl(VpcId=vpc_id)

    network_acl_id = network_acl["NetworkAcl"]["NetworkAclId"]

    resp = conn.describe_network_acls(NetworkAclIds=[network_acl_id])

    result = resp["NetworkAcls"]

    result.should.have.length_of(1)
    result[0]["NetworkAclId"].should.equal(network_acl_id)
```

bblommers:
Hi @anardy, you're right, it should be in models.py, ideally.

The test looks good to me!

bblommers:
Closing - this is now part of moto >= 1.3.15.dev875

PR #3003, opened 2020-05-18
Enhancement/describe network acls

anardy:
Related to issue #2997

Includes:

- Change the method from `get_all_network_acls` to `describe_network_acls`
- Add unit test for this new method

I hope that I have implemented correctly.

### DIFF
--- a/tests/test_ec2/test_network_acls.py
+++ b/tests/test_ec2/test_network_acls.py
@@ -295,7 +295,7 @@ def test_describe_network_acls():
     result[0]["NetworkAclId"].should.equal(network_acl_id)
 
     resp2 = conn.describe_network_acls()["NetworkAcls"]
-    resp2.should.have.length_of(len(resp2))
+    resp2.should.have.length_of(1)
 
     with assert_raises(ClientError) as ex:
         conn.describe_network_acls(NetworkAclIds=["1"])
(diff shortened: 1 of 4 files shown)


tests/test_ec2/test_network_acls.py
```diff
@@ -275,3 +275,32 @@ def test_duplicate_network_acl_entry():
...
+    network_acl_id = network_acl["NetworkAcl"]["NetworkAclId"]
+
+    resp = conn.describe_network_acls(NetworkAclIds=[network_acl_id])
+    result = resp["NetworkAcls"]
+
+    result.should.have.length_of(1)
+    result[0]["NetworkAclId"].should.equal(network_acl_id)
+
+    resp2 = conn.describe_network_acls()["NetworkAcls"]
+    resp2.should.have.length_of(len(resp2))
```
bblommers, line 298: This should probably be `should.have.length_of(1)` :)

anardy: @bblommers Thank you for your reply and attention.

bblommers: Well, don't thank me yet - looks like the tests are failing now because of this change!
https://travis-ci.org/github/spulec/moto/builds/688611941

anardy: Hi @bblommers. I found out what is happening.

When you execute `create_network_acl` method, are created 2 Network ACLs. The first is marked as a default and other is not.
When AWS account is created, consequently the default Network ACL is also created.
Therefore, in this test scenario the result should be 3.
This can also be seen in the test the `create_network_acl` method where the result is 3, [follow the code](https://github.com/spulec/moto/blob/80b64f9b3ff57515db1fc07329bf8e5f519597aa/tests/test_ec2/test_network_acls.py#L20).

PR #3007, opened 2020-05-18
SFN - fix InvalidARN exception on start_execution

### LINKED ISSUE #3005: Certain StepFunctions exceptions uncatchable
I am having an issue catching certain exceptions that are being thrown by only some of the StepFunction related methods.  I have managed to break it down to this simple example to illustrate my problem:

```
import pytest
import moto
from moto import mock_stepfunctions
import boto3

def test_sfn_describe_state_machine():
    with mock_stepfunctions():
        with pytest.raises(ClientError):
            client = boto3.client("stepfunctions", region_name="ap-southeast-2")
            client.describe_state_machine(stateMachineArn="asd")

def test_sfn_start_execution():
    with mock_stepfunctions():
        with pytest.raises(ClientError):
            client = boto3.client("stepfunctions", region_name="ap-southeast-2")
            client.start_execution(stateMachineArn="asd")

```

Running this with pytest (or anything, that part isn't relevant) will result in the first test passing, with the `ClientError` successfully being caught.  The second test will fail.  From what I can see, the two methods are raising two different exceptions:

1st) `botocore.errorfactory.InvalidArn: An error occurred (InvalidArn) when calling the DescribeStateMachine operation: Invalid State Machine Arn: 'asd'`
2nd) `moto.stepfunctions.exceptions.InvalidArn: Invalid State Machine Arn: 'asd'`

From what I can see, they're both calling the same code path to validate the ARN, but only one is raising exception namespace to `moto`.  I've also had a look at the tests in the `moto` repository itself and added one that follows the pattern of my failing test, and it fails there too - so I don't think this is an environment issue.

I'm sure it's something obvious, but it's got my stumped at the moment.

aidan-:
This fixes #3005 .

I also fixed a duplicate test name, but I don't think that has any affect on anything.

### DIFF
--- a/moto/stepfunctions/responses.py
+++ b/moto/stepfunctions/responses.py
@@ -95,7 +95,10 @@ class StepFunctionResponse(BaseResponse):
     def start_execution(self):
         arn = self._get_param("stateMachineArn")
         name = self._get_param("name")
-        execution = self.stepfunction_backend.start_execution(arn, name)
+        try:
+            execution = self.stepfunction_backend.start_execution(arn, name)
+        except AWSError as err:
+            return err.response()
         response = {
             "executionArn": execution.execution_arn,
             "startDate": execution.start_date,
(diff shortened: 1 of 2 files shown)


aidan-:
@bblommers fixed and squashed commits to avoid noise!

aidan-:
@bblommers Looks like the build is still failing, I think it's caused by this test:

```
test_auth.test_access_denied_for_run_instances ... FAIL
```
But this PR shouldn't be affecting that code..

Issue #3008, opened 2020-05-19
Meta - Build fails with botocore 1.16.12


bblommers:
Failing test:
https://travis-ci.org/github/spulec/moto/jobs/688592478

Same problem as https://github.com/spulec/moto/issues/2774 - opening a new issue for visibility.

I've asked for some clarification in the botocore-project:
https://github.com/boto/botocore/issues/2051

PR #3010, opened 2020-05-19
Fix autoscaling tags

### LINKED ISSUE #3009: CloudFormation autoscaling group PropagateAtLaunch not setting properly
When creating a CloudFormation autoscaling group, PropagateAtLaunch for tags is being reported as 'False' even though it was set to 'True' in the template

zbrookle:
Fixes #3009


zbrookle:
Fixes #3009

zbrookle:
@spulec Is there any way that you could re run the pipeline? It only failed on python 2.7 and I'm fairly certain that the test that failed is related to a network issue rather than a change I made

aidan-:
@zbrookle Build failure is probably related to #3008 .  My PR build (#3007) is failing with the same issue.

bblommers:
That's correct, @aidan-. Hopefully we can get the build fixed in #3008 soon - after that I'll merge both your PRs

zbrookle:
@bblommers Ok great sounds good!

PR #3019, opened 2020-05-23
Corrected bug in IAM delete_role() due to overloading of name 'role' …

jweite:
Re Issue 3016


jweite:
Should be good to go now Bert.  Thanks for the feedback.

PR #3020, opened 2020-05-23
Implemented IAM delete_instance_profile

jweite:
Issue 3017

### DIFF
--- a/tests/test_iam/test_iam.py
+++ b/tests/test_iam/test_iam.py
@@ -207,6 +207,24 @@ def test_remove_role_from_instance_profile():
 
 
 @mock_iam()
+def test_delete_instance_profile():
+    conn = boto3.client("iam", region_name="us-east-1")
+    conn.create_role(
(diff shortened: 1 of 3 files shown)


tests/test_iam/test_iam.py
```diff
@@ -206,6 +206,24 @@ def test_remove_role_from_instance_profile():
...
+    conn.add_role_to_instance_profile(
+        InstanceProfileName="my-profile", RoleName="my-role"
+    )
+    with assert_raises(conn.exceptions.DeleteConflictException):
+        conn.delete_instance_profile(InstanceProfileName="my-profile")
+    conn.remove_role_from_instance_profile(
+        InstanceProfileName="my-profile", RoleName="my-role"
+    )
+    conn.delete_instance_profile(InstanceProfileName="my-profile")
+
```
bblommers, line 225: Can you add some assertions here that the instance_profile was actually deleted?
I guess simply be calling get_instance_profile and verifying that it now fails

jweite:
Should be good now Bert.  Thanks for the review and feedback.

Issue #3018, opened 2020-05-23
Need s3 methods get_bucket_encryption, put_bucket_encryption, delete_bucket_encryption


jweite:
Not implemented in 1.3.14 or master.

bblommers:
This should be available as of moto >= 1.3.15.dev877

PR #3022, opened 2020-05-23
implemented s3 default encryption methods

jweite:
Issue #3018


moto/s3/responses.py
```diff
@@ -466,6 +466,13 @@ def _bucket_response_get(self, bucket_name, querystring):
                     is_truncated="false",
                 ),
             )
+        elif "encryption" in querystring:
+            bucket = self.backend.get_bucket(bucket_name)
```
bblommers, line 470: Would you mind adding the logic to a `the get_bucket_encryption` in models.py? Makes this a bit more readable.
```
encryption = self.backend.get_bucket_encryption(..)
if not encryption..
```

PR #3028, opened 2020-05-26
Fix: Cloudwatch delete Alarm status code handling on invalid alarm name

usmangani1:
https://github.com/localstack/localstack/issues/2447

@bblommers  Need your help in getting this small PR merged.


bblommers:
Hey @usmangani1, thanks for your PR! From the docs: `In the event of an error, no alarms are deleted.`
So I guess we'll have loop twice, first to check that all alarms exist, and another loop to actually delete.
Would be good as well to have a test case for this scenario (one call with multiple alarms, some existent and some non-existent)

bblommers:
This is now available in moto >= 1.3.15.dev880

PR #3029, opened 2020-05-27
Fix: SecretsManager Added VersionIdsToStages key in describe_secret function

usmangani1:
https://github.com/localstack/localstack/issues/2429

@bblommers  Need your help in getting this small change merged

### DIFF
--- a/tests/test_secretsmanager/test_secretsmanager.py
+++ b/tests/test_secretsmanager/test_secretsmanager.py
@@ -733,19 +733,22 @@ def test_put_secret_value_versions_differ_if_same_secret_put_twice():
 def test_put_secret_value_maintains_description_and_tags():
     conn = boto3.client("secretsmanager", region_name="us-west-2")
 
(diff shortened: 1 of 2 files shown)


tests/test_secretsmanager/test_secretsmanager.py
```diff
@@ -752,6 +752,7 @@ def test_put_secret_value_maintains_description_and_tags():
         {"Key": "Mykey", "Value": "Myvalue"},
     ]
     assert secret_details["Description"] == "desc"
+    assert secret_details["VersionIdsToStages"] is not None
```
bblommers, line 755: Could you assert that it matches the expected dict, whatever that is?

There's quite a bit of logic around versions/stages (which I'm not sure is completely implemented..), so it would be good to be specific in what moto currently supports.

bblommers:
This is now available as of moto >= 1.3.15.dev882

PR #3036, opened 2020-05-29
Deserialize List sub Map attribute correctly

jmphilli:


bblommers:
Hi @jmphilli, what's the issue you're trying to address/fix here?

PR #3037, opened 2020-06-01
Fix incorrect response for put-targets action

victou:
Hi,

In the AWS documentation, events put-targets action should return something like this:

```
[
    'FailedEntries' => [
        [
            'ErrorCode' => '<string>',
            'ErrorMessage' => '<string>',
            'TargetId' => '<string>',
        ],
        // ...
    ],
    'FailedEntryCount' => <integer>,
]
```
https://docs.aws.amazon.com/cli/latest/reference/events/put-targets.html#output
https://docs.aws.amazon.com/aws-sdk-php/v3/api/api-events-2015-10-07.html#puttargets

### DIFF
--- a/moto/events/responses.py
+++ b/moto/events/responses.py
@@ -217,7 +217,10 @@ class EventsHandler(BaseResponse):
                 "ResourceNotFoundException", "Rule " + rule_name + " does not exist."
             )
 
-        return "", self.response_headers
+        return (
+            json.dumps({"FailedEntryCount": 0, "FailedEntries": []}),
+            self.response_headers,
+        )
 
     def remove_targets(self):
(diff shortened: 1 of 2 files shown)


victou:
Hi @bblommers ,

ok i will add test case and run black on this file.

Thanks.

victou:
> Hi @victou, thanks for your PR! Can you add a small test case to verify this format is being returned correctly?
> 
> Looks like the linting is failing as well, so you might have to run black on this file to get the CI build passing: https://travis-ci.org/github/spulec/moto/jobs/693390887

it's done.

bblommers:
This is now part of moto >= 1.3.15.dev883

Issue #3038, opened 2020-06-02
Changes are missing when creating a change set


jeremietharaud:
**moto version:**
1.3.15 (spulec/moto: master)

**module:**
mock_cloudformation

**action:**
`create_change_set`
When we create a change set and describe it using the following code, the result is not the same as running using boto3:
```
dummy_template = {
    "AWSTemplateFormatVersion": "2010-09-09",
    "Description": "Stack 1",
    "Resources": {
        "BucketResource": {
            "Type": "AWS::S3::Bucket",
            "Properties": {
                "BucketName": "moto-test-123456789",
            },
        }
    },
}

dummy_template_json = json.dumps(dummy_template)

@mock_cloudformation
def test_change_set():
    cf_conn = boto3.client("cloudformation", region_name="eu-west-1")
    cf_conn.create_change_set(
        StackName="NewStack",
        TemplateBody=dummy_template_json,
        ChangeSetName="NewChangeSet",
        ChangeSetType="CREATE",
    )

    print(cf_conn.describe_change_set(ChangeSetName="NewChangeSet"))

if __name__ == '__main__':
    test_change_set()
```

**expected behavior:**
```
aws cloudformation describe-change-set --change-set-name NewChangeSet --stack-name NewStack
{
    "Changes": [
        {
            "Type": "Resource",
            "ResourceChange": {
                "Action": "Add",
                "LogicalResourceId": "BucketResource",
                "ResourceType": "AWS::S3::Bucket",
                "Scope": [],
                "Details": []
            }
        }
    ],
    "ChangeSetName": "NewChangeSet",
    "ChangeSetId": "arn:aws:cloudformation:eu-west-1:664685189390:changeSet/NewChangeSet/ae79ab51-a4a8-4f4f-8306-af6117f2a1ef",
    "StackId": "arn:aws:cloudformation:eu-west-1:664685189390:stack/NewStack/0b05cf30-a0e0-11ea-9051-0aa5f1c3ad6c",
    "StackName": "NewStack",
    "Description": "Stack 1",
    "Parameters": null,
    "CreationTime": "2020-05-28T12:37:50.669Z",
    "ExecutionStatus": "AVAILABLE",
    "Status": "CREATE_COMPLETE",
    "StatusReason": null,
    "NotificationARNs": [],
    "RollbackConfiguration": {},
    "Capabilities": [],
    "Tags": null
}
```

**actual behavior:**
```
{
  'ChangeSetName': 'NewChangeSet',
[...]

PR #3039, opened 2020-06-02
Add missing changes when creating a change set

jeremietharaud:
Hi,

This PR fixes issue #3038 by adding the changes property when describing a change set.
Example:
```
{
    "Changes": [
        {
            "Type": "Resource",
            "ResourceChange": {
                "Action": "Add",
                "LogicalResourceId": "BucketResource",
                "ResourceType": "AWS::S3::Bucket",
                "Scope": [],
                "Details": []
            }
        }
    ],
...
}
```

When creating a change set, ``stack_template`` variable has to be initialized withan empty dict instead of ``template`` variable because the stack is not created at this time. If we don't do that, the call to ``diff`` function gives nothing (e.g. no change) because it compares the stack_template to itself.

### DIFF
--- a/moto/cloudformation/models.py
+++ b/moto/cloudformation/models.py
@@ -219,7 +219,12 @@ class FakeStack(BaseModel):
         self.stack_id = stack_id
         self.name = name
         self.template = template
-        self._parse_template()
+        if template != {}:
+            self._parse_template()
+            self.description = self.template_dict.get("Description")
+        else:
+            self.template_dict = {}
+            self.description = None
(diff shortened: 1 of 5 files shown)


jeremietharaud:
Hi @bblommers, I fixed the linting issue.
Thank you.

jeremietharaud:
> Looks like the build fails because of linting - happy to merge once that passes
> 
> https://travis-ci.org/github/spulec/moto/jobs/693774396

OK fixed it

PR #3040, opened 2020-06-02
Make sure the UTC tz is included in the bucket creation timestamp

### LINKED ISSUE #2665: S3 CreationDate not correctly formatted
I think there's a small regression in version 1.3.15.3 of the library, as used by Localstack.

This appears to be generating incorrectly formatted `CreationDate` fields in S3 ListBucket responses, which breaks some clients (AWS Java SDK chokes on the formatting; AWS CLI does not).

More background is captured in this Localstack issue: https://github.com/localstack/localstack/issues/1899

I suspect this issue can be simply fixed by amending this line to format the creation date according to ISO8601: https://github.com/spulec/moto/commit/ab0853cddc6bf9a0789999751a5d698186b8ebd0#diff-f28b3d402b1d050e247f518f3b94865bR1313

timvlaer:
Fix for the TZ in `CreationDate` of s3 buckets, fixes #2665 (similar to the fix for #2904 )

`bucket.creation_date.isoformat()` doesn't include timezone info if it's utc. The AWS Java SDK fails on this missing field:

```
com.amazonaws.util.DateUtils.parseISO8601Date("2020-06-02T08:43:57.470630")
// java.lang.IllegalArgumentException: Invalid format: "2020-06-02T08:43:57.470630" is too short
```

### DIFF
--- a/moto/s3/models.py
+++ b/moto/s3/models.py
@@ -5,6 +5,7 @@ import json
 import os
 import base64
 import datetime
+import pytz
 import hashlib
 import copy
 import itertools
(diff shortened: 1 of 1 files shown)


spulec:
Thanks!

PR #3045, opened 2020-06-03
Cloudformation support for EventSourceMapping

### LINKED ISSUE #3044: EventSourceMapping cloudformation
After somer research I noticed there's the model `EventSourceMapping` has no support for cloudformation's actions.

gmcrocetti:
Fixes #3044. Did some refactoring to improve readability.

It would be great to hear some tips about these [lines](https://github.com/spulec/moto/pull/3045/files#diff-2df06bdd95e3c78b9335da9f260a3472R650-R653)

### DIFF
--- a/moto/awslambda/models.py
+++ b/moto/awslambda/models.py
@@ -558,32 +558,8 @@ class EventSourceMapping(BaseModel):
         self.function_arn = spec["FunctionArn"]
         self.event_source_arn = spec["EventSourceArn"]
 
-        # BatchSize service default/max mapping
-        batch_size_map = {
-            "kinesis": (100, 10000),
-            "dynamodb": (100, 1000),
-            "sqs": (10, 10),
-        }
-        source_type = self.event_source_arn.split(":")[2].lower()
-        batch_size_entry = batch_size_map.get(source_type)
-        if batch_size_entry:
-            # Use service default if not provided
-            batch_size = int(spec.get("BatchSize", batch_size_entry[0]))
-            if batch_size > batch_size_entry[1]:
-                raise ValueError(
-                    "InvalidParameterValueException",
-                    "BatchSize {} exceeds the max of {}".format(
-                        batch_size, batch_size_entry[1]
-                    ),
-                )
-            else:
-                self.batch_size = batch_size
-        else:
-            raise ValueError(
-                "InvalidParameterValueException", "Unsupported event source type"
-            )
-
         # optional
+        self.batch_size = spec.get("BatchSize")
         self.starting_position = spec.get("StartingPosition", "TRIM_HORIZON")
         self.enabled = spec.get("Enabled", True)
         self.starting_position_timestamp = spec.get("StartingPositionTimestamp", None)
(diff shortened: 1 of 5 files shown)


moto/awslambda/models.py
```diff
@@ -548,40 +548,63 @@ def delete(self, region):
...
 
-        # optional
-        self.starting_position = spec.get("StartingPosition", "TRIM_HORIZON")
-        self.enabled = spec.get("Enabled", True)
-        self.starting_position_timestamp = spec.get("StartingPositionTimestamp", None)
+        source_type = self._get_service_source_from_arn(self.event_source_arn)
+        batch_size_for_source = batch_size_service_map[source_type]
+
+        if batch_size is None:
+            self._batch_size = batch_size_for_source[0]
```
bblommers, line 600: This refactoring changes the logical flow - before we'd throw an error if batch_size is None.

moto/awslambda/models.py
```diff
@@ -595,23 +618,42 @@ def get_configuration(self):
...
+        lambda_backend = lambda_backends[region_name]
+        esms = lambda_backend.list_event_source_mappings(
+            event_source_arn=properties["EventSourceArn"],
+            function_name=properties["FunctionName"],
+        )
+
+        for esm in esms:
+            if esm.logical_resource_id in resource_name:
+                lambda_backend.delete_event_source_mapping
+                esm.delete(region_name)
```
bblommers, line 656: I'm guessing you want to delete the `resource` here, not the region :)

moto/awslambda/models.py
gmcrocetti: Yes, indeed ! This was one of the facts which motivated my refactoring. We shouldn't validate if the type of the eventsource's arn is correct during batch size's attribution and that's why I've created [this](https://github.com/spulec/moto/pull/3045/files#diff-2df06bdd95e3c78b9335da9f260a3472R577). What you think ?

moto/awslambda/models.py
gmcrocetti: Maybe I missunderstood but this deletes the resource. The problem here is doing this "crazy" `in` condition veryfing if it's possible to delete based on names.

gmcrocetti:
> Hi @gmcrocetti, thanks for your PR!
> 
> Left some review comments. Looks like the build is failing as well: https://travis-ci.org/github/spulec/moto/jobs/694093245

Hi @bblommers . Thanks for you impressive fast review ! 👍 
We're facing two errors which I'll try to explain:
1. test_lambda.test_update_event_source_mapping. The old solution responsible to validate batch's size was doing it only once (at instanciation time). We're setting the batch size for a sqs source to 15, which is not allowed, thus, easy to solve;
2. tests.test_cloudformation.test_cloudformation_stack_crud.test_create_stack_lambda_and_dynamodb. This one is harder and hope to hear from you. There's no ddb stream created at the moment we're creating our EventSource. This error was "masked" due to the try block in the test.

moto/awslambda/models.py
bblommers: Ah, right, I understand. Your solution is way more elegant then 👍

moto/awslambda/models.py
bblommers: Yup, gotcha. I should've checked the signature of the delete-method - I assumed you wanted to write `esm.delete(resource_name)`. All good.

gmcrocetti:
>     1. It doesn't look like our CF implementation even supports creating a DynamoDB stream. Leave that with me for now - I'll try and implement that later this week.

Thanks @bblommers . Fixed 1.

PR #3047, opened 2020-06-03
#3046 - DynamoDB - Add Key-size Validation for BatchGetItem

### LINKED ISSUE #3046: DynamoDB - BatchGetItem should throw an error when requesting more than 100 items
Docs: 
https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/dynamodb.html#DynamoDB.Client.batch_get_item

> If you request more than 100 items, BatchGetItem returns a ValidationException with the message "Too many items requested for the BatchGetItem call."

The docs fail to mention that there are two different use cases: request more than 100 keys for a single table, or request more than a 100 keys across all tables. DDB throws different error messages for both scenarios.

PR that handles both cases incoming shortly.

bblommers:
Fixes #3046

### DIFF
--- a/moto/dynamodb2/responses.py
+++ b/moto/dynamodb2/responses.py
@@ -371,6 +371,26 @@ class DynamoHandler(BaseResponse):
 
         results = {"ConsumedCapacity": [], "Responses": {}, "UnprocessedKeys": {}}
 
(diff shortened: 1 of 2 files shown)


spulec:
Thanks!

PR #3049, opened 2020-06-03
chore: refine python-jose dependency

mattsb42-aws:
Starting in 3.1.0[1], `python-jose` isolated all cryptographic operations into independent backends[2]. The unused dependencies cannot be easily removed, but selecting the `[cryptography]` extra installs the dependencies for the `pyca/cryptography` backend and forces all cryptographic operations onto that backend.

[1] https://github.com/mpdavis/python-jose/blob/master/CHANGELOG.md#310----2019-12-10
[2] https://github.com/mpdavis/python-jose#cryptographic-backends

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -41,7 +41,7 @@ install_requires = [
     "PyYAML>=5.1",
     "pytz",
     "python-dateutil<3.0.0,>=2.1",
-    "python-jose<4.0.0",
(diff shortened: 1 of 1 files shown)


mattsb42-aws:
TLDR Travis reference: It looks like the Python 3.6 server mode test run is failing with some patching errors that seem to be unrelated to this dependency. All other test runs are succeeding (including non-server-mode 3.6).

PR #3053, opened 2020-06-06
Improve implementation coverage

bblommers:
Moves/renames some S3 methods to ensure that the coverage script picks them up - no logic changes.


spulec:
Thanks!

PR #3058, opened 2020-06-08
Enhancement : API-Gateway Put Integration Response - Adding support f…

usmangani1:
API-Gateway Put Integration Response - Adding support for contentHandling.

https://github.com/localstack/localstack/issues/2273

### DIFF
--- a/moto/apigateway/responses.py
+++ b/moto/apigateway/responses.py
@@ -387,6 +387,7 @@ class APIGatewayResponse(BaseResponse):
             elif self.method == "PUT":
                 selection_pattern = self._get_param("selectionPattern")
                 response_templates = self._get_param("responseTemplates")
+                content_handling = self._get_param("contentHandling")
                 integration_response = self.backend.create_integration_response(
                     function_id,
                     resource_id,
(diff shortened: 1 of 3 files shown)


usmangani1:
> API-Gateway Put Integration Response - Adding support for contentHandling.
> 
> [localstack/localstack#2273](https://github.com/localstack/localstack/issues/2273)

@bblommers  Need your help in getting this small enhancement merged,

moto/apigateway/responses.py
```diff
@@ -387,13 +387,15 @@ def integration_responses(self, request, full_url, headers):
             elif self.method == "PUT":
                 selection_pattern = self._get_param("selectionPattern")
                 response_templates = self._get_param("responseTemplates")
+                content_handling = self._get_param("contentHandling")
```
bblommers, line 390: According to [the docs](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/apigateway.html#APIGateway.Client.put_integration_response), this param is not mandatory

tests/test_apigateway/test_apigateway.py
```diff
@@ -543,7 +543,9 @@ def test_integration_response():
         statusCode="200",
         selectionPattern="foobar",
         responseTemplates={},
+        contentHandling="CONVERT_TO_BINARY"
```
bblommers, line 546: Can you assert that the param can be retrieved as well? (In one of the tests - doesn't have to be this particular one.)

Also, please remove the contentHandling-param for some of the tests, to verify it's not mandatory.

bblommers:
This is now available in moto >= 1.3.15.dev888

PR #3062, opened 2020-06-10
Fix: Support streaming upload from requests.

g-cassie:
Resolves a bug where streaming uploads made via requests to a presigned s3 url were failing because their request bodies are file-like objects.

See https://2.python-requests.org/en/master/user/advanced/#streaming-uploads for details of this feature in `requests`.


bblommers:
This is now part of moto >= 1.3.15.dev891

PR #3064, opened 2020-06-10
Instance object never sets iam_instance_profile

albinvass:
Hi!

I was trying to set up some tests that makes sure an instance profile is attached to an ec2 instance on creation but ran into the issue that the test object was missing the iam_instance_profile attribute.

I added a small test to show what's fails with the following TraceBack:

```
  File "/home/avass/git/github/albinvass/moto/.tox/py37/lib/python3.7/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/home/avass/git/github/albinvass/moto/moto/core/models.py", line 98, in wrapper
  File "/home/avass/git/github/albinvass/moto/tests/test_ec2/test_instances.py", line 219, in test_instance_iam_instance_profile
    assert instance_profile_arn == instance.iam_instance_profile["Arn"]
TypeError: 'NoneType' object is not subscriptable
```

### DIFF
--- a/tests/test_ec2/test_instances.py
+++ b/tests/test_ec2/test_instances.py
@@ -211,7 +211,6 @@ def test_instance_iam_instance_profile():
         MinCount=1,
         MaxCount=1,
         IamInstanceProfile={
-            "Name": "fake-instance-profile",
             "Arn": instance_profile_arn,
         },
     )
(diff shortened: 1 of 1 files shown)


bblommers:
Thanks for raising this, and providing a test case, @albinvass!

bblommers:
This functionality is now implemented with #5617

PR #3072, opened 2020-06-14
CloudFormation - Check stack name in use

### LINKED ISSUE #1954: CloudFormation allows for creating stacks with duplicate names

bblommers:
Fixes #1954 

Based on work by @grahamlyons in #1956, which can be closed if this PR is approved

### DIFF
--- a/moto/cloudformation/responses.py
+++ b/moto/cloudformation/responses.py
@@ -50,6 +50,12 @@ class CloudFormationResponse(BaseResponse):
             for item in self._get_list_prefix("Tags.member")
         )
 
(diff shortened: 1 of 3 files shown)


spulec:
Thank you!

PR #3073, opened 2020-06-14
Improve parsing of string values that represents booleans during block device mapping construction

sf-mkirilov:
This PR fixes a bug that I introduced during https://github.com/spulec/moto/pull/2999


sf-mkirilov:
> Hey @sf-mkirilov, looks like this changes fail on Python2:
> https://travis-ci.org/github/spulec/moto/jobs/698178912

Indeed, fixed.

PR #3074, opened 2020-06-14
EC2 : Fix - modified volume type in ec2 describe images.

usmangani1:
EC2 : Fix - modified volume type in ec2 describe images.
https://github.com/localstack/localstack/issues/2551

### DIFF
--- a/moto/ec2/responses/amis.py
+++ b/moto/ec2/responses/amis.py
@@ -125,7 +125,7 @@ DESCRIBE_IMAGES_RESPONSE = """<DescribeImagesResponse xmlns="http://ec2.amazonaw
                 <snapshotId>{{ image.ebs_snapshot.id }}</snapshotId>
                 <volumeSize>15</volumeSize>
                 <deleteOnTermination>false</deleteOnTermination>
-                <volumeType>{{ image.root_device_type }}</volumeType>
+                <volumeType>standard</volumeType>
               </ebs>
             </item>
           </blockDeviceMapping>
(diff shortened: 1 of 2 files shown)


usmangani1:
> EC2 : Fix - modified volume type in ec2 describe images.
> [localstack/localstack#2551](https://github.com/localstack/localstack/issues/2551)

@bblommers ,Need your help in getting this small PR merged.

bblommers:
Where does the volume-type come from, from the user in LocalStack?
The default value is 'standard' (https://github.com/spulec/moto/blob/master/moto/ec2/models.py#L1299) - as far as I can see, other values will always be specified by the user

I agree there should be some validation, to ensure that values are always one of `standard | io1 | gp2 | sc1 | st1` - but I don't think restricting the value to 'standard' is the right solution.

usmangani1:
> Where does the volume-type come from, from the user in LocalStack?
> The default value is 'standard' (https://github.com/spulec/moto/blob/master/moto/ec2/models.py#L1299) - as far as I can see, any values that are returned will always be specified by the user when creating an AMI/instance
> 
> I agree there should be some validation, to ensure that values are always one of `standard | io1 | gp2 | sc1 | st1` - but I don't think restricting the value to 'standard' is the right solution.

@bblommers 
I think we are passing root_device_type as volume_type and some of the ami's present in this location have ebs as root_device_type and hence moto is giving out ebs as volume_type. 
https://github.com/spulec/moto/blob/master/moto/ec2/resources/amis.json#L26

Can you please recommend on what has to be done here ?

bblommers:
@usmangani1 Ah, I see - didn't realize they were two separate attributes. We're not doing anything with volumeType at the moment in EC2, so setting it to `standard` is fine for me.

I've just linted, so assuming the build now passes I'll merge in a couple of hours 👍

bblommers:
This is now available in moto >= 1.3.15.dev911

PR #3077, opened 2020-06-17
fix to use json for yaml scanner error

shano:
Valid json can have `\t` characters, this currently throws a yaml scanner error so json.loads never gets called instead.

Also the loader arg in json.loads is invalid.

Can add tests if needed to exemplify this.


shano:
Sure @bblommers! I've just added a test whereby a json string with `\t` characters is validly passed to the json.loads function rather than yaml throwing an unexpected exception.

Let me know if I should change anything. 👍

bblommers:
This is now available in moto >= 1.3.15.dev913

PR #3083, opened 2020-06-20
S3: Delete Object Tagging

### LINKED ISSUE #1732: delete_object_tagging deletes the entire object
Expected behaviour:
Issuing delete_object_tagging on an object to delete the tags on the object but not the object itself.

Actual behaviour:
Object in S3 is actually deleted when the tags are deleted

Test case:
```
@mock_s3
def test_get_put_delete_object_tagging_boto3():
    makedir("/tmp/test", "/tmp/test")
    local_file = f"/tmp/test/local"
    touch(local_file)
    session = boto3.Session(profile_name="moto", region_name="eu-west-1")
    s3_client = session.client('s3')
    s3_client.create_bucket(Bucket="bucket")
    s3_client.put_object(
        Bucket="bucket",
        Key="remote",
        Body="".encode('utf8')
    )

    print("Listing before deleting tags")
    for obj in s3_client.list_objects_v2(Bucket="bucket").get("Contents", []):
        print(obj.get("Key"))

    s3_client.delete_object_tagging(
        Bucket="bucket",
        Key="remote"
    )

    print("Listing after deleting tags")
    for obj in s3_client.list_objects_v2(Bucket="bucket").get("Contents", []):
        print(obj.get("Key"))
```

Output:
```
Listing before deleting tags
remote
Listing after deleting tags
```
I was expecting to still see the file there after removing the tags.
I'm using the python mocks.

Let me know if you need more information.

bblommers:
Fixes #1732

### DIFF
--- a/IMPLEMENTATION_COVERAGE.md
+++ b/IMPLEMENTATION_COVERAGE.md
@@ -7093,7 +7093,7 @@
 - [X] delete_bucket_tagging
 - [ ] delete_bucket_website
 - [X] delete_object
-- [ ] delete_object_tagging
+- [x] delete_object_tagging
 - [ ] delete_objects
 - [ ] delete_public_access_block
 - [ ] get_bucket_accelerate_configuration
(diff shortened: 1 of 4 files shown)


spulec:
Thanks!

PR #3086, opened 2020-06-22
Support for autoscaling policies in run_jobflow, add_instance_group a…

### LINKED ISSUE #3085: Support Autoscaling in EMR
Accept an AutoScalingPolicy in run_job_flow and add_instance_group.  Return AutoScalingPolicy in list_instance_groups.  Support put_auto_scaling_policy and remove_auto_scaling_policy operations.

jweite:
Fixes #3085 

Support for autoscaling policies in run_jobflow, add_instance_group and list_instance_groups.
Support for cluster_id parameter substitution in autoscaling policy cloudwatch alarm dimensions.
New methods put_autoscaling_policy and remove_autoscaling_policy support.

I faced several obstacles to this otherwise straightforward extension, and want to call them out here, as those in-the-know may have recommendations for better ways (i.e. without the utility functions I added...), so I anticipate discussion about these.

The first was in passing the "deep" AutoScalingPolicy in add_instance_group, as existing code made use of the flattened querystring parameters, retreived via _get_list_prefix().  _get_list_prefix only seems to be able to deal with param structures that are themselves flat beyond the prefix.  Prior efforts to support EbsConfiguration params (also deep) were handled by a purpose-built function that unflattened them.  I elected instead to create a generic Unflattener utility.  I'm not in love with the approach, but didn't feel up for writing a _get_list_prefix variant that could deal with  params of depth in the list.  See moto/emr/responses.py and utils.py.

For put_auto_scaling_policy I went with _get_parm() to fetch the policy param.  It preserves the params depth.  But doesn't convert key format from Camel to underscore.  I added a generic utility for doing this to dict/list structures, also in moto/emr/utils.py.  I perform this transformation in the model to all AutoScalingPolicies.

### DIFF
--- a/moto/emr/models.py
+++ b/moto/emr/models.py
@@ -7,7 +7,12 @@ from boto3 import Session
 from dateutil.parser import parse as dtparse
 from moto.core import BaseBackend, BaseModel
 from moto.emr.exceptions import EmrError
-from .utils import random_instance_group_id, random_cluster_id, random_step_id
+from .utils import (
+    random_instance_group_id,
+    random_cluster_id,
+    random_step_id,
+    CamelToUnderscoresWalker,
+)
 
 
 class FakeApplication(BaseModel):
@@ -28,6 +33,7 @@ class FakeBootstrapAction(BaseModel):
 class FakeInstanceGroup(BaseModel):
     def __init__(
         self,
+        cluster_id,
         instance_count,
         instance_role,
         instance_type,
@@ -36,8 +42,10 @@ class FakeInstanceGroup(BaseModel):
         id=None,
         bid_price=None,
         ebs_configuration=None,
+        auto_scaling_policy=None,
     ):
         self.id = id or random_instance_group_id()
+        self.cluster_id = cluster_id
 
         self.bid_price = bid_price
         self.market = market
@@ -53,7 +61,7 @@ class FakeInstanceGroup(BaseModel):
         self.role = instance_role
         self.type = instance_type
         self.ebs_configuration = ebs_configuration
-
+        self.auto_scaling_policy = auto_scaling_policy
         self.creation_datetime = datetime.now(pytz.utc)
         self.start_datetime = datetime.now(pytz.utc)
         self.ready_datetime = datetime.now(pytz.utc)
@@ -62,6 +70,34 @@ class FakeInstanceGroup(BaseModel):
 
     def set_instance_count(self, instance_count):
         self.num_instances = instance_count
+
+    @property
+    def auto_scaling_policy(self):
+        return self._auto_scaling_policy
+
+    @auto_scaling_policy.setter
+    def auto_scaling_policy(self, value):
+        if value is None:
+            self._auto_scaling_policy = value
+            return
+        self._auto_scaling_policy = CamelToUnderscoresWalker.parse(value)
+        self._auto_scaling_policy["status"] = {"state": "ATTACHED"}
+        # Transform common ${emr.clusterId} placeholder in any dimensions it occurs in.
+        if "rules" in self._auto_scaling_policy:
+            for rule in self._auto_scaling_policy["rules"]:
+                if (
+                    "trigger" in rule
+                    and "cloud_watch_alarm_definition" in rule["trigger"]
+                    and "dimensions" in rule["trigger"]["cloud_watch_alarm_definition"]
+                ):
+                    for dimension in rule["trigger"]["cloud_watch_alarm_definition"][
+                        "dimensions"
+                    ]:
+                        if (
+                            "value" in dimension
+                            and dimension["value"] == "${emr.clusterId}"
+                        ):
+                            dimension["value"] = self.cluster_id
 
 
 class FakeStep(BaseModel):
@@ -319,7 +355,7 @@ class ElasticMapReduceBackend(BaseBackend):
         cluster = self.clusters[cluster_id]
         result_groups = []
         for instance_group in instance_groups:
-            group = FakeInstanceGroup(**instance_group)
+            group = FakeInstanceGroup(cluster_id=cluster_id, **instance_group)
             self.instance_groups[group.id] = group
             cluster.add_instance_group(group)
             result_groups.append(group)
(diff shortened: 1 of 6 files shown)


moto/emr/models.py
```diff
@@ -63,6 +71,34 @@ def __init__(
...
+    @property
+    def auto_scaling_policy(self):
+        return self._auto_scaling_policy
+
+    @auto_scaling_policy.setter
+    def auto_scaling_policy(self, value):
+        if value is None:
+            self._auto_scaling_policy = value
+            return
+        self._auto_scaling_policy = CamelToUnderscoresWalker.parse(value)
```
jweite, line 83: Here is the generic Camel to underscore conversion of the deep AutoScalingPolicy parameter.

moto/emr/responses.py
```diff
@@ -329,6 +331,8 @@ def run_job_flow(self):
                 ig["instance_count"] = int(ig["instance_count"])
                 # Adding support to EbsConfiguration
                 self._parse_ebs_configuration(ig)
+                # Adding support for auto_scaling_policy
```
jweite, line 334: A generic "unflattening" of flattened querystring params, a la the immediately preceding EbsConfig special case, but using a generic mechanism.

moto/emr/responses.py
```diff
@@ -76,6 +76,8 @@ def add_instance_groups(self):
             item["instance_count"] = int(item["instance_count"])
             # Adding support to EbsConfiguration
             self._parse_ebs_configuration(item)
+            # Adding support for auto_scaling_policy
+            Unflattener.unflatten_complex_params(item, "auto_scaling_policy")
```
jweite, line 80: A generic "unflattening" of flattened querystring params, a la the immediately preceding EbsConfig special case, but using a generic mechanism.

moto/emr/responses.py
```diff
@@ -989,3 +1113,120 @@ def terminate_job_flows(self):
     <RequestId>2690d7eb-ed86-11dd-9877-6fad448a8419</RequestId>
   </ResponseMetadata>
 </TerminateJobFlowsResponse>"""
+
+PUT_AUTO_SCALING_POLICY = """<PutAutoScalingPolicyResponse xmlns="http://elasticmapreduce.amazonaws.com/doc/2009-03-31">
+  <PutAutoScalingPolicyResult>
+    <ClusterId>{{cluster_id}}</ClusterId>
+    <InstanceGroupId>{{instance_group.id}}</InstanceGroupId>
+    {% if instance_group.auto_scaling_policy is not none %}
```
jweite, line 1121: This block is a duplicate of one in ADD_INSTANCE_GROUP_RESPONSE.  Are there any patterns established in moto for re-use of common templated blocks?

bblommers: As far as I'm aware, there's no existing pattern. But with a block this size, it would be good to only declare it once.

The best approach in terms of extendability is, IMO, to change the `response_template()` signature in moto/core/responses.py, and allow for inline templates. Something like this:  
```
def response_template(self, source, inline_templates = {})
  # inline_templates = {"AUTOSCALING_POLICY": "AUTOSCALING_POLICY_TEMPLATE"}
  ..
  template = self.environment.get_template(template_id)
  for k, v in inline_templates:
    template.substibute(k, self.environment.get_template(v))
  return template
```

moto/emr/responses.py
bblommers: Can you leave a note/todo above the `_parse_ebs_configuration method`, to indicate it can/should be rewritten using the same Unflattener?
I'll create an issue later on to record this rewrite as technical debt

tests/test_emr/test_emr_boto3.py
```diff
@@ -476,6 +476,118 @@ def test_run_job_flow_with_instance_groups():
...
+    resp = client.put_auto_scaling_policy(
+        ClusterId=cluster_id,
+        InstanceGroupId=core_instance_group["Id"],
+        AutoScalingPolicy=auto_scaling_policy,
+    )
+
+    auto_scaling_policy_with_cluster_id = _patch_cluster_id_placeholder_in_autoscaling_policy(
+        auto_scaling_policy, cluster_id
+    )
+    del resp["AutoScalingPolicy"]["Status"]
```
bblommers, line 557: Let's assert the `Status` arg has the right value, before discarding it

moto/emr/utils.py
```diff
@@ -37,3 +38,109 @@ def steps_from_query_string(querystring_dict):
...
+    @staticmethod
+    def unflatten_complex_params(input_dict, param_name):
+        """ Function to unflatten (portions of) dicts with complex keys.  The moto request parser flattens the incoming
+        request bodies, which is generally helpful, but for nested dicts/lists can result in a hard-to-manage
+        parameter exposion.  This function allows one to selectively unflatten a set of dict keys, replacing them
+        with a deep dist/list structure named identically to the root component in the complex name.
+
+        Complex keys are composed of multiple components
+        separated by periods. Components may be prefixed with _, which is stripped.  Lists indexes are represented
+        with two components, 'member' and the index number. """
```
bblommers, line 53: Proper documentation is always good to see :+1:

moto/emr/responses.py
jweite: Do you object to me just making the change as part of this PR?  It's (almost) as simple as replacing the call to _parse_ebs_configuraiton with unflatten_complex_params(item, "ebs_configuration").

Almost, because the unflattener parses every attribute as a string, while _parse_ebs_configuration knows which are ints, and the template as written needs one of them to be.  Addition of the jinja2 int filter fixes that, though.

bblommers: That would be even better!

moto/emr/responses.py
jweite: I'm going with something similar, but which enables the use of jinja2's {% include 'INLINE_TEMPLATE_NAME' %}, by dropping the inline templates into the loader dict.  To make inline templates resolveable by their name, they get stored in the loader dict keyed with their name, which is different than for top-level templates (keyed by object id), but not harmful in my opinion.  Let me know if you disagree.

moto/emr/utils.py
jweite: Tx!

tests/test_emr/test_emr_boto3.py
jweite: Done.

moto/emr/responses.py
jweite: Done.

moto/emr/responses.py
jweite: Done.

jweite:
Bert, I'm getting an _ERROR: Package 'rsa' requires a different Python: 2.7.16 not in '>=3.5, <4'_ complaint when I try to _make init_ in a python2.7 virtualenv on mac.  Any guidance on setting up for 2.7?

jweite:
Looks like server-mode travis tests have gone wonky.  I see the same failures happening in some other recent PR's commits.

bblommers:
@jrbeilke Not sure why the RSA dependency isn't failing for Travis - it realistically should without compatibility.

On the ServerMode tests - correct, our master branch is currently failing as well. Will hold off on merging until that's fixed. 

I'm happy with the PR as is though - thanks again for your work!

jweite:
Thanks Bert, I appreciate the review and look forward to the merge.

bblommers:
Hi @jweite, would you mind merging master into your branch? The ServerMode tests are now fixed, so Travis should be happy again if we pull the fix here as well

PR #3088, opened 2020-06-25
Fix: Running commands inside an ec2 are breaking due to region.

usmangani1:
Fix: Running commands inside an ec2 are breaking due to region.

https://github.com/localstack/localstack/issues/2618

### DIFF
--- a/moto/ssm/models.py
+++ b/moto/ssm/models.py
@@ -278,7 +278,8 @@ class SimpleSystemManagerBackend(BaseBackend):
         self._resource_tags = defaultdict(lambda: defaultdict(dict))
         self._commands = []
         self._errors = []
-
+        # assigning the default region as us-east-1
+        self._region = "us-east-1"
         # figure out what region we're in
         for region, backend in ssm_backends.items():
             if backend == self:
--- a/tests/test_ssm/test_ssm_boto3.py
+++ b/tests/test_ssm/test_ssm_boto3.py
@@ -12,7 +12,7 @@ import json
 from botocore.exceptions import ClientError, ParamValidationError
 from nose.tools import assert_raises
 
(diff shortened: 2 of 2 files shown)


bblommers:
Hi @usmangani1, there is no default region for SSM. If I don't have a region configured, running the simplest command:   
`aws ssm send-command --document-name example`  
against AWS gives me an explicit error:   
`You must specify a region. You can also configure your region by running "aws configure".`

Not sure what the fix is for the original issue - does the user have a default region?

eerajoshi:
Getting 500 error while executing below steps
1. aws --endpoint-url=http://localhost:4566 iam create-instance-profile --instance-profile-name DEMOPROFILE
2. aws --endpoint-url=http://localhost:4566 iam add-role-to-instance-profile --instance-profile-name DEMOPROFILE --role-name demoRole
3. aws --endpoint-url=http://localhost:4566 ec2 associate-iam-instance-profile --iam-instance-profile Name=DEMOPROFILE --instance-id i-5cc861a9086ef7807
ERROR – 
b'<!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 3.2 Final//EN">\n<title>500 Internal Server Error</title>\n<h1>Internal Server Error</h1>\n<p>The server encountered an internal error and was unable to complete your request. Either the server is overloaded or there is an error in the application.</p>\n'

Please suggest the solution

eerajoshi:
Is this fixed?

bblommers:
Hi @eerajoshi, this looks to be unrelated to the SSM issue - can you open a new issue for this?
Executing the second statement already gives me an error that the role-name `demoRole` does not exist, so I guess we're still missing one step to reproduce this

PR #3089, opened 2020-06-25
Add support for template urls in cfn validation

shano:
Added as boto supports both TemplateBody and TemplateUrl

* Adds TemplateURL as a validate_template option
* Adds a test to validate this


shano:
No worries, thanks for merging @bblommers!

bblommers:
This is now available as of moto >= 1.3.15.dev916

Issue #3090, opened 2020-06-26
elbv2 create_rule fails with (ValidationError) when calling the CreateRule operation: A condition value must be specified


adriank-convoy:
## Summary

Call to elbv2 `create_rule()` method fails when specifying a Condition with a PathPatterConfig:
```
"PathPatternConfig": {"Values": ["/test_repo/test_branch/*"]}
```

## Error:
```
Traceback (most recent call last):
  File "convoyml/test/bug_example.py", line 68, in <module>
    bug_repro()
  File "/Users/akreuziger/virtualenvs/py3/lib/python3.7/site-packages/moto/core/models.py", line 88, in wrapper
    result = func(*args, **kwargs)
  File "/Users/akreuziger/virtualenvs/py3/lib/python3.7/site-packages/moto/core/models.py", line 88, in wrapper
    result = func(*args, **kwargs)
  File "convoyml/test/bug_example.py", line 54, in bug_repro
    "ContentType": "application/json",
  File "/Users/akreuziger/virtualenvs/py3/lib/python3.7/site-packages/botocore/client.py", line 276, in _api_call
    return self._make_api_call(operation_name, kwargs)
  File "/Users/akreuziger/virtualenvs/py3/lib/python3.7/site-packages/botocore/client.py", line 586, in _make_api_call
    raise error_class(parsed_response, operation_name)
botocore.exceptions.ClientError: An error occurred (ValidationError) when calling the CreateRule operation: A condition value must be specified
```

## Script to reproduce
```python
import os
from moto import (
    mock_ec2,
    mock_elbv2,
)
import boto3


@mock_ec2
@mock_elbv2
def bug_repro():
    # VPC
    vpc_id = boto3.client("ec2").create_vpc(CidrBlock="10.10.0.0/16")["Vpc"]["VpcId"]
    # Subnet for the load balancer
    subnet_id = boto3.client("ec2").create_subnet(
        CidrBlock="10.10.0.0/24", VpcId=vpc_id
    )["Subnet"]["SubnetId"]
    # Load balancer
    load_balancer_arn = boto3.client(service_name="elbv2").create_load_balancer(
        Name="my-alb", Subnets=[subnet_id]
    )["LoadBalancers"][0]["LoadBalancerArn"]
    # The port 80 HTTP listener
    listener_arn = boto3.client(service_name="elbv2").create_listener(
        LoadBalancerArn=load_balancer_arn,
        Protocol="HTTP",
        Port=80,
        DefaultActions=[
[...]

PR #3091, opened 2020-06-26
Fix 'Values' casing bug in elbv2 create_rule

adriank-convoy:
Fixes https://github.com/spulec/moto/issues/3090


adriank-convoy:
Bug was not related to Value/value case, going to open another PR with the actual fix

PR #3092, opened 2020-06-26
Fix condition filtering bug in elbv2.create_rule()

adriank-convoy:
Fixes https://github.com/spulec/moto/issues/3090, I wasn't able to get tests running locally, but I verified it fixes the repro script from the linked issue. The problem is that the values returned from `_condition.items()` are `('field', 'path-pattern'), ('path_pattern_config._values.member.1', '/test_repo/test_branch/*')`, so the key doesn't actually start with `"values.member"`

### DIFF
--- a/moto/elbv2/responses.py
+++ b/moto/elbv2/responses.py
@@ -356,7 +356,7 @@ class ELBV2Response(BaseResponse):
             condition = {}
             condition["field"] = _condition["field"]
             values = sorted(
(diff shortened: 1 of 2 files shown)


adriank-convoy:
Yeah, sorry was just being lazy... I got tests running locally, and modified the existing `test_handle_listener_rules` unit test to add in a condition using `PathPatternConfig` (which also caught the same bug when updating a rule).

bblommers:
This is now available as of moto >= 1.3.15.dev917

PR #3100, opened 2020-06-29
SSM: Use EC2 region

bblommers:
Passes in the region to SSMBackend, instead of cycling through the dict of known backends.
This aligns with how all other backends handle the region.

Fixes https://github.com/localstack/localstack/issues/2618

### DIFF
--- a/moto/ssm/models.py
+++ b/moto/ssm/models.py
@@ -1,11 +1,11 @@
 from __future__ import unicode_literals
 
 import re
+from boto3 import Session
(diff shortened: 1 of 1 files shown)


mikegrima:
Should this rely on the EC2 backend for determining region, or should this use the SSM backend for determining region? Each service should define their regional backends in botocore.

bblommers:
Fair point @mikegrima - I've updated the logic to retrieve SSM's own regions

PR #3102, opened 2020-06-30
SSM documents

Dreadwall:
Implemented some of the AWS SSM document API calls (Get, Delete, Create, Update, Describe, Update Default Version, and List).


bblommers:
Hi @Dreadwall, thanks for those changes. Would you mind merging master into your branch? The ServerMode tests are now fixed, so Travis should be happy again if we pull the fix here as well

PR #3103, opened 2020-06-30
Cloudformation: Fix - validate template yml fixes

shano:
This change fixes:
* Replace call to non-existent exception yaml.ParserError
* Catches yaml scanner error for valid json with tabs
* Supply yaml loader to ensure yaml loading throws exception validly for json with tabs and doesn't try to load the json incorrectly

### DIFF
--- a/moto/cloudformation/responses.py
+++ b/moto/cloudformation/responses.py
@@ -365,8 +365,8 @@ class CloudFormationResponse(BaseResponse):
         except (ValueError, KeyError):
             pass
         try:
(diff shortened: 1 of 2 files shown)


shano:
Thanks @bblommers !

bblommers:
Hi @shano, would you mind merging master into your branch? The ServerMode tests are now fixed, so Travis should be happy again if we pull the fix here as well

shano:
Cool, done @bblommers

PR #3070, opened 2020-06-13
Sagemaker support

porthunt:
Solves flake8 errors from #2125.

Resolves #2124

PR #3105, opened 2020-07-01
Sagemaker models

jweite:
This PR contains my initial work on SageMaker Notebook operations (create, describe, list, start, stop) merged with the Model operation code from porthunt's PR 3070, as agreed to by porthunt (https://github.com/spulec/moto/pull/3070#issuecomment-652014582), with a few adjustments (return model arn, raise ValidationError when model not found).


jweite:
Anyone have input on the failure of test_batch.test_terminate_job (python 3.7 non-server-mode)?  I've seen that sporadically in my dev env, and presumed it was environmental (i.e. because I'm developing on mac).

bblommers:
Hi @jweite, would you mind merging master in this branch as well? Will review this PR afterwards.

I've seen `test_terminate_job` fail on Travis occasionally as well. As far as I know it's a timing issue, not environmental. Haven't had a chance to look into why it fails, though

jweite:
Fixed the issue with the server-mode tests.  Rebased to July 7 master.

I have coded a number of additional SageMaker sub-services in another branch of my fork for use on my project.  I'll contribute them too, but want to complete review/merge of this initial contribution first, so I can adjust similar issues in these other commits before PR-ing.

moto/sagemaker/__init__.py
```diff
@@ -0,0 +1,6 @@
+from __future__ import unicode_literals
+from .models import sagemaker_backends
+
+sagemaker_backend = sagemaker_backends["us-east-1"]
+mock_sagemaker = sagemaker_backend.decorator
+mock_sagemaker_deprecated = sagemaker_backend.deprecated_decorator
```
bblommers, line 6: Let's remove the `deprecated` decorator - we're moving away from those anyway

moto/sagemaker/models.py
```diff
@@ -0,0 +1,364 @@
...
+    ):
+        self.ModelName = model_name
+        self.creation_time = datetime.now().strftime("%Y-%m-%d %H:%M:%S")
+        self.Containers = containers
+        self.Tags = tags
+        self.EnableNetworkIsolation = False
+        self.VpcConfig = vpc_config
+        self.PrimaryContainer = primary_container
+        self.ExecutionRoleArn = execution_role_arn or "arn:test"
+        self.ModelArn = self.arn_for_model_name(self.ModelName, region_name)
```
bblommers, line 32: Variables should be in `snake_case_format`, i.e. no upper cases and underscores in between words

(Edit: Goes for all models, VpcConfig and Container as well)

moto/sagemaker/models.py
```diff
@@ -0,0 +1,364 @@
...
+        return {
+            k: v for k, v in response_object.items() if v is not None and v != [None]
+        }
+
+    @property
+    def response_create(self):
+        return {"ModelArn": self.ModelArn}
+
+    @staticmethod
+    def arn_for_model_name(model_name, region_name):
```
bblommers, line 46: `model_name` and `region_name` should both be part of the Model-instance, so this can be a non-static method without any args. (Same as `FakeSagemakerNotebookInstance().arn`)

moto/sagemaker/models.py
```diff
@@ -0,0 +1,364 @@
...
+
+    def reset(self):
+        region_name = self.region_name
+        self.__dict__ = {}
+        self.__init__(region_name)
+
+    def create_model(self, **kwargs):
+        model_obj = Model(
+            self.region_name,
+            kwargs.get("ModelName"),
```
bblommers, line 229: Use named variables for long argument lists like this - reduces the chance that we're passing in arguments in the wrong order

`Model(region_name=self.region_name, model_name=kwargs.get("ModelName"), ...)`

moto/sagemaker/models.py
```diff
@@ -0,0 +1,364 @@
...
+            Model.arn_for_model_name(model_name, self.region_name)
+        )
+        raise RESTError(
+            error_type="ValidationException", message=message, template="error_json",
+        )
+
+    def list_models(self):
+        models = []
+        for model in self._models.values():
+            model_response = model.response_object.copy()
```
bblommers, line 256: Let's use the  buildin deepcopy here (`from copy import deepcopy`)

moto/sagemaker/models.py
```diff
@@ -0,0 +1,364 @@
...
+            model_response.update(model.response_create)
```
bblommers, line 257: `response_create` adds the model_arn - isn't that already part of the model-class in the first place? 
In other words - why is this line necessary?

moto/sagemaker/models.py
```diff
@@ -0,0 +1,364 @@
...
+            role_arn,
+            subnet_id=subnet_id,
+            security_group_ids=security_group_ids,
+            kms_key_id=kms_key_id,
+            tags=tags,
+            lifecycle_config_name=lifecycle_config_name,
+            direct_internet_access=direct_internet_access
+            if direct_internet_access is not None
+            else "Enabled",
+            volume_size_in_gb=volume_size_in_gb if volume_size_in_gb is not None else 5,
```
bblommers, line 301: If I'm not mistaken, we can drop the if-else statements here.

To test this: extend `test_create_notebook_instance_minimal_params` to pass in `direct_internet_access=None, volume_size_in_gb=None`. The boto3-client should simply drop the arguments from the incoming request, and subsequently use the default function values from this method

moto/sagemaker/responses.py
```diff
@@ -0,0 +1,136 @@
...
+    def start_notebook_instance(self):
+        notebook_instance_name = self._get_param("NotebookInstanceName")
+        notebook_instance = self.sagemaker_backend.get_notebook_instance(
+            notebook_instance_name
+        )
+        notebook_instance.start()
+        return 200, {}, json.dumps("{}")
+
+    @amzn_request_id
+    def stop_notebook_instance(self):
```
bblommers, line 113: Can we extract the logic from these 3 methods (describe, start, stop) into models.py?
The response-class shouldn't know anything about the inner workings of the NotebookInstance-class

moto/sagemaker/responses.py
```diff
@@ -0,0 +1,136 @@
...
+        return 200, {}, json.dumps("{}")
+
+    @amzn_request_id
+    def delete_notebook_instance(self):
+        notebook_instance_name = self._get_param("NotebookInstanceName")
+        self.sagemaker_backend.delete_notebook_instance(notebook_instance_name)
+        return 200, {}, json.dumps("{}")
+
+    @amzn_request_id
+    def list_tags(self):
```
bblommers, line 128: Same here actually, this can be extracted as well - the Responses-class shouldn't know that how tags are stored

moto/server.py
```diff
@@ -102,6 +102,10 @@ def infer_service_region_host(self, environ):
                 # If Newer API version, use dynamodb2
                 if dynamo_api_version > "20111205":
                     host = "dynamodb2"
+        elif service == "sagemaker":
+            host = "api.sagemaker.{region}.amazonaws.com".format(
```
bblommers, line 106: Nice find - that must have been a fun bug to track down

tests/test_sagemaker/test_sagemaker_notebooks.py
```diff
@@ -0,0 +1,228 @@
...
+@mock_sagemaker
+def test_describe_nonexistent_model():
+    sagemaker = boto3.client("sagemaker", region_name=TEST_REGION_NAME)
+
+    with assert_raises(ClientError) as e:
+        resp = sagemaker.describe_model(ModelName="Nonexistent")
+    assert_true(
+        e.exception.response["Error"]["Message"].startswith("Could not find model")
+    )
+    pass
```
bblommers, line 228: Unnecessary

jweite:
Thank you for the review and comments Bert.  Will get on this shortly.

moto/sagemaker/models.py
jweite: The rationale behind this non-pythonic style is to match the case of the API response, such that a simple generic mechanism for generating the response can be utilized.  I believe @porthunt adapted this technique from the ECR models.

While I'm certainly a believer in consistent coding style, I favor simpler, shorter code and fewer transformations over that.

If you still feel strongly about this, would you point me toward the moto module you feel implements simple response generation from snake_case model attributes?  Tx.

bblommers: ECR keeps snake_case, actually and only generates camelCase on request: https://github.com/spulec/moto/blob/master/moto/ecr/models.py#L27

bblommers: Feel free to just copy paste the relevant bits btw - it's used in ECS as well, so I'd be happy to refactor these methods into a utility method/class afterwards

jweite: Thanks for the pointer and clarification Bert, will do.

jweite: Bert, for your notes wrt ECR's BaseObject's camel-caser for future refactoring:
- It intentionally leaves the key's first char lowercase, as appropriate for ECR, but not for Sagemaker.  I had to modify it.
- It leaves keys without underscores unmodified.  In Sagemaker I need them to be capitalized.

moto/sagemaker/models.py
jweite: Bert, the reason for the _staticmethod arn_for_model_name()_ is to be able to format an arn for a resource _that doesn't exist_, typically to be included in an error message, so we can pretty closely replicate the actual error messages returned by boto3 against live AWS.  You'll see this done in SageMakerModelBackend.describe_model().

Rather than duplicate that logic for formatting actual resource object's arns, I use this staticmethod also to initialize a resource object's actual arn attribute (or implement its property).

moto/sagemaker/models.py
jweite: I think it's still necessary, since argument direct_internet_access could be _None_ (if explicitly passed in as such), so this ternary would be needed to default the value passed to FakeSagemakerNotebookInstance to "Enabled" when that's the case.

tests/test_sagemaker/test_sagemaker_notebooks.py
jweite: Bert, the reason I started adding these "make sure the exception was thrown for the expected reason" checks was in case ClientError gets raised for other reasons than the one we expected.  I recently had a unit test like this one happily passing, despite there being a fault in the unit-under-test that resulted in ClientErrors being raised for a different reason than the one expected.

Let me know if you object to me leaving this here.

moto/sagemaker/__init__.py
jweite: Done

moto/sagemaker/models.py
jweite: Done.

moto/sagemaker/models.py
jweite: Done.

moto/sagemaker/models.py
jweite: Done.

moto/sagemaker/responses.py
jweite: Done.

moto/sagemaker/responses.py
jweite: Done.

moto/server.py
jweite: Just call it an educational opportunity.  ;)

tests/test_sagemaker/test_sagemaker_notebooks.py
bblommers: My bad, I should have been more explicit. The assert should definitely stay - the `pass` statement isn't necessary :)

moto/sagemaker/models.py
bblommers: Yup, you're right

moto/sagemaker/models.py
jweite: Agreed.  Done.

moto/sagemaker/models.py
```diff
@@ -0,0 +1,399 @@
...
+            containers=kwargs.get("Containers", []),
+            tags=kwargs.get("Tags", []),
+        )
+
+        self._models[kwargs.get("ModelName")] = model_obj
+        return model_obj.response_create
+
+    def describe_model(self, model_name=None):
+        for model in self._models.values():
+            # If a registry_id was supplied, ensure this repository matches
```
bblommers, line 263: Minor nitpick, but I only just noticed this now. The comment doesn't really apply, and I guess the logic can be simplified to:
```
if model_name == model_name:
  return response_object
```
Everything else looks good - happy to merge after this

jweite: TBH, I'm not sure why this was coded as a linear search.  The _models bucket is keyed with the model_name.  I re-wrote that using dict get().  Done.

jweite:
Scratching my head on these build fails.  They center around batch.test_submit_job, and are inconsistent.  One time it passed 3.7 local mode and failed server mode, once the opposite.  But always OK in 2.7 and 3.8 (so far).  Can't reproduce locally (though I do have intermittent problems with local batch.test_terminate_job).   Feels like a race condition or thread crash, but not sure why it's only started showing up with this latest change.

bblommers:
@jweite That's a known bug, actually - wasn't introduced by your changes. I've been meaning to look into it, but haven't had time yet. So don't worry about that failure

jweite:
Thanks @bblommers, appreciate the update.

I created a looping version of test_submit_job, and have reproduced this failure at least once locally.  I also can reproduce the fails of test_terminate_job pretty consistently the same way.  I haven't got a handle on the test_submit_job problem, but think I see the race between batch/models.py:Job.run() and the test case to get to [here](https://github.com/spulec/moto/blob/3e2a5e7ee80e32feab1bdce30c7fd9bf9bd2569a/moto/batch/models.py#L499).  

I don't think this gets to the heart of the matters, increasing the time for the faux-job to progress before terminating it to 3s ([here](https://github.com/spulec/moto/blob/3e2a5e7ee80e32feab1bdce30c7fd9bf9bd2569a/tests/test_batch/test_batch.py#L871)) eliminated the failure for me in multiple (100+) runs.  (At its original 2s it fails for me within the first 10 runs).

jweite:
Here's a theory on submit_job fail (in which the logs don't contain the expected 'hello' message: the "dodgy hack" in Jobs.run() checks for stop ever 0.15s, but only grabs the logs every 10x around (1.5s) to avoid some limit on grabbing docker logs.  I'm thinking that that means there's a chance that logging that occurred between the last 1.5s interval and the stop will be lost.  It may need a final pull of the logs when actually stopped (though may have to wait up to 1.5s after the stop before it's allowed to do this, which obviates the value of the accelerated stop checking.)

bblommers:
Thanks for the investigation @jweite - I've increased the timeouts a bit as a temp fix. Hopefully that improves things

PR #3106, opened 2020-07-01
Fix: Ec2 - add destinationIpv6CIDR support.

usmangani1:
Fix: Ec2 - add destinationIpv6CIDR support.

https://github.com/localstack/localstack/issues/2651


bblommers:
Hi @usmangani1, would you mind merging master into your branch? The ServerMode tests are now fixed, so Travis should be happy again if we pull the fix here as well

PR #3107, opened 2020-07-03
Fix: Ec2 - add VPCPeeringConnections region response support.

ninhkd:
Fix: Ec2 - add VPCPeeringConnections region response support.


mikegrima:
It looks like you just need to run `black moto/ tests/`, and commit the changes.

PR #3116, opened 2020-07-10
Config - implement Organization Conformance Pack functionality

gruebel:
I added all endpoints with some sanity checks, but there are still parts missing. The code doesn't pull the conformance pack from S3 and all the Config Rules inside that template would result in a creation of those Config Rules in all accounts. Due to the lack of an implementation of Config Rules at the moment, this should/could be done an other time.

### DIFF
--- a/moto/config/models.py
+++ b/moto/config/models.py
@@ -1167,7 +1167,11 @@ class ConfigBackend(BaseBackend):
         pack = self.organization_conformance_packs.get(name)
 
         if pack:
+            pack.conformance_pack_input_parameters = input_parameters
             pack.delivery_s3_bucket = delivery_s3_bucket
+            pack.delivery_s3_key_prefix = delivery_s3_key_prefix
+            pack.excluded_accounts = excluded_accounts
+            pack.last_update_time = datetime2int(datetime.utcnow())
         else:
             pack = OrganizationConformancePack(
                 region,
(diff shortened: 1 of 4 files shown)


moto/config/models.py
```diff
@@ -1110,6 +1142,134 @@ def put_evaluations(self, evaluations=None, result_token=None, test_mode=False):
...
+
+        pack = self.organization_conformance_packs.get(name)
+
+        if pack:
+            pack.conformance_pack_input_parameters = input_parameters
+            pack.delivery_s3_bucket = delivery_s3_bucket
+            pack.delivery_s3_key_prefix = delivery_s3_key_prefix
+            pack.excluded_accounts = excluded_accounts
+            pack.last_update_time = datetime2int(datetime.utcnow())
+            pack._status = "UPDATE_SUCCESSFUL"
```
bblommers, line 1176: Can we haven an update()-method in the OrganisationConformancePack, to handle the value update?
The model itself should really be responsible for updating `last_update_time` and `status`

moto/config/responses.py
```diff
@@ -159,3 +159,46 @@ def put_evaluations(self):
             self._get_param("TestMode"),
         )
         return json.dumps(evaluations)
+
+    def put_organization_conformance_pack(self):
+        conformance_pack = self.config_backend.put_organization_conformance_pack(
+            self.region,
+            self._get_param("OrganizationConformancePackName"),
```
bblommers, line 166: Can we use named arguments here? Would reduce the chance of introducing an error due to the arguments being out of order

gruebel: done

moto/config/models.py
gruebel: done

gruebel:
@bblommers Not really sure, why there is an error in the tests for batch, maybe you could restart the build or you have some idea what's going on, because it only happened ont he Python 3.8 server mode build.

PR #3120, opened 2020-07-11
Fix :  cloudFormation dynamodb : delete resource on delete stack

usmangani1:
Fix :  cloudFormation dynamodb : delete resource on delete stack

https://github.com/localstack/localstack/issues/2699
https://github.com/localstack/localstack/issues/2697

### DIFF
--- a/moto/cloudformation/models.py
+++ b/moto/cloudformation/models.py
@@ -10,7 +10,7 @@ from moto.compat import OrderedDict
 from moto.core import BaseBackend, BaseModel
 from moto.core.utils import iso_8601_datetime_without_milliseconds
 
-from .parsing import ResourceMap, OutputMap, parse_and_delete_resource
+from .parsing import ResourceMap, OutputMap
 from .utils import (
     generate_changeset_id,
     generate_stack_id,
--- a/moto/cloudformation/models.py
+++ b/moto/cloudformation/models.py
@@ -712,11 +712,6 @@ class CloudFormationBackend(BaseBackend):
         if name_or_stack_id in self.stacks:
             # Delete by stack id
             stack = self.stacks.pop(name_or_stack_id, None)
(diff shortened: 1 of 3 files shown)


moto/dynamodb2/models/__init__.py
```diff
@@ -438,6 +438,18 @@ def create_from_cloudformation_json(
         )
         return table
 
+    @classmethod
+    def delete_from_cloudformation_json(
+            cls, resource_name, cloudformation_json, region_name
+    ):
+        print('resource name', resource_name)
```
bblommers, line 445: Remove the print statement please

moto/cloudformation/models.py
```diff
@@ -712,6 +712,11 @@ def delete_stack(self, name_or_stack_id):
         if name_or_stack_id in self.stacks:
             # Delete by stack id
             stack = self.stacks.pop(name_or_stack_id, None)
+            template = stack.template_dict
```
bblommers, line 715: Resources should already be deleted as part of the `stack.delete()` call: https://github.com/spulec/moto/blob/master/moto/cloudformation/models.py#L363 , so I don't think these lines are necessary

tests/test_cloudformation/test_cloudformation_stack_crud.py
```diff
@@ -188,6 +224,14 @@ def test_describe_stack_by_stack_id():
     stack_by_id.stack_name.should.equal("test_stack")
 
 
+@mock_cloudformation_deprecated
```
bblommers, line 227: I should have mentioned this before - please add a test for boto3 as well (`@mock_cloudformation`)

The test could do with a few asserts as well - please verify that the table exists first, but no longer exist after deletion

bblommers:
This is now part of moto >= 1.3.15.dev976

PR #3121, opened 2020-07-12
S3 - Persist metadata for Presigned URL

### LINKED ISSUE #2104: Metadata not available when using pre-signed urls

bblommers:
Fixes #2104 

Also adds validation for error scenarios: metadata set when creating the URL should also be send through when PUTing the URL

I ran the tests as they are against AWS for validation.

### DIFF
--- a/moto/s3/responses.py
+++ b/moto/s3/responses.py
@@ -1079,6 +1079,10 @@ class ResponseObject(_TemplateEnvironmentMixin, ActionAuthenticatorMixin):
             if key:
                 if not key.acl.public_read and not signed_url:
                     return 403, {}, ""
+            elif signed_url:
+                # coming in from requests.get(s3.generate_presigned_url())
+                if self._invalid_headers(request.url, dict(headers)):
+                    return 403, {}, S3_INVALID_PRESIGNED_PARAMETERS
 
         if hasattr(request, "body"):
             # Boto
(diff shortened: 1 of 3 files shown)


spulec:
Thanks!

PR #3122, opened 2020-07-13
Organizations - implement AWS Service Access functionality

gruebel:
I added a link, where it is possible to find the Service Principals, which support AWS Organizations, but I couldn't find any API endpoint to query it, that's why I added the list in the class.


bblommers:
This is now part of moto >= 1.3.15.dev963

PR #3097, opened 2020-06-27
Feature: DynamoDB: Support projectiontype KEYS_ONLY for GSI/LSI

bblommers:
Fixes #3055 

GSI's/LSI's can be configured to only store a subset of the data, to save space/money, by specifying a ProjectionType.
This PR adds support for the projection type KEYS_ONLY.

The ProjectionType 'ALL' is supported by default.
The ProjectionType 'INCLUDE' is not yet supported, but can be easily added later on.

PR #3125, opened 2020-07-14
Fix dynamodb2 KEYS_ONLY Indexes

mfogel:
It appears there's a bug in the implementation of #3097. When a dynamo index is marked `KEYS_ONLY` it includes not just the index hash & range key, but also the table hash & range key (if the range key exists).

The two tests added in #3097 both did not quite pass when ran directly against dynamo. This PR

- fixes those two tests so they pass when ran against dynamo
- adjusts the implementation of `LocalSecondaryIndex` and `GlobalSecondaryIndex` so that table keys can be included when the index is `KEYS_ONLY`

Note I can see a few different ways this could have been implemented - I chose the one that looked like the simplest from what I can see, but if you'd like a different implementation let me know I can make adjustments.

Cheers & as always thanks a ton for this project. Super helpful.

### DIFF
--- a/moto/dynamodb2/models/__init__.py
+++ b/moto/dynamodb2/models/__init__.py
@@ -283,17 +283,18 @@ class SecondaryIndex(BaseModel):
         if self.projection:
             if self.projection.get("ProjectionType", None) == "KEYS_ONLY":
                 allowed_attributes = ",".join(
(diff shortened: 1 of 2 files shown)


bblommers:
This is now part of moto >= 1.3.15.dev965

PR #3126, opened 2020-07-14
WIP - performance template caching

sf-mkirilov:


bblommers:
An alternative fix for this is now added to Moto in #5308, so I'll close this one.

Thanks for your work on this @sf-mkirilov!

PR #3142, opened 2020-07-16
Basic Support for Endpoints, EndpointConfigs and TrainingJobs

jweite:
Re issue 3141.

### DIFF
--- a/tests/test_sagemaker/test_sagemaker_endpoint.py
+++ b/tests/test_sagemaker/test_sagemaker_endpoint.py
@@ -206,7 +206,6 @@ def test_delete_endpoint():
     assert_true(
         e.exception.response["Error"]["Message"].startswith("Could not find endpoint")
     )
-    pass
 
 
 def _create_model(boto_client, model_name):
--- a/tests/test_sagemaker/test_sagemaker_endpoint.py
+++ b/tests/test_sagemaker/test_sagemaker_endpoint.py
@@ -1,0 +1,247 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+import datetime
+import boto3
+from botocore.exceptions import ClientError, ParamValidationError
+import sure  # noqa
+
+from moto import mock_sagemaker
+from moto.sts.models import ACCOUNT_ID
+from nose.tools import assert_true, assert_equal, assert_raises
+
+TEST_REGION_NAME = "us-east-1"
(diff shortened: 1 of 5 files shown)


tests/test_sagemaker/test_sagemaker_endpoint.py
```diff
@@ -0,0 +1,247 @@
...
+        e.exception.response["Error"]["Message"].startswith("Could not find endpoint")
+    )
+    pass
+
+
+def _create_model(boto_client, model_name):
+    resp = boto_client.create_model(
+        ModelName=model_name,
+        PrimaryContainer={
+            "Image": "382416733822.dkr.ecr.us-east-1.amazonaws.com/factorization-machines:1",
```
bblommers, line 216: You can use the mocked Account ID here, instead of your own (from moto.core import ACCOUNT_ID). 
Same ID is used in the other test file, test_sagemaker_training.py.

tests/test_sagemaker/test_sagemaker_endpoint.py
```diff
@@ -0,0 +1,247 @@
...
+    assert_true(
+        e.exception.response["Error"]["Message"].startswith("Could not find endpoint")
+    )
+
+    with assert_raises(ClientError) as e:
+        sagemaker.delete_endpoint(EndpointName=endpoint_name)
+    assert_true(
+        e.exception.response["Error"]["Message"].startswith("Could not find endpoint")
+    )
+    pass
```
bblommers, line 209: Don't need the pass-statements :)

tests/test_sagemaker/test_sagemaker_endpoint.py
jweite: Hey Bert, that's not a mock image URL, but the URL for one of the AWS pre-built SageMaker container images that one might use for training or an endpoint.  They're region-specific, and you usually you get them from the SageMaker library:

  from sagemaker.amazon.amazon_estimator import get_image_uri
  container = get_image_uri(region_name, 'linear-learner')

I wanted to test with something as close to reality as possible, but not add a dependency on the sagemaker library just for this.

tests/test_sagemaker/test_sagemaker_endpoint.py
jweite: Me and my extra debugging breakpoint lines!

jweite:
Thanks Bert, I appreciate your review and merge!

PR #3145, opened 2020-07-17
Fix: Put Events API

loganasherjones:
Hi All,

Thanks for the library. I wanted to use it to test an application I'm working on that is using the PutEvents API and noticed that there is a discrepancy for the events service between what Amazon says it returns and what moto returns. This change fixes that issue.

* [AWS API Documentation for Put Events](https://docs.aws.amazon.com/eventbridge/latest/APIReference/API_PutEvents.html)

The current response looks like this:

```
{
    "ResponseMetadata": {
        "HTTPStatusCode": 200,
        "HTTPHeaders": {"server": "amazon.com"}, 
        "RetryAttempts": 0
    }
}
```

This merge request changes to match the API documentation:

```
{
    "FailedEntryCount": 0,
    "Entries": [
        {"EventId": "uuid"}
    ]
}
```

Thanks again and let me know if you have any questions, comments or concerns!

### DIFF
--- a/moto/events/models.py
+++ b/moto/events/models.py
@@ -340,17 +340,13 @@ class EventsBackend(BaseBackend):
     def put_events(self, events):
         num_events = len(events)
 
-        response = {"FailedEntryCount": 0, "Entries": []}
-        for _ in events:
-            response["Entries"].append({"EventId": str(uuid4())})
-
         if num_events < 1:
             raise JsonRESTError("ValidationError", "Need at least 1 event")
         elif num_events > 10:
             raise JsonRESTError("ValidationError", "Can only submit 10 events at once")
 
         # We dont really need to store the events yet
-        return response
+        return [{"EventId": str(uuid4())} for _ in events]
 
     def remove_targets(self, name, ids):
         rule = self.rules.get(name)
--- a/moto/events/models.py
+++ b/moto/events/models.py
@@ -8,6 +8,8 @@ from moto.core import BaseBackend, BaseModel
 from moto.sts.models import ACCOUNT_ID
 from moto.utilities.tagging_service import TaggingService
 
+from uuid import uuid4
+
 
 class Rule(BaseModel):
     def _generate_arn(self, name):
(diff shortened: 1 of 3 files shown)


moto/events/models.py
```diff
@@ -338,13 +340,17 @@ def put_targets(self, name, targets):
     def put_events(self, events):
         num_events = len(events)
 
+        response = {"FailedEntryCount": 0, "Entries": []}
```
bblommers, line 343: The format of the response should still be the responsibilty of responses.py.
Maybe this method can simply return a tuple with the required information: `return (failed_entries, entries_list)`

loganasherjones: Hi @bblommers thanks for the review. 

I based this implementation on the way the kinesis `put_events` worked. I think the reason that this response is generated in the models is because the order of the events needs to be retained based on the way they came in. If we were to return a tuple, we'd need some way to re-combine the tuple into the original order. I can definitely do that, but I think it adds an unnecessary level of complexity in the implementation.

If you still feel strongly that the response shouldn't be generated here, just let me know and I'll come up with an implementation. I just wanted you to know the reasoning why I implemented it this way.

loganasherjones: Here is an alternate implementation, let me know if you like this better.

**responses.py**
```python
def put_events(self):
        events = self._get_param("Entries")

        entries = self.events_backend.put_events(events)

        failed_count = len([e for e in entries if "ErrorCode" in e])
        response = {
            "FailedEntryCount": failed_count,
            "Entries": entries,
        }

        return json.dumps(response)
```

**models.py**

```python
def put_events(self, events):
        num_events = len(events)

        if num_events < 1:
            raise JsonRESTError("ValidationError", "Need at least 1 event")
        elif num_events > 10:
            raise JsonRESTError("ValidationError", "Can only submit 10 events at once")

        # We dont really need to store the events yet
        return [{"EventId": uuid4()} for _ in events]
```

bblommers: Much better @loganasherjones - I'd be happy with that

loganasherjones: pushed!

bblommers:
This is now part of moto >= 1.3.15.dev974

Issue #2958, opened 2020-05-04
Ecs update-service does not work with ARN


rifqifatih:
Hi, 

I tried updating ecs by using serviceArn as parameter for service and resulting in `An error occurred (ServiceNotFoundException) when calling the UpdateService operation: The service arn:aws:ecs:us-east-1:012345678910:service/test_ecs_service does not exist`.

IIRC updating with arn should be possible in boto3 / awscli. Here's the test to reproduce:

```
@mock_ecs
def test_update_service_with_arn():
    client = boto3.client("ecs", region_name="us-east-1")
    _ = client.create_cluster(clusterName="test_ecs_cluster")
    _ = client.register_task_definition(
        family="test_ecs_task",
        containerDefinitions=[
            {
                "name": "hello_world",
                "image": "docker/hello-world:latest",
                "cpu": 1024,
                "memory": 400,
                "essential": True,
                "environment": [
                    {"name": "AWS_ACCESS_KEY_ID", "value": "SOME_ACCESS_KEY"}
                ],
                "logConfiguration": {"logDriver": "json-file"},
            }
        ],
    )
    response = client.create_service(
        cluster="test_ecs_cluster",
        serviceName="test_ecs_service",
        taskDefinition="test_ecs_task",
        desiredCount=2,
    )
    response["service"]["desiredCount"].should.equal(2)
    response["service"]["serviceArn"].should.equal(
        "arn:aws:ecs:us-east-1:012345678910:service/test_ecs_service"
    )
    response = client.update_service(
        cluster="test_ecs_cluster",
        service="arn:aws:ecs:us-east-1:012345678910:service/test_ecs_service",
        taskDefinition="test_ecs_task",
        desiredCount=0,
    )
    response["service"]["desiredCount"].should.equal(0)
    response["service"]["schedulingStrategy"].should.equal("REPLICA")
```

bblommers:
Can you verify that is the case against a real AWS instance?
All the [documentation](https://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_UpdateService.html) only talks about the `name of the service` - for parameters where it accepts the name _or ARN_ it's always explicitly mentioned. 

This could be just the docs being out of date, of course. :)

rifqifatih:
Yes, confirmed by running `aws ecs update-service --cluster $CLUSTER_NAME --service $SERVICE_ARN --desired-count 1`

[...]

PR #3147, opened 2020-07-17
ECS - UpdateService - Allow service ARN to be passed in

### LINKED ISSUE #3139: ecs.update_service does not handle service_arns, just services names.
## Actual:
When running:
```
ecs_client.update_service(
            cluster=cluster_arn,
            service=service_arn,
            desiredCount=0
)
```
I'm getting `ServiceNotFoundException` raised, even though I'm using service and cluster ARNs returned by a call to `describe_services`, after creating the services with moto.

## Expected:
Service/Cluster names and ARNs should both be supported.

## Walkaround:
I changed my application logic to make it testable with moto, and it worked, but I shouldn't have to do that, especially that both `boto3` and AWS API allow using Names/ARNs interchangeably.

bblommers:
Fixes #3139, #2958


spulec:
Thanks!

PR #3149, opened 2020-07-18
Fix : EC2 - Added ownerId filter for describe instances

usmangani1:
Fix : EC2 - Added ownerId filter for describe instances
https://github.com/localstack/localstack/issues/2646


bblommers:
This is now part of moto >= 1.3.15.dev975

PR #3153, opened 2020-07-19
Add basic implement for cognito-idp create_resource_server

cm-iwata:
Implements cognito-idp create_resource_server.

### DIFF
--- a/moto/cognitoidp/exceptions.py
+++ b/moto/cognitoidp/exceptions.py
@@ -2,6 +2,7 @@ from __future__ import unicode_literals
 
 import json
 from werkzeug.exceptions import BadRequest
+from moto.core.exceptions import JsonRESTError
 
 
 class ResourceNotFoundError(BadRequest):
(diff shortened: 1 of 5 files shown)


moto/cognitoidp/exceptions.py
```diff
@@ -42,3 +43,11 @@ def __init__(self, message):
         self.description = json.dumps(
             {"message": message, "__type": "NotAuthorizedException"}
         )
+
+
+class InvalidParameterException(JsonRESTError):
```
cm-iwata, line 48: Existing error class extends  "BadRequest".
But it will be return http status code for  error code.
Error code should represent type of exception class.
I think  we should not use "BadRequest" class

tests/test_cognitoidp/test_cognitoidp.py
```diff
@@ -1344,6 +1344,50 @@ def test_admin_update_user_attributes():
...
+    res["ResourceServer"]["Name"].should.equal(name)
+    res["ResourceServer"]["Scopes"].should.equal(scopes)
+
+    with assert_raises(ClientError) as ex:
+        client.create_resource_server(
+            UserPoolId=user_pool_id, Identifier=identifier,
+            Name=name, Scopes=scopes)
+
+    ex.exception.operation_name.should.equal("CreateResourceServer")
+    ex.exception.response["Error"]["Code"].should.equal("InvalidParameterException")
```
cm-iwata, line 1384: Error code should represent type of exception class.
Not http status code.

bblommers: Yup - AWS really dropped the ball on the naming scheme

PR #3158, opened 2020-07-20
RAM - implement CRUD endpoints

gruebel:
It is a basic implementation of the AWS Resource Access Manager plus an endpoint for enabling sharing with AWS Organizations.

A lot of functionality is missing, so feel free to open up some issues and when I have time, I will implement them :)


moto/ram/models.py
```diff
@@ -0,0 +1,257 @@
...
+
+    def update(self, **kwargs):
+        self.allow_external_principals = kwargs.get(
+            "allowExternalPrincipals", self.allow_external_principals
+        )
+        self.last_updated_time = datetime.utcnow()
+        self.name = kwargs.get("name", self.name)
+
+    def _create_arn(self):
+        id = "{0}-{1}-{2}-{3}-{4}".format(
```
bblommers, line 152: This looks like a UUID - does it make sense to that instead? (`from uuid import uuid4`?)

tests/test_ram/test_ram.py
```diff
@@ -0,0 +1,362 @@
...
+    # when
+    response = client.create_resource_share(name="test")
+
+    # then
+    resource = response["resourceShare"]
+    resource["allowExternalPrincipals"].should.be.ok
+    resource["creationTime"].should.be.a(datetime)
+    resource["lastUpdatedTime"].should.be.a(datetime)
+    resource["name"].should.equal("test")
+    resource["owningAccountId"].should.equal("123456789012")
```
bblommers, line 26: You can import the ACCOUNT_ID from `moto.core`, no need to hardcode it everytime

moto/ram/models.py
gruebel: Good eye! Somehow didn't realize it 😄

tests/test_ram/test_ram.py
gruebel: Good point, I got used to hardcode the results of my tests.

bblommers:
This is now part of moto >= 1.3.15.dev980

kmcquade:
@bblommers @gruebel - thought I'd point out - RAM support is not noted in the README

bblommers:
Thanks @kmcquade - I've added it now

PR #3162, opened 2020-07-21
Fix: Api-Gateway ApiKeyAlreadyExists headers change.

usmangani1:
Fix: Api-Gateway ApiKeyAlreadyExists headers change.
https://github.com/localstack/localstack/issues/2753


bblommers:
The ServerMode tests are failing here, because it will automatically return some default headers (content-type, content-length, etc)
https://travis-ci.org/github/spulec/moto/jobs/712837260

Considering the simplicity of the test, I'd be OK with it if this test was only run in non Server Mode

bblommers (approved):
LGTM. Thanks @usmangani1!

bblommers:
This is now part of moto >= 1.3.15.dev1036

PR #3164, opened 2020-07-21
When generating multi-part upload IDs remove `/` characters

dkleissa:
Currently when generating an S3 multi-part upload ID, occasionally an upload ID will contain a `/` character. This is not consistent with how the actual API creates IDs (it does contain special characters `.` and `-` characters, but not `/`).

This small PR simply filters out `/` characters when generating S3 multi-part upload IDs in a similar way as how the code was already filtering out `=` and `+` characters.

### DIFF
--- a/moto/s3/models.py
+++ b/moto/s3/models.py
@@ -280,7 +280,7 @@ class FakeMultipart(BaseModel):
         self.parts = {}
         self.partlist = []  # ordered list of part ID's
         rand_b64 = base64.b64encode(os.urandom(UPLOAD_ID_BYTES))
-        self.id = rand_b64.decode("utf-8").replace("=", "").replace("+", "")
+        self.id = rand_b64.decode("utf-8").replace("=", "").replace("+", "").replace("/", "")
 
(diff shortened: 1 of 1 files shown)


dkleissa:
The way multipart upload IDs are being generated in moto results in a string that is based on the Base64 character set. This includes `/`. When AWS actually generates IDs, they contain `a-zA-Z0-9-_.` from what I can tell. From empirical testing, I'm assuming they use characters that are safe to put in a URL.

I noticed this discrepancy in moto because of how we use the generated IDs. We have a service that manages large, versioned data for users that runs directly on top of S3. When using moto during testing, if an ID ended up having a slash in it, a routing error would result in our service because it would end up not matching any endpoints due to the extra slashes. So you'd get this confusing behavior where every time you ran the tests a different set of 0-N tests would fail.

While this is unique to us and not necessarily directly related to moto functioning (if you're not trying to put the IDs in a URL), anyone that is using these IDs in a similar way during testing would run into a similar problem.

Issue #3168, opened 2020-07-23
CloudFormation create_stack Fails with Long Stack Name and S3 Bucket in Template


jweite:
The following unit test fails:

`
@mock_cloudformation
def test_boto3_create_stack_s3_long_name():
    cf_conn = boto3.client("cloudformation", region_name="us-east-1")

    stack_name = "MyLongStackName01234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012"

    template = '{"Resources":{"HelloBucket":{"Type":"AWS::S3::Bucket"}}}'

    cf_conn.create_stack(StackName=stack_name, TemplateBody=template)

    cf_conn.get_template(StackName=stack_name)["TemplateBody"].should.equal(
        json.loads(template, object_pairs_hook=OrderedDict)
    )
`

Moto tries to use a bucket name that's longer than the 63 char max.  CloudFormation truncates the StackName portion of the bucket name such that it doesn't exceed the limit.

bblommers:
This was fixed as of moto >= 1.3.15.dev1010

PR #3169, opened 2020-07-23
Fixed Failures in CloudFormation Provisioning of S3 Buckets When Stack has Long Name...

jweite:
...which causes the default S3 bucket name's length to exceed its 63 char limit.
Issue #3168.

### DIFF
--- a/tests/test_cloudformation/test_cloudformation_stack_crud_boto3.py
+++ b/tests/test_cloudformation/test_cloudformation_stack_crud_boto3.py
@@ -647,6 +647,21 @@ def test_boto3_create_stack():
 
 
 @mock_cloudformation
+def test_boto3_create_stack_s3_long_name():
+    cf_conn = boto3.client("cloudformation", region_name="us-east-1")
+
+    stack_name = "MyLongStackName01234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012"
+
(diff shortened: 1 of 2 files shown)


tests/test_cloudformation/test_cloudformation_stack_crud_boto3.py
```diff
@@ -646,6 +646,21 @@ def test_boto3_create_stack():
...
+    stack_name = "MyLongStackName01234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012"
+
+    template = '{"Resources":{"HelloBucket":{"Type":"AWS::S3::Bucket"}}}'
+
+    cf_conn.create_stack(StackName=stack_name, TemplateBody=template)
+
+    cf_conn.get_template(StackName=stack_name)["TemplateBody"].should.equal(
+        json.loads(template, object_pairs_hook=OrderedDict)
+    )
+
```
bblommers, line 662: Can you add an assert on the length of the S3-bucket as well, and ensure it still starts with the stack_name?

jweite: Done.  Also noticed my default bucket names could be mixed case: fixed this too.  
Thanks Bert.

jweite:
Hi @bblommers, anything else to-do in this PR?

bblommers:
Nope, all good @jweite - thanks for the reminder

PR #3173, opened 2020-07-24
Support --filters option in secretsmanager:ListSecrets

chriskilding:
Support the `--filters` option in `secretsmanager:ListSecrets`.

AWS documentation for the option:

- [Blog post](https://aws.amazon.com/blogs/security/identify-arrange-manage-secrets-easily-using-enhanced-search-in-aws-secrets-manager/)
- [CLI](https://docs.aws.amazon.com/cli/latest/reference/secretsmanager/list-secrets.html)
- [API](https://docs.aws.amazon.com/secretsmanager/latest/apireference/API_ListSecrets.html)

## To do

Filter value matcher:

- [x] String match
- [x] Any word match

Filter keys:

- [x] No key
- [x] Invalid key
- [x] Duplicate key (e.g. two filters for the 'name' key)
- [x] name
- [x] description
- [x] tag-key
- [x] tag-value
- [x] all

Filter values:

- [x] No values
- [x] One value
- [x] Multiple values

Filters:

- [x] No filters
- [x] One filter
- [x] Multiple filters


chriskilding:
Description of the "all" filter:

> You can also perform a blanket search without specifying any search attributes. The search result will include all secrets that contain the specified search value in either the secret name, description, tag key, or tag value. For example, searching "secrets for" will return any secret with secret name, tag key, tag value, or description that includes both "secrets" and "for" in the same field.

chriskilding:
Ready for a review now. In particular I'd appreciate a second pair of eyes on the filtering logic, as if there are any mistakes, they're most likely to be there. (There isn't a formal spec for the AWS filter behaviour, so I've had to prod the implementation with test queries, and rely on the blog post for details.)

moto/secretsmanager/list_secrets/filters.py
```diff
@@ -0,0 +1,44 @@
...
+        for tag in secret["tags"]:
+            if _matcher(k, tag["Key"]):
+                return True
+    return False
+
+
+def tag_value(secret, tag_values):
+    for v in tag_values:
+        for tag in secret["tags"]:
+            if _matcher(v, tag["Value"]):
```
bblommers, line 33: Can a tag-value ever be None? If so, the list_secrets call will probably fail

chriskilding: Good catch. Have added a test and handling for it.

PR #3174, opened 2020-07-24
Discard master-slave language

bopopescu:
For diversity reasons, it would be nice to try to avoid 'master' and 'slave' terminology in this repository which can be associated to slavery. The master-slave terminology could be problematic for people in several countries which has the history of slavery like Romania, USA and many others. Thank you for considering the proposal. Let me know if any changes in the PR are needed, I would be happy to implement them.


bblommers:
Hi @bopopescu, thanks for raising this issue and the PR. I'm very much in favour of renaming this, and the main/subordinate terms are a good replacement IMO.  
  
The way moto is set up, we only intercept the actual HTTP request that the boto3-client makes. So when it comes to the incoming and outgoing data in these requests, we have to conform to the exact format that boto3 uses. Changes like `self._get_param("Instances.MainInstanceType")` and `<AdditionalMainSecurityGroups>` break the schema and the interaction with the boto3-client. 

The moment AWS decides to change these, we can follow suit - but until then we only have control over our own internal variable names.

cc @spulec

spulec:
I agree with @bblommers : I am in support of something like this, but don't see a reasonable way to do this as long as AWS is using these terms. It seems we are stuck mimicking what they do unless there is a clever workaround that I can't see.

bblommers:
Something that is under our control; the branch naming. Looks like Github is in favour of renaming branches from master to `main`, and they're working now to make this as effortless as possible 'later this year': https://github.com/github/renaming

bblommers:
FYI, @spulec: Renaming the default branch is now possible, per the link above.

PR #3175, opened 2020-07-24
Various changes to organizations endpoint

hacosta:


moto/organizations/responses.py
```diff
@@ -115,6 +120,11 @@ def list_policies(self):
             self.organizations_backend.list_policies(**self.request_params)
         )
 
+    def delete_policy(self):
+        return json.dumps(
```
bblommers, line 124: The backend-method doesn't seem to return anything. It probably reads better to rewrite this to:
```
self.organizations_backend.delete_policy(**self.request_params)
return json.dumps({})
```

tests/test_organizations/test_organizations_boto3.py
```diff
@@ -410,28 +410,72 @@ def test_describe_policy_exception():
 
 
 @mock_organizations
-def test_attach_policy():
```
bblommers, line 413: Don't think this test needs to change? The attach_policy functionality hasn't changed

tests/test_organizations/test_organizations_boto3.py
```diff
@@ -479,6 +523,57 @@ def test_attach_policy_exception():
...
+    with assert_raises(ClientError) as e:
+        response = client.update_policy(PolicyId=non_existent_policy_id)
+    ex = e.exception
+    ex.operation_name.should.equal("DeletePolicy")
+    ex.response["Error"]["Code"].should.equal("400")
+    ex.response["Error"]["Message"].should.contain("PolicyNotFoundException")
+
+
+@mock_organizations
+def test_update_policy_exception():
```
bblommers, line 565: Please use unique names for the tests - `test_update_nonexisting_policy_with_description`?

hacosta:
Thanks for the feedback and sorry about the noise. LMK if this works.

bblommers:
This is now part of moto >= 1.3.15.dev999

PR #3179, opened 2020-07-25
CF : Added support for get template summary

usmangani1:
CF : Added support for get template summary


bblommers:
This is now part of moto >= 1.3.15.dev1001

PR #3180, opened 2020-07-25
FIX : IAM - Added support for pathPrefix in list_users_function

usmangani1:
FIX : IAM - Added support for pathPrefix in list_users_function

https://github.com/localstack/localstack/issues/2767

### DIFF
--- a/moto/iam/models.py
+++ b/moto/iam/models.py
@@ -947,11 +947,7 @@ class AccountSummary(BaseModel):
 
 
 def filter_items_with_path_prefix(path_prefix, items):
-    filtered_items = []
-    for role in items:
-        if role.path.startswith(path_prefix):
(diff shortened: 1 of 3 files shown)


moto/iam/models.py
```diff
@@ -946,6 +946,14 @@ def _users(self):
         return len(self._iam_backend.users)
 
 
+def filter_items_with_path_prefix(path_prefix, items):
+    filtered_items = []
+    for role in items:
+        if role.path.startswith(path_prefix):
+            filtered_items.append(role)
+    return filtered_items
```
bblommers, line 954: I think this is simple enough to convert to a list-comprehension:
`return [role for role in items if role.path.startswith(path_prefix)]`

PR #3181, opened 2020-07-26
Fix : SQS - Added support for attribute labels for send_message function

ninhkd:
Issue fixed:
https://github.com/localstack/localstack/issues/2731


ninhkd:
Hi @bblommers, thanks for your comment. I have updated the exception logic and add negative integration tests. Please have a look. Thanks

bblommers:
This is now part of moto >= 1.3.15.dev1005

PR #3182, opened 2020-07-26
Bugfix: S3 time precision issue fixed

macnev2013:
As per amazon s3's documentation, they are not storing time with milliseconds precision.
While in moto it was storing it in milliseconds. Which was causing s3 sync command with --exact-timestamp to download the files everytime when syncing data from s3 to local.
This fix solved the bug.

### DIFF
--- a/moto/core/utils.py
+++ b/moto/core/utils.py
@@ -187,7 +187,7 @@ def iso_8601_datetime_with_milliseconds(datetime):
 
 
 def iso_8601_datetime_without_milliseconds(datetime):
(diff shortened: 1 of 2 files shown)


macnev2013:
> Thanks for the PR, @macnev2013!
> Looks like the tests are now failing with these changes: https://travis-ci.org/github/spulec/moto/builds/711887910

Thanks for a quick response.
I changed was making change in core util function which was used by other services. So it was breaking.
As a fix i created a separate function for S3 which is used only from S3 module.

Issue #3127, opened 2020-07-14
cloudformation/parsing.py has an incomplete `UNDOCUMENTED_NAME_TYPE_MAP`


richford:
## Issue
[`cloudformation/parsing.py`](https://github.com/spulec/moto/blob/master/moto/cloudformation/parsing.py#L102) wisely has an `UNDOCUMENTED_NAME_TYPE_MAP` because the name type references in the [Amazon documentation](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-name.html) are incomplete. However, there are additional resource types that should be added. For my purposes in particular, I would like to add
```python
{
    "AWS::Batch::ComputeEnvironment": "ComputeEnvironmentName",
    "AWS::Batch::JobDefinition": "JobDefinitionName",
    "AWS::Batch::JobQueue": "JobQueueName",
    "AWS::EC2::LaunchTemplate": "LaunchTemplateName",
}
```
and I will submit a PR to that effect.

But, combing through the [cloudformation documentation](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-template-resource-type-ref.html), I noticed that the following resource types could also be added:
```python
{
    "AWS::AccessAnalyzer::Analyzer": "AnalyzerName",
    "AWS::AmazonMQ::Broker": "BrokerName",
    "AWS::AmazonMQ::Configuration": "Name",
    "AWS::Amplify::App": "Name",
    "AWS::Amplify::Branch": "BranchName",
    "AWS::Amplify::Domain": "DomainName",
    "AWS::ApiGateway::Authorizer": "Name",
    "AWS::ApiGateway::DomainName": "DomainName",
    "AWS::ApiGateway::RequestValidator": "Name",
    "AWS::ApiGateway::RestApi": "Name",
    "AWS::ApiGateway::Stage": "StageName",
    "AWS::ApiGateway::UsagePlan": "UsagePlanName",
    "AWS::ApiGateway::VpcLink": "Name",
    "AWS::ApiGatewayV2::Api": "Name",
    "AWS::ApiGatewayV2::Authorizer": "Name",
    "AWS::ApiGatewayV2::DomainName": "DomainName",
    "AWS::ApiGatewayV2::Model": "Name",
    "AWS::ApiGatewayV2::Stage": "StageName",
    "AWS::ApiGatewayV2::VpcLink": "Name",
    "AWS::AppConfig::Application": "Name",
[...]

PR #3185, opened 2020-07-27
#3127 - Decentralize CF naming responsibilities

bblommers:
First pass at decentralizing the CF responsibility to individual models.
Each Model that wants to be recognized by CloudFormation's parser now needs to:
 - Extend the moto.core.CloudFormationModel class
 - Implement the abstract method `cloudformation_type`
 - Implement the abstract method `cloudformation_name_type`
 - Implement the `create/update/delete_from_cloudformation_json` methods (as before)

This removes the need to explicitly list all known/supported models in our CloudFormation-service.

TODO: 
 

 -  [ ] Change all other models to subclass CloudFormationModel (Only DynamoDB::Table is done atm)
 - [ ] Add the create_from_cloudformation_json-methods to `CloudFormationModel`, to ensure clarity on what to do when implementing a new model
 -  [ ] Instead of cycling through the MODEL_LIST at runtime, we can probably dynamically recreate the structures we have now: 
```
MODEL_MAP = {model.cloudformation_type(): model for model in CloudFormationModel.__subclasses__()
NAME_TYPE_MAP = {model.cloudformation_type(): model.cloudformation_name_type() for model in CloudFormationModel.__subclasses__()
```
That would eliminate the need for the two new for-loops

### DIFF
--- a/moto/cloudformation/parsing.py
+++ b/moto/cloudformation/parsing.py
@@ -13,7 +13,6 @@ from moto.cloudwatch import models as cloudwatch_models
 from moto.cognitoidentity import models as cognitoidentity_models
 from moto.compat import collections_abc
 from moto.datapipeline import models as datapipeline_models
-from moto.dynamodb2 import models as dynamodb2_models
 from moto.ec2 import models as ec2_models
 from moto.ecs import models as ecs_models
 from moto.elb import models as elb_models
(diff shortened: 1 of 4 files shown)


bblommers:
Closing - this has now been merged as part of #3201

PR #3190, opened 2020-07-27
Enforce parameter naming

JordanReiter:
Parameters are not allowed to start with `ssm` or `aws`. This PR adds error messages which
correspond exactly to the error messages returned by boto3.

### DIFF
--- a/moto/ssm/exceptions.py
+++ b/moto/ssm/exceptions.py
@@ -78,6 +78,13 @@ class InvalidDocumentOperation(JsonRESTError):
         )
 
 
+class AccessDeniedException(JsonRESTError):
+    code = 400
+
+    def __init__(self, message):
+        super(AccessDeniedException, self).__init__("AccessDeniedException", message)
+
+
 class InvalidDocumentContent(JsonRESTError):
     code = 400
 
--- a/moto/ssm/models.py
+++ b/moto/ssm/models.py
@@ -27,6 +27,7 @@ from .exceptions import (
     ParameterNotFound,
     DocumentAlreadyExists,
     InvalidDocumentOperation,
+    AccessDeniedException,
     InvalidDocument,
     InvalidDocumentContent,
     InvalidDocumentVersion,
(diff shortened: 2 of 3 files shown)


bblommers:
Hi @JordanReiter, looks like these changes cause the Python2 build to fail: https://travis-ci.org/github/spulec/moto/builds/712328265

JordanReiter:
> Hi @JordanReiter, looks like these changes cause the Python2 build to fail: https://travis-ci.org/github/spulec/moto/builds/712328265

Yes, there were two lines making use of unsupported f-strings. Fixed in ffdce63.

bblommers:
This is now part of moto >= 1.3.15.dev1006

snimark:
I know this is closed, but I'm curious why the code was structured to prevent adding a parameter like this for testing the original base image:

```
@mock_ssm
def test_find_aws_ami():
    ssm_client = boto3.client('ssm', region_name='us-east-1')
    amzn2_ami_id = 'ami-411ba5e2f00000000'

    # Create the dummy latest Amazon Linux 2 reference
    ssm_client.put_parameter(
            Name='/aws/service/ami-amazon-linux-latest/amzn2-ami-hvm-x86_64-gp2',
            Value=amzn2_ami_id,
            Type='String'
    )

    session = create_image.get_session()
    base_ami_id = create_image.find_aws_ami(session)

    assert base_ami_id == amzn2_ami_id, "Base AMI id is not correct"

    return base_ami_id
```

When probing for the parameter name, this is what is presented:
```
$ aws ssm get-parameters --names /aws/service/ami-amazon-linux-latest/amzn2-ami-hvm-x86_64-gp2 --region us-east-1
{
    "Parameters": [
        {
            "Name": "/aws/service/ami-amazon-linux-latest/amzn2-ami-hvm-x86_64-gp2",
            "Type": "String",
            "Value": "ami-0947d2ba12ee1ff75",
            "Version": 33,
            "LastModifiedDate": 1600909956.765,
            "ARN": "arn:aws:ssm:us-east-1::parameter/aws/service/ami-amazon-linux-latest/amzn2-ami-hvm-x86_64-gp2",
            "DataType": "text"
        }
    ],
    "InvalidParameters": []
}
```

This is in accordance with the guide found here: https://aws.amazon.com/blogs/compute/query-for-the-latest-amazon-linux-ami-ids-using-aws-systems-manager-parameter-store/.

Is there another way to mock up a test for this?

bblommers:
Hi @snimark, preventing users from accidentally setting an `/aws/..`-parameter follows the boto3 behaviour, so I think this was the correct approach.

Adding default parameters like the latest AMI sounds like a good enhancement to moto, but until that's implemented, you might have to monkey-patch the `get-parameters`-call.

snimark:
OK, thanks for the answer.

Issue #3178, opened 2020-07-25
Implement UserId parameter for ModifySnapshotAttribute function.


aiell0:
# Issue Description
The boto3 function `ec2modify_snapshot_attribute()` was not implemented completely and has the potential to cause loss of code coverage in test suites where it is used.

# Code
```
from moto import mock_ec2
import boto3

@mock_ec2
def snapshot_modify_attribute():
    ec2 = boto3.resource("ec2", region_name="us-east-1")
    snapshot = ec2.Snapshot("snap-b230e53a")
    snapshot.modify_attribute(
        Attribute="createVolumePermission",
        OperationType="add",
        UserIds=["123456789123"]
    )

snapshot_modify_attribute()
```

## Expected Result
```
{
    'ResponseMetadata': {
        '...': '...',
    },
}
```

## Actual Result
`moto.ec2.exceptions.MotoNotImplementedError: The UserId parameter for ModifySnapshotAttribute has not been implemented in Moto yet. Feel free to open an issue at https://github.com/spulec/moto/issues`

Don't mind if I do!

## System Information
* boto3-1.14.28
* botocore-1.17.28
* moto-1.3.14 installed using pipenv and using python mocks.

bblommers:
This is now part of moto >= 1.3.15.dev1016

PR #3192, opened 2020-07-28
Implement UserIds for Snapshot Attributes

aiell0:
## Changelog
* Modernize snapshot attribute testing.
* Implement UserId parameter for snapshot attributes.
* Add tests for UserId.
* Implements https://github.com/spulec/moto/issues/3178


tests/test_ec2/test_elastic_block_store.py
```diff
@@ -453,128 +453,175 @@ def test_snapshot_filters():
     )
 
 
-@mock_ec2_deprecated
```
bblommers, line 456: Please keep the existing test case in tact, to ensure we're still supporting the deprecated use case.

The snapshot_with_userid should be a separate test case

aiell0: I added the old test case back in, as well as rebased with master. Thanks!

PR #3197, opened 2020-07-28
Fix SQS tag list from CloudFormation resource creation

whummer:
Moving this PR here - raised originally by @hudovisk [here](https://github.com/localstack/moto/pull/8), but I believe this change should better be merged upstream. Thanks!

---
The method `create_from_cloudformation_json` of the Sqs resource does not handle the difference of format of the Tags field in the resource template and the format expected in Sqs resource class.

In cfn resource template Tags is specified as a list of dicts. But the Sqs resource expects that the tags field be a single dict.
This behaviour causes a crash when a queue is created with tags from `create_from_cloudformation_json` and later the list_queue_tags is called because it tries to call `items` from `queue.tags` but tags is actually a list of dicts.

### DIFF
--- a/moto/core/utils.py
+++ b/moto/core/utils.py
@@ -356,3 +356,14 @@ def tags_from_query_string(
             else:
                 response_values[tag_key] = None
     return response_values
+
+
+def tags_from_cloudformation_tags_list(tags_list):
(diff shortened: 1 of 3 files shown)


bblommers:
This is now part of moto >= 1.3.15.dev1009

Issue #3187, opened 2020-07-27
Implement S3 CloudFormation Update Support


jweite:
On CloudFormation update_stack operation with a stack defining an S3 bucket:

AttributeError: type object 'FakeBucket' has no attribute 'update_from_cloudformation_json'

jweite:
@bblommers, this one's got more worms in the can than I anticipated.  AFAICT moto CloudFormation support doesn't know about the three different update behaviors (https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/using-cfn-updating-stacks-update-behaviors.html).  It seems to treat them all as _Replacements_: 

https://github.com/spulec/moto/blob/6adee0cbaf7a1971fff8809cc07dc4b23a08cc2e/moto/cloudformation/parsing.py#L340 always slams-in a new resource name regardless of whether the template is using any Replacement-update-type properties.

Handing this flexibly would be an interesting challenge.  Backend models would need an opportunity to weigh-in on the type of update a template's properties imply.  I'm thinking to declare a well-known method name for this that the cfn code can call on the model to ask it to evaluate the template properties.  If the method's not present (as it wouldn't be for most backend models) then moto assumes Replacement for backward compatibility.

1) Does this make sense to you?
2) Is there a better way?
3) How would others know to start providing this model method?

jweite:
Alternatively, the model class' update_from_cloudformation_json could make the decision, but would have to selectively ignore the name attribute in the CloudFormation json provided to it.

bblommers:
@jweite 
>  I'm thinking to declare a well-known method name for this that the cfn code can call on the model to ask it to evaluate the template properties  

This is the best way to go, IMO. There are too many resource-types to have that logic centralized in one place.
There's a draft PR I started for another issue, that you might be able to use as a base: #3185 
Would be good to make the 'well_known_method' (whatever the name will be) also part of the `CloudFormationModel`-class

jweite:
@bblommers, I read through your draft PR commit and get where you're coming from.  Makes sense to me, and I can see how it can cover a more delegated approach to other CloudFormation processing needs like determining update-type.

While I'd like to push forward with this issue, I don't want us to be cursing each other out at merge time.  Do you have any ETA for merging your PR 3127 work soon?  As a draft I imagine you wanted to incubate the idea a while.  I don't want to get ahead of that process by adding _CloudFormationModel_ (with an _evaluate_cloudformation_update_type()_ method) before you do.

[...]

PR #3199, opened 2020-07-29
S3 cloudformation update

jweite:
Re Issue #3187

Mere's a first cut for early review and discussion.  It utilizes the abstract base class technique @bblommers  prototyped in PR3185 to move CloudFormation resource type and name attribute to the model object.  I had anticipated needing that functionality to allow the cloudformation module to determine the type of an update from a cloudformation_update_type() method in the model object and exclude the forcing of name params into the update json, but subsequent guidance from @bblommers pointed toward a technique where the name param is only forced into the cloudformation_json for creates and deletes.  So the abstract base class isn't strictly needed here after all, but since it's a valuable step toward cloudformation support decentralization, I'm happy to leave it in unless there are objections.

Key thing to point out are that the S3 FakeS3Bucket now have create, uddate and delete_from_cloudformation_json methods.  The update_from_cloudformation_json method determines whether the update is a replacement update, and if so, creates a new bucket and deletes the old.  Otherwise it updates-in-place.

Unit tests of in-place update (encryption) and replacement update (name) are present.  Still TBD is a test to confirm that a replacement update using the same name fails.

So far I only support the Encryption properties.  Property support will get a little verbose, so I wanted to do initial review before getting into that.

### DIFF
--- a/moto/autoscaling/models.py
+++ b/moto/autoscaling/models.py
@@ -6,7 +6,7 @@ from boto.ec2.blockdevicemapping import BlockDeviceType, BlockDeviceMapping
 from moto.ec2.exceptions import InvalidInstanceIdError
 
 from moto.compat import OrderedDict
-from moto.core import BaseBackend, BaseModel
+from moto.core import BaseBackend, BaseModel, CloudFormationModel
 from moto.ec2 import ec2_backends
 from moto.elb import elb_backends
 from moto.elbv2 import elbv2_backends
@@ -74,7 +74,7 @@ class FakeScalingPolicy(BaseModel):
             )
 
 
-class FakeLaunchConfiguration(BaseModel):
+class FakeLaunchConfiguration(CloudFormationModel):
     def __init__(
         self,
         name,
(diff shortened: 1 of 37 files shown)


jweite:
Another consult @bblommers.  As you see some transformation of the cloudformation json into AWS API terms was necessary, and expect that similar will be required for the other properties.  I elected to do that in straight python in funcs in the newly added s3 cloud_formation.py (to be renamed cloudformation.py...) as this was the least verbose.  But IMO the code isn't intuitively obvious to the reader as doing what's conceptually a very simple transform.  Use of ordered dicts to preserve the original XML ordering may be a complicator.

Are there any transformation libraries / techniques in use in moto that might be well used for these functions and result in clearer code?

bblommers:
I'm not aware of any transformation libraries in use @jweite. The current approach is the most used technique for conversions like that.
Some function-level documentation on the nature of the transformation would be enough for me, to help with the clarity.

jweite:
Thanks @bblommers.  I'm in need of better Kinesis and IAM CloudFormation support, which I've started in another branch.   If you've any comments on this work I'd appreciate it, so I can align the other efforts to it.

jweite:
Thanks for the review @bblommers.  Very interested to see how #3201 evolve.

I've been plowing ahead with more complete CloudFormation handling for Kinesis Streams and IAM (needed for a project I'm working, as usual) that builds upon this foundation.  In addition to cloudformation_name_type() and cloudformation_type() 'm consistently adding:

- create_from_cloudformation_json()
- update_from_cloudformation_json()
- delete_from_cloudformation_json()
- @property physical_resource_id()

which are used by the standard cloudformation code (parsing.py).  Perhaps they might go into the CloudFormationModel ABC?

jweite:
Just saw #3201.  Awesome breadth @richford & @bblommers!  I'll rebase my in-progress IAM and Kinesis Cloudformation work on this.  @bblommers, let me know if the s3 work merges cleanly, and if not, I'll wrangle it to be based on 3201 too.

PR #3200, opened 2020-07-29
Organizations - implement Delegated Administrator functionality

gruebel:
I tried my best to mimic AWS behaviour, but it definitely feels a bit inconsistent. I also changed some of the RESTErrors to dedicated Exceptions.

### DIFF
--- a/moto/organizations/models.py
+++ b/moto/organizations/models.py
@@ -15,6 +15,8 @@ from moto.organizations.exceptions import (
     AccountNotFoundException,
     ConstraintViolationException,
     AccountAlreadyRegisteredException,
+    AWSOrganizationsNotInUseException,
+    AccountNotRegisteredException,
 )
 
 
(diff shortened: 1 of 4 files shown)


moto/organizations/models.py
```diff
@@ -606,5 +643,94 @@ def disable_aws_service_access(self, **kwargs):
...
+            )
+            if account:
+                raise AccountNotRegisteredException
+
+            raise AccountNotFoundException
+
+        admin.remove_service_principal(service)
+
+        if not admin.services:
+            self.admins.remove(admin)
```
bblommers, line 733: Is this intended to be (essentially) a silent failure?

bblommers:
Build is failing, @gruebel: https://travis-ci.org/github/spulec/moto/builds/713072295

gruebel:
Yeah 😄 
I was surprised all the builds were failing, usually the moto server tests fail for me from time to time.

moto/organizations/models.py
gruebel: I'm not sure, why there should be any silent failure here 🤔 
The idea behind this part is, when an account has no delegated admin permissions anymore, it has to be removed from the list of delegated admins completely.

bblommers: Right, gotcha. So an admin always has >=1 service. Wasn't sure about the expected behaviour.

bblommers:
Would you mind merging with master, @gruebel? Happy to merge this afterwards

moto/organizations/models.py
gruebel: yeah, correct!

gruebel:
@bblommers all done and the build is green 😄

bblommers:
This is now part of moto >= 1.3.15.dev1013

PR #3202, opened 2020-07-30
Config for IAM roles and policies

nick123pig:
I've tried to add some new AWS Config resources. Before I get too far into adding more resources, I want to make sure that my approach is in line with what's in the CONFIG_README.

### DIFF
--- a/moto/config/models.py
+++ b/moto/config/models.py
@@ -69,29 +69,6 @@ RESOURCE_MAP = {
     "AWS::IAM::Policy": policy_config_query,
 }
 
-CONFIG_REGIONS = [
-    "af-south-1",
-    "ap-east-1",
-    "ap-northeast-1",
-    "ap-northeast-2",
-    "ap-south-1",
-    "ap-southeast-1",
-    "ap-southeast-2",
-    "ca-central-1",
-    "eu-central-1",
-    "eu-north-1",
-    "eu-south-1",
-    "eu-west-1",
-    "eu-west-2",
-    "eu-west-3",
-    "me-south-1",
-    "sa-east-1",
-    "us-east-1",
-    "us-east-2",
-    "us-west-1",
-    "us-west-2",
-]
-
 
 def datetime2int(date):
     return int(time.mktime(date.timetuple()))
--- a/moto/config/models.py
+++ b/moto/config/models.py
@@ -48,6 +48,7 @@ from moto.config.exceptions import (
 from moto.core import BaseBackend, BaseModel
 from moto.s3.config import s3_account_public_access_block_query, s3_config_query
 from moto.core import ACCOUNT_ID as DEFAULT_ACCOUNT_ID
+
 from moto.iam.config import role_config_query, policy_config_query
 
 POP_STRINGS = [
@@ -67,6 +68,29 @@ RESOURCE_MAP = {
     "AWS::IAM::Role": role_config_query,
     "AWS::IAM::Policy": policy_config_query,
 }
+
+CONFIG_REGIONS = [
+    "af-south-1",
+    "ap-east-1",
+    "ap-northeast-1",
+    "ap-northeast-2",
+    "ap-south-1",
+    "ap-southeast-1",
+    "ap-southeast-2",
+    "ca-central-1",
+    "eu-central-1",
+    "eu-north-1",
+    "eu-south-1",
+    "eu-west-1",
+    "eu-west-2",
+    "eu-west-3",
+    "me-south-1",
+    "sa-east-1",
+    "us-east-1",
+    "us-east-2",
+    "us-west-1",
+    "us-west-2",
+]
 
 
 def datetime2int(date):
@@ -979,6 +1003,7 @@ class ConfigBackend(BaseBackend):
                 limit,
                 next_token,
                 resource_region=resource_region,
+                aggregator=self.config_aggregators.get(aggregator_name).__dict__,
             )
 
         resource_identifiers = []
@@ -989,7 +1014,6 @@ class ConfigBackend(BaseBackend):
                 "ResourceType": identifier["type"],
                 "ResourceId": identifier["id"],
             }
-
             if identifier.get("name"):
                 item["ResourceName"] = identifier["name"]
 
--- a/moto/config/models.py
+++ b/moto/config/models.py
@@ -47,8 +47,8 @@ from moto.config.exceptions import (
 
 from moto.core import BaseBackend, BaseModel
 from moto.s3.config import s3_account_public_access_block_query, s3_config_query
-
 from moto.core import ACCOUNT_ID as DEFAULT_ACCOUNT_ID
+from moto.iam.config import role_config_query, policy_config_query
 
 POP_STRINGS = [
     "capitalizeStart",
@@ -64,6 +64,8 @@ DEFAULT_PAGE_SIZE = 100
 RESOURCE_MAP = {
     "AWS::S3::Bucket": s3_config_query,
     "AWS::S3::AccountPublicAccessBlock": s3_account_public_access_block_query,
+    "AWS::IAM::Role": role_config_query,
+    "AWS::IAM::Policy": policy_config_query,
 }
 
 
--- a/moto/core/models.py
+++ b/moto/core/models.py
@@ -27,8 +27,8 @@ from .utils import (
     convert_flask_to_responses_response,
 )
 
-
 ACCOUNT_ID = os.environ.get("MOTO_ACCOUNT_ID", "123456789012")
+CONFIG_BACKEND_DELIM = "\x1e"  # Record Seperator "RS" ASCII Character
 
 
 class BaseMockAWS(object):
@@ -768,15 +768,29 @@ class ConfigQueryModel(object):
 
     def aggregate_regions(self, path, backend_region, resource_region):
         """
-        Returns a list of "region\1eresourcename" strings
+        This method will is called for both aggregated and non-aggregated calls for config resources. 
+        It will figure out how to return the full list of resources for a given regional backend and append them to a final list.
+        It produces a list of both the region and the resource name with a delimiter character (CONFIG_BACKEND_DELIM, ASCII Record separator, \x1e).
+        IE: "us-east-1\x1ei-1234567800" 
+        
+        Each config-enabled resource has a method named `list_config_service_resources` which has to parse the delimiter  
+        ...
+        :param path: - A dict accessor string applied to the backend that locates the resource.
+        :param backend_region:
+        :param resource_region:
+        :return: - Returns a list of "region\x1eresourcename" strings
         """
 
         filter_region = backend_region or resource_region
         if filter_region:
             filter_resources = list(self.backends[filter_region].__dict__[path].keys())
-            return map(
-                lambda resource: "{}\1e{}".format(filter_region, resource),
-                filter_resources,
+            return list(
+                map(
+                    lambda resource: "{}{}{}".format(
+                        filter_region, CONFIG_BACKEND_DELIM, resource
+                    ),
+                    filter_resources,
+                )
             )
 
         # If we don't have a filter region
@@ -784,7 +798,7 @@ class ConfigQueryModel(object):
         for region in self.backends:
             this_region_resources = list(self.backends[region].__dict__[path].keys())
             for resource in this_region_resources:
-                ret.append("{}\1e{}".format(region, resource))
+                ret.append("{}{}{}".format(region, CONFIG_BACKEND_DELIM, resource))
         return ret
 
 
--- a/moto/core/models.py
+++ b/moto/core/models.py
@@ -722,8 +722,6 @@ class ConfigQueryModel(object):
         :param backend_region: The region for the backend to pull results from. Set to `None` if this is an aggregated query.
         :param resource_region: The region for where the resources reside to pull results from. Set to `None` if this is a
                                 non-aggregated query.
-        :param aggregator: If an aggregated query, this will be the `ConfigAggregator instance from the backend. Set to `None`
-                           if a non-aggregated query. Useful if you need special logic based off the aggregator (ie IAM)
         :return: This should return a list of Dicts that have the following fields:
             [
                 {
--- a/moto/core/models.py
+++ b/moto/core/models.py
@@ -766,6 +766,27 @@ class ConfigQueryModel(object):
         """
         raise NotImplementedError()
 
+    def aggregate_regions(self, path, backend_region, resource_region):
+        """
+        Returns a list of "region\1eresourcename" strings
+        """
+
+        filter_region = backend_region or resource_region
+        if filter_region:
+            filter_resources = list(self.backends[filter_region].__dict__[path].keys())
+            return map(
+                lambda resource: "{}\1e{}".format(filter_region, resource),
+                filter_resources,
+            )
+
+        # If we don't have a filter region
+        ret = []
+        for region in self.backends:
+            this_region_resources = list(self.backends[region].__dict__[path].keys())
+            for resource in this_region_resources:
+                ret.append("{}\1e{}".format(region, resource))
+        return ret
+
 
 class base_decorator(object):
     mock_backend = MockAWS
--- a/moto/core/models.py
+++ b/moto/core/models.py
@@ -768,16 +768,22 @@ class ConfigQueryModel(object):
 
     def aggregate_regions(self, path, backend_region, resource_region):
         """
-        This method will is called for both aggregated and non-aggregated calls for config resources.
+        This method is called for both aggregated and non-aggregated calls for config resources.
+
         It will figure out how to return the full list of resources for a given regional backend and append them to a final list.
         It produces a list of both the region and the resource name with a delimiter character (CONFIG_BACKEND_DELIM, ASCII Record separator, \x1e).
         IE: "us-east-1\x1ei-1234567800"
 
-        Each config-enabled resource has a method named `list_config_service_resources` which has to parse the delimiter
+        You should only use this method if you need to aggregate resources over more than one region.
+        If your region is global, just query the global backend directly in the `list_config_service_resources` method
+
+        If you use this method, your config-enabled resource must parse the delimited string in it's `list_config_service_resources` method.
         ...
-        :param path: - A dict accessor string applied to the backend that locates the resource.
-        :param backend_region:
-        :param resource_region:
+        :param path: - A dict accessor string applied to the backend that locates resources inside that backend.
+                       For example, if you passed path="keypairs", and you were working with an ec2 moto backend, it would yield the contents from
+                       ec2_moto_backend[region].keypairs
+        :param backend_region: - Only used for filtering; A string representing the region IE: us-east-1
+        :param resource_region: - Only used for filtering; A string representing the region IE: us-east-1
         :return: - Returns a list of "region\x1eresourcename" strings
         """
 
--- a/moto/core/models.py
+++ b/moto/core/models.py
@@ -28,7 +28,6 @@ from .utils import (
 )
 
 ACCOUNT_ID = os.environ.get("MOTO_ACCOUNT_ID", "123456789012")
-CONFIG_BACKEND_DELIM = "\x1e"  # Record Seperator "RS" ASCII Character
 
 
 class BaseMockAWS(object):
@@ -723,6 +722,8 @@ class ConfigQueryModel(object):
         :param backend_region: The region for the backend to pull results from. Set to `None` if this is an aggregated query.
         :param resource_region: The region for where the resources reside to pull results from. Set to `None` if this is a
                                 non-aggregated query.
+        :param aggregator: If an aggregated query, this will be the `ConfigAggregator instance from the backend. Set to `None`
+                           if a non-aggregated query. Useful if you need special logic based off the aggregator (ie IAM)
         :return: This should return a list of Dicts that have the following fields:
             [
                 {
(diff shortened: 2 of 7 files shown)


mikegrima:
Excellent! I've been wanting to add this feature myself. I will try to take a look at this later tomorrow.

moto/core/models.py
```diff
@@ -715,6 +715,27 @@ def get_config_resource(
         """
         raise NotImplementedError()
 
+    def aggregate_regions(self, path, backend_region, resource_region):
+        """
+        Returns a list of "region\1eresourcename" strings
```
mikegrima, line 720: What is the `1e`?

mikegrima: Update the comments to be clear that this is to be used for both aggregated and non-aggregated calls. It will figure out how to return the full list of resources for a given regional backend.

I'm going to assume that this _doesn't_ work for global types (or the odd partial global types, like S3)?

moto/iam/config.py
```diff
@@ -0,0 +1,173 @@
+import json
+
+from moto.core.exceptions import InvalidNextTokenException
+from moto.core.models import ConfigQueryModel
+from moto.iam import iam_backends
+
+
+class RoleConfigQuery(ConfigQueryModel):
+    def list_config_service_resources(
```
mikegrima, line 9: Because the Config support is ... to say the least complex :), please clone over the docstrings similar to what I wrote for S3, but tailored for IAM.

This will be helpful for future implementations of resource types that are regional. Currently S3 is the only one implemented and S3 is a super edge case due to how it works (it's globally name-spaced but regionally resident).

moto/iam/config.py
```diff
@@ -0,0 +1,173 @@
...
+
+        role = self.backends["global"].roles.get(resource_id, {})
+
+        if not role:
+            return
+
+        if resource_name and role.name != resource_name:
+            return
+
+        # Format the bucket to the AWS Config format:
```
mikegrima, line 74: s/bucket/iam resource

moto/iam/config.py
```diff
@@ -0,0 +1,173 @@
...
+        resource_ids,
+        resource_name,
+        limit,
+        next_token,
+        backend_region=None,
+        resource_region=None,
+    ):
+        # For aggregation -- did we get both a resource ID and a resource name?
+        if resource_ids and resource_name:
+            # If the values are different, then return an empty list:
```
mikegrima, line 100: This logic is true for S3, but is it true for IAM managed policies?

moto/iam/config.py
```diff
@@ -0,0 +1,173 @@
...
+
+        policy = self.backends["global"].managed_policies.get(resource_id, {})
+
+        if not policy:
+            return
+
+        if resource_name and policy.name != resource_name:
+            return
+
+        # Format the bucket to the AWS Config format:
```
mikegrima, line 158: s/bucket/iam resource

moto/iam/config.py
```diff
@@ -0,0 +1,173 @@
...
+        resource_ids,
+        resource_name,
+        limit,
+        next_token,
+        backend_region=None,
+        resource_region=None,
+    ):
+        # For aggregation -- did we get both a resource ID and a resource name?
+        if resource_ids and resource_name:
+            # If the values are different, then return an empty list:
```
mikegrima, line 20: This logic is true for S3, but is it true for IAM roles?

I know that Roles have an ID that is some `AROAxxx` format, and not the name of the role.

tests/test_iam/test_iam.py
```diff
@@ -2879,3 +2880,450 @@ def test_delete_role_with_instance_profiles_present():
...
+def test_delete_account_password_policy_errors():
+    client = boto3.client("iam", region_name="us-east-1")
+
+    client.delete_account_password_policy.when.called_with().should.throw(
+        ClientError, "The account policy with name PasswordPolicy cannot be found."
+    )
+
+
+@mock_iam
+def test_role_list_config_discovered_resources():
```
mikegrima, line 2895: Add a test condition for filtering on name and ID. This should reveal an issue with IAM role names and IDs being different.

tests/test_iam/test_iam.py
```diff
@@ -2879,3 +2880,450 @@ def test_delete_role_with_instance_profiles_present():
...
+    # The role gets a random ID, so we have to grab it
+    role = result[0]
+    assert role["type"] == "AWS::IAM::Role"
+    assert len(role["id"]) == len(random_resource_id())
+    assert role["id"] == role["name"]
+    assert role["region"] == "global"
+
+
+@mock_iam
+def test_policy_list_config_discovered_resources():
```
mikegrima, line 2928: Ditto for the filtering test case.

tests/test_iam/test_iam.py
```diff
@@ -2879,3 +2880,450 @@ def test_delete_role_with_instance_profiles_present():
     role_names = [role["RoleName"] for role in iam.list_roles()["Roles"]]
     assert "Role1" in role_names
     assert "Role2" not in role_names
+
+
+@mock_iam
```
mikegrima, line 2885: Add in a test case with `@mock_config` where you test that the AWS Config APIs themselves properly return a correct IAM object back out.  It doesn't have to be too thorough, but it basically tests that the front end and backend logic all works together.

moto/core/models.py
nick123pig: `\1e` was actually supposed to be `\x1e` which is the ascii record separator. I changed it. I also updated the comments to make it clear.

moto/iam/config.py
nick123pig: complex doesn't even begin to describe it. I have tried to add a bunch more comments

moto/iam/config.py
nick123pig: fixed

moto/iam/config.py
nick123pig: it wasn't; removed

moto/iam/config.py
nick123pig: fixed

moto/iam/config.py
nick123pig: it wasn't; removed

tests/test_iam/test_iam.py
nick123pig: good catch; refactored the backend a little bit to return the role id and name separately

tests/test_iam/test_iam.py
nick123pig: fixed

tests/test_iam/test_iam.py
nick123pig: added this test, as well as some additional "test writing criteria" in the config readme to cover all the bases

mikegrima:
@nick123pig I'm going to _try_ to complete my review today. I have a bunch of other tasks on my plate I need to complete before I can dedicate time for this.

moto/iam/config.py
```diff
@@ -0,0 +1,190 @@
...
+            lambda policy: not policy.split(CONFIG_BACKEND_DELIM)[1].startswith(
+                "arn:aws:iam::aws"
+            ),
+            self.aggregate_regions("managed_policies", "global", None),
+        )
+
+        if not policy_list:
+            return [], None
+
+        # Pagination logic:
```
mikegrima, line 120: Ditto above. Need to filter the results before pagination.

tests/test_iam/test_iam.py
```diff
@@ -2879,3 +2880,661 @@ def test_delete_role_with_instance_profiles_present():
...
+        ConfigurationAggregatorName="test_aggregator",
+        Limit=1,
+    )
+    first_agg_result = agg_result["ResourceIdentifiers"][0]["ResourceId"]
+    assert agg_result["ResourceIdentifiers"][0]["ResourceType"] == "AWS::IAM::Role"
+    assert len(first_agg_result) == len(random_resource_id())
+    assert agg_result["ResourceIdentifiers"][0]["SourceAccountId"] == ACCOUNT_ID
+    assert agg_result["ResourceIdentifiers"][0]["SourceRegion"] == "global"
+
+    # Test aggregated pagination
```
mikegrima, line 3258: Try adding in a test with filtering and pagination. This should result in errors without the filtering logic implemented.

moto/core/models.py
```diff
@@ -766,6 +766,47 @@ def get_config_resource(
...
 
+    def aggregate_regions(self, path, backend_region, resource_region):
+        """
+        This method is called for both aggregated and non-aggregated calls for config resources.
+
+        It will figure out how to return the full list of resources for a given regional backend and append them to a final list.
+        It produces a list of both the region and the resource name with a delimiter character (CONFIG_BACKEND_DELIM, ASCII Record separator, \x1e).
+        IE: "us-east-1\x1ei-1234567800"
+
+        You should only use this method if you need to aggregate resources over more than one region.
```
mikegrima, line 777: This function was used previously in your code for both aggregation and non-aggregation contexts. Perhaps rename the function? (or remove it?)

moto/iam/config.py
```diff
@@ -0,0 +1,221 @@
...
+        # The resource name is a user-assigned string like "MyDevelopmentAdminRole"
+        # Stored in moto backend with the AWS-assigned random string like "AROA0BSVNSZKXVHS00SBJ"
+
+        # Grab roles from backend; need the full values since names and id's are different
+        role_list = list(self.backends["global"].roles.values())
+
+        if not role_list:
+            return [], None
+
+        # Filter by resource name or ids
```
mikegrima, line 31: Typo on precedence.

Also, for aggregation you need to add a check in for both resource names and resource IDs provided. You will need do something along the lines of:
```
        if resource_name or resource_ids:
            filtered_roles = []
            # resource_name takes precedence over resource_ids
            if resource_name:
                for role in role_list:
                    if role.name == resource_name:
                        filtered_roles = [role]
                        break

                # For aggregation: if both are passed in, need to verify both values are correct.
                if resource_ids:   
                    if role.id not in resource_ids:
                        return [], None

            else:
                for role in role_list:
                    if role.id in resource_ids:
                        filtered_roles.append(role)
```

moto/iam/config.py
```diff
@@ -0,0 +1,221 @@
...
+        # but it's default set be default, and moto's config doesn't yet support
+        # custom configuration recorders, we'll just behave as default.
+        policy_list = filter(
+            lambda policy: not policy.arn.startswith("arn:aws:iam::aws"), policy_list,
+        )
+
+        if not policy_list:
+            return [], None
+
+        # Filter by resource name or ids
```
mikegrima, line 136: Ditto above.

tests/test_iam/test_iam.py
```diff
@@ -2885,3 +2886,806 @@ def test_delete_role_with_instance_profiles_present():
...
+    agg_result = config_client.list_aggregate_discovered_resources(
+        ResourceType="AWS::IAM::Policy",
+        ConfigurationAggregatorName="test_aggregator",
+        Limit=1,
+    )
+    first_agg_result = agg_result["ResourceIdentifiers"][0]["ResourceId"]
+    assert agg_result["ResourceIdentifiers"][0]["ResourceType"] == "AWS::IAM::Policy"
+    assert len(first_agg_result) == len(random_policy_id())
+    assert agg_result["ResourceIdentifiers"][0]["SourceAccountId"] == ACCOUNT_ID
+    assert agg_result["ResourceIdentifiers"][0]["SourceRegion"] == "global"
```
mikegrima, line 3591: So in aggregated queries, the `SourceRegion` is not `global`. It's the region for where the config recorder resides. This should also be true for non-aggregated queries as well.

I think the _least offensive thing to do here_ is make this a static variable, like `ACCOUNT_ID`, which will allow users to mock that out to whatever region they want it to be.

And to add more complexity :), this will also impact the filtering for aggregated queries.

moto/core/models.py
nick123pig: I removed it for now; I can add it back when I do something that's truly "regional". This whole PR has been a bit of a learning journey.

tests/test_iam/test_iam.py
nick123pig: Good catch here. I went a bit of different direction (hopefully it's not too offensive). I am trying to replicate the IAM config functionality as close as possible to real life, and getting the "duplicate" resources are an important piece I'm trying to test in my application. 

I first went about it by treating IAM differently in `ConfigBackend#list_aggregate_discovered_resources`, but that just messed up all of the pagination/tokens, so I just moved that logic into the IAM specific config queries. That required me to add a new argument (the aggregator) into the signature for `list_config_service_resources`.

moto/config/models.py
```diff
@@ -68,6 +69,29 @@
     "AWS::IAM::Policy": policy_config_query,
 }
 
+CONFIG_REGIONS = [
```
mikegrima, line 72: Why is this required? Regions for the service should be part of botocore. If you do need to get a list of all active regions, you can do something along the lines of:

```
import boto3
...
CONFIG_REGIONS = boto3.Session().get_available_regions("config")
```

This looks at the botocore JSON model and does not make a call out to AWS.

moto/iam/config.py
```diff
@@ -45,10 +48,54 @@ def list_config_service_resources(
...
-        # sorted_role_ids matches indicies of sorted_roles
-        sorted_role_ids = list(map(lambda role: role.id, sorted_roles))
+        if aggregator:
+            # IAM is a little special; Roles are created in us-east-1 (which AWS calls the "global" region)
+            # However, the resource will return in the aggregator (in duplicate) for each region in the aggregator
+            # Therefore, we'll need to find out the regions where the aggregators are running, and then duplicate the resource there
+
+            # In practice, it looks like AWS will only duplicate these resources if you've "used" any roles in the region, but since
+            # we can't really tell if this has happened in moto, we'll just bind this to the regions in your aggregator
+            from moto.config.models import CONFIG_REGIONS
```
mikegrima, line 58: Import at the top.

moto/iam/config.py
```diff
@@ -45,10 +48,54 @@ def list_config_service_resources(
...
+                            ),  # this is only for sorting, isn't returned outside of this functin
+                            "type": "AWS::IAM::Role",
+                            "id": role.id,
+                            "name": role.name,
+                            "region": region,
+                        }
+                    )
+
+            # Pagination logic, sort by role id
+            sorted_roles = sorted(duplicate_role_list, key=lambda role: role["_id"])
```
mikegrima, line 88: Why do you need sorted roles and sorted role ids? With regards to the token selection below you can get the index of the role object, and then pull out the ID.

moto/iam/config.py
```diff
@@ -150,10 +206,55 @@ def list_config_service_resources(
...
+                            ),  # this is only for sorting, isn't returned outside of this functin
+                            "type": "AWS::IAM::Policy",
+                            "id": policy.id,
+                            "name": policy.name,
+                            "region": region,
+                        }
+                    )
+
+            # Pagination logic, sort by role id
+            sorted_policies = sorted(
```
mikegrima, line 246: Ditto above.

tests/test_iam/test_iam.py
mikegrima: I am curious about this. You should only see duplicates of resources in your aggregator if you have multiple in-account/in-region config recorders that record the same resource (and aggregating it in 1 spot).

For example:
If you have a config recorder in us-east-1, us-west-2, and eu-west-1 that are all configured to record IAM resources, AND those recorders are being aggregated by the aggregator `foo`, then when you query against `foo` you will see **3** results: one for each config region.

mikegrima: ^^ And this only applies to Global resources. For regional resources, the aggregator will list all resources regardless of the region per the note we left here https://github.com/spulec/moto/blob/master/moto/core/models.py#L706-L710.

IAM is an edge case here as it's unlike most AWS resource types :)

mikegrima: It might be easier for you to change the pagination logic to make the token a concatenation of the `role_id + config_recorder_region`.

moto/config/models.py
nick123pig: That's a handy trick! I had no idea you could do that. Replaced!

tests/test_iam/test_iam.py
nick123pig: @mikegrima Sorry - I didn't even see these responses. I should look closer next time!

I agree with your example and had a slight misunderstanding on how things worked. Of course, this now poses even more complexity. IAM returned aggregated resources are now to depend on:

- the existing list criteria already built in moto (filters, aggregation) 
- the recorder status (ie is the recorder running)
- the recorder setup (ie is it recording the resource type you want, as well is it a part of includeGlobalResourceTypes)

Also, I am not sure what "enables" the IAM policies to show up in the recorder, because if you have an "mostly untouched/empty" region, the IAM resources don't show up in both non-aggregated or aggregated queries. Phrased differently...I have 4 region-specific recorders running in my test AWS account, where two of the regions have "stuff" (instances, lambda functions, etc), but two don't. I only see the roles and policies in the regions where I have "stuff". Granted, I've only had those recorders "empty region" recorders on for ~4 hours, so maybe it's an issue with propagation. I do see 8 resources (default vpc's, etc) in those "empty regions". I'll check back in 24 hours to see if they show up.

I am not sure I follow your suggestion to mock out the recorder variable. It seems like I should instead just implement the resource config queries to read from the config recorders off the moto backend. I'd like to solidify the approach before I dive in. Does building "native" support for recorders sound like a good idea?
 
Also, apologies for all the back and forth; this is a learning process for me!

mikegrima: > the existing list criteria already built in moto (filters, aggregation)
the recorder status (ie is the recorder running)
the recorder setup (ie is it recording the resource type you want, as well is it a part of includeGlobalResourceTypes)

When I initially developed the Config feature, I didn't make it a requirement to pre-establish a recorder. I only made the requirement to pre-establish the aggregator for aggregated queries, as AWS Config will raise an exception if one is not existing. 

As such, now that I'm thinking of it... Maybe it _does_ make sense to just have it return all the regions as duplicates, in aggregated queries (since we assume that for the aggregator, it will query all the regional backends, which we document here: https://github.com/spulec/moto/blob/master/CONFIG_README.md#how-does-this-work). For non-aggregated queries, there won't be any duplicates, so that won't be an issue - we just assume that there is a recorder in that region recording that resource type.

mikegrima: If this makes sense then your PR should basically be done.

nick123pig: sorry for the delay; that sounds good to me!

mikegrima: I think the only thing left to do is get the unit tests to pass.

moto/core/models.py
```diff
@@ -723,6 +722,8 @@ def list_config_service_resources(
         :param backend_region: The region for the backend to pull results from. Set to `None` if this is an aggregated query.
         :param resource_region: The region for where the resources reside to pull results from. Set to `None` if this is a
                                 non-aggregated query.
+        :param aggregator: If an aggregated query, this will be the `ConfigAggregator instance from the backend. Set to `None`
```
mikegrima, line 725: It's been a while since I last looked at this...

Is this still needed?

moto/config/models.py
```diff
@@ -977,6 +980,7 @@ def list_aggregate_discovered_resources(
                 limit,
                 next_token,
                 resource_region=resource_region,
+                aggregator=self.config_aggregators.get(aggregator_name).__dict__,
```
mikegrima, line 983: Is providing the aggregator needed following the other changes that were made?

nick123pig: Hi There; sorry for the delay (just got back from holiday). The aggregator is needed - because it will need to know what regions to send back (either the `AllAwsRegions` flag or the list of `AwsRegions`).  I added a comment that tries to better explain it. I should be up-to-the-latest master and passing all the tests!

mikegrima: Taking a look now

moto/iam/config.py
```diff
@@ -0,0 +1,321 @@
...
+
+            # In practice, it looks like AWS will only duplicate these resources if you've "used" any roles in the region, but since
+            # we can't really tell if this has happened in moto, we'll just bind this to the regions in your aggregator
+            aggregated_regions = []
+            aggregator_sources = aggregator.get(
+                "account_aggregation_sources"
+            ) or aggregator.get("organization_aggregation_source")
+            for source in aggregator_sources:
+                source_dict = source.__dict__
+                if source_dict["all_aws_regions"]:
```
mikegrima, line 64: Will this always be present in the dictionary? I.e. should this use `source_dict.get("all_aws_regions")` instead?

moto/iam/config.py
```diff
@@ -0,0 +1,321 @@
...
+
+            # In practice, it looks like AWS will only duplicate these resources if you've "used" any policies in the region, but since
+            # we can't really tell if this has happened in moto, we'll just bind this to the regions in your aggregator
+            aggregated_regions = []
+            aggregator_sources = aggregator.get(
+                "account_aggregation_sources"
+            ) or aggregator.get("organization_aggregation_source")
+            for source in aggregator_sources:
+                source_dict = source.__dict__
+                if source_dict["all_aws_regions"]:
```
mikegrima, line 221: Same as above.

mikegrima (commented):
It all looks good, but please check the 2 comments I left. If that's good to go, then I will MERGE this in!!

moto/iam/config.py
mikegrima: Need to fix this one as well.

mikegrima (changes requested):
Forgot to fix line 221

mikegrima (approved):
Merging after passing tests!

PR #3204, opened 2020-07-30
Fix resource groups tests

gruebel:
With the new boto3 version the handling for resource groups changed.


gruebel:
Hi @bblommers,

could you rebuild this PR, it is a fix which is needed for the previous PRs. I don't understand why the batch test is failing, I had the same problem in the past. Very mysterious!

PR #3206, opened 2020-07-31
Fix DynamoDb2 ExpressionAttributeNames can start with a number

iainb:
When using pynamodb's support for transactions it makes use of of `ExpressionAttributeNames` that look like `#0` `#1` etc. According to the [ExpressionAttributeName documentation](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Expressions.ExpressionAttributeNames.html) and when testing against dynamodb-local these names work. However, when testing with moto they fail.

This PR modifies the regex for an expression attribute to allow for names that start with a number. This fix allows for tests in my codebase to pass.

As an aside - when reading the documentation it specifies that `ExpressionAttributeNames` must be alphanumeric, however, looking at the regex and moto tests `_` appears to be used. Possibly I'm slightly confused or this regex is used in multiple places??

### DIFF
--- a/moto/dynamodb2/parsing/tokens.py
+++ b/moto/dynamodb2/parsing/tokens.py
@@ -109,7 +109,7 @@ class ExpressionTokenizer(object):
 
     @classmethod
     def is_expression_attribute(cls, input_string):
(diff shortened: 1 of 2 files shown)


bblommers:
The underscore is probably another bug - will merge this for now.

PR #3208, opened 2020-08-02
fix: clear pending messages when call purge_queue

yktakaha4:
I'm using this product through localstack, and always helped me.

Compared to the real SQS behavior, it seems that some messages may not be retrieved after calling `purge_queue`.
I thought it was due to lack of clearing `_pending_messages` in `purge_queue`, but I'm not confident...
(I'm using FIFO Queue in my test code because that was exactly where I encountered the problem.)


bblommers:
This is now part of moto >= 1.3.15.dev1020

PR #3215, opened 2020-08-04
Lambda reserved concurrency

gmcrocetti:
Related with #3067. Still **missing** implementation for provisioned concurrency.

### DIFF
--- a/moto/awslambda/responses.py
+++ b/moto/awslambda/responses.py
@@ -377,7 +377,13 @@ class LambdaResponse(BaseResponse):
         pass
 
     def _delete_function_concurrency(self, request):
-        pass
+        function_name = self.path.rsplit("/", 2)[-2]
+        resp = self.lambda_backend.delete_function_concurrency(function_name)
+
+        if resp is None:
+            return 204, {}, "{}"
+        else:
+            return 404, {}, "{}"
 
     def _put_function_concurrency(self, request):
         function_name = self.path.rsplit("/", 2)[-2]
--- a/moto/awslambda/responses.py
+++ b/moto/awslambda/responses.py
@@ -143,7 +143,7 @@ class LambdaResponse(BaseResponse):
 
     def function_concurrency(self, request, full_url, headers):
         http_method = request.method
-        self.setup_class(request, full_url, headers)  # No idea about this
+        self.setup_class(request, full_url, headers)
 
         if http_method == "GET":
             return self._get_function_concurrency(request)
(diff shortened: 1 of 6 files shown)


moto/awslambda/responses.py
```diff
@@ -359,3 +372,31 @@ def _put_code(self):
             return 200, {}, json.dumps(resp)
         else:
             return 404, {}, "{}"
+
+    def _get_function_concurrency(self, request):
+        function_name = self.path.rsplit("/", 2)[-2]
+        resp = self.lambda_backend.get_function_concurrency(function_name)
+
+        if resp:
```
bblommers, line 380: The if-else statements (for all three, get/put/delete) don't seem to be necessary - the backend always returns one value

gmcrocetti: Ooops. I've forgot to validate function's name in the backend. We must return a 404 in this case, right ?

gmcrocetti:
Hey @bblommers, tests are failing at "server mode". I have no idea why, any tips ?

moto/awslambda/responses.py
bblommers: Yes, correct. This is the error as returned from AWS:
```
botocore.errorfactory.ResourceNotFoundException: An error occurred (ResourceNotFoundException) when calling the PutFunctionConcurrency operation: Function not found: [arn]
```

moto/awslambda/responses.py
```diff
@@ -359,3 +372,38 @@ def _put_code(self):
...
+
+    def _get_function_concurrency(self, request):
+        path_function_name = self.path.rsplit("/", 2)[-2]
+        function_name = self.lambda_backend.get_function(path_function_name)
+
+        if function_name is None:
+            return 404, {}, "{}"
+
+        resp = self.lambda_backend.get_function_concurrency(path_function_name)
+        return 200, {}, json.dumps({"ReservedConcurrentExecutions": resp})
```
bblommers, line 384: The concurrency-attribute might not be set yet, in which case the response would simply be `json.dumps({})`
Would be good to have a test case for this scenario as well

bblommers:
@gmcrocetti The master branch is failing as well - don't think it's related to these changes :+1:

moto/awslambda/responses.py
gmcrocetti: Nice catch !!

gmcrocetti: Hey @bblommers . Ran the same request using awscli in debug mode:
image
Looks like it's right :)

bblommers: Fair enough. I ran it using boto3 directly, and then the field isn't returned at all - I guess that's the internal logic in boto that strips None-fields :)

gmcrocetti:
Hi @bblommers , any updates ? Can we do smth to help you ?

bblommers:
Hi @gmcrocetti, the main branch is now passing again. Would you mind merging in the latest changes? That should make this build pass as well

gmcrocetti:
> Hi @gmcrocetti, the main branch is now passing again. Would you mind merging in the latest changes? That should make this build pass as well

Not working yet, failing due to a timeout. Can you please restart the job ?

bblommers (approved):

bblommers:
All done @gmcrocetti - thanks for your PR!

bblommers:
This is now part of moto >= 1.3.15.dev1029

PR #3218, opened 2020-08-06
Iam cloudformation update, singificant cloudformation refactoring

jweite:
I set out to add CloudFormation support for a few IAM entities I need in my current project, and quickly found myself adding special cases to funcs in the CloudFormation parsing.py to cope with things like un-specified names and deletion of resources without unique identities.  I found myself struggling to understand the resulting logic, with the troubling sense that it had unnecessarily grown unmaintainable, and perhaps some of the existing logic wasn't quite right.  IMO the generalized logic for creating, updating and deleting stacks should be pretty straightforward.  So I got up my courage and wrote them how I figured they should be, made my IAM models work with them, and then fixed up the rest.  

Listing my general strategies first:
-	Resource names never get injected into the json.  They are only provided to models' create/update/delete_from_cloudformation_json methods as parameters (resource_name / new_resource_name)
-	The name passed to these is _either_ the caller specified name from the json, _or_ a generated name.  The model’s CF code should generally not be responsible for deciding which to use.
-	Some models _DO_ need to generate their own name.  The CF code that follows create_from_cloudformation_json() gets the final name from the created resource.
-	update_from_cloudformation_json() is expected to delete/recreate the resource when a “replacement” type parameter changes.  While _name_ is the most common replacement parameter, others exist, and re-creation is required so that new generated names get used for the replacing resources if the caller didn’t specify an explicit name param.
-	update_from_cloudformation_json() is expected to _not_ delete/recreate the resource when a NoInterruption or UpdatesWithSomeInterruption parameter changes.  It must do an in-place update of the resource so the resource name doesn’t change, particularly if generated.
-	ResourceMap.update() and delete() look up the resource_name of the existing resource to be updated/deleted, they don't generate it.  (A generated name cannot not match that of the existing resource).  update_from_cloudformation_json() is provided a (proposed) new_resource_name (which contains either the user specified name or a generated name) and uses it for resource recreation in replacement updates.
-	delete_from_cloudformation_json generally should use resource_name to find the resource it wants to delete (when its resources are named, at least).   Note, though, that without a resource name a model probably cannot implement a reliable delete_from_cloudformation_json.  (i.e. when owning models get deleted before the sub-resources they own).
-	I want to avoid implementing model delete() methods.  Models can rarely delete themselves without help of their backend.  IMO a model shouldn’t know its backend (which is typically how delete()’s were implemented).  delete_from_cloudformation_json is a class-level method which IMO can more appropriately know its backend.

Some notes:

ResourceMap.delete() should be modified to delete in reverse-dependency order.  As is, there’s at least one case where a delete required the existence of the resource’s “parent” to find it, which had been deleted already.  I took the shortcut and re-coded it to be able to delete itself.

One interesting twist with the CloudFormationModel ABC: Not all of the current CloudFormationModel subclasses implement update/delete_from_cloudformation_json.  ResourceMap.delete() already introspects models to see if they have a delete() method, and needed to similarly do so if they have delete_from_cloudformation_json methods.  The ABC, though, assures that hasattr always returns true for delete_from_cloudformation_json on CloudFormationModel subclasses.  A technique of looking for the __isabstractmethod__ attribute on the function was found to reliably tell whether a subclass had implemented the method.  The method still silently ignores the non-deletes when no deletion methods are provided.

A similar technique was put into place for update_from_cloudformation_json that will cause a failure if it’s not implemented.  There are no tests that exhibit this problem today.  (Though end-users could experience it).

Some resources just don’t have physical ids.  (i.e. Route53 RecordSetGroups).   I haven’t messed with their XXX_from_cloudformation_json() methods, but without some sort of unique ID it will be challenging to implement delete_from_cloudformation successfully.  (Typically this func will need to know what resource “owns” the resource in question, and deletes it through an operation on its parent.  But a) the CF code for deletion doesn’t execute deletes in dependency order, so an owner can get deleted ahead of the resources it owns, and b) the json passed to delete may not always have all the info needed to find the resource in question.

The looped retrier may have been expected to deal with a, but in some circumstances one must look up the owner to do the deletion, so a premature deletion always thwarts that.

Model’s update_cloudformation_json must always determine if the update is are replacement update, deleting/re-creating if it is, updating in-place if it isn’t.  Some current ones only do one or the other, which is incorrect.  This logic is necessary so that a) generated names change upon replacement, and b) generated names don’t change otherwise.

The generate_resource_name has special cases in it.  These should be moved to the model via yet-another ABC method in a future update.

After making all unit tests pass, I did an inspection of the XXX_from_cloudformation_json methods of models that implement them and modified a few that were inappropriately doing their own resource naming.  A very few that required unique naming techniques were left as is.

### DIFF
--- a/moto/awslambda/models.py
+++ b/moto/awslambda/models.py
@@ -685,9 +685,12 @@ class EventSourceMapping(CloudFormationModel):
         )
 
         for esm in esms:
-            if esm.logical_resource_id in resource_name:
-                lambda_backend.delete_event_source_mapping
+            if esm.uuid == resource_name:
                 esm.delete(region_name)
+
+    @property
+    def physical_resource_id(self):
+        return self.uuid
 
 
 class LambdaVersion(CloudFormationModel):
--- a/moto/cloudformation/parsing.py
+++ b/moto/cloudformation/parsing.py
@@ -326,15 +326,20 @@ def parse_and_update_resource(logical_id, resource_json, resources_map, region_n
         logical_id, resource_json, resources_map
     )
     original_resource = resources_map[logical_id]
-    new_resource = resource_class.update_from_cloudformation_json(
-        original_resource=original_resource,
-        new_resource_name=new_resource_name,
-        cloudformation_json=resource_json,
-        region_name=region_name,
-    )
-    new_resource.type = resource_json["Type"]
-    new_resource.logical_resource_id = logical_id
-    return new_resource
+    if not hasattr(
+        resource_class.update_from_cloudformation_json, "__isabstractmethod__"
+    ):
+        new_resource = resource_class.update_from_cloudformation_json(
+            original_resource=original_resource,
+            new_resource_name=new_resource_name,
+            cloudformation_json=resource_json,
+            region_name=region_name,
+        )
+        new_resource.type = resource_json["Type"]
+        new_resource.logical_resource_id = logical_id
+        return new_resource
+    else:
+        return None
 
 
 def parse_and_delete_resource(resource_name, resource_json, resources_map, region_name):
--- a/moto/cloudformation/parsing.py
+++ b/moto/cloudformation/parsing.py
@@ -246,12 +246,18 @@ def generate_resource_name(resource_type, stack_name, logical_id):
         return "{0}{1}".format(
             stack_name[:max_stack_name_portion_len], right_hand_part_of_name
         ).lower()
+    elif resource_type == "AWS::IAM::Policy":
+        return "{0}-{1}-{2}".format(stack_name[:5], logical_id[:4], random_suffix())
     else:
         return "{0}-{1}-{2}".format(stack_name, logical_id, random_suffix())
 
 
 def parse_resource(
-    logical_id, resource_json, resources_map, add_name_to_resource_json=True
+    logical_id,
+    resource_json,
+    resources_map,
+    add_name_to_resource_json=True,
+    generate_name=True,
 ):
     resource_type = resource_json["Type"]
     resource_class = resource_class_from_type(resource_type)
@@ -264,9 +270,12 @@ def parse_resource(
         return None
 
     resource_json = clean_json(resource_json, resources_map)
-    resource_name = generate_resource_name(
-        resource_type, resources_map.get("AWS::StackName"), logical_id
-    )
+    if generate_name:
+        resource_name = generate_resource_name(
+            resource_type, resources_map.get("AWS::StackName"), logical_id
+        )
+    else:
+        resource_name = logical_id
     resource_name_property = resource_name_property_from_type(resource_type)
     if resource_name_property:
         if "Properties" not in resource_json:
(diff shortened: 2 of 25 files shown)


bblommers:
Hi @jweite , the main branch is now passing again. Would you mind merging in the latest changes? That should make this build pass as well

jweite:
Hmm, it's still failing: server mode tests.  I'll run those in the morning locally and see if it's me.

jweite:
Hi @bblommers, tests.test_awslambda.test_lambda.test_invoke_function_get_ec2_volume (in TEST_SERVER_MODE) is failing for me in the current master.  I think the key line in the stacktrace is:

line 283
 in send\n\xa0\xa0\xa0\xa0raise EndpointConnectionError(endpoint_url=request.url
 error=e)\n\x1b[32mEND RequestId: 25a3aee6-e5db-1d7b-dab8-d55bcc4a7b81\x1b[0m\n\x1b[32mREPORT RequestId: 25a3aee6-e5db-1d7b-dab8-d55bcc4a7b81\tInit Duration: 582.33 ms\tDuration: 3000.00 ms\tBilled Duration: 3000 ms\tMemory Size: 128 MB\tMax Memory Used: 47 MB\t\x1b[0m\n\n{"errorType":"EndpointConnectionError"
"errorMessage":"**Could not connect to the endpoint URL: \\"http://motoserver:5000/\\**""
"stackTrace":["  File \\"/var/task/lambda_function.py\\"

Also, for the first time in my experience I ran into file descriptor exhaustion (ulimit -n == 256) by server.py in TEST_SERVER_MODE.  From some light diagnosis It looks like the Lambda mocks may not be closing connections.  Generally server.py has 50-60 open files, but after the lambda tests run these increase to 280+ and remains around there until server.py termination.  lsof reports that these are mostly of the form:

_Python    65604 jweite   22u     unix 0xc710dc1eae8337cd         0t0             ->0xc710dc1eae83269d_

bblommers:
Hey @jweite, Travis is only unhappy about some missing regions in the new tests. (`boto3.client("iam")` should be `boto3.client("iam", region_name="..."`)

The `Could not connect to the endpoint URL: \"http://motoserver:5000/\"` error is just that the hostname isn't configured, as far as I know. How are you running these tests? It should work when following the same steps as configured in `.travis.yml`

I wasn't aware that the Lambda mocks wasn't not closing any connections - that sounds like a third, unrelated, issue :)

jweite:
Close!  
@bblommers, is the make lint fail on 3.7 caused by my commit?  I develop mainly in 3.7 and make lint passes locally.

bblommers:
@jweite No, there was an issue yesterday with the latest version of black that caused lint to fail.
That's fixed on master now - I'm restarting these builds to double check that it passes

bblommers (approved):
LGTM. Thanks again!

bblommers:
This is now part of moto >= 1.3.15.dev1033

jweite:
Awesome, thanks again for the review and merge @bblommers.

PR #3227, opened 2020-08-11
Convert `cognitoidp_backends.keys()` to list before attempting to access by index

soulshake:
Otherwise a `TypeError: 'dict_keys' object is not subscriptable` exception is raised in Python 3.


bblommers (changes requested):
Hi @soulshake, would you mind adding a unit test for this? It's not clear to me when this error occurs

bblommers:
This is now superceded by #3488

PR #3229, opened 2020-08-11
[SecretsManager] Fix put_secret_value function in case SecretId is not exist 

ninhkd:
In AWS PutSecretValue function document (https://docs.aws.amazon.com/secretsmanager/latest/apireference/API_PutSecretValue.html), the secret must already exist. I have fixed and update the integration tests. Please have a look. Thanks

Issue fixed:
https://github.com/localstack/localstack/issues/2831


bblommers (changes requested):
Thanks for the PR, @ninhkd! Would you mind merging with master? That should make this build pass now

bblommers:
This has already been fixed by #3413

PR #3232, opened 2020-08-12
Add update/delete actions from cloudformation for SNS service

ninhkd:
In SNS service, it already had create_from_cloudformation_json function but still missing update_from_cloudformation_json and delete_from_cloudformation_json functions. I have added these function and add integration tests to verify. Thanks

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,22 +1,12 @@
-FROM alpine:3.6
-
-RUN apk add --no-cache --update \
-    gcc \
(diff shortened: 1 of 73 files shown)


bblommers (changes requested):
Thanks for the PR, @ninhkd! Can you also add a test for boto3, with the non-deprecated decorators? (`@mock_cloudformation`, `mock_sns`)

PR #2151, opened 2019-04-13
Adds cognito-idp sign_up endpoint.

dpretty:
Adds cognito-idp sign_up endpoint. Related issue: #2150

PR #3249, opened 2020-08-20
cognito-idp sign_up

fadur:
Piggy backing off of PR #2151


bblommers (changes requested):
Thanks for your PR, @fadur! Would you mind adding a few more tests? Would be good to also cover `initiate_auth`, and some negative tests to cover the branches `user_pool is None` and `if not user`

fadur:
Hi @bblommers 
@xsphrx have been collaborating to add support for cognitoidp please see #3260 where initiate_auth is covered.

bblommers:
Looks like everything is now covered in #3260, so I'll close this. Thanks for the PR!

PR #3247, opened 2020-08-19
Set lambda credentials' variables in runtime

kam193:
AWS Lambda runtime sets a few envs, according to:
https://docs.aws.amazon.com/lambda/latest/dg/configuration-envvars.html#configuration-envvars-runtime

This PR set credentials and two other basic envs. It's another approach to fix #3225 (also try in #3220)

PR #3250, opened 2020-08-20
Security group egress ip permissions fix

jmbollard:
Updated moto/ec2/responses/security_groups.py:  Add support for Descriptions on egress rules.  This change 'broke' the response for the default egress rule on security group so I made the following change. 
Update moto/ec2/models.py: Change ip ranges on the default security group to be a dict instead of a string and removed for loop on ip_ranges looking for cidr of 0.0.0.0/0. This resolved issues caused by responses fix.  All ec2 tests complete successfully, but not sure if changing this to a dict had implications I missed.
Updated tests/test_ec2/test_security_groups.py: Test egress rules with Descriptions in, and updated revoke test.


kam193:
Hi,
since #3247 is already merged, you can now sync with master and drop copy&paste :)

bblommers (approved):
LGTM. Thanks @jmbollard and @spillin!

bblommers:
This is now part of moto >= 1.3.15.dev1030

Issue #3251, opened 2020-08-20
Send SMS message fails with topic not found


nykolaslima:
SNS allows publish directly to a phone number in order to send SMS messages.

Moto SNS does not support it since it gives an exception of topic not found when using publish operation directly to a phone number.

e.g.
```
self.sns_client.publish(
                    PhoneNumber="+5511964448877",
                    Message="testing SMS message")
```

This code throws the following error:
```
botocore.errorfactory.InvalidParameterValueException: An error occurred (ParameterValueInvalid) when calling the Publish operation: Could not find topic associated with phone number
```

nykolaslima:
As a workaround, you could create an SNS topic and subscribe to it using the phone number:
```
        sns = mock_sns()
        sns.start()

        sns_client = boto3.client('sns')
        sns_client.create_topic(Name="some-topic")
        resp = sns_client.create_topic(Name="some-topic")
        arn = resp["TopicArn"]
        sns_client.subscribe(TopicArn=arn, Protocol="sms", Endpoint='+5511964448877')
```

kam193:
Looks like duplicate of #2198

kam193:
@bblommers This can be also closed by #3253

bblommers:
Thanks for raising this @nykolaslima - this is now fixed as of moto >= 1.3.15.dev1027

nykolaslima:
Thank you ❤

PR #3253, opened 2020-08-20
[SNS] Mock sending directly SMS

### LINKED ISSUE #2198: Sending SMS via SNS without topic fails
Moto version: 1.3.8
Boto3 version: 1.9.42

Example:
```
import boto3
from moto import mock_sns
mock = mock_sns()
mock.start()
client = boto3.client('sns')
phone_number = "<insert valid E164 formatted phone number>"
client.publish(PhoneNumber=phone_number, Message='Test message'))
```

Expected:
No errors

Actual:
```
botocore.errorfactory.InvalidParameterValueException: An error occurred (ParameterValueInvalid) when calling the Publish operation: Could not find topic associated with phone number
```

Extra info:
The same code without Moto works and sends an SMS to my phone, also the AWS cli tool accepts only a phone number and message.

It seems moto tries to connect a phone number to a SNS topic, while this is not required by AWS. See https://github.com/spulec/moto/blob/master/moto/sns/responses.py#L304

kam193:
Proper behaviour when publishing to PhoneNumber is sending
message directly to this number, without any topic or
previous confirmation.

https://docs.aws.amazon.com/sns/latest/dg/sns-mobile-phone-number-as-subscriber.html

Fix #2198 and duplicated #3251

### DIFF
--- a/moto/ses/models.py
+++ b/moto/ses/models.py
@@ -202,7 +202,7 @@ class SESBackend(BaseBackend):
                 if sns_topic is not None:
                     message = self.__generate_feedback__(msg_type)
                     if message:
-                        sns_backends[region].publish(sns_topic, message)
+                        sns_backends[region].publish(message, arn=sns_topic)
 
     def send_raw_email(self, source, destinations, raw_data, region):
         if source is not None:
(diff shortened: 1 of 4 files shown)


bblommers (changes requested):
Hi @kam193, thanks for this PR! Good to see a fix for this now. 

Would you mind merging with master? That should fix any build errors

kam193:
Hi @bblommers! I rebased the PR earlier today. The fail was caused by my fault - in test I try to check local backend, also when tests run in server mode. I hope now it should pass

bblommers (approved):
LGTM! Thanks @kam193

PR #3254, opened 2020-08-21
Fix:s3 List Object response:delimiter 

usmangani1:
Fix:s3 List Object delimiter in response
https://github.com/localstack/localstack/issues/2859


bblommers (changes requested):
Hi @usmangani1, looks like the tests are failing after this change:
https://travis-ci.org/github/spulec/moto/jobs/722522741

bblommers (approved):
LGTM. Thanks @usmangani1!

PR #3260, opened 2020-08-25
added cognito-idp initiate_auth and PASSWORD_VERIFIER challenge to respond_to_auth_challenge

xsphrx:


bblommers (changes requested):
Hi @xsphrx, thanks for your PR!

Could you add negative tests as well? There are quit a few exception-cases - would be good to get test coverage for that

bblommers (approved):
LGTM. Thanks @xsphrx!

bblommers:
This is now part of moto >= 1.3.15.dev1037

PR #3261, opened 2020-08-26
Fix:SQS:Added Non existent queue name in ERROR RESPONSE

usmangani1:
Fix:SQS:Added Non existent queue name in ERROR RESPONSE
https://github.com/localstack/localstack/issues/2878


bblommers (approved):

bblommers:
This is now part of moto >= 1.3.15.dev1035

Issue #3262, opened 2020-08-26
SFN.Client.exceptions.ExecutionAlreadyExists Not implemented


ciaranevans:
**TL;DR: When calling `start_execution`, if a execution exists already with the provided name, throw `SFN.Client.exceptions.ExecutionAlreadyExists`**

## Expected behaviour:

When using a `mock_stepfunctions` like so:

```python
import boto3
from moto import mock_stepfunctions

def get_mock_step_function_arn(step_function_client):
    """
    Create a single node Step Function that passes
    """
    step_function = step_function_client.create_state_machine(
        name="TestStepFunction",
        roleArn="arn:aws:iam::123456789012:role/TestRole",
        definition=json.dumps(
            {
                "StartAt": "Pass",
                "States": {"Pass": {"Type": "Pass", "Result": "Blah", "End": True}},
            }
        ),
    )
    return step_function["stateMachineArn"]

@mock_stepfunctions
def test_that_duplicate_execution_fails():
    step_function_client = boto3.client('stepfunctions')
    mock_step_function_arn = get_mock_step_function_arn(step_function_client)
    
    step_function_client.start_execution(stateMachineArn=mock_step_function_arn, name="a-name", input="")
    # Below should throw SFN.Client.exceptions.ExecutionAlreadyExists
    step_function_client.start_execution(stateMachineArn=mock_step_function_arn, name="a-name", input="")
```

I expect the second call to `start_execution` to fail, throwing `SFN.Client.exceptions.ExecutionAlreadyExists` as the name of the execution has already been used.

## Actual behaviour:

The actual behaviour is the execution happens, and a `list_executions` call returns both the original and the duplicate.

## Versions

**moto** `1.3.14`
**boto** `2.49.0`
**boto3** `1.14.45`
**botocore** `1.17.45`

All were installed via `pip` and I'm *not* using the standalone `moto`

ciaranevans:
I'd be interested in contributing a fix to this, but don't want to dive in without bouncing ideas off folks first!

I'm thinking https://github.com/spulec/moto/blob/c0953a229f959de79d22f3ed60f273eaa1800fcf/moto/stepfunctions/models.py#L206-L216
would have something like
```python
 def _execution_exists(self, name)
     for execution in self.executions:
         if execution.name == name:
             return True
     return False

 def start_execution(self, state_machine_arn, name=None): 
     state_machine_name = self.describe_state_machine(state_machine_arn).name
     if _execution_exists(name):
[...]

PR #3263, opened 2020-08-26
Address SFN.Client.exceptions.ExecutionAlreadyExists Not implemented

ciaranevans:
Addresses: #3262 

* Adds `ExecutionAlreadyExists` AWS Error
* Pin black to `19.10b0`, otherwise we need to update `regex` and `click` (and reformat 500+ files)
* Adds test that confirms exception thrown when a duplicate execution is found
* Fix issue with `ecdsa` failing to install with `make init`


tests/test_stepfunctions/test_stepfunctions.py
```diff
@@ -404,6 +404,27 @@ def test_state_machine_start_execution_with_custom_name():
     execution["startDate"].should.be.a(datetime)
 
 
+@mock_stepfunctions
+@mock_sts
+def test_state_machine_start_execution_fails_on_duplicate_execution_name():
+    client = boto3.client("stepfunctions", region_name=region)
+    #
```
Ricool06, line 411: Unnecessary comment
```suggestion

```

setup.py
```diff
@@ -40,6 +40,7 @@ def get_version():
     "werkzeug",
     "PyYAML>=5.1",
     "pytz",
+    "ecdsa<0.15",
```
bblommers, line 43: Do we need this dependency?

bblommers (changes requested):
Thanks for the PR, @ciaranevans! Looks like the build is failing: https://travis-ci.org/github/spulec/moto/builds/721370544

setup.py
ciaranevans: Hi @bblommers, when I ran `make init` I received:
```
Installed /Users/ciaran/.pyenv/versions/3.7.8/envs/moto-3.7.8-fresh/lib/python3.7/site-packages/pyasn1-0.4.8-py3.7.egg
error: ecdsa 0.15 is installed but ecdsa<0.15 is required by {'python-jose'}
make: *** [init] Error 1
```

Hence me pinning `ecdsa` to less than 0.15.

Unsure why the Python 3.7 build is failing, it looks like `make lint` is falling over, but this ran fine locally for me (on `3.7.8`)

ciaranevans: The Python 2 build is failing on the `f` strings, is there a preferred method of handling this?

bblommers: String concatenation is usually done with the format-function: `"{0}".format(var)`, but old-school concatenation is acceptable as well.

Black was failing due to this issue: https://github.com/psf/black/issues/1628. Locally you might still have an old version of black, that's why it was working.
The build that is starting now should work again.. 🤞

ciaranevans: Awesome, I did get a ping from a friend about the black issue. Fingers crossed!

Thanks for the quick replies too @bblommers

ciaranevans: Looks like it's failed again on the lint

ciaranevans: So @bblommers I've taken a look, the new `black` version requires `regex` and `click` to be bumped. If we bump these and run `make lint`:

```
75 files would be reformatted, 583 files would be left unchanged.
``` 

I'd maybe argue that the bump in black should be a separate change, this PR could pin black to the pre 20 version?

ciaranevans: At a slight loss on build failures now, Python 2 looks like it's maybe because of the black formatting stripping the `u`'s? 

No idea why the recent Python 3.8 build failed https://travis-ci.org/github/spulec/moto/jobs/721473224 

Would appreciate some insights into these 😄

bblommers: The 3.8 failure was intermittant, so that's typically fixed by restarting the build.

Looks like pinning black worked though - it's all green now. 👍

bblommers (approved):
LGTM. Thanks for the PR, and investigating the black-conundrum!

setup.py
ciaranevans: Awesome!

Thanks for the merge, what's the usual turnaround from a master merge to a PyPi installable? 

Would you like me to raise and issue to bump up black and address the fun that entails?

bblommers: Turnaround is around 10 minutes! A dev-build is automatically triggered after every commit to master. 
This is now part of moto >= 1.3.15.dev1032

Black is already aware of these bugs: https://github.com/psf/black/issues/1628

ciaranevans: Magic! 

Looking forward to plugging it in and trying it out! I'll probably be back for some more Step Function stuff 😇

Issue #3138, opened 2020-07-16
ecs.create_service and ecs.describe_service do not handle tags correctly


mkielar:
In #2450 and #2338 handling tags has been added, however boto3 supports setting and reading tags in `create_service` and `describe_service`.

## Actual:
While one can call this:
```
service_matching_response: Dict = ecs_client.create_service(
            cluster=cluster_response['clusterArn'],
            serviceName=service_name_matching,
            taskDefinition=task_definition_response['taskDefinitionArn'],
            desiredCount=2,
            tags=[{
                'key': 'environment',
                'value': 'dev-i1234'
            }]
        )['service']
```
and it creates the service, the response from that method does not contain `tags` key.
Similarly, when calling:
```
result: Dict = self._ecs_client.describe_services(
            cluster=services_in_event.cluster_arn,
            services=services_in_event.service_arns,
            include=['TAGS']
)
```
the result does not contain `tags` key.

## Expected:
`tags` and `include` should be handled properly and `tags` should be returned.

ipmb:
Also applies to `describe_task_definition`

gruebel:
@bblommers can you close this issue, the functionality was added with above PR.

PR #3265, opened 2020-08-28
include=["TAGS"] for describe_task_definition

ipmb:
refs #3138 

First time working with `moto`. Is this the right approach for a fix? Let me know and I can apply it to `ecs.describe_service` and write tests.

### DIFF
--- a/moto/ecs/responses.py
+++ b/moto/ecs/responses.py
@@ -86,10 +86,10 @@ class EC2ContainerServiceResponse(BaseResponse):
 
     def describe_task_definition(self):
         task_definition_str = self._get_param("taskDefinition")
-        data, tags = self.ecs_backend.describe_task_definition(task_definition_str)
+        data = self.ecs_backend.describe_task_definition(task_definition_str)
         resp = {"taskDefinition": data.response_object, "failures": []}
         if "TAGS" in self._get_param("include", []):
-            resp["tags"] = tags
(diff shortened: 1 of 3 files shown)


moto/ecs/responses.py
```diff
@@ -87,7 +87,10 @@ def list_task_definitions(self):
     def describe_task_definition(self):
         task_definition_str = self._get_param("taskDefinition")
         data = self.ecs_backend.describe_task_definition(task_definition_str)
-        return json.dumps({"taskDefinition": data.response_object, "failures": []})
+        resp = {"taskDefinition": data.response_object, "failures": []}
+        if "TAGS" in self._get_param("include", []):
+            resp["tags"] = self.ecs_backend.task_definitions[data.family][data.revision].tags
```
bblommers, line 92: Would be good to have a utility method for this, i.e. `self.ecs_backend.get_tags_for_task(..)`

bblommers (changes requested):
Hi @ipmb, thanks for your PR! This looks like a good approach, yes - just a minor niggle to make it more readable.

Keep in mind that you need to run `black` (in Python3) for the build to pass: https://travis-ci.org/github/spulec/moto/builds/721908719

moto/ecs/responses.py
ipmb: turns out there already is! I found `list_tags_for_resource`

bblommers (approved):
LGTM. Thanks @ipmb!

PR #3266, opened 2020-08-28
WIP: OpenAPI support for API Gateway

jeamland:
This is based on @ChocPanda's work. I've rebased it to the latest main branch and got all the tests passing. I'm posting this for further advice and discussion as to what else needs to be done.

### DIFF
--- a/moto/apigateway/models.py
+++ b/moto/apigateway/models.py
@@ -500,7 +500,7 @@ class RestAPI(BaseModel):
         self.stages = {}
         self.resources = {}
         self.models = {}
-        self.add_child("/")  # Add default child
+        self.default = self.add_child("/")  # Add default child
 
     def __repr__(self):
         return str(self.id)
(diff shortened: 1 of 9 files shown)


moto/apigateway/models.py
```diff
@@ -754,6 +765,86 @@ def get_rest_api(self, function_id):
...
+            raise InvalidOpenApiModeException()
+
+        if api_doc.get("swagger") is not None or (
+            api_doc.get("openapi") is not None and api_doc["openapi"][0] != "3"
+        ):
+            raise InvalidOpenApiDocVersionException()
+
+        if mode == "overwrite":
+            self.resources = {}
+            self.add_child("/")  # Add default child
```
bblommers, line 779: Do we need to assign this to self.default again?

moto/apigateway/models.py
```diff
@@ -754,6 +765,86 @@ def get_rest_api(self, function_id):
...
+                        ).items():
+                            self.create_method_response(
+                                function_id, resource.id, method_type, response_code
+                            )
+                    else:
+                        self.create_integration(
+                            function_id=function_id,
+                            resource_id=resource.id,
+                            method_type=method_type,
+                            integration_type=method_doc[
```
bblommers, line 811: This becomes a bit difficult to read. Might be better to assign temp variables first, before calling the `create_integration`-function
(I.e. `int_type = method_doc[".."]["type"]`, etc)

bblommers (changes requested):
Thanks for the PR, @ChocPanda and @jeamland! Left a few comments.

The tests as is cover the basic use cases, but it would be good to extend them a bit. Negative tests, invalid inputs, overwriting API's and verifying that the originals are gone, etc

Issue #3222, opened 2020-08-09
Memory and CPU missing in ECS task definition


ipmb:
Calling the `ecs` client's `register_task_definition` method with `cpu` and `memory` does not save them as part of the task definition.

```python
>>> import os
>>> os.environ['AWS_ACCESS_KEY_ID'] = 'testing'
>>> os.environ['AWS_SECRET_ACCESS_KEY'] = 'testing'
>>> os.environ['AWS_SESSION_TOKEN'] = 'testing'
>>> os.environ['AWS_SECURITY_TOKEN'] = 'testing'
>>> os.environ['AWS_DEFAULT_REGION'] = 'us-east-1'
>>> import moto
>>> mock = moto.mock_ecs()
>>> mock.start()
>>> import boto3
>>> ecs = boto3.client("ecs")
>>> kwargs = {'containerDefinitions': [{'environment': [{'name': 'APPNAME', 'value': 'my-app'}, {'name': 'CNB_PROCESS_TYPE', 'value': 'release'}], 'image': '123456789012.dkr.ecr.us-east-1.amazonaws.com/my-app-stage:b35ce1cc53cd399d4c1143b4506c9820c2e3e739', 'linuxParameters': {'initProcessEnabled': True}, 'logConfiguration': {'logDriver': 'awslogs', 'options': {'awslogs-group': '/aws/ecs/default/my-app', 'awslogs-region': 'us-east-1', 'awslogs-stream-prefix': 'release'}}, 'name': 'release', 'secrets': []}], 'cpu': '256', 'executionRoleArn': 'arn:aws:iam::123456789012:role/paaws/ecs-tasks-execution-my-app-service-role', 'memory': '512'}
>>> resp = ecs.register_task_definition(family='my-app-release', **kwargs)
>>> resp["taskDefinition"]
[...]

PR #3267, opened 2020-08-28
ECS task definition cpu/memory

ipmb:
refs #3222 

The rules around this are a little confusing. I haven't covered all the use cases, but I think this is an improvement.

Going from https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task_definition_parameters.htm

## Task Definition

> When you register a task definition, you can specify the total cpu and memory used for the task. This is separate from the cpu and memory values at the container definition level. If using the EC2 launch type, these fields are optional. If using the Fargate launch type, these fields are required and there are specific values for both cpu and memory that are supported. 

### cpu

> This field is optional for tasks using the Fargate launch type, and the only requirement is that the total amount of CPU reserved for all containers within a task be lower than the task-level cpu value. 

### memory

> The total amount of memory reserved for all containers within a task must be lower than the task memory value, if one is specified.

> If using the Fargate launch type, this parameter is optional.

> If using the EC2 launch type, you must specify either a task-level memory value or a container-level memory value. If you specify both a container-level memory and memoryReservation value, memory must be greater than memoryReservation. If you specify memoryReservation, then that value is subtracted from the available memory resources for the container instance on which the container is placed. Otherwise, the value of memory is used. 

### memoryReservation

> If a task-level memory value is not specified, you must specify a non-zero integer for one or both of memory or memoryReservation in a container definition. If you specify both, memory must be greater than memoryReservation. If you specify memoryReservation, then that value is subtracted from the available memory resources for the container instance on which the container is placed. Otherwise, the value of memory is used.


bblommers (changes requested):
Looks like the build is failing: https://travis-ci.org/github/spulec/moto/builds/722091623

Would be good to get some tests for this as well

bpandola:
@ipmb Hi there. This had gone a bit stale.  I brought in the latest changes from `master` and re-ran CI.  The tests are still failing based on some of the changes you introduced.  Please let us know if you're still interested in working on this or if you've moved on.  Thanks.

ipmb:
Thanks @bpandola! I am still interested (I maintain this patch locally), but time has been hard to come by. I'll do my best.

bpandola:
Closing this due to lack of activity.  Feel free to update and re-open any time.

PR #3268, opened 2020-08-28
ecs.run_task count default to 1

ipmb:
This mimics boto3's undocumented behavior.


bblommers (changes requested):
Thanks for the PR, @ipmb! Can we get a simple unit test to cover this? Happy to merge after that

ipmb:
Test added. I piggybacked on an existing test to avoid duplicating the code to bootstrap the necessary dependencies.

ipmb:
@bblommers could you take another look at this one? I added a test and updated the branch.

bblommers (approved):
LGTM! Thanks again @ipmb

PR #3272, opened 2020-09-01
Fix deprecation due to base64.decodestring

tirkarthi:
Similar to https://github.com/spulec/moto/pull/1854


bblommers (approved):
LGTM. Thanks @tirkarthi!

PR #3274, opened 2020-09-01
Add introduction of using scaffold.py

toshitanian:
Add introduction of using scaffold.py. That can automatically generate template code of new services and method of existing services.


bblommers (approved):
Good idea to add this @toshitanian - thanks!

PR #3278, opened 2020-09-02
added organizations detach_policy response, model, and tests, issue #…

harvester843:
…3239

### DIFF
--- a/moto/organizations/models.py
+++ b/moto/organizations/models.py
@@ -821,5 +821,33 @@ class OrganizationsBackend(BaseBackend):
 
         return dict(Root=root.describe())
 
+    def detach_policy(self, **kwargs):
+        policy = self.get_policy_by_id(kwargs["PolicyId"])
+        if re.compile(utils.ROOT_ID_REGEX).match(kwargs["TargetId"]) or re.compile(
+            utils.OU_ID_REGEX
+        ).match(kwargs["TargetId"]):
+            ou = next((ou for ou in self.ou if ou.id == kwargs["TargetId"]), None)
+            if ou is not None:
+                if ou in ou.attached_policies:
+                    ou.attached_policies.remove(policy)
+                    policy.attachments.remove(ou)
+            else:
+                raise RESTError(
+                    "OrganizationalUnitNotFoundException",
+                    "You specified an organizational unit that doesn't exist.",
+                )
+        elif re.compile(utils.ACCOUNT_ID_REGEX).match(kwargs["TargetId"]):
+            account = next(
+                (a for a in self.accounts if a.id == kwargs["TargetId"]), None
+            )
+            if account is not None:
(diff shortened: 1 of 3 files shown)


moto/organizations/models.py
```diff
@@ -821,5 +821,33 @@ def disable_policy_type(self, **kwargs):
 
         return dict(Root=root.describe())
 
+    def detach_policy(self, **kwargs):
+        policy = self.get_policy_by_id(kwargs["PolicyId"])
+        if re.compile(utils.ROOT_ID_REGEX).match(kwargs["TargetId"]) or re.compile(
```
bblommers, line 826: This can be made more readable by extracting these into local/class variables.
Can we rewrite these regex-statements to say `root_id_regex.match(target_id)`?

Whether `root_id_regex` is defined doesn't matter much IMO, as long as it's only done once, instead of every call to `detach_policy`

tests/test_organizations/test_organizations_boto3.py
```diff
@@ -467,6 +467,90 @@ def test_attach_policy():
...
+    client.attach_policy(PolicyId=policy_id, TargetId=account_id)
+    with assert_raises(ClientError) as e:
+        response = client.detach_policy(PolicyId=policy_id, TargetId="r-xy85")
+    ex = e.exception
+    ex.operation_name.should.equal("DetachPolicy")
+    ex.response["Error"]["Code"].should.equal("400")
+    ex.response["Error"]["Message"].should.contain(
+        "OrganizationalUnitNotFoundException"
+    )
+    with assert_raises(ClientError) as e:
```
bblommers, line 526: Let's split these tests into separate methods. Ideally each test case only tests one error scenario at the time

bblommers (changes requested):
Hi @harvester843, thanks for your PR! I've left some comments

moto/organizations/models.py
harvester843: @bblommers I have updated based on your comments. I hope this satisfies the requirements.

tests/test_organizations/test_organizations_boto3.py
harvester843: @bblommers I have created separate tests for each exception

bblommers (approved):
LGTM. Thanks @harvester843!

dpowley:
@bblommers @harvester843 Can you all confirm that this should be?
`if policy in account.attached_policies:`

https://github.com/spulec/moto/blob/5aefbb1e51e0b97583bb96bd8c3c596dbca44a31/moto/organizations/models.py#L868

attach_policy has the same issue, but it defaults to working due to the `not in` condition.

bblommers:
@dpowley No, that doesn't look right - thanks for pointing it out. Will create a PR to fix this

Issue #3275, opened 2020-09-01
Enhance mock_iot to support describe_endpoint and register_certificate_without_ca


zhil3:
We wanna enhance the moto framework by adding two functions in **iot_mock** module.

Currently **describe_endpoint** and **register_certificate_without_ca** are what we need, and probably there will be more coming but depends on what we wanna implement in near future.

I have a branch already implemented the describe_endpoint with unittest working as well. 
Now working on register_certificate_without_ca. 

What we want is that we could push a branch so that we could git clone it in our pipeline, and have PR reviewed/merged as soon as possible that we don't have to keep a repo in our pipeline always.

Thank You.

zhil3:
register_certificate_without_ca  is done with unittest as well

bblommers:
Hi @zhil3, thanks for raising this. 

PR's are always welcome!

zhil3:
@bblommers Thanks for the reply. But I don't have permission to push a branch thus how to raise the PR?

zhil3:
Thanks @bblommers for the info. 
And could you please help up speedy up the PR process by any chances? 
We are kind of in an urgent status of using those functions inside our pipelines.

Thank You

PR #3279, opened 2020-09-02
Add describe_endpoint and register_certificate_without_ca in iot_mock…

zhil3:
Related issue: https://github.com/spulec/moto/issues/3275
Add **IOT** functions **describe_endpoint** and **register_certificate_without_ca**

### DIFF
--- a/moto/iot/models.py
+++ b/moto/iot/models.py
@@ -20,6 +20,7 @@ from .exceptions import (
     InvalidStateTransitionException,
     VersionConflictException,
 )
+from moto.utilities.utils import random_string
 
 
 class FakeThing(BaseModel):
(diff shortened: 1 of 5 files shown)


moto/iot/models.py
```diff
@@ -374,6 +374,61 @@ def to_dict(self):
...
+            "iot:Data-ATS",
+            "iot:CredentialProvider",
+            "iot:Jobs",
+        ]:
+            raise InvalidRequestException(
+                " An error occurred (InvalidRequestException) when calling the DescribeEndpoint "
+                "operation: Endpoint type %s not recognized." % endpoint_type
+            )
+        self.region_name = region_name
+        data_identifier = "".join(
```
bblommers, line 390: There's already a `_random_string` method in IoTBackend - would be good to extract that into a utils.py, and reuse it here

moto/iot/models.py
```diff
@@ -495,6 +551,12 @@ def describe_thing_type(self, thing_type_name):
             raise ResourceNotFoundException()
         return thing_types[0]
 
+    def describe_endpoint(self, endpoint_type):
+        self.endpoint = FakeEndpoint(endpoint_type, self.region_name)
```
bblommers, line 555: Can we assign this on initialization, and just return it here?
The `if not` seems definitely unnecessary

tests/test_iot/test_iot.py
```diff
@@ -463,6 +463,37 @@ def test_list_things_with_attribute_and_thing_type_filter_and_next_token():
...
 
 
+@mock_iot
+def test_endpoints():
+    client = boto3.client("iot", region_name="ap-northeast-1")
+
+    # iot:Data
+    endpoint = client.describe_endpoint(endpointType="iot:Data")
+    endpoint.should.have.key("endpointAddress").which.should_not.contain("ats")
+    endpoint.should.have.key("endpointAddress").which.should.contain("iot")
```
bblommers, line 473: I think it makes sense to make these checks more strict, i.e. `should.contain("iot.ap-northeast-1.amazonaws.com")`

bblommers (changes requested):
Hi @zhil3, thanks for your PR! Left a few comments

moto/iot/models.py
zhil3: Thinking better to follow other modules with a Object ( class ) to handle the returned values.
Remove if not.

tests/test_iot/test_iot.py
zhil3: sounds good.

moto/iot/models.py
zhil3: sure thing

moto/iot/models.py
zhil3: > Can we assign this on initialization, and just return it here?
> The `if not` seems definitely unnecessary

Move the assignment here, with using _random_string without put it into uilts.py which may cause other functions failures.
Remove if not.

zhil3: I think the class object needs to be there for endpoint since it will call return obj.to_dict() as others, will keep consistency in case there may be some usages incompatibilities ?
remove `if not`

bblommers (approved):
LGTM. Thanks @zhil3!

PR #3281, opened 2020-09-03
Fix #1728 by introducing per-service extras_require

### LINKED ISSUE #1728: Dependencies overhead

davidszotten:
Rebase of #2676 to fix merge conflicts

Hope you don't mind @slafs .  @spulec noted on #2676 that the current state was probably mergeable as is, so hoping we can do just that, since the merge conflicts in setup.py get tricky to resolve.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,2 +1,2 @@
 #  Please add requirements to setup.py
--e .
+-e .[all]
--- a/setup.py
+++ b/setup.py
@@ -33,22 +33,13 @@ install_requires = [
     "boto>=2.36.0",
     "boto3>=1.9.201",
     "botocore>=1.12.201",
-    "cryptography>=2.3.0",
     "requests>=2.5",
     "xmltodict",
     "six>1.9",
     "werkzeug",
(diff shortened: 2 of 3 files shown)


spulec:
Thanks!

PR #3284, opened 2020-09-03
Implement Execution inputs for Step Functions

### LINKED ISSUE #3283: Include ability to add input to Step Function Execution (and have it returned in describe_execution)
**TL;DR** Moto should actually handle `input` for the Execution model [here](https://github.com/spulec/moto/blob/master/moto/stepfunctions/models.py#L29-L47)

@bblommers I intend on implementing this, just raising the issue so I can A: link the PR to something and B: not forget I need to do this 🤣 

I'm planning on adding `input` as a property of `Execution` and implementing this back into [describe_execution](https://github.com/spulec/moto/blob/master/moto/stepfunctions/responses.py#L126-L141)

This is so that when a user runs `describe_execution` against a moto Step Function, they get something like [this](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/stepfunctions.html#SFN.Client.describe_execution):
```
{
    'executionArn': 'string',
    'stateMachineArn': 'string',
    'name': 'string',
    'status': 'RUNNING'|'SUCCEEDED'|'FAILED'|'TIMED_OUT'|'ABORTED',
    'startDate': datetime(2015, 1, 1),
    'stopDate': datetime(2015, 1, 1),
    'input': 'string',
    'output': 'string'
}
```

ciaranevans:
Closes #3283 

* Allows for `start_execution(<..>, input=<an-input>)` - Not just defaulting to `'{}'`
* Add test to ensure that a custom input is returned correctly by `describe_execution`
* Remove unnecessary `as exc`'s

### DIFF
--- a/moto/stepfunctions/models.py
+++ b/moto/stepfunctions/models.py
@@ -34,6 +34,7 @@ class Execution:
         state_machine_name,
         execution_name,
         state_machine_arn,
+        execution_input,
     ):
         execution_arn = "arn:aws:states:{}:{}:execution:{}:{}"
         execution_arn = execution_arn.format(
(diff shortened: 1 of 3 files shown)


bblommers (approved):
LGTM. Thanks @ciaranevans!

Issue #2992, opened 2020-05-14
Unsubscribe of in-existent SNS subscription does not throw NotFoundException


tomers:
The SNS model's unsubscribe fails on KeyError when trying to unsubscribe an SNS subscription that does not exists.

Instead, the code should be:
```
def unsubscribe(self, subscription_arn):
        try:
            self.subscriptions.pop(subscription_arn, None)
        except KeyError:
            raise SNSNotFoundError("Topic with arn {0} not found".format(subscription_arn))
```

Reference: https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/sns.html#SNS.Client.unsubscribe

tomers:
In `test_unsubscribe_from_deleted_topic` I notice:
```
    # Now delete hanging subscription
    client.unsubscribe(SubscriptionArn=subscription_arn)

    subscriptions = client.list_subscriptions()["Subscriptions"]
    subscriptions.should.have.length_of(0)

    # Deleting it again should not result in any error
    client.unsubscribe(SubscriptionArn=subscription_arn)
```

However in my code I get an exception...
My moto version: Docker taken from https://github.com/spulec/moto.git tags/1.3.14
```
Traceback (most recent call last):
  File "/usr/lib/python3.6/site-packages/werkzeug/serving.py", line 323, in run_wsgi
    execute(self.server.app)
  File "/usr/lib/python3.6/site-packages/werkzeug/serving.py", line 312, in execute
    application_iter = app(environ, start_response)
  File "/usr/lib/python3.6/site-packages/moto/server.py", line 135, in __call__
    return backend_app(environ, start_response)
  File "/usr/lib/python3.6/site-packages/flask/app.py", line 2464, in __call__
    return self.wsgi_app(environ, start_response)
  File "/usr/lib/python3.6/site-packages/flask/app.py", line 2450, in wsgi_app
    response = self.handle_exception(e)
  File "/usr/lib/python3.6/site-packages/flask/app.py", line 1867, in handle_exception
    reraise(exc_type, exc_value, tb)
  File "/usr/lib/python3.6/site-packages/flask/_compat.py", line 39, in reraise
    raise value
  File "/usr/lib/python3.6/site-packages/flask/app.py", line 2447, in wsgi_app
    response = self.full_dispatch_request()
  File "/usr/lib/python3.6/site-packages/flask/app.py", line 1952, in full_dispatch_request
    rv = self.handle_user_exception(e)
[...]

PR #3285, opened 2020-09-04
fix: deleting non-existent topic shouldn't raise KeyError

okeyokoro:
Deleting a non-existent topic shouldn't raise KeyError

Related Issues:
- https://github.com/spulec/moto/issues/2992 <- slightly different, here they called `unsubscribe` instead of `delete_topic`

### DIFF
--- a/moto/sns/models.py
+++ b/moto/sns/models.py
@@ -426,7 +426,10 @@ class SNSBackend(BaseBackend):
         return self._get_values_nexttoken(self.topics, next_token)
 
     def delete_topic(self, arn):
-        self.topics.pop(arn)
+        try:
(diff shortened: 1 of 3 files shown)


bblommers (changes requested):
Hi @okeyokoro, thanks for your PR! Looks like the tests are failing: https://travis-ci.org/github/spulec/moto/jobs/724034950

okeyokoro:
@bblommers thanks for picking this up! tests pass now 🙂

bblommers (approved):
LGTM. Thanks @okeyokoro!

bblommers:
This is now part of moto >= 1.3.16.dev2

PR #3286, opened 2020-09-04
Add Step Function Execution Input validation

ciaranevans:
* Validates that the input is a valid JSON string
* Adds `InvalidExecutionInput` from exceptions [here](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/stepfunctions.html#SFN.Client.start_execution)
* Adds another test to cover valid custom inputs to `start_execution`
* Non-string inputs are already caught via `moto.core.responses._get_param`

### DIFF
--- a/moto/stepfunctions/exceptions.py
+++ b/moto/stepfunctions/exceptions.py
@@ -38,6 +38,11 @@ class InvalidName(AWSError):
     STATUS = 400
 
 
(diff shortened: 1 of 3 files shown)


ciaranevans:
@bblommers CI looks happy on these 🥳

spulec:
Looks great; thank you!

PR #3287, opened 2020-09-04
Enhancement:Ec2:Vpc End Point Services

usmangani1:
Enhancement:Ec2:Vpc End Point Services

### DIFF
--- a/moto/ec2/responses/vpcs.py
+++ b/moto/ec2/responses/vpcs.py
@@ -193,8 +193,44 @@ class VPCs(BaseResponse):
 
     def describe_vpc_endpoint_services(self):
         vpc_end_point_services = self.ec2_backend.get_vpc_end_point_services()
-        template = self.response_template(DESCRIBE_VPC_ENDPOINT_RESPONSE)
+        template = self.response_template(DESCRIBE_VPC_ENDPOINT_SERVICES_RESPONSE)
         return template.render(vpc_end_points=vpc_end_point_services)
+
+    def describe_vpc_endpoints(self):
+        vpc_endpoint_ids = self._get_multi_param("VpcEndpointId")
+        vpc_end_points = self.ec2_backend.describe_vpc_end_point(vpc_endpoint_ids)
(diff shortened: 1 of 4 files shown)


moto/ec2/responses/vpcs.py
```diff
@@ -486,3 +522,65 @@ def describe_vpc_endpoint_services(self):
...
+                <requesterManaged>true</requesterManaged>
+                <rreationTimestamp>{{ vpc_end_point.created_at }}</rreationTimestamp>
+                <ownerId>12345678912</ownerId>
+            {% endfor %}
+        </item>
+    </vpcEndpointSet>
+</DescribeVpcEndpointsResponse>"""
+
+DELETE_VPC_ENDPOINT_RESPONSE = """<DeleteVpcEndpointsResponse xmlns="http://ec2.amazonaws.com/doc/2016-11-15/">
+    <unsuccessful/>
```
bblommers, line 578: Can we add a TODO comment here? From the docs:
`Information about the VPC endpoints that were not successfully deleted.`

Don't mind that this isn't implemented yet, but the wording makes it a bit confusing atm - my initial reading was that the entire deletion  was unsuccessful.

bblommers (changes requested):
Hi @usmangani1, thanks for the PR!

Can we extend the tests a bit? There are quite a few code branches that aren't covered yet, e.g.:
 - modify_vpc_endpoint(AddRouteTableIds=..)
 - modify_vpc_endpoint(AddSubnetIds=..)
 - modify_vpc_endpoint(RemoveSubnetIds=...)
 - modify_vpc_endpoint(AddSecurityGroupIds=..)
 - modify_vpc_endpoint(RemoveSecurityGroupIds=...)

as well as various negative tests, like when the delete or modify endpoint is called with an unknown VPC.

Ideally they all come in separate methods, to make it easier to read/debug later on.

bpandola:
Superseded by the following PRs, which together contain the API endpoints that were to be added:

#3524
#4169
#5530

PR #3293, opened 2020-09-08
#3290 - Add additional dependencies in extras_require

### LINKED ISSUE #3290: Fails to install cryptography for `moto-1.3.15`
Inside a python:3.8-slim docker container, `pip install moto` does not install `cryptography` on the latest moto release (moto-1.3.15). This causes subsequent commands to fail with `ModuleNotFoundError: No module named 'cryptography'` when invoking `moto` mocks.

To see this, jump into a fresh docker:
`docker run --rm -it python:3.8-slim /bin/bash`

Then inside run:
```
pip install moto
pip list
```
And note that `cryptography` is not installed.

Now run:
```
pip install moto==1.3.14
pip list
```

And note that `cryptography` is now installed.

An example of the failures seen when using the new release (ie. having only run `pip install moto`):
```
/usr/local/lib/python3.8/site-packages/moto/iam/__init__.py:2: in <module>
    from .models import iam_backend
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

    import base64
    import hashlib
    import os
    import random
    import string
    import sys
    from datetime import datetime
    import json
    import re
    
>   from cryptography import x509
E   ModuleNotFoundError: No module named 'cryptography'

/usr/local/lib/python3.8/site-packages/moto/iam/models.py:12: ModuleNotFoundError
```

bblommers:
Fixes #3290 

S3 is dependent on IAM, which is dependent on the cryptography-module.
This PR adds S3 as an additional service (`pip install moto[s3]`) to ensure it installs the required dependencies.

This also adds all dependencies to the [server], as at the moment people would have to use `pip install moto[server,all]` to get the full functionality.

FYI @spulec - this might need a new release, considering it broke 1.3.15

Edit: Considering it was a breaking change in the API, we could make this a major version change to reflect this? (2.0.0)

Edit 2: Marking it as draft, because the SNS/SQS services also need additional dependencies - will need to do more testing to verify other services

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -99,25 +99,27 @@ all_extra_deps = [
     _dep_sshpubkeys_py2,
     _dep_sshpubkeys_py3,
 ]
+all_server_deps = all_extra_deps + ['flask']
 
 # TODO: do we want to add ALL services here?
 # i.e. even those without extra dependencies.
 # Would be good for future-compatibility, I guess.
 extras_per_service = {
-    "ec2": [_dep_cryptography, _dep_sshpubkeys_py2, _dep_sshpubkeys_py3],
     'acm': [_dep_cryptography],
-    'iam': [_dep_cryptography],
-    'cloudformation': [_dep_PyYAML, _dep_cfn_lint],
-    'cognitoidp': [_dep_python_jose, _dep_python_jose_ecdsa_pin],
     'awslambda': [_dep_docker],
     'batch': [_dep_docker],
+    'cloudformation': [_dep_PyYAML, _dep_cfn_lint],
+    'cognitoidp': [_dep_python_jose, _dep_python_jose_ecdsa_pin],
+    "ec2": [_dep_cryptography, _dep_sshpubkeys_py2, _dep_sshpubkeys_py3],
+    'iam': [_dep_cryptography],
     'iotdata': [_dep_jsondiff],
+    's3': [_dep_cryptography],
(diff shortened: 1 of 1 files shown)


jamesgunnfiveai:
It's up to you folks as I have very limited experience with the library, but bumping to 2.0.0 for this fix may cause more pain as projects that have pinned the major version (mine included) will then get stuck on a broken release (1.3.15) requiring manual intervention to get CI etc. working again. It may be better to fix as 1.3.16 _and_ release again as 2.0.0.

bblommers:
True - ideally we revert this completely, release 1.3.16, and then merge the change and release as 2.0

spulec:
I have merged this, [reverted the impact](https://github.com/spulec/moto/commit/ed9668107847233ec6f695a494269825e01ba932) of the dependency changes, and have released `1.3.16`. We can undo my commit when we are ready for a larger release.

Issue #3298, opened 2020-09-09
Support Transcribe Service


jweite:

bblommers:
Initial support for Transcribe now available as of moto >= 1.3.16.dev40.

Will close this - feel free to open any new tickets for subsequent developement.

PR #3299, opened 2020-09-09
Transcribe Medical Support

jweite:
A first step toward Transcribe support, Issue #3298.

Please be aware of the technique used to advance Transcription and Vocabulary jobs through states, as a side-effect of get-XXX calls.  This is a pragmatic solution to providing multiple states in a controlled manner, but certainly different than anything else I've seen in moto, and perhaps surprising to those who use it.


moto/transcribe/__init__.py
```diff
@@ -0,0 +1,7 @@
+from __future__ import unicode_literals
+
+from .models import transcribe_backends
+
+transcribe_backend = transcribe_backends["us-east-1"]
+mock_transcribe = transcribe_backend.decorator
+mock_transcribe_deprecated = transcribe_backend.deprecated_decorator
```
bblommers, line 7: We're slowly trying to move away from the deprecated decorators - let's just stick to the normal one

moto/transcribe/models.py
```diff
@@ -0,0 +1,401 @@
...
+        self, region_name, vocabulary_name, language_code, vocabulary_file_uri,
+    ):
+        self._region_name = region_name
+        self.vocabulary_name = vocabulary_name
+        self.language_code = language_code
+        self.vocabulary_file_uri = vocabulary_file_uri
+        self.vocabulary_state = None
+        self.last_modified_time = None
+        self.failure_reason = None
+        self.download_uri = "https://s3.us-east-1.amazonaws.com/aws-transcribe-dictionary-model-us-east-1-prod/{}/medical/{}/{}/input.txt".format(
```
bblommers, line 165: Is the region always the same for this URL, or is it dependent on `self._region_name`?

moto/transcribe/models.py
```diff
@@ -0,0 +1,401 @@
...
+                message="The requested job couldn't be found. Check the job name and try your request again.",
+                template="error_json",
+            )
+
+    def delete_medical_transcription_job(self, medical_transcription_job_name):
+        try:
+            del self.medical_transcriptions[medical_transcription_job_name]
+        except KeyError:
+            raise RESTError(
+                error_type="BadRequestException",
```
bblommers, line 278: Let's define these errors as stand-alone classes in a separate exceptions.py-file.
Ideally we only define the template-argument once in an parent, transcribe-wide, error class.

bblommers (changes requested):
Hey @jweite, thanks for the PR. Apologies for the delay, but finally had some time to look at this - left a few comments.

moto/transcribe/__init__.py
jweite: Done.

moto/transcribe/models.py
jweite: Confirmed that that component of the URI truly varies by region.  Done.

moto/transcribe/models.py
jweite: Done.

bblommers (approved):
LGTM. Thanks again @jweite!

jweite:
Thanks for the review and merge Bert!

PR #3303, opened 2020-09-10
Fix:SQS Receive Message (MessageAttributes) in response

usmangani1:
Fix:SQS Receive Message (MessageAttributes) in response

https://github.com/localstack/localstack/issues/2965

### DIFF
--- a/moto/sqs/models.py
+++ b/moto/sqs/models.py
@@ -521,6 +521,14 @@ class Queue(CloudFormationModel):
             }
 
 
+def _filter_message_attributes(message, input_message_attributes):
+    filtered_message_attributes = {}
+    for key, value in message.message_attributes.items():
+        if key in input_message_attributes:
+            filtered_message_attributes[key] = value
+    message.message_attributes = filtered_message_attributes
+
+
 class SQSBackend(BaseBackend):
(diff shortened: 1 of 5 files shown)


bblommers (changes requested):
Hi @usmangani1, looks like the tests are failing: https://travis-ci.org/github/spulec/moto/builds/725852437

Can we keep the tests untouched? It should be fine to call receive_message without the new param, and the existing tests already cover that.
This just need additional tests to verify that the functionality of the new param

usmangani1:
> Hi @usmangani1, looks like the tests are failing: https://travis-ci.org/github/spulec/moto/builds/725852437
> 
> Can we keep the tests untouched? It should be fine to call receive_message without the new param, and the existing tests already cover that.
> This just need additional tests to verify that the functionality of the new param

@bblommers  modified existing tests a bit as we are asserting the attributes to be present in the result.After this change we will receive the message attributes in the response only when explicitly requested hence we had to modify the existing tests.

bblommers (approved):
Thanks @usmangani1!

bblommers:
This is now part of moto >= 1.3.16.dev47

Issue #3304, opened 2020-09-10
Application Autoscaling Only Supports a Limited Set of the Possible Resource ID Formats


jweite:
Application Autoscaling only supports _three-component slash-delimited_ resource IDs, the style used by ECS.  There are a number of other valid variations to this format, documented in [RegisterScalableTarget](https://docs.aws.amazon.com/autoscaling/application/APIReference/API_RegisterScalableTarget.html).

My specific attempts to register a SageMaker resource ID in the form documented [here](https://docs.aws.amazon.com/sagemaker/latest/dg/endpoint-auto-scaling-add-policy.html) fail with "Unsupported service namespace, resource type or scalable dimension."

This happens on current master.

PR #3305, opened 2020-09-10
ApplicationAutoscaling: support for all the current various forms of resource_id

jweite:
Re Issue #3304.

### DIFF
--- a/moto/applicationautoscaling/models.py
+++ b/moto/applicationautoscaling/models.py
@@ -135,7 +135,17 @@ def _target_params_are_valid(namespace, r_id, dimension):
         try:
             valid_dimensions = [d.value for d in ScalableDimensionValueSet]
             d_namespace, d_resource_type, scaling_property = dimension.split(":")
-            resource_type, cluster, service = r_id.split("/")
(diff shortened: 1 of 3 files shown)


moto/applicationautoscaling/models.py
```diff
@@ -135,7 +135,17 @@ def _target_params_are_valid(namespace, r_id, dimension):
         try:
             valid_dimensions = [d.value for d in ScalableDimensionValueSet]
             d_namespace, d_resource_type, scaling_property = dimension.split(":")
-            resource_type, cluster, service = r_id.split("/")
+            if r_id.startswith("arn:aws:comprehend"):
```
bblommers, line 138: Can we extract this into a separate method, and add some comments to make it clear what r_id should look like?

bblommers (changes requested):
Thanks for the PR @jweite - just have a single comment.

moto/applicationautoscaling/models.py
jweite: Done.  I think this build is OK - it's the recurring batch issue.  It built fine on my Travis.

bblommers (approved):
LGTM

Issue #3307, opened 2020-09-11
boto3 1.14.59 breaks s3 test_presigned_url_restrict_parameters


jweite:
s3 generate_presigned_url exception message for unknown parameters has changed (new param added).

bblommers:
This is now fixed as of moto >= 1.3.16.dev5

PR #3308, opened 2020-09-11
Change to test_s3 method test_presigned_url_restrict_parameters to to…

jweite:
…lerate change in exception messages, spurred by boto3 1.14.59 release.

Re Issue #3307


bblommers (approved):
Thanks @jweite! That's a much better way to assert the response

Issue #3312, opened 2020-09-12
S3 DeleteObject operation doesn't return x-amz-* headers


lsutic:
The S3 DeleteObject operation should (when applicable) return `x-amz-*` headers. In particular `x-amz-version-id` and `x-amz-delete-marker`: https://docs.aws.amazon.com/AmazonS3/latest/API/API_DeleteObject.html

Moto doesn't.

See: https://github.com/spulec/moto/pull/3313 for fix

bblommers:
Thanks again for raising this, and providing a PR, @lsutic. 

This should now be part of moto >= 1.3.16.dev6

PR #3313, opened 2020-09-12
Include response headers when deleting objects.

lsutic:
This PR fixes issue https://github.com/spulec/moto/issues/3312


bblommers (approved):
LGTM. Thanks @lsutic!

PR #3314, opened 2020-09-13
List dependencies for services - add integration test

bblommers:
This PR aims to fix the dependency debacle better known as #3290

-  Adds additional dependencies for specific services
-  Adds cryptography to the base set of dependencies, as it's required for IAM - which is required for many other services
-  Reduces the number of inter-service dependencies
-  Changes the README to have the `Installation`-header at the top - easily visible for people that this has changed

- Includes a test script to verify whether the correct dependencies are listed for each service. 
Use `make int_test` to run the script. 

Info on what this actually does is at the top of the script itself.

I made some minor changes in the test setup to accommodate this integration/dependency test:
 - Tests to verify CF creation are now moved to separate files called `test_service_cloudformation.py`
 - Tests to verify whether multiple services integrate are now moved to a separate file called `test_integration.py`.

There are no actual changes to the unit tests.


spulec:
Thanks for all of the work on this! The code looks good.

Any thoughts on when we should be running `int_test`. I can understand the hesitation to make it part of Travis. Maybe we make that part of the release procedure?

bblommers:
That was my thinking exactly, @spulec. Not sure at which step in the release process, so I kept it at an optional manual step for now.

Issue #1007, opened 2017-07-13
CORS support for standalone server


ludw:
I'm using moto in standalone server mode to develop a javascript web app that does uploads to S3. I have run into some issues with CORS headers. 

When attempting to upload to a bucket I get the following error from Chrome:
`XMLHttpRequest cannot load http://localhost:5000/mybucket/2193d9e1-7fff-43e3…cessKeyId=&Signature=OBKOJKaEv%2BjWj7F7kREog%2FGJH6A%3D&Expires=1499941268. Response to preflight request doesn't pass access control check: No 'Access-Control-Allow-Origin' header is present on the requested resource. Origin 'http://127.0.0.1:4200' is therefore not allowed access.`

I can see two ways to solve this:

1. Update the standalone server to make it possible to set CORS headers. This can be done rather easily with flask-cors, but the CORS headers will apply globally for all resources. I have attempted this approach on a fork of this repo and it seems to work.

2. Use CORS config set on the buckets in the standalone server to set the headers only on the buckets that have them configured. I haven't looked into how complicated this would be to implement but it feels like it would be a lot more work.

Approach 1 would add a new dependency of flask-cors and configuration for CORS headers must also be put somewhere. Command line arguments to the stand alone server is one option, but there are quite a few options for flask-cors that could be useful to expose (http://flask-cors.corydolphin.com/en/latest/api.html). 

Approach 2 feels like implementing too much of S3s logic for a mock library, but maybe it is a cleaner solution in the end. 

Moto version: 1.0.1

JackDanger:
@ludw The added dependence of flask-cors wouldn't be a problem (particularly if it were only required when Moto were used in server mode). EC2 also has some CORS support so implementing only for S3 would require duplicated effort later.

I'd love to see a PR from your work on option 1. Thanks for exploring this!

ludw:
Just an update on this, I have sadly not had any time to work on this, and no longer have the setup that caused this to begin with.

Do you think we should close the issue or is it better to leave it open?

bblommers:
This is now part of moto >= 1.3.16.dev7

PR #3316, opened 2020-09-15
Enable CORS from everywhere using flask-cors.

lsutic:
Fixes issue https://github.com/spulec/moto/issues/1007

This is required for browser-based uploads to a standalone Moto S3 server during front-end development.


bblommers (approved):
LGTM. Thanks @lsutic!

Issue #3315, opened 2020-09-15
cognito-idp Filter doesn't strip strings when filtering values


GuyMoses:
The following code works when using boto3 (but not in moto):

`client.list_users(UserPoolId=user_pool_id, Filter=f'email = "{email}"')["Users"]`

This is because moto's list_users implementation doesn't strip the `Filter` value.

Moto's implementation was merged here:
[https://github.com/spulec/moto/pull/2727/files](url)

The fix could be by striping name & value at line 295 @ (`cognitoidp/respones.py`)
something like:
```
    def list_users(self):
        user_pool_id = self._get_param("UserPoolId")
        limit = self._get_param("Limit")
        token = self._get_param("PaginationToken")
        filt = self._get_param("Filter")
        users, token = cognitoidp_backends[self.region].list_users(
            user_pool_id, limit=limit, pagination_token=token
        )
        if filt:
            name, value = filt.replace('"', "").split("=")
            users = [
                user
                for user in users
                for attribute in user.attributes
                if attribute["Name"] == name.strip() and attribute["Value"] == value.strip()
            ]
        response = {"Users": [user.to_json(extended=True) for user in users]}
        if token:
            response["PaginationToken"] = str(token)
        return json.dumps(response)
```

PR: [https://github.com/spulec/moto/pull/3317](url)

bblommers:
This is now available as of moto >= 1.3.16.dev9

PR #3317, opened 2020-09-15
[fix] cognito-idp list_users Filter arg now support spaces

GuyMoses:
Fixes issue #3315


bblommers (approved):
LGTM - thanks for the PR @GuyMoses!

Issue #3224, opened 2020-08-10
S3 - Copying Objects that have StorageClass = 'GLACIER' but have been restored


ruthbovell:
LocalStack Version - 0.11.2 (which uses Moto Version - 1.3.15.12)

I am having a problem with the mocked S3 copying a source object that has the StorageClass GLACIER but has also been restored. The fake file to be copied has this information returned when headObject is called:

```
{
     "Restore":"ongoing-request=\"false\", expiry-date=\"Tue, 11 Aug 2020 09:30:37 GMT\"",
     "LastModified":"2020-08-10T09:30:37.000Z",
     "ContentLength":28,"ETag":"\"93e5f39b77d1ef26ccd1dc1f3ac1c3fe\"",
     "ContentType":"binary/octet-stream",
     "Metadata":{},
     "StorageClass":"GLACIER"
}
```

In this case, because the source object has been restored (has value "ongoing-request=false" for the Restore property) it can be copied even though its StorageClass is GLACIER. However, the copyObject method is throwing this error:
```
Error [ObjectNotInActiveTierError]: Unexpected message handler failure: The source object of the COPY operation is not in the active tier and is only stored in Amazon Glacier.
```
It may be that a change is needed here https://github.com/spulec/moto/blob/master/moto/s3/responses.py#L1254. Instead of only checking if StorageClass equals GLACIER and throwing `ObjectNotInActiveTierError`, head the object and check whether the Restore property has "ongoing-request=\"false\" as its value when the StorageClass is GLACIER as this would be valid for copying.

bblommers:
Thanks for raising this @ruthbovell - will mark it as an enhancement

bblommers:
Little delayed, but now that #3318 has been merged, this should now be fixed as of moto >= 1.3.16.dev12

PR #3318, opened 2020-09-15
Issue 3224 s3 copy glacier object

ruthbovell:
Issue #3224 (enhancement) - S3 - Copying Objects that have StorageClass = 'GLACIER' but have been restored 

Fix for allowing restored Glacier and Deep Archive objects to be copied without throwing error.


bblommers (approved):
LGTM. Thanks @ruthbovell!

PR #3329, opened 2020-09-20
feature added: support for api RolePermissionsBoundary

macnev2013:
This PR adds support for `PutRolePermissionsBoundary` and `DeleteRolePermissionsBoundary`.
Fixed the response template `GET_ROLE_TEMPLATE` for RolePermissionBoundary support.
Added test case for `PutRolePermissionsBoundary` and `DeleteRolePermissionsBoundary`.

This PR should fix the issue in localstack: [localstack/localstack#2888](https://github.com/localstack/localstack/issues/2888)

### DIFF
--- a/moto/iam/models.py
+++ b/moto/iam/models.py
@@ -1292,6 +1292,23 @@ class IAMBackend(BaseBackend):
         role.max_session_duration = max_session_duration
         return role
 
+    def put_role_permissions_boundary(self, role_name, permissions_boundary):
+        if permissions_boundary and not self.policy_arn_regex.match(
+            permissions_boundary
+        ):
+            raise RESTError(
+                "InvalidParameterValue",
+                "Value ({}) for parameter PermissionsBoundary is invalid.".format(
+                    permissions_boundary
+                ),
+            )
+        role = self.get_role(role_name)
+        role.permissions_boundary = permissions_boundary
+    
+    def delete_role_permissions_boundary(self, role_name):
+        role = self.get_role(role_name)
+        role.permissions_boundary = None
+
     def detach_role_policy(self, policy_arn, role_name):
         arns = dict((p.arn, p) for p in self.managed_policies.values())
         try:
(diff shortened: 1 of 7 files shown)


macnev2013:
Hey @bblommers . Can you help me understand why tests are failing?
I haven't made any changes in the test cases which are failing.

It would be great if you can have a quick look at it.
Thanks.

moto/iam/models.py
```diff
@@ -1292,6 +1292,23 @@ def update_role(self, role_name, role_description, max_session_duration):
         role.max_session_duration = max_session_duration
         return role
 
+    def put_role_permissions_boundary(self, role_name, permissions_boundary):
+        if permissions_boundary and not self.policy_arn_regex.match(
+            permissions_boundary
+        ):
+            raise RESTError(
```
bblommers, line 1299: Hi @macnev2013, can you add a unit test to verify that this error is thrown?

bblommers (changes requested):
Hi @macnev2013, just have a single comment.
Don't worry about that test - it's an intermittent failure, and a second run usually succeeds.

moto/iam/models.py
macnev2013: Hey @bblommers . Thanks for the update.

Looks like we already have a test cases for it.
[https://github.com/spulec/moto/blob/master/tests/test_iam/test_iam.py#L2380-L2386](https://github.com/spulec/moto/blob/master/tests/test_iam/test_iam.py#L2380-L2386)

bblommers: That's for the `create_role`-method though. It's the same logic, but we should still test the negative scenario against `put_role_permissions_boundary`

nevilmacwan1: Hey @bblommers . I have added test case for `put_role_permissions_boundary`.
Please let me know if anything else is required.

macnev2013: Hey @bblommers. I have added test case for put_role_permissions_boundary.
Please let me know if anything else is required.

bblommers (approved):
LGTM. Thanks @macnev2013!

PR #3331, opened 2020-09-22
Make IoT certificate ID generation deterministic and prevent duplicate certificates from being created

### LINKED ISSUE #3320: IoT Core register_certificate_without_ca behavior doesn't match boto3
**moto version:** 1.3.6
**boto3 version:** 1.14.60
Moto is installed via pipenv and used with pytest.

When using boto3, trying to use the [register_certificate_without_ca](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/iot.html#IoT.Client.register_certificate_without_ca) endpoint, trying to register a certificate that already exists will throw an exception.

```
An error occurred (ResourceAlreadyExistsException) when calling the RegisterCertificateWithoutCA operation: The certificate is already provisioned or registered
```

In moto, trying to register the same certificate twice will not result in this exception. I believe this exception is important for robustness testing.

Please let me know what you think, I'd be more than happy to submit a match for this before the end of the week if you agree.

### LINKED ISSUE #3321: Moto generates incorrect certificate id for IoT Core
**moto version:** 1.3.6
**boto3 version:** 1.14.60
Moto is installed via pipenv and used with pytest.

This is where moto generates the IoT certificate id:
https://github.com/spulec/moto/blob/master/moto/iot/models.py#L133

AWS actually generates this certificate ID by getting the SHA256 of the certificate in DER format. I've found this via https://stackoverflow.com/questions/55847788/how-does-aws-iot-generate-a-certificate-id and confirmed with an AWS support ticket. 

Part of our app relies on being able to deterministically generate the certificate ID, which currently is not possible to test with moto.

Please let me know what you think, I'd be more than happy to submit a match for this before the end of the week if you agree.

BenD10:
Fixes #3320

When using boto3, trying to register a certificate that already
exists will throw a ResourceAlreadyExistsException. Moto should
follow the same pattern to allow testing error handling code in
this area.

---

Fixes #3321

As per https://stackoverflow.com/questions/55847788/how-does-aws-iot-generate-a-certificate-id,
the IoT certificate ID is the SHA256 fingerprint of the
certificate. Since moto doesn't generate full certificates we will
instead use the SHA256 hash of the passed certificate pem.

### DIFF
--- a/moto/iot/exceptions.py
+++ b/moto/iot/exceptions.py
@@ -52,3 +52,11 @@ class DeleteConflictException(IoTClientError):
     def __init__(self, msg):
         self.code = 409
         super(DeleteConflictException, self).__init__("DeleteConflictException", msg)
+
+
+class ResourceAlreadyExistsException(IoTClientError):
+    def __init__(self, msg):
+        self.code = 409
+        super(ResourceAlreadyExistsException, self).__init__(
+            "ResourceAlreadyExistsException", msg or "The resource already exists."
+        )
--- a/moto/iot/models.py
+++ b/moto/iot/models.py
@@ -19,6 +19,7 @@ from .exceptions import (
     InvalidRequestException,
     InvalidStateTransitionException,
     VersionConflictException,
+    ResourceAlreadyExistsException,
 )
 from moto.utilities.utils import random_string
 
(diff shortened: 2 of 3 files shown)


bblommers (approved):
LGTM. Thanks @BenD10!

PR #3337, opened 2020-09-25
WIP: Introducing VPC Flow Logs

ljakimczuk:
Hi! 

I'm new to `moto` :) I would like to add support for the VPC Flow Logs. I'm open to and suggestions, let me know if the approach I took fits the rest of the library.

### DIFF
--- a/moto/ec2/models.py
+++ b/moto/ec2/models.py
@@ -3662,22 +3662,21 @@ class FlowLogsBackend(object):
                             "LogDestination: {0} does not exist.".format(arn.resource),
                             )
                         )
-
-            flow_logs = FlowLogs(
-                self,
-                flow_log_id,
-                resource_id,
-                traffic_type,
-                log_destination_type,
-                log_destination,
-                log_format,
-                log_group_name,
-                deliver_logs_permission_arn,
-                max_aggregation_interval,
-            )
-
-            self.flow_logs[flow_log_id] = flow_logs
-            flow_logs_set.append(flow_logs)
+                else:
+                    flow_logs = FlowLogs(
+                        self,
+                        flow_log_id,
+                        resource_id,
+                        traffic_type,
+                        log_destination_type,
+                        log_destination,
+                        log_format,
+                        log_group_name,
+                        deliver_logs_permission_arn,
+                        max_aggregation_interval,
+                    )
+                    self.flow_logs[flow_log_id] = flow_logs
+                    flow_logs_set.append(flow_logs)
 
         return flow_logs_set, unsuccessful
 
--- a/moto/ec2/models.py
+++ b/moto/ec2/models.py
@@ -29,11 +29,6 @@ from moto.core.utils import (
 )
 from moto.core import ACCOUNT_ID
 
-from moto.logs.models import logs_backends
-from moto.logs.exceptions import ResourceNotFoundException
-from moto.s3.models import s3_backend
-from moto.s3.exceptions import MissingBucket
-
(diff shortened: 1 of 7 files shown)


moto/ec2/models.py
```diff
@@ -3511,6 +3524,295 @@ def modify_subnet_attribute(self, subnet_id, attr_name, attr_value):
...
+        log_format,
+        max_aggregation_interval,
+    ):
+        # Guess it's best to put it here due to possible
+        # lack of them in the CloudFormation template
+        max_aggregation_interval = (
+            "600" if max_aggregation_interval is None else max_aggregation_interval
+        )
+        log_destination_type = (
+            "cloud-watch-logs" if log_destination_type is None else log_destination_type
```
bblommers, line 3702: Can we have a test where LogDestinationType is explicitly set to `cloud-watch-logs`? At the moment there are only tests for s3, and tests where the param is not provided (None)

tests/test_ec2/test_flow_logs.py
```diff
@@ -0,0 +1,621 @@
+from __future__ import unicode_literals
+
+# Ensure 'assert_raises' context manager support for Python 2.6
```
bblommers, line 3: Moto does not support Python 2.6, so I don't think we need this

tests/test_ec2/test_flow_logs.py
```diff
@@ -0,0 +1,621 @@
...
+import tests.backport_assert_raises  # noqa
+from nose.tools import assert_raises
+
+import boto3
+
+# boto3.set_stream_logger(name='botocore')
```
bblommers, line 9: Not necessary

moto/ec2/models.py
```diff
@@ -3511,6 +3524,295 @@ def modify_subnet_attribute(self, subnet_id, attr_name, attr_value):
...
+                self.get_vpc(resource_id)
+            elif resource_type == "Subnet":
+                # Validate Subnets exist
+                self.get_subnet(resource_id)
+            elif resource_type == "NetworkInterface":
+                # Validate NetworkInterfaces exist
+                self.get_network_interface(resource_id)
+
+            if log_destination_type == "s3":
+                arn = log_destination.split(":", 5)[5]
```
bblommers, line 3737: It hasn't been released yet, but we're introducing a dependency model where the user would only have to install the dependencies for the specific service he's interested in. 
(See the top of the Readme for installation instructions on what that would look like.)

The new `moto.logs` and `moto.s3` imports at the top would mean that everyone who uses EC2, would also have to install dependencies for Logs and S3.
Can we move the import statements inside the if/elif statement, so the import only happens when we absolutely need it? 
That way, only users who use S3 as a destination type would have to install dependencies for S3.

bblommers (changes requested):
Hi @ljakimczuk, thanks for your PR, and welcome to Moto! I left a few comments, but  the overall approach looks very good.

moto/ec2/models.py
ljakimczuk: Sure, I'll add this

moto/ec2/models.py
ljakimczuk: Haven't thought that, it sounds smart, let me change that and thanks for catching that! :)

bblommers (approved):
LGTM. Thanks again, @ljakimczuk!

ljakimczuk:
@bblommers thank you, glad you liked that!

PR #3338, opened 2020-09-26
Fix:EC2 Tags in create vpc and create subnet

usmangani1:
Fix:EC2 Tags in create vpc and create subnet
https://github.com/localstack/localstack/issues/3031


bblommers (approved):
LGTM - thanks for the PR!

bblommers:
This is now part of moto >= 1.3.16.dev36

PR #3340, opened 2020-09-27
feat: update events remove_targets API response

giovtorres:
This PR updates the response for the Events `RemoveTargets` API call.

https://docs.aws.amazon.com/eventbridge/latest/APIReference/API_RemoveTargets.html

The response will now return a JSON formatted object including the `FailedEntryCount` and `FailedEntries` as expected from the AWS API.

### DIFF
--- a/moto/events/models.py
+++ b/moto/events/models.py
@@ -368,9 +368,12 @@ class EventsBackend(BaseBackend):
 
         if rule:
             rule.remove_targets(ids)
-            return True
-
-        return False
+            return {"FailedEntries": [], "FailedEntryCount": 0}
+        else:
+            raise JsonRESTError(
+                "ResourceNotFoundException",
+                "An entity that you specified does not exist",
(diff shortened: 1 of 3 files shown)


moto/events/models.py
```diff
@@ -368,9 +368,12 @@ def remove_targets(self, name, ids):
 
         if rule:
             rule.remove_targets(ids)
-            return True
-
-        return False
+            return {"FailedEntries": [], "FailedEntryCount": 0}
+        else:
+            raise JsonRESTError(
```
bblommers, line 373: Would you mind adding a unit test for this error scenario?

bblommers (changes requested):

bblommers (approved):
LGTM. Thanks for the PR, @giovtorres!

bblommers:
This is now part of moto >= 1.3.16.dev38

PR #3342, opened 2020-09-28
Update IMPLEMENTATION_COVERAGE.md

Ellynas:
ec2 `run_instances()` seems to be implemented [here](https://github.com/spulec/moto/blob/3bc18455a2dad70148dc095435a55b2a43eaeac1/moto/ec2/responses/instances.py#L48) and should be marked as such


bblommers (approved):
LGTM - thanks for raising/fixing this @Ellynas!

PR #3345, opened 2020-09-29
Fix : SQS Create Queue with attributes

usmangani1:
Fix : SQS Create Queue with attributes
https://github.com/localstack/localstack/issues/3035


bblommers (approved):
LGTM. Thanks @usmangani1!

PR #3346, opened 2020-09-29
Fix:SQS md5 calculation for custom string data type.

usmangani1:
Fix:SQS md5 calculation for custom string data type.

https://github.com/localstack/localstack/issues/2976

### DIFF
--- a/moto/sqs/models.py
+++ b/moto/sqs/models.py
@@ -43,7 +43,7 @@ DEFAULT_SENDER_ID = "AIDAIT2UOQQY3AUEKVGXU"
 
 MAXIMUM_MESSAGE_LENGTH = 262144  # 256 KiB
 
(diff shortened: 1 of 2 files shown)


bblommers (approved):
LGTM

tkent (commented):
These changes will not fix the issue. They reflect a fundamental misunderstanding of how custom data types for message attributes work. Custom data types for SQS message attributes follow a naming convention of {STANDARD_TYPE}.{CUSTOM_PATTERN_BASED_VALUE}.

The value you added, String.custom is just one possibility for String types and what I happened to use as an example in localstack/localstack#2976

Please see the AWS documentation on the subject.

https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-message-metadata.html#sqs-message-attributes

PR #3347, opened 2020-09-30
Add `set_default_policy_version` to the IAM backend

ljakimczuk:
Playing with the Ansible's `iam_managed_policy` I noticed it uses the `set_default_policy_version` which is currently not supported in `moto`. Would you be interested in this PR?


bblommers (approved):
PR's are always welcome, @ljakimczuk. LGTM - thanks!

PR #3348, opened 2020-09-30
Fix issue with wrong parameter signed to secret_binary during secret rotation in Secrets Manager

erka:
There is an issue with secret rotation. During the call secret_binary has got the value of the description and the description has got the tags value.


bblommers (approved):
LGTM. Thanks @erka!

bblommers:
This is now part of moto >= 1.3.16.dev48

PR #3349, opened 2020-09-30
[SecretsManager] Handle missing secrets versions

jbmchuck:
The get_secret_value method should raise ResourceNotFoundException
if a secret exists but the provided VersionId does not.

### DIFF
--- a/moto/secretsmanager/models.py
+++ b/moto/secretsmanager/models.py
@@ -187,7 +187,7 @@ class SecretsManagerBackend(BaseBackend):
         secret_binary=None,
         description=None,
         tags=[],
-        **kwargs,
+        **kwargs
     ):
 
         # error if secret exists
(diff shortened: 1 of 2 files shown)


moto/secretsmanager/models.py
```diff
@@ -179,7 +187,7 @@ def create_secret(
         secret_binary=None,
         description=None,
         tags=[],
-        **kwargs
+        **kwargs,
```
bblommers, line 190: Looks like the build is failing because of this comma: https://travis-ci.org/github/spulec/moto/builds/731710916

bblommers (changes requested):

moto/secretsmanager/models.py
jbmchuck: Thanks - fixed.

bblommers (approved):
LGTM. Thanks @jbmchuck!

bblommers:
This is now part of moto >= 1.3.16.dev48

PR #3354, opened 2020-10-02
do not depend on deprecated nose

pgajdos:
Second attempt, still humble.


pgajdos:
I think I am out of ideas with these two failures, could you please advise @spulec?

pgajdos:
(Actually, three failures and four for 2.7, respectively.)

mcepl:
Probably replaced by https://github.com/spulec/moto/pull/3361.

pgajdos:
@mcepl No problem. Thanks you care of it!

PR #3355, opened 2020-10-02
fix(dynamodb2): Fix update_item nested insert

sarangj:
Started working on a testcase to repro this issue when I encountered it in one 
of my codebases, but then the fix seemed pretty simple so figured I'd throw this
up here!

When comparing old and new values when doing a nested item update, the
`!=` implementation fails when the value being compared is `None`. This
results in an exception when trying to insert a new item into a nested
map. So just do a quick check that the original value is exists before
doing the comparison, as the `None` default is what is tripping this.

### DIFF
--- a/moto/dynamodb2/responses.py
+++ b/moto/dynamodb2/responses.py
@@ -826,7 +826,7 @@ class DynamoHandler(BaseResponse):
                         original.get(key, None), changed[key]
                     )
                     for key in changed.keys()
-                    if changed[key] != original.get(key, None)
+                    if key not in original or changed[key] != original[key]
                 }
             elif type(changed) in (set, list):
                 if len(changed) != len(original):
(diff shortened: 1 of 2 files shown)


sarangj:
hmmm, looks like I broke a different thing, didn't notice the test failure when I ran tests locally, taking a look at that

moto/dynamodb2/responses.py
```diff
@@ -826,7 +826,7 @@ def _build_updated_new_attributes(self, original, changed):
                         original.get(key, None), changed[key]
                     )
                     for key in changed.keys()
-                    if changed[key] != original.get(key, None)
+                    if key not in original or changed[key] != original[key]
```
sarangj, line 829: here, would we also need to guard against the case where the attribute exists but was originally an explicit `None`, eg, check
```
if original.get(key, None) is not None ...
```

bblommers (approved):
LGTM. Thanks @sarangj!

bblommers:
This is now part of moto >= 1.3.16.dev50

PR #3356, opened 2020-10-02
Updates to FlowLogs and IMPLEMENTATION_COVERAGE

ljakimczuk:
Hello! Small update here, what do you think about it? I guess the `Unsuccessful` class may look ugly since it's not used anywhere outside the `FlowLogs`, so maybe it's a good idea to remove it.


bblommers (approved):
Yup, that looks like a sensible simplification. Thanks @ljakimczuk!

PR #3367, opened 2020-10-08
`iot:DeleteThingGroup` should return success for non-existent groups

### LINKED ISSUE #3026: IoT Core delete_thing_group does not behave like real boto3 for nonexistent group
### moto version: 
1.3.15

### module: 
`mock_iot`

### action: 
`delete_thing_group`

### expected behavior:

boto3's response is the same whether or not the thing group existed or not. It does not throw any exception if requested to delete a nonexistent thing group.

### actual behavior:

moto throws a `ResourceNotFoundException` if the thing group does not exist: 

    botocore.errorfactory.ResourceNotFoundException: An error occurred (ResourceNotFoundException) when calling the DeleteThingGroup operation: The specified resource does not exist

bpandola:
Closes #3026

### DIFF
--- a/moto/iot/models.py
+++ b/moto/iot/models.py
@@ -924,8 +924,12 @@ class IoTBackend(BaseBackend):
                 + thing_group_name
                 + " when there are still child groups attached to it"
             )
-        thing_group = self.describe_thing_group(thing_group_name)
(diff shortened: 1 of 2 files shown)


bblommers (approved):
LGTM. Thanks again @bpandola!

PR #3370, opened 2020-10-09
Fixes acm:describe_certificate for imported certificates having no SAN extension

### LINKED ISSUE #3371: [BUG] acm client: describe_certificate() does not work, unittest is broken as well.
While describing imported acm certificate,  if the certificate does not have any SANs, [get_extension_for_oid()](https://github.com/spulec/moto/blob/d00cefa25cfd338caaabcefc7a79e6e803f577cf/moto/acm/models.py#L296)  could throw a [cryptography.x509.ExtensionNotFound](https://cryptography.io/en/latest/x509/reference/#cryptography.x509.Extensions.get_extension_for_oid ).  

This is clearly not being handled.

The [unit test](https://github.com/spulec/moto/blob/d00cefa25cfd338caaabcefc7a79e6e803f577cf/tests/test_acm/test_acm.py#L129) should have failed here.  But unfortunately, it is overwritten by next function with same name. 

#3370 resolves this.

nom3ad:
https://github.com/spulec/moto/blob/c4b630e20fb9004ef4b1eab345ceab5ef9e4c2ff/moto/acm/models.py#L230-L235

While describing imported acm certificate, if the certificate does not have any SANs, `get_extension_for_oid()` could throw a `cryptography.x509.ExtensionNotFound.`

The [unit test](https://github.com/spulec/moto/blob/d00cefa25cfd338caaabcefc7a79e6e803f577cf/tests/test_acm/test_acm.py#L129) should have failed here. But unfortunately, it is overwritten by next function with same name.

This PR fixes `describe_certificate()` to handle imported certificates with missing SAN extension. unit test is updated to run the corresponding tests.

Fixes #3371

### DIFF
--- a/moto/acm/models.py
+++ b/moto/acm/models.py
@@ -293,9 +293,12 @@ class CertBundle(BaseModel):
             key_algo = "EC_prime256v1"
 
         # Look for SANs
-        san_obj = self._cert.extensions.get_extension_for_oid(
-            cryptography.x509.OID_SUBJECT_ALTERNATIVE_NAME
-        )
+        try:
+            san_obj = self._cert.extensions.get_extension_for_oid(
+                cryptography.x509.OID_SUBJECT_ALTERNATIVE_NAME
+            )
+        except cryptography.x509.ExtensionNotFound:
+            san_obj = None
         sans = []
         if san_obj is not None:
             sans = [item.value for item in san_obj.value]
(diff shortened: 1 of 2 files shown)


bblommers (approved):
LGTM. Nice find with the unit tests! Thanks @nom3ad

PR #3373, opened 2020-10-09
support Tags parameter in ACM import_certificate() and request_certificate() methods

### LINKED ISSUE #3377: ACM : import_certificate() and request_certificate() are not respecting Tags parameter
Currently moto does not respect  `Tags` parameter for boto3 ACM client methods: [import_certificate()](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/acm.html#ACM.Client.import_certificate) and [request_certificate()](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/acm.html#ACM.Client.request_certificate).   It skips tags silently. 

To reproduce.  Add following unit test to  [test_acm.py](https://github.com/nom3ad/moto/blob/743dd46399295063981284101981c0be0f69fe02/tests/test_acm/test_acm.py)
```python
@mock_acm
def test_import_certificate_with_tags():
    client = boto3.client("acm", region_name="eu-central-1")

    resp = client.import_certificate(
        Certificate=SERVER_CRT,
        PrivateKey=SERVER_KEY,
        CertificateChain=CA_CRT,
        Tags=[{"Key": "Environment", "Value": "QA"}],
    )
    arn = resp["CertificateArn"]
    resp = client.list_tags_for_certificate(CertificateArn=arn)
    tags = {item["Key"]: item.get("Value", "__NONE__") for item in resp["Tags"]}
    tags.should.contain("Environment")
```

nom3ad:
Implements `Tags` parameter which is acceptable in boto3 ACM client methods: [import_certificate()](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/acm.html#ACM.Client.import_certificate) and [request_certificate()](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/acm.html#ACM.Client.request_certificate) APIs

Fixes #3377

### DIFF
--- a/tests/test_acm/test_acm.py
+++ b/tests/test_acm/test_acm.py
@@ -38,12 +38,24 @@ def test_import_certificate():
     client = boto3.client("acm", region_name="eu-central-1")
 
     resp = client.import_certificate(
-        Certificate=SERVER_CRT, PrivateKey=SERVER_KEY, CertificateChain=CA_CRT
+        Certificate=SERVER_CRT, PrivateKey=SERVER_KEY, CertificateChain=CA_CRT,
+        Tags=[ 
+            { 'Key': 'Environment', 'Value': 'QA' }, 
+            { 'Key': 'KeyOnly' }, 
+        ]
     )
-    resp = client.get_certificate(CertificateArn=resp["CertificateArn"])
+    arn = resp["CertificateArn"]
 
+    resp = client.get_certificate(CertificateArn=arn)
     resp["Certificate"].should.equal(SERVER_CRT.decode())
     resp.should.contain("CertificateChain")
+
+    resp = client.list_tags_for_certificate(CertificateArn=arn)
+    tags = {item["Key"]: item.get("Value", "__NONE__") for item in resp["Tags"]}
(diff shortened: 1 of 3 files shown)


tests/test_acm/test_acm.py
```diff
@@ -38,13 +38,24 @@ def test_import_certificate():
     client = boto3.client("acm", region_name="eu-central-1")
 
     resp = client.import_certificate(
-        Certificate=SERVER_CRT, PrivateKey=SERVER_KEY, CertificateChain=CA_CRT
+        Certificate=SERVER_CRT,
```
bblommers, line 41: Would you mind creating a new test with the tags? Would be good to keep the existing test as is, to verify that it still works without passing in tags

bblommers (changes requested):

tests/test_acm/test_acm.py
nom3ad: Sounds right.  :+1:   
I have added few more new tests and better tag handling.

bblommers (approved):
LGTM. Thanks for the PR @nom3ad!

PR #3376, opened 2020-10-10
Implement additional filters for `ecs.ListTasks`

### LINKED ISSUE #1785: ECS list tasks should handle filters like 'desiredStatus', 'family', 'service_name'
Currently when calling the `list_task()` method on ECS with the `desiredStatus` set to 'STOPPED' all the tasks are returned. I would expect to only get the stopped tasks. 

When looking at the code probably a filter on status should be added to this method: https://github.com/spulec/moto/blob/master/moto/ecs/models.py#L612

bpandola:
Closes #1785

### DIFF
--- a/moto/ecs/models.py
+++ b/moto/ecs/models.py
@@ -940,10 +940,30 @@ class EC2ContainerServiceBackend(BaseBackend):
                 )
             )
 
+        if family:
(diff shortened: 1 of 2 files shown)


bblommers (approved):
LGTM. Thanks @bpandola!

PR #3380, opened 2020-10-11
issue-3379 iam list_roles: implement PathPrefix, MaxItems and Marker

waynemetcalfe:
I also noticed the regex validation is not performed on PathPrefix but figured that was a wider issue and does not belong in this PR


bblommers (approved):
LGTM. Thanks @waynemetcalfe!

bblommers:
This is now part of moto >= 1.3.16.dev62

PR #3383, opened 2020-10-13
Add KMS Support to EBS Encrypted Volumes

### LINKED ISSUE #3248: EBS volumes don't store KmsKeyId
```python
from moto import mock_ec2
import boto3

mock = mock_ec2()
mock.start()

resource = boto3.resource("ec2")

kwargs = {
    "AvailabilityZone": "ap-southeast-2a",
    "Size": 8,
    "Encrypted": True,
    "KmsKeyId": "key"    
}

volume = resource.create_volume(**kwargs)

print(volume.kms_key_id)
```

The above prints `None`. I was expecting it to print `key`.

moto 1.3.14

bpandola:
NOTE: A client-provided `KmsKeyId` is simply stored as-is, and is *not* validated
against the KMS backend.  This is in keeping with other moto backends (RDS, Redshift)
that currently also accept unvalidated customer master key (CMK) parameters, but could
be an area for future improvement.

Closes #3248

### DIFF
--- a/tests/test_ec2/test_elastic_block_store.py
+++ b/tests/test_ec2/test_elastic_block_store.py
@@ -13,6 +13,7 @@ import sure  # noqa
 
 from moto import mock_ec2_deprecated, mock_ec2
 from moto.ec2.models import OWNER_ID
+from moto.kms import mock_kms
 
 
 @mock_ec2_deprecated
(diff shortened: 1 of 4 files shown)


tests/test_ec2/test_elastic_block_store.py
```diff
@@ -915,3 +916,53 @@ def test_search_for_many_snapshots():
...
+        )
+    ex.exception.response["Error"]["Code"].should.equal("InvalidParameterDependency")
+    ex.exception.response["Error"]["Message"].should.contain("KmsKeyId")
+
+
+@mock_kms
+@mock_ec2
+def test_create_encrypted_volume_without_kms_key_should_use_default_key():
+    resource = boto3.resource("ec2", region_name="us-east-1")
+    volume = resource.create_volume(
```
bblommers, line 936: Can we add a sanity check here, that `kms.describe_key()` fails before the volume is created?
Just to make the behaviour explicit, that it's only `created automatically when you first create an encrypted resource in an AWS service.`

bblommers (changes requested):
Hey @bpandola -  left two minor comments.

tests/test_ec2/test_elastic_block_store.py
```diff
@@ -915,3 +916,53 @@ def test_search_for_many_snapshots():
...
+    volume.kms_key_id.should.equal(default_ebs_key_arn)
+
+
+@mock_ec2
+def test_create_volume_with_kms_key():
+    resource = boto3.resource("ec2", region_name="us-east-1")
+    volume = resource.create_volume(
+        AvailabilityZone="us-east-1a", Encrypted=True, KmsKeyId="key", Size=10
+    )
+    volume.kms_key_id.should.equal("key")
```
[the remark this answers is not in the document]
bpandola: Good catch! Not only was this not covered in a test, but it's apparently never worked properly!  I fixed this by coercing the Python bool to an AWS string bool in response templates.

tests/test_ec2/test_elastic_block_store.py
bpandola: Good suggestion.  I also added a check that subsequent requests use the same default key (just to make sure the create once logic works).

bpandola:
Closing/Reopening to retry flaky test...

bblommers (approved):
LGTM - thanks!

RiyaGoel12:
Hi Team,

I have tested the fix on my local. It works fine for 'create_volume' API but not working for 'run_instances' api. Please include this feature for the volumes created using in 'run_instances'.
Also, please let me know when can we expect this feature to be released.

Regards,
Riya Goel

bpandola:
@RiyaGoel12 Open a new issue with a minimum failing test case and I'll take a look.  Thanks.

PR #3389, opened 2020-10-15
Add Support for SNS Topic `KmsMasterKeyId` Attribute

### LINKED ISSUE #3216: SNS KmsMasterKeyId not returned as part of get_topic_attributes response
when sns mock is used to get topic attributes:

`response = client.get_topic_attributes(
    TopicArn='string'
)`

KmsMasterKeyId - The ID of an AWS-managed customer master key (CMK) for Amazon SNS or a custom CMK - is not included as part of the response.

bpandola:
We do not do any validation of the `KmsMasterKeyId` attribute, and simply
store it as-as.  This mimics the behavior in AWS, where the key is not
validated until it is actually used (when publishing[1]).

[1]: https://docs.aws.amazon.com/sns/latest/dg/sns-server-side-encryption.html

Closes #3216

### DIFF
--- a/moto/sns/models.py
+++ b/moto/sns/models.py
@@ -45,6 +45,7 @@ class Topic(CloudFormationModel):
         self.account_id = DEFAULT_ACCOUNT_ID
         self.display_name = ""
         self.delivery_policy = ""
+        self.kms_master_key_id = ""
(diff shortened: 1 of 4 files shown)


bblommers (approved):
LGTM. Thanks @bpandola!

PR #3392, opened 2020-10-15
FIX:Add secrets Manager Tag resource Funtionality

usmangani1:
FIX:Add secrets Manager Tag resoruce Funtionality

https://github.com/localstack/localstack/issues/3102

### DIFF
--- a/moto/secretsmanager/models.py
+++ b/moto/secretsmanager/models.py
@@ -565,13 +565,13 @@ class SecretsManagerBackend(BaseBackend):
 
         return secret.arn, secret.name
 
-    def update_secret_tags(self, secret_id, tags):
+    def tag_resource(self, secret_id, tags):
 
         if secret_id not in self.secrets.keys():
             raise SecretNotFoundException()
 
         secret = self.secrets[secret_id]
-        old_tags = secret["tags"]
+        old_tags = secret.tags
 
         for tag in tags:
             old_tags.append(tag)
--- a/moto/secretsmanager/models.py
+++ b/moto/secretsmanager/models.py
@@ -565,6 +565,19 @@ class SecretsManagerBackend(BaseBackend):
 
         return secret.arn, secret.name
 
+    def update_secret_tags(self, secret_id, tags):
+
+        if secret_id not in self.secrets.keys():
+            raise SecretNotFoundException()
+
+        secret = self.secrets[secret_id]
+        old_tags = secret["tags"]
+
+        for tag in tags:
+            old_tags.append(tag)
+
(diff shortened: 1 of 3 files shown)


moto/secretsmanager/models.py
```diff
@@ -565,6 +565,19 @@ def restore_secret(self, secret_id):
 
         return secret.arn, secret.name
 
+    def update_secret_tags(self, secret_id, tags):
```
bblommers, line 568: Let's keep the naming consistency here as in `responses.py`, so call this method `tag_resource` as well. 
This is done by all methods, to make it clear it implements that specific SDK call.

moto/secretsmanager/models.py
```diff
@@ -565,6 +565,19 @@ def restore_secret(self, secret_id):
...
+
+        if secret_id not in self.secrets.keys():
+            raise SecretNotFoundException()
+
+        secret = self.secrets[secret_id]
+        old_tags = secret["tags"]
```
bblommers, line 574: This is an instance of `FakeSecret`, so you can access the tags by attribute (secret.tags) 
(Tests are currently failing because of this)

moto/secretsmanager/models.py
```diff
@@ -565,6 +565,19 @@ def restore_secret(self, secret_id):
 
         return secret.arn, secret.name
 
+    def update_secret_tags(self, secret_id, tags):
+
+        if secret_id not in self.secrets.keys():
+            raise SecretNotFoundException()
```
bblommers, line 571: Can we get a test for this error scenario

tests/test_secretsmanager/test_secretsmanager.py
```diff
@@ -912,3 +912,21 @@ def test_update_secret_marked_as_deleted_after_restoring():
...
+        Tags=[
+            {"Key": "FirstTag", "Value": "SomeValue"},
+            {"Key": "SecondTag", "Value": "AnotherValue"},
+        ],
+    )
+    secrets = conn.list_secrets()
+    assert secrets["SecretList"][0].get("Tags") == [
+        {"Key": "FirstTag", "Value": "SomeValue"},
+        {"Key": "SecondTag", "Value": "AnotherValue"},
+    ]
```
bblommers, line 932: Can we call tag_resource again here with different tags, to assert that we're extending tags, not overwriting the existing tag-set

bblommers (changes requested):

bblommers (approved):
LGTM. Thanks @usmangani1!

PR #3393, opened 2020-10-16
Including Message attributes when ALL is passed

usmangani1:
Including Message attributes when ALL is passed


bblommers (changes requested):
Hi @usmangani1, can  you add a test for this scenario? Cheers

bblommers (approved):
Thanks @usmangani1!

bblommers:
This is now part of moto >= 1.3.16.dev70

PR #3396, opened 2020-10-20
Add missing `Fn::GetAtt` attributes to S3 bucket mock

10xjs:
Prevent valid CloudFormation templates containing S3 `Fn::GetAtt` from raising unexpected `ValidationError`.

Spec for valid cases referenced from https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-s3-bucket.html#aws-properties-s3-bucket-return-values.

Addresses an issue reported here: https://github.com/localstack/aws-cdk-local/issues/1

### DIFF
--- a/moto/s3/models.py
+++ b/moto/s3/models.py
@@ -523,10 +523,7 @@ class LifecycleAndFilter(BaseModel):
 
         for key, value in self.tags.items():
             data.append(
-                {
-                    "type": "LifecycleTagPredicate",
-                    "tag": {"key": key, "value": value},
-                }
+                {"type": "LifecycleTagPredicate", "tag": {"key": key, "value": value},}
             )
 
         return data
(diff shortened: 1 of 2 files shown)


bblommers (changes requested):
Hi @10xjs, thanks for your PR! Looks like the linting fails: https://travis-ci.org/github/spulec/moto/jobs/737265323

10xjs:
Looks like I forgot to format the changed files with `black` 🤦‍♂️.

10xjs:
@bblommers I'm struggling to get the CI suite to pass. I notice that `black` failed on https://github.com/spulec/moto/pull/3396/commits/af459607b94ca3dabd4a2c08265ddc98f14bab39 failed in Python 3.7 only, however, reformating under a fresh install on 3.7 also fails.

bblommers:
Hey @10xjs - Looks like Travis failed to start Docker for some reason. Those tests now do pass. 
(Different tests now fail, because of a breaking change in an unrelated module.. :facepalm:  Will merge this PR after I get that fixed.)

bblommers (approved):

bblommers:
Thanks again for the PR @10xjs! This is now part of moto >= 1.3.16.dev77

PR #3409, opened 2020-10-23
Fix: Adding alarm arn to describe alarms response

usmangani1:
Fix: Adding alarm arn to describe alarms response
https://github.com/localstack/localstack/issues/3130


tests/test_cloudwatch/test_cloudwatch.py
```diff
@@ -51,6 +53,7 @@ def test_create_alarm():
     list(alarm.ok_actions).should.equal(["arn:ok"])
     list(alarm.insufficient_data_actions).should.equal(["arn:insufficient"])
     alarm.unit.should.equal("Seconds")
+    alarm.alarm_arn.should.equal(make_arn_for_alarm("us-east-1", ACCOUNT_ID, "tester"))
```
bpandola, line 56: You might not want to couple this check so tightly to the underlying implementation.  If a future version of `moto` supports, say, a user-definable account id, this test will break.  You really just want to make sure the alarm arn is in the response.  You could go a bit further and check that it starts with `arn` and contains the `tester` resource id.  That would be a perfectly adequate test without importing an implementation detail (`make_arn_for_alarm`).

tests/test_cloudwatch/test_cloudwatch_boto3.py
```diff
@@ -139,6 +141,9 @@ def test_describe_alarms_for_metric():
     )
     alarms = conn.describe_alarms_for_metric(MetricName="cpu", Namespace="blah")
     alarms.get("MetricAlarms").should.have.length_of(1)
+    alarms.get("MetricAlarms")[0].get("AlarmArn").should.equal(
+        make_arn_for_alarm("eu-central-1", ACCOUNT_ID, "testalarm1")
+    )
```
bpandola, line 146: Same here...

moto/cloudwatch/models.py
```diff
@@ -77,8 +77,10 @@ def __init__(
         insufficient_data_actions,
         unit,
         actions_enabled,
+        region="us-east-1",
     ):
         self.name = name
+        self.alarm_arn = make_arn_for_alarm(region, DEFAULT_ACCOUNT_ID, name)
```
bpandola, line 83: Minor nit: why not just call this property `arn`?  `FakeAlarm.alarm_arn` seems redundant, especially considering the other properties (e.g. `name` instead of `alarm_name`).

tests/test_cloudwatch/test_cloudwatch.py
bblommers: Agreed

bblommers (changes requested):
Hey @usmangani1, thanks for your PR! See @bpandola comment

bblommers (approved):
LGTM - thanks @usmangani1!

bblommers:
This is now part of moto >= 1.3.16.dev87

Issue #3416, opened 2020-10-28
Add support for SageMaker Notebook Lifecycle Config


jweite:

bblommers:
This is now part of moto >= 1.3.16.dev84

PR #3417, opened 2020-10-28
Notebook Lifecycle Config create, describe and delete

jweite:
Re Issue #3416.

### DIFF
--- a/moto/sagemaker/models.py
+++ b/moto/sagemaker/models.py
@@ -8,7 +8,7 @@ from datetime import datetime
 from moto.core import ACCOUNT_ID, BaseBackend, BaseModel
 from moto.core.exceptions import RESTError
 from moto.sagemaker import validators
-from .exceptions import MissingModel
+from .exceptions import MissingModel, ValidationError
 
 
 class BaseObject(BaseModel):
@@ -285,11 +285,7 @@ class FakeEndpointConfig(BaseObject):
             message = "Value '{}' at 'instanceType' failed to satisfy constraint: Member must satisfy enum value set: {}".format(
                 instance_type, VALID_INSTANCE_TYPES
             )
-            raise RESTError(
-                error_type="ValidationException",
-                message=message,
-                template="error_json",
-            )
+            raise ValidationError(message=message)
 
     @property
     def response_object(self):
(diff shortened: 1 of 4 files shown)


moto/sagemaker/responses.py
```diff
@@ -239,3 +239,38 @@ def delete_training_job(self):
...
+                ),
+                on_create=self._get_param("OnCreate"),
+                on_start=self._get_param("OnStart"),
+            )
+            response = {
+                "NotebookInstanceLifecycleConfigArn": lifecycle_configuration.notebook_instance_lifecycle_config_arn,
+            }
+            return 200, {}, json.dumps(response)
+        except AWSError as err:
+            return err.response()
```
bblommers, line 258: Is the try-except necessary? I don't see any errors being raised in models.py

moto/sagemaker/models.py
```diff
@@ -677,6 +718,58 @@ def get_notebook_instance_tags(self, arn):
...
+            del self.notebook_instance_lifecycle_configurations[
+                notebook_instance_lifecycle_config_name
+            ]
+        except KeyError:
+            message = "Unable to delete Notebook Instance Lifecycle Config '{}'. (Details: Notebook Instance Lifecycle Config does not exist.".format(
+                FakeSageMakerNotebookInstanceLifecycleConfig.arn_formatter(
+                    notebook_instance_lifecycle_config_name, self.region_name
+                )
+            )
+            raise RESTError(
```
bblommers, line 767: Let's extract both RESTerror into a ValidationError in exceptions.py, and

moto/sagemaker/models.py
```diff
@@ -677,6 +718,58 @@ def get_notebook_instance_tags(self, arn):
...
+
+    def delete_notebook_instance_lifecycle_config(
+        self, notebook_instance_lifecycle_config_name
+    ):
+        try:
+            del self.notebook_instance_lifecycle_configurations[
+                notebook_instance_lifecycle_config_name
+            ]
+        except KeyError:
+            message = "Unable to delete Notebook Instance Lifecycle Config '{}'. (Details: Notebook Instance Lifecycle Config does not exist.".format(
```
bblommers, line 762: Can we get a unit test for this scenario?

bblommers (changes requested):

moto/sagemaker/responses.py
jweite: Ah, copy-paste from another operation.  But on further thought, there is an important exception I need to be throwing on create: already exists.  So I'm leaving the try/except in place here and adding code to detect duplicate name and raise in the backend method.

moto/sagemaker/models.py
jweite: Done, and changed all existing raises of RESTError in this module to use it.

moto/sagemaker/models.py
jweite: Done.

bblommers (approved):
LGTM. Thanks @jweite!

PR #3420, opened 2020-10-28
Fix `dynamodb:TransactGetItems` response for items that do not exist

### LINKED ISSUE #3404: bug: dynamodb2 transact_get_items should return empty Map for item, if item not found
AWS documentation https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_TransactGetItems.html
"If a requested item could not be retrieved, the corresponding ItemResponse object is Null, or if the requested item has no projected attributes, the corresponding ItemResponse object is an empty Map."

Short example:
Put into some table only one value
    "Get": {
      "Key": {
        "primaryKey": {
          "S": "primarykey1"
        }
      },
      "TableName": "TestTable"
    }
Then transact_get_items, from both AWS and moto dynomodb2
[
  {
    "Get": {
      "Key": {
        "primaryKey": {
          "S": "primarykey1"
        }
      },
      "TableName": "TestTable"
    }
  }
  {
    "Get": {
      "Key": {
        "primaryKey": {
          "S": "primarykey2"
        }
      },
      "TableName": "TestTable"
    }
  }
]

AWS response: [
  {
    "Get": {
      "Key": {
        "primaryKey": {
          "S": "primarykey1"
        }
      },
      "TableName": "TestTable"
    }
  },
  {}
]

Moto dynamodb2 response
[
  {
    "Get": {
      "Key": {
        "primaryKey": {
          "S": "primarykey1"
        }
      },
      "TableName": "TestTable"
    }
  }    # HERE THE SECOND EMPTY ITEM IS MISSING
]

bpandola:
If the requested item has no projected attributes, the corresponding ItemResponse
object is an empty Map.[1]  Verified against real AWS.

Fix existing general test case and add an explicit test case to cover this scenario.

[1]: https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_TransactGetItems.html#API_TransactGetItems_ResponseElements

Fixes #3404

### DIFF
--- a/moto/dynamodb2/responses.py
+++ b/moto/dynamodb2/responses.py
@@ -906,6 +906,7 @@ class DynamoHandler(BaseResponse):
                 return self.error(er, "Requested resource not found")
 
             if not item:
+                responses.append({})
                 continue
 
             item_describe = item.describe_attrs(False)
--- a/tests/test_dynamodb2/test_dynamodb.py
+++ b/tests/test_dynamodb2/test_dynamodb.py
@@ -4285,7 +4285,8 @@ def test_valid_transact_get_items():
         ]
     )
     res["Responses"][0]["Item"].should.equal({"id": {"S": "1"}, "sort_key": {"S": "1"}})
-    len(res["Responses"]).should.equal(1)
+    len(res["Responses"]).should.equal(2)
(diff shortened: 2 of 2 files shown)


bblommers (approved):
LGTM. Thanks @bpandola!

PR #3423, opened 2020-10-28
Fix update for application autoscaling register target

erinnachen:
Also adds suspended_state as an updatable attribute for ApplicationAutoScaling::ScalableTarget


bblommers (approved):
LGTM. Thanks @erinnachen!

PR #3427, opened 2020-10-30
Stepfunctions improvements

### LINKED ISSUE #3137: stepfunction list_executions does not support nextToken
When calling the `list_excutions` method of boto, it is meant to return a paginated response of executions up to the `maxResults` you have specified (with a ceiling of 1000).  It does not seem to be doing this.

```
import boto3
from moto import mock_stepfunctions


@mock_stepfunctions
def test_next_token():
    boto3.setup_default_session()
    sfn_client = boto3.client("stepfunctions", region_name="us-east-1")
    mock_sfn = sfn_client.create_state_machine(
        name="my-sfn", definition="my-definition", roleArn="arn:aws:iam::000000000000:role/my-sfn",
    )
    arn = mock_sfn["stateMachineArn"]
    
    for _ in range(100):
        sfn_client.start_execution(stateMachineArn=arn)

    response = sfn_client.list_executions(stateMachineArn=arn, statusFilter="RUNNING", maxResults=10)

    # The response should only contain 10 executions
    assert len(response["executions"]) == 10
    # And it should contain a nextToken so we can get the next 10, etc.
    assert "nextToken" in response
```

bpandola:
Implement filtering for stepfunctions:ListExecutions

Add pagination to Step Functions endpoints

Implements a generalized approach to pagination via a decorator method for the
Step Functions endpoints.  Modeled on the real AWS backend behavior, `nextToken`
is a dictionary of pagination information encoded in an opaque string.

With just a bit of metadata hard-coded (`utils.PAGINATION_MODEL`), backend `list`
methods need only be decorated with `@paginate` and ensure that their returned
entities are sorted to get full pagination support without any duplicated code
polluting the model.

Closes #3137

### DIFF
--- a/moto/stepfunctions/exceptions.py
+++ b/moto/stepfunctions/exceptions.py
@@ -46,3 +46,11 @@ class InvalidExecutionInput(AWSError):
 class StateMachineDoesNotExist(AWSError):
     TYPE = "StateMachineDoesNotExist"
     STATUS = 400
+
+
+class InvalidToken(AWSError):
+    TYPE = "InvalidToken"
+    STATUS = 400
+
+    def __init__(self, message="Invalid token"):
+        super(InvalidToken, self).__init__("Invalid Token: {}".format(message))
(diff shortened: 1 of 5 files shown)


bblommers (approved):
This is awesome, @bpandola - thanks!
Looks like the decorator can be easily reused across other services as well, so that should be really useful.

PR #3271, opened 2020-08-30
Add kinesisvideo

toshitanian:
Added basic operations of `kinesisvideo`
- `create_stream`
- `describe_stream`
- `list_streams`
- `delete_stream`
- `get_data_endpoint`

PR #3428, opened 2020-10-30
Implementation of core AWS Media Live endpoins

jordan-dimov:
I can't seem to get the mocked response into the test. The medialive unit test `test_create_channel_succeeds` always fails, complaining that Channel is an empty dict, even though the response should contain some channel data.


moto/medialive/responses.py
```diff
@@ -0,0 +1,47 @@
...
+            input_specification=input_specification,
+            log_level=log_level,
+            name=name,
+            request_id=request_id,
+            reserved=reserved,
+            role_arn=role_arn,
+            tags=tags,
+        )
+        # TODO: adjust response
+        return json.dumps(dict(channel=channel))
```
bpandola, line 42: I didn't pull your branch to test this, but I think this just needs to be: `return json.dumps(dict(Channel=channel))`

jordan-dimov: Well that's what I thought, but when I do this it gets even stranger - I end up with NO trace of anything channel-related in the response. So I wonder what else I'm missing in the layers in-between.

bpandola: Ok, I dug into this a bit further.  Check out my commit here: https://github.com/bpandola/moto/commit/c3822f1cb8e118e7e0dec4316052a7ff2599fd15

This should get you going again.  See my commit message for more information.

jordan-dimov: @bpandola  this was super helpful. Thank you!

jordan-dimov:
@bpandola Would you be able to please review this code so far and let me know if you spot anything obviously wrong or missing? I only plan on adding a few more endpoints, in the same style. 

In particular, should I be doing more validation on the request schema, or does moto take care of that already?

bpandola:
@jordan-dimov I think you're making things harder on yourself than necessary by duplicating all that tuple/dict code.  Have a look at the implementation for AWS Kinesis Video (another `rest-json`-based AWS API) here:
https://github.com/spulec/moto/pull/3271/files
In particular, note the `Stream` class in `models.py` and the use of its `to_dict()` method.  You'll want to create a similar `Channel` class for your service.

`moto` doesn't take care of any request validation unless you explicitly code it.  Generally speaking, a "happy path" implementation of a new service is better than no implementation at all.  Again, look to the linked change-set for how to do some basic validation (e.g. raising an error when someone tries to describe a channel that doesn't exist).
https://github.com/spulec/moto/pull/3271/files#diff-cec5bb80f2e70733e7f985acb0e54add02d92af41e6994a301a9d74717acc92dR113

jordan-dimov:
@bpandola Thank you for the advice, I've now refactored. 

While working on two new endpoints (start_channel and stop_channel) I ran into an unexpected problem. While these two endpoints are very simple, and no different from already implemented ones (e.g. almost identical to describe_channel, delete_channel), my new unit tests fail. I get the following error in `moto/core/responses.py`  (in the method `_get_action_from_method_and_request_uri`): 

    ('^/prod/tags/(?P<resource-arn>[^/]*)$', 'CreateTags')
    *** sre_constants.error: bad character in group name 'resource-arn' at position 16

I think the problem is the dash (`-`) character. These URL matchers (`regexp_and_names`) are constructed automatically, so not sure how that dash ended up there. 

You can see a full stack if you try to run the MediaLive unit tests.

Any advice?

bpandola:
@jordan-dimov You'll need to add `.replace("-", "_")` here: https://github.com/spulec/moto/blob/master/moto/core/responses.py#L324

jordan-dimov:
@bpandola I have now finished implementing the core MediaLive endpoints (might do more later - but not as part of this PR). Would you please be able to review and advise on next steps to get that into a release?

moto/medialive/models.py
```diff
@@ -0,0 +1,298 @@
...
+        elif self.state == "UPDATING":
+            self.state = self._previous_state
+            self._previous_state = None
+
+
+class MediaLiveBackend(BaseBackend):
+    def __init__(self, region_name=None):
+        super(MediaLiveBackend, self).__init__()
+        self.region_name = region_name
+        self._channels = {}
```
bpandola, line 118: These should be OrderedDicts() so they work the same in Python 2/3.

moto/medialive/models.py
```diff
@@ -0,0 +1,298 @@
+from __future__ import unicode_literals
+from uuid import uuid4
+from boto3 import Session
+from moto.core import BaseBackend
+
+
+class Input:
```
bpandola, line 7: This should inherit from `moto.core.BaseModel`

moto/medialive/models.py
```diff
@@ -0,0 +1,298 @@
...
+    def _resolve_transient_states(self):
+        # Resolve transient states before second call
+        # (to simulate AWS taking its sweet time with these things)
+        if self.state in ["CREATING"]:
+            self.state = "DETACHED"  # or ATTACHED
+        elif self.state == "DELETING":
+            self.state = "DELETED"
+
+
+class Channel:
```
bpandola, line 55: Inherit from `BaseModel`

moto/medialive/urls.py
```diff
@@ -0,0 +1,14 @@
...
+url_bases = [
+    "https?://medialive.(.+).amazonaws.com",
+]
+
+
+response = MediaLiveResponse()
+
+
+url_paths = {
+    "{0}/.*$": response.dispatch,
```
bpandola, line 13: You'll need to add paths for `/prod/channels`, `/prod/inputs`, etc. for this to work in moto server mode.  See `urls.py` for other services for examples (`polly` is similar).

tests/test_medialive/test_server.py
```diff
@@ -0,0 +1,17 @@
...
+"""
+Test the different server responses
+"""
+
+
+@mock_medialive
+def test_medialive_list():
+    backend = server.create_backend_app("medialive")
+    test_client = backend.test_client()
+    # do test
```
bpandola, line 17: You should put something here.  Even just `list_channels` would be fine and check for the empty response.  (See Redshift's `test_server.py` for an example).

bpandola (commented):
Python 2 tests and the Server mode tests are failing.  I've added some comments that will get you closer to those passing, but there may be other changes needed.

tests/test_medialive/test_medialive.py
```diff
@@ -0,0 +1,332 @@
...
+import sure  # noqa
+from moto import mock_medialive
+from uuid import uuid4
+
+region = "eu-west-1"
+
+
+def _create_input_config(name, **kwargs):
+    role_arn = kwargs.get(
+        "role_arn", "arn:aws:iam::123456789012:role/TestMediaLiveInputCreateRole"
```
bblommers, line 13: Please use the global constant here (`from moto.core import ACCOUNT_ID)`

tests/test_medialive/test_medialive.py
```diff
@@ -0,0 +1,332 @@
...
+        Type=input_type,
+        Tags=tags,
+        Vpc=vpc_config,
+    )
+    return input_config
+
+
+def _create_channel_config(name, **kwargs):
+    role_arn = kwargs.get(
+        "role_arn", "arn:aws:iam::123456789012:role/TestMediaLiveChannelCreateRole"
```
bblommers, line 53: Same here

jordan-dimov:
Thank you @bpandola , @bblommers  - I've now fixed everything that was raised. Still don't understand why the server tests are failing in Travis CI - doesn't seem to have much to do with my code, though I could be wrong.

moto/medialive/urls.py
```diff
@@ -0,0 +1,15 @@
...
+
+url_bases = [
+    "https?://medialive.(.+).amazonaws.com",
+]
+
+
+response = MediaLiveResponse()
+
+
+url_paths = {
```
bpandola, line 12: You have to add `url_paths` for *all* possible request uris that you have mocked.  The ones you've added only cover a `POST` to either of those paths, but you also need `/prod/channels/<channel-id>`, `/prod/channels/<channel-id>/start`, etc. 

Example: https://github.com/spulec/moto/blob/master/moto/managedblockchain/urls.py

bpandola (changes requested):

jordan-dimov:
Quick! Before I mess up something else :)

bblommers (approved):
LGTM. Thanks for all the work, @jordan-dimov!

bblommers:
And thanks for the review and help, @bpandola :)

PR #3431, opened 2020-11-02
Fix: eventName for a deleted record should be REMOVE instead of DELETE

### LINKED ISSUE #3400: eventName for a deleted record should be REMOVE instead of DELETE
When using Moto to mock DynamoDB streams, a deleted record has the eventName = "DELETE" right now:
https://github.com/spulec/moto/blob/9eb58eea411214ded617ea45ce9a4407a8f95194/moto/dynamodb2/models/__init__.py#L243-L250

AWS's DynanoDB Stream specification states that this should actually be `"REMOVE"`:
https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_streams_Record.html#DDB-Type-streams_Record-eventName

I am on Moto version 1.3.16 and using the dynamodb2.start() and stop() methods

bpandola:
Verified API documentation[1] against the real AWS backend.

[1]: https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_streams_Record.html#DDB-Type-streams_Record-eventName

Fixes #3400

### DIFF
--- a/moto/dynamodb2/models/__init__.py
+++ b/moto/dynamodb2/models/__init__.py
@@ -245,7 +245,7 @@ class StreamShard(BaseModel):
         if old is None:
             event_name = "INSERT"
         elif new is None:
-            event_name = "DELETE"
+            event_name = "REMOVE"
         else:
             event_name = "MODIFY"
         seq = len(self.items) + self.starting_sequence_number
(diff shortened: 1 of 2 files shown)


bblommers (commented):
LGTM

PR #3432, opened 2020-11-03
Fix: describe/list attribute discrepancy in Secrets Manager

### LINKED ISSUE #3406: list secretsManager returns "VersionIdsToStages" field but the aws apis call it "SecretVersionsToStages"
The response to list_secrets from aws secretsmanager, has a field in the secret_list_entry that is misnamed. I discovered this when making a call to list_secrets via the latest localstack/localstack docker container 0.12.0 (which includes this library).

"VersionIdsToStages" should instead be "SecretVersionsToStages"

I think all that needs changing is the response model here:
https://github.com/spulec/moto/blob/14980371d796ef849b1ea401ce2c028911b537d1/moto/secretsmanager/models.py#L138

You can see the "SecretVersionsToStages" field in the "ResponseSyntax" example, in the aws api docs here:
https://docs.aws.amazon.com/secretsmanager/latest/apireference/API_ListSecrets.html

And in the java sdk here:
https://github.com/aws/aws-sdk-java/blob/83f1a7c1bef7bab8f809f5123275848678fb807b/aws-java-sdk-secretsmanager/src/main/java/com/amazonaws/services/secretsmanager/model/transform/SecretListEntryJsonUnmarshaller.java#L101

localstack is including moto.secretsmanager here:
https://github.com/localstack/localstack/blob/d43a58d563a5a912de64d4e9c8bcac28a96d3857/localstack/services/secretsmanager/secretsmanager_starter.py#L2

bpandola:
`secretsmanager:DescribeSecret` returns `VersionIdsToStages`
`secretsmanager:ListSecrets` returns the same information in `SecretVersionsToStages`

* Verified fix against real AWS backend.

Fixes #3406

### DIFF
--- a/moto/secretsmanager/models.py
+++ b/moto/secretsmanager/models.py
@@ -136,6 +136,7 @@ class FakeSecret:
             "DeletedDate": self.deleted_date,
             "Tags": self.tags,
             "VersionIdsToStages": version_id_to_stages,
+            "SecretVersionsToStages": version_id_to_stages,
         }
 
     def _form_version_ids_to_stages(self):
--- a/tests/test_secretsmanager/test_list_secrets.py
+++ b/tests/test_secretsmanager/test_list_secrets.py
@@ -43,9 +43,11 @@ def test_list_secrets():
 
     assert secrets["SecretList"][0]["ARN"] is not None
     assert secrets["SecretList"][0]["Name"] == "test-secret"
+    assert secrets["SecretList"][0]["SecretVersionsToStages"] is not None
     assert secrets["SecretList"][1]["ARN"] is not None
     assert secrets["SecretList"][1]["Name"] == "test-secret-2"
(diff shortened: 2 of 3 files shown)


moto/secretsmanager/models.py
```diff
@@ -136,6 +136,7 @@ def to_dict(self):
             "DeletedDate": self.deleted_date,
             "Tags": self.tags,
             "VersionIdsToStages": version_id_to_stages,
```
bblommers, line 138: Does this mean this field is no longer necessary?

bblommers (changes requested):

moto/secretsmanager/models.py
bpandola: They're both required.  The original one for the `describe_secret` response and the one I added for the `list_secrets` response.

bblommers: Ah, gotcha - didn't realize that this dict feeds both methods

bblommers (approved):
LGTM!

Issue #3433, opened 2020-11-03
Add support for Amazon Forecast 


pwrmiller:
I would like to add basic endpoints for Amazon Forecast, then expand to improve endpoint implementation coverage for Amazon Forecast.

bblommers:
Initial implementation of Forecast is now available as of moto >= 1.3.16.dev95!

PR #3434, opened 2020-11-03
Adds some basic endpoints for Amazon Forecast

pwrmiller:
This pull request adds basic endpoints for dataset group creation for `@mock_forecast`, along with test coverage. This is to address #3433


pwrmiller:
Hi! This is my first contribution to `moto`. I've tried to follow your contributing guidelines as best possible - please let me know if there's anything else you need on this one. 

I'd like to add additional forecast endpoint coverage moving forward, hopefully on a weekly cadence/ as time permits.

moto/forecast/models.py
```diff
@@ -0,0 +1,183 @@
...
+        self._validate()
+
+    def update(self, dataset_arns):
+        self.dataset_arns = dataset_arns
+        self.last_modified_date = iso_8601_datetime_without_milliseconds(datetime.now())
+
+    def _validate(self):
+        errors = []
+
+        self._update_validation_errors(errors, self._validate_dataset_group_name())
```
bblommers, line 69: The `_update_validation_errors` helper-method** feels unnecessary. IMO it will be more readable like this:
`errors.extend(self.get_dataset_group_name_errors())`
where `get_dataset_group_name_errors()` returns a list with 0 or more errors.

tests/test_forecast/test_forecast.py
```diff
@@ -0,0 +1,209 @@
...
+import boto3
+import sure  # noqa
+
+from parameterized import parameterized
+from botocore.exceptions import ClientError
+from nose.tools import assert_raises
+
+from moto import mock_sts, mock_forecast
+from moto.core import ACCOUNT_ID
+from moto.forecast.models import DatasetGroup
```
bblommers, line 12: Unnecessary import

tests/test_forecast/test_forecast.py
```diff
@@ -0,0 +1,209 @@
...
+    "EC2_CAPACITY",
+    "WORK_FORCE",
+    "WEB_TRAFFIC",
+    "METRICS",
+]
+
+
+@parameterized(valid_domains)
+@mock_forecast
+@mock_sts
```
bblommers, line 29: The `mock_sts` is not necessary for any of these tests.
(They were necessary for the StepFunctions example, which I guess you based this on, because SFN does some token validation that requires STS)

tests/test_forecast/test_forecast.py
```diff
@@ -0,0 +1,209 @@
...
+    )
+
+
+@mock_forecast
+@mock_sts
+def test_forecast_dataset_group_create_invalid_domain():
+    name = "example_dataset_group"
+    client = boto3.client("forecast", region_name=region)
+
+    with assert_raises(ClientError):
```
bblommers, line 46: Can we have some minimal assertions, that the correct error was thrown? Typically this would look like:
```
with assert_raises(ClientError) as ex:
    do_something()
ex.exception.response["Error"]["Message"].should.equal("This is what's wrong: ...")
```

moto/forecast/models.py
```diff
@@ -0,0 +1,183 @@
...
+from moto.core.utils import iso_8601_datetime_without_milliseconds
+from .exceptions import (
+    InvalidInputException,
+    ResourceAlreadyExistsException,
+    ResourceNotFoundException,
+    ValidationException,
+)
+
+
+class Domain(Enum):
```
bblommers, line 19: The Enum doesn't seem to be in use anymore?

bblommers (changes requested):
Thanks for your PR @pwrmiller! I left a few comments

moto/forecast/models.py
pwrmiller: Thanks for the tip - I've refactored to accommodate your suggestion.

tests/test_forecast/test_forecast.py
pwrmiller: Agreed - refactored, removed.

tests/test_forecast/test_forecast.py
pwrmiller: Thanks - I wasn't sure if it was required or not; I've removed it and run tests again without issue.

tests/test_forecast/test_forecast.py
pwrmiller: Of course. I've added assertions to all exception checking tests based on this suggestion.

moto/forecast/models.py
pwrmiller: Yes. I had it in here originally, but it was proving to be less readable to use the enum. I'll remove it and all references to it.

pwrmiller:
The failed test was unrelated to the changes made here (`test_batch.test_submit_job ... FAIL`), and only in one environment. Have you seen flaky tests in that location before?

How shall we proceed?

bblommers:
@pwrmiller Yup, that's a known issue. I've restarted the build, and it's OK now. Thanks again!

bblommers (approved):

PR #3436, opened 2020-11-04
ECR: Fix "imageDigest" value in ecr.list_images() response

lithammer:


bblommers (changes requested):
Thanks for the PR @lithammer! Would you mind adding a simple test/assertion for this? Happy to merge after that

lithammer:
I added some simple assertions to an existing test following how the other assertions about `imageDigest` is made. It's pretty low effort though I guess.

bblommers (approved):
LGTm! Thanks @lithammer

PR #3439, opened 2020-11-05
Tech Debt - Remove duplicate AWSError classes

bblommers:
Moves the AWSError to moto.core.exceptions, and ensures all services reference the same class.


spulec:
Thanks!

PR #3444, opened 2020-11-08
AWS Lambda - Only instantiate Docker when invoking functions

### LINKED ISSUE #3359: Breaking changes in docker
## Reporting Bugs

Hey, 
If there is no docker deamon running, then using `mock_lambda` with the newest version of docker-py (4.3.1) yields the following error:

```
docker.errors.DockerException: Error while fetching server API version: ('Connection aborted.', ConnectionRefusedError(61, 'Connection refused'))
```
Version 4.3.0 does not have this problem.

bblommers:
…ker running

Fixes #3359 

This ensures that Docker is only instantiated when trying to invoke a Lambda. Put differently, when Docker is not running, it will only throw an error when trying to invoke a Lambda-function.

Other Lambda calls (create_function, list_, etc) will still work even if Docker is not running.

Edit: `@batch` has the same problem, so this PR now fixes both services

(My initial approach was to simply downgrade the `docker-py` dependency to <=4.3.0, but that's not really a solution - just a workaround.)


spulec:
Thanks Bert!

PR #3445, opened 2020-11-08
Travis: Use Focal-distribution, so we no longer have to downgrade Docker

bblommers:
Focal is Ubuntu version 20.04, which is the latest available in Travis: https://docs.travis-ci.com/user/reference/linux

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,4 @@
-dist: bionic
+dist: focal
 language: python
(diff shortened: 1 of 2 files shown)


spulec:
I'm good with this change once we get Travis green

bblommers:
Looks like it's working now @spulec - Travis just needed a kick

Issue #3449, opened 2020-11-09
Route53: Implement GetChange


relvira:
Hello 👋🏼 

The Route53 mock doesn't implement [GetChange](https://docs.aws.amazon.com/Route53/latest/APIReference/API_GetChange.html). I'm currently working on something that would tremendously benefit from having this endpoint implemented in the mock for local dev and testing. 

I'd be happy to contribute and work on a pull request for moto to support this endpoint if you deem appropriate.

Thanks 🙇🏼!

bblommers:
Hi @relvira, PR's are always welcome!

There's a Contributing-document that should have all the info you need, but let me know if you need any help implementing this.

relvira:
Thanks for your reply @bblommers!

I've started a local branch on my machine to play with the codebase and got the GetChange endpoint working.

Looks like you're not currently handling change IDs as [rrset_response](https://github.com/spulec/moto/blob/master/moto/route53/responses.py#L97) always returns the same ID in [CHANGE_RRSET_RESPONSE](https://github.com/spulec/moto/blob/master/moto/route53/responses.py#L277).

Would you be happy with me implementing a simple method that does something similar? The idea would be to add a new url path in `route53/urls.py`

```
r"{0}/(?P<api_version>[\d_-]+)/change/(?P<change_id>[^/]+)$": Route53().get_change,
````

implement a `get_change` in `route53/responses.py` that'd be something like:

```
def get_change(self, request, full_url, headers):
  self.setup_class(request, full_url, headers)
   if request.method == "GET":
        parsed_url = urlparse(full_url)
        changeid = parsed_url.path.rstrip("/").rsplit("/", 1)[1]
        template = Template(GET_CHANGE_RESPONSE)
        return 200, headers, template.render(changeid=changeid)

GET_CHANGE_RESPONSE = """<?xml version="1.0" encoding="UTF-8"?>
<GetChangeResponse xmlns="https://route53.amazonaws.com/doc/2013-04-01/">
   <ChangeInfo>
      <Status>INSYNC</Status>
      <SubmittedAt>2010-09-10T01:36:41.958Z</SubmittedAt>
      <Id>/change/{{ changeid }}</Id>
   </ChangeInfo>
</GetChangeResponse>"""
```

`Status` and `SubmittedAt` would remain static, but not change ID. Another option would be for the Route53 to support generating and storing change ID for new operations and actually use the information from the backend. Let me know what you think.

bblommers:
[...]

PR #3453, opened 2020-11-10
Route53: Implement dummy GetChange endpoint

relvira:
## Description
This PR implements a dummy [GetChange](https://docs.aws.amazon.com/Route53/latest/APIReference/API_GetChange.html) endpoint for the Route53 mock within moto. This change has been discussed in https://github.com/spulec/moto/issues/3449. 

The Route53 mock in moto does not store Change IDs in it's backend at the moment, this is why I created a dummy endpoint that always returns `INSYNC` status together with the same `SubmittedAt` date for all requests.

Ideally, the Route53 mock backend would store Change IDs the same way it stores hostedzones, healthchecks, etc. I've discussed the possibility of doing this with @bblommers and he's keen on that, but for now I'll stick with this approach as the ideal option would require significant changes in the route53 mock to support and store those change IDs.

## Why
GetChange is used by existing tooling around Route53, which makes supporting this endpoint in moto very advantageous when using moto in local development environments as well as testing for Route53. 

## Testing
There's no testing available in this PR at the moment given this endpoint is dummy. If maintainers would like me to implement an integration test for this we can make it work.


relvira:
@bblommers would you mind taking a look at this pull request when you have some time? 

Thanks for your help! 🙇🏼

relvira:
@spulec Would you be able to take a look at my pull request? Let me know if there's anything I can do to get the ball rolling.

Thank you! 🙇🏼

bblommers (changes requested):
Hey @relvira, thanks for your PR! Would be good to have a simple unit test, just to check that the mapping is done correctly.

FYI, we've just changed our test framework to be pytest (instead of nose). You might need to merge in the latest changes from master to prevent any merge conflicts and see an appropriate example test.

relvira:
hey @bblommers - sure I can do that. Would you have any example I can reuse for the test? I'm not familiar with pytest. If not that's fine, I can figure out something.

bblommers:
@relvira You can just add a method to the existing test-suite for Route53: https://github.com/spulec/moto/blob/master/tests/test_route53/test_route53.py
Plenty of examples in there.

Just make sure that you're following an example that uses the `@mock_route53`-decorator (not `@mock_route53_deprecated`).

relvira:
Thanks for your guidance @bblommers. I've submitted a new PR https://github.com/spulec/moto/pull/3486 as I've messed up my branch for this PR due to the pytest merge. 

Would you mind heading there for a review? 🙇🏼

Issue #3441, opened 2020-11-06
DynamoDB#update_item() - AttributeUpdates should have action=PUT by default


gsamaras:
Boto provides a parameter with a default, but moto doesn't seem to follow this, which lead to an error when trying to mock DynamoDB in a Unit Test.

Quoting `boto3.DynamoDB.Client.update_item()` [documentation](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/dynamodb.html#DynamoDB.Client.update_item)(emphasis mine):

>Action (string) --Specifies how to perform the update. Valid values are PUT (**default**), ...

So in my code to-be-tested I was doing:
```
def my_upload_record_to_dynamodb(...):
    ...
    dataclass_attributes[current_field.name] = {'Value': attribute_value}
    record_key = {'record_id': {'S': record_id}}
    self._dynamodb_client.update_item(TableName=table_name, Key=record_key, AttributeUpdates=dataclass_attributes)
```

and my unit test looked like this:
```
import unittest
import boto3

from moto import mock_dynamodb2
from unittest.mock import MagicMock, patch

class TestElasticSearch(unittest.TestCase):
    def setUp(self):
        self.es_record = ...

    @mock_dynamodb2
    @patch('boto3.client')
    def test_upload_record_to_dynamodb(self, mock_boto3_sts_client):
        table_name = 'my_table'
        dynamodb = boto3.resource('dynamodb', region_name='eu-west-1')

        table = dynamodb.create_table(
            TableName=table_name,
            KeySchema=[
                {
                    'AttributeName': 'record_id',
                    'KeyType': 'HASH'
                },
            ],
            AttributeDefinitions=[
                {
                    'AttributeName': 'record_id',
                    'AttributeType': 'S'
                },

            ]
        )
        # Wait until the table exists.
        table.meta.client.get_waiter('table_exists').wait(TableName=table_name)

        my_upload_record_to_dynamodb(table_name, self.es_record)
```

which was failing with:
```
[CPython37-test:stdout]         es_reporter = ElasticSearchReporter(self.role_arn, None)
[CPython37-test:stdout]         es_reporter.my_upload_record_to_dynamodb(
[CPython37-test:stdout]             ElasticsearchIndices.ANALYSIS_BACKEND_SESSIONS_ALPHA,
[CPython37-test:stdout] >           self.es_record
[CPython37-test:stdout]         )
[CPython37-test:stdout] 
...
[...]

PR #3454, opened 2020-11-10
dynamodb2 support for default Action ('Put') in update_item

gsamaras:
Linked GitHub Issue to be fixed by this PR: https://github.com/spulec/moto/issues/3441.

Implemented the fix to use the default value for `Action`, according to `boto3.DynamoDB.Client.update_item()` [documentation](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/dynamodb.html#DynamoDB.Client.update_item), which is 'Put'.

Added a unit test to make sure the default value is used in case of absence of `Action` passed by the caller.

Tests pass:

```
...
test_dynamodb.test_update_item_with_list ... ok
test_dynamodb.test_update_item_with_no_action_passed_with_list ... ok
...
```

as well as Black check.

### DIFF
--- a/moto/dynamodb2/models/__init__.py
+++ b/moto/dynamodb2/models/__init__.py
@@ -109,7 +109,10 @@ class Item(BaseModel):
 
     def update_with_attribute_updates(self, attribute_updates):
         for attribute_name, update_action in attribute_updates.items():
-            action = update_action["Action"]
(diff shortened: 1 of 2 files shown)


bblommers (approved):
LGTM. Thanks, @gsamaras!

Issue #3455, opened 2020-11-10
Support EMR Security Configurations and Kerberos Attributes


jweite:

bblommers:
This is now supported as of moto >= 1.3.16.dev127

PR #3456, opened 2020-11-10
Added support for EMR Security Configurations and Kerberos Attributes.

jweite:
Re Issue #3455.

### DIFF
--- a/tests/test_emr/test_emr_boto3.py
+++ b/tests/test_emr/test_emr_boto3.py
@@ -107,7 +107,15 @@ def test_describe_cluster():
     args["Instances"]["EmrManagedSlaveSecurityGroup"] = "slave-security-group"
     args["Instances"]["KeepJobFlowAliveWhenNoSteps"] = False
     args["Instances"]["ServiceAccessSecurityGroup"] = "service-access-security-group"
+    args["KerberosAttributes"] = {
+        "Realm": "MY-REALM.COM",
+        "KdcAdminPassword": "SuperSecretPassword2",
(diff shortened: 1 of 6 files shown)


bblommers (approved):

tests/test_emr/test_emr_boto3.py
```diff
@@ -983,3 +993,53 @@ def test_tags():
...
+    resp["CreationDateTime"].should.be.a("datetime.datetime")
+
+    resp = client.describe_security_configuration(Name=security_configuration_name)
+    resp["Name"].should.equal(security_configuration_name)
+    resp["SecurityConfiguration"].should.equal(security_configuration)
+    resp["CreationDateTime"].should.be.a("datetime.datetime")
+
+    client.delete_security_configuration(Name=security_configuration_name)
+
+    with assert_raises(ClientError) as ex:
```
bblommers, line 1033: Would you mind merging in master?
We switched to pytest as our testing framework (instead of nose), and as a result, these two will have to change to `with pytest.assert_raises..`

bblommers (changes requested):

tests/test_emr/test_emr_boto3.py
jweite: Done.  

New RDS unit test test_create_db_instance_with_tags failed for me, presumably because (per commit comments) it requires botocore >= 1.18.17, so I updated that in requirements.   Also .gitignoring directory htmlcov, which is generated by pytest (at least in my env).

bblommers (approved):
LGTM. Thanks @jweite!

PR #3459, opened 2020-11-12
Fix: `TagList` missing in rds:DescribeDBInstance response

### LINKED ISSUE #3458: RDS: TagList missing in describe_db_instance
There doesn't seem to be any support for tags in describe_db_instance. It seems to be on purpose, I couldn't find any test dealing with tags in this api call

The AWS API returns a TagList property when the instance is tagged.

(Edit: remove references to create_db_instance, which does in fact support the Tags property)

bpandola:
Previously, tags were only available via rds:ListTagsForResource, but are now
included in the Create/DescribeDBInstance responses as of Botocore 1.18.17[1]

[1]: https://github.com/boto/botocore/commit/f29d23c53ec0477e58450744986b0f8f23c97da1#diff-d10722c0e11ded323c8d240066d7ed31e93a1e6423d54e091b7d54b86e6bd4e0

Fixes #3458

### DIFF
--- a/moto/rds2/models.py
+++ b/moto/rds2/models.py
@@ -280,6 +280,14 @@ class Database(CloudFormationModel):
                 <Port>{{ database.port }}</Port>
               </Endpoint>
               <DBInstanceArn>{{ database.db_instance_arn }}</DBInstanceArn>
+              <TagList>
+              {%- for tag in database.tags -%}
+                <Tag>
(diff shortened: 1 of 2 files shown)


bblommers (approved):

bblommers (approved):
LGTM. Thanks!

PR #3461, opened 2020-11-12
Fix:CloudWatch List metrics with dimensions

usmangani1:
Fix:CloudWatch List metrics with dimensions
https://github.com/localstack/localstack/issues/3222

### DIFF
--- a/moto/cloudwatch/models.py
+++ b/moto/cloudwatch/models.py
@@ -155,6 +155,17 @@ class FakeAlarm(BaseModel):
         self.state_updated_timestamp = datetime.utcnow()
 
 
+def are_dimensions_same(metric_dimensions, dimensions):
+    for dimension in metric_dimensions:
+        is_dimension_found = False
+        for new_dimension in dimensions:
+            if dimension.name == new_dimension.get('Name') and  dimension.value==new_dimension.get('Value'):
+                is_dimension_found = True
+        if not is_dimension_found:
+            return False
+    return True
+
+
 class MetricDatum(BaseModel):
     def __init__(self, namespace, name, value, dimensions, timestamp):
         self.namespace = namespace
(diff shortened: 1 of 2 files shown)


moto/cloudwatch/models.py
```diff
@@ -354,6 +372,16 @@ def delete_alarms(self, alarm_names):
         for alarm_name in alarm_names:
             self.alarms.pop(alarm_name, None)
 
+    # def add_metric(self, namespace, metric_data, new_metric_info):
```
bblommers, line 375: Unnecessary

bblommers (changes requested):
So, I did some testing, and it looks like the `are_dimension_same`-method doesn't quite capture the right logic.
In AWS itself, the metric data is overwritten when the dimensions are exactly the same, but metric data is added if the dimension is only partially different.

Consider this test case:
```
 conn.put_metric_data(
    ..., 
        Dimensions=[{"Name": "Name", "Value": "B"}],
)

result = conn.list_metrics(
        Namespace="tester", Dimensions=[{"Name": "Name", "Value": "B"}]
    )["Metrics"]
result.should.equal([
{'Namespace': 'tester',
  'MetricName': 'metric',
  'Dimensions': [{'Name': 'Name', 'Value': 'B'}]}
])

conn.put_metric_data(
    ...
        Dimensions=[{"Name": "Name", "Value": "B"}, {"Name": "Name2", "Value": "C"}],
)

result = conn.list_metrics(
    Namespace="tester", Dimensions=[{"Name": "Name", "Value": "B"}]
)["Metrics"]
result.should.equal(
[{'Namespace': 'tester',
  'MetricName': 'metric',
  'Dimensions': [{'Name': 'Name', 'Value': 'B'}]},
 {'Namespace': 'tester',
  'MetricName': 'metric',
  'Dimensions': [{'Name': 'Name', 'Value': 'B'}, {'Name': 'Name2', 'Value': 'C'}]}]
)
```

This passes against AWS itself, but fails against moto currently.

bblommers (approved):
LGTM

bblommers:
This is now part of moto >= 1.3.16.dev134

PR #3462, opened 2020-11-12
Fix:RDS:add DBParameterGroupArn in describe-db-parameter-groups & cre…

usmangani1:
Fix:RDS:add DBParameterGroupArn in describe-db-parameter-groups & create-db-parameter-group
https://github.com/localstack/localstack/issues/3219

### DIFF
--- a/moto/rds2/models.py
+++ b/moto/rds2/models.py
@@ -9,7 +9,8 @@ from boto3 import Session
 from jinja2 import Template
 from re import compile as re_compile
 from moto.compat import OrderedDict
-from moto.core import BaseBackend, BaseModel, CloudFormationModel
+from moto.core import BaseBackend, BaseModel, CloudFormationModel, ACCOUNT_ID
+
 from moto.core.utils import iso_8601_datetime_with_milliseconds
 from moto.ec2.models import ec2_backends
 from .exceptions import (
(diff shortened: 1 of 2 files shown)


moto/rds2/models.py
```diff
@@ -1463,20 +1463,26 @@ def to_xml(self):
         return template.render(option_group_option_setting=self)
 
 
+def make_rds_arn(region, name):
+    return "arn:aws:rds:{0}:REDACTED:pg:{1}".format(region, name)
```
bpandola, line 1467: I think you should use `moto.core.ACCOUNT_ID` here.  It seems the original reporter of the localstack issue was just trying to avoid exposing his own account id and not actually suggesting that the Arn should be redacted.

tests/test_rds2/test_rds2.py
```diff
@@ -1518,6 +1520,9 @@ def test_create_db_parameter_group():
     db_parameter_group["DBParameterGroup"]["Description"].should.equal(
         "test parameter group"
     )
+    db_parameter_group["DBParameterGroup"]["DBParameterGroupArn"].should.equal(
+        "arn:aws:rds:{0}:REDACTED:pg:{1}".format(region, pg_name)
```
bblommers, line 1524: The moto.core.ACCOUNT_ID has to be replaced here as well (same in the other test)

bblommers (changes requested):

bblommers (approved):
Thanks @usmangani1!

bblommers:
This is now part of moto >= 1.3.16.dev154

Issue #3448, opened 2020-11-09
DynamoDB update_item using ReturnValues='UPDATED_NEW' fails with "Value must be a nonempty dictionary whose key is a valid dynamodb type"


pstDnb:
The following code runs fine towards DynamoDB, but fails with moto:
```
import time

import boto3
import moto


@moto.mock_dynamodb2()
def test_update_does_not_work():
    dynamo_client = boto3.resource("dynamodb", region_name="eu-west-1")
    dynamo_client.create_table(TableName="ZzTable3",
                               KeySchema=[{'AttributeName': 'HashKey1', 'KeyType': 'HASH'}],
                               AttributeDefinitions=[{'AttributeName': 'HashKey1', 'AttributeType': 'S'}],
                               ProvisionedThroughput={'ReadCapacityUnits': 5, 'WriteCapacityUnits': 5})

    time.sleep(5)

    table_data = [{
        "HashKey1": "HashKeyValue1",
        "listValuedAttribute1": [
            "a",
            "b"
        ]}]

    dynamodb_table = dynamo_client.Table("ZzTable3")
    for record in table_data:
        dynamodb_table.put_item(Item=record)

    table_data[0]["listValuedAttribute1"][1] = "c"

    dynamodb_table.update_item(
        Key={"HashKey1": "HashKeyValue1"},
        UpdateExpression=f"SET listValuedAttribute1=:a",
        ExpressionAttributeValues={':a': table_data[0]["listValuedAttribute1"]},
        ReturnValues='UPDATED_NEW')
```
The following error is shown in the output:
```
self = <boto3.dynamodb.types.TypeDeserializer object at 0x000002D3CD64DB70>
value = None

    def deserialize(self, value):
        """The method to deserialize the DynamoDB data types.
    
        :param value: A DynamoDB value to be deserialized to a pythonic value.
            Here are the various conversions:
    
            DynamoDB                                Python
            --------                                ------
            {'NULL': True}                          None
            {'BOOL': True/False}                    True/False
            {'N': str(value)}                       Decimal(str(value))
            {'S': string}                           string
            {'B': bytes}                            Binary(bytes)
            {'NS': [str(value)]}                    set([Decimal(str(value))])
            {'SS': [string]}                        set([string])
            {'BS': [bytes]}                         set([bytes])
            {'L': list}                             list
            {'M': dict}                             dict
    
        :returns: The pythonic value of the DynamoDB type.
        """
    
        if not value:
[...]

Issue #3460, opened 2020-11-12
Does not work with latest `responses`


rossmacarthur:
Moto has a hack where it sets the default responses mock's `_find_match` method to a custom one. The latest changes in `responses` have changed the return type of this method.

https://github.com/spulec/moto/blob/47dbad291e717059a93f5cf795a9be08cfc872cc/moto/core/models.py#L248-L267

See the [`responses`](https://github.com/getsentry/responses) change here https://github.com/getsentry/responses/commit/cf3c5a685ae95bee1a714fd36cfb77d448a4b5ec

garribas:
We are also affected by this bug and had to force installation of previous version of `responses` (0.12.0), with 0.12.1 it fails as described in the bug report:
https://github.com/spulec/moto/blob/master/setup.py#L43

bblommers:
This is now fixed as of moto >= 1.3.16.dev122

Jlrine2:
This doesn't seem to work with 1.3.16.dev122, Though I was able to get it working by using on of the 1.3.17.devX installs (specifically dev240)

PR #3463, opened 2020-11-12
Fixed issue 3448 for DynamoDB update_item

oidebrett:
#### Reference Issue: #3448
DynamoDB update_item using ReturnValues='UPDATED_NEW' fails 

#### What does this implementation fix/address?
When updating DynamoDB item using UpdateExpressions with ReturnValues='UPDATED_NEW' and where the update is a list with indexes where the updated value equals the original value then an error is shown "TypeError: Value must be a nonempty dictionary whose key is a valid dynamodb type"

This fixes a bug in the build_updated_new_attributes() definition. This method should always return the changed value, even if they are the same.

The change was simply to return changed as opposed to returning None.


moto/dynamodb2/responses.py
```diff
@@ -841,7 +840,7 @@ def _build_updated_new_attributes(self, original, changed):
             elif changed != original:
```
bblommers, line 840: I think we can get rid of this `elif` now

tests/test_dynamodb2/test_dynamodb.py
```diff
@@ -2412,6 +2412,44 @@ def update(col, to, rv):
...
+    r = update('a', ['a', 'c'], "UPDATED_NEW")
+    assert r["Attributes"] == {'listValuedAttribute1': ['a', 'c']}
+
+    r = update('a', {'a', 'c'}, "UPDATED_NEW")
+    assert r["Attributes"] == {'listValuedAttribute1': {'a', 'c'}}
+
+    r = update('a', {1,2}, "UPDATED_NEW")
+    assert r["Attributes"] == {'listValuedAttribute1': {1, 2}}
+
+    with assert_raises(ClientError) as ex:
```
bblommers, line 2450: We've just changed to pytest instead of nose for our testing framework, so this will have to change to `with pytest.raises(..)`.
If you merge in the latest changes to master, you should the same change in the other tests

bblommers (changes requested):

oidebrett:
I updated with the changes requested. Hope thats fine now. I am new to using the code format checked black and when running it for my test case I get the message:
black --check moto/ tests/test_dynamodb2/test_dynamodb.py 
would reformat /home/ec2-user/environment/projects/moto/tests/test_dynamodb2/test_dynamodb.py
Oh no!   
1 file would be reformatted, 428 files would be left unchanged.

I wasnt sure what I needed to do with this message.

bblommers:
Yup, looks good now @oidebrett - thanks!

If you run `black moto tests` (without the check-argument) it will reformat the files for you, just need to commit them afterwards.

oidebrett:
I am getting errors in the travis-ci jobs that have nothing to do with these changes. I am not sure if I have done something wrong. If I have I am happy to make the changes on this PR.

bblommers:
Those errors were related to #3460, nothing to do with the changes on this PR.

LGTM - thanks @oidebrett!

bblommers (approved):

PR #3466, opened 2020-11-13
Fix failures with latest responses library (0.12.1)

### LINKED ISSUE #3460: Does not work with latest `responses`

garribas:
Fixes #3460 

Before:
```
$ pip freeze | grep responses
responses==0.12.1

$make test
...
Ran 3601 tests in 705.949s

FAILED (SKIP=1, errors=28, failures=6)
```

After:
```
$ pip freeze | grep responses
responses==0.12.1

$make test
...
Ran 3601 tests in 591.731s

OK (SKIP=1)
```

### DIFF
--- a/moto/core/models.py
+++ b/moto/core/models.py
@@ -30,7 +30,7 @@ from .utils import (
 )
 
 ACCOUNT_ID = os.environ.get("MOTO_ACCOUNT_ID", "123456789012")
-RESPONSES_VERSION = pkg_resources.get_distribution('responses').version
+RESPONSES_VERSION = pkg_resources.get_distribution("responses").version
 
 
 class BaseMockAWS(object):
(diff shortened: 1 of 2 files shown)


davidszotten:
should we consider trying to detect version or return type rather than forcing the latest version of the library?

ypankovych:
Is this gonna bump the major version?

garribas:
@davidszotten imho implementing a monkey patch that deals with different versions of a third party library is bad long term from a maintainability perspective.

@ypankovych I'm not familiar with moto's release process and how versions are bumped, happy to bump version if you point me how to do it.

Current and past versions of moto are currently broken if doing a fresh install. If this approach isn't enticing we can instead change setup.py to not install responses versions post 0.12.0.

bblommers (changes requested):
Thanks for the PR @garribas! I agree with @davidszotten on this - I'd prefer a solution where we detect the return type.

> imho implementing a monkey patch that deals with different versions of a third party library is bad long term from a maintainability perspective.

Agreed, but it's the lesser of two evils IMO. There are so many projects using moto, there will always be some who can't upgrade a specific dependency. Having some techdebt is preferable to locking people out from using Moto altogether.

garribas:
@bblommers I've updated the PR to detect version (detecting return type looked too fragile), relies on pkg_version (setuptools dependency) and distutils (built-in) to detect responses lib version.

davidszotten:
1) i wonder if we just want to different versions of the whole function, and then detect version before monkeypatching instead of having to check (twice) at runtime. i think this would be more readable

2) consider using importlib.metadata (falling back to the backport importlib-metadata) which is way faster than pkg_resources

moto/core/models.py
```diff
@@ -252,12 +255,22 @@ def _url_matches(self, url, other, match_querystring=False):
 
 
 def _find_first_match(self, request):
+    match_failed_reasons = []
     for i, match in enumerate(self._matches):
-        if match.matches(request):
-            return match
-
-    return None
+        if LooseVersion(RESPONSES_VERSION) < LooseVersion('0.12.1'):
```
bblommers, line 260: For readability, would you mind rewriting this to only have the single if-statement? 
Whether it's a separate function or everything in the same function is the same to me.

Usage of 'pkg_resources' is fine IMO.

You might have to run 'black' as well afterwards, as Travis is currently complaining about it.

bblommers (changes requested):

bblommers (approved):
LGTM. Thank you so much @garribas!

mludvig:
Confirming that this change fixes our problem with `responses==0.12.1`, thanks!

Any chance this can be released to pypi sometime soon?

bblommers:
A dev-release is already available on PyPi, @mludvig - it is part of moto >= 1.3.16.dev122

flying-sheep:
I don’t get it. 1.3.16 was released in september and `1.3.16.dev122 < 1.3.16`.

Does moto use the wrong version tags?

Issue #3339, opened 2020-09-27
DynamoDB supports empty values


icevita:
https://github.com/spulec/moto/blob/55e7caccfe2482c90e4fb52214ef9e031ccdda1e/moto/dynamodb2/responses.py#L24

https://aws.amazon.com/about-aws/whats-new/2020/05/amazon-dynamodb-now-supports-empty-values-for-non-key-string-and-binary-attributes-in-dynamodb-tables/

bblommers:
Thanks for raising this @icevita! Will mark it as an enhancement.

bblommers:
This is now supported as of moto >= 1.3.16.dev126

PR #3467, opened 2020-11-13
Add support for empty strings in non-key dynamo attributes

richunger:
https://github.com/spulec/moto/issues/3339


bblommers (approved):
LGTM. Thanks @richunger!

Issue #3469, opened 2020-11-13
XML encoding is missing


ayushghosh:
When generating XML for the responses the values/attributes are not escaped. Thus it breaks when the payload contains characters that need to be escaped to be compliant.
Ex: I was using though https://github.com/localstack/localstack
When creating route53 record with escaped character &
`{
            "Comment": "CREATE/DELETE/UPSERT a record ",
            "Changes": [{
            "Action": "UPSERT",
                        "ResourceRecordSet": {
                                    "Name": "b.example.com",
                                    "Type": "TXT",
                                    "TTL": 300,
                                 "ResourceRecords": [{ "Value": "MyValue&NewValue"}]
}}]
}`

Response:
`
<ListResourceRecordSetsResponse xmlns="https://route53.amazonaws.com/doc/2012-12-12/">
   <ResourceRecordSets>
      <ResourceRecordSet>
                <Name>b.example.com.</Name>
                <Type>TXT</Type>
                    <TTL>300</TTL>
                <ResourceRecords>
                    <ResourceRecord>
                        <Value>MyValue&NewValue</Value>
                    </ResourceRecord>
                </ResourceRecords>
            </ResourceRecordSet>
   </ResourceRecordSets>
   <IsTruncated>false</IsTruncated>
</ListResourceRecordSetsResponse>
`

This generates error: Unable to parse response (not well-formed (invalid token): line 19, column 47), invalid XML received. Further retries may succeed:

I was able to make the jinja templates to work by escaping the attribute with "escape" filter

bblommers:
Thanks for raising this @ayushghosh! Marking it as a bug.

oidebrett:
I can recreate the issue with the following test case:

@mock_route53
def test_list_resource_records_with_ampersand():
    conn = boto3.client("route53", region_name="us-east-1")
    create_hosted_zone_response = conn.create_hosted_zone(
        Name="db.",
        CallerReference=str(hash("foo")),
        HostedZoneConfig=dict(PrivateZone=True, Comment="db"),
    )
    hosted_zone_id = create_hosted_zone_response["HostedZone"]["Id"]

    def create_resource_record_set(rec_type, rec_name):
        payload = {
            "Comment": "create {} record {}".format(rec_type, rec_name),
            "Changes": [
                {
                    "Action": "UPSERT",
                    "ResourceRecordSet": {
                        "Name": rec_name,
                        "Type": rec_type,
                        "TTL": 300,
[...]

PR #3473, opened 2020-11-16
Fix XML encoding in Route53 JInja2 Templates #3469

ayushghosh:
Use Jinja2 escape functionality to escape html attributes in value response of route53
Fix for: https://github.com/spulec/moto/issues/3469


bblommers (changes requested):
Looks like Travis is failing because of the linting.
Would you mind running black on these files? Happy to merge after that

https://travis-ci.org/github/spulec/moto/jobs/744023235

bblommers (approved):
LGTM - very clean solution.

Thanks @ayushghosh!

Issue #3464, opened 2020-11-13
BUG: update_configuration for lambda is trying to set a property


SophieMJ:
## Reporting Bugs

When using moto for testing `update_function_configuration`, this block of code is being called: https://github.com/spulec/moto/blob/47dbad291e717059a93f5cf795a9be08cfc872cc/moto/awslambda/models.py#L293-L310

However, this is giving an error provided below:
```
elif key == “VpcConfig”:
       self.vpc_config = value
              AttributeError: can’t set attribute

../../.virtualenvs/lambda-zvljBF5N/lib/python3.8/site-packages/moto/awslambda/models.py:323: AttributeError
```

This is because on L308 it's trying to set a value to `vpc_config` which is a property (not an attribute) that doesn't have a setter. I think changing it to `self._vpc_config = value` should fix the problem, but any suggestions/comments are welcome.

bblommers:
Hey @SophieMJ, thanks for raising this! Looks like that should be as simple as changing it to `_vpc_config`, yes. 
Marking it as a bug.

bblommers:
This is now fixed as of moto >= 1.3.16.dev130

PR #3479, opened 2020-11-17
fixed issue in update_configuration for lambda when setting VPCconfig

oidebrett:
#### Reference Issue
https://github.com/spulec/moto/issues/3464

#### What does this implementation fix?
This implementation fixes the bug in awslambda/models.py when updating the VPCconfig. The code in def update_configuration was trying to set a value to vpc_config which is a property (not an attribute) that doesn't have a setter. Changing to setting self._vpc_config = value fixes the problem.

This PR also updates the test case "test_update_configuration" in tests/test_awslambda/test_lambda.py to add the VpcConfig property and an assertion to verify the change.


bblommers (approved):
LGTM - thanks @oidebrett!

PR #3480, opened 2020-11-18
Fix Race Condition in batch:SubmitJob

### LINKED ISSUE #3475: Travis CI jobs failing intermittently
`test_batch.test_submit_job` fails intermittently across all Python versions and regardless of whether `TEST_SERVER_MODE=false` or `TEST_SERVER_MODE=true`.

<details>
<summary>Nose Failure</summary>
<p>

```
======================================================================
FAIL: test_batch.test_submit_job
----------------------------------------------------------------------
  File "/home/travis/virtualenv/python3.7.6/lib/python3.7/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/home/travis/build/spulec/moto/moto/core/models.py", line 99, in wrapper
  File "/home/travis/build/spulec/moto/moto/core/models.py", line 99, in wrapper
  File "/home/travis/build/spulec/moto/moto/core/models.py", line 99, in wrapper
  [Previous line repeated 2 more times]
  File "/home/travis/build/spulec/moto/tests/test_batch/test_batch.py", line 750, in test_submit_job
    [event["message"] for event in resp["events"]].should.equal(["hello"])
  File "/home/travis/virtualenv/python3.7.6/lib/python3.7/site-packages/sure/__init__.py", line 387, in wrapper
    raise AssertionError(e)
AssertionError: given
X = []
    and
Y = ['hello']
Y has 1 items whereas X has only 0
>>  raise AssertionError(e)
```

</p>
</details>  

The test is still failing occasionally after the recent switch from nose to pytest.

<details>
<summary>Pytest Failure</summary>
<p>

```
=================================== FAILURES ===================================
_______________________________ test_submit_job ________________________________

self = <sure.AssertionBuilder object at 0x7fd94262df60>, args = (['hello'],)
kw = {}

    @wraps(func)
    def wrapper(self, *args, **kw):
        try:
>           value = func(self, *args, **kw)

../../../virtualenv/python3.6.10/lib/python3.6/site-packages/sure/__init__.py:385: 

self = <sure.AssertionBuilder object at 0x7fd94262df60>, what = ['hello']
epsilon = None

    @assertionmethod
    def equal(self, what, epsilon=None):
        """compares given object ``X``  with an expected ``Y`` object.
    
        It primarily assures that the compared objects are absolute equal ``==``.
    
        :param what: the expected value
        :param epsilon: a delta to leverage upper-bound floating point permissiveness
        """
    
        try:
            comparison = DeepComparison(self.obj, what, epsilon).compare()
            error = False
        except AssertionError as e:
            error = e
            comparison = None
    
        if isinstance(comparison, DeepExplanation):
            error = comparison.get_assertion(self.obj, what)
    
        if self.negative:
            if error:
                return True
    
            msg = '%s should differ from %s, but is the same thing'
            raise AssertionError(msg % (safe_repr(self.obj), safe_repr(what)))
    
        else:
            if not error:
                return True
>           raise error
E           AssertionError: given
E           X = []
E               and
E           Y = ['hello']
E           Y has 1 items whereas X has only 0

../../../virtualenv/python3.6.10/lib/python3.6/site-packages/sure/__init__.py:673: AssertionError
```

</p>
</details>

This "flipper test" was mentioned and acknowledged as a problem in #3161, but I wasn't able to find an existing issue tracking it.  I will investigate.

bpandola:
The `test_submit_job` in `test_batch.py` kicks off a job, calls `describe_jobs`
in a loop until the job status returned is SUCCEEDED, and then asserts against
the logged events.

The backend code that runs the submitted job does so in a separate thread. If
the job was successful, the job status was being set to SUCCEEDED *before* the
event logs had been written to the logging backend.

As a result, it was possible for the primary thread running the test to detect
that the job was successful immediately after the secondary thread had updated
the job status but before the secondary thread had written the logs to the
logging backend.  Under the right conditions, this could cause the subsequent
logging assertions in the primary thread to fail.

Additionally, the code that collected the logs from the container was using
a "dodgy hack" of time.sleep() and a modulo-based conditional that was
ultimately non-deterministic and could result in log messages being dropped
or duplicated in certain scenarios.

In order to address these issues, this commit does the following:

* Carefully re-orders any code that sets a job status or timestamp
  to avoid any obvious race conditions.
* Removes the "dodgy hack" in favor of a much more straightforward
  (and less error-prone) method of collecting logs from the container.
* Removes arbitrary and unnecessary calls to time.sleep()

Before applying any changes, the flaky test was failing about 12% of the
time.  Putting a sleep() call between setting the `job_status` to SUCCEEDED
and collecting the logs, resulted in a 100% failure rate.  Simply moving
the code that sets the job status to SUCCEEDED to the end of the code block,
dropped the failure rate to ~2%.  Finally, removing the log collection
hack allowed the test suite to run ~1000 times without a single failure.

Taken in aggregate, these changes make the batch backend more deterministic
and should put the nail in the coffin of this flaky test.

Closes #3475

### DIFF
--- a/moto/batch/models.py
+++ b/moto/batch/models.py
@@ -392,7 +392,6 @@ class Job(threading.Thread, BaseModel, DockerModel):
         """
         try:
             self.job_state = "PENDING"
-            time.sleep(1)
 
             image = self.job_definition.container_properties.get(
                 "image", "alpine:latest"
@@ -425,8 +424,8 @@ class Job(threading.Thread, BaseModel, DockerModel):
 
             self.job_state = "RUNNABLE"
             # TODO setup ecs container instance
-            time.sleep(1)
 
+            self.job_started_at = datetime.datetime.now()
             self.job_state = "STARTING"
             log_config = docker.types.LogConfig(type=docker.types.LogConfig.types.JSON)
             container = self.docker_client.containers.run(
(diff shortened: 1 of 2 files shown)


bpandola:
🤞 the Travis run succeeds and doesn't make me look like a fool.  😆

bblommers (approved):
Nice one, @bpandola! This is a much cleaner approach- well done!

PR #3481, opened 2020-11-18
Py3: use unittest.mock instead of mock

singingwolfboy:


singingwolfboy:
Could someone please review this pull request? I just rebased it.

tests/test_packages/test_httpretty.py
```diff
@@ -1,7 +1,11 @@
 # #!/usr/bin/env python
 # -*- coding: utf-8 -*-
 from __future__ import unicode_literals
-import mock
+
```
bblommers, line 4: This import is unused, as far as I can tell

bblommers (changes requested):
Hi @singingwolfboy, thanks for this PR!
Would you mind moving the try-catches to `compat.py`? The file is already in place in both the moto and tests directory - having all compatibility imports in one place will make it easier to keep track of them.

Looks like the CI is failing at the moment btw, but I haven't looked into why. Let me know if you need a hand with that.

bblommers (changes requested):
Nice @singingwolfboy - happy to merge once the build passes

singingwolfboy:
@bblommers I finally got the tests to pass! Can you check my code once again, to make sure that the changes I made are acceptable?

bblommers:
Thanks for getting the build fixed,  @singingwolfboy. 
The try-except brings back some painful memories of #3541, which (I think) fixed the same problem in PY2, introduced by `mock==4.0.3`. 

I'll merge this for now, and have a look later on to see if my patch can be reverted with this solution.

bblommers (approved):

PR #3485, opened 2020-11-19
Fix: redshift:EnableSnapshotCopy raises incorrect Exception

bpandola:
The previous code was trying to raise a botocore ClientError directly, which
was actually generating a secondary AttributeError because the arguments passed
to ClientError() were incorrect.

This replaces the ClientError() call with a proper moto exception class for
Redshift and fixes the test assertions appropriately.

Supersedes #1957


bblommers (approved):
Thanks!

bblommers:
This is now part of moto >= 1.3.16.dev136

PR #3486, opened 2020-11-19
Route53: Implement dummy GetChange endpoint

relvira:
## Description
This PR implements a dummy GetChange endpoint for the Route53 mock within moto. This change has been discussed in #3449.

The Route53 mock in moto does not store Change IDs in it's backend at the moment, this is why I created a dummy endpoint that always returns INSYNC status together with the same SubmittedAt date for all requests.

Ideally, the Route53 mock backend would store Change IDs the same way it stores hostedzones, healthchecks, etc. I've discussed the possibility of doing this with @bblommers and he's keen on that, but for now I'll stick with this approach as the ideal option would require significant changes in the route53 mock to support and store those change IDs.

## Why
GetChange is used by existing tooling around Route53, which makes supporting this endpoint in moto very advantageous when using moto in local development environments as well as testing for Route53.

## Testing
I've added one test to ensure `get_change` always return the `Id` it was asked for as part if it's response and `Status` is `INSYNC`.

### DIFF
--- a/moto/route53/responses.py
+++ b/moto/route53/responses.py
@@ -243,6 +243,15 @@ class Route53(BaseResponse):
 
             return 200, headers, template.render()
 
+    def get_change(self, request, full_url, headers):
+        self.setup_class(request, full_url, headers)
+
+        if request.method == "GET":
+            parsed_url = urlparse(full_url)
+            change_id = parsed_url.path.rstrip("/").rsplit("/", 1)[1]
+            template = Template(GET_CHANGE_RESPONSE)
+            return 200, headers, template.render(change_id=change_id)
+
 
 LIST_TAGS_FOR_RESOURCE_RESPONSE = """
 <ListTagsForResourceResponse xmlns="https://route53.amazonaws.com/doc/2015-01-01/">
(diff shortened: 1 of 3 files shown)


moto/route53/responses.py
```diff
@@ -382,3 +391,12 @@ def list_or_change_tags_for_resource_request(self, request, full_url, headers):
 DELETE_HEALTH_CHECK_RESPONSE = """<?xml version="1.0" encoding="UTF-8"?>
     <DeleteHealthCheckResponse xmlns="https://route53.amazonaws.com/doc/2013-04-01/">
 </DeleteHealthCheckResponse>"""
+
+GET_CHANGE_RESPONSE = """<?xml version="1.0" encoding="UTF-8"?>
+<GetChangeResponse xmlns="https://route53.amazonaws.com/doc/2013-04-01/">
+   <ChangeInfo>
+      <Status>INSYNC</Status>
+      <SubmittedAt>2010-09-10T01:36:41.958Z</SubmittedAt>
+      <Id>/change/{{ change_id }}</Id>
```
bblommers, line 400: Looks like the test fails because of this, as the ID should just be {{change_id}} (without '/change/')

Happy to merge once Travis is happy.

bblommers (changes requested):

moto/route53/responses.py
relvira: oh! that's definitely a typo! I'll fix that now.

bblommers (approved):
LGTM. Thanks @relvira!

PR #3488, opened 2020-11-20
Fix: Python 2/3 Incompatibility

bpandola:
Previous code would raise `TypeError: 'dict_keys' object is not subscriptable`
when run under Python 3.

* Re-write code in Python 2/3 compatible way.
* Add clarifying comment.
* Add test coverage.

Supersedes #3227


bblommers (approved):
LGTM

bblommers:
This is now part of moto >= 1.3.16.dev137

PR #3489, opened 2020-11-22
Fix: EMR `ReleaseLabel` validation does not respect semantic versioning

### LINKED ISSUE #3474: Using CustomAmiId when calling run_job_flow incorrectly throws ValidationException for all EMR versions between 5.10.0 & 5.31.0
Right now the check in `run_job_flow` to see if a CustomAmiId is [valid](https://github.com/spulec/moto/blob/master/moto/emr/responses.py#L311) is done with default python string comparison instead of in a manner which would respect semantic versioning.
So `"emr-5.7.0"` is always considered greater than any value between `"5.1..."`(like 5.10.0) & `"5.6..."`(like 5.60.0) which is incorrect behavior

Reproducible like such:
```python
import boto3
from moto import mock_emr

run_job_flow_args = dict(
    Instances={
        "InstanceCount": 3,
        "KeepJobFlowAliveWhenNoSteps": True,
        "MasterInstanceType": "c3.medium",
        "Placement": {"AvailabilityZone": "us-east-1a"},
        "SlaveInstanceType": "c3.xlarge",
    },
    JobFlowRole="EMR_EC2_DefaultRole",
    LogUri="s3://mybucket/log",
    Name="cluster",
    ServiceRole="EMR_DefaultRole",
    VisibleToAllUsers=True,
    ReleaseLabel="emr-5.31.0", # Semantically greater than 5.7.0 and EMR actually supports CustomAmiId
    CustomAmiId="ami-1234567"
)

@mock_emr
def reproduce_issue():
    client = boto3.client("emr", region_name="us-east-1")
    client.run_job_flow(**run_job_flow_args)
```

bpandola:
I changed the existing test to use a ReleaseLabel version specifically called out in the issue report as not working.

Fixes #3474

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -3,6 +3,7 @@
 exclude_lines =
     if __name__ == .__main__.:
     raise NotImplemented.
+    return NotImplemented
     def __repr__
 
 [run]
--- a/moto/emr/responses.py
+++ b/moto/emr/responses.py
@@ -13,7 +13,7 @@ from moto.core.responses import xml_to_json_response
 from moto.core.utils import tags_from_query_string
 from .exceptions import EmrError
 from .models import emr_backends
-from .utils import steps_from_query_string, Unflattener
+from .utils import steps_from_query_string, Unflattener, ReleaseLabel
 
 
 def generate_boto3_response(operation):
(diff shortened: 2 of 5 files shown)


bblommers (approved):
LGTM. Thanks @bpandola!

PR #3491, opened 2020-11-23
Added instance type offerings and enhanced instance types

nuwang:
Adds support for:
- A more comprehensive instance type list based on the existing instance_types.json file.
- Filtering instance types
- Querying instance type offerings
- Filtering instance type offerings


nuwang:
ping @spulec @JackDanger

bblommers (changes requested):
Hi @nuwang, thanks for your PR!

Having an explicit list like this is not really maintainable, especially considering that we'd have to extend this to support all regions (not just us-east/us-west).
Is is possible to make it a list of instance_types only, and return the same list for every region?

I'm not too familiar with the (un)availability of specific instance types for specific regions, but I'm hoping there aren't too many differences between regions

nuwang:
Hi @bblommers. Thanks for reviewing. Since the availability of instance types is affected by the available capacity in a particular zone, I think it would be useful for moto to support scenarios where some regions just don't have the requested instance type. Clients using the instance type offering api should ideally account for that, and expect instance types to be unavailable, and this would help them to test that explicitly. Perhaps we could always return at least one instance type by default, say t2.nano, to handle regions in which no instance types at all are defined?

nuwang:
Also, I forgot to mention that I generated this offerings list using the existing [instances.json](https://github.com/spulec/moto/blob/master/moto/ec2/resources/instance_types.json) file, by adding just two availability zones to each instance type, and modified a few to test for specific test conditions.

bblommers:
@nuwang Gotcha, I didn't think of the scenario of wanting to test for unavailability.
In that case, a full list might be a better solution after all. Just thinking out loud here, but potentially with a script that periodically downloads the full list from AWS itself per region.

@spulec Any thoughts on this?

spulec:
Since [Availability Zone names don't map to the same location across accounts](https://aws.amazon.com/premiumsupport/knowledge-center/vpc-map-cross-account-availability-zones/), I'm a bit nervous about having these be different across AZs.

If I am a developer using Moto and I know that us-east-1a in my AWS account supports a given instance type, that seems reasonable to me to potentially hardcode it. It would feel weird if Moto then said the instance type wasn't supported.

I'm a bit more comfortable with them varying by Region if we think there is a maintainable way to keep that up to date with the real availabilities. Can we write a script that uses these APIs for real and dumps it to a json?

bblommers:
@nuwang Hope you don't mind - I'll have a stab at creating this script now

nuwang:
@bblommers Not at all, much appreciated!

PR #3492, opened 2020-11-23
Support Cognito ADMIN_USER_PASSWORD_AUTH auth flow

robjwilkins:
Uses the user credentials pattern for the ADMIN_NO_SRP_AUTH flow, and applies
them to the ADMIN_USER_PASSWORD_AUTH auth flow for Cognito admin_initiate_auth
requests.


bblommers (changes requested):
Thanks for the PR @robjwilkins! 

Would you mind adding a unit test for each path? You might have to run `black` as well afterwards to make Travis happy.

tests/test_cognitoidp/test_cognitoidp.py
```diff
@@ -1198,7 +1198,7 @@ def test_admin_delete_user():
     caught.should.be.true
 
 
-def authentication_flow(conn):
+def authentication_flow(conn, auth_flow):
```
bblommers, line 1201: Looks like there are other tests that use this method, so we can probably just default the `auth_flow` param to `"ADMIN_NO_SRP_AUTH"`

bblommers (changes requested):

tests/test_cognitoidp/test_cognitoidp.py
robjwilkins: Thanks. I've been trying to run the tests locally but getting all sorts of errors I think due to python3 and installing the necessary project dependencies 😢
Will have another look today

bblommers (changes requested):
Thanks for making the change @robjwilkins! 
Would you mind providing a clean PR? Happy to merge after that

robjwilkins:
Closing this until I've got the changes working

robjwilkins:
Hopefully fixed the test problems 🤞

Issue #3494, opened 2020-11-23
Go SDK doesn't work with moto (+ localstack)


chris922:
**how to reproduce the issue**
Use `latest` version of localstack (not 0.12.2) and Go SDK. E. g. with the following code:

```go
import (
	"github.com/aws/aws-sdk-go/aws"
	"github.com/aws/aws-sdk-go/aws/credentials"
	"github.com/aws/aws-sdk-go/aws/session"
	"github.com/aws/aws-sdk-go/service/eventbridge"
	"testing"
)

func TestEventbridge(t *testing.T) {
	session := session.Must(session.NewSession())
	events := eventbridge.New(session, &aws.Config{
		Region: aws.String("eu-central-1"),
		Endpoint: aws.String("http://localhost:4566/"),
		Credentials: credentials.NewStaticCredentials("test", "test", ""),
	})

	resp, err := events.ListEventBuses(&eventbridge.ListEventBusesInput{})
	if err != nil {
		print(err.Error())
	} else {
		print(resp.String())
	}
}
```

**what you expected to happen**

I expect that the operation finishes successfully and does not return an error.

**what actually happens**

I get an error and the (localstack) logs contains the following:

```
2020-11-23T22:58:31:WARNING:localstack.multiserver: Unable to create moto backend app for empty service: "None"
2020-11-23 22:58:31,028:API: Error on request:
Traceback (most recent call last):
  File "/home/christian/Development/git/localstack/.venv/lib/python3.8/site-packages/werkzeug/serving.py", line 323, in run_wsgi
    execute(self.server.app)
  File "/home/christian/Development/git/localstack/.venv/lib/python3.8/site-packages/werkzeug/serving.py", line 312, in execute
    application_iter = app(environ, start_response)
  File "/home/christian/Development/git/localstack/.venv/lib/python3.8/site-packages/moto/server.py", line 168, in __call__
    return backend_app(environ, start_response)
TypeError: 'NoneType' object is not callable
```

**what version of Moto you're using**

1.3.15

**how you installed it**

Via localstack

**possible fix**

I reported a similar issue already on LocalStack (see https://github.com/localstack/localstack/issues/3255). The problem is, that the Go SDK signs the requests with the service name `EventBridge` and not `events`.
Therefore I checked if moto also uses the service name from the signing header and I found this in the function `infer_service_region_host` of the `server.py`.

I will create a PR with a possible fix that I verified by modifying the files on my system and re-running the mentioned reproducer.

[...]

PR #3495, opened 2020-11-23
#3494 fix using EventBridge via Go SDK

chris922:
As already mentioned in #3494 I verified this fix by modifying this file locally on my system and re-running the reproducer given in the issue.

The issue is that the Go SDK uses `EventBridge` in the SigV4 header instead of `events` as expected by moto.

### DIFF
--- a/moto/server.py
+++ b/moto/server.py
@@ -93,6 +93,11 @@ class DomainDispatcherApplication(object):
                 # S3 is the last resort when the target is also unknown
                 service, region = DEFAULT_SERVICE_REGION
 
+        if service == "EventBridge":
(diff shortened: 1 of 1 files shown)


bblommers (changes requested):
Hi @chris922, thanks for the PR! Would you mind running `black` on this file? 

(Not sure what could be wrong in only 4 lines, but Travis is still unhappy about it..)

chris922:
Ah, the linter doesn't like whitespaces in empty lines.. I removed them and pushed the update. Hopefully it works now :)

bblommers (approved):
LGTM!

PR #3496, opened 2020-11-24
Redshift improvements

bpandola:
I discovered and fixed a number of (minor) Redshift issues while working on an unrelated task.

I took care to separate all these changes into atomic commits, so if you don't squash them it should provide a better experience for anyone using `git blame` down the road.


bblommers (approved):
LGTM. Thanks @bpandola!

bblommers:
This is now part of moto >= 1.3.16.dev144

PR #3501, opened 2020-11-25
Mark sts.get_caller_identity as implemented

sim1234:
Ref: https://github.com/spulec/moto/search?q=get_caller_identity


bblommers (approved):
LGTM

PR #3505, opened 2020-11-29
FIX:Ec2:Added:Tags for EIP

usmangani1:
FIX:Ec2:Added:Tags for EIP
https://github.com/localstack/localstack/issues/3228


bblommers (changes requested):
Some of the other tests are failing. It looks like get_tags() can fail if there are no tags:

https://travis-ci.org/github/spulec/moto/jobs/746562068

bblommers (approved):
LGTM

bblommers:
This is now part of moto >= 1.3.16.dev171

PR #3507, opened 2020-12-01
Add basic get_execution_history implementation for Step Functions

ciaranevans:
## What I am changing

This PR adds a very simplistic implementation of the [`get_execution_history`](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/stepfunctions.html#SFN.Client.get_execution_history) method on the Step Function client. For now it only uses the parameter `executionArn`.

I've hard coded two histories to be returned:

**A successful execution (Presumably useful to 99% of users) with events :**
```
ExecutionStarted -> PassStateEntered -> PassStateExited -> ExecutionSucceeded
```

**A failed execution (We have a use case where we need failures in the execution history for our testing) with events:**
```
ExecutionStarted -> FailStateEntered -> ExecutionFailed
```

One can enforce which history they want with the environment variable `EXECUTION_HISTORY_TYPE`. If `SUCCESS` (Which is the default if no env var is present) then we return the successful history otherwise, if it's `FAILURE` we return the failed history.

---

I've also added `format` as a Makefile command, it simply calls the formatting equivalents of the `lint` command.

## How I did it

* Add `format` to `Makefile` 
* Moved `start_execution` to be a function of the `StateMachine` class
* Added `executions` as an attribute of the `StateMachine` class
* Refactor `start_execution`, `stop_execution`, `list_executions`, and `describe_execution` in `StepFunctionBackend` to correctly operate on the executions for the correct state machine
* Added `_get_state_machine_for_execution` as a helper function in `StepFunctionBackend` to return the associated `StateMachine` for a given executions ARN
* Added `get_execution_history` as a function of the `StateMachine` class - this returns a failure or a success event history, depending on the `EVENT_HISTORY_TYPE` environment variable
* Add an except block to `stop_execution` in `moto/stepfunctions/responses.py` to accommodate failure cases
* Add `get_execution_history` in `moto/stepfunctions/responses.py`
* Add unit test to cover case where `stop_execution` would get into a failure state
* Add unit tests to cover success history, failure history, and failure case of `get_execution_history` invocations

## How you can test it

```bash
# In the repo root
$ pytest ./tests/test_stepfunctions
```

## Questions

I'll namedrop @bblommers 😇  - I think the `EVENT_HISTORY_TYPE` environment variable usage needs documenting somewhere but I couldn't decide _where_... is the README OK? If so, where abouts?

Again, all the thanks for the `moto` crew! ❤️ 🚀

### DIFF
--- a/moto/stepfunctions/models.py
+++ b/moto/stepfunctions/models.py
@@ -264,7 +264,7 @@ class Execution:
                     },
                 },
             ]
-        else:
+        elif execution_history_type == "FAILURE":
             return [
                 {
                     "timestamp": iso_8601_datetime_with_milliseconds(
--- a/moto/stepfunctions/models.py
+++ b/moto/stepfunctions/models.py
@@ -1,6 +1,8 @@
 import json
+import os
 import re
 from datetime import datetime
+from dateutil.tz import tzlocal
 
 from boto3 import Session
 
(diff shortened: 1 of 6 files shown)


ciaranevans:
Is there a way to pass environment variables in tests for the test server runs?

I'm seeing a failure and I guess the patching isn't working across to the server..

moto/stepfunctions/models.py
```diff
@@ -176,6 +208,104 @@ def __init__(
         self.status = "RUNNING"
         self.stop_date = None
 
+    def get_execution_history(self, roleArn):
+        execution_history_type = os.environ.get("EXECUTION_HISTORY_TYPE", "SUCCESS")
```
bblommers, line 212: This should probably move to settings.py, probably prefixed with `SFN_` to make it more obvious where it's used.

As to documentation: We don't really have any similar environment variables, or a good place to document them. Maybe just add a comment to settings.py with a short explanation.

tests/test_stepfunctions/test_stepfunctions.py
```diff
@@ -797,12 +805,32 @@ def test_state_machine_stop_execution():
...
+        )
+        client.stop_execution(executionArn=unknown_execution)
+    ex.value.response["Error"]["Code"].should.equal("ExecutionDoesNotExist")
+    ex.value.response["Error"]["Message"].should.contain("Execution Does Not Exist:")
+
+
 @mock_stepfunctions
 @mock_sts
 def test_state_machine_describe_execution_after_stoppage():
-    account_id
```
bblommers, line 803: Ehm.. what. Why would anyone have this in a test.
Nice find :)

moto/stepfunctions/models.py
```diff
@@ -346,23 +476,22 @@ def update_state_machine(self, arn, definition=None, role_arn=None):
...
             account_id=self._get_account_id(),
-            state_machine_name=state_machine_name,
             execution_name=name or str(uuid4()),
-            state_machine_arn=state_machine_arn,
             execution_input=execution_input,
         )
-        self.executions.append(execution)
         return execution
 
     def stop_execution(self, execution_arn):
```
bblommers, line 488: This should be slightly refactored as well, to `state_machine.stop_execution()`

bblommers (changes requested):

moto/stepfunctions/models.py
ciaranevans: @bblommers does `settings.py` exist already? Or are you asking me to make one for Step Functions specifically?

tests/test_stepfunctions/test_stepfunctions.py
ciaranevans: Hah, just raised an eyebrow when I saw it!

moto/stepfunctions/models.py
ciaranevans: Cool, will do!

moto/stepfunctions/models.py
bblommers: It already exist: `moto/settings.py`

ciaranevans: 🙃  Wow VS Code search failed me there. Cool will do!

ciaranevans: Okay so doing that makes actually setting it using mock.patch a lot harder - is there a recommended way to change a setting in a unit test style?

I guess this problem will exist when trying to use moto also.

ciaranevans: As the value is assigned as soon as 'settings' is imported, the consumer of `moto` is stuck with whatever value it assigns first

bblommers: Ah, good point. The other settings don't need to be recomputed during testing, so this is a first.

Easiest way to get around that is probably to just make it a function, instead of a property:
`def get_..(): return os.environ.get(.., ..)`

The alternatives would probably be to either:
 - patch settings.py:property, if that's possible
 - or reload the settings.py (using importlib.reload)

Neither are pretty IMO

ciaranevans: If we're happy with it as a function in `settings.py` I can certainly do that! I'll put it in and push it up!

bblommers (approved):
LGTM. Thanks @ciaranevans!

PR #3508, opened 2020-12-02
EC2: describe instance type offerings

bblommers:
Supercedes #3491 

Contains a script that lists all instance-type-offerings, separated across region and location-type (region/availability-zone/availability-zone-id)

Don't have access to the US gov and China-regions, so I haven't added the offerings for those regions.


bblommers:
@nuwang Had to create a new PR, as I don't have permissions to push to your branch.
Made a small change to the way the filter works, as it's possible to specify multiple filters- and only instance types that adhere to all filters should be returned. Let me know if I've missed anything.

@spulec Would you mind taking a look as well?

spulec:
Looks great; thank you both!

PR #3509, opened 2020-12-02
Fix:Route53 list-resource-change-records-template

usmangani1:
Fix:Route53 list-resource-change-records-template
https://github.com/localstack/localstack/issues/3172


bblommers (changes requested):
Some of the tests are breaking because of missing fields, but this looks like the right approach.

If you take the full output of the `to_xml()` method, it should work.
I'm assuming that the `to_xml` method is no longer used after that, so we can remove that. (It shouldn't be part of the model anyway..)
Could do with a test as well, to verify the content of the <ResourceRecord> is properly escaped

bblommers:
Closing this in favour of #4673

PR #3513, opened 2020-12-03
Fix: ClusterType, NumberOfNodes not validated when modifying Redshift…

bpandola:
… cluster

By definition, `single-node` clusters can only consist of 1 node. Likewise,
`multi-node` clusters must have 2 or more nodes.

* Ensure `ClusterType` parameter is either `multi-node` or `single-node`.
* Ensure proper validation of `NumberOfNodes` parameter based on `ClusterType`
  parameter.
* Fix existing test case that incorrectly allowed a `multi-node` cluster to
  consist of 1 node.
* Add dedicated test for resizing a cluster from `single-node` to `multi-node`
  and back again.

Behavior and error messages have been verified against a real AWS backend.

### DIFF
--- a/moto/redshift/models.py
+++ b/moto/redshift/models.py
@@ -632,6 +632,20 @@ class RedshiftBackend(BaseBackend):
         cluster_identifier = cluster_kwargs.pop("cluster_identifier")
         new_cluster_identifier = cluster_kwargs.pop("new_cluster_identifier", None)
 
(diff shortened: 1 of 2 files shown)


bblommers (approved):
LGTM

bblommers:
This is now part of moto >= 1.3.16.dev169

Issue #3478, opened 2020-11-17
dev release numbering doesn't work with pip-tools because it's not PEP 440 compliant


richunger:
According to https://www.python.org/dev/peps/pep-0440/#summary-of-permitted-suffixes-and-relative-ordering
dev releases should be pre release numbers, not post. So, 1.3.16devX should be 1.3.17devX.

Practically, this means that pip-tools will take a requirements.in file with
moto>=1.3.16dev100
and output a requirements.txt file with
moto==1.3.16

Workaround is to specify an exact dependency in requirements.in:
moto==1.3.16dev100

but then you won't pick up new dev releases as they come out, or the final 1.3.17.

richunger:
My clever colleague suggested a better workaround: `moto>=1.3.16dev100,!=1.3.16`

bblommers:
Thanks for raising this @richunger - this is now fixed.
The first release post 1.3.16 is now 1.3.17.dev221

PR #3518, opened 2020-12-04
Fix dev versions

flying-sheep:
You’re using versions wrong, [see PEP 440](https://www.python.org/dev/peps/pep-0440/#developmental-releases).

> The developmental release segment consists of the string .dev, followed by a non-negative integer value. Developmental releases are ordered by their numerical component, **immediately before the corresponding release** […]

i.e.: `1.3.16.devX` < `1.3.16`

This PR uses proper types for the version code which compare correctly. 
An alternative to `1.3.17.devX` would be to use `1.3.16.postX`, which would properly sort as newer than `1.3.16`

The order on PyPI demonstrates that `1.3.16.devX` releases sort as older than `1.3.16`:

grafik


andersk (approved):
This fix is important for those who need to depend on `moto>=1.3.16.dev122` because of #3460, but then get “upgraded” back to `1.3.16`. Please merge.

andersk:
Could be marked as Fixes: #3478.

bblommers (changes requested):
Hi @flying-sheep, this change generates a version in an invalid format:
`1.3.17.dev0.172`

I've updated the version manually for now, to unblock people - 1.3.17.dev221 is the first properly named release.

flying-sheep:
@bblommers fixed! Too bad `Version`s are immutable, otherwise the code could have been sth. clean like:

```py
dev_version = init_version.copy()
dev_version.dev = commits_since
return dev_version
```

bblommers:
Closing this in favor of #4016

Issue #3231, opened 2020-08-12
DynamoDB: Update item doesn't validate against attributes


meshuamam:
Might be related to https://github.com/spulec/moto/issues/2396

Moto doesn't return an error when using putting the wrong attribute type.

Example:
When updating an attribute defined as:
```
table = dynamodb.create_table(
    ...
    AttributeDefinitions = [{
        'AttributeName': 'this_is_a_number',
        'AttributeType': 'N'
    }]
)
```
with "put item", using a string value such as "123123jkasdfsdf", moto doesn't return a 400 with a client error.

Expected:
"400 status" and a "The parameter cannot be converted to a numeric value" message (or similar)

Actual: 
2xx status

Version: 1.3.15.dev1023
Using through pynamodb
Installed as a mock, boto versions:
boto3==1.14.30  
boto==2.48.0   
botocore==1.17.30

bblommers:
Thanks for raising this @meshuamam - marking this as an enhancement

PR #3519, opened 2020-12-05
DynamoDB - Allow StringSets to be passed to update_item()

### LINKED ISSUE #3506: DynamoDB: DynamoDB.Client.update_item() Doesn't Work with String Set
**Summary**
`DynamoDB.Client.update_item()` expects `list` as input for String Set according to the [documentation](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/dynamodb.html#DynamoDB.Client.update_item):
```
SS (list) --
An attribute of type String Set. For example:

"SS": ["Giraffe", "Hippo" ,"Zebra"]
```
However **moto** accepts `set` only for `SS` since this [commit](https://github.com/spulec/moto/commit/a2aefc49b486f8191541f033007b6b3cf81ed723). And the logic is not reachable as **boto3** would fire `ParamValidationError` before reaching **moto**.

**Details**
I tested with the following code snippet.
```
    @mock_dynamodb2
    def test_dynamodb_update_item(self):
        dynamodb = boto3.resource('dynamodb', region_name='eu-west-1')

        table = dynamodb.create_table(
            TableName='test',
            KeySchema=[
                {'AttributeName': 'record_id', 'KeyType': 'HASH'},
            ],
            AttributeDefinitions=[
                {'AttributeName': 'record_id', 'AttributeType': 'S'},
            ],
        )
        table.meta.client.get_waiter('table_exists').wait(TableName='test')
        attribute = {
            'test_field': {
                'Value': {
                    'SS': ['test1', 'test2'],
                },
                'Action': 'PUT'
            }
        }

        boto3.client('dynamodb').update_item(
            TableName='test',
            Key={'record_id': {'S': 'testrecord'}},
            AttributeUpdates=attribute,
        )
```

**Expected Behavior**
Test to proceed without error.

**Actual Behavior**
```
self = <[AttributeError("'DynamoType' object has no attribute 'value'") raised in repr()] DynamoType object at 0x7faa6958da50>
type_as_dict = 'test'

    def __init__(self, type_as_dict):
        if type(type_as_dict) == DynamoType:
            self.type = type_as_dict.type
            self.value = type_as_dict.value
        else:
            self.type = list(type_as_dict)[0]
>           self.value = list(type_as_dict.values())[0]
E           AttributeError: 'str' object has no attribute 'values'
```
Call stack:
```
moto/core/models.py:323: in __call__
    request, request.url, request.headers
moto/core/responses.py:202: in dispatch
    return cls()._dispatch(*args, **kwargs)
moto/core/responses.py:312: in _dispatch
    return self.call_action()
moto/core/utils.py:277: in _wrapper
    response = f(*args, **kwargs)
moto/dynamodb2/responses.py:86: in call_action
    response = getattr(self, endpoint)()
moto/dynamodb2/responses.py:777: in update_item
    condition_expression=condition_expression,
moto/dynamodb2/models/__init__.py:1287: in update_item
    item.update_with_attribute_updates(attribute_updates)
moto/dynamodb2/models/__init__.py:121: in update_with_attribute_updates
    self.attrs[attribute_name] = DynamoType({"L": new_value})
moto/dynamodb2/models/dynamo_type.py:64: in __init__
    self.value = [DynamoType(val) for val in self.value]
moto/dynamodb2/models/dynamo_type.py:64: in <listcomp>
    self.value = [DynamoType(val) for val in self.value]
```

bblommers:
Fixes #3506 

The core of this issue is a lack of input validation (raised in #3231).

Until that's fixed, there are two choices when determining the data type: 
1) check which type is supplied by the user
2) check the type of the user-supplied value.

As the user-supplied value of a StringSets might be a list, we have to accept the type supplied by the user.

When proper validation is implemented, and we know 1) and 2) match up, we can refactor this entire logic flow to always accept the type supplied by the user.

### DIFF
--- a/moto/dynamodb2/models/__init__.py
+++ b/moto/dynamodb2/models/__init__.py
@@ -128,10 +128,10 @@ class Item(BaseModel):
             new_value = list(update_action["Value"].values())[0]
             if action == "PUT":
                 # TODO deal with other types
-                if isinstance(new_value, list):
-                    self.attrs[attribute_name] = DynamoType({"L": new_value})
-                elif isinstance(new_value, set):
+                if set(update_action["Value"].keys()) == set(["SS"]):
                     self.attrs[attribute_name] = DynamoType({"SS": new_value})
+                elif isinstance(new_value, list):
+                    self.attrs[attribute_name] = DynamoType({"L": new_value})
                 elif isinstance(new_value, dict):
                     self.attrs[attribute_name] = DynamoType({"M": new_value})
                 elif set(update_action["Value"].keys()) == set(["N"]):
--- a/tests/test_dynamodb2/test_dynamodb.py
+++ b/tests/test_dynamodb2/test_dynamodb.py
@@ -5628,3 +5628,24 @@ def test_transact_get_items_should_return_empty_map_for_non_existent_item():
     items.should.have.length_of(2)
     items[0].should.equal({"Item": item})
     items[1].should.equal({})
+
+
+@mock_dynamodb2
+def test_dynamodb_update_item_fails_on_string_sets():
+    dynamodb = boto3.resource("dynamodb", region_name="eu-west-1")
+    client = boto3.client("dynamodb", region_name="eu-west-1")
+
+    table = dynamodb.create_table(
+        TableName="test",
+        KeySchema=[{"AttributeName": "record_id", "KeyType": "HASH"},],
+        AttributeDefinitions=[{"AttributeName": "record_id", "AttributeType": "S"},],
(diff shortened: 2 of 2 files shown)


spulec (approved):

bpandola:
@bblommers I took a look at #3231 some time ago, but abandoned it as the scope of the work and size of the changeset started to grow a bit unwieldy.  I'll try to take another look, unless you're actively working on it.

bblommers:
@bpandola I haven't looked at a solution at all, actually, so it's all yours if you want to :)

PR #3522, opened 2020-12-06
Pass the default ECS cluster and raise accurate exceptions

jmsanders:
There are two fundamental changes to behavior in this PR that brings the Mock ECS implementation more in line with the behavior of real ECS:

- For many actions, the "default" cluster is now passed if a cluster is not provided. Most notably, this (correctly) changes the behavior of Mock ECS's ListTasks action to only list tasks for a single cluster.
- Many exceptions have been changed from custom moto exceptions to mirror the actual ECS exception.

I expect that both of these could represent breaking changes to users' tests - but they're changes that better represent the actual behavior of ECS.

This does not provide full coverage of exceptions. In general, I ran these operations against actual ECS resources and cross-referenced the documentation to figure out what actual exceptions should be thrown and what the messages should be. Consequently, I didn't update any exceptions that took more than trivial amount of time to reproduce with real resources.

There's also a lot of opportunity for refactoring of both existing tests and implementation code and to add some additional test coverage - in general, I added exceptions testing where it was easy but didn't go so far as to add tests where the action wasn't already tested (for example, PutAttributes) or where the change was relatively trivial (for example, switching from throwing `"{0} is not a cluster"` to `ClusterNotFoundException`)

### DIFF
--- a/moto/ecs/exceptions.py
+++ b/moto/ecs/exceptions.py
@@ -5,11 +5,9 @@ from moto.core.exceptions import RESTError, JsonRESTError
 class ServiceNotFoundException(RESTError):
     code = 400
 
-    def __init__(self, service_name):
+    def __init__(self):
         super(ServiceNotFoundException, self).__init__(
-            error_type="ServiceNotFoundException",
-            message="The service {0} does not exist".format(service_name),
-            template="error_json",
+            error_type="ServiceNotFoundException", message="Service not found."
         )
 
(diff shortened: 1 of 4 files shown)


bblommers (approved):
LGTM. Thanks for the PR, @jmsanders!

bblommers:
This is now part of moto >= 1.3.16.dev176

PR #3527, opened 2020-12-08
Enabled should be lower-cased

pavel-puchkin:
As per sample response: https://docs.aws.amazon.com/sns/latest/api/API_GetEndpointAttributes.html


bblommers (changes requested):
Thanks for providing a PR for this, @pavel-puchkin! Would you mind adding an assertion to our tests, to verify this value is coming through correctly? Happy to merge afterwards

pavel-puchkin:
Closing this in flavor of #3543

PR #3528, opened 2020-12-09
test_stepfunctions: Fix failures under non-UTC timezone

andersk:
These tests were passing with `TZ=UTC`, but under a non-UTC timezone they were failing:

```
E       AssertionError: given
E       X = [{'timestamp': datetime.datetime(2020, 1, 1, 0, 0, tzinfo=tzutc()), …}, …]
E           and
E       Y = [{'timestamp': datetime.datetime(2020, 1, 1, 0, 0, tzinfo=tzlocal()), …}, …]
E       X[0]['timestamp'] != Y[0]['timestamp']
```

With this fix, they pass either way.

Cc @ciaranevans, who introduced these tests in #3507.

### DIFF
--- a/tests/test_stepfunctions/test_stepfunctions.py
+++ b/tests/test_stepfunctions/test_stepfunctions.py
@@ -6,7 +6,7 @@ import os
 import sure  # noqa
 import sys
 from datetime import datetime
-from dateutil.tz import tzlocal
+from dateutil.tz import tzutc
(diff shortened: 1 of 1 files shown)


ciaranevans:
Woops! 😅  Thanks @andersk !

bblommers (approved):
LGTM. Thanks for the PR, @andersk!

PR #3529, opened 2020-12-09
Leave the global responses configuration alone

### LINKED ISSUE #3264: moto should not take over `responses` defaults
We are using moto to mock AWS services (namely S3); and responses to control our access to third party APIs.

However, moto unconditionnally configures a global pass-through in `responses`:
At import time, `moto` configures responses:
https://github.com/spulec/moto/blob/master/moto/core/models.py#L249-L252

Once moto has been imported, it is no longer possible to use `responses` in one's code.

Since moto uses `@mock_XXX` idioms, it should only apply its configuration on code that is wrapped by such decorators/context managers.

andersk:
Instead of modifying `responses._default_mock`, create our own `responses.RequestsMock` object that we can modify as needed without interfering with other users of the `responses` library.

Fixes #3264.

### DIFF
--- a/moto/apigateway/models.py
+++ b/moto/apigateway/models.py
@@ -41,6 +41,7 @@ from .exceptions import (
     RestAPINotFound,
     ModelNotFound,
 )
+from ..core.models import responses_mock
 
 STAGE_URL = "https://{api_id}.execute-api.{region_name}.amazonaws.com/{stage_name}"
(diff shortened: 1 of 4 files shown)


spulec (approved):

bblommers (approved):
Thanks for the PR @andersk!

PR #3533, opened 2020-12-09
Fix:SNS Create FIFO Topic

usmangani1:
Fix:SNS Create FIFO Topic
https://github.com/localstack/localstack/issues/3336

### DIFF
--- a/moto/sns/models.py
+++ b/moto/sns/models.py
@@ -393,16 +393,16 @@ class SNSBackend(BaseBackend):
             attributes.get("FifoTopic")
             and attributes.get("FifoTopic").lower() == "true"
         ):
-            fails_constraints = not re.match(
-                r"^[a-zA-Z0-9_.-]{1,256}$", name
-            ) or not name.endswith(".fifo")
+            fails_constraints = not re.match(r"^[a-zA-Z0-9_.-]{1,256}fifo$", name)
+            msg = "Fifo Topic names must end with .fifo"
+
         else:
             fails_constraints = not re.match(r"^[a-zA-Z0-9_-]{1,256}$", name)
+            msg = "Topic names must be made up of only uppercase and lowercase ASCII letters, numbers, underscores, and hyphens, and must be between 1 and 256 characters long."
 
         if fails_constraints:
-            raise InvalidParameterValue(
-                "Topic names must be made up of only uppercase and lowercase ASCII letters, numbers, underscores, and hyphens, and must be between 1 and 256 characters long."
-            )
+            raise InvalidParameterValue(msg)
+
         candidate_topic = Topic(name, self)
         if attributes:
             for attribute in attributes:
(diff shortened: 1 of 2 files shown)


moto/sns/models.py
```diff
@@ -388,7 +388,17 @@ def update_sms_attributes(self, attrs):
...
 
     def create_topic(self, name, attributes=None, tags=None):
-        fails_constraints = not re.match(r"^[a-zA-Z0-9_-]{1,256}$", name)
+
+        if (
+            attributes.get("FifoTopic")
+            and attributes.get("FifoTopic").lower() == "true"
+        ):
+            fails_constraints = not re.match(
+                r"^[a-zA-Z0-9_.-]{1,256}$", name
```
bblommers, line 397: This regex would allow something like `name.something.fifo`, which is not valid.
A better regex would probably be `^[a-zA-Z0-9_-]{1,256}.fifo$`

tests/test_sns/test_topics_boto3.py
```diff
@@ -522,6 +522,16 @@ def test_untag_resource_error():
...
+@mock_sns
+def test_create_fifo_topic():
+    conn = boto3.client("sns", region_name="us-east-1")
+    response = conn.create_topic(
+        Name="test_topic.fifo", Attributes={"FifoTopic": "true"}
+    )
+
+    assert "TopicArn" in response
+
+
```
bblommers, line 534: Can we add a negative test as well? Just to assert that the correct error is thrown when creating a FIFO-queue without `.fifo` suffic

bblommers (changes requested):

moto/sns/models.py
```diff
@@ -388,11 +388,21 @@ def update_sms_attributes(self, attrs):
         self.sms_attributes.update(attrs)
 
     def create_topic(self, name, attributes=None, tags=None):
-        fails_constraints = not re.match(r"^[a-zA-Z0-9_-]{1,256}$", name)
+
+        if (
+            attributes.get("FifoTopic")
+            and attributes.get("FifoTopic").lower() == "true"
+        ):
+            fails_constraints = not re.match(r"^[a-zA-Z0-9_.-]{1,256}fifo$", name)
```
bblommers, line 396: Need to move the `.` inside the regex:
`r"^[a-zA-Z0-9_-]{1,256}\.fifo$"`

Should have said this earlier, but worth creating another test case that verifies that `queue.name.fifo` also throws an error. (That name would be considered valid with the current regex, but shouldn't be.)

bblommers (changes requested):

wellermiranda:
💯 

Any chance do fix it?

bblommers (changes requested):
Sorry for the delay @usmangani1, and thanks for making the changes.
Looks like there's one more test failing due to an NPE:
https://travis-ci.org/github/spulec/moto/jobs/748858407

Happy to merge once that's fixed

bblommers (approved):
LGTM - thanks @usmangani1!

bblommers:
This is now part of moto >= 1.3.16.dev220

PR #3534, opened 2020-12-10
Add glue.batch_update_partition

dkuntz2:
This change would add support for the Glue `batch_update_partition` method.

### DIFF
--- a/moto/glue/responses.py
+++ b/moto/glue/responses.py
@@ -208,6 +208,35 @@ class GlueResponse(BaseResponse):
 
         return ""
 
+    def batch_update_partition(self):
+        database_name = self.parameters.get("DatabaseName")
+        table_name = self.parameters.get("TableName")
+        table = self.glue_backend.get_table(database_name, table_name)
+
+        errors_output = []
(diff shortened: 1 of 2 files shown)


moto/glue/responses.py
```diff
@@ -208,6 +208,35 @@ def update_partition(self):
...
+
+        errors_output = []
+        for entry in self.parameters.get("Entries"):
+            part_to_update = entry["PartitionValueList"]
+            part_input = entry["PartitionInput"]
+
+            try:
+                table.update_partition(part_to_update, part_input)
+            except PartitionNotFoundException:
+                errors_output.append(
```
bblommers, line 224: Hi @dkuntz2, thanks for the PR!
Can we get a unit test for this branch as well? Would be good to assert that some batch-entries are created, but errors are returned for failing batch-entries.

bblommers (changes requested):

moto/glue/responses.py
dkuntz2: Test case added

bblommers (approved):
LGTM. Thanks again @dkuntz2!

PR #3537, opened 2020-12-10
Fix: FIFO DLQ locks message group id

irahulranjan:


moto/sqs/models.py
```diff
@@ -805,6 +805,7 @@ def receive_messages(
                     >= queue.redrive_policy["maxReceiveCount"]
                 ):
                     messages_to_dlq.append(message)
+                    queue.pending_messages.remove(message)
```
bblommers, line 808: For readability, I think it's better to just move the `queue.pending_messages.add(message)` underneath this entire if-statement.
If the message needs to be send to DLQ, it will do just that, and `continue` with the next message.
If not, we'll add it to the pending-messages queue (and mark as received, etc).

That way we don't have to explicitly remove it at all.

bblommers (changes requested):

moto/sqs/models.py
irahulranjan: done

bblommers (approved):
LGTM. Thanks for the PR @irahulranjan!

PR #3538, opened 2020-12-10
fix Fifo queue delivers duplicate message(s)

irahulranjan:


moto/sqs/models.py
```diff
@@ -478,6 +478,16 @@ def messages(self):
         ]
 
     def add_message(self, message):
+        if self.fifo_queue:
+            if self.attributes.get("ContentBasedDeduplication") == "true":
```
BeyondEvil, line 482: Since the outer `if` has no `else`, you should be able to shorten this to:

```python
if self.fifo_queue and self.attributes.get("ContentBasedDeduplication") == "true":
```

Also, have you accounted for this case (from `aws sqs create-queue help`):

```
If  you  send one message with ContentBasedDeduplication enabled
and then another message with a MessageDeduplicationId  that  is
the  same  as  the one generated for the first MessageDeduplica-
tionId , the two messages are treated as duplicates and only one
copy of the message is delivered.
```

tests/test_sqs/test_sqs.py
```diff
@@ -2283,3 +2283,88 @@ def test_send_message_fails_when_message_size_greater_than_max_message_size():
...
+def test_fifo_queue_send_multiple_duplicate_messages_after_five_minutes():
+
+    sqs = boto3.resource("sqs", region_name="us-east-1")
+    msg_queue = sqs.create_queue(
+        QueueName="test-queue-dlq.fifo",
+        Attributes={"FifoQueue": "true", "ContentBasedDeduplication": "true"},
+    )
+
+    msg_queue.send_message(MessageBody="first", MessageGroupId="1")
+    time.sleep(301)
```
BeyondEvil, line 2313: Do you really want to pause the test for 5 minutes?

Would it make more sense to have the deduplication time in `models.py` as a constant instead?

Then for testing purposes, you can mock the constant to something like 10 instead?

tests/test_sqs/test_sqs.py
```diff
@@ -2283,3 +2283,88 @@ def test_send_message_fails_when_message_size_greater_than_max_message_size():
     ex.response["Error"]["Message"].should.contain(
         "{} bytes".format(message_size_limit)
     )
+
+
+@mock_sqs
+def test_fifo_queue_send_multiple_duplicate_messages():
```
BeyondEvil, line 2289: We should be able to merge all these tests into one or two using [pytest.mark.parametrize](https://docs.pytest.org/en/stable/parametrize.html)

Also, I see one test-case missing: Different `MessageBody`'s with same `MessageDeduplicationId` should give a queue length of `1`

moto/sqs/models.py
irahulranjan: Moved the condition to a single if and added the test case

tests/test_sqs/test_sqs.py
irahulranjan: made 2 parameterized tests
kept 2 independent tests for deduplication time and sha256 deduplication id. I did this for better readability.

tests/test_sqs/test_sqs.py
irahulranjan: thanks for pointing out the approach. Did the changes.

tests/test_sqs/test_sqs.py
```diff
@@ -2286,58 +2291,58 @@ def test_send_message_fails_when_message_size_greater_than_max_message_size():
...
+    "msg_1, msg_2, dedupid_1, dedupid_2, expectedCount",
+    [
+        ("msg1", "msg1", "1", "1", 1),
+        ("msg1", "msg1", "1", "2", 2),
+        ("msg1", "msg2", "1", "1", 1),
+        ("msg1", "msg2", "1", "2", 2),
+    ],
+)
+def test_fifo_queue_deduplication_with_id(
+    msg_1, msg_2, dedupid_1, dedupid_2, expectedCount
```
BeyondEvil, line 2304: [nitpick]

`expectedCount` => `expected_count`

[/nitpick]

irahulranjan: resolved

irahulranjan:
@BeyondEvil can you please help me out with the CI failure. I have run the tests in my machine using python 2.7, 3.6, 3.7, and 3.8 and all the tests are passing.

bblommers:
Hey @irahulranjan, the test are all failing in server mode, where Moto is run in a standalone Docker container.
The `@mock.patch` has no effect there, as it wouldn't change the value of `DEDUPLICATION_TIME_IN_SECONDS` inside the Docker container.

I don't really see how we would be able to test this inside a Docker container, so we can skip it in server mode like this:
https://github.com/spulec/moto/blob/master/tests/test_sqs/test_sqs.py#L1491

irahulranjan:
@bblommers Thanks for the help.

BeyondEvil:
> Hey @irahulranjan, the test are all failing in server mode, where Moto is run in a standalone Docker container.
> The `@mock.patch` has no effect there, as it wouldn't change the value of `DEDUPLICATION_TIME_IN_SECONDS` inside the Docker container.
> 
> I don't really see how we would be able to test this inside a Docker container, so we can skip it in server mode like this:
> https://github.com/spulec/moto/blob/master/tests/test_sqs/test_sqs.py#L1491

Not exactly sure how the docker setup works, but you could change the `DEDUPLICATION_TIME_IN_SECONDS` to try and get it's value from the environment instead:

```python
DEDUPLICATION_TIME_IN_SECONDS = os.getenv('DEDUPLICATION_TIME_IN_SECONDS', 300)
```

Then the test can set the env var to `1`, either that will then get picked up by the docker runner automagically or it can be passed to the container when started.

Hope that makes sense @irahulranjan @bblommers

bblommers:
@BeyondEvil True, there are ways around it, but none of them pretty.
Changing the constant to read the environment variable means changing the code to make it more testable - which is always a code smell.
Passing the env-var to the Docker container means that value is set for the duration of the container lifetime - and we really only want to change it for the duration of a single test.

The weird travis.yml-notation is not setting the variable, but is used as a list of configurations. Travis will execute the tests once per configuration option (per python-version). That's why the [list of test suites](https://travis-ci.org/github/spulec/moto/builds/749419595) is executed twice per language, one with the env var set to True, one set to False.

bblommers (approved):
Thanks for the PR @irahulranjan!

BeyondEvil:
> @BeyondEvil True, there are ways around it, but none of them pretty.
> Changing the constant to read the environment variable means changing the code to make it more testable - which is always a code smell.

Fair enough.

> Passing the env-var to the Docker container means that value is set for the duration of the container lifetime - and we really only want to change it for the duration of a single test.

True.

It's already merged. But does moto support some type of configuration framework? I guess something like that could've been leveraged in a scenario like this.

> The weird travis.yml-notation is not setting the variable, but is used as a list of configurations. Travis will execute the tests once per configuration option (per python-version). That's why the [list of test suites](https://travis-ci.org/github/spulec/moto/builds/749419595) is executed twice per language, one with the env var set to True, one set to False.

Oh, really. Wow, that is weird. 😅 

Thanks for taking the time to explain. 🙏

bblommers:
> It's already merged. But does moto support some type of configuration framework? I guess something like that could've been leveraged in a scenario like this.
> 

No, not yet! I agree that would be useful, and will become more useful as we're getting more config options. One day.. 🤞

bblommers:
This is now part of moto >= 1.3.16.dev181

PR #3524, opened 2020-12-06
Fix:Ec2-VPC:Added functionality describe-vpc-endpoint

usmangani1:
Fix:Ec2-VPC:Added functionality describe-vpc-endpoint
https://github.com/localstack/localstack/issues/3310

PR #3539, opened 2020-12-10
Fix owner ID in describe-vpc-endpoints response.

sworisbreathing:
Refs #3524


bblommers (approved):
LGTM - thanks for the PR @sworisbreathing!

bblommers:
This is now part of moto >= 1.3.16.dev179

PR #3543, opened 2020-12-11
Enabled SNS endpoint attr must be lowercased

neoascetic:
As per sample response:
https://docs.aws.amazon.com/sns/latest/api/API_GetEndpointAttributes.html


bblommers (changes requested):
Hey @neoascetic, looks like this test is failing as well: `test_set_endpoint_attributes`. Happy to merge once Travis passes

bblommers (approved):
LGTM - Thanks!

PR #3546, opened 2020-12-13
RFC/Draft for configuration utility

BeyondEvil:
@bblommers made a [wish](https://github.com/spulec/moto/pull/3538#issuecomment-744018436) and I figured why not get the ball rolling?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,7 +24,16 @@ install:
     else
       PYTHON_DOCKER_TAG=${TRAVIS_PYTHON_VERSION}-stretch
     fi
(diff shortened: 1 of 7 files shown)


BeyondEvil:
Server tests are failing, likely due to the `configuration.ini` file missing or being in a different/wrong location.

Before I spend time debugging, maybe someone else can chime in with the correct solution to the problem? Afaict it should be present. The docker has the entire repo swallowed and I added the file to the MANIFEST. 🤷‍♂️

BeyondEvil:
Ping @spulec

spulec:
So far we have been using the settings file for this type of thing: https://github.com/spulec/moto/blob/master/moto/settings.py

Any thoughts on why this would be better? Or is this something else where it isn't designed to be configurable by the user?

BeyondEvil:
> So far we have been using the settings file for this type of thing: https://github.com/spulec/moto/blob/master/moto/settings.py
> 
> Any thoughts on why this would be better? Or is this something else where it isn't designed to be configurable by the user?

A couple of things off the top of my head:

- Provides support for different configurations between Production and Development/Test
- Uses a more idiomatic approach to configuration

Wether or not it's user-configurable is really up to us (implementation).

PR #3549, opened 2020-12-14
ENI: DescribeNetworkInterfaces `attachment.instance-id` filter

Preskton:
Related issue: #3550 

## Description

Allows consumers to filter ENIs by the `attachment.instance-id` filter, if present. Today, attempting to filter by `attachment.instance-id` yields an Exception similar to:

```
moto.ec2.exceptions.FilterNotImplementedError: The filter 'attachment.instance-id' for DescribeNetworkInterfaces has not been implemented in Moto yet. Feel free to open an issue at https://github.com/spulec/moto/issues
```

## TODO 

- [x] Open an issue
- [x] Test this actually works
- [x] Unit tests
- [x] Lint, formatting, etc.
- [x] Better description


bblommers:
Thanks for the opening a PR @Preskton! Will convert it to a draft - let me know when you think it's ready for review.

Preskton:
@bblommers since this is my first contribution, I have a couple of quick questions:

Would you mind taking a quick peek at the change and let me know if a) the change is reasonable/directionally correct, and b) if the unit test appears in line with how other items are tested?

Second, when I run `black`, there are a handful of errors outside of my current change's scope. Should I change them as part of this PR?

bblommers:
This looks like the right approach @Preskton, both in terms of the actual change and the testing.

Black shouldn't touch anything else, really, as it is already formatted in master.
Which version are you using?

Preskton:
Thanks for taking a peek, @bblommers. Turns out my py installs were absolutely wrecked. Cleaned them up and was able to get things going correctly.

I now have working tests - could you please provide feedback if this is an appropriate style? I was getting a little confused between if I should use client/resources or hit the `backend`s directly.

Also, what guidance can you provide around clearing this warning? I'd like to knock it out before I merge.

```
tests/test_ec2/test_elastic_network_interfaces.py::test_elastic_network_interfaces_get_by_attachment_instance_id
  /code/Preskton/moto/moto/ec2/models.py:567: PendingDeprecationWarning: Could not find AMI with image-id:ami-d3adb33f, in the near future this will cause an error.
  Use ec2_backend.describe_images() to find suitable image for your test
    warnings.warn(
```

Thanks in advance!

tests/test_ec2/test_elastic_network_interfaces.py
```diff
@@ -401,6 +401,43 @@ def test_elastic_network_interfaces_get_by_description():
     enis.should.have.length_of(0)
 
 
+@mock_ec2_deprecated
+@mock_ec2
+def test_elastic_network_interfaces_get_by_attachment_instance_id():
+    conn = boto.connect_vpc("the_key", "the_secret")
```
bblommers, line 407: Creating the VPC can be done with the boto3-client as well.

`@mock_ec2_deprecated' makes it possible to mock boto, but were trying to phase out support for boto at some point. So new tests shouldn't have any `@.._deprecated` decorators.

tests/test_ec2/test_elastic_network_interfaces.py
```diff
@@ -401,6 +401,43 @@ def test_elastic_network_interfaces_get_by_description():
...
+    subnet = conn.create_subnet(vpc.id, "10.0.0.0/18")
+    ec2_resource = boto3.resource("ec2", region_name="us-east-1")
+    ec2_client = boto3.client("ec2", region_name="us-east-1")
+
+    security_group1 = conn.create_security_group(
+        "test security group #1", "this is a test security group"
+    )
+
+    create_instances_result = ec2_resource.create_instances(
+        ImageId="ami-d3adb33f", MinCount=1, MaxCount=1
```
bblommers, line 418: The warning pops up for quite a few tests - I think it's to do with the fact that this ImageId does not exist in our mocked world.
Don't worry this for now - we'll clean all those warnings up in one go (at some point..)

bblommers (changes requested):

tests/test_ec2/test_elastic_network_interfaces.py
Preskton: Do you have an example where we're using the new style and attaching an ENI? The only other ENI attach example I could find was using the deprecated style.

bblommers: It is be very similar using boto3. Haven't tested it out, but this should work:
```
eni1 = ec2_client.create_network_interface(SubnetId=..., Groups=[..])
ec2_client.attach_network_interface(DeviceIndex=1, InstanceId=..., NetworkInterfaceId=...)
```

bblommers:
Hope you don't mind me touching this PR @Preskton - I just fixed up the test to get it ready

PR #3552, opened 2020-12-16
IoT: Added Rules implementation

sim1234:


tests/test_iot/test_iot.py
```diff
@@ -1991,3 +1991,133 @@ def test_list_job_executions_for_thing():
...
+    # errors
+
+    # get not existing rule
+    try:
+        client.get_topic_rule(ruleName=name)
+    except ClientError as exc:
+        error_code = exc.response["Error"]["Code"]
+        error_code.should.equal("ResourceNotFoundException")
+    else:
+        raise Exception("Should have raised error")
```
bblommers, line 2077: Error handling should follow this format:
```
with pytest.raises(..):
    ....
````
Like it's done in an earlier test case: https://github.com/spulec/moto/blob/master/tests/test_iot/test_iot.py#L607

bblommers (changes requested):
Thanks for the PR @sim1234! Would you mind splitting up the tests into multiple test cases? Ideally each method should only test a single functionality.

tests/test_iot/test_iot.py
sim1234: @bblommers I've updated the tests implementation. Please check it out.

bblommers (changes requested):
Looks like this is failing in Python2:
https://travis-ci.org/github/spulec/moto/jobs/754133587

Date formatting is usually done like this in Moto, to ensure it works in both Py2 and Py3:
`datetime.datetime.utcnow().strftime("%Y-%m-%d %H:%M:%S")`

bblommers (approved):
LGTM. Thanks again, @sim1234!

bblommers:
This is now part of moto >= 1.3.16.dev201

PR #3555, opened 2020-12-16
Tech Debt - Remove unknown ImageID in internal tests

bblommers:
Unknown image ID's would show this warning:
`PendingDeprecationWarning: Could not find AMI with image-id:ami-123456, in the near future this will cause an error.\nUse ec2_backend.describe_images() to find suitable image for your test`

The build log was completely full of warnings like this.
Not all warnings are gone with this change, but it's a start.


spulec (approved):

bblommers:
This is now part of moto >= 1.3.16.dev197

PR #3556, opened 2020-12-17
Run tests in parallel

bblommers:
Improve test performance by running tests in parallel.
(Server mode still runs sequentially, as the reset()-method is not threadsafe)

Runs 4 processes/cores in parallel, [as suggested by Travis](https://docs.travis-ci.com/user/common-build-problems/#parallel-processes).

(TravisCI has been a little finicky lately, so this will probably fail initially.. but at least it fails quickly? ¯\_(ツ)_/¯)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -28,7 +28,7 @@ test-only:
 test-coverage:
 	rm -f .coverage
 	rm -rf cover
-	pytest -sv --cov=moto --cov-report xml ./tests/ $(TEST_EXCLUDE)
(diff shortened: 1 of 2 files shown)


bblommers:
Note to self: Travis fails for Python3.6 because of an old `pytest` plugin. Has 5.4.2 installed, but needs to be >= 6.0.0

bblommers:
Going to close this, as this will in practice not reduce the total test run time until we fix the ServerMode tests

PR #3560, opened 2020-12-20
Fix:Added nat-gateway tags

usmangani1:
Fix:Added nat-gateway tags

https://github.com/localstack/localstack/issues/3377


bblommers (approved):
LGTM. Thanks @usmangani1!

PR #3562, opened 2020-12-22
Add get_function_configuration support for Lambda

enoodle:
This adds support for `boto3.client('lambda').get_function_configuration(FunctionName='blalbla')`

### DIFF
--- a/moto/awslambda/responses.py
+++ b/moto/awslambda/responses.py
@@ -131,6 +131,8 @@ class LambdaResponse(BaseResponse):
         self.setup_class(request, full_url, headers)
         if request.method == "PUT":
             return self._put_configuration(request)
+        if request.method == "GET":
+            return self._get_function_configuration(request, full_url, headers)
         else:
             raise ValueError("Cannot handle request")
 
@@ -300,6 +302,15 @@ class LambdaResponse(BaseResponse):
         else:
             return 404, {}, "{}"
 
+    @staticmethod
+    def _update_function_configuration(configuration, qualifier):
+        configuration = {k: v for k, v in configuration.items()}
+        if qualifier is None or qualifier == "$LATEST":
+            configuration["Version"] = "$LATEST"
+        if qualifier == "$LATEST":
+            configuration["FunctionArn"] += ":$LATEST"
+        return configuration
+
     def _get_function(self, request, full_url, headers):
         function_name = unquote(self.path.rsplit("/", 1)[-1])
         qualifier = self._get_param("Qualifier", None)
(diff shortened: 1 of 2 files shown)


moto/awslambda/responses.py
```diff
@@ -300,6 +302,15 @@ def _delete_function(self, request, full_url, headers):
         else:
             return 404, {}, "{}"
 
+    @staticmethod
+    def _update_function_configuration(configuration, qualifier):
```
bblommers, line 306: The function name is a bit misleading, as it doesn't update - only copy and rename.

moto/awslambda/responses.py
```diff
@@ -300,6 +302,15 @@ def _delete_function(self, request, full_url, headers):
...
+        configuration = {k: v for k, v in configuration.items()}
```
bblommers, line 307: Do we need the copy? As far as I can tell, the entire `configuration` object is created from scratch when calling `fn.get_code()`

tests/test_awslambda/test_lambda.py
```diff
@@ -506,6 +506,70 @@ def test_get_function():
...
+    result["Environment"].should.contain("Variables")
+    result["Environment"]["Variables"].should.equal({"test_variable": "test_value"})
+
+    # Test get function with qualifier
+    result = conn.get_function_configuration(
+        FunctionName="testFunction", Qualifier="$LATEST"
+    )
+    result["Version"].should.equal("$LATEST")
+    result["FunctionArn"].should.equal(
+        "arn:aws:lambda:us-west-2:{}:function:testFunction:$LATEST".format(ACCOUNT_ID)
```
bblommers, line 565: The region can also be formatted, with `_lambda_region`

tests/test_awslambda/test_lambda.py
```diff
@@ -506,6 +506,70 @@ def test_get_function():
...
+        Code={"S3Bucket": "test-bucket", "S3Key": "test.zip"},
+        Description="test lambda function",
+        Timeout=3,
+        MemorySize=128,
+        Publish=True,
+        Environment={"Variables": {"test_variable": "test_value"}},
+    )
+
+    result = conn.get_function_configuration(FunctionName="testFunction")
+    # this is hard to match against, so remove it
```
bblommers, line 537: Don't think we need to remove these keys, as we're not verifying equality on the entire dict.

bblommers (changes requested):
Thanks for your PR @enoodle! Left a few comments.

moto/awslambda/responses.py
enoodle: :+1:

moto/awslambda/responses.py
enoodle: what about `_set_configuration_qualifier` ?

bblommers: Sounds good!

bblommers (changes requested):
Thanks for making the changes @enoodle. Looks like the linter fails atm (https://travis-ci.org/github/spulec/moto/builds/753454806) - happy to merge once that's resolved.

enoodle:
@bblommers I fixed the linting issue.

bblommers (approved):
LGTM!

Issue #3188, opened 2020-07-27
Support lambda layers


benkeil:
Version: master

It looks like lambda layers are not supported. Already works with aws sam: https://github.com/awslabs/serverless-application-model

bblommers:
This is now part of moto >= 1.3.16.dev207

PR #3563, opened 2020-12-23
adding layers support for lambda backend

enoodle:
This adds basic support for layers to the lambda resource.
Adding support for `boto3.client('lambda').publish_layer_version(...)` and `boto3.client('lambda').list_layer_versions(LayerName='')` and also supporting Layer information passed to functions.

It does **not** implement all the layers API and it is **not** adding the layers to the lambda invocation simulation.

Fixes: #3188

### DIFF
--- a/moto/awslambda/models.py
+++ b/moto/awslambda/models.py
@@ -287,7 +287,9 @@ class Layer(object):
         return {
             "LayerName": self.name,
             "LayerArn": self.layer_arn,
-            "LatestMatchingVersion": self.layer_versions[str(self._latest_version)].get_layer_version()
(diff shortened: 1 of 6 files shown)


bblommers (changes requested):
Hi @enoodle, thanks for the PR!

Could you add a CloudFormation-test as well, to verify that `create_from_cloudformation_json` works as expected?

Looks like the build is failing as well: https://travis-ci.org/github/spulec/moto/builds/753836363

enoodle:
@bblomers I have added those tests and fixed the formatting issues.

bblommers (approved):
LGTM - thanks again @enoodle!

PR #3578, opened 2021-01-03
Add launchType parameter to ECS

### LINKED ISSUE #3551: launchType in ECS create_service
In `moto/ecs/responses.py` the function `create_services` doesn't pickup the launchType parameter.  As a result, the `describe_service` operation doesn't return the launchType passed during creation.

gruebel:
Fixes #3551 

I also implemented the missing logic for launchType to `create_task_set`. And I changed the tests I touched to use the `should` syntax instead of `asserts`.


bblommers (approved):
LGTM. Thanks @gruebel!

PR #3579, opened 2021-01-03
Add validation to put_events

### LINKED ISSUE #3384: Add request validation to EventBridge
EventBridge is supported using @moto.mock_events, however, the requests are not validated, and so you end up not receiving errors, when sending in incorrect requests, which, as an example, omit the 'Source' argument in the put_events request

gruebel:
Fixes #3384 

It is a quite interesting behaviour of the `put_events` endpoint, because each event is validated separately and the status is send back for all of them.

### DIFF
--- a/moto/events/exceptions.py
+++ b/moto/events/exceptions.py
@@ -1,0 +1,9 @@
+from __future__ import unicode_literals
+from moto.core.exceptions import JsonRESTError
+
+
+class ValidationException(JsonRESTError):
+    code = 400
+
+    def __init__(self, message):
+        super(ValidationException, self).__init__("ValidationException", message)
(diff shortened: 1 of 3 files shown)


bblommers (changes requested):
Hey @gruebel, would you mind splitting up the test scenarios in multiple methods? 
Would be good to also test a scenario with multiple events, some valid, some invalid.

gruebel:
I also removed the weird `test_rule_tagging_sad` test and added the correct way of testing them.

bblommers (changes requested):
Looks like it's failing in Py2:
`AttributeError: 'module' object has no attribute 'JSONDecodeError'`

gruebel:
@bblommers is there actually a plan to drop Python 2 support at some point?

bblommers:
There's no plan yet, @gruebel. There's still a significant portion of people that are stuck with Py2, and we want to keep Moto available to them as long as possible.
Thanks for the PR!

bblommers (approved):

PR #3580, opened 2021-01-04
Bump cryptography library

daveygit2050:
From Python Safety:

Cryptography 3.3 no longer allows loading of finite field Diffie-Hellman parameters of less than 512 bits in length. This change is to conform with an upcoming OpenSSL release that no longer supports smaller sizes. These keys were already wildly insecure and should not have been used in any application outside of testing.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -34,7 +34,7 @@ def get_version():
 install_requires = [
     "boto3>=1.9.201",
     "botocore>=1.12.201",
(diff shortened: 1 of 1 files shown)


bblommers (approved):
Thanks @daveygit2050!

bblommers:
This is now part of moto >= 1.3.16.dev190

PR #3582, opened 2021-01-06
Fix ResourceNotFoundError exception message

mallamanis:
From `None` to the `identity_pool_id`.


bblommers (changes requested):
Hi @mallamanis, thanks for the PR! Would you mind adding a unit test for this? Happy to merge after that

mallamanis:
@bblommers Done. I updated an existing test that was already there. Note that the test previously had an incorrect indentation for the fine-grained asserts of `pytest.raises`, which is also fixed now.

bblommers (approved):
LGTM - Thanks @mallamanis!

PR #3586, opened 2021-01-10
Fix MessageNotInflight error

### LINKED ISSUE #3435: botocore.parsers.ResponseParserError changing visibility on message not in flight
How to reproduce the bug:

Run moto docker in server mode (I used the latest image at https://hub.docker.com/r/motoserver/moto)

Create a queue, send a message, receive the message, and try to set the visibility on the (now no longer valid) message:

`queue = boto3.resource('sqs', endpoint_url='http://moto:5000').create_queue(QueueName='test')`
`msg = base64.b64encode(json.dumps({'test': 'test'}).encode('utf-8')).decode('utf-8')`
`queue.send_message(MessageBody=msg)`
`sqs_msg = queue.receive_messages()[0]`
`sqs_msg.change_visibility(VisibilityTimeout=1)`

Expected:
`botocore.exceptions.ClientError`

What happens:
`*** botocore.parsers.ResponseParserError: Unable to parse response (syntax error: line 1, column 0), invalid XML received. Further retries may succeed:
b'Invalid request: The message referred to is not in flight.'`

Possible cause:
MessageNotInflight is the only exception in sqs/exceptions.py that does not inherit from RESTError. If I do make it inherit from RESTError, the issue goes away.

Version: Server mode using https://hub.docker.com/r/motoserver/moto latest (digest ef4e4172daf2) using docker-compose on the default port (5000).

gruebel:
Fixes #3435 

I'm not sure, if the error text is correct, because I couldn't reproduce the error. It is documented to be a valid error, but no idea, how to actually force it.

### DIFF
--- a/moto/sqs/exceptions.py
+++ b/moto/sqs/exceptions.py
@@ -2,9 +2,13 @@ from __future__ import unicode_literals
 from moto.core.exceptions import RESTError
 
 
-class MessageNotInflight(Exception):
-    description = "The message referred to is not in flight."
+class MessageNotInflight(RESTError):
     status_code = 400
+
+    def __init__(self):
+        super(MessageNotInflight, self).__init__(
+            "MessageNotInflight", "The message referred to is not in flight."
+        )
 
 
 class ReceiptHandleIsInvalid(RESTError):
--- a/moto/sqs/responses.py
+++ b/moto/sqs/responses.py
@@ -114,17 +114,11 @@ class SQSResponse(BaseResponse):
         except ValueError:
             return ERROR_MAX_VISIBILITY_TIMEOUT_RESPONSE, dict(status=400)
 
-        try:
-            self.sqs_backend.change_message_visibility(
-                queue_name=queue_name,
-                receipt_handle=receipt_handle,
-                visibility_timeout=visibility_timeout,
-            )
-        except MessageNotInflight as e:
-            return (
-                "Invalid request: {0}".format(e.description),
-                dict(status=e.status_code),
-            )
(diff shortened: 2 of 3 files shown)


bpandola:
@gruebel Just so know you're not alone:  I looked into this issue back in November and I simply could not find any way to reproduce the error on a real AWS backend.

bblommers:
Hmm.. not a fan of unreproducible bugs :) 

What is the AWS output for this scenario?

gruebel:
@bblommers nothing just a 200 status response. You can invoke the `change_message_visibility` method multiple times with the same `ReceiptHandle` and you always get positive response.

bblommers:
I'm checking with the user who raised this on how to reproduce the error, will mark this as draft to prevent it from being merged until we know for sure what the desired behaviour is.

wallowa:
Thank you for looking into addressing this. I posted an update in the original bug report: https://github.com/spulec/moto/issues/3435. As noted by @gruebel, AWS allows calling change_message_visibility multiple times (you can also delete a message multiple times).

bblommers:
From what I can see, we shouldn't raise the error here: 
https://github.com/spulec/moto/blob/master/moto/sqs/models.py#L884

The following method will fail if we remove that if-statement: `test_change_message_visibility_on_visible_message`
However - running this test against AWS itself also fails as it stands, as the SQSError is not thrown in the last assert. So I guess we can remove that test alltogether

Haven't been able to test the idempotent delete-message use case yet.

bblommers:
I've (finally..) verified the behaviour against AWS - this is now fixed in #4655

PR #3591, opened 2021-01-16
IAM - Delete Role/InstanceProfile via CloudFormation

### LINKED ISSUE #3583: Feature Request: IAM: Delete InstanceProfile and Role via CF
Both the InstanceProfile and Roles cannot be deleted via CloudFormation at the moment.

Needed to do some debugging for another issue when this popped up, so the simplest implementation would be:

In class InstanceProfile:
```
@classmethod
def delete_from_cloudformation_json(
    cls, resource_name, cloudformation_json, region_name
):
    iam_backend.delete_instance_profile(resource_name)

def delete_role(self, role_name):
    self.roles = [role for role in self.roles if role.name != role_name]
```

In class Role:
```
@classmethod
def delete_from_cloudformation_json(
    cls, resource_name, cloudformation_json, region_name
):
    role_name = cloudformation_json["Properties"]["RoleName"]
    for profile_name, profile in iam_backend.instance_profiles.items():
        profile.delete_role(role_name=role_name)

    iam_backend.delete_role(role_name)
```

bblommers:
Fixes #3583

### DIFF
--- a/moto/cloudformation/parsing.py
+++ b/moto/cloudformation/parsing.py
@@ -347,8 +347,9 @@ def parse_and_update_resource(logical_id, resource_json, resources_map, region_n
         return None
 
 
-def parse_and_delete_resource(resource_name, resource_json, resources_map, region_name):
-    resource_class, resource_json, _ = parse_resource(resource_json, resources_map)
+def parse_and_delete_resource(resource_name, resource_json, region_name):
+    resource_type = resource_json["Type"]
+    resource_class = resource_class_from_type(resource_type)
     if not hasattr(
         resource_class.delete_from_cloudformation_json, "__isabstractmethod__"
     ):
@@ -663,9 +664,7 @@ class ResourceMap(collections_abc.Mapping):
                 ].physical_resource_id
             else:
                 resource_name = None
-            parse_and_delete_resource(
-                resource_name, resource_json, self, self._region_name
-            )
+            parse_and_delete_resource(resource_name, resource_json, self._region_name)
             self._parsed_resources.pop(logical_name)
 
         self._template = template
@@ -726,7 +725,7 @@ class ResourceMap(collections_abc.Mapping):
                             ]
 
                             parse_and_delete_resource(
-                                resource_name, resource_json, self, self._region_name,
+                                resource_name, resource_json, self._region_name,
                             )
 
                         self._parsed_resources.pop(parsed_resource.logical_resource_id)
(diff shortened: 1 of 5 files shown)


moto/iam/models.py
```diff
@@ -633,7 +646,7 @@ def delete_policy(self, policy_name):
 
     @property
     def physical_resource_id(self):
-        return self.id
+        return self.name
```
bblommers, line 649: This was an (unrelated) bug in the implementation. Verified it against AWS itself - the physical resource id is always the role name

moto/cloudformation/parsing.py
```diff
@@ -343,8 +343,9 @@ def parse_and_update_resource(logical_id, resource_json, resources_map, region_n
         return None
 
 
-def parse_and_delete_resource(resource_name, resource_json, resources_map, region_name):
-    resource_class, resource_json, _ = parse_resource(resource_json, resources_map)
```
bblommers, line 347: The resources_map is a LazyDict, where resources are generated on the fly. If a resource is referenced that doesn't exist yet, it is automatically created. This lead to the workflow:
 - everything is created
 - call delete_stack
 - Delete the role
 - Start deleting the InstanceProfile (that references the role)
 - Parse the map to get all info from the InstanceProfile
 - Recreate the referenced role, as it no longer exist..
 - Finish deleting the InstanceProfile

Which lead to the fun bug where the role was deleted, but still existed afterwards.

Dict ordering ensured that it only occured in Python2. In Python3, it would delete the resources alphabetically; the InstanceProfile first, and the Role afterwards.

bblommers: Note that none of the implementations actually use this map, as far as I know, so the fact that it's now unparsed is irrelevant

moto/iam/models.py
```diff
@@ -553,6 +553,19 @@ def create_from_cloudformation_json(
 
         return role
 
+    @classmethod
+    def delete_from_cloudformation_json(
+        cls, resource_name, cloudformation_json, region_name
+    ):
+        for profile_name, profile in iam_backend.instance_profiles.items():
+            profile.delete_role(role_name=resource_name)
```
bblommers, line 561: A minor nitpick on this implementation:
In an ideal world, the InstanceProfile#delete would read the `cloudformation_json`, get the roles that are attached to it, and delete the role only from our InstanceProfile.
Same for policies.

The current implementation could lead to a scenario where if another profile is created outside of CF, the attached role will still be deleted. AWS would throw an error in that scenario, for trying to delete a role attached to an an existing profile.

PR #3605, opened 2021-01-22
Fix ec2 filter by empty tag value

### LINKED ISSUE #3603: Empty tags are not considered on filtering
I've been creating AMIs with tags such as RELEASE with an empty value (I just don't add anything on the console or an empty string "" in boto3/moto)

After that, I filter using those tags. The problem I've found is that when using MOTO to mock boto3 calls the describe_iamges() behaves differently compared to boto3.

When I try to filter with a key-value where the value is empty ("") the filter is not taken into account. So instead of filtering only those images that have a RELEASE tag with an empty value, all images are returned whether they have this filter or not.

I've verified that this is not the behavior on boto3 with a real AWS account. And is not. When using boto3 only those AMIs with the value "" are returned. This only happens when using empty values in the filtering, when using any other values the correct AMIs are returned.

Some code to try to reproduce
````

@pytest.fixture(scope='function')
def aws_credentials():
    os.environ['AWS_ACCESS_KEY_ID'] = 'testing'
    os.environ['AWS_SECRET_ACCESS_KEY'] = 'testing'
    os.environ['AWS_SECURITY_TOKEN'] = 'testing'
    os.environ['AWS_SESSION_TOKEN'] = 'testing'
    os.environ['AWS_DEFAULT_REGION'] = 'eu-west-1'


@pytest.fixture(scope='function')
def ec2(aws_credentials):
    with mock_ec2():
        yield boto3.client('ec2', region_name=os.getenv('AWS_DEFAULT_REGION'))

def fake_images_release_list(ec2):
    fake_images = []

    instances = ec2.run_instances(ImageId='foo', MinCount=1, MaxCount=1)["Instances"]

    # create images from instance
    for i in range(0, 9):
        with freeze_time(f"2021-01-01 08:00:0{i}"):
            image = ec2.create_image(InstanceId=instances[0]["InstanceId"],
                                     Name=f"MyFooAMI{i}",
                                     Description="Test")

            response = ec2.create_tags(Resources=[image["ImageId"]],
                                       Tags=[
                                           {
                                               'Key': 'Base_AMI_Name',
                                               'Value': 'Deep Learning Base AMI (Amazon Linux 2) Version 31.0',
                                           },
                                           {
                                               'Key': 'OS_Version',
                                               'Value': 'AWS Linux 2',
                                           }
                                       ])
            fake_images.append(image)

    # Add release tags to some of the images in the middle
    for image in fake_images[3:6]:
            response = ec2.create_tags(Resources=[image["ImageId"]],
                                       Tags=[{'Key': 'RELEASE', 'Value': ''}])

    return fake_images


def test_get_latest_image_tag_as_release3(ec2):
    fake_images = fake_images_release_list(ec2)

    filter = [{"Name": 'tag:Base_AMI_Name', "Values": ['Deep Learning Base AMI (Amazon Linux 2) Version 31.0']},
              {"Name": 'tag:OS_Version', "Values": ['AWS Linux 2']},
              {"Name": 'tag:RELEASE', "Values": ['']}]

    assert len(ec2.describe_images(Owners=['self'], Filters=filter)["Images"]) == 3
````

Instead of the 3 images mark as release in the middle, the 9 are amis are returned

### PR DESCRIPTION
**Pull request by bpandola**
Return `None` instead of an empty string when the tag key does not exist
and replace the falsy check with a more explicit `is None`, which allows
empty string values to correctly pass through the filter comparator.

Behavior confirmed against a real AWS backend.

Closes #3603

### DIFF
--- a/tests/test_ec2/test_amis.py
+++ b/tests/test_ec2/test_amis.py
@@ -876,3 +876,27 @@ def test_ami_snapshots_have_correct_owner():
 
         for snapshot in snapshots_rseponse["Snapshots"]:
             assert owner_id == snapshot["OwnerId"]
+
+
+@mock_ec2
+def test_ami_filter_by_empty_tag():
+    ec2 = boto3.resource("ec2", region_name="us-west-1")
+    client = boto3.client("ec2", region_name="us-west-1")
+
+    instance = ec2.create_instances(ImageId=EXAMPLE_AMI_ID, MinCount=1, MaxCount=1)[0]
+    image_id = client.create_image(
+        InstanceId=instance.instance_id, Name="test-image", Description="test ami",
+    )["ImageId"]
+    ec2.create_tags(
+        Resources=[image_id], Tags=[{"Key": "empty", "Value": ""}],
+    )
+
+    resp = client.describe_images(Filters=[{"Name": "tag-key", "Values": ["empty"]}])
+    assert len(resp["Images"]) == 1
+    assert resp["Images"][0]["ImageId"] == image_id
+
+    resp2 = client.describe_images(Filters=[{"Name": "tag:empty", "Values": [""]}])
+    assert len(resp["Images"]) == 1
+    assert resp["Images"][0]["ImageId"] == image_id
+
+    assert resp["Images"] == resp2["Images"]
--- a/tests/test_ec2/test_amis.py
+++ b/tests/test_ec2/test_amis.py
@@ -879,30 +879,6 @@ def test_ami_snapshots_have_correct_owner():
 
 
 @mock_ec2
-def test_ami_filter_by_empty_tag():
-    ec2 = boto3.resource("ec2", region_name="us-west-1")
-    client = boto3.client("ec2", region_name="us-west-1")
-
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: tests/test_ec2/test_amis.py [hunk 1]
```diff
@@ -878,6 +878,30 @@ def test_ami_snapshots_have_correct_owner():
...
 
+@mock_ec2
+def test_ami_filter_by_empty_tag():
+    ec2 = boto3.resource("ec2", region_name="us-west-1")
+    client = boto3.client("ec2", region_name="us-west-1")
+
+    instance = ec2.create_instances(ImageId=EXAMPLE_AMI_ID, MinCount=1, MaxCount=1)[0]
+    image_id = client.create_image(
+        InstanceId=instance.instance_id, Name="test-image", Description="test ami",
+    )["ImageId"]
```
bblommers, line 889: Can you create a few images, and verify that we're only returning the one with the tags applied?
The fix looks alright to me, just want to make sure that we're explicitly testing the error case that was supplied:

> So instead of filtering only those images that have a RELEASE tag with an empty value, all images are returned whether they have this filter or not.

bblommers (changes requested):

bblommers (approved):
LGTM - thanks @bpandola!

Issue #3609, opened 2021-01-23
mock_redshift get_cluster_credentials support


dreamorosi:
On moto 1.3.16 the [Redshift `get_cluster_credentials()`](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/redshift.html#Redshift.Client.get_cluster_credentials) endpoint is not implemented.

As detailed in the documentation, the method:
> Returns a database user name and temporary password with temporary authorization to log on to an Amazon Redshift database.

and has signature:
```py
response = client.get_cluster_credentials(
    DbUser='string',
    DbName='string',
    ClusterIdentifier='string',
    DurationSeconds=123,
    AutoCreate=True|False,
    DbGroups=[
        'string',
    ]
)
```

My proposal is to implement the method in the following way:
* `ClusterIdentifier` parameter must pass validation against existing clusters, if not `ClusterNotFoundError` is raised.
* `DurationSeconds` must have a value between `900` and `3600`, if not `InvalidParameterValueError` is raised.
* `DbUser` can have any value, no validation on the value is applied; this is consistent with the moto's current implementation of `create_cluster`.
* `AutoCreate` default value is `False` (undocumented but tested against real Redshift Cluster);
*  * if `AutoCreate==False`; then returned `DbUser` is prefixed with `IAM:`.
*  * if `AutoCreate==True`; then returned `DbUser` is prefixed with `IAMA:`.
* `DbGroups` and `DbName` parameters are ignored.
* The password generated is an alphanumeric string with no actual value and length of 32 characters.

If the above is okay I am happy to open a PR.

bblommers:
Sounds good to me @dreamorosi - PR's are always welcome!

bblommers:
This is now part of moto >= 1.3.16.dev215

PR #3611, opened 2021-01-23
Added redshift.get_cluster_credentials

dreamorosi:
Implemented as described in #3609.

### DIFF
--- a/moto/redshift/models.py
+++ b/moto/redshift/models.py
@@ -1,6 +1,8 @@
 from __future__ import unicode_literals
 
 import copy
+import random
+import string
 import datetime
 
 from boto3 import Session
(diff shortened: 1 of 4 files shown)


dreamorosi:
@bblommers Latest CI build failed for Python 2.7 on already existing tests. The ones that I have added in the PR are all passed.
Thoughts?

moto/redshift/models.py
```diff
@@ -941,6 +943,36 @@ def delete_tags(self, resource_name, tag_keys):
         resource = self._get_resource_from_arn(resource_name)
         resource.delete_tags(tag_keys)
 
+    @staticmethod
+    def _generate_random_alphanum():
```
bblommers, line 947: Let's move this into utilities.utils.py, there's a similar method there already - makes sense to have them in the same place.

moto/redshift/models.py
```diff
@@ -941,6 +943,36 @@ def delete_tags(self, resource_name, tag_keys):
...
+        printable = list(string.ascii_letters + string.digits + string.punctuation)
```
bblommers, line 948: Minor nitpick - but the method name (alphanum) is incorrect if punctuation is allowed

moto/redshift/models.py
```diff
@@ -941,6 +943,36 @@ def delete_tags(self, resource_name, tag_keys):
...
+
+        random_alphanum = random.sample(printable, k=32)
+        random_alphanum = "".join(random_alphanum)
+        return random_alphanum
+
+    def get_cluster_credentials(
+        self, cluster_identifier, db_user, auto_create, duration_seconds
+    ):
+        try:
+            assert duration_seconds >= 900 and duration_seconds <= 3600
```
bblommers, line 959: This can be a simple `if ..: raise `, no need for a try-except

bblommers (changes requested):
Thanks for the PR @dreamorosi - I've left some comments.

The failure in Py2 looks to be caused by #3613, we'll get that sorted independently.

moto/redshift/models.py
dreamorosi: True, my bad 😅

moto/redshift/models.py
dreamorosi: I removed it entirely and now using the existing one; I didn't see it at first, thanks for pointing it out.

bblommers (approved):
LGTM - thanks @dreamorosi!

PR #3614, opened 2021-01-24
Pin sshpubkeys dependency to last version to support Py2

### LINKED ISSUE #3613: Dependency sshpubkeys dropped support for Py2.7
sshpubkeys==3.1.0 supports Py2.7, but sshpubkeys==3.2.0 only supports Python >= 3.6

PR with a fix incoming.

bblommers:
Fixes #3613


bblommers:
This is fixed as of moto >= 1.3.16.dev214

PR #3615, opened 2021-01-25
Assume synchronous Lambda invocation by default

EpicWink:
If `InvocationType` is not provided to `boto.client("lambda").invoke`, then the request header `x-amz-invocation-type` is not present, which breaks an assumption in the model's implementation `moto.awslambda.models.LambdaFunction.invoke`.

In the tests, this results in the response not having a `LogResult` key (when `LogType="Tail"`), or trying to delete a non-existent response header key `x-amz-log-result` (when `LogType` is not provided).

This PR updates the assumption to match the current API definition.

### DIFF
--- a/moto/awslambda/models.py
+++ b/moto/awslambda/models.py
@@ -621,7 +621,8 @@ class LambdaFunction(CloudFormationModel, DockerModel):
 
         # Get the invocation type:
         res, errored, logs = self._invoke_lambda(code=self.code, event=body)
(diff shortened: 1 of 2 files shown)


EpicWink:
The failure for Python 3.7 is due to `black` failing on code that wasn't changed. The failure for Python 2 is due to syntax issue with dict-unpacking

bblommers (approved):
LGTM - thanks @EpicWink!

bblommers:
This is now part of moto >= 1.3.16.dev218

PR #3616, opened 2021-01-25
Address `boto` deprecation warnings

bpandola:
This commit eliminates the following warning (of which there are currently dozens):

../boto/ec2/connection.py:582:
  PendingDeprecationWarning: The current get_all_instances implementation will be replaced with get_all_reservations.

`boto` isn't likely to ever make good on this warning, but doing the replacement will
declutter the `moto` test output.


bblommers (approved):
Nice one @bpandola

bblommers:
This is now part of moto >= 1.3.16.dev219

This removed 47 warnings overall

PR #3618, opened 2021-01-26
Use environment markers for Python 2 dependency pins

andersk:
PyPI does not run `setup.py` separately for each Python version to resolve requirements, so these conditional requirements need to be properly specified in the metadata, not in code.


bblommers (approved):
LGTM - thanks @andersk!

PR #3619, opened 2021-01-26
requirements-dev.txt: Remove requirements copied from setup.py

andersk:
Most of these were already pulled in via `-r requirements.txt` → `-e .[all]`; add the remainder with `-e [all,server]`.


bblommers (approved):
:+1:

PR #3620, opened 2021-01-27
Tech Debt - Use known AMI in internal tests

bpandola:
Moto's implementation of autoscaling:CreateLaunchConfiguration is a little too loose,
allowing the ImageId parameter to be omitted and defaulting it to None, which results
in dozens of deprecation warnings:

`PendingDeprecationWarning: Could not find AMI with image-id:ami-123456, in the near future this will cause an error.`

This commit refactors calls to CreateLaunchConfiguration to explicitly pass in a known
AMI ImageId in order to avoid these warnings.

The integrity of the tests is unaffected by this change.


bblommers (approved):
LGTM - thanks!

PR #3625, opened 2021-01-28
Address pytest warnings

bpandola:
This should just about do it for cleaning up the pytest output.

Commits are atomic and probably best left un-squashed for git blame/history purposes.


bpandola:
I'll rebase and reopen this when the build is fixed.

PR #3626, opened 2021-01-28
Cloudwatch get metric data return by period

irahulranjan:
addresses localstack/localstack#3493


bblommers (changes requested):
Thanks for the PR, @irahulranjan.
Would you mind extending the test case with more datapoints? I wan to make sure that we are picking up multiple datapoints within a period, so that the logic around the statistic (sum/min/max) is still correct.

bblommers (approved):
LGTM

bblommers:
This is now part of moto >= 1.3.17.dev241

PR #3631, opened 2021-01-29
Add support for the force argument for ecs.delete_service

william-richard:
This argument allows you to delete a service even when its DesiredCount is non-zero.

https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/ecs.html#ECS.Client.delete_service


bblommers (approved):
LGTM. Thanks @william-richard!

bblommers:
This is now part of moto >= 1.3.17.dev231

Issue #3477, opened 2020-11-17
SNS errors do not match what AWS actually returns


neoascetic:
For example, when I am trying to fetch the attributes of not existing platform endpoint, moto gives me the following response:

```
<?xml version="1.0" encoding="UTF-8"?>
<ErrorResponse>
      <Errors>
        <Error>  
          <Code>NotFound</Code>
          <Message>Endpoint with arn arn:aws:sns:us-east-1:000000000000:endpoint/Bar/Foo/8c20d50c-82d5-4d39-93c8-100cc64fa512 not found</Message>
        </Error>
     </Errors>
     <RequestID>7a62c49f-347e-4fc4-9331-6e8eEXAMPLE</RequestID>
</ErrorResponse>
```

while AWS returns something like this:

```
<ErrorResponse xmlns="http://sns.amazonaws.com/doc/2010-03-31/">
  <Error>
    <Type>Sender</Type>
    <Code>NotFound</Code>
    <Message>Endpoint does not exist</Message>
  </Error>
  <RequestId>ac6c2567-45f5-5295-a271-7d49406cbe7c</RequestId>
</ErrorResponse>
```

I guess exceptions defined in https://github.com/spulec/moto/blob/master/moto/sns/exceptions.py must use [this template](https://github.com/spulec/moto/blob/19fc76f4669d545b56cfd79c55377f46379d546d/moto/sns/responses.py#L1095-L1102) to render themselves instead?

bblommers:
Thanks for raising this @neoascetic!

Looks like you're right - the exceptions should define a custom template ([like this](https://github.com/spulec/moto/blob/master/moto/sagemaker/exceptions.py#L11)). Marking it as a bug.

aantillonl:
I'd like to chip in with this issue, please check the PR and let me know your comments

bblommers:
This is now fixed as of moto >= 1.3.17.dev235

PR #3632, opened 2021-01-29
Update SNS get_endpoint_attributes response for Not Found endpoints #3477

aantillonl:
Currently the response to `get_endpoint_attributes` for a non existent endpoint
differ from boto and moto, in the following way (shown in JSON instead of xml
for shortness):

* **boto**: `{ErrorResponse:  { Error: { Code: "... ", Type: "... ", Sender: "..." }}`
* **moto**: ` { ErrorResponse: { Errors: [ { Error: { Code: "... ", Type: "... ", Sender: "..." }}]}}`

The template used to generate the NotFound error was the default for [RESTError](https://github.com/spulec/moto/blob/master/moto/core/exceptions.py).

Moreover, although `RESTError` has another template called [SINGLE_ERROR_RESPONSE](https://github.com/spulec/moto/blob/master/moto/core/exceptions.py#L8). 
That at first sight looks like the right one, it lacks the `ErrorResponse` node at the root
of the XML, which causes an error when parsing that XML later in the application, and
results in a `ClientError` instead of a `NotFound` exception.

This PR takes the approach suggested in the [issue conversation](https://github.com/spulec/moto/issues/3477#issuecomment-729090978).
Although there is one thing I personally do not entirely like, which is the fact there is yet 
another variation of this template:

* One in [moto/core/exceptions.py](https://github.com/spulec/moto/blob/master/moto/core/exceptions.py#L8)
* One in [moto/sns/responses.py](https://github.com/spulec/moto/blob/master/moto/sns/responses.py#L1095)
* Plus the one added in this commit

An alternative, which avoids creating the third template is not to catch the exception in [moto/sns/models.py](https://github.com/spulec/moto/blob/master/moto/sns/models.py#L608), 
allowing the exception to "bubble" up to [moto/sns/responses.py](https://github.com/spulec/moto/blob/master/moto/sns/responses.py#L544)
where it can be caught, and return an error response using the already existing
template in that module.

The issue with this later approach, would be that it differs a bit more from the current
implementation, since `moto/sns/models.py` does catch all exceptions for all other
functions, and it is also nicer to catch exceptions sooner than later.

### DIFF
--- a/moto/sns/exceptions.py
+++ b/moto/sns/exceptions.py
@@ -1,12 +1,24 @@
 from __future__ import unicode_literals
 from moto.core.exceptions import RESTError
 
+NOT_FOUND_ENDPOINT_ERROR = """<ErrorResponse xmlns="http://sns.amazonaws.com/doc/2010-03-31/">
+  <Error>
+    <Type>{{ sender }}</Type>
+    <Code>{{ error_type }}</Code>
+    <Message>{{ message }}</Message>
+  </Error>
+  <RequestId>9dd01905-5012-5f99-8663-4b3ecd0dfaef</RequestId>
+</ErrorResponse>"""
+
 
 class SNSNotFoundError(RESTError):
     code = 404
 
-    def __init__(self, message):
-        super(SNSNotFoundError, self).__init__("NotFound", message)
+    def __init__(self, message, **kwargs):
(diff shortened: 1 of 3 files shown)


bblommers:
Looks like a duplicate of #3637, so I'll close this

PR #3640, opened 2021-01-31
Remove redundant test code

bpandola:
These tests, when run, do not execute any `moto` code. They fail the
parameter validation check in `botocore`, which raises an exception
before ever sending a request.  These tests do not cover or verify
any `moto` behavior and have been removed.

* Where possible, tests were modified to exercise `moto` backend code 
   instead of `botocore` code.
* The weird diff in `test_elbv2` is because the original commit that added 
  `test_fixed_response_action_listener_rule_validates_status_code` accidentally 
   copy/pasted the test into itself one indentation level. (https://github.com/spulec/moto/commit/2b8bdc9bca85d97852e0bd9b4fddccd009554d35)


bblommers (approved):
:+1:

PR #3641, opened 2021-01-31
Add failures output to ecs.describe_services

### LINKED ISSUE #2898: ECS Service Info
When getting the info for ECS services using *describe_services* the real boto3 will return a list of failures and report that a service was "MISSING" in a list at the 'failures' key of the json response. However, moto reports an empty failures list

gruebel:
Fixes #2898

I also changed the Account ID to the main one and added a `_get_cluster()` method to centralize the `ClusterNotFoundException`.


bblommers (approved):
LGTM. Thanks @gruebel!

PR #3644, opened 2021-02-01
Fix: IoT does not work in server mode

### LINKED ISSUE #1631: IoT does not work in server mode
**Versions:** 
moto 1.3.3
awscli 1.15.14
python 3.6.1

**How to reproduce:**
```
pip install moto
pip install moto[server]
moto server iot
aws --endpoint-url http://localhost:5000 iot create-thing --thing-name testing

```

The create thing request results in a 404:
```
An error occurred (404) when calling the CreateThing operation: <!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 3.2 Final//EN">
<title>404 Not Found</title>
<h1>Not Found</h1>
<p>The requested URL was not found on the server.  If you entered the URL manually please check your spelling and try again.</p>

```
This occurs for all iot and iot-data endpoints.

**Possible Resolution:**
The code seems to be implemented, but the url paths are not defined in  Adding the URL's seems to solve the routing issues, I've done this in the branch [here](https://github.com/spulec/moto/compare/master...brianrower:feature/fix-iot-server)

**Problems:**
BUT, for the device shadow API to work (which is iot-data) you have to call create-thing, which is an iot api call. I'm unclear how to run iot and iot-data in the same process to enable this.
The work around I came up with is:

- modify your hosts file to point iot.us-west-2.amazonaws.com and data.iot.us-west-2.amazonaws.com to localhost
- run "moto server" with no parameters
- aws --endpoint-url http://iot.region.amazonaws.com:5000 iot create-thing --thing-name testing
- aws --endpoint-url http://data.iot.us-west-2.amazonaws.com:5000 iot-data update-thing-shadow --thing-name testing --payload '{"state": { "desired": {"testing":true}}}' testing

The workaround succeeds because "get_backend_for_host" in server.py lookups up which service to call based on the domain name and allows both iot and iot-data to be hit in the same process.
Is this the preferred method of interacting wtih moto standalone server? If so I'll add documentation to instruct users in this way. If not, what other way is there?

bpandola:
Closes #1631

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,13 +1,10 @@
 SHELL := /bin/bash
 
 ifeq ($(TEST_SERVER_MODE), true)
-	# exclude test_iot and test_iotdata for now
-	# because authentication of iot is very complicated
-
 	# exclude test_kinesisvideoarchivedmedia
 	# because testing with moto_server is difficult with data-endpoint
 
-	TEST_EXCLUDE :=  -k 'not (test_iot or test_kinesisvideoarchivedmedia)'
+	TEST_EXCLUDE :=  -k 'not test_kinesisvideoarchivedmedia'
 else
 	TEST_EXCLUDE :=
 endif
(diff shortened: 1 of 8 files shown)


bblommers (approved):
👍

PR #3647, opened 2021-02-01
Remove duplicate template of SNS error response #3646

### LINKED ISSUE #3646: SNS Response Template Duplicate
The following response templates are essentially the same:

* [moto/sns/exceptions.py](https://github.com/spulec/moto/blob/master/moto/sns/exceptions.py#L4)
* [moto/sns/responses.p](https://github.com/spulec/moto/blob/master/moto/sns/responses.py#L1095)

The one in `exceptions.py` was added in e9dc5ed  to correct the response 
returned by get_endpoint_attributes for a non-existent endpoint. However,
after some more inspection, it was found that a valid template already existed
in `responses.py`

The template from `exceptions.py` could be removed after all, since it is the
one added later, and the one from `responses.py` could be used for the same
purpose

aantillonl:
In e9dc5ed a custom error template was added to correct the response
returned by get_endpoint_attributes for a non-existent endpoint.

However, after some more inspection, it was found that a valid template
already existed in moto/sns/responses.py. This commit removes the 
template introduced in e9dc5ed and takes into use the existing template.

It is a marginal improvement, but helps keeping a cleaner code base.

fixes #3646

### DIFF
--- a/moto/sns/exceptions.py
+++ b/moto/sns/exceptions.py
@@ -1,23 +1,11 @@
 from __future__ import unicode_literals
 from moto.core.exceptions import RESTError
 
-NOT_FOUND_ENDPOINT_ERROR = """<ErrorResponse xmlns="http://sns.amazonaws.com/doc/2010-03-31/">
-  <Error>
-    <Type>{{ sender }}</Type>
-    <Code>{{ error_type }}</Code>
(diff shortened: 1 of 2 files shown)


bblommers (approved):
👍

PR #3650, opened 2021-02-02
 S3 - Allow uploads using a PUT request without content-type

### LINKED ISSUE #3363: moto_server S3 HTTP PUT/GET
using moto 1.3.16

Not sure if moto actually supports this and I'm just being daft, or if this isn't supported to begin with.
I've tried searching through documentation and other issues/PRs with no luck.

I am running a standalone `moto_server s3` as I need to proxy NGINX HTTP GET/PUT requests to it, but for the purposes of getting-it-working I have cut out the NGINX middleman for now.
I also have some pytests which use `@mock_s3` for creating a test bucket.

I have a couple of issues:
- My pytest tests *seem* to work, successfully creating a bucket and `PUT`ing a file in the correct bucket, however with no contents according to `http://localhost:5000/moto-api`, failing the last stage of one of my tests:
```python
    data_input = b'some data 123 321'
    req = urllib.request.Request(url='localhost:5000/mybucket/test.txt', data=data_input, method='PUT')

    with urllib.request.urlopen(req) as f:
        assert f.status == 200  # PASSED
    
    req = urllib.request.Request(url='localhost:5000/mybucket/test.txt', method='GET')
    with urllib.request.urlopen(req) as f:
        assert f.status == 200   # PASSED

        data_response = f.read()
        assert data_response == data_input  # <- FAILURE: b'' != b'some data 123 321'
```
- If I manually PUT a file using `curl` I again get no contents when using `curl -X PUT localhost:5000/mybucket/test.txt -d 'test123'`, however when using `curl -X PUT localhost:5000/mybucket/test.txt --upload-file ./test.txt` it works, creating a file with *actual* contents for a change. The problem is if I try to access the file via HTTP GET (either through `curl` or my browser) I get a 403 error. As part of the bucket creation (using `@mock_s3`) I have tried attaching a bucket policy allowing [read-only permission for anonymous users](https://docs.aws.amazon.com/AmazonS3/latest/dev/example-bucket-policies.html#example-bucket-policies-use-case-2), however this has not made any difference.

TL;DR: I want to be able to PUT and GET files (and their contents) unauthenticated with REST/without boto3.
Just to clarify, all my tests **work** when proxying NGINX to an actual live S3 bucket (with v4 authentication).

Apologies if I haven't been too clear, I'm writing this after a day of tinkering not having achieved much.

On a completely unrelated note, I thought I'd use this opportunity to ask if `moto_server` has a daemon/background mode.

## Update
Turns out adding `req.add_header("Content-Type", "text/plain")` to the PUT request makes it all work.

In which case, this issue could now be a question about whether or not `moto_server` should accept/store data without a content-type. S3 REST allows it therefore I believe (granted, as a new user of moto in general) moto_server should accommodate this.

bblommers:
Fixes #3363

### DIFF
--- a/moto/s3/responses.py
+++ b/moto/s3/responses.py
@@ -1146,10 +1146,17 @@ class ResponseObject(_TemplateEnvironmentMixin, ActionAuthenticatorMixin):
         else:
             # Flask server
             body = request.data
-            # when the data is being passed as a file
-            if request.files and not body:
-                for _, value in request.files.items():
-                    body = value.stream.read()
+            if not body:
+                # when the data is being passed as a file
+                if request.files:
+                    for _, value in request.files.items():
+                        body = value.stream.read()
+                elif hasattr(request, "form"):
+                    # Body comes through as part of the form, if no content-type is set on the PUT-request
+                    # form = ImmutableMultiDict([('some data 123 321', '')])
+                    form = request.form
+                    for k, _ in form.items():
+                        body = k
 
         if body is None:
             body = b""
--- a/tests/test_s3/test_s3.py
+++ b/tests/test_s3/test_s3.py
@@ -5,7 +5,7 @@ import datetime
 import sys
 import os
 from boto3 import Session
-from six.moves.urllib.request import urlopen, Request
+from six.moves.urllib.request import urlopen
 from six.moves.urllib.error import HTTPError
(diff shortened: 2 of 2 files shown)


bblommers:
Codecov wasn't updating/pulling in the latest config from master. Created a new PR to force a proper coverage report: #3699

PR #3657, opened 2021-02-03
SQS - Raise error message when providing an invalid value for the queue_url parameter

### LINKED ISSUE #3655: SQS - send_message should not accept the queue-name as a value
Example code:

```
queue = sqs.Queue(queue_url="queue_name")
queue.send_message(MessageBody="this is a test message")
```

This should fail, as the queue_url-parameter should be the entire URL.
Note: This error should be thrown for every operation, I'm just using `send_message` as an example

bblommers:
Fixes #3655


tests/test_sqs/test_sqs.py
```diff
@@ -229,20 +229,15 @@ def test_get_queue_url_errors():
...
-        sqs.get_queue_by_name(QueueName="non-existing-queue")
-    ex = err.value
-    ex.operation_name.should.equal("GetQueueUrl")
-    ex.response["Error"]["Code"].should.equal("AWS.SimpleQueueService.NonExistentQueue")
-    ex.response["Error"]["Message"].should.equal(
-        "The specified queue non-existing-queue does not exist for this wsdl version."
-    )
 
     with pytest.raises(ClientError) as err:
         sqs.Queue("http://whatever-incorrect-queue-address").load()
```
bblommers, line 234: I started out with a more complete URL validator, until I saw this test. Running this against AWS itself results in this NonExistentQueue-error.

I'd expected it to return an InvalidAddress, because the URL does not adhere to the usual `http://amazonaws.com/queue/queue-name`, but I guess AWS doesn't do a full URL validation either

tests/test_sqs/test_sqs.py
```diff
@@ -311,20 +311,15 @@ def test_get_queue_url_error_not_exists():
 @mock_sqs
 def test_get_nonexistent_queue():
     sqs = boto3.resource("sqs", region_name="us-east-1")
-    with pytest.raises(ClientError) as err:
```
bblommers, line 314: This particular check wasl already part of the previous test, so there's no need to replicate it

Issue #3665, opened 2021-02-08
Add KmsKeyId to Redshift Cluster


jnsf-cg:
When creating a new [Redshift cluster](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-redshift-cluster.html#aws-resource-redshift-cluster-properties), specifying the `KmsKeyId` does not persist to the redshift backend. This results in the `KmsKeyId` not being returned when querying for clusters using the [describe_clusters](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/redshift.html#Redshift.Client.describe_clusters) boto3 function.

bblommers:
This is now part of moto >= 1.3.17.dev246

PR #3666, opened 2021-02-08
Add KmsKeyId to Redshift Cluster

jnsf-cg:
Add the KmsKeyId property when creating a cluster so that it is also returned when querying the describe_clusters endpoint.

Resolves https://github.com/spulec/moto/issues/3665

### DIFF
--- a/tests/test_redshift/test_redshift.py
+++ b/tests/test_redshift/test_redshift.py
@@ -41,6 +41,7 @@ def test_create_cluster_boto3():
         datetime.datetime.now(create_time.tzinfo) - datetime.timedelta(minutes=1)
     )
     response["Cluster"]["EnhancedVpcRouting"].should.equal(False)
+    response["Cluster"]["KmsKeyId"].should.equal("")
 
 
 @mock_redshift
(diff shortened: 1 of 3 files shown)


bblommers (changes requested):
Thanks for the PR @jnsf-cg! 
Would you mind adding a unit test for this, to confirm that the attribute is properly stored/returned?

tests/test_redshift/test_redshift.py
```diff
@@ -64,6 +65,31 @@ def test_create_cluster_with_enhanced_vpc_routing_enabled():
     response["Cluster"]["EnhancedVpcRouting"].should.equal(True)
 
 
+@mock_redshift
+def test_create_and_describe_cluster_with_kms_key_id():
```
jnsf-cg, line 69: @bblommers I combined the create and describe into a single test to validate that the `KmsKeyId` is used. I added a check in the previous unit test to check that KmsKeyId is empty if not provided. If you'd rather me create another test with a signature of 

```python
test_create_and_describe_without_kms_key_id()
```

I can do that as well.

jnsf-cg:
Not sure why the linter is complaining with:

```
black --check moto/ tests/
would reformat /home/runner/work/moto/moto/tests/test_redshift/test_redshift.py
```

I even ran `black tests/test_redshift` locally and it says 3 files left unchanged.

bblommers:
Which version of black are you running? We ran into issues with the latest version, so for moto it's pinned to `black==19.10b0`

jnsf-cg:
> Which version of black are you running? We ran into issues with the latest version, so for moto it's pinned to `black==19.10b0`

Ah that was it. Thanks!

bblommers (approved):
LGTM! Thanks again

Issue #3667, opened 2021-02-09
better s3 post support for success_action_redirect


wynnw:
The success_action_redirect value that is supported by s3 posts doesn't match how s3 works in production in a couple of ways that does not allow me to fully test my s3 integration code. Missing features are:

- automatic addition of the bucket and key values to the success_action_redirect query args. I can't find exact documentation for this behavior, but this page references it:
https://docs.aws.amazon.com/AmazonS3/latest/API/RESTObjectPOST.html

and it's how it works and we rely  on that behavior.

- automatic replacement of ${filename} in the form key value. Same issue as the above with documentation.

bblommers:
This is now part of moto >= 1.3.17.dev248

wynnw:
Thanks for the reviews! Great project, has saved us a lot of time - we're big fans.

PR #3668, opened 2021-02-09
#3667: Update the s3 post functionality to better support success_action_redirect

wynnw:
…irect

- Add the bucket/key values to the redirect url like s3 does, which
supports code that relies on the key value being there on the
redirect.
- Add support for replacing ${filename} in the key value with the actual
filename from the form upload.

See Issue #3667

### DIFF
--- a/moto/s3/responses.py
+++ b/moto/s3/responses.py
@@ -1,5 +1,6 @@
 from __future__ import unicode_literals
 
+import os
 import re
 import sys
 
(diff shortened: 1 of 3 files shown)


wynnw:
The tests aren't passing on this, but they aren't on the standard master branch either without my commit. Hangs in the test_lambda.py code after 4 failures.

bblommers (changes requested):
Hi @wynnw, thanks for your PR!

The tests might be failing locally because you don't have docker running/installed.
The only CI test that is failing is `test_creating_presigned_post`, because it's doesn't expect the Location-header to have the new bucket/key parameters.

Would be good to have a new unit test for the `${filename}` change as well

wynnw:
I just pushed up a commit to address those test issues.

I've got docker configured and running but still can't get all the tests to pass. Sounds like they are for you though? This latest commit should fix the test failure you mentioned.

wynnw:
Well this passed so I think I've got it. Sorry for the python2.7 breakage churn, but it looks all good now.

bblommers (approved):
LGTM @wynnw. Thanks for making the changes!

wynnw:
Thanks for the reviews and working with me on this. This allows us to write some really great integration tests that have been missing for us.

PR #3669, opened 2021-02-09
Add support for RDS resource filtering

### LINKED ISSUE #3656: Feature Request: `describe_db_instances` method does not filter the resources based on the `Filters` parameter provided
Here's a code snippet - 
```
response = self.client.describe_db_instances(
            Filters=[{
                'Name': 'db-instance-id',
                'Values': [rds_arn]
            }]
        )
```
It returns all the RDS in the response instead of just the filtered one.

bpandola:
I initially tried to leverage the filter code that already exists for EC2, but the RDS filters work slightly differently so I ended up implementing a unique solution.  The basic idea is that each model class (DBInstance/DBSnapshot) defines its supported filters by mapping filter names to the equivalent object attributes.  This keeps all the specific filter details properly contained and allows for all of the code that actually implements the filtering to be totally generic.  All of the generic filtering code is covered by a fairly comprehensive test suite that defines exactly how it all works, which may prove useful for someone implementing filtering for additional RDS endpoints or perhaps another service altogether down the road.

* Extensive testing was performed against real AWS endpoints in order to
  nail down the filter behavior under various scenarios, ensuring that
  `moto` returns the proper response or error.
* Full test coverage of all utility functions as well as several
  filter/parameter combinations.

Closes #3656


bblommers (changes requested):
Nice one! Would you mind splitting up the snapshot/instance test cases? I realize there's some setup across all of them, but it would be nice to only test a single feature/error per method

bpandola:
Codecov report is clearly incorrect (or I must not be understanding how it's calculating the diff). 

In any case, I have made the requested changes.  Regards.

bblommers (approved):
Awesome. Thanks @bpandola!

bblommers:
> 
> 
> Codecov report is clearly incorrect (or I must not be understanding how it's calculating the diff).

It looks like the coverage report is posted immediately after the first build passes. Py3.7 is typically the fastest, but that means that the coverage report does not cover any Py2/servermode specific code.
Will see if I can change this.

The full report can still be seen here, btw: https://codecov.io/gh/spulec/moto/pull/3669/ 
That does look correct.

PR #3670, opened 2021-02-09
Functionality added to SES service

pinzon:
This patch adds the functions test_render_template and update_template so the SES service.
Also adds validation to the function add_template.

Each new function has its corresponding test.

This should close issue localstack/localstack/issues/3466

### DIFF
--- a/tests/test_ses/test_ses_boto3.py
+++ b/tests/test_ses/test_ses_boto3.py
@@ -496,7 +496,9 @@ def test_render_template():
         TemplateData=json.dumps({"name": "John", "favoriteanimal": "Lion"}),
     )
 
-    conn.test_render_template.when.called_with(**kwargs).should.throw(ClientError)
+    with pytest.raises(ClientError) as ex:
+        conn.test_render_template(**kwargs)
+    ex.value.response["Error"]["Code"].should.equal("TemplateDoesNotExist")
 
     conn.create_template(
         Template={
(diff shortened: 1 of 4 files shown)


pinzon:
Hi @bblommers sorry to bother you. Could you review my PR? 
Is there something else I should do? some tests failed but I don't think they have to do with my changes.
Thank you.

tests/test_ses/test_ses_boto3.py
```diff
@@ -484,3 +485,66 @@ def test_create_ses_template():
...
+            "TemplateName": "MyTestTemplate",
+            "SubjectPart": "Greetings, {{name}}!",
+            "TextPart": "Dear {{name}},"
+            "\r\nYour favorite animal is {{favoriteanimal}}.",
+            "HtmlPart": "<h1>Hello {{name}},"
+            "</h1><p>Your favorite animal is {{favoriteanimal}}.</p>",
+        }
+    )
+    result = conn.test_render_template(**kwargs)
+    assert True == result["RenderedTemplate"].__contains__("Subject: Greetings, John!")
```
bblommers, line 512: Please use the should-flow for these assertions:
`template.should.contain(...)`

tests/test_ses/test_ses_boto3.py
```diff
@@ -484,3 +485,66 @@ def test_create_ses_template():
...
+@mock_ses
+def test_render_template():
+    conn = boto3.client("ses", region_name="us-east-1")
+
+    kwargs = dict(
+        TemplateName="MyTestTemplate",
+        TemplateData=json.dumps({"name": "John", "favoriteanimal": "Lion"}),
+    )
+
+    conn.test_render_template.when.called_with(**kwargs).should.throw(ClientError)
```
bblommers, line 499: When testing error scenarios, we prefer to use pytest.raises, as that makes it easier to assert that the correct error was thrown:
```
  # something that fails
err  = ex.value.response["Error"]
err["Code"].should.equal("...")
err["Message"].should.equal("...")
```

bblommers (changes requested):
Thanks for the PR, @pinzon! I've left a few comments.
Looks like the build is failing as well, because of a Py2 issue: https://github.com/spulec/moto/pull/3670/checks?check_run_id=1872455785

pinzon:
Hi @bblommers, my last commit should fix that error and I changed the tests as requested.
Thank you for your help.

bblommers (approved):
LGTM. Thanks again @pinzon!

PR #3671, opened 2021-02-10
events: fix archive event pattern match check

elopez:
There is a missing `return True` for the positive match case in
matches_pattern, causing all valid patterns to fail.

### DIFF
--- a/moto/events/models.py
+++ b/moto/events/models.py
@@ -287,6 +287,8 @@ class Archive(CloudFormationModel):
             if event_value not in pattern_value:
                 return False
 
+        return True
+
     def get_cfn_attribute(self, attribute_name):
         from moto.cloudformation.exceptions import UnformattedGetAttTemplateException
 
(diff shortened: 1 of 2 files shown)


bblommers (changes requested):
Hi @elopez, thanks for the PR! Would you mind adding a unit test for this, to verify this now works correctly?

elopez:
Hi! I'll try to get some time to write the test soon, but if anyone else wants to do it meanwhile, feel free 👍

elopez:
Hi @bblommers, I updated one of the existing tests to check this case as well. The current test only considered the empty pattern and the "non-empty but not matching" pattern cases, so I added an instance of the "non-empty and matching" case.

bblommers (approved):
LGTM - thanks @elopez!

bblommers:
This is now part of moto >= 1.3.17.dev270

PR #3672, opened 2021-02-10
moved iteritems backwards compatibility to six

pwrmiller:
I noticed in testing today that we weren't using `six` for the `iteritems` import. This avoids the needless dependency on `future` in this file by using `six` instead. This matches how `iteritems` is imported across the remainder of the project.


bblommers (approved):
Nice @pwrmiller - thanks!

PR #3680, opened 2021-02-12
Fix heartbeatTimeout of NONE and polling responses when there are no tasks

redparham:
While using moto swf with AWS Java Flow SDK 1.11.22, I found some incompatibilities with activity heartbeatTimeout and decision/activity polling.

### heartbeatTimeout
When activities don't define a heartbeatTimeout, moto/swf/models/activity_task.py fails and returns:
`ValueError: invalid literal for int() with base 10: 'NONE'`
when this code runs:
`
        heartbeat_timeout_at = self.last_heartbeat_timestamp + int(
            self.timeouts["heartbeatTimeout"]
        )
`

### decision/activity polling
When poll_for_decision_task and poll_for_activity_task in moto/swf/responses.py execute, they currently return an empty string for taskToken:
`return json.dumps({"startedEventId": 0, "taskToken": ""})`
or
```
return json.dumps(
    {"previousStartedEventId": 0, "startedEventId": 0, "taskToken": ""}
)
```
But this results in an exception thrown by my Java decision/activity pollers:
```
java.lang.IllegalArgumentException: No implementation was found for null
	at com.amazonaws.services.simpleworkflow.flow.worker.AsyncDecisionTaskHandler.createDecider(AsyncDecisionTaskHandler.java:111)
	at com.amazonaws.services.simpleworkflow.flow.worker.AsyncDecisionTaskHandler.handleDecisionTask(AsyncDecisionTaskHandler.java:49)
	at com.amazonaws.services.simpleworkflow.flow.worker.DecisionTaskPoller.pollAndProcessSingleTask(DecisionTaskPoller.java:201)
	at com.amazonaws.services.simpleworkflow.flow.worker.GenericWorker$PollServiceTask.run(GenericWorker.java:94)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
	at java.base/java.lang.Thread.run(Thread.java:834)
```
It can be confirmed here that decision pollers don't expect an empty string for taskToken but rather null:
https://github.com/aws/aws-swf-flow-library/blob/b2fee40ca289f6cd882b3289d65795d10f0b092b/src/main/java/com/amazonaws/services/simpleworkflow/flow/worker/DecisionTaskPoller.java#L198
```
        if (result == null || result.getTaskToken() == null) {
            return null;
        }
```
And the same for activity pollers:
https://github.com/aws/aws-swf-flow-library/blob/b2fee40ca289f6cd882b3289d65795d10f0b092b/src/main/java/com/amazonaws/services/simpleworkflow/flow/worker/ActivityTaskPoller.java#L174
```
        if (result == null || result.getTaskToken() == null) {
            if (log.isDebugEnabled()) {
                log.debug("poll request returned no task");
            }
            return null;
        }
```

### DIFF
--- a/moto/swf/models/activity_task.py
+++ b/moto/swf/models/activity_task.py
@@ -73,7 +73,10 @@ class ActivityTask(BaseModel):
     def first_timeout(self):
         if not self.open or not self.workflow_execution.open:
             return None
-        # TODO: handle the "NONE" case
+
+        if self.timeouts["heartbeatTimeout"] == "NONE":
+            return None
+
         heartbeat_timeout_at = self.last_heartbeat_timestamp + int(
             self.timeouts["heartbeatTimeout"]
         )
--- a/moto/swf/responses.py
+++ b/moto/swf/responses.py
@@ -446,9 +446,7 @@ class SWFResponse(BaseResponse):
         if decision:
             return json.dumps(decision.to_full_dict(reverse_order=reverse_order))
         else:
-            return json.dumps(
-                {"previousStartedEventId": 0, "startedEventId": 0, "taskToken": ""}
-            )
+            return json.dumps({"previousStartedEventId": 0, "startedEventId": 0})
 
     def count_pending_decision_tasks(self):
         domain_name = self._params["domain"]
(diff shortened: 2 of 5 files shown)


bblommers (approved):
LGTM - thanks @redparham!

PR #3682, opened 2021-02-12
Minor RDS Clean up

bpandola:
Just some minor clean up of some things I found while working on something else.

Changes the DBInstanceNotFound message to match what is returned from a real AWS backend.

Removes a duplicate test that was not getting run (and wouldn't have succeeded if it had run).

Makes the existing test (that was duplicated) a little more explicit.


bblommers (approved):
LGTM!

PR #3685, opened 2021-02-13
support - intial commit to kick off trusted advisor checks

ccggeo:
Hi,

I noticed in the week there was not one yet so this is a first attempt at the implementation for the support feature.

Currently only returns the call from `describe_trusted_advisor_checks` and as this is just a static response have included it as part of a json.

The contribution page did not say anything about a minimum number of operations to add so I kept it small as this is my first PR to moto. I would like to do more for this module but would like to know what you think of my PR for feedback

Thanks!

### DIFF
--- a/moto/support/responses.py
+++ b/moto/support/responses.py
@@ -1,0 +1,24 @@
+from __future__ import unicode_literals
+from moto.core.responses import BaseResponse
+from .models import support_backends
+import json
+
+
+class SupportResponse(BaseResponse):
+    SERVICE_NAME = "support"
+
+    @property
+    def support_backend(self):
+        return support_backends[self.region]
+
+    def describe_trusted_advisor_checks(self):
+        language = self._get_param("language")
+        checks = self.support_backend.describe_trusted_advisor_checks(
+            language=language,
+        )
+
+        response = {"checks": []}
+        for item in checks:
+            response["checks"].append(item)
+
+        return json.dumps(dict(response))
--- a/moto/support/responses.py
+++ b/moto/support/responses.py
@@ -17,8 +17,4 @@ class SupportResponse(BaseResponse):
             language=language,
         )
 
-        response = {"checks": []}
-        for item in checks:
-            response["checks"].append(item)
-
-        return json.dumps(dict(response))
+        return json.dumps({"checks": checks})
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -5,5 +5,6 @@ include moto/ec2/resources/instance_type_offerings/*/*.json
 include moto/ec2/resources/amis.json
 include moto/cognitoidp/resources/*.json
 include moto/dynamodb2/parsing/reserved_keywords.txt
+include moto/support/resources/*.json
 recursive-include moto/templates *
 recursive-include tests *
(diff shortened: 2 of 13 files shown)


moto/support/responses.py
```diff
@@ -0,0 +1,24 @@
...
+    def support_backend(self):
+        return support_backends[self.region]
+
+    def describe_trusted_advisor_checks(self):
+        language = self._get_param("language")
+        checks = self.support_backend.describe_trusted_advisor_checks(
+            language=language,
+        )
+
+        response = {"checks": []}
```
bblommers, line 20: It looks like this whole section can be simplified as `return json.dumps({"checks": checks})`

bblommers (changes requested):
Hi @ccggeo, thanks for the PR! Always exciting to see support for another service.

There's no minimum required for a PR at all - every bit helps.

The tests are currently failing in server mode, potentially because the new json-file isn't listed yet in the Manifest.in file. (I haven't debugged it yet, so I can't say for certain -but that's definitely something that's currently missing.)  
If you want to run the server tests yourself, you can use:
```
pip install .
moto_server
TEST_SERVER_MODE=true pytest -sv ./tests/test_support
```

ccggeo:
@bblommers thanks for the review have implemented your suggestion

Getting a weird error on travis unrelated to the testing content. I have purged my .pyc files and rename the test to no success. I think it might be related to the caching stages on Travis, as despite naming the test server to test_server_support.py it still fails.
Any ideas? I 

`==================================== ERRORS ====================================
______________ ERROR collecting tests/test_support/test_server.py ______________
import file mismatch:
imported module 'test_server' has this __file__ attribute:
  /home/runner/work/moto/moto/tests/test_medialive/test_server.py
which is not the same as the test file we want to collect:
  /home/runner/work/moto/moto/tests/test_support/test_server.py
HINT: remove __pycache__ / .pyc files and/or use a unique basename for your test file modules`

bblommers:
So far so good on the Ci.. but if it still fails, it might be because pytest expects an __init__.py file in the tests/test_support folder

ccggeo:
> So far so good on the Ci.. but if it still fails, it might be because pytest expects an `__init__.py` file in the tests/test_support folder

Good shout, i also thought this might happen and have included here. Hopefully that does it 

https://github.com/spulec/moto/blob/c02513e2d96b606fc92fb7aa60c15d8732716d80/tests/test_support/__init__.py

ccggeo:
`        if http.status_code >= 300:
E           botocore.exceptions.ClientError: An error occurred (500) when calling the DescribeTrustedAdvisorChecks operation (reached max retries: 4): <!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 3.2 Final//EN">
E           <title>500 Internal Server Error</title>
E           <h1>Internal Server Error</h1>
E           <p>The server encountered an internal error and was unable to complete your request. Either the server is overloaded or there is an error in the application.</p>`

Getting this error on the build server state. It is strange because this is working locally. Any ideas?

bblommers:
@ccggeo Our CI runs moto in a Docker container during the server tests. The error only showed up inside Docker because the relative paths are slightly different, so the json file couldn't be found. That part should be fixed now.

ccggeo:
Thanks so much for the help with the resource file

bblommers (approved):
LGTM. Thanks again @ccggeo!

bblommers:
This is now part of moto >= 1.3.17.dev262

PR #3687, opened 2021-02-13
Improve autoscaling:CreateLaunchConfiguration request validation

bpandola:
Something I discovered when cleaning up the "missing AMI" warnings (#3620) was that the implementation of CreateLaunchConfiguration was a bit too permissive.

AWS requires certain parameters to be mutually inclusive.

Moto wasn't doing anything with the InstanceId parameter, which is now made clear with a TODO.


bblommers (approved):
LGTM - thanks!

bblommers:
This is now part of moto >= 1.3.17.dev261

PR #3690, opened 2021-02-14
Feature/#3683 Add symmetric key capability to moto

rvodden:
Early draft PR for review - not ready for merge, but certainly ready for comment. Things still to do:

- [ ] `test_schedule_key_deletion` was broken locally when I pulled. Need to investigate that.
- [ ] Many of the tests, naturally, assume symmetric keys - I need to rename them to make that obvious (i.e. `test_decrypt_unknown_key_id` becomes `test_symmetric_decrypt_unknown_key_id`) and write a matching asymmetric tests (i.e. `test_asymmetric_decrypt_unknown_key_id`)
- [ ] Some of the failure modes don't match AWS, for example, `AccessDeniedError` seems to get thrown when `InvalidKeyUsageException` might be more appropriate.
- [x] There is something up with 512 byte pad and hash lengths - cryptography moans that there is not sufficient key material.
- [ ] Fix the build for Python = =2.7
- [ ] Add some more of the missing asymmetric endpoints namely
    - [x] `get_public_key`
    - [ ] `generate_data_key_pair`
    - [ ] `generate_data_key_pair_without_plaintext`
    - [x] `generate_random`
    - [x] `sign`
    - [x] `verify`

### DIFF
--- a/IMPLEMENTATION_COVERAGE.md
+++ b/IMPLEMENTATION_COVERAGE.md
@@ -6191,7 +6191,7 @@
 - [X] get_key_policy
 - [X] get_key_rotation_status
 - [ ] get_parameters_for_import
-- [ ] get_public_key
+- [X] get_public_key
 - [ ] import_key_material
 - [ ] list_aliases
 - [ ] list_grants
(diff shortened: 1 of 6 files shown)


rvodden:
Turns out the salt length is specified in bytes not bits, so I had a salt length 8 times too large!

bblommers:
> certainly ready for comment

Looks good so far! Will periodically check back in to review this.
Let me know if you need any help.

key:
Does anyone have an update on this pull request?

bblommers:
There have been a lot of improvements to KMS recently, so I think all of these features are now covered for. Closing.

PR #3700, opened 2021-02-17
Fix: DynamoDB:GetItem throws wrong error when table doesn't exist

### LINKED ISSUE #3695: Mocked DynamoDB throws wrong error when table doesn't exist
## Impact

I am unable to use moto to accurately test code that detects when a DynamoDB table is not replicated in the current region

## Reproduce

```python
import boto3

@mock_dynamodb2
def repro():
  client = boto3.client('dynamodb', 'us-east-1')
  result = client.get_item(TableName='foo', Key={'site-id': {'S': 'foo'}})
```

## Expected result

`ResourceNotFound` raised. Here's the stack trace from boto3 for a live API request:

```
  File "test.py", line 3, in <module>
    client.get_item(TableName='foo', Key={'site-id': {'S': 'foo'}})
  File "/usr/local/lib/python3.9/site-packages/botocore/client.py", line 357, in _api_call
  File "/usr/local/lib/python3.9/site-packages/botocore/client.py", line 676, in _make_api_call
botocore.errorfactory.ResourceNotFoundException: An error occurred (ResourceNotFoundException) when calling the GetItem operation: Requested resource not found
```

## Actual result

```
  File "venv/lib/python3.9/site-packages/moto/core/models.py", line 100, in wrapper
  File "<stdin>", line 4, in test_my_model_save
  File "venv/lib/python3.9/site-packages/botocore/client.py", line 337, in _api_call
  File "venv/lib/python3.9/site-packages/botocore/client.py", line 656, in _make_api_call
botocore.exceptions.ClientError: An error occurred (ValidationException) when calling the GetItem operation: Validation Exception
```

## Versions
- Python 3.9.0
- Moto 1.3.16
- botocore 1.18.0

bpandola:
Closes #3695

### DIFF
--- a/moto/dynamodb2/responses.py
+++ b/moto/dynamodb2/responses.py
@@ -349,6 +349,12 @@ class DynamoHandler(BaseResponse):
 
     def get_item(self):
         name = self.body["TableName"]
+        table = self.dynamodb_backend.get_table(name)
+        if table is None:
+            return self.error(
+                "com.amazonaws.dynamodb.v20120810#ResourceNotFoundException",
+                "Requested resource not found",
+            )
         key = self.body["Key"]
         projection_expression = self.body.get("ProjectionExpression")
         expression_attribute_names = self.body.get("ExpressionAttributeNames", {})
@@ -359,14 +365,9 @@ class DynamoHandler(BaseResponse):
 
         try:
             item = self.dynamodb_backend.get_item(name, key, projection_expression)
-        except KeyError:
-            er = "com.amazonaws.dynamodb.v20120810#ResourceNotFoundException"
-            msg = "Requested resource not found"
-            return self.error(er, msg)
         except ValueError:
             er = "com.amazon.coral.validate#ValidationException"
-            msg = "Validation Exception"
-            return self.error(er, msg)
+            return self.error(er, "Validation Exception")
         if item:
             item_dict = item.describe_attrs(attributes=None)
             item_dict["ConsumedCapacity"] = {"TableName": name, "CapacityUnits": 0.5}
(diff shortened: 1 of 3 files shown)


moto/dynamodb2/responses.py
```diff
@@ -359,9 +359,14 @@ def get_item(self):
 
         try:
             item = self.dynamodb_backend.get_item(name, key, projection_expression)
-        except ValueError:
-            er = "com.amazon.coral.validate#ValidationException"
-            return self.error(er, "Validation Exception")
+        except ValueError as e:
+            if str(e) == "No table found":
```
bblommers, line 363: Is it possible to throw a custom exception when the table does not exist? I feel like this would be a cleaner approach:
```
try:
...
except TableNotFound:
...
except ValueError:
...
```

bblommers (changes requested):
~~LGTM - thanks @bpandola!~~

Edit: See comment

moto/dynamodb2/responses.py
bpandola: That's what I did initially, but unfortunately it breaks other tests.  I opted for the simplest solution, though I agree with you that's it's not the most elegant.  Fixing this properly requires fairly major surgery, which I'm pretty hesitant to undertake.

bpandola: ...looked at this again.  Decided to just fix the reported issue, without touching anything else.  Let me know what you think.

bblommers: Yeah, I think this makes sense, to fail as early as possible. Thanks!

bblommers (approved):

Issue #3276, opened 2020-09-02
[feature] allow lambda to run with podman as an alternative to docker


yuvalif:
* allows you to containers not as root
* cgroups v2
* better support on centos8/fedora OSs

should be an alternative, and not a replacement, as docker is more widespread than podman

bblommers:
Hi @yuvalif, thanks for raising this. Hadn't heard of Podman before. 

Is there a specific use case for Lambdas where Podman would give a noticable improvement? From what I can read, it would behave exactly the same in our limited use case. (I.e., run t he container, execute some python code and return the result.)

yuvalif:
the main improvement is around installation and setup of docker (especially in case of centos8/fedora OSs), as well as the fact that non-privileged containers don't need root permissions in podman.
don't think there is functional advantage.

yan12125:
I give it a try, and get blocked as Podman hasn't implemented all Docker APIs. After some patches to moto and docker-py, the blocker comes to the `put_archive` function [1][2].

I'm interested in this as I'm maintaining the Arch Linux package and I'd like to run the test suite of moto without extra privileges.

[1] https://github.com/spulec/moto/blob/9f0f230d130a839fb2de6bfc97af4182360fbcdb/moto/awslambda/models.py#L135
[2] https://github.com/containers/podman/issues/6050

yan12125:
Podman 3.0 supports all needed Docker APIs, and only a few changes are needed in moto => #3702

bblommers:
This is now supported in moto >= 1.3.17.dev276

PR #3702, opened 2021-02-17
Support Podman for mocking Lambda

yan12125:
Podman supports all Docker APIs used in moto since version 3.0. Note
that Podman requires pulling the image before creating a container
using a fully-qualified image name (e.g., "docker.io/library/busybox"
instead of "busybox").

Test plan:
$ podman system service -t 0
$ DOCKER_HOST="unix://$XDG_RUNTIME_DIR/podman/podman.sock" pytest

Fixes https://github.com/spulec/moto/issues/3276

### DIFF
--- a/moto/awslambda/models.py
+++ b/moto/awslambda/models.py
@@ -575,7 +575,9 @@ class LambdaFunction(CloudFormationModel, DockerModel):
                         if settings.TEST_SERVER_MODE
                         else {}
                     )
-                    self.docker_client.images.pull("docker.io/lambci/lambda", tag=self.run_time)
+                    self.docker_client.images.pull(
+                        "docker.io/lambci/lambda", tag=self.run_time
+                    )
                     container = self.docker_client.containers.run(
                         "lambci/lambda:{}".format(self.run_time),
                         [self.handler, json.dumps(event)],
--- a/moto/awslambda/models.py
+++ b/moto/awslambda/models.py
@@ -51,7 +51,7 @@ from moto.sqs import sqs_backends
 from moto.dynamodb2 import dynamodb_backends2
 from moto.dynamodbstreams import dynamodbstreams_backends
 from moto.core import ACCOUNT_ID
-from moto.utilities.docker_utilities import DockerModel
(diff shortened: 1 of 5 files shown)


moto/awslambda/models.py
```diff
@@ -131,6 +131,7 @@ def __enter__(self):
                 volumes = {self.name: {"bind": "/tmp/data", "mode": "rw"}}
             else:
                 volumes = {self.name: "/tmp/data"}
+            self._lambda_func.docker_client.images.pull("docker.io/library/alpine")
```
bblommers, line 134: Let's use the parse_image_name method here as well, i.e. `parse_image_name("alpine")` and `parse_image_name("lambci/lambda")`

moto/utilities/docker_utilities.py
```diff
@@ -31,3 +31,17 @@ def replace_adapter_send(*args, **kwargs):
...
+    # podman does not support short container image name out of box - try to make a full name
+    if ":" in image_name:
+        image_repository, image_tag = image_name.split(":", 1)
+    else:
+        image_repository = image_name
+        image_tag = "latest"
+    if "/" not in image_repository:
+        image_repository = "library/" + image_repository
+    if len(image_repository.split("/")) < 3:
+        image_repository = "docker.io/" + image_repository
```
bblommers, line 46: Can you make the `docker.io` configurable, just in case people need to use a different repository? Ideally it's an env-variable in settings.py

bblommers (changes requested):

moto/utilities/docker_utilities.py
yan12125: Sure! I've also found a bug in this function. Will update both soon.

yan12125: Done! I added `$DEFAULT_CONTAINER_REGISTRY`. Please tell me if you want a better name.

moto/awslambda/models.py
yan12125: Done!

bblommers (approved):
LGTM. Thanks for the PR and making the changes, @yan12125!

PR #3705, opened 2021-02-17
feature/refresh_trusted_advisor_check

ccggeo:
Hi,

Have created another small PR to add refresh_trusted_advisor_check.

I am returned a random status from the response, please let me know if this is not ideal and I will work towards resolving.

Thanks!


bblommers (changes requested):
Ideally the returned status is deterministic, instead of random. 
Hi @ccggeo, thanks for the PR!

We can mock the advancement in statuses (none -> enqueued -> processing -> success), by advancing the status everytime the `refresh_trusted_advisor_check` method is called.

In pseudocode:
```
statuses = {}
if "check_id" not in statuses:
  statuses[check_id] = "none"
else:
  if statuses[check_id] == "none":
    statuses[check_id] = "enqueued"
  elif:
    etc...
return statuses[check_id]
```

Check how we do this in the Transcribe-status: https://github.com/spulec/moto/blob/9bc6bded6e6472f04a45f0bc429acc926c2cdc08/moto/transcribe/models.py#L260

bblommers (approved):
LGTM

bblommers:
This is now part of moto >= 1.3.17.dev279

Issue #3706, opened 2021-02-17
create_user_pool_domain not returning data if no cert config is given


baohx2000:
Moto version 1.3.16

Per AWS docs, you should be able to create a domain without supplying domain config (a certificate)
Doing so, AWS will just use their internal cognito domain to create the login page.

Related code:
https://github.com/spulec/moto/blob/master/moto/cognitoidp/models.py#L192-L198

The above code should probably be something like:

```python
    def _distribution_name(self):
        if self.custom_domain_config and "CertificateArn" in self.custom_domain_config:
            hash = hashlib.md5(
                self.custom_domain_config["CertificateArn"].encode("utf-8")
            ).hexdigest()
            return "{hash}.cloudfront.net".format(hash=hash[:16])
        hash = hashlib.md5(self.user_pool_id).hexdigest()
        return "{hash}.amazoncognito.com".format(hash=hash[:16])
```

I am not a python dev so above may not be right.

bblommers:
This is now part of moto >= 1.3.17.dev277

baohx2000:
@bblommers just curious - how long before a docker image will be tagged for .17?
No huge hurry since I'm just building from the patch branch for the moment.

PR #3707, opened 2021-02-17
pool domain should always return a domain

baohx2000:
Refs https://github.com/spulec/moto/issues/3706

### DIFF
--- a/tests/test_cognitoidp/test_cognitoidp.py
+++ b/tests/test_cognitoidp/test_cognitoidp.py
@@ -169,7 +169,6 @@ def test_describe_user_pool_domain():
     result["DomainDescription"]["Domain"].should.equal(domain)
     result["DomainDescription"]["UserPoolId"].should.equal(user_pool_id)
     result["DomainDescription"]["AWSAccountId"].should_not.be.none
-    result["DomainDescription"]["CloudFrontDomain"].should_not.be.none
 
 
 @mock_cognitoidp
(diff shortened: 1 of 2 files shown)


bblommers (changes requested):
Hi @baohx2000, thanks for the PR!
Would you mind adding a test to verify that this behaves as it should?

If I'm understanding the test case correctly, you can just add a single assertion to `tests/test_cognitoidp/test_cognitoidp.py::test_describe_user_pool_domain`

tests/test_cognitoidp/test_cognitoidp.py
```diff
@@ -137,6 +137,7 @@ def test_create_user_pool_domain():
     user_pool_id = conn.create_user_pool(PoolName=str(uuid.uuid4()))["UserPool"]["Id"]
     result = conn.create_user_pool_domain(UserPoolId=user_pool_id, Domain=domain)
     result["ResponseMetadata"]["HTTPStatusCode"].should.equal(200)
+    result["CloudFrontDomain"].should_not.be.none
```
bblommers, line 140: Typo - this should be `result["DomainDescription"]["CloudFrontDomain"]`

bblommers (changes requested):

tests/test_cognitoidp/test_cognitoidp.py
baohx2000: Kind of confused... that structure would not match the next test result
`result["CloudFrontDomain"].should.equal("e2c343b3293ee505.cloudfront.net")`

The test log is gigantic and takes forever to load on github. Will try to take a look again when I get to the office.

baohx2000: Fixed

bblommers (approved):
LGTM

PR #3708, opened 2021-02-17
dynamodb: fix deleting last set element (w/attr name)

ikonst:
When deleting last element(s) of a set, the attribute itself is cleared from the container. This is fixing support for specifying such attributes through attribute name placeholders in update expressions.

### DIFF
--- a/moto/dynamodb2/parsing/executors.py
+++ b/moto/dynamodb2/parsing/executors.py
@@ -165,8 +165,18 @@ class DeleteExecutor(NodeExecutor):
         # the last item in the set, we have to remove the attribute.
         if not string_set_list:
             element = self.get_element_to_action()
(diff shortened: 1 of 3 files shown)


bblommers (changes requested):
Hi @ikonst, thanks for the PR!
Can you add a user-facing unit test as well, to make it explicit how an end user would run into this scenario? I'm guessing it's just a slightly modified variant of the `test_dynamodb.py::test_set_attribute_is_dropped_if_empty_after_update_expression` test

ikonst:
@bblommers Thanks for the pointer! Added.

bblommers (approved):
LGTM. Thanks again, @ikonst!

bblommers:
This is now part of moto >= 1.3.17.dev280

PR #3709, opened 2021-02-18
Update stackset create operation to correctly return the resource's ID.

umgelurgel:
'FakeStackSet' does not have a 'stackset_id' attribute, the current implementation silently returns a blank ID.


bblommers (changes requested):
Hi @umgelurgel, thanks for the PR! Can you add a test for this, to verify that the ID is now returned properly?

umgelurgel:
Thanks for the quick response @bblommers ! Test added.

bblommers (approved):
LGTM

bblommers:
This is now part of 1.3.17.dev278

PR #3713, opened 2021-02-19
Fix stackset describe/delete. Fix stackset update.

umgelurgel:
StackSets can be described/deleted using their name or id (as per docs [here](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/cloudformation.html#CloudFormation.Client.describe_stack_set) and [here](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/cloudformation.html#CloudFormation.Client.delete_stack_set)). 

Updating StackSet parameters should support omitting providing a value and specifying a `UsePreviousValue` flag [instead](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/cloudformation.html#CloudFormation.Client.update_stack_set
), as it does for Stack updates. Refactored the mechanism for resolving the parameters to be shared by both FakeStacks and FakeStackSets.

### DIFF
--- a/moto/cloudformation/models.py
+++ b/moto/cloudformation/models.py
@@ -494,6 +494,8 @@ class CloudFormationBackend(BaseBackend):
 
     def get_stack_set(self, name):
         stacksets = self.stacksets.keys()
+        if name in stacksets:
+            return self.stacksets[name]
         for stackset in stacksets:
             if self.stacksets[stackset].name == name:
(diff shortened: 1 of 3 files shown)


bblommers (approved):
LGTM. Thanks @umgelurgel!

bblommers:
This is now part of moto >= 1.3.17.dev281

PR #3718, opened 2021-02-22
Implements core MediaConnect endpoints

jordan-dimov:
This PR aims to implement the following MediaConnect endpoints:

  * create_flow
  * describe_flow
  * delete_flow
  * list_flows
  * start_flow
  * stop_flow
  * tag_resource
  * list_tags_for_resource


jordan-dimov:
@bpandola Could you please help me figure out why unit tests are failing? I don't understand the error message about __pycache__

bblommers (changes requested):
Hi @jordan-dimov, you might need to add an __init__.py file in the test_mediaserver-directory

jordan-dimov:
That fixed it, thank you @bblommers !

moto/mediaconnect/responses.py
```diff
@@ -0,0 +1,88 @@
...
+        next_token = self._get_param("nextToken")
+        flows, next_token = self.mediaconnect_backend.list_flows(
+            max_results=max_results, next_token=next_token,
+        )
+        return json.dumps(dict(flows=flows, nextToken=next_token))
+
+    def describe_flow(self):
+        flow_arn = self._get_param("flowArn")
+        if flow_arn:
+            flow_arn = flow_arn.replace("%3A", ":")
```
bblommers, line 46: It looks like this can simply be replace with `unquote(arn)`
```
try:
    from urllib import unquote
except ImportError:
    from urllib.parse import unquote
```
That's how ARN's are decoded in other services.

Do we need the if-statement? It looks like the param should always be there, and if it isn't, we should fail anyway - so it might as well become `flow_arn = unquote(self._get_param("flowArn"))`

moto/mediaconnect/models.py
```diff
@@ -0,0 +1,191 @@
...
+            availability_zone=availability_zone,
+            entitlements=entitlements,
+            name=name,
+            outputs=outputs,
+            source=source,
+            source_failover_config=source_failover_config,
+            sources=sources,
+            vpc_interfaces=vpc_interfaces,
+        )
+        flow.description = "A Moto test flow"
```
bblommers, line 105: Just curious - what's the value that AWS outputs here?
Strange that there's no way to pass this during creation

moto/mediaconnect/models.py
```diff
@@ -0,0 +1,191 @@
...
+            return new_data
+        return data
+
+    def resolve_transient_states(self):
+        if self.status in ["STARTING"]:
+            self.status = "ACTIVE"
+        if self.status in ["STOPPING"]:
+            self.status = "STANDBY"
+        if self.status in ["UPDATING"]:
+            self.status = self._previous_status
```
bblommers, line 55: `self._previous_status` doesn't seem to be used anywhere else.
Wouldn't the status become "ACTIVE" after it's finished being updated? (I've never used this service, but that seems like the best value to use)

bblommers: I suppose it's not really relevant now, only after we implement `update_flow()`

bblommers (changes requested):

moto/mediaconnect/responses.py
jordan-dimov: You're right. I've made this change. Thank you!

moto/mediaconnect/models.py
jordan-dimov: I wondered about this too. Seems to be inconsistent. AWS docs say: "This description appears only on the MediaConnect console and is not visible outside of the current AWS account. "

So I'm assuming the API doesn't support setting the description in the request, but you can set it manually from the AWS console, and if you do, you'll see it in the Describe response. 

On the other hand, for Entitlements, Sources and Outputs of a flow you _can_ provide descriptions in the request.

moto/mediaconnect/models.py
jordan-dimov: Correct, that's only necessary for the `update_` operations (which I am not going to implement in this PR, as I have no need for them currently). 

I did something similar for MediaLive, where I _did_ implement the `update_channel()` method. The reason I needed to keep track of `_previous_status` there is because it was important to return to whatever the channel state was before the update - and it could be either `IDLE` or `RUNNING`, because you can update a channel before or after it is started. 

Likewise, for MediaConnect, I'm expecting that you can perform update operations before or after starting a flow, so the `_previous_status` could potentially be one of `STANDBY` or `ACTIVE`.

bblommers (approved):
LGTM. Thanks again for the PR, @jordan-dimov!

PR #3723, opened 2021-02-23
Add the status field to ECS task definitions

william-richard:
Also I found extra tests for describe_task_definition and deregister_task_definition that were not being run,
so I changed their names so they are found by pytest and made them pass.  I also added checks to them for the new
status field.


bblommers (approved):
LGTM. Thanks @william-richard!

bblommers:
This is now part of moto >= 2.0.1.dev287

PR #3724, opened 2021-02-23
Fixes for IAM Role Description field in responses from list_roles and create_roles

jbmchuck:
`list_roles` and `create_roles` should return the `Description` field if it is non-null.

### DIFF
--- a/tests/test_iam/test_iam.py
+++ b/tests/test_iam/test_iam.py
@@ -4043,13 +4043,11 @@ def test_list_roles_with_description():
     # Ensure the Description is included in role listing as well
     conn.list_roles().get("Roles")[0].get("Description").should.equal(description)
 
+
 @mock_iam()
 def test_list_roles_without_description():
     conn = boto3.client("iam", region_name="us-east-1")
-    resp = conn.create_role(
-        RoleName="my-role",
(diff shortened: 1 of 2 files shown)


tests/test_iam/test_iam.py
```diff
@@ -4029,6 +4029,46 @@ def test_list_roles_none_found_returns_empty_list():
     assert len(roles) == 0
 
 
+@mock_iam()
+def test_list_roles_with_description():
```
bblommers, line 4033: The logic seems to be the same as `test_list_roles_with_empty_string_description` - would you mind parameterizing this method to cover both cases? IMO that would be easier to maintain
```
@pytest.mark.parametrize("desc", ["", "Test Description"])
def test_list_roles_with_description(desc):
```

bblommers (changes requested):
Hi @jbmchuck, thanks for the PR! Please see my comment - happy to merge once that's addressed

tests/test_iam/test_iam.py
jbmchuck: Thanks - updated.

bblommers (approved):
LGTM

PR #3734, opened 2021-02-26
Fix:update rule

usmangani1:
Fix:update rule

https://github.com/localstack/localstack/issues/3652


moto/events/models.py
```diff
@@ -492,10 +492,22 @@ def list_targets_by_rule(self, rule, next_token=None, limit=None):
 
         return return_obj
 
+    def update_rule(self, rule, **kwargs):
+        rule.event_pattern = kwargs.get("EventPattern") or rule.event_pattern
+        rule.schedule_exp = kwargs.get("ScheduleExpression") or rule.schedule_exp
+        rule.state = kwargs.get("State") or rule.state
+        rule.description = kwargs.get("Description") or rule.description
+        rule.role_arn = kwargs.get("RoleArn") or rule.role_arn
+        rule.event_bus_name = kwargs.get("EventBusName", "default") or rule.event_bus_name
```
bblommers, line 501: The or-statement will never be hit, as the expression will always become 'default'. Need to use kwargs.get() without the if_none parameter

bblommers (changes requested):
Hi @usmangani1, thanks for the PR! Got one comment, and it looks like the linter fails as well.

bblommers (approved):
LGTM - thanks!

bblommers:
This is now part of moto >= 2.0.1.dev297

PR #3735, opened 2021-02-27
Add eventbridge replay

gruebel:
Added the needed endpoints for the replay functionality. The state is quite often faked, because the real replay had to be done asynchron, which I didn't want to implement. I also implemented the possibility to send events to a Log Group target.


bblommers (approved):
LGTM - thanks @gruebel!

bblommers:
This is now part of moto >= 2.0.1.dev295

PR #3736, opened 2021-02-28
Fix AttributeNames for sqs.receive_message

### LINKED ISSUE #3607: sqs receive_message returns AttributeNames when not requested
This behavior deviates from the [boto3 documentation](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/sqs.html#SQS.Client.receive_message). No default option is listed for `AttributeNames`, however moto returns, unconditionally, attributes such as `ApproximateReceiveCount`. This leads to false positive unit tests in which attributes are relied upon, but are not available when interacting without moto.

https://github.com/spulec/moto/blob/2bdba88389e5c3ee83bcbf719bbbaa1da0f81ede/moto/sqs/responses.py#L549

python 3.6
moto== 1.3.16
boto==2.49.0
boto3=1.16.31
botocore=1.19.36

Below are two test cases:
`test_default_behavior` currently fails and **should not** according to boto3 execution of the same call.
`test_requested_message_attributes` currently passes and **should** according to boto3 execution of the same call.

```
import unittest
import boto3
from moto import mock_sqs

@mock_sqs
class TestIssue(unittest.TestCase):
    def test_default_behavior(self):
        sqs = boto3.resource("sqs")
        queue = sqs.create_queue(QueueName="a_queue")
        queue.send_message(MessageBody="health", DelaySeconds=0)
        session = boto3.session.Session()
        sqs_client = session.client("sqs")
        response = sqs_client.receive_message(
            QueueUrl=queue.url,
            MaxNumberOfMessages=1,
            # AttributeNames=[
            #     'ApproximateReceiveCount',
            # ],
            VisibilityTimeout=30,
            WaitTimeSeconds=2,
        )
        message = response["Messages"][0]
        self.assertEqual("health", message["Body"])
        self.assertIsNone(message["Attributes"])

    def test_requested_message_attributes(self):
        sqs = boto3.resource("sqs")
        queue = sqs.create_queue(QueueName="a_queue")
        queue.send_message(MessageBody="health", DelaySeconds=0)
        session = boto3.session.Session()
        sqs_client = session.client("sqs")
        response = sqs_client.receive_message(
            QueueUrl=queue.url,
            MaxNumberOfMessages=1,
            AttributeNames=[
                "ApproximateReceiveCount",
            ],
            VisibilityTimeout=30,
            WaitTimeSeconds=2,
        )
        message = response["Messages"][0]
        self.assertEqual("health", message["Body"])
        self.assertIsNotNone(message["Attributes"])
        self.assertIsNotNone(message["Attributes"]["ApproximateReceiveCount"])
        self.assertEqual("1", message["Attributes"]["ApproximateReceiveCount"])
```

gruebel:
Fixes #3607

### DIFF
--- a/moto/sqs/responses.py
+++ b/moto/sqs/responses.py
@@ -354,7 +354,9 @@ class SQSResponse(BaseResponse):
         queue_name = self._get_queue_name()
         message_attributes = self._get_multi_param("message_attributes")
         if not message_attributes:
-            message_attributes = extract_input_message_attributes(self.querystring,)
+            message_attributes = extract_input_message_attributes(self.querystring)
+
+        attribute_names = self._get_multi_param("AttributeName")
 
         queue = self.sqs_backend.get_queue(queue_name)
 
@@ -394,7 +396,12 @@ class SQSResponse(BaseResponse):
             return ERROR_MAX_VISIBILITY_TIMEOUT_RESPONSE, dict(status=400)
 
         messages = self.sqs_backend.receive_messages(
-            queue_name, message_count, wait_time, visibility_timeout, message_attributes
+            queue_name,
+            message_count,
+            wait_time,
+            visibility_timeout,
+            message_attributes,
+            attribute_names,
         )
         template = self.response_template(RECEIVE_MESSAGE_RESPONSE)
         return template.render(messages=messages)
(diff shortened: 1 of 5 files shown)


bblommers (changes requested):
Hi @gruebel, thanks for the PR! Would you mind adding `MessageDeduplicationId ` and `MessageGroupId` as well?

moto/sqs/responses.py
```diff
@@ -396,8 +398,30 @@ def receive_message(self):
...
         )
+
+        attributes = {
+            "approximate_first_receive_timestamp": False,
+            "approximate_receive_count": False,
+            "sender_id": False,
+            "sent_timestamp": False,
+        }
+
+        if "All" in attribute_names:
```
bblommers, line 409: These if-statements can probably become a for-loop, using the `moto.utils.underscore_to_camelcase` method to streamline the conversion

tests/test_sqs/test_sqs.py
```diff
@@ -796,6 +801,134 @@ def test_send_receive_message_timestamps():
...
+    message1["Attributes"].should.have.key("SenderId")
+    message1["Attributes"].should.have.key("SentTimestamp")
+    message2["Attributes"].should.have.key("ApproximateFirstReceiveTimestamp")
+    message2["Attributes"]["ApproximateReceiveCount"].should.equal("1")
+    message2["Attributes"].should.have.key("SenderId")
+    message2["Attributes"].should.have.key("SentTimestamp")
+
+
+@mock_sqs
+def test_send_receive_message_with_attribute_name_receive_time():
```
bblommers, line 841: Don't know how easy it would be to combine all these methods with a pytest.mark.parametrize?
There is a lot of duplication here

bblommers (changes requested):

tests/test_sqs/test_sqs.py
gruebel: good point, I will give it a try

gruebel:
@bblommers please have a look regarding the parameterized tests.

moto/sqs/responses.py
```diff
@@ -396,8 +398,32 @@ def receive_message(self):
...
+            camelcase_name = camelcase_to_underscores(attribute_name)
+            if camelcase_name == "all":
+                attributes["approximate_first_receive_timestamp"] = True
+                attributes["approximate_receive_count"] = True
+                attributes["message_deduplication_id"] = True
+                attributes["message_group_id"] = True
+                attributes["sender_id"] = True
+                attributes["sent_timestamp"] = True
+                attributes["sequence_number"] = True
+            elif camelcase_name in attributes:
```
bblommers, line 422: If I'm reading this right, the if/else can be simplified:
```
if camelcase_name == "all" or camelcase_name in attributes:
    attributes[camelcase_name] = True
```

bblommers (changes requested):
The tests look good to me @gruebel - thanks

gruebel:
#3751 after the fix is merged, this is good to go!

bblommers (approved):
LGTM - thanks for making the changes!

bblommers:
This is now part of moto >= 2.0.2.dev5

PR #3742, opened 2021-03-02
Fix:Add functionality authorize-cluster-security-group-ingress

usmangani1:
Fix:Add functionality authorize-cluster-security-group-ingress

https://github.com/localstack/localstack/issues/3637

### DIFF
--- a/moto/redshift/models.py
+++ b/moto/redshift/models.py
@@ -28,6 +28,7 @@ from .exceptions import (
     SnapshotCopyGrantAlreadyExistsFaultError,
     SnapshotCopyGrantNotFoundFaultError,
     UnknownSnapshotCopyRegionFaultError,
+    ClusterSecurityGroupNotFoundFaultError,
 )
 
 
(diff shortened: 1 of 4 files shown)


moto/redshift/models.py
```diff
@@ -749,6 +751,16 @@ def delete_cluster_security_group(self, security_group_identifier):
             return self.security_groups.pop(security_group_identifier)
         raise ClusterSecurityGroupNotFoundError(security_group_identifier)
 
+    def authorize_cluster_security_group_ingress(self, security_group_name, cidr_ip):
+        security_group = self.security_groups[security_group_name]
+        if not security_group:
+            raise ClusterSecurityGroupNotFoundFaultError("")
```
bblommers, line 757: Can we get a unit test for this scenario? I'm guessing the message should have a value as well - would be good to assert on that

bblommers (changes requested):

bblommers (approved):
LGTM - thanks!

bblommers:
This is now part of moto >= 2.0.2.dev6

PR #3751, opened 2021-03-05
Fix Osaka region problem

gruebel:
A few days ago Osaka became a regular AWS region, so I added the missing AZs.

https://aws.amazon.com/about-aws/whats-new/2021/03/announcing-the-newly-expanded-aws-asia-pacific-osaka-region/


gruebel:
@bblommers Would be great if you could have a look, all the tests are now failing.

bblommers (approved):
Nice one! Thanks @gruebel

bblommers:
This is fixed as of moto >= 2.0.2.dev5

PR #3752, opened 2021-03-05
Feature: Kinesis - list_shards()

### LINKED ISSUE #1635: Kinesis List Shards Endpoint
Hello, I would like to implement the list_shards() endpoint for moto, if nobody has started on it yet.  I notice it hasn't been checked in the IMPLMENTATION_COVERAGE document.

bblommers:
Fixes #1635

The StepFunctions-module has a `@paginate`-decorator - most of this PR involves extracting and reusing tthis decorator across multiple services.

Changes:
 - `moto/utilities/Paginator` now contains a globally useable decorator for pagination
 - CognitoIDP now uses this decorator, instead of it's own
 - Logs now uses this decorator, instead of a per-method implementation
 - And finally, actually implemented the Kinesis::list_shard functionality with pagination


tests/test_kinesis/test_kinesis_boto3.py
```diff
@@ -0,0 +1,140 @@
+import boto3
+
+from moto import mock_kinesis
+
+import sure  # noqa
+
+
+@mock_kinesis
+def test_split_shard():
```
bblommers, line 9: There was a boto-test for this functionality, but not boto3 yet

tests/test_logs/test_logs.py
```diff
@@ -494,3 +492,48 @@ def test_describe_log_groups_paging():
     resp = client.describe_log_groups(nextToken="invalid-token")
     resp["logGroups"].should.have.length_of(0)
     resp.should_not.have.key("nextToken")
+
+
+@mock_logs
+def test_describe_log_streams_simple_paging():
```
bblommers, line 498: There was no testss at all for pagination in describe_log_streams(). I created the test first, and then switched to the global decorator to verify the implementation hasn't changed

moto/cognitoidp/utils.py
```diff
@@ -7,6 +7,34 @@
 import base64
 
 
+PAGINATION_MODEL = {
```
bblommers, line 10: This model was used in SFN first, but can be reused for every service like this:
```
method:
    input_token: name of the method-kwarg that contains the token
    limit_key: name of the method-kwarg that contains the limit/max/nr of results the user asked for
    page_ending_range_keys: attributes used to create unique hash for the item, required to identify the starting point for the next token

File: moto/kinesis/models.py [hunk 4]
```diff
@@ -574,6 +576,11 @@ def merge_shards(self, stream_name, shard_to_merge, adjacent_shard_to_merge):
                 record.partition_key, record.data, record.explicit_hash_key
             )
 
+    @paginate(pagination_model=PAGINATION_MODEL)
+    def list_shards(self, stream_name, limit=None, next_token=None):
+        stream = self.describe_stream(stream_name)
+        return [shard.to_json() for shard in stream.shards.values()]
```
bblommers, line 582: The model can simply return the entire list of (applicable) items.
The decorator will convert this into a `(curated_item_list, token)`-tuple

moto/utilities/Paginator.py
```diff
@@ -0,0 +1,139 @@
...
+        if self._param_checksum:
+            token_dict["parameterChecksum"] = self._param_checksum
+        range_keys = []
+        for (index, attr) in enumerate(self._page_ending_range_keys):
+            if type(next_item) == dict:
+                range_keys.append(next_item[attr])
+            else:
+                range_keys.append(getattr(next_item, attr))
+        token_dict["pageEndingRangeKey"] = "|".join(range_keys)
+        return TokenEncoder().encode(token_dict)
```
bpandola, line 112: This should be `self._token_encoder.encode(token_dict)`.  (A minor oversight from back when I first submitted this code.)

moto/utilities/Paginator.py
```diff
@@ -0,0 +1,139 @@
+from functools import wraps
```
bpandola, line 1: Any particular reason why this filename is capitalized?

bblommers: Oversight - thanks for pointing it out

moto/utilities/Paginator.py
bblommers: Fixed

tests/test_logs/test_logs.py
```diff
@@ -480,11 +478,11 @@ def test_describe_log_groups_paging():
 
     resp = client.describe_log_groups(limit=2)
     resp["logGroups"].should.have.length_of(2)
-    resp["nextToken"].should.equal("/aws/lambda/FileMonitoring")
```
bpandola, line 483: The `nextToken` value here is just the `logGroupName` when using a real AWS backend.  I don't know if it matters for moto or not... pagination will still work, but the implementation will not match AWS.  I've given some thought to a truly generic paginator for moto, one that does all the heavy lifting but defers some implementation details to the individual services.  I can't say I'm close to submitting anything for review, but it should be possible, which would be nice.  What you're doing here in pulling this out into the utilities module is certainly a start.

bblommers: That's annoying - I assumed the `logGroupName`was just used as token because it's the easiest implementation.

I suppose the easy way would be to extend the pagination_token model with new `hash_value`/`is_hashed` configuration options, but that's not a pretty solution.

PR #3753, opened 2021-03-05
Fix:SES-Added params check in template

usmangani1:
Fix:SES-Added params check in template

https://github.com/localstack/localstack/issues/3664

### DIFF
--- a/tests/test_ses/test_ses_boto3.py
+++ b/tests/test_ses/test_ses_boto3.py
@@ -516,6 +516,14 @@ def test_render_template():
     result["RenderedTemplate"].should.contain("<h1>Hello John,</h1>")
     result["RenderedTemplate"].should.contain("Your favorite animal is Lion")
 
+    kwargs = dict(
+        TemplateName="MyTestTemplate", TemplateData=json.dumps({"name": "John"}),
+    )
+
+    with pytest.raises(ClientError) as ex:
+        conn.test_render_template(**kwargs)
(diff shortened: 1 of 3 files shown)


tests/test_ses/test_ses_boto3.py
```diff
@@ -516,6 +516,14 @@ def test_render_template():
     result["RenderedTemplate"].should.contain("<h1>Hello John,</h1>")
     result["RenderedTemplate"].should.contain("Your favorite animal is Lion")
 
+    kwargs = dict(
+        TemplateName="MyTestTemplate", TemplateData=json.dumps({"name": "John"}),
+    )
+
+    with pytest.raises(ClientError) as ex:
+        conn.test_render_template(**kwargs)
+    assert ex.value.response["Error"]["Code"] == "MissingRenderingAttributeException"
```
bblommers, line 525: Can we get an assert on the message as well?
Just tested it out against AWS, the message it should give is:
`Attribute 'favoriteanimal' is not present in the rendering data.`

(If there are multiple attributes missing, it will only mention the first one)

bblommers (changes requested):
Hi @usmangani1, thanks for the PR!
Can we change the template a bit to include random spaces in the vars? This is acceptable with AWS, and I want to make sure that the regex substitution still works like that:
```
"Dear {{ name }},"
"\r\nYour favorite animal is {{favoriteanimal  }}.",
```

bblommers (approved):
LGTM - thanks!

Issue #3748, opened 2021-03-04
Creating an auto scaling group removes  the previously existing instances from the load balancer


gipi:
I'm experiencing a weird behaviour with ``moto`` trying to unit test some fabric scripts used to manage code on machines under a load balancer: for the real setup I'm using ``ansible`` and i'm creating a couple of machines, registering them when the load balancer is created; at this point I'm also associating an autoscaling group to the same load balancer. On the real AWS, querying with ``describe_load_balancers()`` the same load balancer I obtain two sets of machines, each one associated with a "Reservation".

With ``moto`` instead the ``create_auto_scaling_group()`` removes the previous instances and I obtain only one set of machines, i.e. the machines associated with the autoscaling group.

It's not clear to me if it's really a bug of ``moto`` but since this is something that happens with the **real** services I thought was worthy a shot and someone can help me.

By the way this is the code used for ansible

```yaml
- name: Provision a set of instances for web
  ec2:
    region: eu-west-1
    key_name: ssh_xxx
    group:
      - "{{ security_group_webapp.group_name }}"
      - "{{ security_group_ssh.group_name }}"
    instance_type: t2.micro
    image: "ami-xxx"
    vpc_subnet_id: "{{ subnet.subnet.id }}"
    assign_public_ip: true
    wait: true
    exact_count: 1
    instance_tags:
      Name: xxx
  register: ec2

- name: Extract EC2 instances ids
  set_fact:
    ec2_ids: "{{ ec2_ids | default([]) |  union([item.id]) }}"
  with_items: "{{ ec2.tagged_instances }}"

- name: Setup ELB
  ec2_elb_lb:
    name: "xxx-elb"
    state: present
    region: "eu-west-1"
    subnets: "{{ subnet.subnet.id }}"
    instance_ids: "{{ ec2_ids }}"
    security_group_ids:
      - "{{ security_group_webapp.group_id }}"
    # validate_certs: no
    listeners:
      - protocol: https # options are http, https, ssl, tcp
        load_balancer_port: 443
        instance_port: 443
        ssl_certificate_id: "arn:aws:iam::027188111508:server-certificate/xxx"
    health_check:
      ping_protocol: https
      ping_port: 443
      ping_path: "/"
      response_timeout: 5 # seconds
      interval: 30 # seconds
      unhealthy_threshold: 2
      healthy_threshold: 2
    wait: yes
    tags:
      Environment: test
      Application: xxx

 - name: Create launch config
   community.aws.ec2_lc:
[...]

PR #3757, opened 2021-03-08
Fix/autoscaling elb interaction

gipi:
This is a first attempt to fix issue #3748 in a pretty hacky way: in the original code the assumption is that all the instances attached to a load balancer are only by an autoscaling group, so when the group is modified it acts on the load balancer so to have only its own instances attached, but this is not how the ELB behaves, you can manually attach isolated machines that remain independent by the group.

For now I'm using an internal attribute to take note of which instances are intended to be deregistered from the load balancer but I'm pretty sure it's not a very good way of do it, maybe I'm missing some subtle point about instances, however the tests seem to pass. So if someone more experienced with the code want to give me feedback I can try to improve it.

I also improved a test to catch the bug, if it's ok I'm going to extend to other tests involving this service.

### DIFF
--- a/moto/autoscaling/models.py
+++ b/moto/autoscaling/models.py
@@ -946,10 +946,10 @@ class AutoScalingBackend(BaseBackend):
         for elb in elbs:
             elb_instace_ids = set(elb.instance_ids)
             self.elb_backend.register_instances(
-                elb.name, group_instance_ids - elb_instace_ids
+                elb.name, group_instance_ids - elb_instace_ids, from_autoscaling=True,
             )
             self.elb_backend.deregister_instances(
-                elb.name, elb_instace_ids - group_instance_ids
+                elb.name, elb_instace_ids - group_instance_ids, from_autoscaling=True,
             )
 
     def update_attached_target_groups(self, group_name):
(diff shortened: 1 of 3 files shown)


bblommers (changes requested):
Hi @gipi, thanks for the PR!
IMO this would be cleaner if we store two separate sets of instance_ids per ELB. One set of 'native' instances, registered directly, and one set of instances controlled by the ASG. 
The autoscaling-service will only have to touch the ASG-controlled instances, and we don't have to worry about accidentally removing any instances it doesn't have control over.

gipi:
@bblommers ok makes sense but this means we should create two separate "family" of de/register calls, one for group and one for the other in the ELB and to me seems a big change.

bblommers:
I'd be in favor of the keyword argument. As long as it has a sensible default, and the only change is in models.py, there should be no impact.

gipi:
I tried the new approach; it needs some adjustments but seems cleaner than before: I split the instance ids in "sparse" (probably a better naming would be "native") and "from autoscaling group" and added the ``from_autoscaling`` as a keyword argument in the necessary methods.

bblommers (changes requested):
Thanks for the reminder @gipi. The ServerMode tests are failing, probably because of the fact that the tests now combine `deprecated` and normal decorators.
Might have to rewrite the test to use only deprecated, or only normal, decorators

bblommers (approved):
LGTM! Thanks @gipi

PR #3759, opened 2021-03-09
Organisations - Detach policy, and asserts it actually happens

bblommers:
As pointed out by @dpowley in #3278:
Detaching a policy was broken, and the existing test didn't actually assert that the detachment was successful.

### DIFF
--- a/moto/organizations/models.py
+++ b/moto/organizations/models.py
@@ -576,7 +576,7 @@ class OrganizationsBackend(BaseBackend):
         ).match(kwargs["TargetId"]):
             ou = next((ou for ou in self.ou if ou.id == kwargs["TargetId"]), None)
             if ou is not None:
-                if ou not in ou.attached_policies:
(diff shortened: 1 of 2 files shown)


dpowley (approved):
LGTM - Thanks for the quick fix

dpowley:
@bblommers What timeline can I expect for a release with this bug fix included? We would use this on day one 😄

bblommers:
@dpowley Probably one or two days after approval from a maintainer! No timeline on when the approval comes in though

dpowley:
> @dpowley Probably one or two days after approval from a maintainer! No timeline on when the approval comes in though

@bblommers I saw there have been a few releases since this one opened up. Any chance we can get this fix out the door soon? It also causes issues with testing DeletePolicy. If the policy is attached to a target, you can't delete the policy without first detaching it from the targets. And there is no way to detach the policy currently (this PR fixes the detachment bug).

PR #3762, opened 2021-03-10
Implementation of core AWS Mediapackage endpoints

FranckNdame:


jordan-dimov:
@bpandola @bblommers  Just FYI, this is a continuation of the AWS Media Services we started contributing in earlier PRs for MediaLive and MediaConnect. This is the corresponding one for MediaPackage.

moto/mediapackage/models.py
```diff
@@ -0,0 +1,202 @@
...
+        manifest_name,
+        mss_package,
+        origination,
+        startover_window_seconds,
+        tags,
+        time_delay_seconds,
+        whitelist,
+    ):
+        arn = "arn:aws:mediapackage:origin_endpoint:{}".format(id)
+        url = "https://origin-endpoint.mediapackage.eu_west_1.amazonaws.com/{}".format(
```
bblommers, line 126: I'm assuming the region in this URL will be dependent on the specified region?
Would be good to have an assertion in one of the tests as well, to verify the URL is returned

bblommers (changes requested):

moto/mediapackage/models.py
bblommers: It's not dependent on region yet - it should be
```
"https://origin-endpoint.mediapackage.{}.amazonaws.com/{}".format(self.region_name, id)
```

jordan-dimov: @bblommers About that - for some reason `self.region_name` ends up being None when we run our tests? Do you know why that may be? 

Also, just as a clarification, that URL is a bit arbitrary and not really used by anything, we just needed a string that kind of looks like a URL.

bblommers: In models.py, you have to pass the region when creating the Backend per region:
`mediapackage_backends[region] = MediaPackageBackend()` -> `mediapackage_backends[region] = MediaPackageBackend(region)`

AnyaChamp: Thank you, I hadn't realised this. I have made the change now.

bblommers (approved):
LGTM. Thanks @AnyaChamp, @bpandola, @FranckNdame!

bblommers:
This is now part of moto >= 2.0.2.dev10

PR #2357, opened 2019-08-06
Existing user now raises a UsernameExistsException

brodie-g:
If a user is attempted to be added to a pool that already contains a user with that username, the UsernameExistsException is thrown, to match AWS behaviour.

resolves https://github.com/spulec/moto/issues/2356

PR #3765, opened 2021-03-11
Cognito IdP: Raise UsernameExistsException from sign_up when user exists

sawyerh:
Similar to `admin_create_user` (#2357), the `sign_up` method in the Cognito identity provider client should raise a `UsernameExistsException` if a user already exists within the user pool

### DIFF
--- a/tests/test_cognitoidp/test_cognitoidp.py
+++ b/tests/test_cognitoidp/test_cognitoidp.py
@@ -1613,15 +1613,10 @@ def test_sign_up_existing_user():
     username = str(uuid.uuid4())
     password = str(uuid.uuid4())
 
-    conn.sign_up(ClientId=client_id, Username=username, Password=password)
-
-    caught = False
-    try:
+    with pytest.raises(ClientError) as err:
         conn.sign_up(ClientId=client_id, Username=username, Password=password)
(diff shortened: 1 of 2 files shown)


tests/test_cognitoidp/test_cognitoidp.py
```diff
@@ -1603,6 +1603,27 @@ def test_sign_up():
...
+    client_id = conn.create_user_pool_client(
+        UserPoolId=user_pool_id, ClientName=str(uuid.uuid4()),
+    )["UserPoolClient"]["ClientId"]
+    username = str(uuid.uuid4())
+    password = str(uuid.uuid4())
+
+    conn.sign_up(ClientId=client_id, Username=username, Password=password)
+
+    caught = False
+    try:
```
bblommers, line 1619: Can we use the pytest.raises construct here, instead of a try-except? That's how most of our negative tests are handled.
```
        ...
err["Code"].should.equal("...")
err["Message"].should.equal(...)
```

bblommers (changes requested):
Hi @sawyerh, thanks for the PR!

bblommers (approved):
LGTM - thanks @sawyerh!

bblommers:
This is now part of moto >= 2.0.3.dev4

Issue #3766, opened 2021-03-12
untag_resource support in secretsmanager


GeorgeVince:
Thanks for creating such a great library!

It would be great if there was support  for untagging resources in `secretsmanager` - I'll go ahead and make a PR.

bblommers:
This is now part of moto >= 2.0.2.dev9

PR #3767, opened 2021-03-12
Secrets manager untag resource

GeorgeVince:
Ticket issue: #3766


bblommers (changes requested):
Hey @GeorgeVince, thanks for the PR!

Looks like the linter is failing. Make sure to use the right version of black for this: moto uses black==19.10b0. 
(Later versions have some bugs in them.)

bblommers (approved):
LGTM - Thanks again!

PR #3770, opened 2021-03-14
Guardduty create+list detectors

juicySec:
Adding the create_detector and list_detectors with appropriate outputs and tests

### DIFF
--- a/moto/guardduty/models.py
+++ b/moto/guardduty/models.py
@@ -3,6 +3,7 @@ from boto3 import Session
 from moto.core import BaseBackend, BaseModel
 from datetime import datetime
 from uuid import uuid4
+from .exceptions import MissingParameterError
 
 
 class GuardDutyBackend(BaseBackend):
@@ -24,7 +25,11 @@ class GuardDutyBackend(BaseBackend):
             "ONE_HOUR",
             "SIX_HOURS",
         ]:
-            return None
+            finding_publishing_frequency = "SIX_HOURS"
+
+        if enable is None or "":
+            raise MissingParameterError(parameter="enabled")
+
         service_role = "AWSServiceRoleForAmazonGuardDuty"
         print(finding_publishing_frequency)
         detector = Detector(
--- a/moto/guardduty/models.py
+++ b/moto/guardduty/models.py
@@ -1,0 +1,77 @@
+from __future__ import unicode_literals
+from boto3 import Session
+from moto.core import BaseBackend, BaseModel
+from datetime import datetime
+from uuid import uuid4
+
+
+class GuardDutyBackend(BaseBackend):
+    def __init__(self, region_name=None):
+        super(GuardDutyBackend, self).__init__()
+        self.region_name = region_name
+        self.detectors = {}
+
+    def reset(self):
+        region_name = self.region_name
+        self.__dict__ = {}
+        self.__init__(region_name)
+
+    def create_detector(
+        self, enable, client_token, finding_publishing_frequency, data_sources, tags
+    ):
+        if finding_publishing_frequency not in [
+            "FIFTEEN_MINUTES",
+            "ONE_HOUR",
+            "SIX_HOURS",
+        ]:
+            return None
+        service_role = "AWSServiceRoleForAmazonGuardDuty"
+        print(finding_publishing_frequency)
+        detector = Detector(
+            self,
+            datetime.now,
+            finding_publishing_frequency,
+            service_role,
+            enable,
+            data_sources,
+            tags,
+        )
+        self.detectors[detector.id] = detector
+        return detector.id
+
+    def list_detectors(self, maxResults, nextToken):
+        detectorids = []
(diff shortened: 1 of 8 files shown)


bblommers:
Hi @juicySec, looks like the tests in ServerMode were failing.  Hope you don't mind, but I made some changes that should get them to pass.
Looks like the GuardDuty-service is a bit picky, so my change makes it more explicit how requests are recognized/handled.

moto/guardduty/models.py
```diff
@@ -0,0 +1,82 @@
...
+    ):
+        if finding_publishing_frequency not in [
+            "FIFTEEN_MINUTES",
+            "ONE_HOUR",
+            "SIX_HOURS",
+        ]:
+            finding_publishing_frequency = "SIX_HOURS"
+
+        if enable is None or "":
+            raise MissingParameterError(parameter="enabled")
```
bblommers, line 31: Would be good to have a test for this scenario as well, using the pytest.raises paradigm ([Example](https://github.com/spulec/moto/blob/master/tests/test_cognitoidp/test_cognitoidp.py#L607))

moto/guardduty/models.py
```diff
@@ -0,0 +1,82 @@
...
+        )
+        self.detectors[detector.id] = detector
+        return detector.id
+
+    def list_detectors(self, maxResults, nextToken):
+        detectorids = []
+        for detector in self.detectors:
+            detectorids.append(self.detectors[detector].id)
+            nextToken = str(maxResults) if len(detectorids) > maxResults else None
+            return detectorids, nextToken
```
bblommers, line 52: If I'm reading this right, it would only return the first detector ID

moto/guardduty/responses.py
```diff
@@ -0,0 +1,58 @@
...
+        client_token = self._get_param("clientToken")
+        finding_publishing_frequency = self._get_param("findingPublishingFrequency")
+        data_sources = self._get_param("dataSources")
+        tags = self._get_param("tags")
+
+        detector_id = self.guardduty_backend.create_detector(
+            enable, client_token, finding_publishing_frequency, data_sources, tags
+        )
+
+        return 200, {}, json.dumps(dict(detectorId=detector_id))
```
bblommers, line 44: With my change, each method now needs to return a tuple in the form of (HTTP Status Code, Headers, Body (as a string)). Bit more code, but it does make it easier to fine-tune the response if required

tests/test_guardduty/test_guardduty.py
```diff
@@ -0,0 +1,33 @@
...
+def test_create_detector():
+    client = boto3.client("guardduty", region_name="us-east-1")
+    response = client.create_detector(
+        Enable=True,
+        ClientToken="745645734574758463758",
+        FindingPublishingFrequency="ONE_HOUR",
+        DataSources={"S3Logs": {"Enable": True}},
+        Tags={},
+    )
+    assert response["DetectorId"] != None
```
bblommers, line 19: Can we get a few assertions in both tests?

moto/guardduty/responses.py
```diff
@@ -0,0 +1,58 @@
...
+
+
+class GuardDutyResponse(BaseResponse):
+    SERVICE_NAME = "guardduty"
+
+    @property
+    def guardduty_backend(self):
+        return guardduty_backends[self.region]
+
+    # @property
```
bblommers, line 14: This commented out code shouldn't be necessary, that's all handled in the parent class (BaseResponse)

moto/guardduty/models.py
```diff
@@ -0,0 +1,82 @@
...
+        self.service_role = service_role
+        self.status = status
+        self.updated_at = updated_at
+        self.datasources = datasources
+        self.tags = tags
+
+
+guardduty_backends = {}
+for region in Session().get_available_regions("guardduty"):
+    guardduty_backends[region] = GuardDutyBackend()
```
bblommers, line 78: The GuardDutyBackend-constructor does have a region_name parameter. Don't know if that parameter is ever needed in the logic, but if so, it can be set here: `GuardDutyBackend(region_name=region)`

bblommers (changes requested):

moto/guardduty/urls.py
```diff
@@ -0,0 +1,13 @@
...
+
+response = GuardDutyResponse()
+
+url_bases = [
+    "https?://guardduty.(.+).amazonaws.com",
+]
+
+
+url_paths = {
+    "{0}/detector$": response.detector,
```
bblommers, line 12: If you want to extend this service with more methods, each method would have to be listed here.
(APIGateway is a good example on what this would look like: https://github.com/spulec/moto/blob/master/moto/apigateway/urls.py)

bblommers:
Hi @juicySec, hope you don't mind - I took the liberty to finish this PR in #4667.
It only implements the create+list feature for now, same as in this PR - but let us know if there are any other features you need!

PR #3773, opened 2021-03-15
fix route table association by internet gateway

tony-dot-sh:
fix route table association by internet gateway per https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_AssociateRouteTable.html

### DIFF
--- a/moto/ec2/models.py
+++ b/moto/ec2/models.py
@@ -4179,7 +4179,7 @@ class RouteTableBackend(object):
         self.route_tables.pop(route_table_id)
         return True
 
-    def associate_route_table(self, route_table_id, subnet_id):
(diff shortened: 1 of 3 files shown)


bblommers (changes requested):
Hi @tony-dot-sh, thanks for the PR! 
Would you mind adding some tests, to verify that we can call this method now with either parameter?

tony-dot-sh:
Added an IGW test here 6eced35de25a398b7b1bf7c61d8f97a9dcf5f1bc and subnet test here f739b95b3c397649bbdfcc71e6d4e9504b70ac2f

bblommers (approved):
LGTM

bblommers:
This is now part of moto >= 2.0.3.dev3

PR #3774, opened 2021-03-15
DNSName key retrieval for list-hosted-zones-by-dns-name

tony-dot-sh:
DNSName key retrieval for list-hosted-zones-by-dns-name per https://docs.aws.amazon.com/Route53/latest/APIReference/API_ListHostedZonesByName.html#API_ListHostedZonesByName_RequestSyntax

### DIFF
--- a/moto/route53/responses.py
+++ b/moto/route53/responses.py
@@ -53,15 +53,16 @@ class Route53(BaseResponse):
         dnsname = query_params.get("dnsname")
 
         if dnsname:
-            dnsname = dnsname[
-                0
-            ]  # parse_qs gives us a list, but this parameter doesn't repeat
-            # return all zones with that name (there can be more than one)
+            dnsname = dnsname[0]
+            if dnsname[-1] != ".":
+                dnsname += "."
             zones = [
                 zone
                 for zone in route53_backend.get_all_hosted_zones()
                 if zone.name == dnsname
             ]
+            template = Template(LIST_HOSTED_ZONES_BY_NAME_RESPONSE)
+            return 200, headers, template.render(zones=zones, dnsname=dnsname)
         else:
             # sort by names, but with domain components reversed
             # see http://boto3.readthedocs.io/en/latest/reference/services/route53.html#Route53.Client.list_hosted_zones_by_name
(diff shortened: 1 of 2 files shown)


moto/route53/responses.py
```diff
@@ -74,9 +75,11 @@ def sort_key(zone):
 
             zones = route53_backend.get_all_hosted_zones()
             zones = sorted(zones, key=sort_key)
+            template = Template(LIST_HOSTED_ZONES_BY_NAME_RESPONSE)
+            return 200, headers, template.render(zones=zones, dnsname=dnsname)
 
-        template = Template(LIST_HOSTED_ZONES_BY_NAME_RESPONSE)
-        return 200, headers, template.render(zones=zones)
+        # template = Template(LIST_HOSTED_ZONES_BY_NAME_RESPONSE)
```
bblommers, line 81: Looks like we can keep the return here, as the code is the same in the `if` and `else` sections

moto/route53/responses.py
```diff
@@ -53,15 +53,16 @@ def list_hosted_zones_by_name_response(self, request, full_url, headers):
         dnsname = query_params.get("dnsname")
 
         if dnsname:
-            dnsname = dnsname[
-                0
-            ]  # parse_qs gives us a list, but this parameter doesn't repeat
-            # return all zones with that name (there can be more than one)
+            dnsname = dnsname[0]
+            if dnsname[-1] != ".":
+                dnsname += "."
```
bblommers, line 58: Might be me misreading the test, but I can't actually see a test case where we hit this if-statement?

bblommers (changes requested):

moto/route53/responses.py
tony-dot-sh: Good eye, forgot to get rid of those. Done in 2abf91cab05ef7a903fb393018dd8bd0c6c656d4

moto/route53/responses.py
tony-dot-sh: Added add'l test in 2abf91cab05ef7a903fb393018dd8bd0c6c656d4

bblommers (approved):
LGTM. Thanks @tony-dot-sh!

bblommers:
This is now part of moto >= 2.0.3.dev2

Issue #3722, opened 2021-02-23
Release 2.0.0 - Missing dependencies? Read this


bblommers:
## New Release

Moto 2.0.0 has just been released - see our [CHANGELOG](https://github.com/spulec/moto/blob/master/CHANGELOG.md) for a full overview of all changes.

#### Note that the installation method has changed since 1.x.
To install moto for a specific service:

```
pip install moto[ec2,s3]
```

This will install Moto, and only the dependencies required for that specific service.
If you don't care about the number of dependencies, or if you want to mock many AWS services:

```
pip install moto[all]
```

Not all services might be covered, in which case you might see a warning:
_moto 1.3.16 does not provide the extra 'service'._
You can ignore the warning, or simply install moto as is:
```
pip install moto
```

jacobsorme:
I seem to be getting errors, `moto/awslambda/models.py", line 10, in <module>
    import docker
ModuleNotFoundError: No module named 'docker'`.  Before I had no problems with my dependencies. Is this related to 2.0.0? Thanks.

amarjandu:
@bblommers is version 1.3.17 going to be released? Or did 2.0.0 take over? I tried searching around, nor can find any tags mentioning it.

bblommers:
@amarjandu No, there will be no 1.3.17 release. We've moved to 2.x because of this breaking change

PR #3775, opened 2021-03-16
Add yaml to install_requires

cowboy-bebug:
Hi there 👋

I was using `mock_ssm` and noticed it requires the `yaml` package (see below).

Screen Shot 2021-03-16 at 3 33 19 PM

Just adding `yaml` as a required dependency 😊

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -35,6 +35,7 @@ install_requires = [
     "cryptography>=3.3.1",
     "requests>=2.5",
     "xmltodict",
+    "yaml",
     "six>1.9",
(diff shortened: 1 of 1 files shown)


bblommers:
Hi @cowboy-bebug, the installation method has changed as of moto 2.x. It's now required to specify which service you need to install: `pip install moto[ssm]`
That should install all required dependencies.

See this issue for more details: https://github.com/spulec/moto/issues/3722

cowboy-bebug:
@bblommers thank you so much!

PR #3776, opened 2021-03-16
feature/support create_case, resolve_case, describe_cases and associated tests

ccggeo:
Hi,

Three methods for create_case, resolve_case, describe_cases and associated tests. Please let me know if I can improve on anything and I will work to resolve.

Thanks!


moto/support/models.py
```diff
@@ -14,6 +16,8 @@ def __init__(self, region_name=None):
         super(SupportBackend, self).__init__()
         self.region_name = region_name
         self.check_status = {}
+        self.case_id = {}
```
bblommers, line 19: The dict should probably be named `cases`

moto/support/models.py
```diff
@@ -58,6 +62,151 @@ def advance_check_status(self, check_id):
...
+        language,
+        issue_type,
+        attachment_set_id,
+    ):
+        # Random case ID
+        random_case_id = "".join(
+            random.choice("0123456789ABCDEFGHIJKLMabcdefghijklm") for i in range(16)
+        )
+        case_id = "case-12345678910-2020-%s" % random_case_id
+        return {"caseId": case_id}
```
bblommers, line 140: This doesn't seem to actually persist the case details, it only returns a random ID

tests/test_support/test_support.py
```diff
@@ -155,3 +161,342 @@ def test_refresh_trusted_advisor_check_cycle_continues_on_full_cycle():
...
+        includeResolvedCases=True,
+        nextToken="test_next_token",
+        maxResults=137,
+        language="test_lanauage",
+        includeCommunications=True,
+    )
+
+    actual_case_id = describe_cases_response["cases"][0]["caseId"]
+
+    actual_case_id.should.equal(case_id_list)
```
bblommers, line 259: Would be good to have some more assertions to verify that the subject/serviceCode/etc is returned as expected

moto/support/responses.py
```diff
@@ -24,3 +24,57 @@ def refresh_trusted_advisor_check(self):
...
+        )
+
+        return json.dumps(create_case_response)
+
+    def describe_cases(self):
+        case_id_list = self._get_param("caseIdList")
+        display_id = self._get_param("displayId")
+        after_time = self._get_param("afterTime")
+        before_time = self._get_param("beforeTime")
+        include_resolved_cases = self._get_param("includeResolvedCases")
```
bblommers, line 62: Reading the docs, this param defaults to False. 
Looks like `include_communications` can default to True

tests/test_support/test_support.py
```diff
@@ -155,3 +161,342 @@ def test_refresh_trusted_advisor_check_cycle_continues_on_full_cycle():
...
+        "pending-customer-action",
+        "reopened",
+        "unassigned",
+        "resolved",
+        "work-in-progress",
+    ]
+    expected_initial_case = [resolve_case_response["initialCaseStatus"]]
+    expected_final_case = "resolved"
+
+    set(expected_initial_case).issubset(
```
bblommers, line 198: Assertions should be on separate lines, that makes it easier to debug if it ever fails.
To simplify this:
```
possible_case_status.should.contain(expected_initial_case)
# Expected_final_case has just been set, so it's obvious that this defaults to resolved?
expected_final_case.should.equal("resolved")
```

tests/test_support/test_support.py
```diff
@@ -155,3 +161,342 @@ def test_refresh_trusted_advisor_check_cycle_continues_on_full_cycle():
...
+    )
+
+    case_id_list = create_case_response["caseId"]
+    describe_cases_response = client.describe_cases(
+        caseIdList=[case_id_list],
+        displayId="test_display_id",
+        afterTime="test_after_time",
+        beforeTime="test_before_time",
+        includeResolvedCases=True,
+        nextToken="test_next_token",
```
bblommers, line 251: Can we also have a test without nextToken/maxResults set? Just to verify that it's not required

tests/test_support/test_support.py
ccggeo: Hey thanks for this. I tried to implement `possible_case_status.should.contain(expected_initial_case)` without subset but struggled to get it to check a `possible_case_status` against `expected_initial_case`. Is there a better way in your experience to achieve this?

bblommers: That doesn't work as a list, you're right. I had to change it to `list.contains(status-as-a-string)` to get it to work.
Made a few more adjustments and used the build-in `pytest.mark.parametrize`, so we don't need an additional dependency. I didn't realize it was there in my initial review, and I figured I might as well change as I'm looking at it anyway. Hope you don't mind :)

bblommers (approved):
LGTM - thanks again @ccggeo!

ccggeo:
@bblommers No worries, thank you for your parameter simplification

bblommers:
This is now part of moto >= 2.0.3.dev6

PR #3777, opened 2021-03-16
S3 - Make put-object-acl return 404 if the object does not exist

tibordp:
Currently trying to set ACL on a key does not exist fails with 

```
AttributeError: 'NoneType' object has no attribute 'set_acl'
```

S3 returns 404 (`NoSuchKey`) if the object referenced by the key does not exist


bblommers (approved):
LGTM - thanks @tibordp!

bblommers:
This is now part of moto >= 2.0.3.dev5

PR #3782, opened 2021-03-18
Update Message spec to respect SQS message class

errietta:
In SQS Message, there's an attribute called [message_id](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/sqs.html#SQS.Message.message_id) which holds the id of the message. This is correctly reflected within moto as far as I can tell from [test cases](https://github.com/spulec/moto/blob/master/tests/test_sqs/test_sqs.py#L2428) but in the model for some reason it's just called `id`. Unless I'm wrong this should be `message_id`

### DIFF
--- a/moto/sqs/models.py
+++ b/moto/sqs/models.py
@@ -71,7 +71,7 @@ DEDUPLICATION_TIME_IN_SECONDS = 300
 
 class Message(BaseModel):
     def __init__(self, message_id, body):
-        self.id = message_id
+        self.message_id = message_id
         self._body = body
         self.message_attributes = {}
         self.receipt_handle = None


errietta:
Will fix tests if someone tells me that I'm not wrong about this being an issue

bblommers:
Hey @errietta, it's not really an issue, as far as I know. The `id` attribute is for internal use only, and transforms into `message_id` before sending it back to the user:
https://github.com/spulec/moto/blob/master/moto/sqs/responses.py#L559

bblommers:
So `message_id` might be better cosmetically, but 
it shouldn't be the cause of any bugs.

errietta:
@bblommers I guess we should not be using `Message()` directly to construct mock messages then?

bblommers:
@errietta I wouldn't, no, as it is open to change without any warning. In my projects I try to only use the boto3 api. More code maybe, but less likely to break in between versions.

errietta:
Ok thanks

PR #3789, opened 2021-03-19
adding physical_resource_id in SubnetRouteTableAssociation, Route and NatGW classes

ismferd:
Adding physical_resource_id in SubnetRouteTableAssociation, Route and NatGW classes in `moto/ec2/models.py` in order to avoid https://github.com/localstack/aws-cdk-local/issues/44

### DIFF
--- a/moto/ec2/models.py
+++ b/moto/ec2/models.py
@@ -4029,6 +4029,10 @@ class SubnetRouteTableAssociation(CloudFormationModel):
     def __init__(self, route_table_id, subnet_id):
         self.route_table_id = route_table_id
         self.subnet_id = subnet_id
+    
+    @property
+    def physical_resource_id(self):
+        return self.route_table_id
 
     @staticmethod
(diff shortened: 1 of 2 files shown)


bblommers (changes requested):
Hi @ismferd, thanks for the PR!
Would you mind adding some tests, to verify this works properly?
Looks like the linter fails as well at the moment - you can run `make lint` to verify whether the formatting is correct

ismferd:
Hi @bblommers ,

I can't see the linter issue at local as you could see below:
image

bblommers:
Which version of black are you using?
The latest has some bugs, so we're using `black==19.10b0`

ismferd:
🥳

bblommers (approved):
LGTM! Thanks again @ismferd

bblommers:
This is now part of moto >= 2.0.3.dev8

PR #3796, opened 2021-03-24
Check S3_IGNORE_SUBDOMAIN_BUCKETNAME environment variable

### LINKED ISSUE #3795: environment variable S3_IGNORE_SUBDOMAIN_BUCKETNAME is ignored for stand alone s3 server
I am running a standalone s3 server on a fixed domain. I want the bucket name to be in the url, not in the domain as those domains will not be resolved. However creating a bucket will fail with `The specified bucket does not exist`. Setting the S3_IGNORE_SUBDOMAIN_BUCKETNAME environment variable to '1' does not solve the issue.

To reproduce:
* Prepare /etc/hosts so I can address a server running on localhost by a domainname 
/etc/hosts
  ````
  127.0.0.1       localhost a.test.com
  ````
* Run the s3 server
  ````bash
  export S3_IGNORE_SUBDOMAIN_BUCKETNAME=1
  moto_server s3 -H 0.0.0.0
  ````
* Test the server using boto3
  ````python
  >>> import boto3
  >>> client = boto3.client('s3', endpoint_url="http://a.test.com:5000", aws_access_key_id="myid", aws_secret_access_key='secret')
  >>> client.create_bucket(Bucket="test")
  ````
  This will generate an error:
  ````
  File "/home/onno/tmp/moto/venv/lib/python3.8/site-packages/botocore/client.py", line 357, in _api_call
  File "/home/onno/tmp/moto/venv/lib/python3.8/site-packages/botocore/client.py", line 676, in _make_api_call
  botocore.errorfactory.NoSuchBucket: An error occurred (NoSuchBucket) when calling the CreateBucket operation: The specified bucket does not exist
  ````
  
What I expected was the same beheavure as when addressing the server as localhost
````python
>>> import boto3
>>> client = boto3.client('s3', endpoint_url="http://localhost:5000", aws_access_key_id="myid", aws_secret_access_key='secret')
>>> client.create_bucket(Bucket="test")
{'ResponseMetadata': {'HTTPStatusCode': 200, 'HTTPHeaders': {'content-type': 'text/html; charset=utf-8', 'content-length': '158', 'access-control-allow-origin': '*', 'server': 'Werkzeug/1.0.1 Python/3.8.5', 'date': 'Wed, 24 Mar 2021 12:42:05 GMT'}, 'RetryAttempts': 0}}
````

I believe this issue is caused by the environment variable S3_IGNORE_SUBDOMAIN_BUCKETNAME being checked in utils.py `bucket_name_from_url` but not in responses.py `subdomain_based_buckets`.

### PR DESCRIPTION
**Pull request by Codeglitches**
This PR fixes #3795 by checking the `S3_IGNORE_SUBDOMAIN_BUCKETNAME` environment variable in the same way `bucket_name_from_url` does in utils.py

With this PR, performing the test as described in #3795 creates the bucket as expected.

Also a unittest was added to test the change

### DIFF
--- a/tests/test_s3/test_s3_utils.py
+++ b/tests/test_s3/test_s3_utils.py
@@ -1,5 +1,4 @@
 from __future__ import unicode_literals
-import os
 import pytest
 from sure import expect
 from moto.s3.utils import (
@@ -9,6 +8,7 @@ from moto.s3.utils import (
     clean_key_name,
     undo_clean_key_name,
 )
+from tests.compat import patch
 
 
 def test_base_url():
@@ -25,12 +25,11 @@ def test_localhost_without_bucket():
     expect(bucket_name_from_url("https://www.localhost:5000/def")).should.equal(None)
 
 
-def test_force_ignore_subdomain_for_bucketnames():
-    os.environ["S3_IGNORE_SUBDOMAIN_BUCKETNAME"] = "1"
-    expect(
-        bucket_name_from_url("https://subdomain.localhost:5000/abc/resource")
-    ).should.equal(None)
-    del os.environ["S3_IGNORE_SUBDOMAIN_BUCKETNAME"]
+def test_force_ignore_subdomain_for_bucketnames(monkeypatch):
+    with patch("moto.s3.utils.S3_IGNORE_SUBDOMAIN_BUCKETNAME", True):
+        expect(
+            bucket_name_from_url("https://subdomain.localhost:5000/abc/resource")
+        ).should.equal(None)
 
 
 def test_versioned_key_store():
--- a/moto/s3/responses.py
+++ b/moto/s3/responses.py
@@ -27,6 +27,7 @@ from moto.packages.httpretty.core import HTTPrettyRequest
 from moto.core.responses import _TemplateEnvironmentMixin, ActionAuthenticatorMixin
 from moto.core.utils import path_url
 from moto.core import ACCOUNT_ID
+from moto.settings import S3_IGNORE_SUBDOMAIN_BUCKETNAME
 
 from moto.s3bucket_path.utils import (
     bucket_name_from_url as bucketpath_bucket_name_from_url,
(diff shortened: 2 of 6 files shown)

### CODE REVIEW

**Review by bblommers — CHANGES_REQUESTED** [review]
Hi @Codeglitches, thanks for the PR! Would you mind adding the envvar to moto/settings.py? Just to ensure that all configuration options are in the same place. Happy to merge afterwards

File: tests/test_s3/test_s3_utils.py [hunk 1]
```diff
@@ -9,6 +8,12 @@
     clean_key_name,
     undo_clean_key_name,
 )
+import sys
+
+if sys.version_info.major == 2 and sys.version_info.minor == 7:
```
bblommers, line 13: You can use `if six.PY2` for this:
https://github.com/spulec/moto/blob/c902de8a098d2537792ec5533feb01085febd5ea/moto/s3/utils.py#L97

bblommers (changes requested):

tests/test_s3/test_s3_utils.py
Codeglitches: Thanks. I will use that

bpandola: Might want to move this conditional import to `moto.compat` and then import from there.

Codeglitches: @bpandola The import is for unittests, personaly I would not move that to `moto.compat`. I could create a `tests.compat` if required?

bblommers: Good point @bpandola. Yeah, if you could, @Codeglitches, that would be useful

Codeglitches: @bblommers do you want it in `moto.compat` or `tests.compat`?

bblommers: Sorry, that wasn't clear - I meant create a `tests.compat`

Codeglitches: I created `tests.compat.py` and used the same import style as in moto.compat (with a try...except)

bblommers (approved):
LGTM - thanks again @Codeglitches!

Issue #3799, opened 2021-03-25
Dynamo query case sensitivity doesn't behave as expected


davidpedrosa:
# Dynamo query case sensitivity doesn't behave as expected

**HINT:** I've been working on a fix. A PR will be coming soon.

In *KeyConditionExpression*  section in https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/dynamodb.html#DynamoDB.Client.query I don't see clearly stated that the comparison using BETWEEN keyword is not case sensitive, but it happens not to be. As for begins_with, it clearly establishes it is case-sensitive and lower-case must be used.

## Testing it with Dynamo

Consider the following example:

```
import boto3

dynamodb = boto3.client('dynamodb')
TABLE_NAME = 'test_ignorecase'
dynamodb.create_table(
    TableName=TABLE_NAME,
    KeySchema=[
        {"AttributeName": "forum_name", "KeyType": "HASH"},
        {"AttributeName": "subject", "KeyType": "RANGE"},
    ],
    AttributeDefinitions=[
        {"AttributeName": "forum_name", "AttributeType": "S"},
        {"AttributeName": "subject", "AttributeType": "S"},
    ],
    ProvisionedThroughput={"ReadCapacityUnits": 5, "WriteCapacityUnits": 5},
)
print("Table {table} was created".format(table=TABLE_NAME))

dynamodb.put_item(TableName=TABLE_NAME, Item={"forum_name": {"S": "the-key"}, "subject": {"S": "100"}})
dynamodb.put_item(TableName=TABLE_NAME, Item={"forum_name": {"S": "the-key"}, "subject": {"S": "199"}})
dynamodb.put_item(TableName=TABLE_NAME, Item={"forum_name": {"S": "the-key"}, "subject": {"S": "250"}})
print("Some sample data was inserted")

results = dynamodb.query(
        TableName=TABLE_NAME,
        KeyConditionExpression="forum_name = :forum_name and subject between :start  and  :end",
        ExpressionAttributeValues={
            ":forum_name": {"S": "the-key"},
            ":start": {"S": "100"},
            ":end": {"S": "200"},
        }
    )
print("between", results["Count"])
results = dynamodb.query(
        TableName=TABLE_NAME,
        KeyConditionExpression="forum_name = :forum_name and subject BETWEEN :start  AnD  :end",
        ExpressionAttributeValues={
            ":forum_name": {"S": "the-key"},
            ":start": {"S": "100"},
[...]

PR #3800, opened 2021-03-25
Improve dynamodb query case sensitivity

davidpedrosa:
See #3799

When using the compact API this would not be an issue, since `Key("whatever").eq("the-key")
        & Key("range-attribute").begins_with("some_text")` would use "begins_with" behind the scenes. 

The same goes for *between*, which would set it to uppercase.

Almost all tests use the compact version, and both conditions are thoroughly being tested.

However, KeyConditionExpression "text" syntax is also supported allowing something like `"KeyConditionExpression="key_attribute = :key_value and BEGINS_WITH(range_attribute, :some_text )"`

This should raise a controlled exception:

```botocore.exceptions.ClientError: An error occurred (ValidationException) when calling the Query operation: Invalid KeyConditionExpression: Invalid function name; function: BEGINS_WITH```

However, moto would raise an unexpected error.

A lowercase clause like this `KeyConditionExpression="key_attribute = :key_value  and range_attribute between :start  and  :end" ` should succeed but moto would raise an unexpected exception.

This PR would fix this two behaviors.

### DIFF
--- a/moto/dynamodb2/responses.py
+++ b/moto/dynamodb2/responses.py
@@ -510,7 +510,7 @@ class DynamoHandler(BaseResponse):
                 range_key_expression_components = range_key_expression.split()
                 range_comparison = range_key_expression_components[1]
 
-                if "AND" in range_key_expression:
+                if " and " in range_key_expression.lower():
                     range_comparison = "BETWEEN"
                     range_values = [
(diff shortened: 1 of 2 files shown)


bblommers (approved):
LGTM - thanks @davidpedrosa!

davidpedrosa:
Great. Thanks!

PR #3801, opened 2021-03-26
Raise error when adding duplicate egress rule to ec2:SecurityGroup

bpandola:
The `InvalidPermission.Duplicate` error was already implemented for inbound rules,
but AWS also returns this error for duplicate outbound rules.

Very minor changes were needed on existing tests that were adding duplicate
outbound rules (when testing the RulesPerSecurityGroupLimitExceeded error).


bblommers (approved):
Nice one @bpandola - thanks!

PR #3805, opened 2021-03-28
Document installing Moto with Homebrew

singingwolfboy:
See: https://github.com/Homebrew/homebrew-core/pull/73903


bblommers (approved):
👍

PR #3806, opened 2021-03-28
Modify SSM `put_parameter()` to raise ValidationException if value is empty string

### LINKED ISSUE #3794: Moto SSM put parameter allows empty string as a value, real service does not
Hello,

I noticed that moto SSM service allows putting parameter with empty string. This is not available in boto or via AWS web UI.
Moto is running in server mode. 

1. boto3==1.17.35
2. botocore==1.20.35
3. moto==2.0.2

Code example that is invoking parameter creation using moto:
`client.ssm.put_parameter(
            Name='test',
            Description='string',
            Value='',
            Type='String',
            Overwrite=True
        )`

It creates the parameter and returns empty string when fetched.

This is, on the other hand, output from boto parameter creation:
`$ aws ssm put-parameter --name "/test" --value "" --profile saml
An error occurred (ValidationException) when calling the PutParameter operation: 1 validation error detected: Value '' at 'value' failed to satisfy constraint: Member must have length greater than or equal to 1.`

That makes inconsistency. Is this a feature somehow or should be fixed?

Regards

TSNoble:
Fixes #3794 by checking `put_parameter()` value for an empty string and raising a ValidationException if this is the case.

### DIFF
--- a/moto/ssm/models.py
+++ b/moto/ssm/models.py
@@ -1284,6 +1284,11 @@ class SimpleSystemManagerBackend(BaseBackend):
     def put_parameter(
         self, name, description, value, type, allowed_pattern, keyid, overwrite, tags,
     ):
+        if value == "":
+            raise ValidationException(
+                "1 validation error detected: Value '' at 'value' failed to satisfy"
+                " constraint: Member must have length greater than or equal to 1."
+            )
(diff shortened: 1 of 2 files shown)


bblommers (approved):
LGTM - thanks @TSNoble!

PR #3808, opened 2021-03-28
Use `MOTO_ACCOUNT_ID` when building default `Rule` ARN

### LINKED ISSUE #3807: mock_events: `Rule` does not use `MOTO_ACCOUNT_ID`
I've noticed while taking a look through the events code that the `Rule` class does uses the account ID `111111111111` during construction when no ARN is specified:

```py
class Rule(CloudFormationModel):
    Arn = namedtuple("Arn", ["service", "resource_type", "resource_id"])

    def _generate_arn(self, name):
        return "arn:aws:events:{region_name}:111111111111:rule/{name}".format(
            region_name=self.region_name, name=name
        )

    def __init__(self, name, region_name, **kwargs):
        self.name = name
        self.arn = kwargs.get("Arn") or self._generate_arn(name)
        self.event_pattern = kwargs.get("EventPattern")
        self.schedule_exp = kwargs.get("ScheduleExpression")
        self.state = kwargs.get("State") or "ENABLED"
        self.description = kwargs.get("Description")
        self.role_arn = kwargs.get("RoleArn")
        self.event_bus_name = kwargs.get("EventBusName") or "default"
        self.targets = []
```

I think it'd make sense to use the environment variable `MOCK_ACCOUNT_ID` as other classes in `moto/events/models.py` do. PR incoming!

TSNoble:
Fixes #3807. Updates default `Rule` ARN building logic to use the `MOTO_ACCOUNT_ID` environment variable. Updates relevant `describe_rule()` unit test.

### DIFF
--- a/moto/events/models.py
+++ b/moto/events/models.py
@@ -28,8 +28,8 @@ class Rule(CloudFormationModel):
     Arn = namedtuple("Arn", ["service", "resource_type", "resource_id"])
 
     def _generate_arn(self, name):
-        return "arn:aws:events:{region_name}:111111111111:rule/{name}".format(
-            region_name=self.region_name, name=name
+        return "arn:aws:events:{region_name}:{account}:rule/{name}".format(
+            region_name=self.region_name, account=ACCOUNT_ID, name=name
         )
 
     def __init__(self, name, region_name, **kwargs):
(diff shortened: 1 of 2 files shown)


bblommers (approved):
Nice @TSNoble - thanks!

PR #3809, opened 2021-03-28
Fix events rule ARN for custom event bus

gruebel:
The rule ARN for rules added to a custom event bus have the event bus name prefixed.

Additionally  added `ManagedBy` and `CreatedBy` for event rules.

Also quite interesting scheduled events can only be added on the default event bus.


bblommers (approved):
LGTM - thanks @gruebel!

PR #3812, opened 2021-03-29
black 20.8b1

singingwolfboy:


bblommers:
Hi @singingwolfboy, the black release we're on is the last stable release, as far as I'm concerned. If you look at the repo (https://github.com/psf/black/issues), there are quite a few issues regarding inconsistent formatting.
The moment black is stable I'd love to upgrade, but we're not there yet.

PR #3815, opened 2021-03-30
Add events target integration for sqs queue

gruebel:
I also cleaned up the targets methods a bit.

During implementation I realised that the rules are stored for all Event Buses centrally, which is kind of wrong, because they are bound to the specific Event Bus. In the past there was only one (default), so it was ok. @bblommers should I refactor it and move the list of rules to the specific Event Bus?  If yes, then I would create a dedicated issue and implement it via a separate PR.

### DIFF
--- a/moto/events/models.py
+++ b/moto/events/models.py
@@ -100,7 +100,10 @@ class Rule(CloudFormationModel):
         if not self._validate_event(event):
             return
 
-        # for now only CW Log groups are supported
+        # supported targets
+        # - CloudWatch Log Group
+        # - EventBridge Archive
+        # - SQS Queue (not FIFO)
         for target in self.targets:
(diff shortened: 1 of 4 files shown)


bblommers:
> 
> 
> should I refactor it and move the list of rules to the specific Event Bus? 

From what I can tell, that wouldn't necessarily improve the code quality. In all CRUD-operations we would have to find the EventBus first, before adding/deleting/listing the rules.
The only positive change would be in `Rule:send_to_targets()`, where we can get rid of the if-statement that checks whether this Rule belongs to the correct EventBus.

From a perfectionalist point of view, yes, it would be good to store them under an EventBus. But from a practical view, I'd argue we wait until there's a functional need for it.

bblommers (approved):
PR looks good!

PR #3817, opened 2021-03-31
Make security rules consistent between direct (backend) and indirect (api) boundaries

bpandola:
[Discovered while working on adding support for [Managed Security Groups in EMR](https://docs.aws.amazon.com/emr/latest/ManagementGuide/emr-man-sec-groups.html)]

Security rules added directly via the backend were unable to be revoked via the API
because the port values were being stored as strings but were always coerced back
to integers by the botocore model.  `"0" != 0`, so the rules would never match,
raising an `InvalidPermissionNotFoundError`.

This change ensures that the port values for a security group rule are always of type
`Union[int, None]`.

No tests needed to be modified as a result of this change.  A new test was added that
explicitly covers the behavior that had been failing.

### DIFF
--- a/moto/ec2/models.py
+++ b/moto/ec2/models.py
@@ -1984,10 +1984,11 @@ class SecurityRule(object):
         self.ip_protocol = ip_protocol
         self.ip_ranges = ip_ranges or []
         self.source_groups = source_groups
+        self.from_port = self.to_port = None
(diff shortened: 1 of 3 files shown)


bblommers (approved):
LGTM - thanks @bpandola!

PR #3822, opened 2021-04-01
Enhancement/3821

### LINKED ISSUE #3821: mock_events: `Archive` contains events that do not match `EventPattern`
**How to reproduce:**

```py
@mock_events
def test_event_not_routed_to_archive_when_detail_does_not_match_pattern():
    # given
    client = boto3.client("events", "eu-central-1")
    event_bus_arn = "arn:aws:events:eu-central-1:{}:event-bus/default".format(
        ACCOUNT_ID
    )
    client.create_archive(
        ArchiveName="no-matches-detail",
        EventSourceArn=event_bus_arn,
        EventPattern=json.dumps({
            "detail": {"foo": ["baz"]}
        })
    )

    # when
    event = {
        "Source": "source",
        "DetailType": "type",
        "Detail": '{"foo": "baz"}'
    }
    client.put_events(Entries=[event])

    # then
    response = client.describe_archive(ArchiveName="no-matches-detail")
    response["EventCount"].should.equal(0)
    response["SizeBytes"].should.equal(0)
```

**Expected behaviour:**

Since the Detail of the event does not match the archive's EventPattern, I would expect there to be no events in the archive and therefore for the test to pass.

**Actual behaviour:**

```
self = 'X is 1 whereas Y is 0', X = 1, Y = 0

    def as_assertion(self, X, Y):
>       raise self.get_assertion(X, Y)
E       AssertionError: given
E       X = 1
E           and
E       Y = 0
E       X is 1 whereas Y is 0

venv/lib/python3.8/site-packages/sure/core.py:57: AssertionError

    @mock_events
    def test_event_not_routed_to_archive_when_detail_does_not_match_pattern():
        # given
        client = boto3.client("events", "eu-central-1")
        event_bus_arn = "arn:aws:events:eu-central-1:{}:event-bus/default".format(
            ACCOUNT_ID
        )
        client.create_archive(
            ArchiveName="no-matches-detail",
            EventSourceArn=event_bus_arn,
            EventPattern=json.dumps({
                "detail": {"foo": ["baz"]}
            })
        )
    
        # when
        event = {
            "Source": "source",
            "DetailType": "type",
            "Detail": '{"foo": "baz"}'
        }
        client.put_events(Entries=[event])
    
        # then
        response = client.describe_archive(ArchiveName="no-matches-detail")
>       response["EventCount"].should.equal(0)

tests/test_events/test_events.py:1346: 
```

**Note:**
I'm planning to raise a PR fixing the issue soon

TSNoble:
Fixes #3821 by adding a `_does_event_match_pattern()` function to `EventsBackend`, which is now used to determine whether an event should be sent to an `Archive`.

Note that the pattern matching code currently only supports list matching i.e. `{"key": ["val1", "val2", ...]}` and recursive dict matching i.e. `{"key1": {"key2": {"key3": ["val1"]}}}` `EventPattern` syntax.

Adds a new unit test verifying that unmatching events are not archived.

### DIFF
--- a/moto/events/models.py
+++ b/moto/events/models.py
@@ -91,6 +91,7 @@ class Rule(CloudFormationModel):
         return True
 
     def _does_event_item_match_pattern_item(self, event_item, pattern_item):
+        #  Only supports "key: [value]" filters currently
         if not event_item:
             return False
         if isinstance(pattern_item, list):
--- a/moto/events/models.py
+++ b/moto/events/models.py
@@ -81,6 +81,23 @@ class Rule(CloudFormationModel):
             if index is not None:
                 self.targets.pop(index)
 
+    def _does_event_match_pattern(self, event, pattern):
+        if not pattern:
+            return True
+        event_pattern_pairs = [(event.get(k), v) for k, v in pattern.items()]
+        for event_item, pattern_item in event_pattern_pairs:
+            if not self._does_event_item_match_pattern_item(event_item, pattern_item):
+                return False
+        return True
+
+    def _does_event_item_match_pattern_item(self, event_item, pattern_item):
+        if not event_item:
+            return False
+        if isinstance(pattern_item, list):
+            return event_item in pattern_item
+        if isinstance(pattern_item, dict):
+            return self._does_event_match_pattern(event_item, pattern_item)
+
     def send_to_targets(self, event_bus_name, event):
         if event_bus_name != self.event_bus_name:
             return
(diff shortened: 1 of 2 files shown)


bblommers (changes requested):
Hi @TSNoble, thanks for the PR!
Can you extend the test case a bit? Would be good to also have tests for positive matches, dicts, lists, etc

TSNoble:
Hey @bblommers. Thanks for the quick feedback! I've extended the test case to include a positive case with a nested dict filter (already includes the negative case), and positive and negative cases with a list filter.

bblommers (approved):
LGTM!

PR #3824, opened 2021-04-01
Add support for EMR-Managed Security Groups

bpandola:
This covers the base case for EMR Clusters provisioned in a private subnet.

Ref: https://docs.aws.amazon.com/emr/latest/ManagementGuide/emr-man-sec-groups.html


moto/emr/models.py
```diff
@@ -501,7 +514,50 @@ def remove_tags(self, cluster_id, tag_keys):
...
+        manager = EmrSecurityGroupManager(self.ec2_backend, subnet.vpc_id)
+        master, slave, service = manager.manage_security_groups(
+            emr_managed_master_security_group,
+            emr_managed_slave_security_group,
+            service_access_security_group,
+        )
+        return master.id, slave.id, service.id
+
     def run_job_flow(self, **kwargs):
+        if "instance_attrs" in kwargs:
```
bblommers, line 555: Do we need the if-statement? The `Instances`-param is required for the run_job_flow-method, so it looks like this kwarg will always be there

moto/emr/models.py
```diff
@@ -2,16 +2,21 @@
 from datetime import datetime
 from datetime import timedelta
 
+import warnings
+
 import pytz
 from boto3 import Session
 from dateutil.parser import parse as dtparse
 from moto.core import BaseBackend, BaseModel
+from moto.ec2 import ec2_backends
```
bblommers, line 11: Can we move the import inside the `run_job_flow`-method?
The EC2-service has some dependencies that are not required for EMR, so users should only have to install them when calling the `run_job_flow`

tests/test_emr/test_emr_boto3.py
```diff
@@ -1045,3 +1045,78 @@ def test_security_configurations():
     ex.value.response["Error"]["Message"].should.match(
         r"Security configuration with name .* does not exist."
     )
+
+
+@mock_emr
+@mock_ec2
```
bblommers, line 1051: There is a script that tests whether all the required dependencies are included in setup.py, `scripts/int_test.sh`
The short version is:
For each service: create a venv, install dependencies for that service only, and run all the tests

This test will cause the script to fail, as calling `mock_ec2` will require some imports that haven't been installed.

TLDR: Would you mind moving this test to a new file, `test_emr_integration.py`?
In general, all tests that cover multiple services should be declared in a file with that specific name, so the script knows to exclude those when testing the dependencies

tests/test_emr/test_utils.py
```diff
@@ -47,3 +50,125 @@ def test_release_label_comparisons():
         "emr-5.60.2",
     ]
     assert releases_sorted == expected
+
+
+@pytest.mark.skipif(
+    settings.TEST_SERVER_MODE, reason="Can't modify backend directly in server mode."
+)
+class TestEmrSecurityGroupManager(object):
```
bblommers, line 58: Same here with this test, I guess - that can be moved to `test_emr_util_integration.py`

moto/emr/utils.py
```diff
@@ -1,8 +1,14 @@
...
+import datetime
 import random
 import re
 import string
-from moto.core.utils import camelcase_to_underscores
+from moto.core.utils import (
+    camelcase_to_underscores,
+    iso_8601_datetime_with_milliseconds,
+)
+from moto.ec2.exceptions import InvalidPermissionDuplicateError
```
bblommers, line 11: Same here - this import should be moved as low as possible

bblommers (changes requested):

bblommers (approved):
LGTM - thanks @bpandola!

PR #3829, opened 2021-04-02
IAM - get_user() #3828

### LINKED ISSUE #3828: The iam.get_user method does not return any tags when mocked.
The `iam.get_user` method does not return any tags when mocked. But when tested against a live environment without moto it **does** return tags.

```python
#!/usr/bin/env python3
import json

import boto3
from moto import mock_iam

mock_iam = mock_iam()
mock_iam.start()
iam = boto3.client('iam')
iam.create_user(
    UserName='TestUser1',
    Tags=[
        {
            'Key': 'TestTag1',
            'Value': '90'
        },
        {
            'Key': 'TestTag2',
            'Value': 'abc'
        }
    ]
)
resp = iam.get_user(
        UserName='TestUser1'
)
print(json.dumps(resp['User'], indent=2, default=str))
```
Returns
```json
{
  "Path": "/",
  "UserName": "TestUser1",
  "UserId": "h3lhk2pzgb39o9zvtak2",
  "Arn": "arn:aws:iam::123456789012:user/TestUser1",
  "CreateDate": "2021-04-02 04:31:41.912000+00:00"
}
```
Expected
```json
{
  "Path": "/",
  "UserName": "TestUser1",
  "UserId": "h3lhk2pzgb39o9zvtak2",
  "Arn": "arn:aws:iam::123456789012:user/TestUser1",
  "CreateDate": "2021-04-02 04:31:41.912000+00:00",
  "Tags": [
    {
        "Key": "TestTag1",
        "Value": "90"
    },
    {
        "Key": "TestTag2",
        "Value": "abc"
    }
  ]
}
```

Jonx1000:
- fixing issue where the get_user method does not return tags

Fixes #3828

Documentation:
https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/iam.html#IAM.Client.get_user

### DIFF
--- a/moto/iam/responses.py
+++ b/moto/iam/responses.py
@@ -484,9 +484,9 @@ class IamResponse(BaseResponse):
                 user = User("default_user")
         else:
             user = iam_backend.get_user(user_name)
-
+        tags = iam_backend.tagger.list_tags_for_resource(user.arn).get("Tags", [])
         template = self.response_template(USER_TEMPLATE)
-        return template.render(action="Get", user=user)
+        return template.render(action="Get", user=user, tags=tags)
 
     def list_users(self):
         path_prefix = self._get_param("PathPrefix")
--- a/tests/test_iam/test_iam.py
+++ b/tests/test_iam/test_iam.py
@@ -4075,7 +4075,8 @@ def test_create_user_with_tags():
     assert resp["User"]["Tags"] == tags
     resp = conn.list_user_tags(UserName=user_name)
     assert resp["Tags"] == tags
-
+    resp = conn.get_user(UserName=user_name)
+    assert resp["User"]["Tags"] == tags
     resp = conn.create_user(UserName="test-create-user-no-tags")
     assert "Tags" not in resp["User"]
 


moto/iam/responses.py
```diff
@@ -484,9 +484,9 @@ def get_user(self):
                 user = User("default_user")
         else:
             user = iam_backend.get_user(user_name)
-
+            tags = iam_backend.tagger.list_tags_for_resource(user.arn).get("Tags", [])
```
bblommers, line 487: Looks like the indentation is incorrect - it shouldn't be part of the `else` statement

bblommers (changes requested):
Hi @Jonx1000, thanks for the PR! 
Can you add a test for this, to verify it's returned correctly? Probably easiest to extend the existing 'tests/test_iam/test_iam.py::test_create_user_with_tags` with a `get_user` call

Jonx1000:
@bblommers thank you I removed it from the conditional.

Jonx1000:
Thanks for pointing that out. I added some tests.

bblommers (approved):
LGTM!

PR #3830, opened 2021-04-02
Add events target integration for sqs fifo queue

gruebel:
I was not sure, how to best warn the user, if you choose a FIFO queue as a target it needs to have content-based deduplication enabled. Otherwise it won't be delivered at all, that's why I added the `warnings.warn` part.


bblommers:
How does AWS handle it without dedup enabled - does it fail silently?

gruebel:
Basically you don't receive any message in the FIFO queue without dedup enabled, maybe they will dump it in a dead letter queue, if you attach it, but I didn't try it out yet.

gruebel:
I checked and it will deliver the message to the dead letter queue, if you don't enable dedup with following error message

_The queue should either have ContentBasedDeduplication enabled or MessageDeduplicationId provided explicitly (Service: AmazonSQS; Status Code: 400; Error Code: InvalidParameterValue; Request ID: xxx; Proxy: null)_

gruebel:
@bblommers do you know where the Python 2 error comes from, I don't think it comes from my changes.

tests/test_events/test_events_targets.py
```diff
@@ -68,6 +68,94 @@ def test_send_to_cw_log_group():
     message["detail"].should.equal({"key": "value"})
 
 
+@mock_events
+@mock_sqs
```
bblommers, line 72: There is a script that tests whether all the required dependencies are included in setup.py, scripts/int_test.sh
The short version is:
For each service: create a venv, install dependencies for that service only, and run the tests for that service

This test will cause the script to fail, as calling mock_sqs will require some imports that haven't been installed for events.

TLDR: Would you mind moving the SQS test to a new file, test_events_integration.py?
In general, all tests that cover multiple services should be declared in a file with that specific name (`test*integration`), so the script knows to exclude those when testing the dependencies

bblommers (changes requested):

bblommers:
> @bblommers do you know where the Python 2 error comes from, I don't think it comes from my changes.

I'm getting the same failures on master, on a fresh venv. Might be a broken dependency - will do some debugging this weekend

tests/test_events/test_events_targets.py
gruebel: I see, wouldn't it be better I just rename this file, because I actually planned to put all kinds of targets in there, one of the next ones will be Lambda and SNS?

bblommers: Sure, that makes sense

bblommers (approved):
LGTM - thanks @gruebel!

PR #3835, opened 2021-04-02
Add prefix and numeric filtering logic for Archive EventPattern

TSNoble:
Adds logic for "prefix" and "numeric" filters in an `Archive` `EventPattern` with related test cases. Also removes a rogue print and reverts to defaulting to `True` if a named filter is not implemented

### DIFF
--- a/moto/events/models.py
+++ b/moto/events/models.py
@@ -561,7 +561,6 @@ class Replay(BaseModel):
 
 
 class EventPattern:
-
     def __init__(self, filter):
         self._filter = json.loads(filter) if filter else None
 
--- a/moto/events/models.py
+++ b/moto/events/models.py
@@ -95,57 +95,6 @@ class Rule(CloudFormationModel):
             if index is not None:
                 self.targets.pop(index)
 
-    def _does_event_match_filter(self, event, filter):
-        if not filter:
-            return True
-        items_and_filters = [(event.get(k), v) for k, v in filter.items()]
-        nested_filter_matches = [
-            self._does_event_match_filter(item, nested_filter)
-            for item, nested_filter in items_and_filters
-            if isinstance(nested_filter, dict)
-        ]
-        filter_list_matches = [
-            self._does_item_match_filters(item, filter_list)
-            for item, filter_list in items_and_filters
-            if isinstance(filter_list, list)
-        ]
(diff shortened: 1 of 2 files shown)


tests/test_events/test_events.py
```diff
@@ -1598,6 +1598,91 @@ def test_archive_with_exists_event_filter():
...
+    )
+    matching_event_1 = {"Source": "", "DetailType": "", "Detail": '{"foo": "bar"}'}
+    matching_event_2 = {"Source": "", "DetailType": "", "Detail": '{"foo": "bar!"}'}
+    not_matching_event = {"Source": "", "DetailType": "", "Detail": '{"foo": "ba"}'}
+    response = client.put_events(
+        Entries=[matching_event_1, matching_event_2, not_matching_event]
+    )
+    response["FailedEntryCount"].should.equal(0)
+    response = client.describe_archive(ArchiveName="with-prefix-filter")
+    response["EventCount"].should.equal(2)
```
bblommers, line 1620: Can we extend the assertions here (for all tests) to verify that it returns the correct events? That would make the tests more robust

moto/events/models.py
```diff
@@ -127,13 +127,24 @@ def _does_item_match_named_filter(self, item, filter):
             item_exists = item is not None
             should_exist = filter_value
             return item_exists if should_exist else not item_exists
+        if filter_name == "prefix":
+            prefix = filter_value
+            return item.startswith(prefix)
+        if filter_name == "numeric":
+            as_function = {"<": lt, "<=": le, "=": eq, ">=": ge, ">": gt}
```
bblommers, line 134: Very clean way of doing this! Nice

bblommers (changes requested):

tests/test_events/test_events.py
TSNoble: Shall do. I was wondering whether it'd be worth pulling the logic out into an `EventPattern` class and testing the filtering method directly, rather than doing so through an `Archive`. I think the same logic is common to `EventPattern`s on `Rule`s, too.

bblommers: Makes sense, if we can reuse the logic anyway

bblommers (approved):
LGTM - thanks @TSNoble!

PR #2834, opened 2020-03-23
Cloudwatch  - Add default metrics for other services

bblommers:
Fixes #2711 

Adds a framework for services to register AWS OOTB metrics in CloudWatch.
Currently only implemented for S3, which registers ```BucketSizeBytes``` and ```NumberOfObjects``` per bucket.

Can be fairly easily extended for EC2 for instance, if we want to return e.g. ```CPUUtilization```

Also adds the missing feature of being able to filter metrics by dimension - before you could only filter by name or namespace.

PR #3839, opened 2021-04-06
Cloudwatch - Return build-in S3 metrics - take 2

bblommers:
Original PR: #2834

Better solution for the original feature request, for S3 to supply build-in metrics such as BucketSizeBytes and NumberOfObjects


bpandola (approved):
Looping through the subclasses is clever.  Looks good to me!  (Assuming the conflicts get resolved, of course.)

PR #3841, opened 2021-04-07
document use with unittest

### LINKED ISSUE #3840: document how to use with unittest.TestCase
I believe the README should contain an example of how to use `moto` with `unittest`. Perhaps another subsection with [usage](https://github.com/spulec/moto#usage).

As far as I can tell, the best way is:

```
import unittest
from moto import mock_s3

class MyTest(unittest.TestCase):
    mock_s3 = mock_s3()
    def setUp(self):
        self.mock_s3.start()

    def tearDown(self):
        self.mock_s3.stop()

    def test(self):
        ...
```

Can someone confirm that this is the best way? If so, I'll write a PR to add this to the README.

mdavis-xyz:
Fixes #3840

### DIFF
--- a/README.md
+++ b/README.md
@@ -299,7 +299,7 @@ have altered some of the mock behavior. In short, you need to ensure that you _a
    This can typically happen if you import a module that has a `boto3` client instantiated outside of a function.
    See the pesky imports section below on how to work around this.
 
-### Example on usage?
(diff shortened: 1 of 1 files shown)


bblommers (approved):
Nice - thanks @mdavis-xyz!

PR #3843, opened 2021-04-09
Make it possible to customize the ACM cert validation wait time.

xiaket:
We are using AWS lambda to create/delete the acm certificates and doing the validation via DNS, when writing unittest for these lambdas we'll have to wait 60 seconds. This PR make it possible to customize this wait time.

### DIFF
--- a/moto/acm/models.py
+++ b/moto/acm/models.py
@@ -5,7 +5,7 @@ import datetime
 from moto.core import BaseBackend, BaseModel
 from moto.core.exceptions import AWSError
 from moto.ec2 import ec2_backends
-from moto.settings import ACM_VALIDATION_WAIT
+from moto import settings
 
 from .utils import make_arn_for_certificate
 
@@ -332,7 +332,7 @@ class CertBundle(BaseModel):
         if (
             self.type == "AMAZON_ISSUED"
             and self.status == "PENDING_VALIDATION"
-            and waited_seconds > ACM_VALIDATION_WAIT
+            and waited_seconds > settings.ACM_VALIDATION_WAIT
         ):
             self.status = "ISSUED"
 
--- a/moto/acm/models.py
+++ b/moto/acm/models.py
@@ -1,6 +1,7 @@
 from __future__ import unicode_literals
 
 import re
+import os
 import datetime
 from moto.core import BaseBackend, BaseModel
 from moto.core.exceptions import AWSError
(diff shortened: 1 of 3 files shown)


moto/acm/models.py
```diff
@@ -324,13 +325,15 @@ def validate_chain(self):
             )
 
     def check(self):
-        # Basically, if the certificate is pending, and then checked again after 1 min
-        # It will appear as if its been validated
+        # Basically, if the certificate is pending, and then checked again after a
+        # while, it will appear as if its been validated. The default wait time is 60
+        # seconds but you can set an environment to change it.
+        wait = os.environ.get("MOTO_ACM_VALIDATION_WAIT", 60)
```
bblommers, line 331: This should be moved into `moto/settings.py`, where all our env variables are stored

bblommers (changes requested):
Hey @xiaket, thanks for the PR!
Would you mind adding a unit test for this? See this test for an example on how we test env vars in other services: https://github.com/spulec/moto/blob/d3ad9d66866413670c31f36d54e939647feedeb1/tests/test_stepfunctions/test_stepfunctions.py#L933

tests/test_acm/test_acm.py
```diff
@@ -530,6 +545,36 @@ def test_request_certificate_issued_status():
         resp["Certificate"]["Status"].should.equal("ISSUED")
 
 
+@mock_acm
+@mock.patch.dict(os.environ, {"MOTO_ACM_VALIDATION_WAIT": 3})
+def test_request_certificate_issued_status_with_wait_in_envvar():
+    # After requesting a certificate, it should then auto-validate after 1 minute
```
bblommers, line 551: Still need to use the SkipTest here, otherwise our ServerMode tests wil fail

bblommers:
Make sure that  you use `black==19.10b0` for linting, as later versions have a few bugs

xiaket:
> Make sure that you use `black==19.10b0` for linting, as later versions have a few bugs

Cheers mate, I'm having trouble with this. :)

xiaket:
Thanks @bblommers. I ended up not mocking the envvar directly but mocked the value from settings(because there's a cache somewhere that I cannot figure out). Hope it looks good now. Please let me know if you have any follow up suggestions/comments. I appreciate your help!

bblommers (approved):
LGTM @xiaket - thanks again!

Issue #3568, opened 2020-12-29
Support Sagemaker Notebook Instance on Cloud Formation


vineeth-450:
Moto's CloudFormation mock does not support AWS::SageMaker::NotebookInstance .

When I execute a create_stack operation on a CloudFormation template that creates AWS::SageMaker::NotebookInstance, the follwing warning is returned:
`UserWarning: Tried to parse AWS::SageMaker::NotebookInstance but it's not supported by moto's CloudFormation implementation`

Are there plans on implementing the above ?

bblommers:
Thanks @vineeth-450 - looks like we're not supporting any SageMaker models at the moment.
Marking it as an enhancement.

bblommers:
This is now part of moto >= 2.0.5.dev17

PR #3845, opened 2021-04-09
Add CloudFormation support for SageMaker Notebook Instances

zachurchill-root:
Issue: https://github.com/spulec/moto/issues/3568

Updates the `FakeSagemakerNotebookInstance` class to inherit from `CloudFormationModel` and adds tests around the creation, deletion, and updating of a Notebook Instance via CloudFormation templates.


bblommers (approved):
LGTM - thanks @zachurchill-root!

PR #3847, opened 2021-04-10
Enhancement/3837

### LINKED ISSUE #3837: Archive Event Pattern issues/enhancements
Seeing the enhancement from @TSNoble regarding EventBridge Archive Pattern, I think I should offer my personal experience here, because I initially implemented it. I think it was a great idea to create a dedicated `EventPattern` class, I also taught about it, but had no time to work on it.

So, how it was implemented is actually not correct at the moment. The Archive Event Pattern has no real meaning, because EventBridge puts everything into a self managed Event Rule `Events-Archive-[name of archive]` there it merges the provided Archive Event Pattern with a special pattern to not archive it's own messages again, when a replay was initiated.
```
{
  "replay-name": [{
    "exists": false
  }]
}
```

That was the reason why I created
https://github.com/spulec/moto/blob/b138d9956bf784a122b31508e27c71c71b9d0483/moto/events/models.py#L103
so there is no special magic for Archive Event Patterns, they are all normal Event Rule Patterns. So, it would be great to move everything there, then also the normal rules can benefit from the new functionality

An other issue is the recent integration of the `exists` filter, it will be only applied on the leaves of the tree not on the branches. This test should pass and not fail
```
bar_not_exists = EventPattern(json.dumps({"detail": {"bar": [{"exists": False}]}}))
assert bar_not_exists.matches_event({"detail": {"bar": {"foo": "bar"}}})
```
https://docs.amazonaws.cn/en_us/eventbridge/latest/userguide/eb-event-patterns-content-based-filtering.html#eb-filtering-exists-matching

TSNoble:
Fixes #3837 by moving event pattern validation logic inside the EventPattern class and using the EventPattern class for filtering on all Rules, rather than those specific to Archives. Also fixes "exists" filtering logic to only match leaf nodes of an event in line with the AWS docs and updates the test logic to verify that this is the case


bblommers (approved):

bblommers:
LGTM. @gruebel Would you mind having a look as well?

TSNoble:
Just realised that the unit test names were still referencing archives, despite the fact that we're testing the event pattern logic directly now. Have updated them

gruebel (approved):
Great job! There is only one thing, maybe it would be better to move the event pattern tests to a dedicated file, something like `test_events_pattern.py`? Because the tests are quite different compared to other tests in `test_events.py`. @bblommers what do you think?

bblommers:
Moving the tests to a separate file does actually make sense @gruebel , as the class is already quite large.
@TSNoble Happy to merge once that's done

TSNoble:
Makes sense. I've moved them into `test_event_pattern.py` now

bblommers:
Awesome. Thanks!

PR #3848, opened 2021-04-10
Support Python 3.9

bblommers:
In order for 3.9 to work, we needed to unpin the `LXML` dependency. This was already committed to the master-branch here: https://github.com/spulec/moto/commit/ede25798362b08500c3f30eae73658666824bb41#diff-2b4945591edfeaa4cf4d3f155e66d4b43d1bda7a55d881d5cf3107f1b05abbbc

So the only thing outstanding is to add 3.9 to the matrix.

Original description of this PR, before the LXML dependency was unpinned:

-----------------------------------------
This `lxml`-dependency is only required for the scaffold-script, not for regular development, so pinning it to any specific version is not necessary as far as I can tell.

The additional Linux-dependencies were required to build lxml from source, as the pinned version did not have a wheel released for 3.8
The latest version supports 3.8 and 3.9 OOTB.


bpandola (commented):
Should you also bump the `after_n_builds` attribute in the `codecov.yml` file to account for the additional test runs?

bblommers:
I should! I knew I was going to forget something..

PR #3855, opened 2021-04-12
Add CloudFormation support for SageMaker Notebook Instance Lifecycle Configs

zachurchill-root:
Issue: https://github.com/spulec/moto/issues/3846

- Updates the `FakeSageMakerNotebookInstanceLifecycleConfig` class to inherit from `CloudFormationModel` and adds tests around the creation, deletion, and updating of a Notebook Instance via CloudFormation templates.
- Additionally, I reduced the overhead for adding the "create", "get_attr", and "delete" CloudFormation tests by using `pytest.mark.parametrize` because these tests will be similar across the SageMaker resources. I kept the "update" tests unparameterized since needing to know how to update and how to inspect the updates varies across the SageMaker resources.

Issue #3846, opened 2021-04-10
Add CloudFormation support for remaining SageMaker resources


zachurchill-root:
As a corollary to https://github.com/spulec/moto/issues/3568, it would be nice to also provide CloudFormation support for the following SageMaker resources: Models, Endpoint Configurations, Endpoints, and Notebook Lifecycle Configurations.

*Note:* Training Jobs [not currently supported](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/AWS_SageMaker.html) by CloudFormation.

zachurchill-root:
Would it be better to break this issue into an issue for each SageMaker resource in order to link a PR per SageMaker resource? Or is it preferable to just knock all of these out in a single PR?

bblommers:
Hey @zachurchill-root, I'm OK with just a single issue and single PR for all of them.
If it turns out to be easier to split the work in multiple PR's, just reference this issue in the PR's instead of closing it.

bblommers:
All Sagemaker models should now be supported by our CloudFormation implementation as of moto >= 2.0.6.dev4

PR #3861, opened 2021-04-15
Add CloudFormation support for SageMaker Models

zachurchill-root:
Issue: https://github.com/spulec/moto/issues/3846

- Updates the `Model` class to inherit from `CloudFormationModel` and adds tests around the creation, deletion, and updating of a Model via CloudFormation templates.
- Although the refactoring of SageMaker CloudFormation tests in https://github.com/spulec/moto/pull/3855 reduced the redundancy in the number of tests, I wasn't satisfied with all of the repeated data being passed into the `pytest.mark.parametrize` decorator. Thus, I decided to create a simple interface for test configuration objects to encapsulate all of the necessary data needed in the `create`, `get_attr`, and `delete` test parameterizations.

*Side-note*: I'm still thinking through how to make the `update` tests less redundant but I may just keep these separate and out of the test configuration objects since there is more knowledge needed to know what to change and how to check the changes.

### DIFF
--- a/moto/sagemaker/models.py
+++ b/moto/sagemaker/models.py
@@ -2,7 +2,6 @@ from __future__ import unicode_literals
 
 import os
 from boto3 import Session
-from copy import deepcopy
 from datetime import datetime
 
 from moto.core import ACCOUNT_ID, BaseBackend, BaseModel, CloudFormationModel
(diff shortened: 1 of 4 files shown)


moto/sagemaker/models.py
```diff
@@ -353,6 +353,83 @@ def arn_for_model_name(model_name, region_name):
...
+        cls, resource_name, cloudformation_json, region_name
+    ):
+        sagemaker_backend = sagemaker_backends[region_name]
+
+        # Get required properties from provided CloudFormation template
+        properties = cloudformation_json["Properties"]
+        execution_role_arn = properties["ExecutionRoleArn"]
+        primary_container = properties["PrimaryContainer"]
+
+        # `create_model` returns `Model.response_create()` so we will
```
bblommers, line 388: It does at the moment.. but it probably shouldn't. Separation of concerns and all that.
Would you mind changing this so that the `sagemaker_backend.create_model` just returns a Model object, and the response.py than calls `response_create()`?
That way, the model is only concerned about storing the data - the response is responsible for formatting that data - and we don't have to manually create the model twice here

bblommers (changes requested):

bblommers (approved):
LGTM - thanks @zachurchill-root!

PR #3864, opened 2021-04-17
sort Route53 resource record sets

zubenkoivan:
According to the [docs](https://docs.aws.amazon.com/Route53/latest/APIReference/API_ListResourceRecordSets.html) resource record sets should have the following order:
```
ListResourceRecordSets sorts results first by DNS name with the labels reversed, for example:

com.example.www.

Note the trailing dot, which can change the sort order when the record name contains characters that appear before . (decimal 46) in the ASCII table. These characters include the following: ! " # $ % & ' ( ) * + , -

When multiple records have the same DNS name, ListResourceRecordSets sorts results by the record type.
```

Also API should return 400 Bad Request if you specify only `Type`.

### DIFF
--- a/moto/route53/models.py
+++ b/moto/route53/models.py
@@ -216,14 +216,9 @@ class RecordSet(CloudFormationModel):
 
 
 def reverse_domain_name(domain_name):
-    labels = domain_name.split(".")
-    if domain_name.endswith("."):
-        labels.pop()
-        labels = list(reversed(labels))
-        labels.append("")
-    else:
-        labels = reversed(labels)
-    return ".".join(labels)
+    if domain_name.endswith("."):  # normalize without trailing dot
+        domain_name = domain_name[:-1]
+    return ".".join(reversed(domain_name.split(".")))
 
 
 class FakeZone(CloudFormationModel):
@@ -271,9 +266,10 @@ class FakeZone(CloudFormationModel):
 
     def get_record_sets(self, start_type, start_name):
         def predicate(rrset):
-            return (
-                reverse_domain_name(rrset.name) < reverse_domain_name(start_name)
-                or rrset.type_ < start_type
+            rrset_name_reversed = reverse_domain_name(rrset.name)
+            start_name_reversed = reverse_domain_name(start_name)
+            return rrset_name_reversed < start_name_reversed or (
+                rrset_name_reversed == start_name_reversed and rrset.type_ < start_type
             )
 
         record_sets = sorted(
--- a/moto/route53/models.py
+++ b/moto/route53/models.py
@@ -1,5 +1,6 @@
 from __future__ import unicode_literals
 
+import itertools
 from collections import defaultdict
 
 import string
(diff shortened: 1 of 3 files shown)


moto/route53/models.py
```diff
@@ -215,9 +216,14 @@ def delete(self, *args, **kwargs):
...
-        domain_name = domain_name[:-1]
-    return ".".join(reversed(domain_name.split(".")))
+    labels = domain_name.split(".")
+    if domain_name.endswith("."):
+        labels.pop()
+        labels = list(reversed(labels))
+        labels.append("")
+    else:
+        labels = reversed(labels)
+    return ".".join(labels)
```
bblommers, line 226: Why the rewrite? The logic seems the same as it was before

moto/route53/models.py
```diff
@@ -264,20 +270,20 @@ def delete_rrset_by_id(self, set_identifier):
         ]
 
     def get_record_sets(self, start_type, start_name):
-        record_sets = list(self.rrsets)  # Copy the list
+        def predicate(rrset):
+            return (
+                reverse_domain_name(rrset.name) < reverse_domain_name(start_name)
+                or rrset.type_ < start_type
```
bblommers, line 276: This should be an _and_-condition

tests/test_route53/test_route53.py
```diff
@@ -113,24 +117,35 @@ def test_rrset():
...
+    rrsets[1].resource_records[0].should.equal("5.6.7.8")
+    rrsets[2].resource_records[0].should.equal("bar")
 
     rrsets = conn.get_all_rrsets(zoneid, name="foo.bar.testdns.aws.com", type="A")
-    rrsets.should.have.length_of(2)
+    rrsets.should.have.length_of(3)
     resource_records = [rr for rr_set in rrsets for rr in rr_set.resource_records]
     resource_records.should.contain("1.2.3.4")
     resource_records.should.contain("5.6.7.8")
+    resource_records.should.contain("bar")
```
bblommers, line 139: This should fail now, as type=A does not match the provided tye TXT.

Can we change the existing assertions here, to validate the order in which the items are returned?
As far as I can tell, there is no definitive validation yet that the new `sorted(..)` logic works, but this request would be a good test

bblommers (changes requested):
Hi @zubenkoivan, thanks for your PR! Please see my comments

moto/route53/models.py
zubenkoivan: Currently "." in the end is always removed, but it should not be according to the example from the docs.

moto/route53/models.py
zubenkoivan: Because this predicate is used in `dropwhile` function it will drop record sets until the first one which has name >= start_name and type >= start_type will be found. So the condition is correct.

tests/test_route53/test_route53.py
zubenkoivan: https://github.com/spulec/moto/pull/3864/files#diff-a426c01b3d9e83d976f9c3f7a209b1f988f37da5cfe37e1546100cd9606d9994R128-R132

This one tests that records have correct order. Will it be enough?

zubenkoivan: According to docs types should not match. You need to return all the types that are greater or equal to the start_type.

moto/route53/models.py
zubenkoivan: I made some tests in AWS Route53 with and without "." in the end. I didn't see any difference in behaviour. May be I miss something in the example provided in the docs. I reverted the changes. Thanks!

moto/route53/models.py
bblommers: Gotcha, my bad

tests/test_route53/test_route53.py
bblommers: Yup - I misread the documentation. All good.

bblommers (approved):
LGTM - thanks again @zubenkoivan!

zubenkoivan:
Thanks for review! @bblommers

PR #3869, opened 2021-04-21
Add the abilitiy to set ECS task definition task role and execution role arns

william-richard:
It doesn't actually use or verify these roles, but I didn't see a good way to do that in the `ecs_backend`.


bblommers:
Complete validation would involve checking against the IAM backend whether these roles exist, similar to how Lambda does it: https://github.com/spulec/moto/blob/master/moto/awslambda/models.py#L954

Persisting them without validation is a sensible first step though, and no doubt already useful for loads of people.

bblommers (approved):
LGTM. Thanks for the PR @william-richard!

PR #3871, opened 2021-04-21
List instances Implementation in EMR

jdeepe:
Implementation of list_instances in emr .  (https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/emr.html#EMR.Client.list_instances)


moto/emr/models.py
```diff
@@ -359,6 +412,7 @@ def __init__(self, region_name):
         self.region_name = region_name
         self.clusters = {}
         self.instance_groups = {}
+        self.instances = []
```
bblommers, line 415: Doesn't look like this is used anywhere - we should be able to retrieve all instances by just iterating over the clusters, if we ever need a full list

moto/emr/models.py
```diff
@@ -390,6 +444,28 @@ def add_instance_groups(self, cluster_id, instance_groups):
...
+        cluster = self.clusters[cluster_id]
+        response = self.ec2_backend.add_instances(
+            EXAMPLE_AMI_ID, instances["instance_count"], "", [], **instances
+        )
+        for instance in response.instances:
+            instance = FakeInstance(
+                cluster_id=cluster_id,
+                instance_role=instances["instance_role"],
+                ec2_instance_id=instance.id,
+                instance_type=instance.instance_type,
```
bblommers, line 457: Is there a specific need for this information to be replicated?
Ideally, we just store the ec2_instance_id, and all the EC2-instance properties (instance_type, public_dns, etc) can be retrieved dynamically when calling `list_instances`

bblommers (changes requested):

moto/emr/models.py
jdeepe: Removed it.

moto/emr/models.py
jdeepe: Pushed the changes.

bblommers (approved):
LGTM. Thanks for  the PR and making the changes, @jdeepe!

PR #2952, opened 2020-05-03
update SQS MaximumMessageSize from 64K to 256K

khneal:
Amazon SQS updated MaximumMessageSize from 64K to 256K back in 2013:
https://aws.amazon.com/about-aws/whats-new/2013/06/18/amazon-sqs-announces-256KB-large-payloads/

Tested with Python 2.7 and 3.7. I updated the one test that failed (as expected, since it tested the value I changed). I also see 4 `test_cloudformation.test_cloudformation_stack_crud_boto3` tests with errors, but the errors also happened before I made any changes:
* `test_create_change_set_from_s3_url`
* `test_create_stack_from_s3_url`
* `test_create_stack_set_from_s3_url`
* `test_update_stack_from_s3_url`

PR #3873, opened 2021-04-23
EMR and SWF - add arn to response

khneal:
Reference:
* https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/emr.html#EMR.Client.list_clusters
* https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/swf.html#SWF.Client.list_domains

### DIFF
--- a/moto/swf/models/domain.py
+++ b/moto/swf/models/domain.py
@@ -1,7 +1,7 @@
 from __future__ import unicode_literals
 from collections import defaultdict
 
-from moto.core import BaseModel
+from moto.core import ACCOUNT_ID, BaseModel
 from ..exceptions import (
     SWFUnknownResourceFault,
     SWFWorkflowExecutionAlreadyStartedFault,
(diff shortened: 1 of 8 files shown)


khneal:
FYI - this is my second contribution to this repo. #2952 was my first.

moto/swf/models/domain.py
```diff
@@ -1,18 +1,19 @@
...
+from moto.core import ACCOUNT_ID, BaseModel
 from ..exceptions import (
     SWFUnknownResourceFault,
     SWFWorkflowExecutionAlreadyStartedFault,
 )
 
 
 class Domain(BaseModel):
-    def __init__(self, name, retention, description=None):
+    def __init__(self, name, retention, description=None, swf_backend=None):
```
bblommers, line 12: I think we should pass in the region directly, instead of passing in the entire backend-object

moto/swf/models/domain.py
```diff
@@ -31,6 +32,10 @@ def to_short_dict(self):
         hsh = {"name": self.name, "status": self.status}
         if self.description:
             hsh["description"] = self.description
+        region = self.swf_backend.region_name if self.swf_backend else "unknown"
```
bblommers, line 35: We should always know which region we're in, so the backend/region should never be None

bblommers (changes requested):
Hi @khneal, thanks for the PR! Please see my comments

moto/swf/models/domain.py
khneal: Done. It was originally optional because I didn't see an obvious region to use for testing... but I just picked a test region like other testing modules.

moto/swf/models/domain.py
khneal: Done. The original implementation followed what I saw in EMR, but this works too. 😃

bblommers (approved):
LGTM 👍

PR #3878, opened 2021-04-25
Fix:SQS Added support for system attributes in sqs

usmangani1:
Fix:SQS Added support for system attributes in sqs

https://github.com/localstack/localstack/issues/3686


moto/sqs/responses.py
```diff
@@ -227,7 +227,9 @@ def send_message(self):
         if len(message) > MAXIMUM_MESSAGE_LENGTH:
             return ERROR_TOO_LONG_RESPONSE, dict(status=400)
 
-        message_attributes = parse_message_attributes(self.querystring)
+        message_attributes, aws_trace_header = parse_message_attributes(
```
bblommers, line 230: Looks like the `parse_message_attributes` now has completely duplicate logic. Would be better if we can reuse the same method:
```
message_attributes = parse_message_attributes(self.querystring)  # default value for key will be "MessageAttribute"
system_attributes = parse_message_attributes(self.querystring, key="MessageSystemAttribute")
```

moto/sqs/responses.py
```diff
@@ -246,6 +248,7 @@ def send_message(self):
             delay_seconds=delay_seconds,
             deduplication_id=message_dedupe_id,
             group_id=message_group_id,
+            aws_trace_header=aws_trace_header,
```
bblommers, line 251: With the above change, we can just store the `system_attributes` as a whole. The exact properties are not relevant to this method. 
For now it may only contain the aws_trace_header, but if AWS ever decides to store more attributes, we don't have to change anything

bblommers (changes requested):

moto/sqs/responses.py
usmangani1: Done,Changed as requested

moto/sqs/responses.py
usmangani1: Done,Changed as requested

moto/sqs/responses.py
```diff
@@ -596,6 +600,12 @@ def list_queue_tags(self):
             <Value>{{ message.group_id }}</Value>
           </Attribute>
           {% endif %}
+          {% if message.aws_trace_header is not none %}
```
bblommers, line 603: This check needs to be modified as well. I'm guessing that's blocking the tests, so when they pass, I'll be happy to merge

bblommers (changes requested):

moto/sqs/responses.py
usmangani1: Done

bblommers (approved):
LGTM

PR #3881, opened 2021-04-28
Rewrite deprecated EC2 tests

bblommers:
See #3842 - adds equivalent boto3 tests for EC2, so that the deprecated tests can be removed without losing any test coverage


moto/ec2/responses/elastic_block_store.py
```diff
@@ -268,7 +268,7 @@ def reset_snapshot_attribute(self):
   <ownerId>{{ snapshot.owner_id }}</ownerId>
   <volumeSize>{{ snapshot.volume.size }}</volumeSize>
   <description>{{ snapshot.description }}</description>
-  <encrypted>{{ snapshot.encrypted }}</encrypted>
+  <encrypted>{{ 'true' if snapshot.encrypted else 'false' }}</encrypted>
```
bblommers, line 271: These changes were required to get the new tests to pass - I guess boto is more flexible in how it parses the response

PR #3398, opened 2020-10-21
Fix: `nextToken` value in `logs:DescribeLogGroups` response

bpandola:
The pagination for this endpoint has been modified to more closely
model the real AWS behavior:
* Log Groups are now sorted alphabetically by `logGroupName`.
* `nextToken` is now a string containing the last `logGroupName` in the
  current response.
* Specifying an invalid `nextToken` does not generate an error, but does
  return an empty group list.
* `nextToken` is not included in the response if there are no additional
  items to return.

Fixes #3395

PR #3883, opened 2021-04-28
Fix: `nextToken` value in `logs:FilterLogEvents` response

### LINKED ISSUE #3882: logs client nextToken failure in filter_log_events
Similar to the problem reported in #3395 for `describe_log_groups`, `filter_log_events` shows the same errors when you try to use the `nextToken` to page through the events.

I'm working on a PR that uses a very similar fix to the one @bpandola implemented in #3398

neil-greenwood:
Plagiarizing freely from @bpandola and his PR #3398, I have
modified the pagination for FilterLogEvents to more closely follow
the real AWS behaviour.

Fixes #3882

### DIFF
--- a/moto/logs/models.py
+++ b/moto/logs/models.py
@@ -366,13 +366,32 @@ class LogGroup:
         if interleaved:
             events = sorted(events, key=lambda event: event["timestamp"])
 
-        if next_token is None:
-            next_token = 0
+        first_index = 0
+        if next_token:
+            try:
+                group, stream, event_id = next_token.split('/')
+                if group != log_group_name:
+                    raise ValueError()
+                first_index = (
+                    next(
+                        index for (index, e) in enumerate(events)
+                        if e["logStreamName"] == stream and e["eventId"] == event_id
+                    )
+                    + 1
+                )
+            except (ValueError, StopIteration):
+                first_index = 0
+                # AWS returns an empty list if it receives an invalid token.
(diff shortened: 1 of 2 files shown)


moto/logs/models.py
```diff
@@ -366,13 +366,35 @@ def filter_log_events(
...
             events = sorted(events, key=lambda event: event["timestamp"])
 
-        if next_token is None:
-            next_token = 0
+        first_index = 0
+        if next_token:
+            try:
+                group, stream, event_id = next_token.split("/")
+                if group != log_group_name:
+                    raise ValueError()
```
bblommers, line 374: Hi @neil-greenwood, thanks for the PR! Can we have a test for this error scenario as well?

bblommers (changes requested):
Hi

moto/logs/models.py
neil-greenwood: Sure.

neil-greenwood:
How important is the python2.7 compatibility? I'm struggling to get the timestamp working. I guess I'm being too fancy, and I should just use `int(time.time())` instead of bothering with the `datetime` library.

bblommers:
> How important is the python2.7 compatibility?

Very - as long as we support it :) `time.time()` should definitely work, or alternatively `datetime.now()`

bblommers (approved):
LGTM!

PR #3887, opened 2021-04-29
S3 - Fix exception for missing versionID

amarjandu:
Fixes exception that is raised if a versionID is not found for calls to s3.


bblommers (approved):
LGTM - thanks for the PR @amarjandu!

Issue #3866, opened 2021-04-18
enforce dynamodb range key size limit


mdavis-xyz:
Dynamodb has a range key [size limit](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Limits.html) of 1024 bytes.
Moto does not enforce this.

# Steps to Reproduce

```
import boto3
from moto import mock_dynamodb2

table_name = 'my-table'
limit = 1024

with mock_dynamodb2():
    client = boto3.client('dynamodb')
    response = client.create_table(
        AttributeDefinitions=[
            {
                'AttributeName': 'my_hash',
                'AttributeType': 'S'
            },
            {
                'AttributeName': 'my_range',
                'AttributeType': 'S'
            },
        ],
        TableName=table_name,
        KeySchema=[
            {
                'AttributeName': 'my_hash',
                'KeyType': 'HASH'
            },
            {
                'AttributeName': 'my_range',
                'KeyType': 'RANGE'
            },
        ]
    )
    
    
    boto3.resource('dynamodb').Table(table_name).put_item(
        Item={
            'my_hash': 'abc',
            'my_range': 'x' * (limit + 1)
        }
    )
```

# Expected Behavior

This script throws an exception at `put_item` because the range key is too large

Here's what happens when you do this on the real dynamodb:

```
botocore.exceptions.ClientError: An error occurred (ValidationException) when calling the PutItem operation: One or more parameter values were invalid: Aggregated size of all range keys has exceeded the size limit of 1024 bytes
```

# Actual behavior

Script does not throw exception.

# Solution

I'm not sure if the solution should be similar to #1874 .

Should I just put an `if` statement inside `def put_item()` [here](https://github.com/spulec/moto/blob/9b3e9328227c795bfe5896a89b19cd4989407b14/moto/dynamodb2/models/__init__.py#L598)?

mdavis-xyz:
While we're at it, we can implement the min limit (1 byte) too, and the limits for the hash key.

mdavis-xyz:
We also probably need to do the same for update table.

Should I use a custom setter function instead?

bblommers:
Hi @mdavis-xyz, thanks for raising this!
Would be nice if the solution is similar to #1874, so have that validation a single place: https://github.com/spulec/moto/blob/master/moto/dynamodb2/models/__init__.py#L62

At the moment, the LimitedSizeDict does not have any context, i.e. it doesn't know what date it stores and which key is hash/range.

[...]

PR #3888, opened 2021-04-30
Enforce dynamodb key size limit [solves #3866]

mdavis-xyz:


mdavis-xyz:
Solves #3866

bblommers (approved):
LGTM - thanks @mdavis-xyz!

PR #3893, opened 2021-05-03
Fix- Raise InvalidParameterValue when change visibility timeout is mo…

usmangani1:
Fix- Raise InvalidParameterValue when change visibility timeout is more than permitted
https://github.com/localstack/localstack/issues/3956

### DIFF
--- a/tests/test_sqs/test_sqs.py
+++ b/tests/test_sqs/test_sqs.py
@@ -811,7 +811,7 @@ def test_change_message_visibility_than_permitted():
     with freeze_time("2015-01-01 12:00:00"):
         conn.create_queue(QueueName="test-queue-visibility")
         queue = sqs.Queue("test-queue-visibility")
-        sqs.send_message(QueueName="blah", MessageBody="derp")
+        conn.send_message(QueueName="blah", MessageBody="derp")
         messages = conn.receive_message(QueueUrl=queue.url)
         messages.should.have.length_of(1)
 
--- a/tests/test_sqs/test_sqs.py
+++ b/tests/test_sqs/test_sqs.py
@@ -818,7 +818,7 @@ def test_change_message_visibility_than_permitted():
         conn.change_message_visibility(
             QueueUrl=queue.url,
             ReceiptHandle=messages.get("Messages")[0].get("ReceiptHandle"),
-            VisibilityTimeout="360",
+            VisibilityTimeout=360,
         )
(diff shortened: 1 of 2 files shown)


tests/test_sqs/test_sqs.py
```diff
@@ -777,6 +777,38 @@ def test_send_receive_message_with_attributes_with_labels():
...
+
+        sqs.change_message_visibility(
+            queue_name="blah",
+            receipt_handle=messages[0].get("ReceiptHandle"),
+            visibility_timeout="360",
+        )
+
+    with freeze_time("2015-01-01 12:05:00"):
+
+        try:
```
bblommers, line 800: Please use the `with pytest.raises`-paradigm for exceptions like this
(https://github.com/spulec/moto/blob/master/tests/test_sqs/test_sqs.py#L241)

tests/test_sqs/test_sqs.py
```diff
@@ -777,6 +777,38 @@ def test_send_receive_message_with_attributes_with_labels():
     )
 
 
+@mock_sqs
+def test_change_message_visibility_than_permitted():
```
bblommers, line 781: We'll need to skip this test in ServerMode (similar to this test https://github.com/spulec/moto/blob/master/tests/test_sqs/test_sqs.py#L1820)

bblommers (changes requested):
A bunch of other tests are now failing as well, some because the `change_visibility`-method is also called internally, for instance when a message is send to AWS Lambda. 
The error should only be raised when the user explicitly calls `change_message_visibility`

tests/test_sqs/test_sqs.py
usmangani1: Done

tests/test_sqs/test_sqs.py
usmangani1: Done

bblommers:
Hi @usmangani1, the `change_message_visibility` method is only available on the [SQS Client](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/sqs.html#client) object, so make sure you're using that:
```
client = boto3.client("sqs")
# res = boto3.resource("sqs")  # The Resource-object does _not_ support this method
client.change_message_visibility()
```

bblommers (approved):
LGTM - thanks!

PR #3896, opened 2021-05-04
`nextToken` value in `logs:describeLogStreams` response

neil-greenwood:
Modified the pagination for FilterLogEvents to more closely follow
the real AWS behaviour.


bblommers (changes requested):
Hi @neil-greenwood, thanks for the PR! Happy to merge once the build passes

bblommers (approved):
👍

PR #3898, opened 2021-05-04
EC2 describe_network_acls: add support for owner-id filter

khneal:
Added `ownerId` to response: https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_DescribeNetworkAcls.html#API_DescribeNetworkAcls_Examples

I thought about adding support for the filter to the parent "TaggedEC2Resource" because many ec2 resources do support it, but wasn't sure if that would be welcome. Somebody would have to check that all the subclasses support it, and possibly add logic to prevent some subclasses from using that filter.

My previous contributions to this project:
https://github.com/spulec/moto/pulls?q=is%3Apr+author%3Akhneal+is%3Aclosed

### DIFF
--- a/moto/ec2/models.py
+++ b/moto/ec2/models.py
@@ -5667,10 +5667,13 @@ class NetworkAclAssociation(object):
 
 
 class NetworkAcl(TaggedEC2Resource):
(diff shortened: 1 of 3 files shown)


bblommers (approved):
LGTM - thanks @khneal!

PR #3899, opened 2021-05-05
allow specifying the service as env var

razvan-moj:
This is required if running as a Github Action which doesn't allow passing parameters to `docker create` but allows environment variables

### DIFF
--- a/moto/server.py
+++ b/moto/server.py
@@ -3,6 +3,7 @@ from __future__ import unicode_literals
 import argparse
 import io
 import json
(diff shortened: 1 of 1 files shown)


bblommers (changes requested):
Hi @razvan-moj, thanks for the PR!

Looks like `black` has an issue with this change (potentially the single quotes that need to be double) - happy to merge once the CI passes

bblommers (approved):
👍

PR #3903, opened 2021-05-05
Implement User Pool MFA Actions

### LINKED ISSUE #3901: [Cognito-IDP] Add Support For User Pool MFA Endpoints
# Description

The endpoints `SetUserPoolMfaConfig` and `GetUserPoolMfaConfig` are not implemented. The goal of this issue is to implement them.

I am actively working on this on my fork, [on this branch](https://github.com/George-lewis/moto/tree/add_user_pool_mfa_actions).

I will raise a PR shortly.

George-lewis:
## Description

Closes #3901

---

- Implements `set_user_pool_mfa_config`
- Implements `get_user_pool_mfa_config`
- Updates `CognitoUserPool`'s JSON representation to include `MfaConfiguration` field so that the `DescribeUserPool` endpoint remains correct
- Creates tests for the new actions

## Question

I have added some checks in the response layer to ensure that the input aligns with what boto3 expects, and then raise a `ValueError` if it doesn't. I don't see anything similar being done anywhere else, so I ask: Should I be raising a different error? Should I be checking at all? 

Thanks.

### DIFF
--- a/moto/cognitoidp/models.py
+++ b/moto/cognitoidp/models.py
@@ -83,6 +83,10 @@ class CognitoIdpUserPool(BaseModel):
         self.creation_date = datetime.datetime.utcnow()
         self.last_modified_date = datetime.datetime.utcnow()
 
+        self.mfa_config = "OFF"
+        self.sms_mfa_config = None
+        self.token_mfa_config = None
+
         self.clients = OrderedDict()
         self.identity_providers = OrderedDict()
         self.groups = OrderedDict()
@@ -105,6 +109,7 @@ class CognitoIdpUserPool(BaseModel):
             "Status": self.status,
             "CreationDate": time.mktime(self.creation_date.timetuple()),
             "LastModifiedDate": time.mktime(self.last_modified_date.timetuple()),
+            "MfaConfiguration": self.mfa_config,
         }
 
     def to_json(self, extended=False):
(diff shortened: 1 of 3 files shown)


bblommers:
Hi @George-lewis, thanks for this PR!

> Should I be raising a different error? Should I be checking at all?

There are a few custom exceptions for CognitoIDP already, defined in exceptions.py, and raised in models.py. It looks like you should be able to reuse the existing InvalidParameterException - raising that should automatically align the error output with what boto3 expects.

Would be good to have a few unit tests just for those error scenarios as well. A good example on how to verify the correct error is raised is in `test_resource_server`: https://github.com/spulec/moto/blob/master/tests/test_cognitoidp/test_cognitoidp.py#L1555

George-lewis:
Hey @bblommers, thank you for your quick reply!

I have updated the validation code to use `InvalidParameterException` instead, and have tested boto3 and adjusted my validation code to more closely match boto3's behaviour. (e.g. sms config and token config are only required if you're enabling mfa.)

I've written a bunch of tests for errors as well. I should have every case covered now.

Let me know what you think. Thanks!

bblommers:
> Boto3 seems to check that `SnsCallerArn` matches a particular pattern, except for some reason this doesn't seem to happen in our tests; We enable SMS MFA with an incorrectly formatted ARN and boto3 doesn't complain
> 
>     * Should we be checking?

Ideally Moto implements the exact same behaviour, including validation cases like that. Rome wasn't build in a day though, so I'm happy for that to be a future enhancement.

bblommers (approved):
Thanks for making the changes!

George-lewis:
@bblommers Is there an approx release date for 2.0.7?

bblommers:
@George-lewis Nothing set in stone yet for a release, but probably at some point next week

bblommers:
> 
> 
> @bblommers Is there an approx release date for 2.0.7?

@George-lewis 2.0.7 was just released!

George-lewis:
> > @bblommers Is there an approx release date for 2.0.7?
> 
> @George-lewis 2.0.7 was just released!

Excellent! Thank you.

PR #3906, opened 2021-05-06
Register classes in `logs/models.py` as models

danielsamuels:
This change allows these object types to appear in the Moto API dashboard


bblommers (approved):
Nice one - thanks @danielsamuels!

Issue #3905, opened 2021-05-06
Calling SecretsManager `rotate_secret` with a `RotationLambdaARN` doesn't invoke the function


danielsamuels:
**How to reproduce the issue**

```python
from moto.tests.test_awslambda.test_lambda import get_test_zip_file1

secretsmanager_client = boto3.client('secretsmanager')
awslambda_client = boto3.client('awslambda')

secret = secretsmanager_client.create_secret(
    Name='secret',
    SecretString=b'',
)
func = awslambda_client.create_function(
    FunctionName="testFunction",
    Runtime="python2.7",
    Role=get_role_name(),
    Handler="lambda_function.lambda_handler",
    Code={"ZipFile": get_test_zip_file1()},
    Description="test lambda function",
    Timeout=3,
    MemorySize=128,
    Publish=True,
)

secretsmanager_client.rotate_secret(
    SecretId=secret['ARN'],
    RotationLambdaARN=func['FunctionArn'],
    RotationRules=dict(
        AutomaticallyAfterDays=30,
    ),
)
```

**What I expected to happen / What actually happens**

When [`rotate_secret`](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/secretsmanager.html#SecretsManager.Client.rotate_secret) is called with a lambda ARN and rotation rules, the Lambda function should be invoked 4 times as part of a [4 step rotation process](https://docs.aws.amazon.com/secretsmanager/latest/userguide/rotating-secrets-lambda-function-overview.html#rotation-explanation-of-steps). Currently in moto, [the Lambda function is not invoked at all](https://github.com/spulec/moto/blob/d3f20694a90209d1dd39f27e528edfa7d170d429/moto/secretsmanager/models.py#L425-L429). 

Presumably there is a desire to not make secretsmanager depend on Docker like awslambda does, but it would be useful if the functionality was possible if both services were enabled.

**Other details**

Moto version 2.0.5
Both standalone server and mocks.

PR #3912, opened 2021-05-07
Support rotating secrets using Lambda [#3905]

danielsamuels:
The Secrets manager rotation process uses an AWS Lambda function
to perform the rotation of a secret. [1]

In fact, it's not possible to trigger rotation of a Secret
without specifying a Lambda function at some point in the life
of the secret:

```
$ aws secretsmanager rotate-secret --secret-id /rotationTest

An error occurred (InvalidRequestException) when calling the RotateSecret operation: No Lambda rotation function ARN is associated with this secret.
```

`moto` can be a little more lenient in this regard and allow
`rotate_secret` to be called without a Lambda function being
present, if only to allow simulation of the `AWSCURRENT` and
`AWSPREVIOUS` labels moving across versions.

However, if a lambda function _has_ been specified when calling
`rotate_secret` [2], it should be invoked therefore providing the
developer with the full multi-stage process [3] which can be
used to test the Lambda function itself and ensuring that full
end-to-end testing is performed. Without this there's no easy
way to configure the Secret in the state needed to provide the
Lambda function with the data in the format it needs to be in
at each step of the invocation process.

[1]: https://docs.aws.amazon.com/secretsmanager/latest/userguide/rotating-secrets-lambda-function-overview.html
[2]: https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/secretsmanager.html#SecretsManager.Client.rotate_secret
[3]: https://docs.aws.amazon.com/secretsmanager/latest/userguide/rotating-secrets-lambda-function-overview.html#rotation-explanation-of-steps

### DIFF
--- a/moto/secretsmanager/models.py
+++ b/moto/secretsmanager/models.py
@@ -9,6 +9,7 @@ import datetime
 from boto3 import Session
 
 from moto.core import BaseBackend, BaseModel
+from moto.awslambda.models import lambda_backends
 from .exceptions import (
     SecretNotFoundException,
     SecretHasNoValueException,
@@ -170,6 +171,7 @@ class SecretsManagerBackend(BaseBackend):
         super(SecretsManagerBackend, self).__init__()
         self.region = region_name
         self.secrets = SecretsStore()
+        self.lambda_backend = lambda_backends[self.region]
 
     def reset(self):
         region_name = self.region
@@ -325,7 +327,11 @@ class SecretsManagerBackend(BaseBackend):
         if secret_id in self.secrets:
             secret = self.secrets[secret_id]
             secret.update(description, tags)
-            secret.reset_default_version(secret_version, version_id)
+
+            if 'AWSPENDING' in version_stages:
+                secret.versions[version_id] = secret_version
+            else:
+                secret.reset_default_version(secret_version, version_id)
         else:
             secret = FakeSecret(
                 region_name=self.region,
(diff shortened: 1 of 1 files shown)


moto/secretsmanager/models.py
```diff
@@ -413,24 +419,44 @@ def rotate_secret(
...
         if rotation_rules:
             secret.auto_rotate_after_days = rotation_rules.get(rotation_days, 0)
         if secret.auto_rotate_after_days > 0:
             secret.rotation_enabled = True
 
-        if "AWSCURRENT" in old_secret_version["version_stages"]:
-            old_secret_version["version_stages"].remove("AWSCURRENT")
+        # Begin the rotation process for the given secret by invoking the lambda function.
+        if secret.rotation_lambda_arn:
+            request_headers = {}
```
bblommers, line 441: The lambda-import should be done as late as possible, so inside this if-statement.
That way, other people can still call other SecretsManager-methods without needing the AWSLambda dependencies

bblommers (changes requested):
Hi @danielsamuels, thanks for the PR! 

Would you mind adding a few tests for this, to verify that everything works as it should?

danielsamuels:
@bblommers I've written an integration test between Secrets Manager and Lambda (and shaved the necessary yaks to make that possible).

moto/secretsmanager/models.py
```diff
@@ -592,6 +651,54 @@ def untag_resource(self, secret_id, tag_keys):
...
+        self, secret_id, version_stage, remove_from_version_id, move_to_version_id
+    ):
+        if secret_id not in self.secrets.keys():
+            raise SecretNotFoundException()
+
+        secret = self.secrets[secret_id]
+
+        if remove_from_version_id:
+            if remove_from_version_id not in secret.versions:
+                raise InvalidParameterException(
```
bblommers, line 664: Thanks for the tests @danielsamuels. Ideally we have some tests for these error scenarios as well, using the regular boto3 Client.
Example on how we do these negative tests in moto can be found here: https://github.com/spulec/moto/blob/master/tests/test_secretsmanager/test_list_secrets.py#L133

bblommers (changes requested):

moto/secretsmanager/models.py
danielsamuels: Added in 5c51a48

bblommers (approved):
Awesome - thanks @danielsamuels!

PR #3914, opened 2021-05-07
Fix separator in filter_log_events nextToken value.

neil-greenwood:
The old separator could be present in some of the values being concatenated.

### DIFF
--- a/moto/logs/models.py
+++ b/moto/logs/models.py
@@ -389,7 +389,7 @@ class LogGroup(BaseModel):
         first_index = 0
         if next_token:
             try:
-                group, stream, event_id = next_token.split("/")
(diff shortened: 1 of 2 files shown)


neil-greenwood:
> Does AWS use '/' as separators? Ideally we'd keep/replicate that behavior

Moto is faking the nextToken value. AWS uses a large hexadecimal string, whereas we just need something to keep track for reading the next page.

neil-greenwood:
Using the / was causing a test failure in my application when I used a real log group name, instead of the 'dummy' value that was in the moto test-case.

bblommers:
Gotcha, that's fine then. Just checking, as AWS does sometimes use literal values as tokens, instead of a hexadecimal value

PR #3919, opened 2021-05-11
ResourceGroup - Add Support for getGroupConfiguration

usmangani1:
ResourceGroup - Add Support for getGroupConfiguration

### DIFF
--- a/moto/resourcegroups/models.py
+++ b/moto/resourcegroups/models.py
@@ -12,7 +12,9 @@ from .exceptions import BadRequestException
 
 
 class FakeResourceGroup(BaseModel):
-    def __init__(self, name, resource_query, description=None, tags=None):
+    def __init__(
+        self, name, resource_query, description=None, tags=None, configuration=None
+    ):
         self.errors = []
         description = description or ""
         tags = tags or {}
@@ -28,6 +30,7 @@ class FakeResourceGroup(BaseModel):
         self.arn = "arn:aws:resource-groups:us-west-1:{AccountId}:{name}".format(
             name=name, AccountId=ACCOUNT_ID
         )
+        self.configuration = configuration
 
     @staticmethod
     def _format_error(key, value, constraint):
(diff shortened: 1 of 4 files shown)


moto/resourcegroups/responses.py
```diff
@@ -22,8 +22,13 @@ def create_group(self):
...
         resource_query = self._get_param("ResourceQuery")
         tags = self._get_param("Tags")
+        configuration = self._get_param("Configuration")
         group = self.resourcegroups_backend.create_group(
-            name=name, description=description, resource_query=resource_query, tags=tags
+            name=name,
+            description=description,
+            resource_query=resource_query,
+            tags=tags,
+            configuration=configuration,
```
bblommers, line 31: This should have it's own test, to confirm that we can create a group that already contains a Configuration

moto/resourcegroups/models.py
```diff
@@ -350,6 +359,16 @@ def update_group_query(self, group_name, resource_query):
         self.groups.by_name[group_name].resource_query = resource_query
         return self.groups.by_name[group_name]
 
+    def get_group_configuration(self, group_name):
+        group = self.groups.by_name.get(group_name)
+        configuration = group.configuration
+        return configuration
+
+    def put_group_configuration(self, group_name, configuration):
+        if configuration:
```
bblommers, line 368: Is this a valid scenario? I'm assuming that the Configuration-param is required when calling this method

bblommers (changes requested):
Hi @usmangani1, please see my comments.

One of the failing tests (test_s3_server_post_without_content_length) was due to an issue with our CI - please merge in the latest changes from master if this test is still failing on the next commit

moto/resourcegroups/responses.py
usmangani1: done

moto/resourcegroups/models.py
usmangani1: Done

bblommers (approved):
LGTM - thank you!

PR #3924, opened 2021-05-12
Fix sqs message retention logic

Joddev:
Sqs should check created timestamp of message, not queue attribute

### DIFF
--- a/tests/test_sqs/test_sqs.py
+++ b/tests/test_sqs/test_sqs.py
@@ -314,6 +314,29 @@ def test_message_retention_period():
 
 
 @mock_sqs
+def test_queue_retention_period():
+    sqs = boto3.resource("sqs", region_name="us-east-1")
+    queue = sqs.create_queue(
+        QueueName="blah", Attributes={"MessageRetentionPeriod": "3"}
+    )
+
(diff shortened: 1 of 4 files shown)


tests/test_sqs/test_sqs.py
```diff
@@ -313,6 +313,29 @@ def test_message_retention_period():
     assert len(messages) == 0
 
 
+@mock_sqs
+def test_queue_retention_period():
```
bblommers, line 317: Do we need this test? From what I can see, the correct behaviour is already coveredin the test above (`test_message_retention_period`)

bblommers (changes requested):
Hi @Joddev, thanks for the PR! Please see my comment

tests/test_sqs/test_sqs.py
bblommers: Ah, ignore me - I just realized why this test is useful.

bblommers (changes requested):
Looks like there are a few SNS-tests that now fail. The change looks good to me though - happy to merge once the build passes

bblommers (approved):
LGTM - thanks @Joddev!

Issue #3921, opened 2021-05-12
Werkzeug 2.0.0 breaks tests


bblommers:
The S3 server test `test_s3_server_post_without_content_length` no longer works with werkzeug 2.0.0, as the calls return a 200 instead of a 411.

I'm submitting a PR to downgrade werkzeug for now, back to the known working version 1.0.1, until we have a solution.

bblommers:
Werkzeug was downgraded as part of https://github.com/spulec/moto/pull/3920

bblommers:
The failing test was a red herring, caused by a mismatch in flask/werkzeug versions. Our CI cached flask to 1.x, but upgraded werkzeug to 2.x, (or vice versa.. I forgot) and they don't work mix'n'match very well.

When upgrading both to 2.x, most of the tests work. The only thing failing is the exception handling in the `managedblockchain` service. This was always handled badly, but didn't break until the `werkzeug` upgrade.

I'm rewriting the exception handling now - TBC..

PR #3925, opened 2021-05-12
Lock Flask

### LINKED ISSUE #3922: Add support for Jinja version 3.0.0
Flask has released a new version that requires Jinja>=3.0.0. There are a lot of improvements in this version upgrade. Currently, this library is requiring that Jinja < 3.0.0.

guiscaranse:
This should fix #3922 and #3921. Pinning only Werkzeug doesn't solve the issue since sometimes flask dependency might be installed first creating an impossible lock state when using tools like Pipenv.

Captura de Tela 2021-05-12 às 13 13 16

This is a temporary fix until this project upgrades to Flask 2

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -99,7 +99,7 @@ all_extra_deps = [
     _dep_sshpubkeys_py2,
     _dep_sshpubkeys_py3,
 ]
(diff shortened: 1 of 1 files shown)


bblommers:
Good catch @guiscaranse - thanks for creating the PR

bblommers (approved):

Issue #3928, opened 2021-05-13
ssm parameters (cloudformation) are not recognized and resolved


sahilshah6196:
Hello @bblommers 
Right now, ssm parameters (specific to cloudformation that have the type `AWS::SSM::Parameter::`) are not recognized and hence not resolved. 

Here is the comment that mentions that as well: https://github.com/spulec/moto/blob/master/moto/cloudformation/parsing.py#L87

To give an entire example,
1. Here is my small template to replicate the error:
```
AWSTemplateFormatVersion: 2010-09-09
 
Parameters:

  SubnetIds:
    Description: Load balancer subnets IDs
    Type: AWS::SSM::Parameter::Value<List<String>>

Resources:

  LoadBalancer:
    Type: AWS::ElasticLoadBalancingV2::LoadBalancer
    Properties:
      IpAddressType: ipv4
      Scheme: internal
      Type: application
      Subnets: !Ref SubnetIds
```

2. Here is the code to create and mock the required services:

```
def _scaffold_network(subnet_ids_ssm, security_group_ids_ssm, vpc_id_ssm):

    ec2_client = boto3.client('ec2', 'us-east-1')

    vpc_response    = ec2_client.create_vpc(CidrBlock='10.0.0.0/16')
    subnet_response = ec2_client.create_subnet(CidrBlock='10.0.1.0/24', VpcId=vpc_response['Vpc']['VpcId'], AvailabilityZone="us-east-1a")

    security_group_response = ec2_client.create_security_group(
        Description=generate_string(),
        GroupName=generate_string(),
        VpcId=vpc_response['Vpc']['VpcId'])

    _add_ssm_parameter(ssm_parameter_name=subnet_ids_ssm, ssm_parameter_value=subnet_response['Subnet']['SubnetId'])
    _add_ssm_parameter(ssm_parameter_name=security_group_ids_ssm, ssm_parameter_value=security_group_response['GroupId'])
    _add_ssm_parameter(ssm_parameter_name=vpc_id_ssm, ssm_parameter_value=vpc_response['Vpc']['VpcId'])


def _add_ssm_parameter(ssm_parameter_name, ssm_parameter_value):

    client = boto3.client('ssm', region_name='us-east-1')

    put_response = client.put_parameter(Name=ssm_parameter_name, Value=ssm_parameter_value)

    response = client.get_parameter(Name=ssm_parameter_name)
```

3. Here is the output (error):

```
[...]

PR #3929, opened 2021-05-13
SSM parameters in cloudformation (`AWS::SSM::Parameter::`) are not recognized and resolved

sahilshah6196:
When using ssm parameters in cloudforamtion, and `Ref` node is replaced with the corresponding `physical_resource_id.`

Have mention more here:

Related issue:
https://github.com/spulec/moto/issues/3928

### DIFF
--- a/moto/cloudformation/parsing.py
+++ b/moto/cloudformation/parsing.py
@@ -41,6 +41,7 @@ from moto.sagemaker import models as sagemaker_models  # noqa
 from moto.sns import models as sns_models  # noqa
 from moto.sqs import models as sqs_models  # noqa
 from moto.stepfunctions import models as stepfunctions_models  # noqa
+from moto.ssm import models as ssm_models, ssm_backend  # noqa
 
 # End ugly list of imports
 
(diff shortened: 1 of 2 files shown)


bblommers (changes requested):
Hi @sahilshah6196, thanks for the PR! Change looks good to me, just need to make sure that the tests passes on our CI

sahilshah6196:
> Hi @sahilshah6196, thanks for the PR! Change looks good to me, just need to make sure that the tests passes on our CI

I believe I needed to specify the `region_name` [here](https://github.com/spulec/moto/blob/17231741297f96a571e455805f0132529ac117e2/tests/test_cloudformation/test_stack_parsing.py#L518)

moto/cloudformation/parsing.py
```diff
@@ -41,6 +41,7 @@
 from moto.sns import models as sns_models  # noqa
 from moto.sqs import models as sqs_models  # noqa
 from moto.stepfunctions import models as stepfunctions_models  # noqa
+from moto.ssm import models as ssm_models, ssm_backend  # noqa
```
bblommers, line 44: Just realized that the `ssm_backend` specifically refers to the backend for us-east-1, so it's not going to work for other regions. 
Out the top of my head, you might need to use `ssm_models.ssm_backends[region]`

bblommers (changes requested):

moto/cloudformation/parsing.py
sahilshah6196: I updated the stack to be in `us-east-1`. Reckon that would work? c90cc01

bblommers: For our test, yeah, but moto users would also be stuck using us-east-1

sahilshah6196: @bblommers Everything looks good now?

PR #3931, opened 2021-05-14
Fix tags for resource in ListTagsForResource

usmangani1:
Fix tags for resource in ListTagsForResource

https://github.com/localstack/localstack/issues/4019

### DIFF
--- a/moto/ssm/responses.py
+++ b/moto/ssm/responses.py
@@ -253,6 +253,10 @@ class SimpleSystemManagerResponse(BaseResponse):
             name, description, value, type_, allowed_pattern, keyid, overwrite, tags
         )
 
+        if tags:
+            tags = {t["Key"]: t["Value"] for t in self._get_param("Tags")}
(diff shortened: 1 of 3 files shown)


moto/ssm/responses.py
```diff
@@ -253,6 +253,10 @@ def put_parameter(self):
             name, description, value, type_, allowed_pattern, keyid, overwrite, tags
         )
 
+        if tags:
```
bblommers, line 256: This logic should be part of the `put_parameter`-method in models.py, for a few reasons:

1) The model should take care of the formatting/logic of extracting tags
2) Storing every property of a parameter should be in a single place/method
2) The `put_parameter` call might fail. Having the logic here could result in tags being added regardless, and still receiving the ParameterAlreadyExists-error

bblommers (changes requested):

moto/ssm/responses.py
usmangani1: Done

bblommers (approved):
LGTM - thanks @usmangani1!

PR #3934, opened 2021-05-15
Fix resource group get-group-query & update-query-query

usmangani1:
Fix resource group get-group-query & update-query-query

### DIFF
--- a/moto/resourcegroups/responses.py
+++ b/moto/resourcegroups/responses.py
@@ -71,6 +71,9 @@ class ResourceGroupsResponse(BaseResponse):
 
     def get_group_query(self):
         group_name = self._get_param("GroupName")
+        group_arn = self._get_param("Group")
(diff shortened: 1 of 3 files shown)


moto/resourcegroups/responses.py
```diff
@@ -165,6 +168,9 @@ def update_group(self):
     def update_group_query(self):
         group_name = self._get_param("GroupName")
         resource_query = self._get_param("ResourceQuery")
+        group_arn = self._get_param("Group")
```
bblommers, line 171: Can we get a test for the `update_group_query` as well? At the moment it just tests the get_group_query twice

usmangani1: Done

bblommers (approved):
LGTM - thanks @usmangani1!

PR #3937, opened 2021-05-18
feat: transition elbv2 state to `active` on first describe

giovtorres:
This change updates the elbv2 load balancer state from `provisioning` to `active` after the first call to `describe_load_balancers()`.


bblommers (approved):
LGTM - thanks for the PR @giovtorres!

PR #3938, opened 2021-05-19
MediaStore delete container and list tags endpoints implemented

arcovoltaico:
Idelete_container and list_tags_for_resource were not implemented. 
I did it following boto3 & AWS documentation. I added the ContainerNotFound exception which it's the actual being thrown when trying to delete an unexistent container. 
All tests are passing.

### DIFF
--- a/tests/test_mediastore/test_mediastore.py
+++ b/tests/test_mediastore/test_mediastore.py
@@ -229,7 +229,7 @@ def test_delete_container():
     response = client.delete_container(ContainerName=container["Name"])
     response["ResponseMetadata"]["HTTPStatusCode"].should.equal(200)
     containers = client.list_containers(NextToken="next-token")["Containers"]
-    container_exists = any(d['Name'] == container_name for d in containers)
+    container_exists = any(d["Name"] == container_name for d in containers)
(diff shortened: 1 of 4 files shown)


tests/test_mediastore/test_mediastore.py
```diff
@@ -192,3 +193,52 @@ def test_get_metric_policy_raises_error_if_container_does_not_have_metric_policy
...
+def test_delete_container():
+    client = boto3.client("mediastore", region_name=region)
+
+    create_response = client.create_container(
+        ContainerName="Awesome container!"
+    )
+    container = create_response["Container"]
+    response = client.delete_container(ContainerName=container["Name"])
+    response["ResponseMetadata"]["HTTPStatusCode"].should.equal(200)
+
```
bblommers, line 235: Can we verify that the deletion was actually successful? Easiest way is probably to verify that the container name does no longer show up in a `list_containers` call.

arcovoltaico: Sure, in fact, I forgot to delete it from the containers list. Hope it's sorted now. Thanks so much!

PR #3940, opened 2021-05-19
Drop dependency on zipp

musicinmybrain:
A search of the repository shows zipp is not imported anywhere, so this PR removes the dependency.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -67,8 +67,6 @@ install_requires += [
     "rsa<=4.0; python_version < '3'",
     "setuptools",
     "setuptools==44.0.0; python_version < '3'",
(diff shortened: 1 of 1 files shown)


bblommers:
Thanks for raising this @musicinmybrain.

Some background: 
This used to be an indirect dependency: `aws-xray-sdk` requires `jsonpickle` requires `importlib-metadata` requires `zipp`.
The entire `importlib-metadata` dependency was dropped as of `jsonpickle` 2x.

Not sure why `zipp` was listed independently though! We can definitely remove it now.

bblommers (approved):

PR #3953, opened 2021-05-23
SSM - load global parameters

### LINKED ISSUE #3949: Support SSM AWS Region Query
## Steps to reproduce

moto: 2.0.7
boto3: 1.17.75

```python3
from moto import mock_ssm
import boto3
import unittest

    def test_ssm_region_query(self):
        with mock_ssm():
            self.assertEqual(
                boto3.client('ssm').get_parameter(
                    Name='/aws/service/global-infrastructure/regions/us-west-1/longName'
                )['Parameter']['Value'],
                'US West (N. California)',
            )
```

Save as test.py and run with python3 -m unittest test.py

## Expected behavior

The test passes.

## Actual behavior

The test fails.

## Context

I am trying to access the recently added global ssm paramters: [https://aws.amazon.com/blogs/aws/new-query-for-aws-regions-endpoints-and-more-using-aws-systems-manager-parameter-store/](https://aws.amazon.com/blogs/aws/new-query-for-aws-regions-endpoints-and-more-using-aws-systems-manager-parameter-store/)

Unfortunately get_parameters_by_path returns an empty list.

```
client.get_parameters_by_path(
    Path='/aws/service/global-infrastructure/regions'
)
```

I tried this

```
client.put_parameter(
    Name='/aws/service/global-infrastructure/regions/us-west-1/longName',
    Value='US West (N. California)'
)
```

but it is giving me this error

```jsx
botocore.exceptions.ClientError: An error occurred (AccessDeniedException) when calling the PutParameter operation: No access to reserved parameter name: /aws/service/global-infrastructure/regions/us-west-1/longName.
```

and trying to access the parameter directly gives me this

```jsx
botocore.errorfactory.ParameterNotFound: An error occurred (ParameterNotFound) when calling the GetParameter operation: Parameter /aws/service/global-infrastructure/regions/us-west-1/longName not found.
```

bblommers:
Fixes #3949 

Docs: https://aws.amazon.com/blogs/aws/new-query-for-aws-regions-endpoints-and-more-using-aws-systems-manager-parameter-store/

### Background: 
AWS stores a bunch of parameters related to the 'global infrastructure' in SSM (region names, available services, etc).
This PR makes those parameters available in moto.

### PR contains:
 - Script to download all those parameters
 - Extension to SSM that loads the parameters in-memory when requested

Individual tests that request a parameter starting with `/aws` incur a 200 to 300 millisecond penalty for loading the data in-memory.
There is no impact on tests that only use client-provided parameters.

### Raw numbers:
 - There are around **25k** parameters in total
 - Filesize of the (storage-optimized version of this) JSON: **1.2MB**

If the current approach is too resource intensive, an alternative could be:

Users create these parameters manually, as part of their test set-up.
At the moment, there is validation in place so that users cannot create any parameters that start with `/aws/...`
We could have an environment variable to disable this validation.
```
def setup:
    os.environ['DISABLE_SSM_VALIDATION'] = True
    ssm.put_parameter("/aws/service/global-infrastructure/regions/us-west-2", ..)
    os.environ["DISABLE_SSM_VALIDATION"] = False

def test_business_logic():
    # some business logic that calls:
    ssm.get_parameter("/aws/service/...")
```
#### Changes required
Throw this PR away, and create a new PR with just the additional environment variable


bblommers:
Considering the tests are timing out, it looks like the CI agrees that the current approach is too resource intensive.
Marking it as draft until I've got a solution for this.

jeking3 (commented):
If you compress the JSON file it might be more palatable.  Given the purpose of moto as a simulator, needing to do additional setup as a consumer which requires a one-time real connection seems less than ideal.

PR #3955, opened 2021-05-24
Media store data Service

arcovoltaico:
At the moment, the decorator is available, but the execution of the test  delete_object does not use the mediaStore-data response, so boto is still called. What I am missing here? Cheers @bblommers


moto/mediastoredata/urls.py
```diff
@@ -0,0 +1,12 @@
...
+from .responses import MediaStoreDataResponse
+
+url_bases = [
+    "https?://data.mediastore.(.+).amazonaws.com",
+]
+
+response = MediaStoreDataResponse()
+
+url_paths = {
+    "{0}/$": response.dispatch,
```
bblommers, line 11: The `delete_object` calls has this signature:
`DELETE http://data.mediastore.eu-west-1.amazonaws.com/foo`

To intercept this call, the url-path has to match this pattern:
`"{0}/(?P<Path>[^/.]+)$": response.dispatch`

Note the `<Path>`-identifier, which allows us to retrieve this variable again in the Response-class:
`item_id = self._get_param("Path")`

moto/backends.py
```diff
@@ -82,6 +82,7 @@
     "mediaconnect": ("mediaconnect", "mediaconnect_backends"),
     "mediapackage": ("mediapackage", "mediapackage_backends"),
     "mediastore": ("mediastore", "mediastore_backends"),
+    "mediastore-data": ("mediastore-data", "mediastoredata_backends"),
```
bblommers, line 85: The folder does not have a dash, so this should be:
```
"mediastore-data": ("mediastoredata", "mediastoredata_backends"),
```

bblommers (changes requested):

moto/mediastoredata/urls.py
bblommers: The `mediapackage`-service has more examples on intercepting URLs like this:
https://github.com/spulec/moto/blob/master/moto/mediapackage/urls.py

arcovoltaico: I think it's ok now.

moto/backends.py
arcovoltaico: Sorted

arcovoltaico:
@bblommers I have at least 2 issues on models, marked with TODO.
One is not getting a param value, and the other is list_items pointing to get_object. Any hint? Thanks a zillion.

bblommers:
Hey @arcovoltaico 

The body of a PUT/POST request is accessible via `self.body`.
The list_items pointing to get_object was due to a bug in Moto's URL-matching logic - see #4021. This should be fixed now - just merge the latest master into your branch, and you should be good to go. Let me know if I've missed anything.

arcovoltaico:
@bblommers I have some issues marked with TODO.  I am failing to return the item(s) from the response. Thanks!

tests/test_mediastoredata/test_mediastoredata.py
```diff
@@ -0,0 +1,70 @@
...
+region = "eu-west-1"
+
+
+@mock_mediastoredata
+def test_put_object():
+    client = boto3.client("mediastore-data", region_name=region)
+    response = client.put_object(Body="011001", Path="foo")
+    response["ResponseMetadata"]["HTTPStatusCode"].should.equal(200)
+    response["StorageClass"].should.equal('TEMPORAL')
+    # TODO:  Why Only ETag is available on Items
```
bblommers, line 19: 'Path' is never returned for the list_items call. See the following example:
https://awscli.amazonaws.com/v2/documentation/api/latest/reference/mediastore-data/list-items.html#examples

From that list, only 'ETag' is currently returned by Moto. If you add type/content_length/name attributes to the response, they should be returned as well.

moto/mediastoredata/responses.py
```diff
@@ -0,0 +1,42 @@
...
+
+    @property
+    def mediastoredata_backend(self):
+        return mediastoredata_backends[self.region]
+
+    def get_object(self):
+        path = self._get_param("Path")
+        range = self._get_param("Range")
+        result = self.mediastoredata_backend.get_object(Path=path, Range=range)
+        return json.dumps(result.to_dict())
```
bblommers, line 20: AWS returns the actual content of the object as the body, but other parameters are returned as headers.
In moto, this can be done by returning a tuple:
```
headers = {"ETag": ..., "Cache-Control": "max-age=6, public", ...}
return "body of item", headers
```

tests/test_mediastoredata/test_mediastoredata.py
```diff
@@ -0,0 +1,70 @@
...
+    ex.value.response["Error"]["Code"].should.equal("ObjectNotFoundException")
+
+
+@mock_mediastoredata
+def test_get_object():
+    client = boto3.client("mediastore-data", region_name=region)
+    client.put_object(Body="011001", Path="foo")
+    response = client.get_object(Path="foo")
+    response["ResponseMetadata"]["HTTPStatusCode"].should.equal(200)
+    # TODO:  Why  the object is not included on the response?
```
bblommers, line 39: The body is returned, as far as I can tell? (`response["Body"]`) See my comment in responses.py on how to return the other parameters (etag/content-type/content-length/etc)

tests/test_mediastoredata/test_mediastoredata.py
```diff
@@ -0,0 +1,70 @@
...
+
+
+@mock_mediastoredata
+def test_delete_object_succeeds():
+    client = boto3.client("mediastore-data", region_name=region)
+    object_path = "foo"
+    new_object = client.put_object(Body="011001", Path=object_path)
+    response = client.delete_object(Path=object_path)
+    response["ResponseMetadata"]["HTTPStatusCode"].should.equal(200)
+    # TODO:  Why Only ResponseMetadata is included on items?
```
bblommers, line 57: `client.list_items()["Items"]` does work

moto/mediastoredata/models.py
```diff
@@ -0,0 +1,113 @@
...
+            "Path": self.path,
+            "ContentSHA256": self.content_sha256,
+        }
+        if exclude:
+            for key in exclude:
+                del data[key]
+        return data
+
+
+class Object(BaseModel):
```
bblommers, line 35: This object doesn't seem to be used. (As far as naming goes - I'd prefer `Object` over `CreatedObject` though)

moto/mediastoredata/models.py
```diff
@@ -0,0 +1,113 @@
...
+from collections import OrderedDict
+
+from boto3 import Session
+
+from moto.core import BaseBackend, BaseModel
+from .exceptions import ClientError
+
+
+class CreatedObject(BaseModel):
+    def __init__(self, *args, **kwargs):
```
bblommers, line 13: If there are only a few, passing in named arguments here will be easier to read then having to parse a dict

moto/mediastoredata/models.py
```diff
@@ -0,0 +1,113 @@
...
+        self._objects = OrderedDict()
+
+    def reset(self):
+        region_name = self.region_name
+        self.__dict__ = {}
+        self.__init__(region_name)
+
+    def put_object(
+        self,
+        Body,
```
bblommers, line 70: Parameters should be lowercase

moto/mediastoredata/models.py
```diff
@@ -0,0 +1,113 @@
...
+        Path,
+        ContentType=None,
+        CacheControl=None,
+        StorageClass="TEMPORAL",
+        UploadAvailability="STANDARD",
+    ):
+        new_object = CreatedObject(
+            Path=Path, ContentSHA256=Body, ETag="etag", StorageClass="TEMPORAL"
```
bblommers, line 78: The `body` and sha256 should be stored separately. Probably easiest to only pass the body, and let the `CreatedObject` figure out the hash-value (`hashlib.sha256(body).hexdigest()`)

tests/test_mediastoredata/test_mediastoredata.py
```diff
@@ -0,0 +1,68 @@
...
+@mock_mediastoredata
+def test_get_object_throws_not_found_error():
+    client = boto3.client("mediastore-data", region_name=region)
+    with pytest.raises(ClientError) as ex:
+        client.get_object(Path="foo")
+    ex.value.response["Error"]["Code"].should.equal("ObjectNotFoundException")
+
+
+@mock_mediastoredata
+def test_get_object():
```
bblommers, line 33: This should assert that the body was successfully persisted, as well

tests/test_mediastoredata/test_mediastoredata.py
```diff
@@ -0,0 +1,68 @@
...
+    object_path = "foo"
+    client.put_object(Body="011001", Path=object_path)
+    response = client.delete_object(Path=object_path)
+    response["ResponseMetadata"]["HTTPStatusCode"].should.equal(200)
+    items = client.list_items()["Items"]
+    len(items).should.equal(0)
+
+
+@mock_mediastoredata
+def test_list_items():
```
bblommers, line 61: Let's have a sanity check here that `list_items()` returns no items at the beginning

tests/test_mediastoredata/test_server.py
```diff
@@ -0,0 +1,23 @@
...
+from moto import mock_mediastoredata
+
+"""
+Test the different server responses
+"""
+
+
+@mock_mediastoredata
+def test_mediastore_lists_containers():
+    backend = server.create_backend_app("mediastoreback")
```
bblommers, line 15: Typo (`server.create_backend_app("mediastore-data")`)

tests/test_mediastoredata/test_server.py
```diff
@@ -0,0 +1,23 @@
...
+    test_client = backend.test_client()
+
+    # res = test_client.get(
```
bblommers, line 18: Haven't tested it, but I think we don't need the headers in this case. `test_client.get("/")` should call `list_items()`, so the accompanying assertion would be that `result.should.contain('"Items": []')`

tests/test_mediastoredata/test_mediastoredata.py
```diff
@@ -0,0 +1,72 @@
...
+
+@mock_mediastoredata
+def test_get_object():
+    client = boto3.client("mediastore-data", region_name=region)
+    client.put_object(Body="011001", Path="foo")
+    response = client.get_object(Path="foo")
+    response["ResponseMetadata"]["HTTPStatusCode"].should.equal(200)
+    response["ResponseMetadata"]["HTTPHeaders"]["path"].should.equal("foo")
+    data = response["Body"].read()
+    data.should.equal(b"body of item")
```
bblommers, line 40: `data.should.equal("011001")`, as that's the data you're sending in the put_object call

moto/mediastoredata/models.py
```diff
@@ -0,0 +1,92 @@
...
+from collections import OrderedDict
+
+from boto3 import Session
+
+from moto.core import BaseBackend, BaseModel
+from .exceptions import ClientError
+
+
+class Object(BaseModel):
+    def __init__(self, Path, Body, ETag, StorageClass="TEMPORAL"):
```
bblommers, line 13: Variables needs to be snakecase (path, body, etag, storage_class)

moto/mediastoredata/models.py
```diff
@@ -0,0 +1,92 @@
...
+        return new_object
+
+    def delete_object(self, path):
+        if path not in self._objects:
+            error = "ObjectNotFoundException"
+            raise ClientError(error, "Object with id={} not found".format(path))
+        del self._objects[path]
+        return {}
+
+    def get_object(self, Path, Range=None):
```
bblommers, line 69: Variables needs to be snakecase

moto/mediastoredata/models.py
```diff
@@ -0,0 +1,92 @@
...
+        objects_found = [item for item in self._objects.values() if item.path == Path]
+        if len(objects_found) == 0:
+            error = "ObjectNotFoundException"
+            raise ClientError(error, "Object with id={} not found".format(Path))
+        return objects_found[0]
+
+    def list_items(self, Path, MaxResults=1000, NextToken=None) -> []:
```
bblommers, line 76: Variables needs to be snakecase

moto/mediastore/models.py
```diff
@@ -83,6 +83,8 @@ def list_containers(self, next_token, max_results):
         return response_containers, None
 
     def list_tags_for_resource(self, name):
+        if name not in self._containers:
+            raise ContainerNotFoundException()
```
bblommers, line 87: This needs a unit test

moto/mediastoredata/exceptions.py
```diff
@@ -0,0 +1,38 @@
+from __future__ import unicode_literals
+
+from moto.core.exceptions import JsonRESTError
+
+
+class MediaStoreDataClientError(JsonRESTError):
+    code = 400
+
+
+class ContainerNotFoundException(MediaStoreDataClientError):
```
bblommers, line 10: Doesn't look like this class is used?

moto/mediastoredata/exceptions.py
```diff
@@ -0,0 +1,38 @@
...
+    def __init__(self, msg=None):
+        self.code = 400
+        super(ContainerNotFoundException, self).__init__(
+            "ContainerNotFoundException",
+            msg or "The specified container does not exist",
+        )
+
+
+class ObjectNotFoundException(MediaStoreDataClientError):
```
bblommers, line 19: Same here

moto/mediastoredata/exceptions.py
```diff
@@ -0,0 +1,38 @@
...
+    def __init__(self, msg=None):
+        self.code = 400
+        super(ObjectNotFoundException, self).__init__(
+            "ObjectNotFoundException", msg or "The specified object does not exist"
+        )
+
+
+class InternalServerError(MediaStoreDataClientError):
```
bblommers, line 27: Same here

moto/mediastoredata/models.py
bblommers, line 76: We still support Py2, which doesn't support `-> []` - so let's remove that

moto/mediastoredata/models.py
```diff
@@ -0,0 +1,92 @@
...
+
+class Object(BaseModel):
+    def __init__(self, Path, Body, ETag, StorageClass="TEMPORAL"):
+        self.path = Path
+        self.body = Body
+        self.content_sha256 = hashlib.sha256(Body.encode("utf-8")).hexdigest()
+        self.etag = ETag
+        self.storage_class = StorageClass
+
+    def to_dict(self, exclude=None):
```
bblommers, line 20: Let's remove the exclude-parameter, as we're not using it in this service

bblommers (changes requested):

bblommers (approved):
LGTM. Thanks for the PR, @arcovoltaico!

PR #3956, opened 2021-05-25
AWS Batch enhancements

thomas-maschler:
This PR adds several enhancements to Moto's AWS Batch Service

- Check exit status of container
- Add support for job dependencies
- Correctly set container properties as specified in Job
- Add AWS_BATCH_JOB_ID to container env variables


tests/test_batch/test_batch.py
```diff
@@ -886,3 +886,362 @@ def _wait_for_job_status(client, job_id, status, seconds_to_wait=30):
...
+        jobQueue=queue_arn,
+        jobDefinition=job_def_arn_success,
+        dependsOn=depends_on,
+    )
+    job_id3 = resp["jobId"]
+
+    future = datetime.datetime.now() + datetime.timedelta(seconds=30)
+
+    while datetime.datetime.now() < future:
+        resp = batch_client.describe_jobs(jobs=[job_id1, job_id2, job_id3])
```
bblommers, line 1108: We're testing that both job2 and job3 fail, right? We don't really need to check for job1, and ideally we verify that both jobs have failed
(Incidentally, that will probably fix the Py2-failure)

bblommers (changes requested):

tests/test_batch/test_batch.py
thomas-maschler: Thanks, for the review. I refactored the test accordingly.

bblommers (approved):
LGTM. Thanks for the PR @thomas-maschler!

Issue #2054, opened 2019-02-08
S3 Metadata name characters not preserved when run as server


thehesiod:
From what I see the server back-end sticks the metadata keys in an `EnvironHeaders` class, which then mucks with both the casing and certain characters like "_".

If I set a key called "src_etag", it gets returned "src-etag"  I believe the `tolower()` is correct, however the "_" -> "-" is not as it's breaking my unittests when moving to moto

spulec:
Hmmm. Is this happening in moto itself or inside of flask/werkzeug?

We will want to find a fix either way, but it will be more annoying if the latter.

thehesiod:
I believe it was in moto.

spulec:
Hmmm, I'm not able to find any reference to `EnvironHeaders` in our codebase.

The following code is working for me. Can you confirm this is still a problem against `master`?

```python
@mock_s3_deprecated
@reduced_min_part_size
def test_multipart_upload_with_headers():
    conn = boto.connect_s3('the_key', 'the_secret')
    bucket = conn.create_bucket("foobar")

    multipart = bucket.initiate_multipart_upload(
        "the-key", metadata={"src_etag": "bar"})
    part1 = b'0' * 10
    multipart.upload_part_from_file(BytesIO(part1), 1)
    multipart.complete_upload()

    key = bucket.get_key("the-key")
    key.metadata.should.equal({"src_etag": "bar"})
```

thehesiod:
@spulec note the title, you need to run moto as a server.  I just re-ran into this issue

spulec:
(sorry for delay on this)

Ah, I see. Do you have code to reproduce?

qqshfox:
Encountering similar issue here. Spawned the server inside of flask/werkzeug just like this https://github.com/ZwickyTransientFacility/alertbase/blob/87d642fbd2ba4d7c77640450f269820571192376/test/test_blobstore.py#L14

The root cause maybe similar to this https://github.com/localstack/localstack/issues/459#issuecomment-346756560.

qqshfox:
Here's the code to reproduce:

```python
import boto3.session
import moto.server
import pytest
import threading
import werkzeug


TEST_BUCKET = 'test-bucket'


@pytest.fixture
def s3_server():
    ip = "127.0.0.1"
    port = "5287"
    app = moto.server.DomainDispatcherApplication(
        moto.server.create_backend_app,
        service="s3",
    )
    app.debug = True

[...]

PR #3958, opened 2021-05-25
Add dot as a valid s3 metadata key

### LINKED ISSUE #3957: S3 custom metadata key should allow dotted values
## Reporting Bugs

S3 custom metadata should allow the use of the "dot" as part of the key name.

The regex in https://github.com/spulec/moto/blob/master/moto/s3/utils.py#L76 prevents users from doing so hence rejecting any metadata of the following form `foo.bar = something`

### Pytest test case to reproduce the issue:

```
import boto3
from moto import mock_s3


class TestS3Metadata:
    def test_s3_metadata(self):
        bucket_name = "test-medata"
        metadata = {
            "foo": "FOO",
            "bar": "BAR",
            "foo.bar": "FOO.BAR"
        }
        with mock_s3():
            s3 = boto3.client("s3")

            # Create an S3 bucket and store a file with metadata
            s3.create_bucket(Bucket=bucket_name)
            s3.put_object(
                Bucket=bucket_name,
                Key="test.txt",
                Body="File with meta",
                Metadata=metadata,
            )
            # Retrieve previously stored object metadata
            obj = s3.head_object(
                Bucket=bucket_name,
                Key="test.txt",
            )

            assert obj['Metadata'] == metadata
```

* Output of the test:

```
>           assert obj['Metadata'] == metadata
E           AssertionError: assert {'bar': 'BAR', 'foo': 'FOO'} == {'bar': 'BAR', 'foo': 'FOO', 'foo.bar': 'FOO.BAR'}
E             Common items:
E             {'bar': 'BAR', 'foo': 'FOO'}
E             Right contains 1 more item:
E             {'foo.bar': 'FOO.BAR'}
E             Full diff:
E             - {'bar': 'BAR', 'foo': 'FOO', 'foo.bar': 'FOO.BAR'}
E             + {'bar': 'BAR', 'foo': 'FOO'}

test.py:30: AssertionError
=================================================================== short test summary info ===================================================================
FAILED test.py::TestS3Metadata::test_s3_metadata - AssertionError: assert {'bar': 'BAR', 'foo': 'FOO'} == {'bar': 'BAR', 'foo': 'FOO', 'foo.bar': 'FOO.BAR'}
====================================================================== 1 failed in 1.01s ======================================================================
```

### Expected result

Test should pass as `foo.bar` is a valid S3 metadata

### What actually happens

Pip installs:
* `moto==2.0.8`
* `boto3==1.17.79`
* `botocore==1.20.79`

Lowess:
Fixes #3957

### DIFF
--- a/tests/test_s3/test_s3.py
+++ b/tests/test_s3/test_s3.py
@@ -65,13 +65,16 @@ def reduced_min_part_size(f):
 
 
 class MyModel(object):
-    def __init__(self, name, value):
+    def __init__(self, name, value, metadata={}):
         self.name = name
         self.value = value
+        self.metadata = metadata
 
     def save(self):
         s3 = boto3.client("s3", region_name=DEFAULT_REGION_NAME)
-        s3.put_object(Bucket="mybucket", Key=self.name, Body=self.value)
+        s3.put_object(
+            Bucket="mybucket", Key=self.name, Body=self.value, Metadata=self.metadata
+        )
 
 
 @mock_s3
@@ -131,6 +134,24 @@ def test_my_model_save():
     body = conn.Object("mybucket", "steve").get()["Body"].read().decode()
 
     assert body == "is awesome"
+
+
+@mock_s3
+def test_object_metadata():
+    """Metadata keys can contain certain special characters like dash and dot"""
+    # Create Bucket so that test can run
+    conn = boto3.resource("s3", region_name=DEFAULT_REGION_NAME)
+    conn.create_bucket(Bucket="mybucket")
+    ####################################
+
+    metadata = {"meta": "simple", "my-meta": "dash", "meta.data": "namespaced"}
+
+    model_instance = MyModel("steve", "is awesome", metadata=metadata)
+    model_instance.save()
+
+    meta = conn.Object("mybucket", "steve").get()["Metadata"]
+
+    assert meta == metadata
 
 
 @mock_s3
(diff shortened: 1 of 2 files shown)


bblommers (changes requested):
Hi @Lowess, thanks for providing a fix for this!
Can you add a unit test as well [here](https://github.com/spulec/moto/blob/master/tests/test_s3/test_s3.py)?

Lowess:
@bblommers test added !

tests/test_s3/test_s3.py
```diff
@@ -133,6 +136,24 @@ def test_my_model_save():
...
 
+@mock_s3
+def test_object_metadata():
+    """Metadata keys can contain certain special characters like underscore and dot"""
+    # Create Bucket so that test can run
+    conn = boto3.resource("s3", region_name=DEFAULT_REGION_NAME)
+    conn.create_bucket(Bucket="mybucket")
+    ####################################
+
+    metadata = {"meta": "simple", "my_meta": "underscore", "meta.data": "namespaced"}
```
bblommers, line 147: The tests are currently failing because of the underscore - would you mind changing this to `"my-meta": "dash"`? Happy to merge once the build passes

(Some background on why/how this is  failing can be found here: https://github.com/spulec/moto/issues/2054)

bblommers (changes requested):

tests/test_s3/test_s3.py
Lowess: Got it, makes sense. test updated.

bblommers (approved):
LGTM!

PR #3965, opened 2021-05-27
Batch - remove potential race condition

bblommers:
The initial ordering of the attribute declaration:
```
job_state = FAILED
job_stopped = True
```

Which lead to the race condition like this:
```diff
! Client thread
# Job thread

! job.run()
# Starting docker service
# Doing it's thing asyncronously...
! describe_jobs()
! No-op, as the jobs are still running
# job_state = FAILED
! job = describe_jobs()
! job.status == FAILED
! job.container != None
@@ FAIL -  job.container is only set if the `job_stopped`-attribute is True @@
# job_stopped = True
```

Switching the ordering of the attribute declaration should solve this.

### DIFF
--- a/moto/batch/models.py
+++ b/moto/batch/models.py
@@ -534,9 +534,9 @@ class Job(threading.Thread, BaseModel, DockerModel):
 
                 result = container.wait()
                 if self.stop or result["StatusCode"] != 0:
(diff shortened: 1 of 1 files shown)


bblommers:
FYI @bpandola

PR #3967, opened 2021-05-28
Enforce SSM Parameter Version Limit

### LINKED ISSUE #3963: SSM Parameter Store Should Only Support 100 Versions
Per https://docs.aws.amazon.com/systems-manager/latest/userguide/sysman-paramstore-versions.html - 

AWS SSM Parameter Store only supports 100 versions. After this point, the lowest version is deleted every time a new version is created. 

This functionality is not currently implemented in SimpleSystemManagerBackend. 

To reproduce, simply add 101 versions to a parameter and then fetch the history. Only the most recent 100 should be returned, instead all 101 are returned.

bpandola:
Behavior verified against a real AWS backend.

Closes #3963

### DIFF
--- a/moto/ssm/exceptions.py
+++ b/moto/ssm/exceptions.py
@@ -115,3 +115,12 @@ class DuplicateDocumentContent(JsonRESTError):
         super(DuplicateDocumentContent, self).__init__(
             "DuplicateDocumentContent", message
         )
+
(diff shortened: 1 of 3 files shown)


bblommers (approved):
LGTM - thanks @bpandola!

Issue #3966, opened 2021-05-27
No support for AWS::ElasticLoadBalancingV2::ListenerRule


sahilshah6196:
## Reporting Bugs

There is no support for: `AWS::ElasticLoadBalancingV2::ListenerRule`

Here is the warning:
`WARNING  moto:parsing.py:219 No Moto CloudFormation support for AWS::ElasticLoadBalancingV2::ListenerRule`

sahilshah6196:
Working on a fix for this

bblommers:
This is now part of moto >= 2.0.9

PR #3969, opened 2021-05-28
Add support for AWS::ElasticLoadBalancingV2::ListenerRule

sahilshah6196:
There is not support for `AWS::ElasticLoadBalancingV2::ListenerRule`. This PR aims to add that.

Issue:
https://github.com/spulec/moto/issues/3966


sahilshah6196:
Hey @bblommers !
I removed a few things that you suggested, but I had to leave a few things cause of the existing unit-tests. What do you think about this PR?
After this one, I will open PR's to add "proper" validation that does not exist in this PR.

moto/elbv2/models.py
```diff
@@ -214,30 +215,23 @@ def __init__(
         self.certificate = certificate
         self.certificates = [certificate] if certificate is not None else []
         self.default_actions = default_actions
-        self._non_default_rules = []
-        self._default_rule = FakeRule(
```
bblommers, line 218: What's the reasoning behind removing the default rule? If I understand the docs correctly, the default rule should be created automatically when creating a listener, regardless of any rules created by the user

moto/elbv2/models.py
```diff
@@ -262,50 +256,103 @@ def create_from_cloudformation_json(
...
+        cls, resource_name, cloudformation_json, region_name
+    ):
+        properties = cloudformation_json["Properties"]
+        elbv2_backend = elbv2_backends[region_name]
+        listener_arn = properties.get("ListenerArn")
+        priority = properties.get("Priority")
+        conditions = properties.get("Conditions")
+        # transform Actions to confirm with the rest of the code and XML templates
+        default_actions = []
+        for i, action in enumerate(properties["Actions"]):
```
bblommers, line 323: The logic here seems to be the same as when creating a AWS::ElasticLoadBalancingV2::Listener, the only difference is whether it's called `Actions` or `DefaultActions`.
Can we extract this while loop in a common method, and reuse it for both?

moto/elbv2/models.py
```diff
@@ -214,30 +215,23 @@ def __init__(
...
+        self._non_default_rules.pop(arn)
 
-    def register(self, rule):
-        self._non_default_rules.append(rule)
-        self._non_default_rules = sorted(
-            self._non_default_rules, key=lambda x: x.priority
+    def register(self, arn, rule):
+        self._non_default_rules[arn] = rule
+        self._non_default_rules = OrderedDict(
+            reversed(list(self._non_default_rules.items()))
```
bblommers, line 234: What's the reasoning behind this change?

moto/elbv2/models.py
```diff
@@ -262,50 +256,103 @@ def create_from_cloudformation_json(
...
+                for redirect_config_key, redirect_config_value in action[key].items():
+                    # need to match the output of _get_list_prefix
+                    redirect_action[
+                        camelcase_to_underscores(key)
+                        + "._"
+                        + camelcase_to_underscores(redirect_config_key)
+                    ] = redirect_config_value
+                default_actions.append(redirect_action)
+            else:
+                raise InvalidActionTypeError(action_type, i + 1)
```
bblommers, line 348: This edge case should have a unit test as well

bblommers (changes requested):

moto/elbv2/models.py
sahilshah6196: I updated it to be an `OrderedDict` instead of a `list` like other "FakeServices". If I do not reverse it then the order of the items inserted is changed. The order in which the items are tested against will changer here:
https://github.com/spulec/moto/blob/master/tests/test_elbv2/test_elbv2.py#L1113-L1115 i.e. `[0]` element will become he `[2]` element ec.

moto/elbv2/models.py
sahilshah6196: I did try and then reverted it: https://github.com/spulec/moto/pull/3969/commits/704a5ab890836d6588dd652d58475c2d0d08bd6f
I do not remember why. Will try it again

moto/elbv2/models.py
sahilshah6196: I already see that here: https://github.com/spulec/moto/blob/master/tests/test_elbv2/test_elbv2.py#L1158
Is that not correct? or did I misunderstand something?

moto/elbv2/models.py
sahilshah6196: Oh yes. The `Action` works a little differently in `ListernRule`. My intention was to fix that in the next PR. 
Here is how: There is a [`ForwardConfig` option in the `ListenerRule`](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-elasticloadbalancingv2-listenerrule-action.html#cfn-elasticloadbalancingv2-listenerrule-action-forwardconfig) that in-turn defines the `TargetGroups` and the `TargetGroupStickinessConfig`. That needs to be added in the check. So really this will need to change when adding proper validations.
What do you think?

sahilshah6196: Example:
```
LoadBalancerListenerRule:
    Type: AWS::ElasticLoadBalancingV2::ListenerRule
    Properties:
      Actions:
        - Type: forward
          ForwardConfig:
            TargetGroups:
              - TargetGroupArn:
                  {{target_group.arn}}
                Weight: {{target_group.weight}}
            TargetGroupStickinessConfig:
              DurationSeconds: 300
              Enabled:
                Fn::If: [ IsStickinessEnabled, true, false ]
      Conditions:
          - Field: path-pattern
            Values:
              - '/*'
      ListenerArn: 
        Ref: LoadBalancerListener
      Priority: 1
```

This, right now, fails with the error:
KeyError: 'TargetGroupArn'

moto/elbv2/models.py
sahilshah6196: Oh yes. The Action works a little differently in ListernRule. My intention was to fix that in the next PR.
Here is how: There is a ForwardConfig option in the ListenerRule that in-turn defines the TargetGroups and the TargetGroupStickinessConfig. That needs to be added in the check. So really this will need to change when adding proper validations.
What do you think?

sahilshah6196: Example:
```
LoadBalancerListenerRule:
    Type: AWS::ElasticLoadBalancingV2::ListenerRule
    Properties:
      Actions:
        - Type: forward
          ForwardConfig:
            TargetGroups:
              - TargetGroupArn:
                  {{target_group.arn}}
                Weight: {{target_group.weight}}
            TargetGroupStickinessConfig:
              DurationSeconds: 300
              Enabled:
                Fn::If: [ IsStickinessEnabled, true, false ]
      Conditions:
          - Field: path-pattern
            Values:
              - '/*'
      ListenerArn: 
        Ref: LoadBalancerListener
      Priority: 1
```
This, right now, fails with the error:
KeyError: 'TargetGroupArn'

moto/elbv2/models.py
bblommers: Looks like it's the same error scenario, but it also needs to be tested when creating the rule using CloudFormation

moto/elbv2/models.py
bblommers: I understand ( ..I think). If future behaviour will create a difference between them, there's no point in making it DRY now.

moto/elbv2/models.py
bblommers: Sorry, I wasn't very clear. Why the change from sorting it by priority?
I don't have any real world experience with these methods, so not sure what AWS does. But from the docs I can see that the CLI does show them in priority order

moto/elbv2/models.py
sahilshah6196: You are correct. My bad. Fixed: ffd63ba

sahilshah6196:
Hey @bblommers 
I fixed/responded to all the comments. Thanks for starting the review

moto/elbv2/models.py
sahilshah6196: Fixed: eb95069
Also changed the priority order from 50 to 500 to show that lowest priority comes first

moto/elbv2/models.py
sahilshah6196: @bblommers Like this? 79268a9

bblommers (approved):
LGTM. Thanks again for the PR and making all the changes @sahilshah6196!

sahilshah6196:
> LGTM. Thanks again for the PR and making all the changes @sahilshah6196!

Will work on a new PR for the remaining validations

Issue #3971, opened 2021-05-31
New Service: wafv2


v-dev:
Hello! I would like to add support for a few wafv2 APIs. Who do I talk to about this?

So far, I ran `scripts/scaffold.py` and had success with a basic mock of `list-web-acls`. Wondering if I should now open a PR, or include more models and/or APIs such as `get-web-acl` and other CRUD operations?

My end goal is to have these used within [localstack](https://github.com/localstack/localstack).

Thank you.

bblommers:
Hi @v-dev, welcome to Moto!

I'd recommend to open a PR now, just to get some early feedback. You can mark it as a draft and keep working on it in the background, if you want to add other models/operations etc.

PR #3973, opened 2021-05-31
WAFV2 : list-web-acls

v-dev:
**Background:**
Related to #3971

Begin adding support for the wafv2 services, starting with the `list-web-acls` API. Currently just returns 2 preloaded mock WebACLs.

First PR in this project. Appreciate the guidance.

**Sample Output:**
```
$ aws --endpoint-url=http://localhost:5000 wafv2 list-web-acls --scope REGIONAL
{
    "WebACLs": [
        {
            "Name": "first-mock-webacl-ad54c7ad-8a89-450f-9673-9c1571c6acc3",
            "Id": "edcf8c35-3615-4b38-a57f-96efb0f67945",
            "Description": "Mock WebACL named first-mock-webacl-ad54c7ad-8a89-450f-9673-9c1571c6acc3",
            "ARN": "arn:aws:wafv2:us-east-1:123456789012:regional/webacl/first-mock-webacl-ad54c7ad-8a89-450f-9673-9c1571c6acc3/fecf928a-4ae5-404e-9791-91ff0d299ffe"
        },
        {
            "Name": "second-mock-webacl-bfbb5048-1037-4fd5-bebc-f1e00f6c1717",
            "Id": "e02e0290-5d37-43d1-a294-7e1339d0530d",
            "Description": "Mock WebACL named second-mock-webacl-bfbb5048-1037-4fd5-bebc-f1e00f6c1717",
            "ARN": "arn:aws:wafv2:us-east-1:123456789012:regional/webacl/second-mock-webacl-bfbb5048-1037-4fd5-bebc-f1e00f6c1717/5f79a574-a240-4a78-9fad-f14a263c6971"
        }
    ]
}
```


moto/wafv2/models.py
```diff
@@ -0,0 +1,66 @@
...
+        """
+        The name must have 1-128 characters. Valid characters: A-Z, a-z, 0-9, - (hyphen), and _ (underscore).
+        """
+        self.Name = "{0}-{1}".format(name, uuid.uuid4())[0:128]
+        self.ARN = make_webacl_arn(US_EAST_1_REGION, self.Name)
+        self.Capacity = 3
+        self.Description = "Mock WebACL named {0}".format(self.Name)
+        self.VisibilityConfig = VisibilityConfig(self.Name)
+        """
+        TODO: create classes for the below fields:
```
v-dev, line 40: There are many [other fields in a WebACL](https://docs.aws.amazon.com/waf/latest/APIReference/API_WebACL.html), and those, in turn, often have many other fields, too. Should I put them all in this `models.py` file, or should I create a `models` package with the other classes?

bblommers: It's fine to have them all in the same file.
For most services, if there's a model with many (nested) fields, we tend to just store them as a dictionary. Depending on how many of the fields are actually used, or whether they're simply stored for retrieval later

moto/wafv2/responses.py
```diff
@@ -0,0 +1,23 @@
...
+
+class WAFV2Response(BaseResponse):
+    SERVICE_NAME = "wafv2"
+
+    @property
+    def wafv2_backend(self):
+        return wafv2_backends[self.region]
+
+
+class WafV2Handler(BaseResponse):
```
bblommers, line 17: These two classes can be combined (`WAFV2Response`/`WavV2Handler`)

ubuntu_init_dev_setup.sh
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/bash
+
+# from https://tutorials.technology/solved_errors/10-fatal-error-libxml-xpath_h-No-such-file-or-directory.html
+sudo apt-get install libxml2-dev libxslt1-dev python-dev
```
bblommers, line 4: This shouldn't be in an separate file. Were you getting this error when running `make init`?

moto/wafv2/models.py
```diff
@@ -0,0 +1,66 @@
...
+    https://docs.aws.amazon.com/waf/latest/APIReference/API_VisibilityConfig.html
+    """
+
+    def __init__(self, metric_name):
+        self.SampledRequestsEnabled = True
+        self.CloudWatchMetricsEnabled = False
+        self.MetricName = metric_name
+
+
+wafv2_backends = {"global": WAFV2Backend(GLOBAL_REGION)}
```
bblommers, line 66: WAF is interesting, in that it combines both regional and global resources. As far as I know, it's the first service in Moto that does it.
The global/cloudfront resources are still persisted in `us-east-``, though. So the backends can still be created as every other service, having a separate backend per region. We'll just have to check for every call whether the scope was set to CLOUDFRONT or REGIONAL.

https://docs.aws.amazon.com/cli/latest/reference/wafv2/list-web-acls.html
>
> To work with CloudFront, you must also specify the Region US East (N. Virginia)

bblommers: The `get_available_regions` command does not return anything for `wafv2` at the moment, so let's use `waf-regional` for now. The list should be the same anyway.
```
wafv2_backends = {}
for region in Session().get_available_regions("waf-regional"):
    wafv2_backends[region] = WAFV2Backend(region)
for region in Session().get_available_regions("waf-regional", partition_name="aws-us-gov"):
    wafv2_backends[region] = WAFV2Backend(region)
for region in Session().get_available_regions("waf-regional", partition_name="aws-cn"):
    wafv2_backends[region] = WAFV2Backend(region)
```

moto/wafv2/utils.py
```diff
@@ -0,0 +1,14 @@
+import uuid
+from collections import namedtuple
+
+ARN = namedtuple("ARN", ["region", "account", "function_name", "version"])
+
+
+def make_arn(service, region, name):
+    return "arn:aws:wafv2:{0}:123456789012:regional/{1}/{2}/{3}".format(
```
bblommers, line 8: The account ID is stored centrally:
`from moto.core import ACCOUNT_ID`

moto/wafv2/responses.py
```diff
@@ -0,0 +1,23 @@
...
+    def list_web_ac_ls(self):
+        wacl = wafv2_backends[GLOBAL_REGION]
```
bblommers, line 19: `self.body` returns `{"Scope": "CLOUDFRONT"}` or  `{"Scope": "REGIONAL"}`, so we can use that to determine which backend to talk to.

moto/wafv2/models.py
bblommers: Attribute-names should be pythonic btw (snake_case, so `self.name` and `self.visibility_config`)

v-dev:
Thank you for all the feedback! I will incorporate and/or come back with more questions as I progress... which is more likely during the weekends.

ubuntu_init_dev_setup.sh
v-dev: It has been a while, but: I believe so, yes.

moto/wafv2/responses.py
v-dev: Combined into `WAFV2Response`

moto/wafv2/utils.py
v-dev: Good to know - updated... thanks!

moto/wafv2/models.py
v-dev: Thank you. Updated.

moto/wafv2/responses.py
v-dev: Used `self.body` and added/updated tests.

bblommers:
Closing in favour of #4076

PR #3975, opened 2021-06-01
Add new resource functionalities

usmangani1:
Add new resource functionalities

### DIFF
--- a/moto/resourcegroups/models.py
+++ b/moto/resourcegroups/models.py
@@ -13,7 +13,13 @@ from .exceptions import BadRequestException
 
 class FakeResourceGroup(BaseModel):
     def __init__(
-        self, name, resource_query, description=None, tags=None, configuration=None
+        self,
+        name,
+        resource_query,
+        description=None,
+        tags=None,
+        configuration=None,
+        resource_arns=[],
     ):
         self.errors = []
         description = description or ""
@@ -31,6 +37,7 @@ class FakeResourceGroup(BaseModel):
             name=name, AccountId=ACCOUNT_ID
         )
         self.configuration = configuration
+        self.resource_arns = resource_arns
 
     @staticmethod
     def _format_error(key, value, constraint):
@@ -221,6 +228,14 @@ class ResourceGroups:
         return group
 
 
+def get_resource_type(arn):
+    if "instance/my-ec2" in arn:
+        return "AWS::EC2::Instance"
+    elif "db:mysql-db" in arn:
+        return "AWS::RDS::DBInstance"
+    return ""
+
+
 class ResourceGroupsBackend(BaseBackend):
     def __init__(self, region_name=None):
         super(ResourceGroupsBackend, self).__init__()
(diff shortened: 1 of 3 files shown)


moto/resourcegroups/models.py
```diff
@@ -221,6 +228,14 @@ def delete(self, name):
         return group
 
 
+def get_resource_type(arn):
+    if "instance/my-ec2" in arn:
```
bblommers, line 232: This only matches a specific test case, it should regex the full ARN:
`arn:aws:ec2:<region>:<accountid>:instance/<id>`

moto/resourcegroups/models.py
```diff
@@ -221,6 +228,14 @@ def delete(self, name):
...
+        return "AWS::EC2::Instance"
+    elif "db:mysql-db" in arn:
```
bblommers, line 234: Same here

moto/resourcegroups/models.py
```diff
@@ -221,6 +228,14 @@ def delete(self, name):
...
+        return "AWS::RDS::DBInstance"
```
bblommers, line 235: And we should have a unit test for RDS as well

moto/resourcegroups/models.py
```diff
@@ -368,6 +383,37 @@ def put_group_configuration(self, group_name, configuration):
...
+    def ungroup_resources(self, group_name, resource_arns):
+        actual_resource_arns = self.groups.by_name[group_name].resource_arns
+        for resource_arn in resource_arns:
+            actual_resource_arns.remove(resource_arn)
+        self.groups.by_name[group_name].resource_arns = actual_resource_arns
+
+    def list_group_resources(self, group_name):
+        return self.groups.by_name[group_name].resource_arns
+
+    def search_resources(self, resource_query):
```
bblommers, line 398: This method should have a few tests as well, covering all possible use cases (no matches, 1 resource-type match, 1 tag match, and combination of the above)

tests/test_resourcegroups/test_resourcegroups.py
```diff
@@ -274,3 +274,56 @@ def test_update_group_query():
...
+        Group=group_response.get("Group"),
+        ResourceArns=["arn:aws:ec2:us-west-2:123456789012:instance/my-ec2-instance1"],
+    )
+    assert result.get("Succeeded") == [
+        "arn:aws:ec2:us-west-2:123456789012:instance/my-ec2-instance1"
+    ]
+
+
+@mock_resourcegroups
+def test_list_group_resources():
```
bblommers, line 310: We should have an additional test case calling `list_group_resources` on a group without any resources

tests/test_resourcegroups/test_resourcegroups.py
```diff
@@ -274,3 +274,56 @@ def test_update_group_query():
...
+        Group=group_response.get("Group"),
+        ResourceArns=["arn:aws:ec2:us-west-2:123456789012:instance/my-ec2-instance1"],
+    )
+    assert result.get("Succeeded") == [
+        "arn:aws:ec2:us-west-2:123456789012:instance/my-ec2-instance1"
+    ]
+
+
+@mock_resourcegroups
+def test_ungroup_resources():
```
bblommers, line 295: This doesn't test anything at the moment. The flow should be:
 - group multiple resources
 - list_group -> returns all data
 - ungroup_resources (some resources only)
 - list_group -> verify only some resources were dropped, but it still returns the others

bblommers (changes requested):

PR #3976, opened 2021-06-01
Fix:Add list connection API's

usmangani1:
Added connection API's

https://github.com/localstack/localstack/issues/4065

### DIFF
--- a/moto/events/responses.py
+++ b/moto/events/responses.py
@@ -409,3 +409,32 @@ class EventsHandler(BaseResponse):
         result = self.events_backend.cancel_replay(name)
 
         return json.dumps(result), self.response_headers
+
+    def create_connection(self):
+        name = self._get_param("Name")
+        description = self._get_param("Description")
+        authorization_type = self._get_param("AuthorizationType")
+        auth_parameters = self._get_param("AuthParameters")
+
+        if authorization_type != "API_KEY":
+            raise NotImplementedError(
+                "EventsBridge with authorization_type {0} is not yet implemented".format(
+                    authorization_type
+                )
+            )
+
+        result = self.events_backend.create_connection(
+            name, description, authorization_type, auth_parameters
+        )
+
+        return (
+            json.dumps(
+                {
+                    "ConnectionArn": result,
+                    "ConnectionState": "CREATING",
+                    "CreationTime": result.creation_time,
+                    "LastModifiedTime": result.creation_time,
+                }
+            ),
+            self.response_headers,
+        )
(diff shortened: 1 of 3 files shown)


moto/events/responses.py
```diff
@@ -409,3 +409,48 @@ def cancel_replay(self):
...
+        )
+
+    def list_connections(self):
+        connections = self.events_backend.list_connections()
+        result = []
+        for connection in connections:
+            result.append(
+                {
+                    "ConnectionArn": connection.arn,
+                    "ConnectionState": "CREATING",
```
bblommers, line 449: What's the end-state for connections? Looking at the docs, I'm assuming the happy flow is: CREATING -> AUTHORIZING -> AUTHORIZED, so we should probably return 'AUTHORIZED' here

moto/events/responses.py
```diff
@@ -409,3 +409,48 @@ def cancel_replay(self):
...
         return json.dumps(result), self.response_headers
+
+    def create_connection(self):
+        name = self._get_param("Name")
+        description = self._get_param("Description")
+        authorization_type = self._get_param("AuthorizationType")
+        auth_parameters = self._get_param("AuthParameters")
+
+        if authorization_type != "API_KEY":
+            raise NotImplementedError(
```
bblommers, line 420: Do we need this check? It doesn't look like we're doing anything with this data anyway, we just store it and return it later.
So which authorization-type is passed in doesn't really matter

tests/test_events/test_events.py
```diff
@@ -2138,3 +2138,28 @@ def test_start_replay_send_to_log_group():
...
+        Description="test description",
+        AuthorizationType="API_KEY",
+        AuthParameters={
+            "ApiKeyAuthParameters": {"ApiKeyName": "test", "ApiKeyValue": "test"}
+        },
+    )
+
+    assert (
+        response.get("ConnectionArn")
+        == "arn:aws:events:eu-central-1:000000000000:connection/test"
```
bblommers, line 2157: Moto's account ID is different:
`from moto.core import ACCOUNT_ID`

tests/test_events/test_events.py
```diff
@@ -2138,3 +2138,28 @@ def test_start_replay_send_to_log_group():
     event_replay["resources"].should.be.empty
     event_replay["detail"].should.equal({"key": "value"})
     event_replay["replay-name"].should.equal("test-replay")
+
+
+@mock_events
+def test_create_and_list_connections():
```
bblommers, line 2144: We should have a separate list_connections test to verify it works without creating any connections first

bblommers (changes requested):

bblommers (approved):
LGTM - thanks @usmangani1!

PR #3977, opened 2021-06-01
SSM Parameter Store Error Message When Requesting Invalid Version

bigahen:
SSM Parameter Store returns a ParameterVersionNotFound error code when requesting a parameter with version where the parameter exists but the requested version does not. 

Provided example below with actual AWS responses. 

image


bblommers (approved):
LGTM. Thanks for the PR @bigahen!

Issue #3870, opened 2021-04-21
Adding EKS Support?


ferruzzi:
Apologies, I didn't see a link to a community mailing list/chat/etc.  I have a usecase where I need to mock a portion of EKS and it doesn't look like that is implemented yet.  Just wanted to make sure there wasn't someone working on that before I go and duplicate work.  If it's not being worked on, I'll look into adding the service.

bblommers:
Hi @ferruzzi, I'm not aware of anyone working on that service, so a PR would be welcome.

Some notes on how to add new services can be found here: https://github.com/spulec/moto/blob/master/CONTRIBUTING.md#missing-services

Let me know if you need any help.

ferruzzi:
Sounds good.  I am going to need it to test a contribution I am adding to the Apache Airflow project.  Assuming the code is to your standards and all that, what's a general approval time on a well-executed PR?

bblommers:
No promises here, but approval tends to happen within 24h. That only creates a dev- release though - there's no fixed timeline on when a full release will be cut.

ferruzzi:
Fair enough.  We would need the code to be publicly available quickly after approval (1-2 weeks).  If the turnaround is going to be longer then we'll have to implement the mocking within the Airflow test package.  What dictates when the next full release gets cut?  Anything I can help with there?  I'd really prefer this was mutually beneficial; if I'm already doing the work to mock the service, it may as well be useful.  Thanks for your time!

bblommers:
Release times really depend on how much has been merged (i.e., whether it's worth it) and maintainer time.
1-2 weeks shouldn't be a problem though. Just remind me after approval, and I'll get on it 👍

ferruzzi:
That sounds great, can definitely work with that.

bblommers:
A single PR is fine, @ferruzzi! Easier to create and easier to review

ferruzzi:
Hi, I'm plugging away at this.  I have the four endpoints done for clusters, working on nodegroups now.  I've run into a snag I can't seem to figure out and could use a hand.  When implementing `list_nodegroups()`, I ran the scaffold.py script to generate the bones as I did previously and have added the following:

in responses.py:
```
    def list_nodegroups(self):
        cluster_name = self._get_param("clusterName")
        max_results = self._get_int_param("maxResults", DEFAULT_MAX_RESULTS)
        next_token = self._get_param("nextToken", DEFAULT_NEXT_TOKEN)
[...]

PR #3981, opened 2021-06-02
Initial EKS Implementaion

ferruzzi:
Added the following EKS endpoints with testing:
- [Create Cluster](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/eks.html#EKS.Client.create_cluster)
- [Create Nodegroup](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/eks.html#EKS.Client.create_nodegroup)
- [Delete Cluster](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/eks.html#EKS.Client.delete_cluster)
- [Delete Nodegroup](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/eks.html#EKS.Client.delete_nodegroup)
- [Describe Cluster](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/eks.html#EKS.Client.describe_cluster)
- [Describe Nodegroup](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/eks.html#EKS.Client.describe_nodegroup)
- [List Clusters](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/eks.html#EKS.Client.list_clusters)
- [List Nodegroups](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/eks.html#EKS.Client.list_nodegroups)

https://github.com/spulec/moto/issues/3870


ferruzzi:
I see the failing tests; looking into it.

ferruzzi:
Hey @bblommers, sorry to bother.   Looks like the 2.7 test is failing because the test_constants file it not found, but it's definitely there in the same directory/module.  What am I missing?

It also looks like all of the 3.* ones are failing due to S3 tests which I didn't touch and aren't included in my "files changed" tab so that may be something to look into?

tests/test_eks/test_eks.py
```diff
@@ -0,0 +1,770 @@
...
+    CLUSTER_NOT_READY_MSG,
+    CLUSTER_NOT_FOUND_MSG,
+    CLUSTER_IN_USE_MSG,
+    LAUNCH_TEMPLATE_WITH_DISK_SIZE_MSG,
+    LAUNCH_TEMPLATE_WITH_REMOTE_ACCESS_MSG,
+    NODEGROUP_NOT_FOUND_MSG,
+    CLUSTER_EXISTS_MSG,
+)
+from moto.utilities.utils import random_string
+from test_eks_constants import (
```
bblommers, line 28: `tests/test_eks/__init__ ` should be renamed to `tests/test_eks/__init__.py` in order for pytest to play nice.
After that, the imports should be relative, with a dot preceding the files: `from .test_eks_constants import..`, `from .test_eks_utils import ..`

moto/eks/responses.py
```diff
@@ -0,0 +1,124 @@
...
+            role_arn=role_arn,
+            resources_vpc_config=resources_vpc_config,
+            kubernetes_network_config=kubernetes_network_config,
+            logging=logging,
+            client_request_token=client_request_token,
+            tags=tags,
+            encryption_config=encryption_config,
+        )
+
+        return 200, {}, json.dumps({"cluster": {**dict(cluster)}})
```
bblommers, line 40: The `**dict` notation doesn't work in Py2.7, as far as I know.
Other services will typically use a `model.to_dict()` method  to control the response format

moto/eks/responses.py
```diff
@@ -0,0 +1,124 @@
...
+            labels=labels,
+            tags=tags,
+            client_request_token=client_request_token,
+            launch_template=launch_template,
+            capacity_type=capacity_type,
+            version=version,
+            release_version=release_version,
+        )
+
+        return 200, {}, json.dumps({"nodegroup": {**dict(nodegroup)}})
```
bblommers, line 79: Same here

bblommers:
Hey @ferruzzi, no problem. See my comments - that should get it working in Py2.

Not sure about the S3-tests, they all pass on my end. Let's see what happens with the init-file renamed 🤞

moto/eks/models.py
```diff
@@ -0,0 +1,467 @@
...
+)
+NODEGROUP_NOT_FOUND_MSG = BASE_MSG + "No node group found for name: {nodegroup_name}."
+
+
+class Cluster:
+    def __init__(
+        self,
+        name,
+        roleArn,
+        resourcesVpcConfig,
```
bblommers, line 90: Variables should be all snake_case

moto/eks/models.py
```diff
@@ -0,0 +1,467 @@
...
+
+
+class ManagedNodegroup:
+    def __init__(
+        self,
+        cluster_name,
+        node_role,
+        nodegroup_name,
+        subnets,
+        regionName,
```
bblommers, line 160: Same here

moto/eks/models.py
```diff
@@ -0,0 +1,467 @@
...
+        yield "logging", self.logging
+        yield "identity", self.identity
+        yield "status", self.status
+        yield "certificateAuthority", self.certificateAuthority
+        yield "clientRequestToken", self.client_request_token
+        yield "platformVersion", self.platformVersion
+        yield "tags", self.tags
+        yield "encryptionConfig", self.encryption_config
+
+    def isActive(self):
```
bblommers, line 149: Same here

moto/eks/utils.py
```diff
@@ -0,0 +1,48 @@
+import inspect
+import re
+
+from boto3 import Session
+
+from moto.eks.exceptions import InvalidParameterException
+
+
+def set_partition(region):
```
bblommers, line 9: This should be named `get_partition`

tests/test_eks/test_eks.py
```diff
@@ -0,0 +1,770 @@
...
+            # Generate a list of the Cluster attributes to be tested when validating results.
+            self.attributes_to_test = attributes_to_test(
+                ClusterInputs, self.existing_cluster_name
+            )
+
+    def _execute(count=1, minimal=True):
+        return client, ClusterTestDataFactory(count, minimal)
+
+    mock_eks().start()
+    client = boto3.client(SERVICE)
```
bblommers, line 91: Can we specify a region name here? Don't really have a preference on which region, as long as it runs on machines without a default. Same for the other 2 invocations.

`boto3.client(SERVICE, region_name=DEFAULT_REGION)`

tests/test_eks/test_eks.py
```diff
@@ -0,0 +1,770 @@
...
+    all_arn_values_should_be_valid(
+        expected_arn_values=expected_arn_values,
+        pattern=RegExTemplates.NODEGROUP_ARN,
+        arn_under_test=generated_test_data.nodegroup_describe_output[
+            NodegroupAttributes.ARN
+        ],
+    )
+
+
+@pytest.mark.freeze_time
```
bblommers, line 506: Moto uses the `freezegun` package for this effect:
```
from freezegun import freeze_time
@freeze_time("2014-01-01 05:00:00")
```

tests/test_eks/test_eks.py
```diff
@@ -0,0 +1,770 @@
...
+def test_create_nodegroup_generates_valid_nodegroup_created_timestamp(NodegroupBuilder):
+    _, generated_test_data = NodegroupBuilder()
+    current_time = iso_8601_datetime_without_milliseconds(datetime.now())
+
+    result_time = iso_8601_datetime_without_milliseconds(
+        generated_test_data.nodegroup_describe_output[NodegroupAttributes.CREATED_AT]
+    )
+
+    result_time.should.equal(current_time)
```
bblommers, line 515: Note that freezing time will not work in ServerMode, as that will start Moto in a completely separated Docker container.
You can use settings.TEST_SERVER_MODE to guard the assertions, and maybe just check that the attribute exists/has the correct format in server mode. (Example: https://github.com/spulec/moto/blob/master/tests/test_sts/test_sts.py#L85

tests/test_eks/test_eks.py
```diff
@@ -0,0 +1,770 @@
...
+def test_describe_nodegroup_throws_exception_when_cluster_not_found(NodegroupBuilder):
+    client, generated_test_data = NodegroupBuilder()
+    expected_exception = ResourceNotFoundException
+    expected_msg = CLUSTER_NOT_FOUND_MSG.format(
+        exception_name=expected_exception.TYPE,
+        method=MethodNames.DESCRIBE_NODEGROUP,
+        cluster_name=generated_test_data.nonexistent_cluster_name,
+    )
+
+    client.describe_nodegroup.when.called_with(
```
bblommers, line 574: Exceptions should be tested using the `with pytest.raises` paradigm, to verify the format is correct. 
Example: https://github.com/spulec/moto/blob/master/tests/test_sts/test_sts.py#L738

bblommers: Same for the other 11 `when.called_with` calls

tests/test_eks/test_eks.py
```diff
@@ -0,0 +1,770 @@
...
+    )
+
+    client.delete_nodegroup.when.called_with(
+        clusterName=generated_test_data.cluster_name,
+        nodegroupName=generated_test_data.nonexistent_nodegroup_name,
+    ).should.throw(expected_exception, expected_msg)
+
+
+# If launch_template is specified, you can not specify instanceTypes, diskSize, or remoteAccess.
+test_cases = [
```
bblommers, line 670: Nice - that's some very thorough testing 👍

bblommers (changes requested):

tests/test_eks/test_eks.py
ferruzzi: ACK, thanks.

moto/eks/models.py
ferruzzi: These need to stay camelCase, that is what the API accepts.  I had left the above two camelCase to align with that, but have changed them now.

moto/eks/responses.py
ferruzzi: I get your comment about the **dict() usage in the responses file, and implemented a to_dict() in the Cluster and ManagedNodegroup models.  I've set up a venv with 2.7 top make sure I catch that stuff in the future.

ferruzzi:
Alright, sorry for the delay.  I think I addressed all of your comments.  If I missed something, please let me know.   I ran the tests under 2.7 and under 3.8 this time to make sure I have the coverage you want.  Sorry for missing that before.

tests/test_eks/test_eks.py
```diff
@@ -746,7 +746,11 @@ def test_create_nodegroup_handles_launch_template_combinations(
                 (NodegroupAttributes.NODEGROUP_NAME, nodegroup_name),
             ]
             # Test Case Values
-            + [_ for _ in [launch_template, instance_types, disk_size, remote_access] if _]
+            + [
+                _
+                for _ in [launch_template, instance_types, disk_size, remote_access]
+                if _
+            ]
```
ferruzzi, line 753: Thanks, not sure how I missed linting.

bblommers:
Hi @ferruzzi, I made some changes to get the server mode tests to run. Looks like the majority of the failures were due to missing URL paths.

Some negative tests are still going to fail (in server mode), because errors are not returned in the expected format. 
Are you able to run a negative test case against AWS itself, for example the test when creating a cluster with a name that already exists, and intercept the HTTP response?

(In order to mimick the expected response from AWS, we typically have to format it similar to this: https://github.com/spulec/moto/blob/master/moto/dynamodb/responses.py#L111, but the exact response tends to change between services...)

ferruzzi:
Thanks for the help on that, and sorry about the linting on that last one.  I'll have a look at the error messages today and see what I can do.

ferruzzi:
[Log dive]  Current issue is this for (most/all) tests:

```
         mock_eks().start()
        client = boto3.client(SERVICE, region_name=REGION)
        yield _execute
>       mock_eks().stop()

tests/test_eks/test_eks.py:94: 
moto/core/models.py:105: in stop
    self.disable_patching()

self = <moto.core.models.ServerModeMockAWS object at 0x7f1858e09910>

    def disable_patching(self):
>       if self._client_patcher:
E       AttributeError: 'ServerModeMockAWS' object has no attribute '_client_patcher'
```

I have a ton of meetings all day today, I may not get to look into it or the error message you mentioned until tomorrow.

bblommers:
Thanks for the curl-response @ferruzzi - that's exactly what I had in mind.

> Is the responses.py file only used in server mode?
No, it's used for everything

> Is there some documentation on what works differently in server mode

We should have some documentation on this, yes. One day. :) 
In the meantime, in a nutshell:
There are three types of tests that verify different behaviours of moto.

1) Unit tests: Verify that the decorator works
2) ServerMode tests: Verify that Moto can act as a standalone server, so that users can point any SDK to it (Python, Java, .NET, etc.)
3) test_server.py: Verify that the build-in Flask server returns the expected raw data, what we would expect to see when using the HTTP API

The test_server tests has the advantage that we have complete control over the request and response, so we can verify scenarios that we can't test with just the boto3-client.

I'm about to push two tests for test_server.py with the correct assertions, and an incomplete implementation to return the correct response for these tests.

tests/test_eks/test_eks.py
```diff
@@ -0,0 +1,811 @@
...
+def test_create_cluster_throws_exception_when_cluster_exists(ClusterBuilder):
+    client, generated_test_data = ClusterBuilder(BatchCountSize.SMALL)
+    expected_exception = ResourceInUseException
+    expected_msg = CLUSTER_EXISTS_MSG.format(
+        exception_name=expected_exception.TYPE,
+        method=MethodNames.CREATE_CLUSTER,
+        cluster_name=generated_test_data.existing_cluster_name,
+    )
+
+    with pytest.raises(expected_exception) as raised_exception:
```
bblommers, line 216: Apologies for not seeing this earlier - this happened to work before, because the exception handling wasn't implemented yet, but it should throw an error now that proper exception handling is in place.

Moto (like AWS) responds with a HTTP 4xx-error in this scenario, which boto3 then converts into a generic ClientError.
Any tests we run should therefore always verify that that specific ClientError is thrown.

Example:
```
    ... something that fails
er["Message"].should.equal("..")
er["Code"].should.equal("..")
```

This is how it works against AWS, and is also the way to verify that we respond with a properly formatted HTTP 4xx-error that boto3 can understand.

When using this specific format, any test that passes the normal unit tests will also pass the ServerMode tests.

You might have to rewrite a few tests - I noticed there were a few more that follow the incorrect format.

bblommers (changes requested):

tests/test_eks/test_eks.py
ferruzzi: I see!   Alright, sounds good.  Between the commit you added above and this, I have a bit more work to do.  Apologies for all the revisions on this one; I'll get it there...

bblommers: This will add a **lot** of features, so some back-and-forth is expected :) 

The error handling is on us, really - there should be a standardized way of doing this. But with a 6+ year old 'organically grown' code base, and plenty of variations on what kind of output AWS/boto3 expects per service, it's not exactly easy to understand.

ferruzzi: Yeah, the fact that the output format of the AWS services vary so much doesn't make this any easier.

I'll get on adding the rest of those tests next week.  To make sure I understand what you have there and I'm not missing something: it looks like the error message format you were looking for happened to be just the exception name in this case, and not something like `"com.amazonaws.dynamodb.v20111205#ResourceNotFoundException"`, or is checking the header like that (test_server.py:L63) a shortcut/workaround?

> 
> Edit:
> The outstanding tasks would be:
> Ensure that the test_server tests pass, so that we know the error response is as expected
> Fix the negative tests per my comment below, so that all unit tests pass
> The ServerMode tests should then pass automagically as well

Thanks for this.

bblommers: > it looks like the error message format you were looking for happened to be just the exception name in this case, and not something like "com.amazonaws.dynamodb.v20111205#ResourceNotFoundException"

Yes, just the exception name. Dynamo is an exception in how it's named(of course..), because of the deprecated API version it refers to.

The header-check was added to ensure Moto's response is exactly the same as your Postman/curl-response. If AWS returns the exception name, so should we

ferruzzi:
Alright.   I think what I am seeing in your commit there is that on `exceptions.py:L13` the `self.description` should represent the expected curl result.  In which case I should add the nodegroup name, addon name, and the short version of the message to that `dumps`, as verified in `test_server.py:L68`.  Is that correct, or is the message being added somewhere behind the scenes that I am missing?

tests/test_eks/test_eks.py
ferruzzi: Alright. I think what I am seeing in your commit there is that on `exceptions.py:L13` the `self.description` should represent the expected curl result. In which case I should add the nodegroup name, addon name, and the short version of the message to that `dumps`, as verified in `test_server.py:L68`. Is that correct, or is the message being added somewhere behind the scenes that I am missing?

bblommers: Yes, that is correct

ferruzzi:
Hey @bblommers .   I've implemented the Server Tests, tested under py2.7 and 3.9, and squashed a few of the latter commits to keep your log a bit cleaner.

ferruzzi:
Looks like there's an issue with one of the tests in server mode:

```
=================================== FAILURES ===================================
________ test_create_nodegroup_throws_exception_when_cluster_not_active ________

NodegroupBuilder = <function NodegroupBuilder.<locals>._execute at 0x7f46e0a448c8>
monkeypatch = <_pytest.monkeypatch.MonkeyPatch object at 0x7f46f8446588>

    def test_create_nodegroup_throws_exception_when_cluster_not_active(
        NodegroupBuilder, monkeypatch
    ):
        client, generated_test_data = NodegroupBuilder(BatchCountSize.SMALL)
        expected_exception = InvalidRequestException
        expected_msg = CLUSTER_NOT_READY_MSG.format(
            clusterName=generated_test_data.cluster_name,
        )
    
        with mock.patch("moto.eks.models.Cluster.isActive", return_value=False):
            with pytest.raises(ClientError) as raised_exception:
                client.create_nodegroup(
                    clusterName=generated_test_data.cluster_name,
                    nodegroupName=random_string(),
>                   **dict(NodegroupInputs.REQUIRED)
                )
E               Failed: DID NOT RAISE <class 'botocore.exceptions.ClientError'>

tests/test_eks/test_eks.py:474: Failed
```

Test lives in [/test_eks.py](https://github.com/spulec/moto/pull/3981/files#diff-d51af7b3831f39694a6bae7f0239b82dc6894e6b4125b72157eb120d2c4fad33):L460

bblommers:
In ServerMode, moto runs in a separate Docker container, so it's not possible to patch the `moto.eks.models.Cluster.isActive`-attribute. 
Probably easiest to only run this test in normal mode. (Example on how we skip tests, just with a different text: https://github.com/spulec/moto/blob/master/tests/test_acm/test_acm.py#L544)

ferruzzi:
Added the "skip if in server mode" check to three tests which manipulate time and one that uses a the patch mentioned above.

bblommers:
Great success! 
The ServerMode tests didn't play nice with the mocking inside the fixtures - not sure why. Mocking everything using decorators works fine though, that's why I changed them all

bblommers (approved):
Thanks for the PR @ferruzzi! Much appreciated

ferruzzi:
Nice!   Thanks!    As mentioned at the beginning of this journey, I'm going to need a that in a new release when you get the time so I can ingest it into an Airflow contribution to add EKS support, which is waiting for it.

bblommers:
Already taken care of @ferruzzi - 2.0.10 was released a couple of hours ago :)

ferruzzi:
Amazing.  Thanks!

PR #3982, opened 2021-06-03
Add Tag support for other resource types then an account

### LINKED ISSUE #3980: support for other resources then accounts on list_tags_for_resources for the organizations resource
Currently the list_tags_for_resource() method of the organizations resource only supports tags on accounts. While the AWS API call supports:

- AWS account
- Organization root
- Organizational unit (OU)
- Policy (any type)

See the API [docs](https://docs.aws.amazon.com/organizations/latest/APIReference/API_ListTagsForResource.html)

This currently results in a InvalidInputException which is incorrect

stromp:
fixes #3980 

- Adding checking for resource type in tag functions
- Adding TargetNotFoundException when no resource found
- Adding support for tags for root OU, OU and Policies
- Adding tests covering the new code
- Adding test for deletion of a tag

### DIFF
--- a/moto/organizations/models.py
+++ b/moto/organizations/models.py
@@ -656,6 +656,25 @@ class OrganizationsBackend(BaseBackend):
             ]
         )
 
+    def _get_resource_for_tagging(self, resource_id):
+        if re.compile(utils.OU_ID_REGEX).fullmatch(resource_id) or re.fullmatch(
+            utils.ROOT_ID_REGEX, resource_id
+        ):
+            resource = next((a for a in self.ou if a.id == resource_id), None)
+        elif re.compile(utils.ACCOUNT_ID_REGEX).fullmatch(resource_id):
+            resource = next((a for a in self.accounts if a.id == resource_id), None)
+        elif re.compile(utils.POLICY_ID_REGEX).fullmatch(resource_id):
+            resource = next((a for a in self.policies if a.id == resource_id), None)
+        else:
+            raise InvalidInputException(
+                "You provided a value that does not match the required pattern."
+            )
+
+        if resource is None:
+            raise TargetNotFoundException
+
+        return resource
+
     def list_targets_for_policy(self, **kwargs):
         if re.compile(utils.POLICY_ID_REGEX).match(kwargs["PolicyId"]):
             policy = next(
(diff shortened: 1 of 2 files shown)


moto/organizations/models.py
```diff
@@ -672,38 +674,39 @@ def list_targets_for_policy(self, **kwargs):
...
-        if account is None:
+    def _get_resource_to_tag(self, resource_id):
+        if re.compile(utils.OU_ID_REGEX).match(resource_id) or re.match(
+            utils.ROOT_ID_REGEX, resource_id
+        ):
+            resource = next((a for a in self.ou if a.id == resource_id), None)
+        elif re.compile(utils.ACCOUNT_ID_REGEX).match(resource_id):
+            resource = next((a for a in self.accounts if a.id == resource_id), None)
+        elif re.compile(utils.POLICY_ID_REGEX).match(resource_id):
+            resource = next((a for a in self.policies if a.id == resource_id), None)
```
bblommers, line 685: This should have a test as well

tests/test_organizations/test_organizations_boto3.py
```diff
@@ -1025,14 +1092,23 @@ def test_list_tags_for_resource_errors():
...
-        client.list_tags_for_resource(ResourceId="000000000000")
+        client.list_tags_for_resource(ResourceId="000x00000A00")
     ex = e.value
     ex.operation_name.should.equal("ListTagsForResource")
     ex.response["ResponseMetadata"]["HTTPStatusCode"].should.equal(400)
     ex.response["Error"]["Code"].should.contain("InvalidInputException")
     ex.response["Error"]["Message"].should.equal(
         "You provided a value that does not match the required pattern."
     )
+    with pytest.raises(ClientError) as e:
```
bblommers, line 1103: Looks like this test is duplicated three times

bblommers (changes requested):

moto/organizations/models.py
stromp: This is implicitly tested by the different tests of tagging the different resources. 

but I can add explicit tests if that is preferred?

tests/test_organizations/test_organizations_boto3.py
stromp: the test is done with 3 different resources,

one test tests tagging an account and checks the tags
second test tests tagging an root and checks the tags
etc.

I can consolidate the tests and do a for loop with the different resource types. or consolidate parts in a helper function. What is your preferred approach?

bblommers: My bad, I didn't mean the entire test - just the call to `client.list_tags_for_resource(ResourceId="000000000000")`. 
That method-call, and the assertion that it throws a TargetNotFoundException, is duplicated.

moto/organizations/models.py
bblommers: Yes, please. I can see it should work, but it's always good to have explicit tests

tests/test_organizations/test_organizations_boto3.py
stromp: You are right, I was also not paying attention.  This should check 3 different functions, list_tags, create_tag etc. will update!

moto/organizations/models.py
stromp: I have added them.

tests/test_organizations/test_organizations_boto3.py
stromp: I fixed the tests. They now all test different responses per account type

bblommers:
Thanks @stromp, that looks good to me. The build is currently failing because of a dependency issue - leave it with me, I'll have a look at that later.

bblommers (approved):

PR #3983, opened 2021-06-03
Media package client error

arcovoltaico:
This PR makes MediaPackage throwing a ClientError, exactly as Boto would do by converting any AWS Error into a ClientError as documented here https://boto3.amazonaws.com/v1/documentation/api/latest/guide/error-handling.html

### DIFF
--- a/moto/mediapackage/exceptions.py
+++ b/moto/mediapackage/exceptions.py
@@ -1,1 +1,13 @@
 from __future__ import unicode_literals
+
+from moto.core.exceptions import JsonRESTError
+
+
+class MediaPackageClientError(JsonRESTError):
+    code = 400
+
+
+# AWS service exceptions are caught with the underlying botocore exception, ClientError
+class ClientError(MediaPackageClientError):
+    def __init__(self, error, message):
+        super(ClientError, self).__init__(error, message)
--- a/tests/test_mediapackage/test_mediapackage.py
+++ b/tests/test_mediapackage/test_mediapackage.py
@@ -1,7 +1,10 @@
 from __future__ import unicode_literals
 
 import boto3
+import pytest
 import sure  # noqa
+from botocore.exceptions import ClientError  # Boto3 will always throw this exception
+
 from moto import mock_mediapackage
 
 region = "eu-west-1"
@@ -11,7 +14,7 @@ def _create_channel_config(**kwargs):
     id = kwargs.get("id", "channel-id")
     description = kwargs.get("description", "Awesome channel!")
     tags = kwargs.get("tags", {"Customer": "moto"})
-    channel_config = dict(Description=description, Id=id, Tags=tags,)
+    channel_config = dict(Description=description, Id=id, Tags=tags)
     return channel_config
 
 
@@ -79,6 +82,16 @@ def test_describe_channel_succeeds():
     )
     describe_response["Description"].should.equal(channel_config["Description"])
     describe_response["Tags"]["Customer"].should.equal("moto")
+
+
+@mock_mediapackage
+def test_describe_unknown_channel_throws_error():
+    client = boto3.client("mediapackage", region_name=region)
+    channel_id = "unknown-channel"
+    with pytest.raises(ClientError) as err:
+        client.describe_channel(Id=channel_id)
+    assert err.value.response["Error"]["Code"] == "NotFoundException"
+    assert err.value.response["Error"]["Message"] == f"'{channel_id}'"
 
 
 @mock_mediapackage
--- a/tests/test_mediapackage/test_mediapackage.py
+++ b/tests/test_mediapackage/test_mediapackage.py
@@ -91,7 +91,8 @@ def test_describe_unknown_channel_throws_error():
     with pytest.raises(ClientError) as err:
         client.describe_channel(Id=channel_id)
     assert err.value.response["Error"]["Code"] == "NotFoundException"
-    assert err.value.response["Error"]["Message"] == f"'{channel_id}'"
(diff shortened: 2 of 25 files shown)


moto/mediapackage/exceptions.py
```diff
@@ -1 +1,13 @@
...
+
+from moto.core.exceptions import JsonRESTError
+
+
+class MediaPackageClientError(JsonRESTError):
+    code = 400
+
+
+# AWS service exceptions are caught with the underlying botocore exception, ClientError
+class ClientError(MediaPackageClientError):
```
bblommers, line 11: Moto intercepts any HTTP requests that are made. Whatever error is thrown by us, will be converted into a HTTP 400 response. Boto3 will in turn throw a ClientError for any 400-error that is returned.

TLDR: The name doesn't matter here. Considering it's purpose, it's probably cleaner to make it a ChannelNotFoundError (with a hardcoded message)

bblommers (changes requested):
Hi @arcovoltaico, thanks for the PR!
Would you mind adding a test for this? There's a clean example on how to to that here: https://github.com/spulec/moto/blob/master/tests/test_s3/test_s3.py#L705

moto/mediapackage/exceptions.py
arcovoltaico: My production code has fragments like this:
```
 try:
            return self.aws_client.update_origin_endpoint(**params)
        except ClientError as error:
            if 'NotFoundException' == error.response['Error']['Code']:
                Logger.info(f'OriginEndpoint {params} was not found')
            else:
                Logger.info(f'Errorr updating media package with params {params} :  {error}')

            raise error
```

And it works nicely catching Boto errors. So my purpose is to make Moto throw the exact error as Boto does. I am not sure if there's a way to make it work if we don't use the same error naming.

arcovoltaico:
> Hi @arcovoltaico, thanks for the PR!
> Would you mind adding a test for this? There's a clean example on how to to that here: https://github.com/spulec/moto/blob/master/tests/test_s3/test_s3.py#L705

This would be my attempt. But I don't know what URL I should use on the get, and how/where are listed/mapped.
Thanks for the help
```
from moto.mediapackage.exceptions import ClientError

@mock_mediapackage
def test_describe_channel():
    backend = server.create_backend_app("mediapackage")
    test_client = backend.test_client()

        res = test_client.get("/channel-id")
    assert err.value.response["Error"]["Code"] == "NotFoundException"
```

bblommers:
You should be able to use the boto3-client:
```
from botocore.exceptions import ClientError  # Boto3 will always throw this exception

@mock_mediapackage
def test_describe_unknown_channel():
    client = boto3.client("mediapackage", region_name=region)

        client.describe_channel(Id="unknown-channel")
    assert err.value.response["Error"]["Code"] == "??"
    assert err.value.response["Error"]["Message"] == "??"    
```

arcovoltaico:
> You should be able to use the boto3-client:
> 
> ```
> from botocore.exceptions import ClientError  # Boto3 will always throw this exception
> 
> @mock_mediapackage
> def test_describe_unknown_channel():
>     client = boto3.client("mediapackage", region_name=region)
> 
>     with pytest.raises(ClientError) as err:
>         client.describe_channel(Id="unknown-channel")
>     assert err.value.response["Error"]["Code"] == "??"
>     assert err.value.response["Error"]["Message"] == "??"    
> ```

so you mean then, catch and raise the error on the response instead of the model? and add the test on test_mediapackage instead of on the test_server?

bblommers:
> so you mean then, catch and raise the error on the response instead of the model? 

Raising it on the model should be fine,

> and add the test on test_mediapackage instead of on the test_server?

but yes. The test should reflect a normal use case, i.e. in what scenario will a user encounter this error

tests/test_mediapackage/test_mediapackage.py
```diff
@@ -81,6 +84,16 @@ def test_describe_channel_succeeds():
...
 
 
+@mock_mediapackage
+def test_describe_unknown_channel_throws_error():
+    client = boto3.client("mediapackage", region_name=region)
+    channel_id = "unknown-channel"
+    with pytest.raises(ClientError) as err:
+        client.describe_channel(Id=channel_id)
+    assert err.value.response["Error"]["Code"] == "NotFoundException"
+    assert err.value.response["Error"]["Message"] == f"'{channel_id}'"
```
bblommers, line 94: Python2 doesn't support f-strings, looks like that's causing the build to fail. Happy to merge once the CI is happy though 👍

arcovoltaico: I fixed it. But although I ran black, version 19.10b0 or even make format, make lint, and no changes were done. But still, CI is failing. Any hint? Sorry for the hassle.

bblommers: @arcovoltaico Python 2 created a unicode-string out of the Exception that was thrown. When running the test, it was comparing a normal string with an unicode-string - and that was failing. 
```
assert "u'unknown-channel'" == "'unknown-channel'"
```
Running this scenario against AWS showed that the error message should be slightly different anyway, so I reworked the logic to return the correct error message

bblommers (approved):
Thanks again for the PR!

arcovoltaico:
> This is now part of moto >= 2.0.9.dev19

Any estimation on the release date? If it takes time I would need to do a provisional fix on our production code. Cheers :-)

bblommers:
@arcovoltaico No promises! But I'll try to release this weekend

bblommers:
2.0.9 has just been released @arcovoltaico

PR #3985, opened 2021-06-03
Fix : Logs - Add start_query functionality

usmangani1:
Fix : Logs - Add start_query functionality

https://github.com/localstack/localstack/issues/4078

### DIFF
--- a/moto/logs/models.py
+++ b/moto/logs/models.py
@@ -8,6 +8,15 @@ from .exceptions import (
     InvalidParameterException,
     LimitExceededException,
 )
+import uuid
+
+
+class LogQuery(BaseModel):
+    def __init__(self, id, start_time, end_time, query):
+        self.id = id
+        self.start_time = start_time
+        self.end_time = end_time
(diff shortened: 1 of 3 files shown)


moto/logs/models.py
```diff
@@ -725,6 +735,16 @@ def delete_subscription_filter(self, log_group_name, filter_name):
 
         log_group.delete_subscription_filter(filter_name)
 
+    def start_query(self, log_group_names, start_time, end_time, query_string):
+
+        for log_group_name in log_group_names:
+            if log_group_name not in self.groups:
+                raise ResourceNotFoundException()
```
bblommers, line 742: Can we get a unit test for this scenario as well?

bblommers (changes requested):

moto/logs/models.py
usmangani1: Done

bblommers (approved):
LGTM!

PR #3989, opened 2021-06-05
Pin Jose-dependency for Py2

bblommers:
`python-jose` had an update to 3.3.0, where support for Py2 was dropped.

Failure log: https://github.com/spulec/moto/runs/2752216996?check_suite_focus=true


passage2016:
python3.5 has the same problem with python2

bblommers:
@passage2016 Moto doesn't officially support Python3.5, so you might have to pin this dependency in your own project to make it work

PR #3993, opened 2021-06-07
Adding support for `ForwardConfig` property in `ListernRule` in CloudFormation

sahilshah6196:
Right now, `ListerRule` does not support the `ForwardConfig` property that comes under the `Action` section in `AWS::ElasticLoadBalancingV2::ListenerRule`. This PR aims to add that.

Here are the AWS docs for the above: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-elasticloadbalancingv2-listenerrule-forwardconfig.html

Follow-up to: https://github.com/spulec/moto/pull/3969


sahilshah6196:
Hello again @bblommers What do you think about this PR?

moto/elbv2/models.py
```diff
@@ -666,14 +695,22 @@ def _validate_actions(self, actions):
...
                 if action_target_group_arn not in target_group_arns:
                     raise ActionTargetGroupNotFoundError(action_target_group_arn)
             elif action_type == "fixed-response":
                 self._validate_fixed_response_action(action, i, index)
             elif action_type in ["redirect", "authenticate-cognito"]:
                 pass
+            # pass if listener rule has forward_config as an Action property
+            elif (
+                action_type == "forward"
+                and "forward_config._target_groups.member.{}._target_group_arn".format(
```
bblommers, line 709: This is not a boolean condition - I'm guessing that should say something like [arn]` == action.data`

bblommers (changes requested):
Thanks for the follow-up @sahilshah6196 - just have one minor comment.

It looks like the `test_handle_listener_rules`-test has become shaky, i.e. passes sometimes but sometimes fails. I'll probably hold off on merging this until we figure out what's going on there.
(Example failure: https://github.com/spulec/moto/runs/2772429387?check_suite_focus=true)

moto/elbv2/models.py
sahilshah6196: @bblommers Looks good now? 14999cb

Issue #2288, opened 2019-07-09
Question: moto_service s3 becoming slow and timing out as tests continue


gmossessian:
I have a test suite of ~1000 tests. Previously we were using the mock_s3 decorator which only worked because our "integration tests" just imported modules from other services and mocked their functionality. This was always been undesirable but has not been a priority to fix until recently. 

I'm attempting to use moto in standalone server mode, hackily starting the service in conftest.py using 

```
def start_local_moto_server():
    global moto_subprocess
    moto_subprocess = subprocess.Popen(
        ['moto_server', 's3'],
        stdout=subprocess.PIPE, stderr=subprocess.PIPE,
        bufsize=0, universal_newlines=True,
    )
    print 'Waiting for moto local service to start...'
    while True:
        moto_log = moto_subprocess.stderr.readline()
        print moto_log
        if 'Running' in moto_log:
            break

    boto3.resource(
        's3', region_name='us-east-1', endpoint_url='http://127.0.0.1:5000'
    ).create_bucket(Bucket=TEST_S3_BUCKET_NAME)
```

As the tests progress, any calls to the s3 client become slower and slower, until finally I get 

`ReadTimeoutError: Read timeout on endpoint URL: "http://localbox:5000/my-bucket/my/key/file.tar"`

Running those tests individually causes all of them to pass. 

I don't really know where to begin debugging this or understanding the issue, or how to provide more information. Could someone please help identify the problem?

gmossessian:
(cross-posting to SO [here](https://stackoverflow.com/questions/56961298/moto-server-s3-slows-down-as-tests-progress))

spulec:
For the most part, moto will store S3 files in memory. It looks like anything greater than 16MB, we will flush to disk.

If you continue to write files under 16MB to moto s3, then the slowness would make sense.

A couple options:
- do you need to persist the same s3 backend for each test? If not, you can issue a reset to the backend to clear it (`http://localhost:5000/moto-api/reset`)
- Maybe our 16MB is a bad config and we should lower it

gmossessian:
That did help with the speed issues! Thanks.

I wonder if that 16mb number could be configurable, as well as the total memory used before flushing all files to disk?

spulec:
Yeah, we can make the individual file size to flush configurable, but we don't have an easy way to make this global.

If anyone has time to make a PR to do that, I would be happy to accept, but otherwise am going to close this out for now. Let me know if there is anything else!

PR #3994, opened 2021-06-08
Reduce DEFAULT_KEY_BUFFER_SIZE to avoid in memory caching of multipart uploads

benediktbrandt:
When uploading large files via multipart uploads, moto's memory consumption increases by the size of the file. 

The reason for this is that `DEFAULT_KEY_BUFFER_SIZE` is set to 16 MB, however multipart upload size is up to the client. If clients use less than 16 MB for each part, moto stores every part in memory. This PR sets the `DEFAULT_KEY_BUFFER_SIZE` to 1KB less of `UPLOAD_PART_MIN_SIZE`, which will ensure that multipart uploads are written to disk. 

Since this is often but potentially not always the desired behavior, this PR allows the setting of the `DEFAULT_KEY_BUFFER_SIZE` value via the environment variable `DEFAULT_KEY_BUFFER_SIZE`.

Additional details:

Moto stores the individual parts of a multipart upload in the `multipart` field of the `FakeKey` object. According to the aws documentation on multipart uploads https://docs.aws.amazon.com/AmazonS3/latest/userguide/qfacts.html the size of a multipart upload is between 5 MB and 5 GB and the selection of the size is up to the client. In local runs I often had the aws cli choose a size of 8MB for each part, causing moto to store each part in memory.

I believe this PR will solve these issues:

https://github.com/spulec/moto/issues/2288
https://github.com/localstack/localstack/issues/682


moto/s3/models.py
```diff
@@ -59,7 +59,9 @@
     "GLACIER",
     "DEEP_ARCHIVE",
 ]
-DEFAULT_KEY_BUFFER_SIZE = 16 * 1024 * 1024
+DEFAULT_KEY_BUFFER_SIZE = int(
```
bblommers, line 62: This should be set in moto/settings.py, to ensure we have all env variables in one place.
Can  you make it a method, as well? As a variable, it can only be set once - with a method, people can set different values for the env variable if and when required, and it will always get the most recent env variable value.

bblommers (changes requested):
Hi @benediktbrandt, thanks for the PR!
Don't worry about the failing test - it's a known issue.

I think it would be best to split this up in 2 PR's:
 - Introduce a variable to override the buffer size
 - Lower the default buffer size

The first PR can be merged immediately, but the second has the potential to have a big impact on users, so that PR should be scheduled for a minor release.

moto/s3/models.py
benediktbrandt: Sounds good, I will make that update.

benediktbrandt:
Thanks @bblommers for the review. Your suggestion of splitting the PR makes a lot of sense. Here is the first PR, the one 
that introduces the environment variable https://github.com/spulec/moto/pull/4001 I will open the second PR (the one that reduces the default value) once the first PR goes into master since the second PR depends on the first.

benediktbrandt:
Closing in favor of https://github.com/spulec/moto/pull/4001 and an upcoming follow up PR that reduces the default size.

PR #3999, opened 2021-06-09
ELBv2 - Simplify tests

bblommers:


bblommers:
FYI @bpandola - this should make it a bit easier to identify why the flaky ELB test fails

bpandola:
@bblommers Thanks!  I figured it out with the help of the additional output from your commit.  I've got a fix here: https://github.com/spulec/moto/pull/4002

PR #4001, opened 2021-06-09
Env variable for default key buffer size

benediktbrandt:
Continuation from: https://github.com/spulec/moto/pull/3994


bblommers (approved):
LGTM - thanks @benediktbrandt!

PR #4002, opened 2021-06-10
Refactor listener rule ARN generation to avoid collisions

bpandola:
The code that generates an ARN for a listener rule was relying on a random
number between 0 and 50 to ensure uniqueness.  As a result, there was a decent
chance of generating a collision, particularly when adding multiple rules.

When a collision occurred, the new rule simply overwrote an existing rule,
causing a test that asserts on an expected number of rules being returned
to fail.

Reference the following runs for related test failures:
https://github.com/spulec/moto/runs/2772429387
https://github.com/spulec/moto/runs/2786330037

### DIFF
--- a/moto/elbv2/models.py
+++ b/moto/elbv2/models.py
@@ -1,5 +1,4 @@
 from __future__ import unicode_literals
-import random
 
 import datetime
 import re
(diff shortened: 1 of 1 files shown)


bblommers (approved):
Nice one @bpandola - thanks for looking into this!

PR #4003, opened 2021-06-10
Reduce default value for DEFAULT_KEY_BUFFER_SIZE

benediktbrandt:
This is a continuation of https://github.com/spulec/moto/pull/3994

This PR reduces the default value for DEFAULT_KEY_BUFFER_SIZE to 1KB less than the UPLOAD_PART_MIN_SIZE. This prevents in memory caching for multi part uploads. To facilitate this change I had to move the constant UPLOAD_PART_MIN_SIZE to the settings.py so that I was able to use it in the `get_s3_default_key_buffer_size` function without circular import issues.


bblommers (approved):
Thanks @benediktbrandt!

bblommers:
Scheduled this for the next minor release.

PR #4006, opened 2021-06-11
Escape Hyphen in Regular Expression

kyledecot:
Currently you cannot have a SSM parameter name that contains a hyphen such as `/my-cool-parameter` due to a missing backslash.

### DIFF
--- a/tests/test_ssm/test_ssm_boto3.py
+++ b/tests/test_ssm/test_ssm_boto3.py
@@ -1394,7 +1394,7 @@ def test_label_parameter_version_invalid_name():
 
     test_parameter_name = "test"
 
-    response = client.label_parameter_version.when.called_with(
+    client.label_parameter_version.when.called_with(
         Name=test_parameter_name, Labels=["test-label"]
     ).should.throw(
         ClientError,
@@ -1415,7 +1415,7 @@ def test_label_parameter_version_invalid_parameter_version():
         Type="String",
     )
 
-    response = client.label_parameter_version.when.called_with(
+    client.label_parameter_version.when.called_with(
         Name=test_parameter_name, Labels=["test-label"], ParameterVersion=5
     ).should.throw(
         ClientError,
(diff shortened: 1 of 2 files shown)


bblommers (changes requested):
Thanks for the fix @kyledecot!
Would you mind adding a test for this as well? Easiest way is probably to extend the existing `test_put_parameter` test (in `tests/test_ssm/test_ssm_boto3.py`), and parametrize it to verify that both `test` and `/my-cool-parameter` are valid names.

(An example of parameterization already exists in the same file, in the `test_describe_parameters_invalid_path`-test)

tests/test_ssm/test_ssm_boto3.py
```diff
@@ -227,76 +227,73 @@ def test_get_parameters_by_path():
...
     len(response["Parameters"]).should.equal(1)
-    response["Parameters"][0]["Name"].should.equal("test")
+    response["Parameters"][0]["Name"].should.equal(name)
     response["Parameters"][0]["Value"].should.equal("value")
     response["Parameters"][0]["Type"].should.equal("String")
     response["Parameters"][0]["Version"].should.equal(1)
     response["Parameters"][0]["LastModifiedDate"].should.be.a(datetime.datetime)
     response["Parameters"][0]["ARN"].should.equal(
-        "arn:aws:ssm:us-east-1:1234567890:parameter/test"
+        f"arn:aws:ssm:us-east-1:1234567890:parameter/{name}"
```
bblommers, line 250: We still support Py2, so f-strings will not work unfortunately.
We usually use "". format

correnarsprowls: Thanks for the feedback. I took care of this 🎉  Since I pushed up the change and not Kyle, it's not running the build.

bblommers (approved):
LGTM - thanks @kyledecot and @correnarsprowls!

PR #4008, opened 2021-06-14
MediaPackage Additional Error Handling

arcovoltaico:
Some describe, delete and update methods were not catching KeyError. This PR addresses those cases by throwing a ClientError.


arcovoltaico:
Sorry. The unit tests are running fine locally on my machine. I don't know why it's throwing a different error here. @bblommers  what I am missing? Cheers

bblommers:
Hi @arcovoltaico, the `@mock_mediapackage`-decorator is missing from the tests. It probably works on your local machine because it runs against AWS itself (and succeeds.)

Note: With the decorator, some of the tests still fail because of a typo in the message. If these tests do succeed against AWS (as is), then the expected error message is correct - just the message that is thrown in Moto should be fixed

PR #4011, opened 2021-06-16
Media package client error additional handling

arcovoltaico:
Some describe, delete and update methods were not catching KeyError. This PR addresses those cases by throwing a ClientError.


bblommers (approved):
LGTM - thanks @arcovoltaico!

Issue #3760, opened 2021-03-09
mock_dynamodb2 fails to update field that is used in a Global Secondary Index


h0wser:
In the newest version of moto 2.0.1, when calling update_item on a field that is used as a key in a GSI, the operation triggers a ValidationException saying that an AttributeValue may not contain an empty string.

**how to reproduce the issue**

Save the following code to .py file and run.

``` python
from moto import mock_dynamodb2
import boto3

with mock_dynamodb2():
    name = "TestTable"
    conn = boto3.client("dynamodb", region_name="eu-west-2", aws_access_key_id="ak", aws_secret_access_key="sk")
    conn.create_table(
        TableName=name,
        KeySchema=[{"AttributeName": "main_key", "KeyType": "HASH"}],
        AttributeDefinitions=[{"AttributeName": "main_key", "AttributeType": "S"},
            {"AttributeName": "index_key", "AttributeType": "S"}],
        ProvisionedThroughput={"ReadCapacityUnits": 5, "WriteCapacityUnits": 5},
        GlobalSecondaryIndexes=[
            {
                "IndexName": "test_index",
                "KeySchema": [
                    {
                        "AttributeName": "index_key",
                        "KeyType": "HASH"
                    }
                ],
                "Projection": {
                    "ProjectionType": "ALL",
                },
                "ProvisionedThroughput": {"ReadCapacityUnits": 1, "WriteCapacityUnits": 1}
            }
        ]
    )

    conn.put_item(
        TableName=name,
        Item={
            "main_key": {"S": "testkey1"},
            "extra_data": {"S": "testdata"},
            "index_key": {"S": "indexkey1"}
        },
    )

    resp = conn.update_item(
        TableName=name,
        Key={"main_key": {"S": "testkey1"}},
        UpdateExpression="set index_key=:new_index_key",
        ExpressionAttributeValues={":new_index_key": {"S": "new_value"}}
    )

    print(resp)
```

**what you expected to happen**
That the attribute index_key updates successfully.

I have validated that this code runs correctly against dynamodb proper without issues. It also works under the 1.3.6 version of moto.

**what actually happens**
```
Traceback (most recent call last):
  File "test.py", line 41, in <module>
    resp = conn.update_item(
  File "/home/h0wser/.local/lib/python3.8/site-packages/botocore/client.py", line 357, in _api_call
    return self._make_api_call(operation_name, kwargs)
[...]

PR #3763, opened 2021-03-10
DynamoDB - only raise exception when GSI key is actually empty

bblommers:
Fixes #3760

PR #4014, opened 2021-06-16
Only raise EmptyKeyAttributeException when the value node is empty

thomassross:
This fixes a bug where all updates to keys (including GSIs) would raise
EmptyKeyAttributeException, even when the new value was not an empty string.


thomassross:
This fixes #3760. It's also a duplicate of #3763 - serves me right for not looking first. Leaving it open because the test case is different from the one in the other PR.

bblommers (approved):
LGTM - thanks @thomassross!

PR #4015, opened 2021-06-17
Disallow termination of protected EMR job flows

bpandola:
Error message verified against real AWS backend.


bblommers (approved):
Nice one - thanks @bpandola!

PR #4017, opened 2021-06-17
Added "isTruncated" attribute for listUsersResponse

pinzon:
This PR adds the isTruncated attribute to the xml for the ListsUserResponse.

According to the documentation mentioned in the issue localstack/localstack#4150, the Java AWS SDK v2 is expecting the attribute when parsing the response.

### DIFF
--- a/moto/iam/responses.py
+++ b/moto/iam/responses.py
@@ -492,10 +492,9 @@ class IamResponse(BaseResponse):
         path_prefix = self._get_param("PathPrefix")
         marker = self._get_param("Marker")
         max_items = self._get_param("MaxItems")
-        is_truncated = self._get_bool_param("IsTruncated", False)
         users = iam_backend.list_users(path_prefix, marker, max_items)
         template = self.response_template(LIST_USERS_TEMPLATE)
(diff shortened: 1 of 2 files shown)


moto/iam/responses.py
```diff
@@ -492,9 +492,10 @@ def list_users(self):
         path_prefix = self._get_param("PathPrefix")
         marker = self._get_param("Marker")
         max_items = self._get_param("MaxItems")
+        is_truncated = self._get_bool_param("IsTruncated", False)
```
bblommers, line 495: This is not a user supplied argument, but depends on whether we should paginate this call. Considering pagination is not yet implemented for this particular method, this attribute can always be False (for now at least)

bblommers (approved):
Hi @pinzon, thanks for the PR! Would you mind adding a test case to verify the attribute is returned correctly?

belosh59:
nice. Thank you guys

pinzon:
@bblommers thank you for the review, I made the change and added to the **list_users test** the assertion of the attribute.

bblommers (approved):
LGTM! Thanks again @pinzon

PR #4019, opened 2021-06-18
Support EMR clusters with multiple master nodes

bpandola:
Reference: https://docs.aws.amazon.com/emr/latest/ManagementGuide/emr-plan-ha-launch.html


bblommers (approved):
LGTM- thanks @bpandola

PR #4024, opened 2021-06-22
Fix validation of values in case of DELETE action

avislav:
In case of DELETE action we don't pass the value if we want to remove the attribute.
There are more cases but it's for you to handle.


bblommers (approved):
LGTM - thanks for the PR @avislav!

PR #4026, opened 2021-06-22
Fixed validation on custom-resource in applicationautoscaling

psheets:
Helper function _get_resource_type_from_resource_id() did not handle custom-resources correctly.  the resource id of a custom-resource does not need to conform to a standard and thus does not need to be validated.


bblommers (approved):
LGTM - thanks for the PR @psheets!

PR #4039, opened 2021-06-27
Fix: SQS: get_queue_attributes

usmangani1:
Fix: SQS: get_queue_attributes

https://github.com/localstack/localstack/issues/4178


bblommers (changes requested):
Thanks for the PR @usmangani1 - happy to merge once the build passes

PR #4042, opened 2021-06-28
SSM parameters in cloudformation (AWS::SSM::Parameter::) are not recognized and resolved for default values

sahilshah6196:
If a particular cloudformation parameter has a `Default` value and is of `Type: AWS:SSM:PARAMETER..`, it is not resolved and hence not parsed. This PR aims to fix that


sahilshah6196:
@bblommers Hello! Could you review this PR? Thanks

moto/cloudformation/parsing.py
```diff
@@ -504,11 +504,26 @@ def transform_mapping(self):
...
+        return actual_value
+
     def load_parameters(self):
         parameter_slots = self._template.get("Parameters", {})
         for parameter_name, parameter in parameter_slots.items():
             # Set the default values.
-            self.resolved_parameters[parameter_name] = parameter.get("Default")
+            value = parameter.get("Default")
+            value_type = parameter.get("Type")
+            if parameter.get("Type").startswith("AWS::SSM::Parameter::") and value:
```
bblommers, line 524: `parameter.get("Type")` was already resolved, so this can be simplified to `if value_type.startswith...`

tests/test_cloudformation/test_stack_parsing.py
```diff
@@ -78,7 +78,10 @@
 ssm_parameter = {
     "Parameters": {
         "SingleParamCfn": {"Type": "AWS::SSM::Parameter::Value<String>"},
-        "ListParamCfn": {"Type": "AWS::SSM::Parameter::Value<List<String>>"},
+        "ListParamCfn": {
+            "Type": "AWS::SSM::Parameter::Value<List<String>>",
+            "Default": "/path/to/list/param",
```
bblommers, line 83: We'll have to verify that this value is actually retrieved from SSM, i.e. extend the test to add something to the parameter store, and assert afterwards that the values match up

bblommers (changes requested):

tests/test_cloudformation/test_stack_parsing.py
sahilshah6196: That is being done here: https://github.com/sahilshah6196/moto/blob/master/tests/test_cloudformation/test_stack_parsing.py#L528
The `ssm_parameter` on line 81 is being used in the test I linked above. 
Sorry if I am misunderstanding the request.

moto/cloudformation/parsing.py
sahilshah6196: @bblommers Absolutely. Fixed: e9585a2

tests/test_cloudformation/test_stack_parsing.py
bblommers: That test only verifies the user-defined value, as that is passed in as part of `FakeStack(parameters=..)`
In order to assert that the default works, we should also create a FakeStack without passing in a value:
```
stack = FakeStack(
            stack_id="test_id",
            name="test_stack",
            template=ssm_parameter_template_json,
            parameters={
                "SingleParamCfn": "/path/to/single/param" # Not passing in a value for ListParamCfn!
            },
            region_name="us-west-1",
        )
       # and copy the same asserts, to verify that the 'ListParamCfn' now has the default value
```

sahilshah6196: Extended that test: 3486f7a
@bblommers That looks good?

sahilshah6196: Got an unwanted change in there, sorry. Removed: 8e65e26

bblommers: Yes, that's exactly what I had in mind

bblommers (approved):
Thanks for the PR @sahilshah6196!

PR #4043, opened 2021-06-29
Fix error message for missing log stream

### LINKED ISSUE #3978: Wrong error message when pushing logs to non-existent stream in existing group
# MWE

```
import boto3
from moto import mock_logs
from time import time

mock_logs().start()

group_name = 'my-group-2'
stream_name = 'my-stream'

client = boto3.client('logs')

client.create_log_group(logGroupName=group_name)

# I forgot to do this in my real code
# so the error from moto/boto should tell me this is the part I forgot
#client.create_log_stream(
#    logGroupName=group_name,
#    logStreamName=stream_name
#)

client.put_log_events(
    logGroupName=group_name,
    logStreamName=stream_name,
    logEvents=[
        {
            'timestamp': int(time()),
            'message': 'test'
        },
    ]
)
```

# Expected behavior

If you run the above script without `moto`, you get:

> botocore.errorfactory.ResourceNotFoundException: An error occurred (ResourceNotFoundException) when calling the PutLogEvents operation: The specified log **stream** does not exist.

# actual behavior

If you run the above script with `moto`, you get:

> botocore.errorfactory.ResourceNotFoundException: An error occurred (ResourceNotFoundException) when calling the PutLogEvents operation: The specified log **group** does not exist

The group *does* exist. The problem with the `put_log_events` call is that the *stream* does not exist. This difference in error message makes it hard to figure out why the whole script does not pass.

bpandola:
Closes #3978

### DIFF
--- a/moto/logs/models.py
+++ b/moto/logs/models.py
@@ -333,7 +333,7 @@ class LogGroup(BaseModel):
         self, log_group_name, log_stream_name, log_events, sequence_token
     ):
         if log_stream_name not in self.streams:
-            raise ResourceNotFoundException()
+            raise ResourceNotFoundException("The specified log stream does not exist.")
         stream = self.streams[log_stream_name]
         return stream.put_log_events(
             log_group_name, log_stream_name, log_events, sequence_token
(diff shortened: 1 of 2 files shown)


bblommers (approved):
LGTM - thanks @bpandola!

PR #4049, opened 2021-06-30
Transform certificates in listener in expected XML

sahilshah6196:
Listener certs needs to be transformed as other properties in elbv2 resoruces

Example: https://github.com/spulec/moto/blob/master/moto/elbv2/models.py#L614


moto/elbv2/models.py
```diff
@@ -1308,41 +1319,23 @@ def modify_listener(
...
-                            raise RESTError(
-                                "CertificateNotFound",
-                                "Certificate {0} not found".format(
-                                    cert["certificate_arn"]
-                                ),
-                            )
-
-                        all_certs.add(cert["certificate_arn"])
-
-                    if default_cert is None:
```
sahilshah6196, line 1339: All this is redundant cause of what is in the AWS docs: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-elasticloadbalancingv2-listener.html#cfn-elasticloadbalancingv2-listener-certificates

> You must provide exactly one certificate if the listener protocol is HTTPS or TLS.

sahilshah6196:
@bblommers Hey! Can you review this PR? Thanks!

moto/elbv2/models.py
```diff
@@ -833,6 +835,15 @@ def create_target_group(self, name, **kwargs):
         self.target_groups[target_group.arn] = target_group
         return target_group
 
+    def convert_and_validate_certificates(self, certificates):
+
+        # transform default certificate to conform with the rest of the code and XML templates
+        if certificates:
+            for cert in certificates:
```
bblommers, line 842: Let's simplify this to `for cert in certificates or []:` (so we can get rid of the if-statement)

tests/test_elbv2/test_elbv2.py
```diff
@@ -1796,7 +1793,7 @@ def test_modify_listener_http_to_https():
             Port=443,
             Protocol="HTTPS",
             SslPolicy="ELBSecurityPolicy-TLS-1-2-2017-01",
-            Certificates=[{"CertificateArn": google_arn, "IsDefault": False}],
+            Certificates=[],
```
bblommers, line 1796: Would you mind adding assertions to this check? I know they weren't there in the first place, but it would be nice to have.
```
with pytest.raises() as ex:
    ...
err["Code"].should.equal(..)
err["Message"].should.equal(..)
```

bblommers (changes requested):

moto/elbv2/models.py
sahilshah6196: Done: 86d2773

tests/test_elbv2/test_elbv2.py
sahilshah6196: Absolutely, makes sense to add it: a9df01f

bblommers (approved):
LGTM - thanks @sahilshah6196!

PR #4050, opened 2021-07-02
SES domains are case-insensitive

### LINKED ISSUE #3947: SES get_dkim_tokens creates duplicate domain
# Steps to reproduce:

```
from moto import mock_ses
import boto3
import unittest

    def test_moto(self):
        domain = 'example.com'
        with mock_ses():
            client = boto3.client('ses')
        
            response = client.list_identities(IdentityType='Domain')
            self.assertEqual(sum([(domain == i) for i in response.get('Identities', [])]), 0)
        
            ret = client.verify_domain_identity(Domain=domain)
            response = client.list_identities(IdentityType='Domain')
            self.assertEqual(sum([(domain == i) for i in response.get('Identities', [])]), 1)
        
            response = client.verify_domain_dkim(Domain=domain)
            response = client.list_identities(IdentityType='Domain')
            self.assertEqual(sum([(domain == i) for i in response.get('Identities', [])]), 1)
                
```

Save as `test.py` and run with `python3 -m unittest test.py`

# Expected behavior

The test passes.

i.e. when `verify_domain_dkim` is called, it checks whether the domain already exists in the list,
and doesn't re-add it as a duplicate.

To confirm the expected behavior, comment out `with mock_ses()` and run against real AWS.

# Actual behavior

The script fails. The final assertion fails.

`verify_domain_dkim` seems to have created a second domain, without checking whether it's already added.

bpandola:
Verified test case against a real AWS backend.

Fixes #3947

### DIFF
--- a/moto/ses/models.py
+++ b/moto/ses/models.py
@@ -137,7 +137,8 @@ class SESBackend(BaseBackend):
         self.email_addresses.append(address)
 
     def verify_domain(self, domain):
-        self.domains.append(domain)
+        if domain.lower() not in self.domains:
+            self.domains.append(domain.lower())
 
     def list_identities(self):
         return self.domains + self.addresses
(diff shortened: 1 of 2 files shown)


bblommers (approved):
Thanks @bpandola!

PR #4055, opened 2021-07-03
sns#create_platform_endpoint: If token and attributes are the same, return endpoint

kohbis:
#4056

### DIFF
--- a/moto/sns/models.py
+++ b/moto/sns/models.py
@@ -584,10 +584,11 @@ class SNSBackend(BaseBackend):
     def create_platform_endpoint(
         self, region, application, custom_user_data, token, attributes
     ):
-        if any(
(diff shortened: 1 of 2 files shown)


bblommers (changes requested):
Hi @kohbis, thanks for the PR! Would you mind adding a test for this as well, in tests/test_sns/test_application_boto3.py?

There is already a test to verify the duplicate exception, so that might need adjusting as well

kohbis:
Hi @bblommers, thanks for confirming the PR. I've added & updated tests.
https://github.com/spulec/moto/pull/4055/commits/5d49275791978177126dcc4fea702aab04d62660
I was wondering if you could check them !

bblommers (approved):
LGTM - thanks for the PR @kohbis!

kohbis:
thanks for review&merge!

Issue #2544, opened 2019-11-07
While creating a alb listener rule for the http-request-method through moto, I am getting ""An error occurred (ValidationError) when calling the CreateRule operation: Condition field 'http-request-method' must be one of '[path-pattern, host-header]" Exception.


satishgampala:
While creating a alb listener rule for the http-request-method through moto, 

I am getting "An error occurred (ValidationError) when calling the CreateRule operation: Condition field 'http-request-method' must be one of '[path-pattern, host-header]" Exception.

Sample code snippet:

```
import boto3
from moto import mock_elbv2

alb_client = boto3.client('elbv2', region_name = 'eu-central-1')


@mock_elbv2
def create_alb_listener_rule():
    alb_client.create_rule(
        Priority=10,
        ListenerArn='string',
        Conditions=[
            {
                "Field": "HttpRequestMethodConfig",
                "HttpRequestMethodConfig": {
                    "Values": [
                        "POST"
                    ]
                }
            }
        ],
        Actions=[
            {
                'Type': 'forward',
                'TargetGroupArn': "XXXXXXXXXXXXXXXXXXX"
            }
        ]
    )
```

Exception:

```
"An error occurred (ValidationError) when calling the CreateRule operation: Condition field 'http-request-method' must be one of '[path-pattern, host-header]"
```

Is moto comaptible with the latest boto3 versions???

satishgampala:
Even after trying with "Field": "http-request-method", then also it fails

spulec:
Thanks for opening. Setting this as a feature request for us to implement `http-request-method`

tabramczyk:
I have same problem with `"Field": "http-header"`.

There are more options now:
> `The field in the HTTP request. The following are the possible values:
>
> - http-header
> - http-request-method
> - host-header
> - path-pattern
> - query-string
> - source-ip`

Source:
https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/elbv2.html#ElasticLoadBalancingv2.Client.create_rule

bblommers:
This is now fixed in #4066

PR #4066, opened 2021-07-12
Support all ELBv2 ListenerRule condition types

eoconnell:
This PR adds support for the following condition types:
- http-header
- http-request-method
- host-header
- path-pattern
- query-string
- source-ip

Addresses #2544

### DIFF
--- a/moto/core/responses.py
+++ b/moto/core/responses.py
@@ -557,6 +557,81 @@ class BaseResponse(_TemplateEnvironmentMixin, ActionAuthenticatorMixin):
                 ]
         return params
 
+    def _get_params(self):
+        """
+        Given a querystring of
+        {
+            'Action': ['CreatRule'],
+            'Conditions.member.1.Field': ['http-header'],
+            'Conditions.member.1.HttpHeaderConfig.HttpHeaderName': ['User-Agent'],
+            'Conditions.member.1.HttpHeaderConfig.Values.member.1': ['curl'],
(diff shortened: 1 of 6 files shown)


bblommers (changes requested):
Hi @eoconnell, thanks for the PR!
Can we parametrize some of the tests? It looks like the create/modify tests all follow the same pattern, just with different conditions. It would be cleaner if we can combine these tests like so:

```
@pytest.mark.parametrize(
    "condition",
    [
        {"Field": "path-pattern", "Values": ["/home"]},
        ..other-path-pattern,
        ..http-header-condition,
        ..etc..
    ],
)
def test_create_rule_with_conditions(condition):
     ...
    conn.create_rule(
        Conditions=[condition]
     )
    ....
    rule["Conditions"].should.equal([condition])
```

eoconnell:
Hi @bblommers, that's a good idea. Let me know what you think of the latest commit. The validation scenarios aren't quite as uniform, but maybe they are more thorough than they need to be. Do you think those are fine the way they are?

eoconnell:
On second thought, I can probably parametrize those as well.

bblommers:
Awesome - thanks for making the change @eoconnell, that looks very clean like this

bblommers (approved):

eoconnell:
Thanks for the suggestion and for taking the time to review.

PR #4079, opened 2021-07-23
Fix security group filters

jl2501:
Security group filters are pretty simple, but the current implementation causes a stack trace when using them in many cases and makes some assumptions about how to map the filter names to attributes that is unfortunately incorrect in many cases.

This approach is based on the idea that in an open source testing project like moto, the most important things are that:
1) it works
2) it is easy for people to extend and understand so more people can contribute

so, I keep every filter as its own separate function and have a dictionary to map the filter's _full_ name to a specific function.
there's plenty of repetition here, but it avoids having convoluted code that people have to really think about before they understand it and can add their own filters / modify a single filter w/out affecting all the rest of them.


bblommers (changes requested):
Hi @jl2501, thanks for the PR!

Looks like not all of the new changes are covered by unit tests yet - would you mind extending them to cover all scenarios?
(Based on the coverage report here: https://codecov.io/gh/spulec/moto/compare/85dc52bd84e505b362853887f6a5f54b762662a7...16169d4962675fa830a7d28d0b538f74532fa2dd/diff)

bblommers:
There's been a few changes to `ec2/models.py` in the last days @jl2501 - you might have to rebase/merge with the latest master to get the tests to run.

jl2501:
sorry for the extended delays on this

jl2501:
forgot to blacken those two files earlier. oops. done now.

bblommers:
> sorry for the extended delays on this

All good @jl2501. Looks like some of the tests are failing - let me know if you need any help with them

bblommers (approved):
Nice one - thanks for the PR @jl2501!

jl2501:
very welcome. thanks for your patience!

Issue #4010, opened 2021-06-15
Build support for EFS functionality


pagreene:
Our architecture depends heavily on EFS, which has relatively few actions to implement. I would like to propose contributing an implementation of support for EFS, submitting this issue as per: https://github.com/spulec/moto/blob/master/CONTRIBUTING.md.

bblommers:
Hi @pagreene, welcome to Moto! PR's are always welcome.

There's a section in the documentation [here](https://github.com/spulec/moto/blob/master/CONTRIBUTING.md#missing-services) with some guidelines on how to get started. Every service is different, so some investigation might still be necessary, but it can at least get you started.

Let me know if you run into any issues -  happy to help.

bblommers:
I'll close this issue, as the initial EFS implementation has now been released.
Feel free to open additional issues/PR's for any other services/methods that are missing.

pagreene:
Sounds good, thanks!

PR #4080, opened 2021-07-24
Add partial support for EFS

pagreene:
This PR adds support for the following actions in EFS, generally centered around file systems and mount targets (_not_ access points):
- create_file_system
- create_mount_target
- delete_file_system
- delete_mount_target
- describe_backup_policy
- describe_file_systems
- describe_mount_target

In addition I have implemented tests for both good and bad behavior. As much as possible an error in usage should produce the same code, error type, and message as `boto3`, and tests are an attempt to enforce this.

Some minor tweaks were also made to the cloudformation module to better match behavior from `boto3`.

This PR is the completion of the work proposed in https://github.com/spulec/moto/issues/4010


moto/efs/models.py
```diff
@@ -0,0 +1,542 @@
...
+
+from boto3 import Session
+
+from moto.core import ACCOUNT_ID, BaseBackend, CloudFormationModel
+from moto.core.utils import (
+    camelcase_to_underscores,
+    get_random_hex,
+    underscores_to_camelcase,
+)
+from moto.ec2 import ec2_backends
```
bblommers, line 22: As of 2.x, Moto supports installation of only specific modules (`pip install moto[efs]`). At the moment that would fail, as EFS does not specify additional dependencies, but EC2 does - so importing any EC2 model will fail with a `module not found` exception.
As far as I can tell, we only need the EC2 reference in create_mount_target/delete_mount_target. Can we lower the import inside those methods?
That way people will be able to install EFS alone without needing any additional dependencies - only when calling those two methods will they receive the notification that EC2 dependencies are not installed.

tests/test_efs/test_file_system.py
```diff
@@ -0,0 +1,384 @@
...
+    # Check the response contents.
+    assert set(resp3.keys()) == {"FileSystems", "Marker"}
+    assert len(resp3["FileSystems"]) == 2
+    assert resp3["Marker"] == resp2["NextMarker"]
+    fs_id_set_3 = {fs["FileSystemId"] for fs in resp3["FileSystems"]}
+    assert fs_id_set_3 & (fs_id_set_1 | fs_id_set_2) == set()
+
+
+def test_describe_file_systems_invalid_marker(efs):
+    try:
```
bblommers, line 316: Moto tends to use pytest.raises to verify error scenarios, which is a bit cleaner IMO:
```
    ..
err["Message"].should.equal(..)
err["Code"].should.equal(..)
```

This should be applied for all every negative test.

bblommers (changes requested):
Hi @pagreene, thanks for the PR! Looks good overall - just left a few comments.

Note that the linter is failing at the moment - you can use `make format` to fix this.
Make sure that you have right version of black installed (different versions will produce different results): `black==19.10b0`

moto/efs/models.py
pagreene: > Hi @pagreene, thanks for the PR! Looks good overall - just left a few comments.
> 
> Note that the linter is failing at the moment - you can use `make format` to fix this.
> Make sure that you have right version of black installed (different versions will produce different results): `black==19.10b0`

Looks like I just forgot to lint a file in cloudformation. That should be fixed now.

pagreene: > As of 2.x, Moto supports installation of only specific modules (`pip install moto[efs]`). At the moment that would fail, as EFS does not specify additional dependencies, but EC2 does - so importing any EC2 model will fail with a `module not found` exception.
> As far as I can tell, we only need the EC2 reference in create_mount_target/delete_mount_target. Can we lower the import inside those methods?
> That way people will be able to install EFS alone without needing any additional dependencies - only when calling those two methods will they receive the notification that EC2 dependencies are not installed.

Unfortunately it is also used (currently) whenever a FileSystem is created, specifically to look up the availability zone ID given an availability zone name. Is there a way to do this without using ec2's `describe_availability_zones`?

tests/test_efs/test_file_system.py
pagreene: > Moto tends to use pytest.raises to verify error scenarios, which is a bit cleaner IMO:
> 
> ```
> with pytest.raises(ClientError) as ex:
>     ..
> err = ex.value.response["Error"]
> err["Message"].should.equal(..)
> err["Code"].should.equal(..)
> ```
> 
> This should be applied for all every negative test.

Alright, this should be all set now. Thanks for the pointer!

moto/efs/models.py
bblommers: Ah, right. Don't think there's another way to to this at at the moment, no.
Looking into to it a bit more, we'll need to refactor the dependencies soon anyway. So let's leave this as is for now - I'll find a better way of doing this later on 👍

pagreene:
Looks like there might be an issue with the version of `ipaddress` causing certain tests to fail.

bblommers:
The `subnet_of` method is only available as of Py3.7, unfortunately: https://docs.python.org/3/library/ipaddress.html#ipaddress.IPv4Network.subnet_of - we still support 3.6

pagreene:
> The `subnet_of` method is only available as of Py3.7, unfortunately: https://docs.python.org/3/library/ipaddress.html#ipaddress.IPv4Network.subnet_of - we still support 3.6

Ah, fiddlesticks. I will change that to work with older versions.

bblommers:
Thanks @pagreene. 
The reason the ServerMode tests are failing at the moment is probably that we're intercepting the correct URL's. In urls.py, the list might be more explicit:
```
url_paths = {
    "{0}/filesystem$": response.dispatch,
    "{0}/mount_target$": response.dispatch,
}
```
This is just an example - it will depend on the exact URLs used by AWS. Let me know if you need some help debugging this.

pagreene:
> Thanks @pagreene.
> The reason the ServerMode tests are failing at the moment is probably that we're intercepting the correct URL's. In urls.py, the list might be more explicit:
> 
> ```
> url_paths = {
>     "{0}/filesystem$": response.dispatch,
>     "{0}/mount_target$": response.dispatch,
> }
> ```
> 
> This is just an example - it will depend on the exact URLs used by AWS. Let me know if you need some help debugging this.

Alright, I think that should now be fixed. The issues ultimately seems to be that the URLs prepend 2015-02-01 for no apparent reason. :roll_eyes: :laughing:

bblommers:
> 
> 
> Alright, I think that should now be fixed. The issues ultimately seems to be that the URLs prepend 2015-02-01 for no apparent reason. 🙄 😆

Yeah, that would be the API version for EFS. It's different for every module, just to make things easier for us. Good find 🙂 

> 
> EDIT: to clarify, I wrote some server tests that are now passing, which I am hoping will come to the same thing as tests passing in "server mode".

Looks like that fixed most of the ServerMode tests, yes.
The server tests are currently mixing local and servermode data, though - that's why those are now failing.
For example, this is roughly what happens when running `test_efs_file_system_delete` in our CI:
 - CI starts a moto server
 - The test creates a file system on that server (`efs.create_file_system(CreationToken="foobarbaz", Backup=True)`)
 - The test starts a in-memory MotoServer (`backend = server.create_backend_app("efs")`)
 - The test then tries to delete the filesystem from the in-memory MotoServer - which it obviously can't find

So to fix these tests, the `file_system_id`-fixture would have to create the file-system in the in-memory MotoServer, instead of makign the boto3-call. Hope that makes sense..

Same goes for the subnet creation. To keep things simple though, I'd be OK with dropping the tests to require a subnet - that behaviour is already tested thoroughly in the standard boto3-tests.

pagreene:
> Looks like that fixed most of the ServerMode tests, yes.
> The server tests are currently mixing local and servermode data, though - that's why those are now failing.
> For example, this is roughly what happens when running `test_efs_file_system_delete` in our CI:


Yeah, I noticed that yesterday afternoon. I am nearly done fixing it, and the fix should be in momentarily.

pagreene:
That should now be all set. Hopefully no more bugs crawl out :crossed_fingers:

bblommers (approved):
LGTM @pagreene - thanks again!

pagreene:
Hi @bblommers, just a note that somehow your way of merging PRs wipes out all the commits made within the PR. This is generally considered poor coding practice as it removes an immense amount of documentation in the form of commit messages. It also significantly obscures credit for work done on the repository by other developers; as far as GitHub is concerned I have only contributed one commit over the last couple of weeks, rather than 62.

Thanks again for your work reviewing everything, and generally maintaining `moto`. :)

bblommers:
Correct, @pagreene, I tend to 'Squash and Merge'. I agree with your arguments in general - it is good form and good documentation to keep a history. 
For this project it is a double-edged sword though, as there are so many PR's/commits that finding a specific change is often difficult. So by 'dumbing' it down to a single commit per PR, the history becomes better readable/searchable.

PR #4082, opened 2021-07-24
Merge LocalStack changes into upstream moto

whummer:
Hi folks! 👋  This PR merges the changes from [LocalStack's fork](https://github.com/localstack/moto) into upstream `moto`. As discussed with @spulec, @bblommers and @thrau, we are going to collaborate more closely going forward, making sure the code bases are kept in sync (and avoiding a hard-fork situation where things diverge too much!).

Some of the changes in this PR stem from a [new test suite](https://github.com/localstack/localstack-terraform-test) we have recently introduced - running the Terraform AWS provider tests against LocalStack (and moto). Going forward, we're thinking of establishing a process where we can add new features into upstream `moto` as long as they are covered by a TF test passing (evidence to be provided, for example, by a screenshot like the one [here](https://github.com/localstack/moto/pull/38)). (This is to be confirmed by @bblommers @spulec - please chime in here!)

The changes have been thoroughly sanity checked, and all non-moto relevant changes (e.g., some LocalStack-specific minor adjustments) have been removed during preparation of this PR.

There is one key question remaining - whether to **drop Python 2.7 support** (which we've decided to do in LocalStack). In case you feel strongly about keeping 2.7 compatibility, we'd have to make a few more adjustments in the PR (but we'd prefer dropping it, given that Python 2 EOL happened more than 1.5 years ago :) ).

Looking forward to getting your feedback! 🚀 Thanks


moto/cloudwatch/models.py
```diff
@@ -178,16 +183,19 @@ def __init__(self, namespace, name, value, dimensions, timestamp, unit=None):
...
             return False
         if name and name != self.name:
             return False
-        for metric in already_present_metrics:
-            if self.dimensions and are_dimensions_same(
-                metric.dimensions, self.dimensions
-            ):
-                return False
+
+        # Note: The logic below appears to be incorrect. We should not skip adding new Metrics
```
bblommers, line 192: What is the reason behind the removal of this logic?
It was added to fix this issue: https://github.com/localstack/localstack/issues/3222

Note that in AWS itself, this scenario would never occur - as the put_metric_data() call fails when trying to store a duplicate metric.
However, as we (moto) do not have this validation yet, we do have to deal with potentially dirty data. So it makes sense IMO to at least keep the outgoing data consistent with what AWS would return

whummer: Good catch, thanks @bblommers . This was part of a general overhaul of the CW metrics logic - we've been working with @usmangani1 (he's part of our team) on [this PR](https://github.com/spulec/moto/pull/3461) that you've kindly helped review, and later discovered that a few more changes were needed.

Just double-checked against the latest version of LocalStack (which contains the code in this PR), and the behavior reported by the user in localstack/localstack#3222 still seems to be accurate:

```
$ awslocal cloudwatch list-metrics --namespace AWS/EC2 --dimensions Name=InstanceId,Value=i-0bada24f4d28a7793
{
    "Metrics": []
}
$ awslocal cloudwatch put-metric-data --namespace AWS/EC2 --metric-data '[{ "MetricName": "CPUUtilization", "Dimensions": [ { "Name": "InstanceId", "Value": "i-0bada24f4d28a7793" } ], "Value": 28, "Unit": "Percent" }]'
$ awslocal cloudwatch list-metrics --namespace AWS/EC2 --dimensions Name=InstanceId,Value=i-0bada24f4d28a7793
{
    "Metrics": [
        {
            "Namespace": "AWS/EC2",
            "MetricName": "CPUUtilization",
            "Dimensions": [
                {
                    "Name": "InstanceId",
                    "Value": "i-0bada24f4d28a7793"
                }
            ]
        }
    ]
}
$ awslocal cloudwatch put-metric-data --namespace AWS/EC2 --metric-data '[{ "MetricName": "CPUUtilization", "Dimensions": [ { "Name": "InstanceId", "Value": "i-0bada24f4d28a7793" } ], "Value": 28, "Unit": "Percent" }]'
$ awslocal cloudwatch list-metrics --namespace AWS/EC2 --dimensions Name=InstanceId,Value=i-0bada24f4d28a7793
{
    "Metrics": [
        {...}  # same as above ...
    ]
}
```

Hope that makes sense. Happy to discuss more and/or work with you on adding more tests to cover different edge cases for CloudWatch Metrics (as well as adding any missing validation) post merge. 👍

bblommers: Hmm. I wonder if the AWS behavior changed then recently, because the error that duplicate dimensions are invalid definitely didn't show up when I was testing this exact scenario in #3461

whummer: Good catch - commented the code back in and re-enabled the disabled test (also added a new test to cover get_metric_statistics using metrics with different dimensions).

bblommers (approved):

PR #4087, opened 2021-07-26
Fix: Adding empty tags to SQS queue on creation time

mgagliardo:
Fixes https://github.com/localstack/localstack/issues/4343, it seems like newly created queues do not add tags if there is an empty among them. This behavior works in AWS but not in moto and is fixed by retrieving the missing params instead of passing `self.tags`


whummer:
LGTM, thanks @mgagliardo 👍 . Awaiting review from @bblommers ...

bblommers (approved):
LGTM!

PR #4089, opened 2021-07-26
Add support for EC2 transit gateway associations

whummer:
Add support for EC2 transit gateway associations. [Migrating PR from fork repo](https://github.com/localstack/moto/pull/33) after recent merge (fixing merge conflicts). 

Screenshot of passed Terraform tests following soon... /cc @macnev2013


bblommers:
Hi @whummer, thanks for the PR!
Instead of the screenshot, can we extend the whitelist of TF tests please: https://github.com/spulec/moto/blob/master/tests/terraform-tests.success.txt
(And/or blacklist any tests here, if you want greater control over which tests to run: https://github.com/spulec/moto/blob/master/tests/terraform-tests.failures.txt)

macnev2013:
Hi @bblommers. Thanks for pointing it out. I'll update the PR with minor few bugfix and add the tf test names as you mentioned. Thanks.

kbalk:
I've been watching this PR with interest because I was getting a segmentation violation error when using moto and a Terraform script containing transit gateway resources.  The same Terraform script works when using AWS itself.

I pulled this PR today and tried the Terraform script again and I'm no longer getting a segment violation, but I'm getting an error.  Would you be interested in trying out this script or seeing the error?  The script creates a transit gateway, route table, route and 2 VPCs; it's a minimized version of something we use in production.

I could wait until this PR is finished and if the problem still exist then report it, but I thought it might make a useful test case.

kbalk:
I am so sorry ... I was using localstack, not the moto server when I ran the Terraform script.  I'm not getting any error from moto when I run the Terraform "apply", but it is having trouble with the "destroy".  It can't seem to destroy the aws_route_table.

I really appreciate the work you're doing and I apologize again for the erroneous comment I made previously.

macnev2013:
Hi @kbalk, thanks for the feedback. We would be happy to fix the bug. Can you please share the script?

kbalk:
The tar file includes a README to explain the files.
[tgw.tar.gz](https://github.com/spulec/moto/files/6904992/tgw.tar.gz)

whummer:
@bblommers - I think this PR should now be ready to review/merge. Please let us know if there are any questions. Thanks!

macnev2013:
Hi @kbalk , Thanks for sharing the script. Looks like the issue is not related to TransitGateway. We will fix this in upcoming PR. Thanks

bblommers (approved):
LGTM! Thanks @whummer @macnev2013 @monty16597

PR #4091, opened 2021-07-26
Fix: Adding fix when setting empty policy attr to queue

mgagliardo:
Fixes https://github.com/localstack/localstack/issues/4225

Fixes an issue with the terraform AWS provider that sets the SQS policy to an empty string, as seen in [the following function](https://github.com/hashicorp/terraform-provider-aws/blob/ccc13f55709c860a2130ecee6e21acde2318e936/aws/resource_aws_sqs_queue_policy.go#L104) via query parameters in the URL, for example:

`Action=SetQueueAttributes&Attribute.1.Name=Policy&Attribute.1.Value=&QueueUrl=http%3A%2F%2Flocalhost%3A4566%2F000000000000%2Fdistributor_queue&Version=2012-11-05`

In the end the `set_queue_attributes` receives the following:
```
odict_items([('Action', ['SetQueueAttributes']), ('Attribute.1.Name', ['Policy']), ('Attribute.1.Value', ['']), ('QueueUrl', ['http://localhost:4566/000000000000/distributor_queue']), ('Version', ['2012-11-05'])])
```

### DIFF
--- a/moto/sqs/responses.py
+++ b/moto/sqs/responses.py
@@ -202,8 +202,17 @@ class SQSResponse(BaseResponse):
 
     def set_queue_attributes(self):
         # TODO validate self.get_param('QueueUrl')
+        attribute = self.attribute
+
+        # Fixes issue with Policy set to empty str
+        attribute_names = self._get_multi_param("Attribute")
+        if attribute_names:
+            for attr in attribute_names:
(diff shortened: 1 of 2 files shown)


bblommers (approved):
LGTM - thanks for the PR @mgagliardo!

PR #4092, opened 2021-07-27
Add support for admin_user_global_sign_out to cognitoidp

andyfase:
This PR adds support for admin_user_global_sign_out for Cognito. 

[admin_user_global_sign_out](https://docs.aws.amazon.com/cli/latest/reference/cognito-idp/admin-user-global-sign-out.html) requires that refresh tokens are invalidated - hence the code removes any user specific refresh tokens. 

I added a check to `initiate_auth` to confirm if the refresh token exists - if it doesnt (which is shouldn't if `admin_user_global_sign_out` has been called) it responds with a ResourceNotFoundException

I've added a test case to cover this code and have run all cognitoidp test cases to confirm the small change to `initiate_auth` has not broken anything.

### DIFF
--- a/moto/cognitoidp/models.py
+++ b/moto/cognitoidp/models.py
@@ -900,6 +900,19 @@ class CognitoIdpBackend(BaseBackend):
         user = user_pool.users[username]
         user.update_attributes(attributes)
 
+    def admin_user_global_sign_out(self, user_pool_id, username):
+        user_pool = self.user_pools.get(user_pool_id)
+        if not user_pool:
+            raise ResourceNotFoundError(user_pool_id)
+
+        if username not in user_pool.users:
+            raise UserNotFoundError(username)
+
+        for token, token_tuple in list(user_pool.refresh_tokens.items()):
+            _, username = token_tuple
+            if username == username:
+                del user_pool.refresh_tokens[token]
+
     def create_resource_server(self, user_pool_id, identifier, name, scopes):
         user_pool = self.user_pools.get(user_pool_id)
         if not user_pool:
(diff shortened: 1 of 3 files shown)


andyfase:
I wonder if we should instead raise `NotAuthorizedException` for this use-case happy to change it if thats the agreed on Exception

bblommers:
Hi @andyfase, thanks for the PR!
> 
> 
> I wonder if we should instead raise `NotAuthorizedException` for this use-case happy to change it if thats the agreed on Exception

Ideally we use whatever name AWS also uses.

All exceptions are converted into CLientErrors though, so in other places we tend to use this format:
```
    ....
err["Message"].should.equal("whatever AWS returns..")
err["Code"].should.equal("whatever AWS returns..")
```

moto/cognitoidp/models.py
```diff
@@ -900,6 +900,19 @@ def admin_update_user_attributes(self, user_pool_id, username, attributes):
         user = user_pool.users[username]
         user.update_attributes(attributes)
 
+    def admin_user_global_sign_out(self, user_pool_id, username):
+        user_pool = self.user_pools.get(user_pool_id)
+        if not user_pool:
+            raise ResourceNotFoundError(user_pool_id)
```
bblommers, line 906: Can we get unit tests for these two exceptions as well, please?

bblommers (changes requested):

moto/cognitoidp/models.py
andyfase: No Problem - additional tests committed

andyfase:
Hey @bblommers I believe I've made the changes requested. 

Note for 
```
err["Message"].should.equal("whatever AWS returns..")
```

The Cognito model code does not generally raise Exceptions with messages defined from what i can tell. I've done it explicitly for the case when a refresh token has been invalidated (aligning with aws boto response) but the other exceptions are inline with the rest of the code where the message itself is not raised explicitly - and not checked in the unit tests.

bblommers (approved):
Awesome - thanks for making the changes @andyfase!

PR #4096, opened 2021-07-27
include associationState when DescribeRouteTables

jrauen:
The AWS API documentation specifies that the [association state](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_RouteTableAssociationState.html) is a part of the [DescribeRouteTables](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_DescribeRouteTables.html) response object. Although it's not a required attribute of the object, it is consistently provided by the API and the Terraform AWS provider expects it to be present. Without including it in the response the provider throws a null pointer reference and testing can fail.


bblommers (approved):
LGTM - thanks for the PR @jrauen!

PR #4106, opened 2021-07-29
Fix EKS list models bug

ferruzzi:
Further testing against the actual API shows that the final state of `nextToken` is `None`, not `"null"`.  The meat of the PR is that change on L481.

 - Support for Fargate Profiles is being added in a coming PR.  `list_fargate_profiles` will reuse the same logic as `list_clusters` and `list_nodegroups`.  This bug is in the affected overlapping code, so that logic has been moved into a helper method as part of the correction.


bblommers (approved):
LGTM - thanks @ferruzzi!

PR #4110, opened 2021-07-29
Fix regex URLs so that a dot is actually interpreted as a dot

### LINKED ISSUE #4099: URL Bases regular expressions not escaping dots
All url bases patterns defined per services don't escape dots which brings to wrong backend identification when moto tries to match urls against patterns.
For example, if you mock both S3 and IoT services and you try to create a bucket starting with "iot" it will fail as it will match with the IoT Service url base.

Here is how to reproduce the issue.
``` python
import boto3
from moto import mock_iot, mock_s3


class TestMockS3:
    @mock_iot
    @mock_s3
    def test_s3_mock(self):
        s3_client = boto3.client("s3", "eu-west-1")
        s3_client.create_bucket(
            ACL="private",
            Bucket="iot-bucket",
            CreateBucketConfiguration={"LocationConstraint": "eu-west-1"},
        )
```
Following error is raised:
**botocore.exceptions.ClientError: An error occurred (404) when calling the CreateBucket operation: Not Found**

In urls.py of IoT service:
`url_bases = ["https?://iot.(.+).amazonaws.com"]` matches with https://iot-bucket.s3.eu-west-1.amazonaws.com
It should be:
`url_bases = ["https?://iot\.(.+)\.amazonaws\.com"]`

The same regexp mistake in url_bases is applicable to all other services.

bblommers:
FIxes #4099

This PR partially fixes a bug so we can create S3 buckets that start with a service name, i.e. `iot-bucket`. 

Note that it is still not possible to create a S3 bucket that start with `{service}.`, such as `iot.bucket` - this will still be interpreted as a call to the IOT-module. This will be fixed with #3793

### DIFF
--- a/moto/datasync/urls.py
+++ b/moto/datasync/urls.py
@@ -2,6 +2,6 @@ from __future__ import unicode_literals
 
 from .responses import DataSyncResponse
 
-url_bases = ["https?://(.*?)(datasync)(.*?).amazonaws.com"]
+url_bases = [r"https?://(.*?)(datasync)\.(.*?).amazonaws.com"]
 
 url_paths = {"{0}/$": DataSyncResponse.dispatch}
--- a/moto/datasync/urls.py
+++ b/moto/datasync/urls.py
@@ -2,6 +2,6 @@ from __future__ import unicode_literals
 
 from .responses import DataSyncResponse
 
-url_bases = [r"https?://(.*?)(datasync)\.(.*?).amazonaws.com"]
+url_bases = [r"https?://(.*\.)?(datasync)\.(.+).amazonaws.com"]
 
 url_paths = {"{0}/$": DataSyncResponse.dispatch}
--- a/moto/glue/urls.py
+++ b/moto/glue/urls.py
@@ -2,6 +2,6 @@ from __future__ import unicode_literals
 
 from .responses import GlueResponse
 
-url_bases = ["https?://glue(.*).amazonaws.com"]
+url_bases = [r"https?://glue\.(.*)\.amazonaws\.com"]
 
 url_paths = {"{0}/$": GlueResponse.dispatch}
--- a/moto/glue/urls.py
+++ b/moto/glue/urls.py
@@ -2,6 +2,6 @@ from __future__ import unicode_literals
 
 from .responses import GlueResponse
 
-url_bases = [r"https?://glue\.(.*)\.amazonaws\.com"]
+url_bases = [r"https?://glue\.(.+)\.amazonaws\.com"]
 
 url_paths = {"{0}/$": GlueResponse.dispatch}
--- a/moto/iam/urls.py
+++ b/moto/iam/urls.py
@@ -1,6 +1,6 @@
 from __future__ import unicode_literals
 from .responses import IamResponse
 
-url_bases = [r"https?://iam\.(.*)amazonaws\.com"]
+url_bases = [r"https?://iam\.(.*\.)?amazonaws\.com"]
 
 url_paths = {"{0}/$": IamResponse.dispatch}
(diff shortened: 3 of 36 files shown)


moto/datasync/urls.py
```diff
@@ -2,6 +2,6 @@
 
 from .responses import DataSyncResponse
 
-url_bases = ["https?://(.*?)(datasync)(.*?).amazonaws.com"]
+url_bases = [r"https?://(.*?)(datasync)\.(.*?).amazonaws.com"]
```
Cox65, line 5: It should be :
`url_bases = [r"https?://(.*\.)?(datasync)\.(.+).amazonaws.com"]` no?
I assume we want to include subdomains of datasync and not something ending with datasync?
And about the region, we should use (.+) as other services to ensure there is something between both dots.
 I'm not really familiar with this AWS service.

moto/sqs/urls.py
```diff
@@ -1,7 +1,7 @@
 from __future__ import unicode_literals
 from .responses import SQSResponse
 
-url_bases = ["https?://(.*?)(queue|sqs)(.*?).amazonaws.com"]
+url_bases = [r"https?://(.*?)(queue|sqs)\.(.*?)amazonaws\.com"]
```
Cox65, line 4: Same comment here regarding subdomain inclusion.
One dot was missing as well before amazonaws.
`url_bases = [r"https?://(.*\.)?(queue|sqs)\.(.*\.)?amazonaws\.com"]`

Tested against:
https://sqs.eu-west-1.amazonaws.com
https://queue.amazonaws.com

moto/sts/urls.py
```diff
@@ -1,6 +1,6 @@
 from __future__ import unicode_literals
 from .responses import TokenResponse
 
-url_bases = ["https?://sts(.*).amazonaws.com(|.cn)"]
+url_bases = [r"https?://sts\.(.*)amazonaws\.com(|.cn)"]
```
Cox65, line 4: `url_bases = [r"https?://sts\.(.*\.)?amazonaws\.com(|.cn)"]`

moto/iam/urls.py
```diff
@@ -1,6 +1,6 @@
 from __future__ import unicode_literals
 from .responses import IamResponse
 
-url_bases = ["https?://iam(.*).amazonaws.com"]
+url_bases = [r"https?://iam\.(.*)amazonaws\.com"]
```
Cox65, line 4: `url_bases = [r"https?://iam\.(.*\.)?amazonaws\.com"]`

if we want to include:
https://iam.amazonaws.com
https://iam.anything.amazonaws.com (in which case is it necessary?)

moto/glue/urls.py
```diff
@@ -2,6 +2,6 @@
 
 from .responses import GlueResponse
 
-url_bases = ["https?://glue(.*).amazonaws.com"]
+url_bases = [r"https?://glue\.(.*)\.amazonaws\.com"]
```
Cox65, line 5: `url_bases = [r"https?://glue\.(.+)\.amazonaws\.com"]`

If region is mandatory for Glue, it should be a plus not a star.

Cox65 (commented):
Some proposals to improve regular expressions for some services.

moto/iam/urls.py
bblommers: IAM is a global service, so boto3 sends everything to `iam.amazonaws.com` AFAIK. I'll leave the URL open for other regions though (`iam.us-east-1.amazonaws.com`), just in case other SDK's behave differently.

bblommers:
Thanks for the review @Cox65 - I've updated the URL's accordingly

moto/datasync/urls.py
bblommers: Not too familiar with it either, so I'm not going to make too many assumptions about the URL formats :)

Cox65 (approved):

PR #4116, opened 2021-08-01
fixed vpc with terraform; added modify vpc tenancy

macnev2013:
### Fixed
- VPC was missing `ownerId` attribute which caused terraform to calculate wrong ARN

### Added
- Added ModifyVpcTenancy API

### TF Test
- All 21 cases under TestAccAWSVpc_ are working now

### DIFF
--- a/moto/ec2/exceptions.py
+++ b/moto/ec2/exceptions.py
@@ -481,6 +481,14 @@ class CidrLimitExceeded(EC2ClientError):
         )
 
 
+class UnsupportedTenancy(EC2ClientError):
+    def __init__(self, tenancy):
+        super(UnsupportedTenancy, self).__init__(
+            "(UnsupportedTenancy",
+            "The tenancy value {0} is not supported.".format(tenancy),
+        )
+
+
 class OperationNotPermitted(EC2ClientError):
     def __init__(self, association_id):
         super(OperationNotPermitted, self).__init__(
--- a/moto/ec2/exceptions.py
+++ b/moto/ec2/exceptions.py
@@ -484,7 +484,7 @@ class CidrLimitExceeded(EC2ClientError):
 class UnsupportedTenancy(EC2ClientError):
     def __init__(self, tenancy):
         super(UnsupportedTenancy, self).__init__(
-            "(UnsupportedTenancy",
+            "UnsupportedTenancy",
             "The tenancy value {0} is not supported.".format(tenancy),
         )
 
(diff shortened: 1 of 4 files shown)


macnev2013:
Hi @bblommers , @whummer - PR is ready to merge. Can you please review it?

moto/ec2/models.py
```diff
@@ -3011,11 +3016,26 @@ def get_filter_value(self, filter_name):
...
+        else:
+            raise UnsupportedTenancy(tenancy)
+        return True
+
     def associate_vpc_cidr_block(
         self, cidr_block, amazon_provided_ipv6_cidr_block=False
     ):
         max_associations = 5 if not amazon_provided_ipv6_cidr_block else 1
 
+        for cidr in self.cidr_block_association_set.copy():
```
whummer, line 3031: Note for @bblommers : This is done to prevent error messages where unused association sets (in state `disassociated`) exist and take up all slots. We've verified this behavior against real AWS. For example, if 1 IPv6 CIDR block association already exists in state `disassociated`, we should still allow associating a new CIDR block here, hence `pop`'ing it from line 3038 below.

moto/ec2/exceptions.py
```diff
@@ -481,6 +481,14 @@ def __init__(self, vpc_id, max_cidr_limit):
         )
 
 
+class UnsupportedTenancy(EC2ClientError):
+    def __init__(self, tenancy):
+        super(UnsupportedTenancy, self).__init__(
+            "(UnsupportedTenancy",
```
whummer, line 487: nitpick: I'm assuming the error type should be changed to `"UnsupportedTenancy"` (without the leading bracket `(`).

moto/ec2/models.py
```diff
@@ -3011,11 +3016,26 @@ def get_filter_value(self, filter_name):
         else:
             return super(VPC, self).get_filter_value(filter_name, "DescribeVpcs")
 
+    def modify_vpc_tenancy(self, tenancy):
```
whummer, line 3019: nitpick: Could be slightly rewritten (to make it more intuitive, checking the parameter assertions early in the method) to:
```
    def modify_vpc_tenancy(self, tenancy):
        if tenancy != "default":
            raise UnsupportedTenancy(tenancy)
        self.instance_tenancy = tenancy
        return True
```

whummer (commented):
LGTM, thanks @macnev2013 . 👍 Only a few minor nitpicks.. Thanks!

macnev2013:
@whummer , I have updated the code.

bblommers (approved):
LGTM! Thanks @macnev2013 and @whummer

PR #4117, opened 2021-08-01
Add support for ConfigService APIs list_tags_for_resource, untag_resource, tag_resource

kbalk:
Implementation of three tag-related ConfigService APIs, with the exception of handling ConfigRule as that resource has not yet been implemented:

- `list_tags_for_resource`
- `tag_resource`
- `untag_resource`

### DIFF
--- a/tests/test_config/test_config_tags.py
+++ b/tests/test_config/test_config_tags.py
@@ -47,7 +47,7 @@ def config_aggregators_info(client):
 @mock_config
 def test_tag_resource():
     """Test the ConfigSource API tag_resource()."""
-    client = boto3.client("config", region_name="us-east-1")
+    client = boto3.client("config", region_name=TEST_REGION)
 
     # Try an ARN when there are no configs instantiated.
     no_config_arn = "no_configs"
@@ -131,7 +131,7 @@ def test_tag_resource():
 @mock_config
 def test_untag_resource():
     """Test the ConfigSource API untag_resource()."""
-    client = boto3.client("config", region_name="us-east-1")
+    client = boto3.client("config", region_name=TEST_REGION)
 
     # Try an ARN when there are no configs instantiated.
     no_config_arn = "no_configs"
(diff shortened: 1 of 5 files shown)


tests/test_config/test_config_tags.py
```diff
@@ -0,0 +1,206 @@
...
+from botocore.exceptions import ClientError
+from botocore.exceptions import ParamValidationError
+import pytest
+
+from moto.config import mock_config
+from moto.config.models import MAX_TAGS_IN_ARG
+from moto.config.models import random_string
+from moto.core import ACCOUNT_ID
+
+TEST_REGION = "us-east-1"
```
bblommers, line 19: This isn't used anywhere

bblommers:
Looks good to me sofar @kbalk. I'll mark the PR as draft, to make sure we don't accidentally merge it untill it's complete - let me know when it's ready for a full review.

tests/test_config/test_config_tags.py
kbalk: It is ... as an argument to put_aggregation_authorization().  In `test_config.py` that argument is hard coded.  Here I made it a define.  See line 111.

kbalk:
@bblommers The easy way to handle ConfigService pagination would be to use code similar to what was written for `list_discovered_resources()` and `list_aggregate_discovered_resources()`.  The nice way would be to use a generic paginator.  There's an open PR https://github.com/spulec/moto/pull/3781 to create one, but work stopped on that PR in March. 

I assume I'll need to use code similar to what exists for ConfigService() for now?

kbalk:
@bblommers I was overthinking `list_tags_for_resource`.  It appears a config instance can only hold 50 tags.  For `list_tags_for_resoucrce` the "limit" arg is ignored (you get all tags regardless) and the response does not include "NextToken".

I've tested it with the CLI and boto3.  I'm wondering if I'm missing something.

kbalk:
@bblommers I think this is ready for review.

bblommers:
> 
> 
> @bblommers I was overthinking `list_tags_for_resource`. It appears a config instance can only hold 50 tags. For `list_tags_for_resoucrce` the "limit" arg is ignored (you get all tags regardless) and the response does not include "NextToken".
> 
> I've tested it with the CLI and boto3. I'm wondering if I'm missing something.

Wouldn't be the first instance where the behaviour and documentation of boto3 differ 🙂

bblommers (approved):
Thanks for the PR @kbalk!

Issue #4118, opened 2021-08-02
SecretsManager doesn't handle KmsKeyId


glungley:
Submitting KmsKeyId in create/update secret doesn't go anywhere and the describe_secret response has a hard coded empty string response

PR #4119, opened 2021-08-02
#4118 Add KmsKeyId Support to secretsmanager

glungley:


moto/secretsmanager/models.py
```diff
@@ -61,6 +61,7 @@ def __init__(
         secret_binary=None,
         description=None,
         tags=[],
+        KmsKeyId=None,
```
bblommers, line 64: Variables should be snake-case (`kms_key_id`) (everywhere)

bblommers (changes requested):
Hi @glungley, thanks for the PR!
Looks like the linter is failing, you might have to run black to get the CI to pass:
```
pip install black==19.10b0
make format
```

moto/secretsmanager/models.py
glungley: Addressed, does a module somewhere else address that the boto API has the values in camel-case?

bblommers:
Hi @glungley, make sure to use the right version of black (`19.10b0`), as other versions will result in different formatting

moto/secretsmanager/models.py
bblommers: Ah, I missed that - no, we'd have to update the `response.py` file as well to read that param:
https://github.com/spulec/moto/blob/master/moto/secretsmanager/responses.py#L43

That goes for both create_ and update_secret.

glungley: Got the unit tests running locally, so have fixed that now. Thanks for the help btw

bblommers (approved):
LGTM - thanks again @glungley!

PR #4120, opened 2021-08-02
Support cognito:user_status filter in cognito-idp list_users

nluk:
Hi, I've added the option to filter users by implicit attributes in `users_list` endpoint:

- username
- status (`enabled` field)
- cognito:user_status (`status` field)

Those aren't stored in `user.attributes`, which lead to an empty response if such a filter was specified, even though they are listed in AWS [docs](https://docs.aws.amazon.com/cognito-user-identity-pools/latest/APIReference/API_ListUsers.html#API_ListUsers_RequestSyntax).


bblommers (approved):
LGTM - thanks for the PR @nluk!

PR #4128, opened 2021-08-03
Update ecr

gruebel:
My plan was to tackle an issue regarding the missing functionality to define policies for ECR repos, but then I realized that this service wasn't updated for quite some time. 

So, I added the missing fields in `create_repository`, fixed CloudFormation functionality, added tagging and also added a default KMS policy, which was missing and therefore failed a couple of the Terraform tests related to ECR.


gruebel:
and the Terraform test output
image

bblommers (approved):
LGTM - thank you @gruebel!

PR #4129, opened 2021-08-03
Fix for 'status' filter, separate inherent_attributes test

nluk:
Hi, coming back to the `inherent_attributes` filter. The `status` filter handling wasn't done correctly: 
- `user.enabled` is a `bool`, so `inherent_attributes['status'](user) == value` is a `bool == str` comparison - obviously failing.
- AWS expects a case-sensitive *Enabled* or *Disabled* value for this filter, as can be seen working correctly here:
image

With that mistake out of the way, I've decided to make a separate test case, to verify **all** `inherent_attributes`.

### DIFF
--- a/tests/test_cognitoidp/test_cognitoidp.py
+++ b/tests/test_cognitoidp/test_cognitoidp.py
@@ -1221,12 +1221,43 @@ def test_list_users():
     result["Users"].should.have.length_of(1)
     result["Users"][0]["Username"].should.equal(username_bis)
 
-    # checking Filter for inherent attributes
-    result = conn.list_users(
-        UserPoolId=user_pool_id, Filter='username = "{}"'.format(username_bis)
-    )
+
+@mock_cognitoidp
+def test_list_users_inherent_attributes():
+    conn = boto3.client("cognito-idp", "us-west-2")
+
+    username = str(uuid.uuid4())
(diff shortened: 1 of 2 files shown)


bblommers (approved):
Nice one - thanks for the follow-up @nluk!

tests/test_cognitoidp/test_cognitoidp.py
```diff
@@ -1221,12 +1221,43 @@ def test_list_users():
...
+        ("username", disabled_user_username, "Username", disabled_user_username),
+        ("status", "Disabled", "Enabled", False),
+        ("cognito:user_status", "CONFIRMED", "UserStatus", "CONFIRMED"),
+    ]
+
+    for filter, filter_value, response_field, response_field_expected_value in filters:
+        result = conn.list_users(
+            UserPoolId=user_pool_id, Filter='{}="{}"'.format(filter, filter_value)
+        )
+        print(result)
```
nluk, line 1258: Oops.

nluk: @bblommers Might want to remove that print, without another PR from my side :disappointed:

bblommers: Oops. I should've spotted that as well! Will remove it 🙂

PR #4131, opened 2021-08-03
Improve the tagging of resources in the elbv2 module

jrauen:
tags were not being added to every kind of resource (listener, listener rule, etc) which was causing certain expected responses to fail.

### DIFF
--- a/moto/elbv2/models.py
+++ b/moto/elbv2/models.py
@@ -134,6 +134,13 @@ class FakeTargetGroup(CloudFormationModel):
             raise TooManyTagsError()
         self.tags[key] = value
 
+    def list_tags(self):
+        return self.tags
+
+    def remove_tag(self, key):
+        if key in self.tags:
+            del self.tags[key]
+
     def health_for(self, target, ec2_backend):
         t = self.targets.get(target["id"])
         if t is None:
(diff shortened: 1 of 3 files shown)


bblommers (changes requested):
Hi @jrauen, thanks for the PR!
Would you mind adding more test cases, to ensure all the new methods/branches are covered?
Especially the error scenarios, it would be really useful to have unit tests for those.

jrauen:
Thanks for the feedback. After spending some time learning the terraform tests repo (very interesting!) I'm getting an odd result where I'm not seeing the same results of the tests that ran after this PR was created. I seem to get a different set of errors each time.

Also, what parts of the PR would you like to see covered by tests that aren't already?

bblommers:
I was mostly looking at the error scenarios in add_tags:
```
if not balancer:
...
if not listener:
..
```
Ideally, every new line of code is covered. There's a overview on which lines are currently not covered here: https://codecov.io/gh/spulec/moto/compare/0f8da52196ec5b66cbaea4ab9cee4c1b5d6ee231...c54545a6b134db03664d32d8d41a1993bc81801a/diff

bblommers:
> 
> 
> Thanks for the feedback. After spending some time learning the terraform tests repo (very interesting!) I'm getting an odd result where I'm not seeing the same results of the tests that ran after this PR was created. I seem to get a different set of errors each time.

Yeah, it looks like the failing TF test (`TestAccAWSProvider`) is a bit finicky - they'll fail one time, but work fine the next. Don't worry about it, as long as the other tests stil work it'll be OK 👍

bblommers:
This behaviour is now part of Moto already, as part of #4956 and #4964, so I'll close this PR. Thanks again for your work on this @jrauen :+1:

PR #4139, opened 2021-08-04
Add ecr tag mutability & image scan config

gruebel:
I also refactored the exceptions a bit and added some convenience function for retrieving the repository and parsing the resource arn.

Terraform tests also passed
image


bblommers:
> 
> 
> @bblommers do you know why the Terraform tests take so long? I took a look into the log file and I'm surprised how much time is wasted just to do the authentication and account checking till the actual resource creation is triggered. I'm afraid adding more and more Terraform tests will result in super long PR checks, they already take ~40min to pass and there are not that many at the moment.

Hey @gruebel - the performance is quite bad indeed, with the full test suite taking just under 5 hours.
I haven't had the opportunity to look into it and see how this can be improved though.

bblommers (approved):
LGTM - thanks for the PR and Terraform test!

PR #4142, opened 2021-08-05
Remove pkg_resources module

bblommers:
As the `setuptools` dependency was removed, the `pkg_resources`-module is no longer available to Moto.

This PR rewrites all functionality that depended on pkg_resources, to instead use the `importlib_metadata` and `pgkdata`-modules.

It also explicitly removes the `pkg_resources`-module from the dependency test, to ensure it works.

(Note: XRay has an implicit dependency on pkg_resources, so we're adding setuptools back to that module only.)

(Note 2: For Py3.8, we could start using the buildin `importlib.metadata`, but for we'll need the backport for 3.6 and 3.7.)

Successful dependency test: https://github.com/bblommers/moto/runs/3253699280?check_suite_focus=true

### DIFF
--- a/moto/config/models.py
+++ b/moto/config/models.py
@@ -1,11 +1,9 @@
-import json
 import re
 import time
 import random
 import string
 
 from datetime import datetime
-import pkg_resources
 
 from boto3 import Session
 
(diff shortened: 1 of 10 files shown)


asherf:
yay! thanks!

whardier:
Nice!  I had recommended setuptools be vanquished.. it's nice to see this was resolved.

PR #4143, opened 2021-08-05
Add support for KMS update-alias

whummer:
Add support for KMS `UpdateAlias` API (covered by Terraform test case). /cc @bblommers


bblommers (approved):
LGTM - thanks @whummer!

PR #4147, opened 2021-08-05
Add tagging support for Cognito identity pools

whummer:
Add tagging support for Cognito identity pools (covered by Terraform test suite). /cc @bblommers


moto/cognitoidentity/responses.py
```diff
@@ -34,10 +36,10 @@ def update_identity_pool(self):
         allow_classic = self._get_bool_param("AllowClassicFlow")
         login_providers = self._get_param("SupportedLoginProviders")
         provider_name = self._get_param("DeveloperProviderName")
-        provider_arns = self._get_multi_param("OpenIdConnectProviderARNs")
```
whummer, line 37: Looks like we were using the wrong methods here - Cognito Identity is already using JSON payloads (and not URL-encoded form data), hence we can simply use `_get_param(..)` here.

whummer:
@bblommers tests are passing now - should be ready to review/merge. (I removed the Cognito entry from `terraform-tests.success.txt` again, as some other features are still required to get the tests passing e2e). Thanks!

bblommers (approved):
LGTM - thanks @whummer!

PR #4148, opened 2021-08-06
Add ecr repo policy

### LINKED ISSUE #4032: Feature request - ecr.set_repository_policy()
This method is not implemented as per today.  It would be nice to have.

I've taken a look at the code, and if you think it is an okay issue, I could try to make a PR.

Thanks for the great library!

gruebel:
Fixes #4032 

I added the `IAMPolicyDocumentValidator` and had to patch some methods, because the ECR policy is not that strict, but I think it is still worth to use it and if users complain about it, we can still remove it and don't do any validation at all 😄 

And also the Terraform tests are working 💪 
image


bblommers (approved):
LGTM! Thanks @gruebel

bblommers:
> 
> 
> if users complain about it, we can still remove it and don't do any validation at all 😄
> 

Yeah, we'll either remove it or introduce a flag to disable it. We'll worry about that when we get there 🙂

PR #4149, opened 2021-08-06
add stack status in execute_change_set call.

### LINKED ISSUE #4141: Cloudformation: ExecuteChangeSet should also change the status of the stack
Hi there,

I was working on a Cloudformation library and I'm using moto to test the behaviour of cfn stacks. One thing I came across is the status of the stack is not changed after an `ExecuteChangeSet` call. In the following pseudo code:

```
    cfn.create_stack(
        StackName=stack.name,
        TemplateBody=old_json,
    )
    assert stack.status == "CREATE_COMPLETE"
    cfn.update_stack(
        StackName=stack.name,
        TemplateBody=new_json,
    )
    assert stack.status == "UPDATE_COMPLETE"
```

I'm expecting the `stack.status` in the last line to be `UPDATE_COMPLETE`, this is inline with Cloudformation behaviour. However, this assertion failed as `stack.status` is still `CREATE_COMPLETE` , that's because in this method:

https://github.com/spulec/moto/blob/79f0cc9e9e845461c337d1d297727483cfd0b651/moto/cloudformation/models.py#L683

We are merely adding events, and we have not changed the status of the stack.

Happy to create a PR to address this issue, please let me know what you think. Thanks!

xiaket:
Resolves #4141 

Thanks for the patience, this PR is ready for review now. Please take a look at the commit message as I've left my rationale there for this bigger than expected change.

Signed-off-by: Kai Xia <kaix+github@fastmail.com>

### DIFF
--- a/moto/cloudformation/models.py
+++ b/moto/cloudformation/models.py
@@ -214,7 +214,6 @@ class FakeStack(BaseModel):
         tags=None,
         role_arn=None,
         cross_stack_resources=None,
-        create_change_set=False,
     ):
         self.stack_id = stack_id
         self.name = name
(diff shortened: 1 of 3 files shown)


xiaket:
Sorry I created some noise here, I was being lazy and used github to run the tests. Also, I think I'll have to change the scope of the work because I'll also have to change the `create_change_set` method. Will bring more updates when available. Apologize again for the spam. Thanks!

bblommers (approved):
Nice one. It does make sense to treat a ChangeSet as it's own thing, rather than as a subclass of a Stack.
Thanks for the PR @xiaket!

PR #4150, opened 2021-08-06
Add Log Service describe_resource_policies, delete_resource_policy

kbalk:
Implementation of two Log Service APIs:

- `describe_resource_policies`
- `delete_resource_policy`

Corrected return value for `put_resource_policy` - should return a dict versus nothing.

Corrected account ID in LogGroup ARN.


bblommers (approved):
LGTM! Thanks for the general code improvements as well (snake_case everywhere, better variable names) - that's always appreciated 👍

PR #4153, opened 2021-08-07
Add ecr lifecycle policy

gruebel:
I created a dedicated lifecycle policy validator, which is definitely not complete, but I think a good start 💪 I divided the tests for it, so I don't write just one test with a long list of paramatrized tests.

And the related Terraform test 🙂 
image


bblommers (approved):
Nice @gruebel - that's an impressive test suite.
Thanks for the PR!

gruebel:
Had to test the validator properly 😄

PR #4159, opened 2021-08-10
Add ecr registry policy

gruebel:
I think this is the last relevant policy endpoint for ECR 😄 was fun to work again on `moto` 💪 

And of course the Terraform tests are looking nice 🙂 
image


bblommers:
>  was fun to work again on `moto` :muscle: 
> 

That's always good to hear :slightly_smiling_face:

PR #4166, opened 2021-08-12
Add ecr image scan

gruebel:
the result from `ecr. describe_image_scan_findings()` is of course totally fake and this implementation is just for completeness and if someone wants to construct some kind of custom code around those methods and needs `moto` to mock those requests 😄


bblommers (approved):
LGTM! Thanks @gruebel

PR #4170, opened 2021-08-13
Add ecr replication config

gruebel:
Sadly there are some problems with the Terraform test, because it seems like the regions are the same and this is not allowed for the same account, if you want to setup up replication.


bblommers:
> 
> 
> Sadly there are some problems with the Terraform test, because it seems like the regions are the same and this is not allowed for the same account, if you want to setup up replication.

That's a shame - would still be useful to include the TF test.
Do you know where the regions come from in the test? Could we override them, by passing environment variables to specify to use different regions?

gruebel:
@bblommers nice, you already fixed it in a different way, so reverted back my change and here is the Terraform output
image

bblommers (approved):
Thanks @gruebel!

PR #4171, opened 2021-08-13
Add AWS ConfigService put_config_rule, delete_config_rule, describe_config_rule

kbalk:
Basic implementation of the AWS ConfigService APIs for:

- `put_config_rule`
- `delete_config_rule`
- `describe_config_rule`

As written, these APIs handle the "accounting" for ConfigRules, but events are not created or triggered, and there's no interaction with the config recorder.


moto/config/models.py
```diff
@@ -411,13 +475,340 @@ def update(
...
+    ]
+    EVENT_SOURCES = ["aws.config"]
+
+    def __init__(
+        self, event_source, message_type, max_execution_frequency=DEFAULT_FREQUENCY
+    ):
+        super().__init__(capitalize_start=True, capitalize_arn=False)
+        if not event_source:
+            # boto3 doesn't have a specific error if this field is missing.
+            raise ParamValidationError(
```
bblommers, line 556: ParamValidationErrors are raised by boto3 itself, before the request ever gets send to Moto - it's like an initial validation.
So we don't need these

moto/config/aws_managed_rules.py
```diff
@@ -0,0 +1,200 @@
+"""List of IDs for AWS Config Managed rules.
+
+Started with the list found here, then went to each page to get the ID:
```
bblommers, line 3: Could we write a script to get this list from boto3? Would be good to have something reproducible, in case this list will change in the future

kbalk: I agree, it would be nice to be able to get the latest list, but I couldn't find a way to do so.  It don't see any APIs for the Config Service that would get that list.  boto3 must have the list for them to perform their validation checks, but if I pick one of those IDs and search through boto3's github repo, it's not finding a match.  There are yaml files related to best practices that list IDs, but they don't have the complete list and refer you to the page where I got the list.

I have a lot more work to do for this PR, but I wanted to commit what I had so far to ensure I didn't accidently lose or delete what I have so far.

moto/config/models.py
kbalk: Thank you.  I should have known better.

moto/config/aws_managed_rules.py
kbalk: @bblommers This just got a lot worse.  I was experimenting a bit and the AWS managed rule constrains the value of the InputParameters and MaximumExecutionFrequency arguments.  I would need that info to perform accurate validation.  I really wish I could find an API that would retrieve that info, otherwise I'd need to do some web scraping and create the structure I could use for validation.

kbalk: https://github.com/awsdocs/aws-config-developer-guide/blob/main/doc_source/managed-rules-by-aws-config.md  -- it's the source for the AWS documentation and Managed Rule list.

bblommers: It's a bit awkward, but it is possible to retrieve all rules programmatically - by enabling all rules first in the Console, getting the full list via the API and then deleting them all again.

The output of `describe_config_rules` can be found here - feel free to use this in your PR:
https://gist.github.com/bblommers/0a451593c7473e95ad74c671c952258a

This should contain all AWS managed rules, and their InputParameters/MaximumExecutionFrequency

Steps on how to do it, if you're interested in reproducing the list:
https://gist.github.com/bblommers/91cf995def518f9057a4afe71ca519a4

**Note that you might incur costs, so follow these steps at your own risk**

moto/config/models.py
bblommers: All good - I prefer too many tests over too few!

moto/config/aws_managed_rules.py
kbalk: Thank you very much!  I'll use your list.  I had hoped it would have been simple to read the AWS markdown files and then convert them to JSON and work from there, but there's no simple markdown to JSON conversion for Python.  It wouldn't be too difficult to write, but your solution will do just fine.

kbalk:
@bblommers Sorry to trouble you again ... I create a lambda for testing purposes using code I copied from other moto tests and while the tests work on my system, they fail during the workflow tests.  I initially zipped the lambda code and put it into a s3 bucket, but it didn't like the constraints, i.e., the region "us-east-1":

```
E           botocore.exceptions.ClientError: An error occurred (InvalidLocationConstraint) when calling the CreateBucket operation: The specified location-constraint is not valid
```

So in my last commit I eliminated the s3 bucket and just included the zipped code.  I'm getting a different error - the server doesn't respond (response code 500).

I don't know what I'm doing wrong or if there's something special I need to be doing for the workflow environment.

moto/config/models.py
bblommers: Turns out I was wrong about this - boto does not intercept this particular error, so catching it manually does make sense.
Incidentally, it's also the cause of the ServerMode error that we're seeing now.

The Moto internals can gracefully deal with Exceptions - as long as it extends a specific werkzeug HTTPException.
Any other exceptions, it doesn't know how to deal with those, so it throws a generic 500-exception instead of the custom crafted exception that Boto3 expects.

The solution would be to create a custom Exception for this, and ensure that it extends the JsonRESTError-class, like all our custom exceptions.

```
class WhateverNameMakesSense(JsonRESTError):
    code = 400

    def __init__(self):
        super().__init__(
            "ParamValidation",
            "Missing required parameter in ConfigRule.SourceDetails: 'EventSource'",
        )
```

The best way to verify that Moto throws the correct error, is to always follow this format in the unit tests:
```
        client.put_config_rule(ConfigRule=custom_rule)
err["Message"].should.equal("Missing required parameter in ConfigRule.SourceDetails: 'EventSource'")
err["Code"].should.equal("ParamValidation")
```
This test will only pass if the error is returned in the correct format.

kbalk: Is it ok if I use `assert` versus the `sure` library ("should.equal")?

I'm also having trouble if I use your suggested code and I think it's because the error isn't really a ClientError.  The error I'm seeing is as follows (using your suggested code):

```
{'Message': '{\n    
"message": "Missing required parameter in ConfigRule.SourceDetails: "EventSource"",\n    
"__type": "ParamValidationError"\n
}', 
'Code': '400'}
```

The expected format of a ClientError is more like this:

```
{'Message': "Value 'foo' at 'configRule.source.sourceDetails.eventSource'
failed to satisfy constraint: Member must satisfy enum value set: {aws.config}", 
'Code': 'ValidationException'}
```

I can see there are other error types in `moto.core.exceptions.py`, but I'm concerned that the users of this API will expect a ClientError.  I can convert the incoming message to a ClientError type, but I'm not sure if you'd be ok with that.  I'll give that a try so I continue adding some missing unit tests, but I'm happy to change it based on your feedback/decision.

bblommers: Make sure that the error msg is valid JSON, i.e. small quotes around the 'EventSource'.
If the message can't be used directly as a JSON key, that's when you see the weird format.

kbalk: Damn you're good!  That was the problem.  Thank you ...

bblommers: > 
> 
> Damn you're good! That was the problem. Thank you ...

Yeah.. I already ran into the same issue when I was debugging. I just forgot to mention it 🙁 

> 
> 
> Is it ok if I use `assert` versus the `sure` library ("should.equal")?

Yeah, that's absolutely fine. I prefer `should.equal` because it reads better IMO, but it's very much a personal preference, so we don't really enforce any style when it comes to assertions.

moto/config/aws_managed_rules.py
kbalk: I started working on the list of AWS managed rules and out of curiosity, I compared the IDs in your list with the IDs in the markdown list I had created.  There are 34 more IDs on the managed list than on yours:

```
{'IAM_GROUP_HAS_USERS_CHECK', 'S3_BUCKET_BLACKLISTED_ACTIONS_PROHIBITED', 'CODEBUILD_PROJECT_ENVVAR_AWSCRED_CHECK', 'EC2_MANAGEDINSTANCE_PLATFORM_CHECK', 'IAM_INLINE_POLICY_BLOCKED_KMS_ACTIONS', 'RESTRICTED_INCOMING_TRAFFIC', 'SERVICE_VPC_ENDPOINT_ENABLED', 'S3_BUCKET_POLICY_NOT_MORE_PERMISSIVE', 'ELB_TLS_HTTPS_LISTENERS_ONLY', 'DESIRED_INSTANCE_TYPE', 'IAM_ROLE_MANAGED_POLICY_CHECK', 'ELB_PREDEFINED_SECURITY_POLICY_SSL_CHECK', 'EC2_MANAGEDINSTANCE_APPLICATIONS_REQUIRED', 'CLOUDWATCH_ALARM_ACTION_CHECK', 'API_GW_ENDPOINT_TYPE_CHECK', 'EC2_MANAGEDINSTANCE_APPLICATIONS_BLACKLISTED', 'ACCESS_KEYS_ROTATED', 'REDSHIFT_CLUSTER_CONFIGURATION_CHECK', 'IAM_USER_UNUSED_CREDENTIALS_CHECK', 'REQUIRED_TAGS', 'CLOUDFORMATION_STACK_DRIFT_DETECTION_CHECK', 'APPROVED_AMIS_BY_TAG', 'FMS_WEBACL_RESOURCE_POLICY_CHECK', 'CLOUDWATCH_ALARM_RESOURCE_CHECK', 'EC2_MANAGEDINSTANCE_INVENTORY_BLACKLISTED', 'APPROVED_AMIS_BY_ID', 'LAMBDA_FUNCTION_SETTINGS_CHECK', 'IAM_POLICY_BLACKLISTED_CHECK', 'DESIRED_INSTANCE_TENANCY', 'CODEBUILD_PROJECT_SOURCE_REPO_URL_CHECK', 'FMS_SHIELD_RESOURCE_POLICY_CHECK', 'IAM_POLICY_IN_USE', 'ELB_CUSTOM_SECURITY_POLICY_SSL_CHECK', 'REDSHIFT_CLUSTER_MAINTENANCESETTINGS_CHECK', 'CLOUDWATCH_ALARM_SETTINGS_CHECK', 'IAM_CUSTOMER_POLICY_BLOCKED_KMS_ACTIONS', 'FMS_WEBACL_RULEGROUP_ASSOCIATION_CHECK'}
```

I was looking at the history of the AWS markdown file and it's updated every month or every other month.

I'm going to write a script to pulldown and parse the markdown files.  I'm assuming this script could update or replace the `aws_managed_rules.py` file during a moto install/release?  But then wouldn't it need to be checked in as well?  Could you tell me where I should commit the script and how it should be invoked (or handle that part if you'd like)?  It all seems a bit messy.

bblommers: Hmm.. interesting. Strange, that there are 34 rules not available via the Console. (Either that, or I just missed something)

Yes, it makes sense to have it replace `aws_managed_rules.py` with each invocation.

All of our other scripts are inside the `scripts` folder. None of them are invoked automatically during release at the moment, they're all manual. So I'd be happy to treat this script the same - just check it in inside that folder, and we'll invoke it manually until there's a better way.

>
>  It all seems a bit messy.

Couldn't agree more!

kbalk: Some of these rules are region-specific.  I wasn't planning on checking for that.  

```
**AWS Region:** All supported AWS regions except Asia Pacific \(Osaka\), Europe \(Milan\), Middle East \(Bahrain\) Region
```

kbalk: I've committed the script `pull_down_aws_managed_rules.py` into the scripts directory.  Couple of notes:

- It's slow because it's reading 192 files at the moment.  If run manually, it'd be nice to print the name of the file it's currently reading, but I removed that for now.
- If it detects an error, it prints an error to syserr.  I wasn't sure what else to do.
- It uses a relative path to put the json file into `moto/moto/config`.
- I output the region info, but I currently don't plan to use it.  That's an exercise for someone else!

Feel free to modify this file and make it work as you see fit.  For now I'll use the JSON output to add ConfigRule validations for managed rules.

kbalk: Oops ... there's a bug in the collection of parameters.

kbalk:
@bblommers It's ready for review.

tests/test_config/test_config_rules.py
```diff
@@ -0,0 +1,728 @@
...
+        client.put_config_rule(ConfigRule=managed_rule)
+    err = exc.value.response["Error"]
+    assert err["Code"] == "InvalidParameterValueException"
+    assert (
+        "The required parameter [alarmActionRequired, "
+        "insufficientDataActionRequired, okActionRequired] is not present "
+        "in the inputParameters" in err["Message"]
+    )
+
+    # If no MaxExecutionFrequency specified, set it to the default.
```
bblommers, line 290: Looks like this is now covered in `test_valid_put_config_custom_rule`?

moto/config/models.py
```diff
@@ -411,13 +452,403 @@ def update(
...
+            return
+
+        # Otherwise, owner == "CUSTOM_LAMBDA"
+        if not source_details:
+            raise InvalidParameterValueException(
+                "SourceDetails should be null/empty if the owner is AWS. "
+                "SourceDetails should be provided if the owner is CUSTOM_LAMBDA"
+            )
+
+        lambda_func = lambda_backends[region].get_function(source_identifier)
```
bblommers, line 641: Can we move the `lambda_backends` import down to this level?
It's quite a slow operation, and it imports some dependencies that might not be necessary for all Config-users, so we should only load it if/when required

bblommers (changes requested):
Looks good overall @kbalk! The test coverage is very thorough, that's always good to see.
Just have two minor comments - happy to merge once that's addressed.

tests/test_config/test_config_rules.py
kbalk: There are two `MaximumExecutionFrequency`, one at the `ConfigRule.Source.SourceDetails` level and one at the `ConfigRule` level.  I am very confused as to how they're related and how to set the default values.  I raised a question in the AWS Forum and never got an answer.  I experimented a bit and between boto3 and the AWS Console, it was no clearer how they interacted.

What's even more confusing is that for a custom rule, you can have multiple rules and they can each have a different  `MaximumExecutionFrequency` value, so how do you determine the value for the one at the ConfigRule level?

This will probably have to be revisited when people complain it's not working as expected or if I ever get an answer from AWS.

In any case, `test_aws_managed_rule_errors()` is testing at the `ConfigRule.Source.SourceDetails` level for AWS managed rules, and `test_valid_put_config_custom_rule` is testing `ConfigRule.Source.SourceDetails` level for custom rules. 

All very messy.

bblommers: Ah, gotcha. Thanks for the explanation.

bblommers (approved):
LGTM! Thanks for the PR and all the work around it, @kbalk!

PR #4173, opened 2021-08-13
IoT: Add VersionsLimitExceededException

number09:
IoT : `create_policy_version`  should throw `VersionsLimitExceededException` if the number of policy versions exceeds 5


bblommers (approved):
LGTM - thanks for the PR @number09!

PR #4174, opened 2021-08-13
Feature S3 Object Lock

pinzon:
While trying to solve issue localstack/localstack#4166 I figure out that moto doesnt have the [Object Lock feature](https://docs.aws.amazon.com/AmazonS3/latest/userguide/object-lock-overview.html) of the S3 service.

So I implemented it, including some validations necessary to solve the issue previously mentioned.  

The quirk of this implementation is that the retention lock is that only forces the lock when is set to COMPLIANCE mode.


bblommers:
Hi @pinzon, thanks for the PR!
Would you mind running the linter on this? The CI currently fails on the formatting.
```
pip install black==19.10b0
make format
make lint
```

pinzon:
@bblommers thanks for the commands, I was running format with the wrong black version.
Just pushed a new commit.

pinzon:
@bblommers I'm not sure what to do to make the tests pass in server mode. 
To me it seems that the headers related to object lock are not being detected at the moto server.

pinzon:
The last commit changes the usage of `headers` to `self.headers` because the s3-object lock params weren't being detected.
Thank you @bblommers for your help.

bblommers (approved):
LGTM - thank you @pinzon!

Issue #4020, opened 2021-06-18
CloudWatch put_metric_alarm does not handle TreatMissingData parameter


Vilsepi:
On moto 2.0.9, when creating a CloudWatch metric alarm, you can pass the `TreatMissingData` parameter, but it is not stored by moto.

**What is expected to happen:** When describing alarms, an alarm should have the `TreatMissingData` property which was passed when creating that alarm.

**What actually happens:** When describing alarms, none of the alarms have `TreatMissingData`, even though it was passed when creating those alarms.

**How to reproduce the issue:**

This code should pass but instead fails the assert.

```
import os
import pytest
from moto import mock_cloudwatch
import boto3

@pytest.fixture(scope="function")
def aws_credentials():
    os.environ["AWS_ACCESS_KEY_ID"] = "testing"
    os.environ["AWS_SECRET_ACCESS_KEY"] = "testing"
    os.environ["AWS_SECURITY_TOKEN"] = "testing"
    os.environ["AWS_SESSION_TOKEN"] = "testing"

@pytest.fixture(scope="function")
def cloudwatch(aws_credentials):
    with mock_cloudwatch():
        yield boto3.client("cloudwatch", region_name="eu-west-1")

def test_moto_issue2(cloudwatch):
    cloudwatch.put_metric_alarm(
        AlarmName="MyAlarm",
        ActionsEnabled=True,
        AlarmActions=["foo_alarm_arn"],
        MetricName="DDoSDetected",
        Namespace="AWS/DDoSProtection",
        Statistic="Maximum",
        Dimensions=[
            {
                "Name": "ResourceArn",
                "Value": "foo_resource_arn"
            }
        ],
        Period=60,
        EvaluationPeriods=1,
        DatapointsToAlarm=1,
        Threshold=0.0,
        TreatMissingData="notBreaching",
        ComparisonOperator="GreaterThanThreshold"
    )
    alarm = cloudwatch.describe_alarms()["MetricAlarms"][0]
    assert alarm["TreatMissingData"] == "notBreaching"

```

I looked at the [responses.py](https://github.com/spulec/moto/blob/03b2009a1c065196be1256836473e8d72c4f65ab/moto/cloudwatch/responses.py#L18) and it seems like it's missing that parameter. Is this code written by hand or generated from AWS interface specifications, i.e. should this be added by hand or by fetching a newer API spec from AWS?

bblommers:
[...]

PR #4175, opened 2021-08-14
Enhanced support for ami filtering

macnev2013:
## Added
- Support for filtering ami based on owner alias
- Added owner_alias attribute in ami class
- Added new ami (nat instance)

PR #4179, opened 2021-08-15
Add cloudwatch tags

gruebel:
Had to also add the possibility to search for AMIs by `amazon` owner alias, because of a Terraform test, which queries that data.

I also added several missing parameters, I think that's all at the moment and Terraform is also happy 😄 

image


bblommers:
Hi @gruebel, looks like the EC2 AMI filter was already added this morning, in this PR: https://github.com/spulec/moto/pull/4175/files

gruebel:
yeah, I also saw that now 😄 looks much better than my simple fix 💪 rebasing right now and have to add an AMI to the json file, which is used by one of the Terraform tests.

gruebel:
Forgot to mention, this also fixes #4020

bblommers (approved):
Awesome - thanks @gruebel!

PR #4184, opened 2021-08-17
added deprecate_thing_type method for IOT

arielb135:


tests/test_iot/test_iot.py
```diff
@@ -2210,3 +2213,40 @@ def test_topic_rule_delete(self):
...
+        )
+
+        res = client.describe_thing_type(thingTypeName=thing_type_name)
+        res["thingTypeMetadata"]["deprecated"].should.equal(False)
+
+    @mock_iot
+    def test_deprecate_thing_type_not_exist(self):
+        client = boto3.client("iot", region_name="ap-northeast-1")
+        thing_type_name = "my-type-name"
+        with pytest.raises(client.exceptions.ResourceNotFoundException):
```
bblommers, line 2248: We should catch `botocore.exceptions.ClientError` here, as that's the error class that boto3 uses.
Ideally we also assert that the right error is thrown:
```
    ....
err["Message"].should.equal("...")
```

bblommers (changes requested):
Hi @arielb135, thanks for the PR!
Looks like the linter is failing - make sure you use the right version of black, as different versions will produce different results.
```
black==19.10b0
```

tests/test_iot/test_iot.py
arielb135: Hi, thanks for that, indeed i have a liner issue - if the next one fails i'll search for the correct one.
the ClientError is the uglier way to catch boto3 exceptions, some clients do offer the "specific" exception:

https://boto3.amazonaws.com/v1/documentation/api/latest/guide/error-handling.html#parsing-error-responses-and-catching-exceptions-from-aws-services

as shown here 
image

bblommers: Fair enough - I misread, and thought it referred to our custom `moto.iot.exceptions.ResourceNotFoundException`

bblommers (approved):
LGTM!

PR #4185, opened 2021-08-17
Fargate

ferruzzi:
Added EKS support for Fargate Profiles with unit testing:
- list_fargate_profiles
- create_fargate_profile
- describe_fargate_profile
- delete_fargate_profile

### DIFF
--- a/moto/eks/models.py
+++ b/moto/eks/models.py
@@ -79,7 +79,9 @@ FARGATE_PROFILE_EXISTS_MSG = (
     "A Fargate Profile already exists with this name in this cluster."
 )
 FARGATE_PROFILE_NEEDS_SELECTOR_MSG = "Fargate Profile requires at least one selector."
-FARGATE_PROFILE_NOT_FOUND_MSG = "No Fargate Profile found with name: {fargateName}."
+FARGATE_PROFILE_NOT_FOUND_MSG = (
+    "No Fargate Profile found with name: {fargateProfileName}."
+)
 FARGATE_PROFILE_SELECTOR_NEEDS_NAMESPACE = (
     "Fargate Profile must have at least one selector with at least one namespace value."
 )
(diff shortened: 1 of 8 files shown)


moto/eks/models.py
```diff
@@ -498,6 +675,51 @@ def validate_launch_template_combination(disk_size, remote_access):
...
+        # Every existing Selector must have a namespace
+        if "namespace" not in selector:
+            raise_exception(message=FARGATE_PROFILE_SELECTOR_NEEDS_NAMESPACE)
+        # If a selector has labels, it can not have more than 5
+        if len(selector.get("labels", {})) > 5:
+            raise_exception(message=FARGATE_PROFILE_TOO_MANY_LABELS)
+
+
+def generate_uuid():
+    return "-".join([random_string(_) for _ in [8, 4, 4, 4, 12]]).lower()
```
bblommers, line 720: Can we use the build-in method here?
```
from uuid import uuid4
return str(uuid4())
```

tests/test_eks/test_eks.py
```diff
@@ -797,6 +847,490 @@ def test_create_nodegroup_handles_launch_template_combinations(
...
+    FargateProfileBuilder,
+):
+    _, generated_test_data = FargateProfileBuilder()
+
+    result_time = iso_8601_datetime_without_milliseconds(
+        generated_test_data.fargate_describe_output[FargateProfileAttributes.CREATED_AT]
+    )
+
+    if settings.TEST_SERVER_MODE:
+        RegExTemplates.ISO8601_FORMAT.match(result_time).should.be.true
```
bblommers, line 1042: The sure-library has a match-method, that we can use AFAIK:
```
RegExTemplates.ISO8601_FORMAT.should.match(result_time)
```
Slightly easier to read.

tests/test_eks/test_eks_constants.py
```diff
@@ -49,6 +51,9 @@
 NODEROLE_ARN_KEY = "nodeRole"
 NODEROLE_ARN_VALUE = "arn:aws:iam::123456789012:role/role_name"
 
+POD_EXECUTION_ROLE_ARN_KEY = "podExecutionRoleArn"
+POD_EXECUTION_ROLE_ARN_VALUE = "arn:aws:iam::123456789012:role/role_name"
```
bblommers, line 55: The account ID is centralized, so we should use that over hardcoded values where possible:
```
from moto.core import ACCOUNT_ID
```
(Same for `NODEROLE_ARN_VALUE `, actually - I must have missed that in the last review)

bblommers (changes requested):
Hi @ferruzzi, thanks for the PR!
You might have to rebase, there's a merge conflict that's stopping the CI from running

tests/test_eks/test_eks.py
ferruzzi: Neat, I didn't know that one.  Thanks!

ferruzzi:
Sorry about that; you are right, I did forget to rebase.  Rebased, which required a force push, and added pr fixes commit.

FYI:  It looks like [this commit](https://github.com/spulec/moto/commit/76094f012bfbd90f583ef47e0ce313f631f00742) broke py2.7 compatibility with a syntax error at `moto/core/models.py::548`

tests/test_eks/test_eks.py
ferruzzi: Looks like the server mode tests don't like that:  https://github.com/spulec/moto/pull/4185/checks?check_run_id=3364472186#step:9:4609

bblommers: Mea culpa! I honestly thought that would work

bblommers (approved):
LGTM! Thanks again @ferruzzi

ferruzzi:
I'm going to be taking some offline time, but afterward I'd like to fix up the EKS tests I contributed previously; I've learned better test-writing practices since that contrib so I'll clean those up and add some support for the EKS waiters I think.  We'll see.  Till then... take care 😁

bblommers:
Awesome - tech debt PR's are always welcome 👍

PR #4187, opened 2021-08-17
Reset self._template during stack update

skinitimski:
update calls diff. diff mutates self by calling load_parameters. but the latter reads the value of self._template, which is out of date when we are doing an update.

**Solution:** set self._template to the given template in diff.

also, change the following identifiers to be more clear (since they represent not templates but resource maps):

* `old_template` -> `old_template_resources`
* `new_template` -> `new_template_resources`

### DIFF
--- a/moto/cloudformation/parsing.py
+++ b/moto/cloudformation/parsing.py
@@ -667,7 +667,6 @@ class ResourceMap(collections_abc.Mapping):
             )
             self._parsed_resources[resource_name] = new_resource
 
-
         for logical_name, _ in resources_by_action["Remove"].items():
             resource_json = old_template_resources[logical_name]
             resource = self._parsed_resources[logical_name]
--- a/moto/cloudformation/parsing.py
+++ b/moto/cloudformation/parsing.py
@@ -607,62 +607,69 @@ class ResourceMap(collections_abc.Mapping):
                 )
 
     def diff(self, template, parameters=None):
+
+        self._template = template
+
         if parameters:
             self.input_parameters = parameters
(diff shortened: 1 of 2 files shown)


bblommers (changes requested):
Hi @skinitimski, thanks for the PR!
What's the scenario where a user runs into this issue? Is it possible to create a full unit test using the `boto3.client()` to replicate the bug?

skinitimski:
> Hi @skinitimski, thanks for the PR!

😁 :+1:

> What's the scenario where a user runs into this issue?

On a stack update, if a Parameter is **added** (in the test, [`TimeoutParameter`](https://github.com/spulec/moto/blob/1589ac1b7864db304f20764497026cbe48cd1c4f/tests/test_cloudformation/test_stack_parsing.py#L46), and that parameter is [referenced](https://github.com/spulec/moto/blob/1589ac1b7864db304f20764497026cbe48cd1c4f/tests/test_cloudformation/test_stack_parsing.py#L50), the reference will fail in `parse_and_create_resource` because the [parameters get loaded](https://github.com/spulec/moto/blob/1589ac1b7864db304f20764497026cbe48cd1c4f/moto/cloudformation/parsing.py#L617) using the old template. The same test, without the changes to the main code, [fails with](https://github.com/skinitimski/moto/runs/3362408347) `KeyError: 'TimeoutParameter'`.



> Is it possible to create a full unit test using the `boto3.client()` to replicate the bug?

I will work on that.

skinitimski:
I need to rework this.

PR #4188, opened 2021-08-17
Translate Cloudformation API data to EventBridge API data

skinitimski:
The CloudFormation API takes a [dictionary for `EventPattern`](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-events-rule.html#aws-resource-events-rule-syntax), whereas the EventBridge API takes [a string only](https://docs.aws.amazon.com/eventbridge/latest/APIReference/API_PutRule.html#API_PutRule_RequestSyntax). The "real" cloudformation backend must do a translation. This PR emulates that behavior.

### DIFF
--- a/moto/events/models.py
+++ b/moto/events/models.py
@@ -236,6 +236,10 @@ class Rule(CloudFormationModel):
     ):
         properties = cloudformation_json["Properties"]
         properties.setdefault("EventBusName", "default")
+
+        if "EventPattern" in properties:
+            properties["EventPattern"] = json.dumps(properties["EventPattern"])
+
         event_name = resource_name
 
         event_pattern = properties.get("EventPattern")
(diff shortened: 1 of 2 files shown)


bblommers (changes requested):
Thanks for the fix @skinitimski!
Can we get an assert on the EventPattern as well, to validate that this attribute is now returned correctly when calling `describe_rule`?

skinitimski:
> Can we get an assert on the EventPattern as well, to validate that this attribute is now returned correctly when calling `describe_rule`?

Good call.

Doneski https://github.com/spulec/moto/pull/4188/commits/d906f05930a5c961ee0edbbc40027274bb71151b

bblommers (approved):
LGTM! Thanks for the fix @skinitimski

PR #4192, opened 2021-08-18
Iot/added deprecation errors

arielb135:
create_thing and update_thing now raise exceptions if thing type is deprecated.
@bblommers - Who is responsible of adding the X to the implementation details? also the release notes - who keeps track of which new methods are implemented? (as i've seen 2.2.4 is not shown yet there)


bblommers:
Hi @arielb135, thanks for the PR!

> 
> @bblommers - Who is responsible of adding the X to the implementation details? also the release notes - who keeps track of which new methods are implemented? (as i've seen 2.2.4 is not shown yet there)

The maintainers will update both before a release, so don't worry about that 👍

bblommers (approved):

PR #4200, opened 2021-08-18
Fix bugs. Cloudformation updates are complex!

skinitimski:
* **Fix bug.** If a cloudformation stack is updated with a new parameter, that parameter should be honored. Several unit tests had bugs where they were not providing parameters required by the (updated) template. Solve this by by setting `self._template` in `ResourceMap.update`, and updating the unit tests.
* **Fix bug.** Do not update reload parameters until after all removed resources are deleted, so that any references from deleted resources to removed parameters can be resolved when deleting the resources.
* **Fix bug.** Per the API, creation of a change set should not modify a stack. The `ResourceMap.diff` method, called in the creation of a FakeChangeSet, was mutating the resource map which was problematic.

This is a complete rework of https://github.com/spulec/moto/pull/4187

### DIFF
--- a/moto/cloudformation/parsing.py
+++ b/moto/cloudformation/parsing.py
@@ -678,6 +678,9 @@ class ResourceMap(collections_abc.Mapping):
         self._resource_json_map = template["Resources"]
 
         for logical_name in resource_names_by_action["Add"]:
+
+            # call __getitem__ to initialize the resource
+            # TODO: usage of indexer to initalize the resource is questionable
             _ = self[logical_name]
 
         tries = 1
--- a/moto/cloudformation/parsing.py
+++ b/moto/cloudformation/parsing.py
@@ -606,63 +606,55 @@ class ResourceMap(collections_abc.Mapping):
                     [self[resource].physical_resource_id], self.tags
                 )
 
-    def diff(self, template, parameters=None):
-        if parameters:
-            self.input_parameters = parameters
-        self.load_mapping()
-        self.load_parameters()
-        self.load_conditions()
+    def build_resource_diff(self, other_template):
 
-        old_template = self._resource_json_map
-        new_template = template["Resources"]
+        old = self._resource_json_map
+        new = other_template["Resources"]
+
+        resource_names_by_action = {"Add": {}, "Modify": {}, "Remove": {}}
 
         resource_names_by_action = {
-            "Add": set(new_template) - set(old_template),
+            "Add": set(new) - set(old),
             "Modify": set(
-                name
-                for name in new_template
-                if name in old_template and new_template[name] != old_template[name]
+                name for name in new if name in old and new[name] != old[name]
             ),
-            "Remove": set(old_template) - set(new_template),
+            "Remove": set(old) - set(new),
         }
+
+        return resource_names_by_action
+
+    def build_change_set_actions(self, template, parameters):
+
+        resource_names_by_action = self.build_resource_diff(template)
+
         resources_by_action = {"Add": {}, "Modify": {}, "Remove": {}}
 
         for resource_name in resource_names_by_action["Add"]:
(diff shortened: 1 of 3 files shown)


moto/cloudformation/parsing.py
```diff
@@ -676,10 +668,22 @@ def update(self, template, parameters=None):
...
+        self._template = template
+        if parameters:
+            self.input_parameters = parameters
+        self.load_mapping()
+        self.load_parameters()
+        self.load_conditions()
+
+        self._resource_json_map = template["Resources"]
+
+        for logical_name in resource_names_by_action["Add"]:
```
bblommers, line 680: What does this for-loop do?

bblommers:
> 
>     * **Fix bug.** If a cloudformation stack is updated with a new parameter, that parameter should be honored. Several unit tests had bugs where they were not providing parameters required by the template. Do this by setting `self._template` in `ResourceMap.update`.

Am I correct in assuming that this will now throw an error, if a user does not supply a value for a new template-parameter?
If so, we should have a unit-test to verify this behaviour, and check the correct error is raised

In general, would you mind adding a few assertions to verify that parameters are now treated correctly. From what I can see, the tests currently are only altered just enough to not fail.

moto/cloudformation/parsing.py
skinitimski: The `__getitem__` method is lazy and creates resources if they do not yet exist. It has the same initialization steps that were in the original loop (original lines 657-662). This was an attempt to DRY that out and use only one set of initialization steps.

It looks pretty silly though.

I could refactor the initialization out of `__getitem__` and into its own method, then call that method here, if that makes more sense.

bblommers: Ah, right, I see. It does look pretty silly, but I guess that's just because it is a non-intuitive way of resource creation in the first place.
I'm OK with keeping it like this - just put a comment there to explain it, please, to make it easier for whoever comes across this code next.

skinitimski:
> Am I correct in assuming that this will now throw an error, if a user does not supply a value for a new template-parameter? If so, we should have a unit-test to verify this behaviour, and check the correct error is raised

Yep, it should. [Cases added](https://github.com/spulec/moto/pull/4200/commits/2ce1de33f8403d0c454f7135a5abfae57014a9de).

> In general, would you mind adding a few assertions to verify that parameters are now treated correctly. From what I can see, the tests currently are only altered just enough to not fail.

Sure. I was using my consuming project to gain confidence, but that does not help moto. Added [a couple more test cases](https://github.com/spulec/moto/pull/4200/commits/ffdf52b6f9140ad38b78b58de5a28f4e5d0d1893) (plus https://github.com/spulec/moto/pull/4200/commits/f420abb66351e901de3e00a600e921f6de5e8513, https://github.com/spulec/moto/pull/4200/commits/7ccaaf3ed996d27ada2dc14e019a45dc4fbd726d), more specific to the problem I set out to fix.

moto/cloudformation/parsing.py
skinitimski: Doneski. https://github.com/spulec/moto/pull/4200/commits/0f7aed7ecf8b124d3274e329a3826da0bf806a5a

I am still trying to understand it, but I am starting to think that `update` should not be trying to initialize added resources at all, given the lazy nature of resources in ResourceMap. Perhaps a prior author of `update` did not want it to be lazy or did not code it correctly to be lazy? Not sure. I will look to follow up with another PR to do something about it.

bblommers: As far as I know, it's pretty close on par with how AWS behaves. And given the complexity of it all, I'm just happy that it works 🙂

bblommers (approved):
Thanks for the PR and all the work, @skinitimski - much appreciated!

Issue #2339, opened 2019-07-26
S3 PUT etag is incompatible with dotnet client


davidje13:
This appears to be the same issue raised in #1219 (though that was closed without resolution)

When PUTting a non-empty file from the AWSSDK.S3 package to a moto server, an exception is thrown by the client:

> ```
> One or more errors occurred. (Expected hash not equal to calculated hash)
> [...]
> ----- Inner Stack Trace -----
>    at Amazon.S3.Internal.AmazonS3ResponseHandler.CompareHashes(String etag, Byte[] hash) in ...\AWSDotNetPublic\sdk\src\Services\S3\Custom\Internal\AmazonS3ResponseHandler.cs:line 213
>    at Amazon.S3.Internal.AmazonS3ResponseHandler.ProcessResponseHandlers(IExecutionContext executionContext) in ...\AWSDotNetPublic\sdk\src\Services\S3\Custom\Internal\AmazonS3ResponseHandler.cs:line 89
>    at Amazon.S3.Internal.AmazonS3ResponseHandler.InvokeAsync[T](IExecutionContext executionContext) in ...\AWSDotNetPublic\sdk\src\Services\S3\Custom\Internal\AmazonS3ResponseHandler.cs:line 60
> ```

We integrate like so:

```csharp
var client = new AmazonS3Client(
    new BasicAWSCredentials("A", "B"),
    new AmazonS3Config
    {
        ForcePathStyle = true, // required to match moto API
        ServiceURL = "http://localhost:5000",
    }
)

await client.PutBucketAsync("test");

var response = await client.PutObjectAsync(new PutObjectRequest
{
    BucketName = "test",
    Key = "file",
    InputStream = new MemoryStream(new byte [] { 1, 2, 3 })
});
```

Talking to the real S3 service with the same code works, and talking to moto using the AWS CLI works. It is only the combination of the dotnet client and moto which fails.

Investigation reveals this is due to a different `etag` calculation in moto. I believe this is something to do with which data is included in the hash algorithm. As a test, I disabled etag generation entirely inside moto:

```patch
diff --git a/moto/s3/models.py b/moto/s3/models.py
index 2a628d68..74fc76a1 100644
--- a/moto/s3/models.py
+++ b/moto/s3/models.py
@@ -145,6 +145,7 @@ class FakeKey(BaseModel):

     @property
     def etag(self):
+        return ""
         if self._etag is None:
             value_md5 = hashlib.md5()
             self._value_buffer.seek(0)
```

[...]

Issue #1219, opened 2017-09-28
"Expected hash not equal to calculated hash" error with S3 put object


copenhas:
First I would like to thank all of you for such a comprehensive AWS mocking library... now on to my issue.

I'm getting an "Expected hash not equal to calculated hash" exception thrown from AWSSDK for .NET when using moto's stand alone s3 server. This always happens when putting an object into the server. 

It appears to be that moto returns an ETag with the wrong (?) hash value in it. I've debugged the AWSSDK to find the [line where the check is made](https://github.com/aws/aws-sdk-net/blob/master/sdk/src/Services/S3/Custom/Internal/AmazonS3ResponseHandler.cs#L137) and verified the that calculated hash matches what I get when calculating the hash manually in .NET and in Python. My Python code is based on the [etag code in moto](https://github.com/spulec/moto/blob/master/moto/s3/models.py#L85).

I have all the relevant code and my test file uploaded to a gist:
https://gist.github.com/copenhas/677cf1b1066af1c53fce9feaa9ff0191

The test file has no line endings, but before it did have multiple Windows line endings creating whitespace at the bottom. This caused differences in hash values between my .NET and Python code. I removed them as a test and found that I was able to generate the same hash value between my .NET and Python code. I'm not positive but Python might be stripping out the carriage return character which could cause some Windows compatibility issues with this library.

Output of the C# program (I've tried .NET Core 2.0 and .NET Full Framework 4.6.1)
```
.NET MD5: "32A0E5FD908F68255E88DA135632BD5F"
MotoTest
Expected hash not equal to calculated hash
Moto MD5: "8929c640fcd7b0a9b8800f6852df4f93"
```

Output of Python script:
```
PS C:\> python .\md5.py
"32a0e5fd908f68255e88da135632bd5f"
```

Exception details coming out of AWSSDK:
```
Amazon.Runtime.AmazonClientException occurred
  HResult=0x80131500
  Message=Expected hash not equal to calculated hash
  Source=<Cannot evaluate the exception source>
  StackTrace:
   at Amazon.S3.Internal.AmazonS3ResponseHandler.CompareHashes(String etag, Byte[] hash) in E:\JenkinsWorkspaces\v3-stage-release\AWSDotNetPublic\sdk\src\Services\S3\Custom\Internal\AmazonS3ResponseHandler.cs:line 213
[...]

PR #4201, opened 2021-08-19
Added handling for V4 signatures in PutObject body

nluk:
Hi, first of all a bit of background:
- I'm using moto in server mode, to mock AWS in java integration tests
- AWS SDK v2 PutObject failed, due to failing ETag [validation](https://github.com/aws/aws-sdk-java-v2/blob/584ccb59e770177aeaa4c3b6bda4e24015b8ece9/services/s3/src/main/java/software/amazon/awssdk/services/s3/checksums/ChecksumsEnabledValidator.java#L147)
- Upon some debugging, I've found out that java client sends chunked body with [signatures](https://docs.aws.amazon.com/AmazonS3/latest/API/sigv4-streaming.html#sigv4-chunked-body-definition), for example executing:
```java
s3Client.putObject(putObjectRequest, RequestBody.fromString("CONTENT"))
```
is on moto side:
`b"7;chunk-signature=bd479c607ec05dd9d570893f74eed76a4b333dfa37ad6446f631ec47dc52e756\r\nCONTENT\r\n0;chunk-signature=d192ec4075ddfc18d2ef4da4f55a87dc762ba4417b3bd41e70c282f8bec2ece0\r\n\r\n"`.

Moto stores the whole body, including signature, same thing for ETag calculation.

This issue may be same as these:
- https://github.com/spulec/moto/issues/1219
- https://github.com/spulec/moto/issues/2339

Also, localstack seems to have fixed this on their side using a regex:
- https://github.com/localstack/localstack/issues/1571 
- https://github.com/localstack/localstack/issues/1611

In my case, their regex failed ('CONTENT' vs 'CONTENT\r\n'), so I've used more of a *parsing* solution. Feel free to discuss any impact it might have. I'm concerned, that having a second `new_body = bytearray()` can be an issue in terms of memory usage, but I'm not sure.

### DIFF
--- a/moto/s3/responses.py
+++ b/moto/s3/responses.py
@@ -1,5 +1,6 @@
 from __future__ import unicode_literals
 
+import io
 import os
 import re
 import sys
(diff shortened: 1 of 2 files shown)


bblommers (approved):
Nice @nluk - this feels like a good solution. The memory footprint is not ideal, but I can't think of a way to reduce this without completely rewriting our request handler.

Thanks for the PR and the investigation!

bblommers:
FYI @whummer - sounds like this fix can be removed from Localstack, now that it's present in Moto?

PR #4204, opened 2021-08-21
Fix IsDefault bug in elbv2 describe_rules

eoconnell:
Looks like a simple copy/paste mistake from the other templates.


bblommers (approved):
Nice catch @eoconnell - thanks for the PR!

Issue #4207, opened 2021-08-22
First request to moto server is slow


thrau:
The moto server starts up instantly when running `python -m moto.server`, but the first request will take a long time. For example:

```
 % time awslocal --endpoint=http://localhost:5000 s3 ls
awslocal --endpoint=http://localhost:5000 s3 ls  0,40s user 0,09s system 6% cpu 7,764 total
```

This call to the `s3` service took 7.7 seconds. Subsequent calls are faster (about 450 ms on my machine)

This has to do with the way moto lazy-loads the backends and searches for a backend based on the host url.
Currently, moto will, in the worst case, import all moto backends on the first http request sent to the moto server because of this method: https://github.com/spulec/moto/blob/7acdfe4febbdfd97389fdbcd0e63fad6ae581903/moto/server.py#L57-L72
Here, `backends.search_backend` needs to import the backend using `_import_backend` to check whether the host matches the backend `url_bases` defined in `moto.<backend>.urls`.
In this case, the host will be `localhost`, so this part here: https://github.com/spulec/moto/blob/7acdfe4febbdfd97389fdbcd0e63fad6ae581903/moto/server.py#L147-L151
will import all backends in alphabetical order (including cloudformation, which alone takes about 2.5 seconds on my machine to import), just to find that there's no backend for the host "localhost".

As discussed with @bblommers, having a static index between the backends and the `url_bases` would be useful to further defer calls to importlib until the backend is actually needed.

bblommers:
This should be fixed in 2.2.7 👍

PR #4209, opened 2021-08-22
Add url_bases index for faster backend lookup

thrau:
This PR address #4207 (as good as it gets imo, gotta import at some point...)

As discussed with @bblommers, it introduces an index between the backend and the `url_bases` attribute defined in the respective backend `urls.py`. The index is constructed at startup time by parsing the `urls.py` files of the backends in plain text (this is blazingly fast, takes a few milliseconds, despite `eval`). This also speeds up subsequent calls quite a bit, as we no longer call importlib and `getattr` for the backend lookup.

I also tried this: but clearly this will again end up importing all parent modules due to relative imports in `urls.py`, so not useful at startup time. however it would be more robust if used to build the index at build time and store it somewhere.
```python
def build_url_index_import():
    index = dict()

    for name, (module_name, _) in BACKENDS.items():
        module = importlib.import_module(f"moto.{module_name}.urls")
        index[name] = getattr(module, "url_bases")

    return index
```

/cc @whummer


thrau:
reflecting on this now, i see that assuming you have access to the source files in this way (using `os.path.dirname(__file__)/...` may be problematic for scenarios like packaging moto in archives.

moto/backends.py
```diff
@@ -122,3 +123,51 @@ def get_model(name, region_name):
...
+def build_url_index():
+    """
+    Builds an index between the backend name and the associated url_bases patterns.
+    :rtype: Dict[str, List[str]]
+    """
+    index = dict()
+    here = os.path.dirname(__file__)
+
+    for name, (module_name, _) in BACKENDS.items():
+        file_path = os.path.join(here, module_name, "urls.py")
```
thrau, line 137: reflecting on this now, i see that assuming you have access to the source files in this way may be problematic for scenarios like packaging moto in archives.

any thoughts?

bblommers: Yeah - maybe that's the reason TF is failing? (Don't have time to test it out myself atm..)
You could try this instead:
```
from moto.utilities.utils import load_resource
load_resource(__name__, f"{module_name}/urls.py", as_json=False)
```
https://github.com/spulec/moto/blob/d278fd6eaad3194b6d2347876d4fe24948f1cb2c/moto/utilities/utils.py#L22:L29

Rebasing with master should fix the linter - I just pushed a fix for that earlier.

thrau: unfortunately that uses `pkgutil`, which will again cause importlib to actually load the modules, which is what we want to avoid.
it seems to me that there's no way around a build-time step that creates this index and packages it with the distribution.

thrau:
So I amended the commit to do something completely different ;-)

It now adds a script `scripts/update_backend_index.py` that generates the index into `moto/backend_index.py` (could be extended in the future), and auto-formats it. the script should be deterministic and idempotent, so we can check in the generated file into version control, and keep it up to date by simply re-running the script.

the only drawback i see now is that it's not clear when the script should be executed, and who executes it.

thrau:
long-term, it may be worth thinking about adding something like a `BackendSpec` or `BackendMeta` concept, that encapsulates this type of information (and also what's currently in `backends.BACKENDS`), and that can be easily located by indexing mechanisms.

thrau:
@bblommers do you see this as a viable solution, or do you have concerns? i understand it's at the guts of the dispatcher, and i'm happy to discuss alternatives :-)

bblommers (approved):
LGTM - thanks @thrau!

thrau:
hey @bblommers, i merged your changes into the PR and rebased. i fixed a couple of conflicts so would be good to do one more round of checks :-)

thrau:
ah i was too slow! ;-) thanks bert.

the only minor thing we should maybe revisit is the print statement. `curl http://localhost:5000` will lead to

```
Unable to find appropriate URL for {}.Remember to add the URL to urls.py, and run script/update_backend_index.py to index it.
```

bblommers:
@thrau I was just going to test this change in production :slightly_smiling_face: 

Thanks for the heads up - I'll fix the print statement in an upcoming PR.

PR #4210, opened 2021-08-22
added support for modify-vpc-connection; fixed vpc peering options

macnev2013:
## Added
- Support for `ModifyVpcPeeringConnectionOptions`
- Support for `VPCPeeringConnectionOptions`

## Fixed
- `peeringOptions` in all apis which contained hardcoded values
- Fixed account ID (Here we are checking if the vpc exsists or not! Code will not be able to find VPC from different account. So we have assumed we are only peering to the current account only ---- TF test wont pass without this test)

## TF_Test
- `TestAccAWSRouteTable_IPv4_To_VpcPeeringConnection`


macnev2013:
Hi @bblommers , Can you review this PR?

bblommers (approved):
LGTM! Thanks @macnev2013

PR #4213, opened 2021-08-23
Minor updates to scaffold.py

kbalk:
- The note output at the end of the script regarding additional files to modify was corrected to remove moto/\_\_init\_\_.py (it's updated by scaffold.py) and to add the document files (docs/index.rst and the implementation readme).
- Made template/test/\_\_init\_\_.py a template file so it would be generated correctly.  Previously an "\_\_init\_\_" file was generated.
- Fixed some pylint issues in scaffold.py.
- Added documentation strings to the templates in the hopes it would encourage that practice.
- Removed "from \_\_future\_\_" statements in templates.


bblommers (approved):
Nice one @kbalk 👍

PR #4214, opened 2021-08-24
Only use backported importlib_metadata on Python < 3.8

andersk:
[`importlib.metadata`](https://docs.python.org/3/library/importlib.metadata.html) is in the standard library in Python ≥ 3.8, so we can avoid installing the backported `importlib_metadata` except on older Python versions, using an [environment marker](https://www.python.org/dev/peps/pep-0496/).


bblommers (approved):
LGTM! Thanks for the PR @andersk

PR #4216, opened 2021-08-24
Cleanup pulldown script a bit, update managed rules

kbalk:
Thought `pull_down_aws_managed_rules.py` could have been written a little better, so I made a few corrections.  When the script was rerun for testing purposes, the JSON containing managed rules showed some updates, so these are included as well.


bblommers (approved):
LGTM!

PR #4232, opened 2021-08-27
enhancement(SSM): Add DataType attribute to SSM Parameters

### LINKED ISSUE #4215: ssm.get_parameter response contains no field DataType
The ssm [get_parameter](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/ssm.html#SSM.Client.get_parameter) function returns a field called `DataType`.  However `mock_ssm()` does not.

 ```python
import os

import boto3
import pytest
from moto import mock_ssm

@pytest.fixture
def aws_credentials():


@pytest.fixture
def ssm_mock(aws_credentials):
    with mock_ssm():
        client = boto3.client("ssm", region_name="us-west-2")
        client.put_parameter(
            Name='parameter-name',
            Description='parameter-description',
            Value='parameter-value',
            Type='String',
            DataType='string'
        )
        yield


def test_get_parameter(ssm_mock):
    ssm = boto3.client("ssm", region_name="us-west-2")
    actual = ssm.get_parameter(Name="parameter-name", WithDecryption=False)

    assert "DataType" in actual["Parameter"]
```

Results

```python
>       assert "DataType" in actual["Parameter"]
E       AssertionError: assert 'DataType' in {'ARN': 'arn:aws:ssm:us-west-2:123456789012:parameter/parameter-name', 'LastModifiedDate': datetime.datetime(2021, 8, 23, 20, 5, 10, 520000, tzinfo=tzlocal()), 'Name': 'parameter-name', 'Type': 'String', ...}
```

jlopezscala:
fixes #4215 

This PR adds the `DataType` attribute to SSM parameters #4215 

`DataType` parameter on SSM is now added to the Parameter class. The parameter is now accepted and validated on the `put_parameter` function and also returned back when getting parameters from the store.

`aws:ec2:image` Data Type parameters are not created instantly on AWS. It depends on the existence of an AMI ID of the provided value so it can keep the value updated. This is still not supported.

`DataType` defaults to `text` as the documentation describes.

### DIFF
--- a/tests/test_ssm/test_ssm_boto3.py
+++ b/tests/test_ssm/test_ssm_boto3.py
@@ -232,13 +232,11 @@ def test_get_parameters_by_path():
 @mock_ssm
 def test_put_parameter(name):
     client = boto3.client("ssm", region_name="us-east-1")
-    data_type = "text"
     response = client.put_parameter(
         Name=name,
         Description="A test parameter",
         Value="value",
         Type="String",
-        DataType=data_type,
     )
 
     response["Version"].should.equal(1)
@@ -250,7 +248,7 @@ def test_put_parameter(name):
     response["Parameters"][0]["Value"].should.equal("value")
     response["Parameters"][0]["Type"].should.equal("String")
     response["Parameters"][0]["Version"].should.equal(1)
-    response["Parameters"][0]["DataType"].should.equal(data_type)
+    response["Parameters"][0]["DataType"].should.equal("text")
     response["Parameters"][0]["LastModifiedDate"].should.be.a(datetime.datetime)
     response["Parameters"][0]["ARN"].should.equal(
         "arn:aws:ssm:us-east-1:{}:parameter/{}".format(ACCOUNT_ID, name)
@@ -276,7 +274,7 @@ def test_put_parameter(name):
     response["Parameters"][0]["Value"].should.equal("value")
     response["Parameters"][0]["Type"].should.equal("String")
     response["Parameters"][0]["Version"].should.equal(1)
-    response["Parameters"][0]["DataType"].should.equal(data_type)
+    response["Parameters"][0]["DataType"].should.equal("text")
     response["Parameters"][0]["LastModifiedDate"].should.equal(
         initial_modification_date
     )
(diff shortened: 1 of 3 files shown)


tests/test_ssm/test_ssm_boto3.py
```diff
@@ -394,19 +407,42 @@ def test_put_parameter_china():
...
 def test_get_parameter():
     client = boto3.client("ssm", region_name="us-east-1")
 
     client.put_parameter(
-        Name="test", Description="A test parameter", Value="value", Type="String"
+        Name="test",
+        Description="A test parameter",
+        Value="value",
+        Type="String",
+        DataType="text",
```
bblommers, line 437: If DataType defaulfs to text, I don't think we need to change this invocation.
The `test_put_parameter` is modified so that we test that the parameter is supported, so all other tests do no need to pass the parameter - just the assertion that the default value is returned.

tests/test_ssm/test_ssm_boto3.py
```diff
@@ -418,10 +454,18 @@ def test_get_parameter_with_version_and_labels():
     client = boto3.client("ssm", region_name="us-east-1")
 
     client.put_parameter(
-        Name="test-1", Description="A test parameter", Value="value", Type="String"
+        Name="test-1",
+        Description="A test parameter",
+        Value="value",
+        Type="String",
+        DataType="text",
```
bblommers, line 461: Same

tests/test_ssm/test_ssm_boto3.py
```diff
@@ -418,10 +454,18 @@ def test_get_parameter_with_version_and_labels():
...
     )
     client.put_parameter(
-        Name="test-2", Description="A test parameter", Value="value", Type="String"
+        Name="test-2",
+        Description="A test parameter",
+        Value="value",
+        Type="String",
+        DataType="text",
```
bblommers, line 468: Same

bblommers (changes requested):
Hi @jlopezscala, thanks for the PR! The logic looks good to me - just a minor nitpick on the test cases

tests/test_ssm/test_ssm_boto3.py
jlopezscala: Good call! Fixed!

bblommers:
Looks like the linter is failing @jlopezscala - you'll have to run black to get the CI to pass:
```
pip install black==19.10b0
make format
```
(Make sure to use that specific version of black , as that's what our CI uses - different versions will produce different formatting.)

bblommers (approved):
Nice one - thanks again for the PR @jlopezscala!

bruce-szalwinski:
Thanks @jlopezscala!

jlopezscala:
Glad to contribute! Thanks for reviewing 🚀

PR #4239, opened 2021-08-28
SecurityGroup Improvements

monty16597:
# Improvements
- Improved SourceGroup implementation in SG.
- Fixed `AuthorizeSecurityGroupEgress` and `AuthorizeSecurityGroupIngress` response.
- SecurityGroup Drifting properties


monty16597:
@bblommers Can you help me with the Response Tests?

monty16597:
Hey @bblommers @whummer ,
We are ready to merge this PR.

monty16597:
Hello @bblommers , Please let me know if anything is needed.

moto/ec2/models.py
```diff
@@ -2019,13 +2019,13 @@ def __init__(
         to_port,
         ip_ranges,
         source_groups,
-        prefix_list_ids=None,
+        prefix_list_ids=[],
```
bblommers, line 2022: Default arguments should always be None, IMO, as having a list as default can lead to really tricky bugs.
(https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments if you want a clear/concise summary of what can go wrong)

moto/ec2/models.py
```diff
@@ -2056,11 +2058,18 @@ def owner_id(self):
     def __eq__(self, other):
         if self.ip_protocol != other.ip_protocol:
             return False
-        if self.ip_ranges != other.ip_ranges:
+        ip_ranges = [item for item in self.ip_ranges if item not in other.ip_ranges]
```
bblommers, line 2061: Should we use `set.symmetric_difference()` here to guarantee equality? 
Using the current comparison, the following scenario would still pass, if I'm not mistaken:
```
self.ip_ranges = [{"CidrIp": "0.0.0.0/0"}]
other.ip_ranges = [{"CidrIp": "1.1.1.1/0"}]
```

moto/ec2/models.py
```diff
@@ -2338,9 +2361,8 @@ def authorize_security_group_ingress(
         from_port,
         to_port,
         ip_ranges,
-        source_group_names=None,
-        source_group_ids=None,
-        prefix_list_ids=None,
+        source_groups=[],
```
bblommers, line 2364: Same issue with the default arg

moto/ec2/models.py
```diff
@@ -2407,30 +2443,54 @@ def revoke_security_group_ingress(
...
```
bblommers, line 2446: Same issue with the default arg

moto/ec2/models.py
```diff
@@ -2442,9 +2502,8 @@ def authorize_security_group_egress(
...
```
bblommers, line 2505: Same issue with the default arg

moto/ec2/models.py
```diff
@@ -2515,24 +2589,14 @@ def revoke_security_group_egress(
...
```
bblommers, line 2592: Same issue with the default arg

bblommers (changes requested):
Apologies for the delay @monty16597 - please see my comments

monty16597:
@bblommers, I have updated the changes.

monty16597:
Also, @bblommers , I wanted to ask if we do support `EC2-classic`, AWS is no more supporting those features, So every time we create SecurityGroup without a `VPC_ID` AWS will consider the default `VPC` for the SG. So wonder if we can do that the same. Any thought?

bblommers:
> 
> 
> Also, @bblommers , I wanted to ask if we do support `EC2-classic`, AWS is no more supporting those features, So every time we create SecurityGroup without a `VPC_ID` AWS will consider the default `VPC` for the SG. So wonder if we can do that the same. Any thought?

There is still a grace period for EC2-classic users, afaik, but there can't be that many users, it doesn't make sense to support that behaviour as our default. So creating SG's in the default VPC sounds good to me.

I'll merge this for now, as it's getting quite big already, but if you want to implement that in a future PR, that would be very welcome 👍

bblommers (approved):
LGTM! Thanks for the PR @monty16597

PR #4245, opened 2021-08-30
DynamoDB - Throw exception when updating key using SET operation

### LINKED ISSUE #3877: Wrong error message when trying to update hashkey or range key in dynamodb
# Steps to Reproduce

## With moto

```
import boto3
table_name = 'test-key-update'

with mock_dynamodb2():

    client = boto3.client('dynamodb')

    client.create_table(
        TableName=table_name,
        KeySchema=[{"AttributeName": "forum_name", "KeyType": "HASH"}],
        AttributeDefinitions=[{"AttributeName": "forum_name", "AttributeType": "S"}],
        )


    client.update_item(
        TableName=table_name,
        Key={
            'forum_name': {
                'S': "x"
            },
        },
        UpdateExpression="set forum_name=:New",
        ExpressionAttributeValues={":New": {"S": "x"}},
    )
```

Run this, and then run the same thing again, but without `mock_dynamodb2()`.

# Expected Behavior

The two exceptions are the same

# Actual behavior

Moto error:

> botocore.exceptions.ClientError: An error occurred (ValidationException) when calling the UpdateItem operation: One or more parameter values were invalid: An AttributeValue may not contain an empty string

Boto error:

> botocore.exceptions.ClientError: An error occurred (ValidationException) when calling the UpdateItem operation: One or more parameter values were invalid: Cannot update attribute forum_name. This attribute is part of the key

(Exact same error for range key update in boto by the way.)

# Consequence

The purpose of moto is to test code that uses boto. In this case, if the code under test is faulty, moto gives a very misleading error.

bblommers:
Fixes #3877 

Adds validation when trying to update the tables' HASH or RANGE key using the UpdateExpression.


tests/test_dynamodb2/test_dynamodb_table_with_range_key.py
```diff
@@ -2069,3 +2070,48 @@ def test_scan_by_index():
...
     assert last_eval_key["range_key"]["S"] == "1"
     assert last_eval_key["lsi_range_key"]["S"] == "1"
+
+
+@mock_dynamodb2
+@pytest.mark.parametrize("create_item_first", [False, True])
+@pytest.mark.parametrize(
+    "expression", ["set h=:New", "set r=:New", "set x=:New, r=:New"]
+)
+def test_update_item_throws_exception_when_updating_hash_or_range_key(
```
bblommers, line 2080: Test is verified against AWS

tests/test_dynamodb2/test_dynamodb.py
```diff
@@ -5932,6 +5932,54 @@ def test_gsi_key_can_be_updated():
     item["main_key"].should.equal({"S": "testkey1"})
 
 
+@mock_dynamodb2
+def test_gsi_key_cannot_be_empty():
```
bblommers, line 5936: This is verified against AWS as well.

tests/test_dynamodb2/test_dynamodb_validation.py
```diff
@@ -42,27 +42,6 @@ def test_valid_update_expression(table):
     ).validate()
 
 
-def test_validation_of_empty_string_key_val(table):
```
bblommers, line 45: This test originally verified that a key cannot be empty. Although true, it should also never get to this point - as keys cannot be set in the first place.

An index key *can* be set, but can never be empty. This behaviour is validated in the new boto3-test, 'test_gsi_key_cannot_be_empty'.

PR #4246, opened 2021-08-30
Add some basic Firehose service APIs

kbalk:
Implements the Firehose service APIs:

- `create_delivery_stream`
- `delete_delivery_stream`
- `describe_delivery_stream`
- `list_delivery_streams`
- `list_tags_for_delivery_stream`
- `put_record`
- `put_record_batch`
- `tag_delivery_stream`
- `untag_delivery_stream`
- `update_destination`

These changes do not allow Kinesis to be used as a source for Firehose; that would need to be implemented as a separate task.  These changes also don't allow CloudWatch to stream logs into Firehose; again, that still needs to be implemented.


kbalk:
@bblommers I need to validate tags for firehose and I noticed there's some nice logic to do so in `moto/moto/resourcegroups/models.py` https://github.com/spulec/moto/blob/cc568c16565e5e9e895a73ce6389b58b5406b0a5/moto/resourcegroups/models.py#L124  

There's also tag validation in `moto/moto/config/models.py` and I think there's one other place where that logic resides.  I could create a copy of the ResourceGroup's validation logic (with some slight modifications since it's a class method) and move it to a common location and then use it for firehose.  That common code code could then be used for other services.  But I'm not sure of the best place to put that common code.

The tag validation logic could be added to the `BaseModel` or `BaseBackend` class.  Or there's a `moto/utilities` directory with odds and ends.  Or `moto/core/utils.py`.  I like the idea of adding it to `BaseModel`, but I wanted to get your opinion.

bblommers:
Is it possible to use `moto/utilities/taggin_service.py`?
The idea behind that class was to have a single unified approach, but we never got around to implementing it retroactive in all services. It does sound like a good candidate to contain shared validation.

kbalk:
> Is it possible to use `moto/utilities/taggin_service.py`?
> The idea behind that class was to have a single unified approach, but we never got around to implementing it retroactive in all services. It does sound like a good candidate to contain shared validation.

Ok.  I wish I knew about that class when I added tagging functions to ConfigService.

kbalk:
@bblommers I'm using black-21.8b0 on my system and seeing no reformatting issues, but the CI tests are using black-19.10b0 and showing that three of my files as needing formatting.  I could revert my version of `black`, but thought I would check with you to see if there were plans to update `black`.

bblommers:
> 
> 
> @bblommers I'm using black-21.8b0 on my system and seeing no reformatting issues, but the CI tests are using black-19.10b0 and showing three of my files as needing formatting. I could revert my version of `black`, but thought I would check with you to see if there were plans to update `black`.

Hi @kbalk, we'll stick with 19.10b0 for a while. Last I checked there were some bugs in the later versions, and I don't want to upgrade until those are resolved.

kbalk:
@bblommers With my last commit, I no longer had lint errors so all of the moto unit tests were run.  Several tests failed and the failures were related to Kinesis logic that contained firehose code:

https://github.com/spulec/moto/blob/7240e8f9657601e8e6e1b3e1f923dd28819cf7b9/moto/kinesis/models.py#L625-L647

... plus some Firehose functions related to tags.

There were some errors in other code that I still need to look at, i.e. problems with the service name and S3 buckets.

In any case, I'm a bit confused at this point.  If I go back to the original problem for this PR, maybe I could just point to the Kinesis logic and not use the Firehose logic I developed.  Or, sort through the Kinesis logic and remove the Firehose logic so it uses the new Firehose logic.  What would you recommend?  I'm also worried about the amount of time this is consuming.

Currently, the work that remains on this PR includes implementing the `put_record` and `put_records` APIs, and adding logic in `create_delivery_stream` to process the Kinesis source parameter.

moto/firehose/urls.py
```diff
@@ -0,0 +1,6 @@
+"""Firehose base URL and path."""
+from .responses import FirehoseResponse
+
+
+url_bases = ["https?://firehose.(.+).amazonaws.com"]
```
bblommers, line 5:
```suggestion
url_bases = ["https?://firehose\.(.+)\.amazonaws.com"]
```
This regex was wrong for all other services as well - we should ensure it matches a literal `.`, otherwise it will match any character. This should fix the test_bucketname_starting_with_service_name[firehose-]-test that's currently failing

kbalk: I wondered about that, but when I checked other services, some escaped the dots, some didn't.  I'll fix the scaffold tool as well.  But why does it work for the other services and fail for firehose?

bblommers:
>
> Or, sort through the Kinesis logic and remove the Firehose logic so it uses the new Firehose logic.

That would be the endgoal, I suppose, but: Rome wasn't build in a day. There's nothing wrong with leaving the Kinesis and Firehose logic separate for now. When someone else really needs this to be integrated, I'm sure they can put in the work.

> 
> I'm also worried about the amount of time this is consuming.

Well, you've put a lot of work in Moto in the past few months, and submitted a enormous amount of features - that's already incredibly appreciated. Don't feel bad for leaving some work for the rest of us 🙂 

Alternatively, if this PR is not something you need ASAP, there's always the option of marking it with a `needs-help` label. That will let others know that it's a work in progress, and they can help and contribute with the remaining features.

moto/firehose/urls.py
bblommers: I submitted a PR some time ago that was supposed to fix all of them - do you remember which services do not escape it yet?

kbalk:
The other thing I was wondering about were the fields shown in the `describe_delivery_stream` output.  Localstack and this PR only show the destination fields that were provided as input for `create_delivery_stream`.  But boto3 shows all the possible fields for a destination for `describe_delivery_stream`.  I'll have to check what the Kinesis code is doing.  But I think it would be good to mimic boto3.  I shouldn't be a big deal to add the other fields as it's just busy work to supply all the fields.

moto/firehose/urls.py
kbalk: There's a long list:

```
cd moto/moto
find . -name "urls.py"  | xargs grep "(.+).amazonaws"
```

kbalk:
From what I'm seeing in the failed unit tests, there's confusion with two backends declaring an endpoint for firehose.  I've transitioned some of the kinesis firehose-specific unit tests and started editing out firehose from kinesis, but I need to complete the `put_record` and `put_record_batch` before I can completely remove firehose from kinesis.

I found some mistakes in my logic as a result of looking at the Kinesis logic, so there is value in the exercise.

bblommers:
Quick FYI @kbalk : we're making some changes to the way modules are loaded.
After rebasing with master, you'll probably run into some merge issues. I'm expecting git will complain about the following files:

`moto/__init__.py` - There are some changes to this file, but you're addition is still correct
`moto/backends.py` - This file no longer needs to be updated, so you can remove your addition. (This information now comes from `moto/__init__.py`)
` moto/backend_index.py`: This file needs to be updated with the new URL's for kinesis/firehose. Update it manually, or call `scripts/update_backend_index.py` to update it automatically.

Some background on why these changes were made can be found in [this PR](https://github.com/spulec/moto/pull/4209).

Apologies for the inconvenience - let me know if you run into any issues.

kbalk:
@bblommers I saw the PR and will make the updates.  Appreciate the heads up!

kbalk:
@bblommers The scaffolds script is still editing `backends.py`:

https://github.com/spulec/moto/blob/b3b326f578dba985978a36ce0201c91a6afab04e/scripts/scaffold.py#L145

bblommers:
Good catch @kbalk  - thanks for letting me know.
That method is no longer used, so it shouldn't break anything, but I'll remove it soon.

kbalk:
> Good catch @kbalk - thanks for letting me know.
> That method is no longer used, so it shouldn't break anything, but I'll remove it soon.

But it should update `backends_index.py`.

bblommers:
For a lot of services the developer has to manually edit the URLs anyway, so the least confusing workflow IMO is to let the developer fix all the URLs, and then let them run the update_backend_index script to auto populate the `backends_index.py`

kbalk:
> For a lot of services the developer has to manually edit the URLs anyway, so the least confusing workflow IMO is to let the developer fix all the URLs, and then let them run the update_backend_index script to auto populate the `backends_index.py`

Makes sense ... I forgot about that script.  There are messages the scaffold script prints at the end of the run -- perhaps a message to run that script?

kbalk:
@bblommers The firehose unit tests fail when running the ServerMode tests in the CI/CD environment.  They fail the first time the firehose boto3 client is invoked.  I'm not sure what I'm missing.  I did discover that `backend_index.py` had two listings for the firehose URL (the kinesis one was added back when that file was created today) and I fixed that in this PR, but it doesn't seem to matter.

Do you have any suggestions as to where I might look?

bblommers:
Looks like it's just the one test that fails, which suggests that the URLs are OK (if other firehose tests do pass).
I'll try to do some debugging either today or this weekend.

kbalk:
> Looks like it's just the one test that fails, which suggests that the URLs are OK (if other firehose tests do pass).
> I'll try to do some debugging either today or this weekend.

I was thinking the other firehose tests were failing too -- but that was before I fixed the URLs.  I feel stupid now!  

At one point I commented out that particular client call and the test then failed on the call that followed.  But again, that might have been before I cleaned up the URLs.  I'll have another go at it ... you've helped by pointing out that the other tests passed.

moto/firehose/models.py
```diff
@@ -0,0 +1,644 @@
...
+        if len(self.delivery_streams) == DeliveryStream.MAX_STREAMS_PER_REGION:
+            raise LimitExceededException(
+                f"You have already consumed your firehose quota of "
+                f"{DeliveryStream.MAX_STREAMS_PER_REGION} hoses. Firehose "
+                f"names: {list(self.delivery_streams.keys())}"
+            )
+
+        # Rule out situations that are not yet implemented.
+        if delivery_stream_encryption_configuration_input:
+            raise NotImplementedError(
```
bblommers, line 211: This should be a warning IMO - we can still continue with the creation, it just will not have the full functionality

moto/firehose/models.py
```diff
@@ -0,0 +1,644 @@
...
+                "A delivery stream with server-side encryption enabled is not "
+                "yet implemented"
+            )
+
+        if destination_name == "Splunk":
+            raise NotImplementedError(
```
bblommers, line 217: Same here with the warning

moto/firehose/models.py
```diff
@@ -0,0 +1,644 @@
...
+
+        delivery_stream = self.delivery_streams.get(delivery_stream_name)
+        if not delivery_stream:
+            raise ResourceNotFoundException(
+                f"Firehose {delivery_stream_name} under accountId "
+                f"{ACCOUNT_ID} not found."
+            )
+
+        if destination_name == "Splunk":
+            raise NotImplementedError(
```
bblommers, line 565: Same here with the warning

moto/firehose/models.py
```diff
@@ -0,0 +1,644 @@
...
+        delivery_stream = self.delivery_streams.get(delivery_stream_name)
+        if not delivery_stream:
+            raise ResourceNotFoundException(
+                f"Firehose {delivery_stream_name} under account {ACCOUNT_ID} "
+                f"not found."
+            )
+
+        self.tagger.delete_all_tags_for_resource(delivery_stream.delivery_stream_arn)
+
+        # The following logic is not applicable for moto as far as I can tell.
```
bblommers, line 268: Agreed - I think we can just delete the logic, no need to have it commented out

scripts/template/lib/urls.py.j2
```diff
@@ -2,7 +2,7 @@
 from .responses import {{ service_class }}Response
 
 url_bases = [
-    "https?://{{ endpoint_prefix }}.(.+).amazonaws.com",
+    r"https?://{{ endpoint_prefix }}\.(.+)\.amazonaws\.com",
```
bblommers, line 5: Nice one 👍

bblommers (changes requested):

tests/test_firehose/test_firehose.py
```diff
@@ -0,0 +1,453 @@
...
+#     client = boto3.client("firehose", region_name=TEST_REGION)
+#     s3_dest_config = sample_s3_dest_config()
+#     stream_name = "test_lookup"
+#
+#     arn = client.create_delivery_stream(
+#         DeliveryStreamName=stream_name, S3DestinationConfiguration=s3_dest_config,
+#     )["DeliveryStreamARN"]
+#
+#     from moto.firehose.models import firehose_backends
+#     delivery_stream = firehose_backends[TEST_REGION].lookup_name_from_arn(arn)
```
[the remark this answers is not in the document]
kbalk: I originally intended to add the logic to support sending CloudWatch logs to Firehose and that logic needed to find the delivery stream given the ARN.  But I pulled that code out and I plan on it being part of a separate PR.  Ironically, supporting the ability to handle logs in Fireose was the point of the original exercise!

I'll get rid of this commented code entirely and then add it back in with that PR.  

Thanks for the tip on how to skip the test -- I was looking for some sort of flag, but didn't think of the obvious one!  Thank you.

moto/firehose/models.py
kbalk: By warning, what do you mean?  moto doesn't seem to use logging.

moto/firehose/models.py
kbalk: Done.

scripts/template/lib/urls.py.j2
kbalk: Just doing my part ...

moto/firehose/models.py
```diff
@@ -0,0 +1,644 @@
...
+        prefix = s3_destination.get("Prefix", "")
+        object_path = self._format_s3_object_path(
+            delivery_stream_name, version_id, prefix
+        )
+
+        batched_data = b"".join([b64decode(r["Data"]) for r in records])
+        try:
+            s3_backend.put_object(bucket_name, object_path, batched_data)
+        except Exception as exc:
+            # This could be better ...
```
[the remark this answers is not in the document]
kbalk: It's interesting ... the boto3 documentation for s3 doesn't list the errors.  The moto code shows InvalidStorageClass or MissingBucket errors.  I could use "except ClientError as exc" and then just re-raise the ClientError exception?

moto/firehose/models.py
bblommers: Sorry, you're right, I was being too quick- I should have clarified. I was talking about the OOTB warning library:

```
import warnings
warnings.warn(...)
``` 

It would be the first instance of Moto using it, afaik!

moto/firehose/models.py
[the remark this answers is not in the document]
bblommers: No, it's OK as it is for now - I was just curious. 👍

moto/firehose/models.py
kbalk: I just checked and there are a couple other places using warnings - cloudformation, ec2, emr, events, elastictranscoder, packages.  I think it might be the first time I've used the warnings package.  

It makes sense to issue a warning for the encryption configuration, but I'm a bit concerned about just issuing a warning for a Splunk destination.  Any attempt to put records to that delivery stream will fail and it would be confusing.

bblommers: There's still a lot of value in testing whatever business logic leads up to the Splunk-connector though - that all becomes impossible if the `create_delivery_stream` throws an Exception.

Personally I wouldn't want to test the Splunk-connector anyway, or push data into Splunk as part of my unit tests and going to the trouble of connecting to it to verify the data has landed. I'd trust that AWS handles that part properly, and all I care about is that the surrounding logic is correct.

kbalk: Ok.  I've changed them to warnings.  Personally I find it frustrating when our Terraform tests fail and I'm not sure what caused the initial error.  But I'm working with a large number of repos that were already tested against AWS and we're now trying to automate those tests using moto.

moto/firehose/models.py
kbalk: Done.

moto/firehose/models.py
kbalk: Done.

bblommers (approved):
LGTM - thanks for all the work @kbalk!

moto/firehose/models.py
bblommers: Thanks. Yeah, I can understand it's frustrating. I expect the majority of users to have more straight forward test cases, and for them a warning will be more useful - hence my insistence 🙂

PR #4253, opened 2021-09-01
Add validation of tags to TaggingService

kbalk:
Got carried away pylinting tagging_service.py while making the additions.


bblommers (approved):
Nice improvement @kbalk - thanks!

PR #4254, opened 2021-09-01
Fix: Issue with empty tags created on SQS queues

mgagliardo:
Relates to https://github.com/localstack/localstack/issues/4343 and https://github.com/localstack/localstack/pull/4365

When tagging an existing SQS queue and one of the values is an empty string, this line breaks the loop because `not(k and v)` does not expect an empty str.

Example:
```
response = self.client.tag_queue(
    QueueUrl=queue_url, Tags={"tag1": "value1", "tag2": "value2", "tag3": ""}
)
```

With this change, it won't break the tag (and hopefully won't break either for the rest of the modules using this, as empty tags are accepted across AWS services AFAIK)

### DIFF
--- a/moto/core/responses.py
+++ b/moto/core/responses.py
@@ -723,7 +723,7 @@ class BaseResponse(_TemplateEnvironmentMixin, ActionAuthenticatorMixin):
                     elif key.endswith(value_end):
                         v = value[0]
 
-            if not (k and v):
+            if not (k and v is not None):
                 break
 
             results[k] = v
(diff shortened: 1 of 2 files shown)


bblommers:
Thanks for the PR @mgagliardo! Does this mean we can revert the logic in #4087? 
If I'm understanding it correctly, the tests in that PR should now be covered by this change as well.

mgagliardo:
> Thanks for the PR @mgagliardo! Does this mean we can revert the logic in #4087?
> If I'm understanding it correctly, the tests in that PR should now be covered by this change as well.

Thanks for the catch @bblommers! I have tested this and indeed it is feasible to revert the change (which I just did on this same PR). Thanks!

bblommers (approved):
LGTM! Thanks again @mgagliardo

Issue #1356, opened 2017-11-21
owner-alias not implemented when filtering AMI using describe_images


laurrentt:
As recommended by this error message:
> `FilterNotImplementedError: The filter 'owner-alias' for DescribeImages has not been implemented in Moto yet. Feel free to open an issue at https://github.com/spulec/moto/issues`

I'm opening this issue to keep track of the support for owner-alias filter in the describe_images method of the ec2 client. 

Here's a code example of the issue:
```
ec2_connection = boto3.client('ec2')
images = ec2_connection.describe_images(
    Filters=[
        {
            'Name': 'name',
            'Values': [
                'amzn-ami-hvm-*',
            ]
        },
        {
            'Name': 'owner-alias',
            'Values': [
                'amazon',
            ]
        },
    ]
)['Images']
```

And here's the exception raised by moto:
```
caught_exception = FilterNotImplementedError(u"The filter 'owner-alias' for DescribeImages has no...oto yet. Feel free to open an issue at https://github.com/spulec/moto/issues",)

    def _check_caught_exception(self, attempt_number, caught_exception):
        # This is implementation specific, but this class is useful by
        # coordinating with the MaxAttemptsDecorator.
        # The MaxAttemptsDecorator has a list of exceptions it should catch
        # and retry, but something needs to come along and actually raise the
        # caught_exception.  That's what this class is being used for.  If
        # the MaxAttemptsDecorator is not interested in retrying the exception
        # then this exception just propogates out past the retry code.
>       raise caught_exception
E       FilterNotImplementedError: The filter 'owner-alias' for DescribeImages has not been implemented in Moto yet. Feel free to open an issue at https://github.com/spulec/moto/issues

/usr/local/lib/python2.7/site-packages/botocore/retryhandler.py:359: FilterNotImplementedError
```

This is not a huge deal for me so I'm not planning to open a PR myself. But maybe someone with the same problem will. 😄

terrycain:
We'd need a comprehensive list of what alias's refer to what ownerid, if someone finds me them, i'll add support.

afgane:
Briefly looking at the docs, it appears it's only three choices: `amazon`, `aws-marketplace`, and `microsoft`: http://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_DescribeImages.html

bblommers:
This was implemented by #4175, and should be available as of moto >= 2.2.3.dev17.
I'll close this, but let us know if we've missed anything.

PR #4255, opened 2021-09-01
EC2 - Verify we can filter by owner-alias

bblommers:


bblommers:
Verifies that #1356 is working

Issue #4248, opened 2021-08-31
Implement sagemaker list_training_jobs


nivla1:
I believe there are no other issues on this endpoint; will take a look and try to implement. Creating this issue as per contributing guidelines.

bblommers:
Hi @nivla1, welcome to Moto! A PR would be very welcome - let us know if you need any help or assistance with that.

PR #4256, opened 2021-09-01
Implement sagemaker list_training_jobs enhancement (issue #4248)

nivla1:
Added mock implementation of sagemaker list_training_jobs. Tried to follow convention of str.format.

Also added prompt_toolkit to dev requirements.

Any comments appreciated

### DIFF
--- a/moto/sagemaker/models.py
+++ b/moto/sagemaker/models.py
@@ -3,11 +3,14 @@ from __future__ import unicode_literals
 import os
 from boto3 import Session
 from datetime import datetime
-
 from moto.core import ACCOUNT_ID, BaseBackend, BaseModel, CloudFormationModel
-from moto.core.exceptions import RESTError
+from moto.core.exceptions import AWSError, RESTError
 from moto.sagemaker import validators
 from .exceptions import MissingModel, ValidationError
+
+
+class AWSValidationException(AWSError):
+    TYPE = "ValidationException"
 
 
 class BaseObject(BaseModel):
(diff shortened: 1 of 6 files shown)


moto/sagemaker/models.py
```diff
@@ -1243,6 +1246,108 @@ def get_training_job_tags(self, arn):
...
+            try:
+                starting_index = int(next_token)
+                if starting_index > len(self.training_jobs):
+                    raise ValueError  # invalid next_token
+            except ValueError:
+                raise AWSValidationException('Invalid pagination token because "{0}".')
+        else:
+            starting_index = 0
+
+        if max_results:
```
bblommers, line 1277: The limiting should probably be done at the very end, after filtering the results for name, creation time, etc.

moto/sagemaker/responses.py
```diff
@@ -274,3 +274,29 @@ def delete_notebook_instance_lifecycle_config(self):
...
+        max_results = self._get_int_param("MaxResults")
+        creation_time_after = self._get_param("CreationTimeAfter")
+        creation_time_before = self._get_param("CreationTimeBefore")
+        last_modified_time_after = self._get_param("LastModifiedTimeAfter")
+        last_modified_time_before = self._get_param("LastModifiedTimeBefore")
+        name_contains = self._get_param("NameContains")
+        status_equals = self._get_param("StatusEquals")
+        sort_by = self._get_param("SortBy", "CreationTime")
+        sort_order = self._get_param("SortOrder", "Ascending")
+        response = self.sagemaker_backend.list_training_jobs(
```
bblommers, line 290: This should be done in a `try-except`, to ensure we properly format any errors that it might throw.
See `create_notebook_instance_lifecycle_config` above for an example

bblommers (changes requested):
Hi @nivla1, thanks for the PR! Please see my comments.

moto/sagemaker/models.py
nivla1: I believe limiting is the first step, unless I'm reading the [aws documentation](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/sagemaker.html#SageMaker.Client.list_training_jobs) wrong.

`When StatusEquals and MaxResults are set at the same time, the MaxResults number of training jobs are first retrieved ignoring the StatusEquals parameter and then they are filtered by the StatusEquals parameter, which is returned as a response.`

bblommers: Ah, you're right. Just did some testing, and it looks like the other params behave the same as well (CreationTimeBefore etc). That's unusual - good catch!

bblommers (approved):
LGTM! Thanks again @nivla1

Issue #2123, opened 2019-03-21
Autoscaling: describe_auto_scaling_instances() fails after attaching a moto instance with attach_instances() when using moto_server


jmeichle:
## Issue Description

When using moto_server (and possibly the decorators), calls to autoscaling describe_auto_scaling_instances() fail with a 500 after calling attach_instances() for a moto EC2 instance into a moto ASG.

## Steps to reproduce

Running moto server this way:

```
moto_server -H 0.0.0.0 -p 3000
```

And then using the following:

```
import boto3
session = boto3.Session()
asg_client = session.client(service_name='autoscaling', region_name='us-east-1', endpoint_url='http://localhost:3000/')
ec2_client = session.client(service_name='ec2', region_name='us-east-1', endpoint_url='http://localhost:3000/')

kwargs = {
    'KeyName': 'foobar',
    'ImageId': 'ami-pytest',
    'MinCount': 1,
    'MaxCount': 1,
    'InstanceType': 'c4.2xlarge',
    'TagSpecifications': [
        {
            'ResourceType': 'instance',
            'Tags': [
                {
                    'Key': 'key',
                    'Value': 'val'
                }
            ]
        },
    ]
}
fake_instance = ec2_client.run_instances(**kwargs)['Instances'][0]
fake_lc = asg_client.create_launch_configuration(
    LaunchConfigurationName='foobar',
    ImageId='ami-pytest',
    KeyName='foobar',
)
fake_asg = asg_client.create_auto_scaling_group(
    AutoScalingGroupName='test-cluster',
    LaunchConfigurationName='foobar',
    MinSize=0,
    MaxSize=1,
    AvailabilityZones=[fake_instance['Placement']['AvailabilityZone']]
)
```

This works great and I can call `asg_client.describe_auto_scaling_groups()` and `asg_client.describe_auto_scaling_instances()` without any problems.

If I attach the fake instance to the ASG via:

```
In [4]: asg_client.attach_instances(
   ...:     InstanceIds=[
   ...:         fake_instance['InstanceId'],
   ...:     ],
   ...:     AutoScalingGroupName='test-cluster'
   ...: )
Out[4]: 
{'ResponseMetadata': {'HTTPHeaders': {'content-length': '221',
   'content-type': 'text/html; charset=utf-8',
   'date': 'Thu, 21 Mar 2019 17:59:56 GMT',
   'server': 'amazon.com',
   'x-amz-crc32': '1138981672',
[...]

PR #4258, opened 2021-09-01
Fix/asg attach instances

mello7tre:
fix to #2123 (attach_instances)
    
    In attach_instances add autoscaling_group property to attached instances
    as an InstanceState property


bblommers (changes requested):
Hi @mello7tre, thanks for the PR!
Would you mind adding the testcase from the original issue here, to showcase that this works?

mello7tre:
> Hi @mello7tre, thanks for the PR!
> Would you mind adding the testcase from the original issue here, to showcase that this works?

do you mean create a test under `tests/test_autoscaling` that create the server and then a client that execute the calls ?

bblommers:
In `tests/test_autoscaling/test_autoscaling.py`, yeah - you don't have to create the server though, just take the decorated test that was posted in the [first comment](https://github.com/spulec/moto/issues/2123#issuecomment-475355108).

bblommers (approved):
Nice one - thanks again @mello7tre!

PR #4259, opened 2021-09-02
Feat/asg lifecyclehook

mello7tre:
Added AutoScalingGroup LifeCycleHook support.

put_lifecycle_hook
describe_lifecycle_hooks
delete_lifecycle_hook

Hooks are not functional and there is no triggering related to Launch or Terminate.
But it can be useful if an application need to use the describe action (this is why i added them).


bblommers (approved):
LGTM! Thanks for the PR @mello7tre

PR #4260, opened 2021-09-02
Add instance lifecycle property to detect spot instances

mello7tre:
Add instance lifecycle property to detect spot
    
    Instance lifecycle property is used to identify spot instances.
    This property is empty for non spot and has value "spot" for spot
    instances.
    When requesting a new spot instance we simply set the property to "spot"
    (SpotInstanceRequest).
    When using runInstance we use the "InstanceMarketOptions.MarketType"
    property to detect if we want to launch a spot instance.


bblommers (approved):
LGTM!

PR #4261, opened 2021-09-02
Add describe_spot_price_history

mello7tre:
Base data is taken from INSTANCE_TYPE_OFFERINGS
productDescription is fixed to: "Linux/UNIX (Amazon VPC)"
spot price is fixed to: 0.00001
timestamp is fixed.
   
Filters support is limited to "availability-zone" and "instance-type"
    
There is no price history support.


moto/ec2/models.py
```diff
@@ -5562,6 +5562,30 @@ def modify_spot_fleet_request(
         return True
 
 
+class SpotPrice(object):
```
bblommers, line 5565: This should be named `SpotPriceBackend`, to make it consistent with other backends

moto/ec2/models.py
```diff
@@ -5562,6 +5562,30 @@ def modify_spot_fleet_request(
...
+    def describe_spot_price_history(self, instance_types=[], filters=None):
```
bblommers, line 5566: Parameters should always default to None, ideally, instead of an empty list. It shouldn't impact the logic in this case

moto/ec2/models.py
```diff
@@ -5562,6 +5562,30 @@ def modify_spot_fleet_request(
...
+                elif key == "instance-type":
+                    return offering.get("InstanceType") in values
+                else:
+                    return False
+
+            return all([matches_filter(key, values) for key, values in filters.items()])
+
+        matches = [o for o in matches if matches_filters(o, filters)]
+
+        if instance_types:
```
bblommers, line 5583: Can we get a test for this as well, to verify that a call to `describe_spot_price_history(InstanceTypes=[..])` works?

bblommers (changes requested):
Thanks for the PR @mello7tre! Please see my comments

bblommers (approved):
LGTM! Thanks for making the changes @mello7tre

Issue #4249, opened 2021-08-31
Missing implementation of Cognito Username/Alias Attributes


leothomas:
First of all, a big thank you for this great tool!

I think I've stumbled upon a missing or incorrect implementation in the Cognito user-pool users. According to the [cognito docs](https://docs.aws.amazon.com/cognito/latest/developerguide/user-pool-settings-attributes.html#user-pool-settings-aliases-settings) when signing up and signing in, there are 2 configuration options (on the user pool) that enable users to use alternative identifiers (phone number, email, etc): 

1. Using the `AliasAttributes` field. In this case the user must supply a Username unique to the user pool and can then sign in with their username or any of the attributes they've specified in the `AliasAttributes` field (provided the email/phone number has been verified)
2. Using the `UsernameAttributes` field. In this case the user does not supply a username, and instead signs up directly with an email or phone number. The user pool generates a GUID for that the user (which become the value of the `sub` field in the JWT) to track it in the UserPool, and all operations are performed using the user's email/phone number as well as their `sub`. When performing a `list-users` or `get-user` operation, the `Username` field will contain the `sub` (uuid) as opposed to the email/phone number used when signing up. 

This issue is concerning the second option (`UsernameAttributes`) - since that's the implementation that we are using in our app, and the implementation we would like to be able to test.

Code sample implementing this behaviour: 
```python

from moto import mock_cognitoidp
import boto3


@mock_cognitoidp()
def moto():
    cognito_client = boto3.client("cognito-idp", region_name="us-east-1")
    return setup_user_pool(cognito_client)


def aws():
    cognito_client = boto3.client("cognito-idp", region_name="us-east-1")
    return setup_user_pool(cognito_client)


def setup_user_pool(cognito_client):
    
    # create user pool
    user_pool_id = cognito_client.create_user_pool(
        PoolName="moto-implementation-test",
        # allow users to sign up using an email address as their username
        UsernameAttributes=["email"],
        Schema=[
            {
                "Name": "sub",
                "AttributeDataType": "String",
                "DeveloperOnlyAttribute": False,
                "Mutable": False,
                "Required": True,
                "StringAttributeConstraints": {"MinLength": "1", "MaxLength": "2048"},
            },
            {
                "Name": "preferred_username",
                "AttributeDataType": "String",
                "DeveloperOnlyAttribute": False,
                "Mutable": True,
                "Required": True,
                "StringAttributeConstraints": {"MinLength": "0", "MaxLength": "2048"},
            },
        ],
[...]

Issue #3741, opened 2021-03-02
cognito-idp: cognito:* claims not available in id and access tokens


pogocode:
As far as I can tell, the tokens returned from the `admin_initiate_auth` do not include the `cognito:*` claims such as `cognito:groups`, `cognito:roles` that are documented at [Adding Groups to a User Pool](https://docs.aws.amazon.com/cognito/latest/developerguide/cognito-user-pools-user-groups.html) and [Using Tokens with User Pools](https://docs.aws.amazon.com/cognito/latest/developerguide/amazon-cognito-user-pools-using-tokens-with-identity-providers.html).

I think that this may involve a change to the following code - I see that there is already facility to include extra data from readable attributes, but this probably doesn't cover this scenario:

https://github.com/spulec/moto/blob/5c04a4d8cc0631da147ee298dbe29e94370b7217/moto/cognitoidp/models.py#L121-L141

I'd be interested in attempting to implement this feature if it is indeed missing (and I've not just misunderstood the code).

bblommers:
Hi @pogocode, thanks for raising this!
If I read it correctly, the 'readable attributes' are supplied by the user in the ReadAttributes-parameter when calling the `create_user_pool_client`-method.
Considering the `cognito:` claims will always be part of the ID token, it probably makes most sense to extend the `create_id_token` method with this information:
https://github.com/spulec/moto/blob/751c5ec362730f143adad902f7fcdb9a67fd2993/moto/cognitoidp/models.py#L143

PR's are always welcome! Let me know if you need any help.

elisoncrum:
Inquiring about progress that @leothomas has possibly made on the contribution on adding the ability for `cognito:groups` to be displayed in the JWT since it is similar to his PR #4262.

bblommers:
The `cognito:groups`-attribute is now part of the access token as of moto >= 2.2.16.dev16.

The AWS docs do not reference `cognito:roles` as part of the token, so that has not been added. Do let us know if that's an oversight, happy to reopen this if some attributes are still missing.

PR #4262, opened 2021-09-02
Userpool UsernameAttributes

leothomas:
Addresses #4249 

## What I did: 
Implemented the logic necessary to enable the  `UsernameAttributes` parameter in user pools. (Ref: [Using Aliases to Simplify User Sign-Up and Sign-In,  Option2 ](https://docs.aws.amazon.com/cognito/latest/developerguide/user-pool-settings-attributes.html#user-pool-settings-aliases-settings). 

**NOTE:** This PR only implements `UsernameAttributes` and not `UsernameAliases` (Option 1 of the above link), although with the work done for this implementation, I suspect it would be quite easy to implement `UsernameAliases`.

## How I did it: 
The `admin_create_user` and `sign_up` methods check if the `UsernameAttributes` parameter is set within the user pool's `extended_config`. 
- If no, proceed as before - create the user with the provided username and use that username as the key under which to store that user in the userpool's users object.
- If yes, validate the provided username against the regex of each of the possible `UsernameAttributes` (`email` or `phone_number` or both). 
   - If there is no match for any `UsernameAttributes` throw an `InvalidParameterException`. 
   - If there is a match, added the provided username to the user's `UserAttributes` under it's attribute type (`email` or `phone_number`) and set the User's username to the value of the internally generate id (called Cognito Sub or just Sub). 

I also added the user's Sub as a item in the user's `UserAttributes`, as it was included in the responses of `admin-get-user` and `list-user` operations that I performed on cognito user pools with `UsernameAttributes` both enabled and disabled.

I implemented a `_get_user(username:str)-> CognitoIdpUser` method that will retrieve a user from a user pool by Username if `UsernameAttributes` is **not** set, and by any of the `UsernameAttributes` (`email` or `phone_number`) or by the user's sub, if the value **is** set. Eg: 
```bash
aws admin-get-user --user-pool-id ${userpool_WITHOUT_username_attributes} --username "a-unique-username" --> success

aws admin-get-user --user-pool-id ${userpool_WITH_username_attributes} --username "email@domain.com"  --> success
aws admin-get-user --user-pool-id ${userpool_WITH_username_attributes} --username "+123456789"  --> success
aws admin-get-user --user-pool-id ${userpool_WITH_username_attributes} --username "266ca5bd-eda1-48c7-9887-f3c685eb2fa9"  --> success
```

I updated any methods that access users by provided username to use this `_get_user()` method. 

I added some unit tests to cover these methods - but they could probably be made a bit more comprehensive - I'm open to adding more if needed.

### DIFF
--- a/moto/cognitoidp/models.py
+++ b/moto/cognitoidp/models.py
@@ -22,7 +22,13 @@ from .exceptions import (
     UserNotConfirmedException,
     InvalidParameterException,
 )
-from .utils import create_id, check_secret_hash
+from .utils import (
+    create_id,
+    check_secret_hash,
+    validate_username_format,
+    flatten_attrs,
+    expand_attrs,
+)
 
 UserStatus = {
     "FORCE_CHANGE_PASSWORD": "FORCE_CHANGE_PASSWORD",
@@ -121,6 +127,20 @@ class CognitoIdpUserPool(BaseModel):
 
         return user_pool_json
 
+    def _get_user(self, username):
+        """Find a user within a user pool by Username or any UsernameAttributes
+        (`email` or `phone_number` or both)"""
+        if self.extended_config.get("UsernameAttributes"):
+            attribute_types = self.extended_config["UsernameAttributes"]
+            for user in self.users.values():
+                if username in [
+                    flatten_attrs(user.attributes).get(attribute_type)
+                    for attribute_type in attribute_types
+                ]:
+                    return user
+
+        return self.users.get(username)
+
     def create_jwt(
         self, client_id, username, token_use, expires_in=60 * 60, extra_data={}
     ):
@@ -129,12 +149,12 @@ class CognitoIdpUserPool(BaseModel):
             "iss": "https://cognito-idp.{}.amazonaws.com/{}".format(
                 self.region, self.id
             ),
-            "sub": self.users[username].id,
+            "sub": self._get_user(username).id,
             "aud": client_id,
             "token_use": token_use,
             "auth_time": now,
             "exp": now + expires_in,
-            "email": self.users[username].username,
+            "email": flatten_attrs(self._get_user(username).attributes).get("email"),
         }
         payload.update(extra_data)
         headers = {"kid": "dummy"}  # KID as present in jwks-public.json
@@ -179,7 +199,7 @@ class CognitoIdpUserPool(BaseModel):
                 attribute = list(
                     filter(
                         lambda f: f["Name"] == readable_field,
-                        self.users.get(username).attributes,
+                        self._get_user(username).attributes,
                     )
                 )
                 if len(attribute) > 0:
@@ -300,7 +320,9 @@ class CognitoIdpUser(BaseModel):
     def __init__(self, user_pool_id, username, password, status, attributes):
         self.id = str(uuid.uuid4())
         self.user_pool_id = user_pool_id
-        self.username = username
+        # Username is None when users sign up with an email or phone_number,
+        # and should be given the value of the internal id generate (sub)
+        self.username = username if username else self.id
         self.password = password
         self.status = status
         self.enabled = True
@@ -314,6 +336,8 @@ class CognitoIdpUser(BaseModel):
         # Groups this user is a member of.
         # Note that these links are bidirectional.
         self.groups = set()
+
+        self.update_attributes([{"Name": "sub", "Value": self.id}])
 
     def _base_json(self):
         return {
@@ -345,12 +369,6 @@ class CognitoIdpUser(BaseModel):
         return user_json
 
     def update_attributes(self, new_attributes):
-        def flatten_attrs(attrs):
-            return {attr["Name"]: attr["Value"] for attr in attrs}
-
-        def expand_attrs(attrs):
-            return [{"Name": k, "Value": v} for k, v in attrs.items()]
-
         flat_attributes = flatten_attrs(self.attributes)
         flat_attributes.update(flatten_attrs(new_attributes))
         self.attributes = expand_attrs(flat_attributes)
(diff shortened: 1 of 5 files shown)


moto/cognitoidp/models.py
```diff
@@ -644,11 +662,52 @@ def admin_create_user(
...
 
         if message_action and message_action == "RESEND":
-            if username not in user_pool.users:
+            if not user_pool._get_user(username):
                 raise UserNotFoundError(username)
-        elif username in user_pool.users:
+        elif user_pool._get_user(username):
             raise UsernameExistsException(username)
 
+        # UsernameAttributes are attributes (either `email` or `phone_number`
```
bblommers, line 670: Good documentation :+1:

moto/cognitoidp/utils.py
```diff
@@ -4,6 +4,12 @@
 import hashlib
 import hmac
 import base64
+import re
+
+FORMATS = {
```
bblommers, line 9: Are these regexes defined by AWS somewhere? I'm a little worried that AWS will have something slightly different, and we'll make it too restrictive

tests/test_cognitoidp/test_cognitoidp.py
```diff
@@ -1145,7 +1295,67 @@ def test_admin_get_user():
...
+    result["UserAttributes"].should.have.length_of(4)
+    _verify_attribute("phone_number", "+123456789")
+    _verify_attribute("email", "test@example.com")
+
+    # verify that the generate user sub is a valid UUID v4
+    [user_sub] = [
+        attr["Value"] for attr in result["UserAttributes"] if attr["Name"] == "sub"
+    ]
+    user_sub_valid = False
+    try:
```
bblommers, line 1344: This shouldn't need a try-catch - we can just call `uuid.UUID(user_sub)` as is, and let it fail if it's invalid

tests/test_cognitoidp/test_cognitoidp.py
```diff
@@ -1164,13 +1374,31 @@ def test_admin_get_missing_user():
...
+@mock_cognitoidp
+def test_admin_get_missing_user_with_username_attributes():
+    conn = boto3.client("cognito-idp", "us-west-2")
+
+    username = "test@example.com"
+    user_pool_id = conn.create_user_pool(
+        PoolName=str(uuid.uuid4()), UsernameAttributes=["email"]
+    )["UserPool"]["Id"]
+
+    caught = False
```
bblommers, line 1386: We should use the `with pytest.raises()` here, instead of manually verifying whether it has failed.
That way we can also verify whether the error message/code is correct

tests/test_cognitoidp/test_cognitoidp.py
```diff
@@ -1222,6 +1450,58 @@ def test_list_users():
...
+        Username=username_bis,
+        UserAttributes=[{"Name": "phone_number", "Value": "+33666666666"}],
+    )
+    result = conn.list_users(
+        UserPoolId=user_pool_id, Filter='phone_number="+33666666666'
+    )
+    result["Users"].should.have.length_of(1)
+    result["Users"][0]["Username"].should_not.equal(username_bis)
+
+    user_sub_valid = False
```
bblommers, line 1485: Same here

tests/test_cognitoidp/test_cognitoidp.py
```diff
@@ -1382,6 +1699,26 @@ def test_admin_delete_user():
...
+    conn = boto3.client("cognito-idp", "us-west-2")
+
+    username = "test@example.com"
+    user_pool_id = conn.create_user_pool(
+        PoolName=str(uuid.uuid4()), UsernameAttributes=["email"]
+    )["UserPool"]["Id"]
+    conn.admin_create_user(UserPoolId=user_pool_id, Username=username)
+    conn.admin_delete_user(UserPoolId=user_pool_id, Username=username)
+
+    caught = False
```
bblommers, line 1713: Same here

bblommers (changes requested):
Hi @leothomas, thanks for the PR! Please see my comments

moto/cognitoidp/utils.py
leothomas: I've only been able to find documentation regarding the phone number: 

> Phone numbers must follow these formatting rules: A phone number must start with a plus (+) sign, followed immediately by the country code. A phone number can only contain the + sign and digits. You must remove any other characters from a phone number, such as parentheses, spaces, or dashes (-) before submitting the value to the service. For example, a United States-based phone number must follow this format: +14325551212.

[Source - under `phone_number`](https://docs.aws.amazon.com/cognito/latest/developerguide/user-pool-settings-attributes.html#user-pool-settings-usernames)

Regarding the email address, I pulled the regex from [here](https://www.geeksforgeeks.org/check-if-email-address-valid-or-not-in-python/)

leothomas:
@bblommers comments have been addressed 😄

moto/cognitoidp/utils.py
bblommers: Fair enough, that should be OK

bblommers (changes requested):
Thanks for making the changes @leothomas - happy to merge once the CI passes.

(Keep in mind that we use `black==19.10b0` as our linter - different versions of black might produce different results.)

tests/test_cognitoidp/test_cognitoidp.py
```diff
@@ -1164,13 +1367,29 @@ def test_admin_get_missing_user():
...
+@mock_cognitoidp
+def test_admin_get_missing_user_with_username_attributes():
+    conn = boto3.client("cognito-idp", "us-west-2")
+
+    username = "test@example.com"
+    user_pool_id = conn.create_user_pool(
+        PoolName=str(uuid.uuid4()), UsernameAttributes=["email"]
+    )["UserPool"]["Id"]
+
+    with pytest.Raises(ClientError) as ex:
```
bblommers, line 1379: Typo - that should be a lower case `raises`

tests/test_cognitoidp/test_cognitoidp.py
```diff
@@ -1382,6 +1682,24 @@ def test_admin_delete_user():
...
+    conn = boto3.client("cognito-idp", "us-west-2")
+
+    username = "test@example.com"
+    user_pool_id = conn.create_user_pool(
+        PoolName=str(uuid.uuid4()), UsernameAttributes=["email"]
+    )["UserPool"]["Id"]
+    conn.admin_create_user(UserPoolId=user_pool_id, Username=username)
+    conn.admin_delete_user(UserPoolId=user_pool_id, Username=username)
+
+    with pytest.Raises(ClientError) as ex:
```
bblommers, line 1696: Same

bblommers (changes requested):

leothomas:
@bblommers apologies for the delay - I've had to shift my priorities a bit. I'll try to loop back around to this as soon as I can. 

In our use case, we also need to make use of the `cognito:groups` attribute, a feature that has already been requested here: #3741 . I'm happy to add that feature to this PR  - or I can open a separate one. Let me know!

bblommers:
No worries @leothomas! Ideally the `cognito:groups` would be a separate PR to make the Git history look nicer, but if one PR makes development a lot easier, I'd be OK with that as well.

bblommers:
All good @leothomas - I know how it goes.

`cognito:groups` is not yet supported, so another PR would be very welcome :+1:

PR #4264, opened 2021-09-05
EC2:create_volume() - Add support for VolumeType-parameter

### LINKED ISSUE #3960: Support non-standard EC2 volumes
## Steps to reproduce

moto: 2.0.7
boto3: 1.17.75

```python
from moto import mock_ec2
import boto3
import unittest

    @mock_ec2
    def test_create_non_standard_volume(self):
        ec2 = boto3.client('ec2', region_name='us-east-1')
        self.assertEqual(
            ec2.create_volume(AvailabilityZone='us-east-1a', Size=100, Iops=1000, VolumeType='io2')['VolumeType'],
            'io2'
        )
```

## Expected behavior
The test passes.

## Actual behavior
The test fails.

bblommers:
Fixes #3960

### DIFF
--- a/moto/ec2/models.py
+++ b/moto/ec2/models.py
@@ -620,7 +620,11 @@ class Instance(TaggedEC2Resource, BotoInstance, CloudFormationModel):
         kms_key_id=None,
     ):
         volume = self.ec2_backend.create_volume(
-            size, self.region_name, snapshot_id, encrypted, kms_key_id
+            size=size,
+            zone_name=self.region_name,
+            snapshot_id=snapshot_id,
+            encrypted=encrypted,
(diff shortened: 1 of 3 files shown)


jeking3 (commented):
This would be improved by checking the VolumeType to see if it is valid, for example adding a test to set the VolumeType=`spam1` and getting the appropriate response back.

https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/ec2.html#EC2.Client.create_volume

Valid types: `VolumeType='standard'|'io1'|'io2'|'gp2'|'sc1'|'st1'|'gp3'`

Not certain what the error response should be, as it is not documented. :|

PR #4267, opened 2021-09-06
Fix SecuirtyGroupRules and added default Ec2-VPC support for SG.

monty16597:
# Fix
- `SecurityGroupRules`
-  Default security groups are now being made by `Ec2Backend`'s `__init__` function.
- `SecurityGroup`'s id is now 17 char long
- By default `create_ssecurity_group` will take default VPC if none is being passed by the user (EC2-VPC only support). 
- test cases
- If it is `SecurityGroup` with `VPC_ID` then `source_group` can only be passed using the `GroupId` parameter. Still`GroupName` is being supported.

# Added
- Support for API `UpdateSecurityGroupRuleDescriptionsIngress`
- Support for API `UpdateSecurityGroupRuleDescriptionsEgress`

# Terraform passed tests
- `TestAccAWSSecurityGroupRule`

### DIFF
--- a/moto/ec2/models.py
+++ b/moto/ec2/models.py
@@ -963,7 +963,7 @@ class InstanceBackend(object):
         new_reservation.id = random_reservation_id()
 
         security_groups = [
-            self.get_security_group_from_name(name) for name in security_group_names
+            self.get_security_group_by_name_or_id(name) for name in security_group_names
         ]
         security_groups.extend(
             self.get_security_group_from_id(sg_id)
@@ -2198,7 +2198,9 @@ class SecurityGroup(TaggedEC2Resource, CloudFormationModel):
     @classmethod
     def _delete_security_group_given_vpc_id(cls, resource_name, vpc_id, region_name):
         ec2_backend = ec2_backends[region_name]
-        security_group = ec2_backend.get_security_group_from_name(resource_name, vpc_id)
+        security_group = ec2_backend.get_security_group_by_name_or_id(
+            resource_name, vpc_id
+        )
         if security_group:
             security_group.delete(region_name)
 
@@ -2293,12 +2295,13 @@ class SecurityGroupBackend(object):
     def create_security_group(
         self, name, description, vpc_id=None, tags=None, force=False
     ):
+
         if not description:
             raise MissingParameterError("GroupDescription")
 
         group_id = random_security_group_id()
         if not force:
-            existing_group = self.get_security_group_from_name(name, vpc_id)
+            existing_group = self.get_security_group_by_name_or_id(name, vpc_id)
             if existing_group:
                 raise InvalidSecurityGroupDuplicateError(name)
         group = SecurityGroup(
@@ -2342,7 +2345,7 @@ class SecurityGroupBackend(object):
         elif name:
             # Group Name.  Has to be in standard EC2, VPC needs to be
             # identified by group_id
-            group = self.get_security_group_from_name(name)
+            group = self.get_security_group_by_name_or_id(name)
             if group:
                 return self._delete_security_group(None, group.id)
             raise InvalidSecurityGroupNotFoundError(name)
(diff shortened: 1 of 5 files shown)


moto/ec2/models.py
```diff
@@ -7761,6 +7880,11 @@ def __init__(self, region_name):
             # backward-compatibility issues
             vpc = self.vpcs.values()[0]
 
+        self.default_vpc = vpc
+
+        # Create the default security group
+        self.create_security_group("default", "default group", vpc.id)
```
bblommers, line 7886: This makes a lot more sense - nice to no longer have the duplicate default SG 👍

tests/test_ec2/test_security_groups.py
```diff
@@ -581,8 +583,8 @@ def test_sec_group_rule_limit():
     ec2_conn.authorize_security_group_egress(
         group_id=sg.id, ip_protocol="-1", src_group_id=other_sg.id
     )
-    # fill the rules up the limit
-    for i in range(1, 100):
+    # fill the rules up the limit, 49 + 1 default rule
+    for i in range(1, 49):
```
bblommers, line 587: What's the reasoning behind this change?

monty16597: Actually, As we discussed on other PR, we are no longer supporting Ec2-Classic for the SG. So by default, there will be always `SecurityGroup` in default `VPC` if no VPC is being passed while creating a SG. So for `EC2-VPC`'s SGs, one can have max of 50 Rules per SG. For egress, there will be by default 1 rule.

However, AWS is still giving backward compatibility and I haven't checked the other services in `moto` itself if they do have those EC2-Classic attributes. So should we make it default behaviour or add `TODO` for the future? thoughts?

monty16597: I think we should add TODO note because there are still many `EC2-Classic` test cases for different services which exist in terraform tests.

bblommers: Yeah, I meant the change from 100 to 50. Is it the EC2-classic rule to allow 100 SG's, or was it always max 50 (and Moto had been wrong the entire time)?

bblommers: > I think we should add TODO note because there are still many `EC2-Classic` test cases for different services which exist in terraform tests.

I'm having a difficult time to wrap my head around how Terraform would do this. Isn't EC2-classic only available/enabled on older accounts? Or is there a way to force EC2-classic behaviour by passing a specific parameter somehow

monty16597: > Yeah, I meant the change from 100 to 50. Is it the EC2-classic rule to allow 100 SG's, or was it always max 50 (and Moto had been wrong the entire time)?

- For EC2-Classic its 100 and for Ec2-VPC its 60 (We have to update it from 50 to 60).

> I'm having a difficult time to wrap my head around how Terraform would do this. Isn't EC2-classic only available/enabled on older accounts? Or is there a way to force EC2-classic behaviour by passing a specific parameter somehow

- IDK as well, they maybe have different accounts for testing purposes with EC2-Classic features or they can ask to AWS. Its Hashicorp, right? AWS can provide them 😉 . As per the [link](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-classic-platform.html), they no longer allow Ec2-Classic for the new ACC. So I guess, there is no way we can enable Ec2-classic features for our new ACCs.

bblommers: My preference would be to switch to VPC as the default behaviour, and retire any EC2-classic features as and when we encounter them.
The TF tests might fail for another year, until it's retired for good, but I expect the vast majority of people to use VPC only. So forcefully retiring our EC2-classic behaviour will improve/increase the supported features for the vast majority of Moto/Localstack users.
Any users who are still on EC2-classic will just have to pin Moto to an older version.

monty16597: Sounds good. We can blacklist those tests as we encounter them.

bblommers: Cool. Looks like we have a conflict to resolve before we can merge, but LGTM otherwise - let me know when this is ready

bblommers (approved):
LGTM! Thanks @monty16597

PR #4274, opened 2021-09-09
Performance improvement - Only reset backends loaded into memory

bblommers:
Follow up to #4209 
Only backends that are loaded into memory will now be reset when calling `/moto-api/reset`, reducing the waiting time drastically.

@whummer @thrau Does LocalStack use the reset-api, and/or you do expect there to be any impact for you?

### DIFF
--- a/moto/backends.py
+++ b/moto/backends.py
@@ -2,6 +2,7 @@ from __future__ import unicode_literals
 
 import importlib
 import moto
+import sys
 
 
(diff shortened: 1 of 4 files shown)


whummer:
Thanks for the heads-up @bblommers - we're not currently using the `reset()` API in LocalStack, so we should be good. 👍 Thanks!

PR #4277, opened 2021-09-10
fixed vpc cidr association integrations with route-tables 

macnev2013:
## Fixed/Added
- attaching CIDR block with VPC now creates a route in all route tables in VPC
- if multiple CIDR blocks are attached to vpc - it creates default `local` routes on creating new route table in the same VPC

## Terraform Test Case
- TestAccAWSRouteTable_VpcMultipleCidrs
- TestAccAWSRouteTable_ConditionalCidrBlock


bblommers (approved):
LGTM! Thanks @macnev2013

PR #2839, opened 2020-03-24
add metric filter support to cloudwatch logs

brady-gsa:

PR #4278, opened 2021-09-10
add metric filter support to cloudwatch logs (updated)

foxmulder900:
Full credit here goes to @brady-gsa, I have just resolved the conflicts with master as well as some linting issues from his original pull-request here:
https://github.com/spulec/moto/pull/2839

If there are any remaining issues, I will happily attempt to work through them in order to get this feature in.

### DIFF
--- a/IMPLEMENTATION_COVERAGE.md
+++ b/IMPLEMENTATION_COVERAGE.md
@@ -7068,7 +7068,7 @@
 
 ## logs
 <details>
-<summary>50% implemented</summary>
+<summary>57% implemented</summary>
 
 - [ ] associate_kms_key
 - [ ] cancel_export_task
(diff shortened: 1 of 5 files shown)


bblommers (changes requested):
Hi @foxmulder900, thanks for resubmitting this and cleaning up!

After the build passes I'll do a full check, but just a general note: there shouldn't be any `assert` statements in the logic. (Something I should've picked up in the original PR as well)

Those will fail alright, but there will not be a very descriptive error message - especially in ServerMode, where it returns a generic 500.

An example of how to transform these into boto3-compliant exceptions can be found here: https://github.com/spulec/moto/pull/3744/files

foxmulder900:
@bblommers Thanks for pointing me in the right direction! I've updated the validation to raise proper exceptions instead of using `assert`. I also resolved an issue I found with how the describe_metric_filter function was using the `metricName` and `metricNamespace` parameters.

bblommers (approved):
LGTM! Thanks for making the changes @foxmulder900

PR #4282, opened 2021-09-12
support for carrier gateway in rt

macnev2013:
## Added
- Support for carrier gateway routes

## Terraform test case
- TestAccAWSRouteTable_IPv4_To_CarrierGateway

Note: I have removed duplicate entry `TestAccAWSRouteTable_RequireRouteTarget` from passing test case list


bblommers (approved):
Nice! Thanks @macnev2013

PR #4285, opened 2021-09-14
fixed route-table vpc-endpoint integration

macnev2013:
# Fixed
- VPC Endpoint integration with route table
- Prefix List as route table destination

# TF Test Passed
- TestAccAWSRouteTable_IPv4_To_InternetGateway
- TestAccAWSRouteTable_GatewayVpcEndpoint


macnev2013:
Hi @bblommers , Can you please review this PR a in a bit detail and see if it will any side-effects on other functionalities? (then we can fix the tests accordingly) 

Thanks

bblommers:
> 
> 
> Hi @bblommers , Can you please review this PR a in a bit detail and see if it will any side-effects on other functionalities? (then we can fix the tests accordingly)
> 
> Thanks

Hi @macnev2013, I'm not aware of anything that would be impacted by this 👍

macnev2013:
Hi @bblommers,
Short Note: I just discovered while using `boto3.resource` to get the routes in route table using `route_table.routes` does not contain any route with IPv6 in the result.
Alternatively we can use `route_table.routes_attribute` to retrive all the routes.
Or we can use `boto3.client` for the same

bblommers (approved):
LGTM! Thanks @macnev2013

PR #4286, opened 2021-09-15
Allow CloudWatch logs to be sent to Firehose

kbalk:
The CloudWatch APIs `put_subscription_filter` and `put_log_events` will now support a Firehose destination.

Data is stored in the S3 bucket specified as a Firehose destination.  That data is written in Base64 encoded gzip format and can be retrieved from that bucket.  

However, I'm not sure I did everything that needs to be done.  Perhaps additional logic will be required at a future date.


bblommers (approved):
LGTM! Thanks @kbalk

PR #4287, opened 2021-09-15
Change message deletion behavior of SQS queue to use discard

silv-io:
Using `remove` in this case causes a `KeyError` when deleting messages with changed visibility (because those messages are not present in `pending_messages`). Changing it to `discard` avoids the error.

Also, the comment suggests the message is only removed from `pending_messages` but as messages which have the correct handle are not appended to the new list they will be removed regardless. The behavior is correct though (tested against AWS).

Relates to [issue in localstack](https://github.com/localstack/localstack/issues/4470)

### DIFF
--- a/moto/sqs/models.py
+++ b/moto/sqs/models.py
@@ -886,12 +886,11 @@ class SQSBackend(BaseBackend):
         ):
             raise ReceiptHandleIsInvalid()
 
+        # Delete message from queue regardless of pending state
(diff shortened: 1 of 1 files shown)


whummer:
@bblommers Can you please help us approve the workflow trigger. Thanks!
image

bblommers (approved):
LGTM! Thanks for the fix @silv-io

PR #4292, opened 2021-09-15
Fix error when AWS::ECS::Service does not include a DesiredCount

psxpaul:
I recently added Moto to a project, and got an error related to the DesiredCount on my Ecs Servie declaration in the CloudFormation template. According to [the AWS documentation](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-ecs-service.html#cfn-ecs-service-desiredcount), DesiredCount is not required in some circumstances.


bblommers (approved):
LGTM! Thanks for the PR @psxpaul, much appreciated.

PR #4294, opened 2021-09-15
Firehose timestamps should be in ISO format

kbalk:
Corrected create and update timestamp formats.


bblommers (approved):
LGTM! Thanks for the follow-up @kbalk

PR #4295, opened 2021-09-16
Publish messages to SNS when CloudFormation NotifcationARNs is set

psxpaul:
When creating/updating a CloudFormation stack, you can supply NotificationARNs in order to receive stack related events. This change will make moto publish those events to the given SNS topics.


bblommers (approved):
Nice one - thanks @psxpaul!

PR #4300, opened 2021-09-16
Implement ACM export_certificate function

psxpaul:
This is an implementation of the [ACM ExportCertificate action](https://docs.aws.amazon.com/acm/latest/APIReference/API_ExportCertificate.html). I changed the google root ca to the aws root ca [found here](https://censys.io/authorities/7a3a96e64a895c66676cc71400ed0e3ce9718be9ec5d46edb4f4a52d96a4414d) because it looks like that is the cert that matches the issuer in the `generate_cert()` method.

### DIFF
--- a/tests/test_acm/test_acm.py
+++ b/tests/test_acm/test_acm.py
@@ -171,12 +171,10 @@ def test_describe_certificate():
 def test_describe_certificate_with_bad_arn():
     client = boto3.client("acm", region_name="eu-central-1")
 
-    try:
+    with pytest.raises(ClientError) as err:
         client.describe_certificate(CertificateArn=BAD_ARN)
-    except ClientError as err:
-        err.response["Error"]["Code"].should.equal("ResourceNotFoundException")
-    else:
-        raise RuntimeError("Should of raised ResourceNotFoundException")
+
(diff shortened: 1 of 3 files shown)


tests/test_acm/test_acm.py
```diff
@@ -176,6 +179,39 @@ def test_describe_certificate_with_bad_arn():
...
+        format=serialization.PrivateFormat.TraditionalOpenSSL,
+        encryption_algorithm=serialization.NoEncryption(),
+    )
+    private_key.should.equal(SERVER_KEY)
+
+
+@mock_acm
+def test_export_certificate_with_bad_arn():
+    client = boto3.client("acm", region_name="eu-central-1")
+    try:
```
bblommers, line 207: Would you mind rewriting this to use pytest.raises? That way we don't have to manually raise the RuntimeError
```
    client.export_certificate(CertificateArn=BAD_ARN, Passphrase="pass")
err.value.response["Error"]["Code"].should.equal("ResourceNotFoundException")
```

bblommers (changes requested):

bblommers (approved):
LGTM! Thanks for the PR @psxpaul

PR #4302, opened 2021-09-17
Fix CloudFormation create_stack returning stack_id with wrong region 

psxpaul:
Calling CloudFormation's CreateStack action seems to always return a StackId with 'us-east-1' in it.


bblommers (approved):
Nice catch @psxpaul - thanks for the PR!

PR #4303, opened 2021-09-18
added update rest api

rajendersaini:
Added implementation add/remove  update-rest-api , I would like to add @jainmnsh he also worked with me on this.


bblommers (approved):
LGTM - thanks for the PR @rajendersaini and @jainmnsh !

PR #4309, opened 2021-09-19
implemented validators endpoints

rajendersaini:
@bblommers  - I have one test case failing - need your help to sort out get_request_validators()


tests/test_apigateway/test_apigateway.py
```diff
@@ -959,6 +959,133 @@ def test_delete_authorizer():
...
+            VALIDATE_REQUEST_BODY: True,
+            VALIDATE_REQUEST_PARAMETERS: True,
+        }
+    )
+
+    # delete validator
+    response = client.delete_request_validator(
+        restApiId=api_id, requestValidatorId=validator_id
+    )
+
```
bblommers, line 1028: We should have a more direct verification that this validator is now gone.
That can be as simple as verifying that this validator is no longer returned when requesting the full list:
```
validators = client.get_request_validators(restApiId=api_id)["items"]
validators.should.have.length_of(0)
```

moto/apigateway/responses.py
```diff
@@ -303,6 +303,54 @@ def restapis_authorizers(self, request, full_url, headers):
...
+                validator = restApi.create_request_validator(
+                    name=name,
+                    validateRequestBody=validateRequestBody,
+                    validateRequestParameters=validateRequestParameters,
+                )
+                return 200, {}, json.dumps(validator)
+        except BadRequestException as e:
+            return self.error("BadRequestException", e.message)
+        except NotFoundException as e:
+            return self.error("NotFoundException", e.message)
```
bblommers, line 334: This should have a (stand alone) test case as well.
Negative tests like this should follow the following format:
https://github.com/spulec/moto/blob/master/tests/test_apigateway/test_apigateway.py#L78

moto/apigateway/responses.py
```diff
@@ -303,6 +303,54 @@ def restapis_authorizers(self, request, full_url, headers):
...
+                    "validateRequestParameters"
+                )
+                validator = restApi.create_request_validator(
+                    name=name,
+                    validateRequestBody=validateRequestBody,
+                    validateRequestParameters=validateRequestParameters,
+                )
+                return 200, {}, json.dumps(validator)
+        except BadRequestException as e:
+            return self.error("BadRequestException", e.message)
```
bblommers, line 332: In what scenario would this exception be raised? Ideally we have a unit test for this

bblommers (changes requested):
Hi @rajendersaini, thanks for the PR! Please see my comments.

As a general note: Would you mind moving all the new tests to a new file, `test_apigateway_validators.py`? The existing file is quite large already, so it makes sense to group all these tests in a separate file.

moto/apigateway/responses.py
rajendersaini: I think I just added this looking at other code, let me remove it for the time being.

moto/apigateway/responses.py
rajendersaini: removed.

moto/apigateway/responses.py
bblommers: No, you were right adding it, actually, it should be there :) The `self.error()` method parses our custom exceptions into a JSON format that is understood by all AWS SDK's (Python/Java/.NET/etc)

tests/test_apigateway/test_apigateway_validators.py
```diff
@@ -0,0 +1,168 @@
...
+            VALIDATE_REQUEST_BODY: True,
+            VALIDATE_REQUEST_PARAMETERS: True,
+        }
+    )
+
+    # delete validator
+    response = client.delete_request_validator(
+        restApiId=api_id, requestValidatorId=validator_id
+    )
+    with pytest.raises(RequestValidatorNotFound):
```
bblommers, line 119: Sorry, I should have been more explicit in the first place. 
Negative tests should be in the following format:
```
    client.get_request_validator(
        restApiId=api_id, requestValidatorId=validator_id
    )
err["Code"].should.equal("NotFoundException")
err["Message"].should.equal("Invalid Request Validator Id specified")
```
This verifies that the exception is formatted correctly. See my other comment on adding the `self.error()` back in.

Changing this will fix the ServerMode test that is currently failing.

bblommers (changes requested):

rajendersaini:
@bblommers  - please approve and merge.

bblommers (approved):
LGTM! Thanks for the PR @rajendersaini

PR #4311, opened 2021-09-20
Fix response rendering for IAM endpoints using GENERIC_EMPTY_TEMPLATE

dominikschubert:
The GENERIC_EMPTY_TEMPLATE jinja template already adds a `...Response` suffix to the XML tag name so some of these were incorrectly leading to a double suffixes such as `<DeleteRolePolicyResponseResponse>`.

Also fixes the `delete_group_policy` endpoint by using the same GENERIC_EMPTY_TEMPLATE to render the response.

### DIFF
--- a/moto/iam/responses.py
+++ b/moto/iam/responses.py
@@ -18,7 +18,7 @@ class IamResponse(BaseResponse):
         policy_arn = self._get_param("PolicyArn")
         iam_backend.detach_role_policy(policy_arn, role_name)
         template = self.response_template(GENERIC_EMPTY_TEMPLATE)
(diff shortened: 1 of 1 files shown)


dominikschubert:
The jinja template for reference: 

```python
GENERIC_EMPTY_TEMPLATE = """<{{ name }}Response>
   <ResponseMetadata>
      <RequestId>7a62c49f-347e-4fc4-9331-6e8eEXAMPLE</RequestId>
   </ResponseMetadata>
</{{ name }}Response>"""
```

bblommers (approved):
Nice catch @dominikschubert - thanks for the PR!

PR #4314, opened 2021-09-20
Fix: Adding ClientRequestToken for SecretsManager update_secret method

mgagliardo:
Fixes https://github.com/localstack/localstack/issues/4618

* Adds ClientRequestToken to the `update_secret` method.
* Improving validation of the param on most of methods using it

### DIFF
--- a/moto/secretsmanager/models.py
+++ b/moto/secretsmanager/models.py
@@ -19,7 +19,7 @@ from .exceptions import (
     InvalidRequestException,
     ClientError,
 )
-from .utils import random_password, secret_arn, get_secret_name_from_arn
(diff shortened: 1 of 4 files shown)


bblommers (changes requested):
Hi @mgagliardo, thanks for the PR! Would you mind adding a few tests for this functionality? 

The `test_secretsmanager/test_server.py` already has some tests to check the token length, those can be extended/duplicated to verify the same behaviour for the UpdateSecret-action.
Verifying that a token is persisted/returned can be done with a regular boto3-test in `test_secretsmanager.py`.

bblommers (approved):
LGTM!

Issue #4284, opened 2021-09-14
Batch job definitions do not capture tags


oakbramble:
I am working with a lambda function which creates a job definition and then cleans up old definitions which have the same tags as the job definition it has just created.

When using moto 2.2.6 to test this, I have run into the issue that the moto job definition does not seem to capture the tags of the job definition. This is the structure of the job definition I get with moto, which does not include tags:

```
"jobDefinitions": [
        {
            "jobDefinitionName": "test_repo-latest",
            "jobDefinitionArn": "arn:aws:batch:eu-west-1:123456789012:job-definition/test_repo-latest:1",
            "revision": 1,
            "status": "ACTIVE",
            "type": "container",
            "parameters": {},
            "retryStrategy": {"attempts": 1},
            "containerProperties": {
                "image": "123456789012.dkr.ecr.eu-west-1.amazonaws.com/test_repo:latest",
                "vcpus": 1,
                "memory": 128,
                "command": ["ls"],
                "jobRoleArn": "arn:aws:iam::123456789012:role/test_customer-test_cloud-test_environment-user-iam-role",
                "volumes": [
                    {
                        "name": "efs",
                        "efsVolumeConfiguration": {
                            "fileSystemId": "fs-de240457",
                            "rootDirectory": "/",
                        },
                    }
                ],
                "environment": [],
                "mountPoints": [{"containerPath": "/efs", "sourceVolume": "efs"}],
                "privileged": False,
                "ulimits": [],
                "resourceRequirements": [],
                "secrets": [],
            },
        }
    ],
```
As opposed to the structure I get when I `describe_job_definitions` with boto3 1.18.36, seen [here](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/batch.html#Batch.Client.describe_job_definitions) in the docs, which includes `tags`, either populated with the given tags, or as an empty dict.

Assuming I haven't just misunderstood how this is working, I would propose adding tags to `JobDefinition`, in the same way that `parameters` is added [here](https://github.com/spulec/moto/blob/9859d66ff8fb7375fadf89fe976452d8c841459b/moto/batch/models.py#L182). If you think this change makes sense, I would be happy to work on it.

bblommers:
Hi @oakbramble, thanks for raising this! That's definitely something that's still missing from Moto, so if you'd like to contribute, that would be very welcome.
[...]

PR #4316, opened 2021-09-21
Add tagging to batch job definitions

oakbramble:
resolves spulec/moto#4284

I have updated batch job definition creation, such that tags will be associated with the definition.
Following [boto3](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/batch.html#Batch.Client.describe_job_definitions), the tags are submitted with the job def as a dict: 
```
tags={
        'string': 'string'
    },
```
The tagging object is used to store the values, but when a batch job definition is retrieved, it will have the tags returned in a dict: 
```
'tags': {
                'string': 'string'
            },
```
If there are no tags, an empty dictionary will be returned:
```
'tags': {},
```
I have submitted at this point to check I am doing something along the lines you would find acceptable and any necessary feedback would be appreciated.

Currently this affects `register_job_definition()` and `describe_job_definitions()` as, admittedly, this solves the issue I need solving, however if there are other endpoints you feel are essential to change or add around tagging job definitions before this can be accepted, then let me know.

### DIFF
--- a/moto/batch/models.py
+++ b/moto/batch/models.py
@@ -217,14 +217,7 @@ class JobDefinition(CloudFormationModel):
         self.tagger.tag_resource(self.arn, tags or [])
 
     def _format_tags(self, tags):
-        formatted_tags = []
-        for item in tags.items():
-            tag = {}
-            tag["Key"] = item[0]
-            tag["Value"] = item[1]
-            print(tag)
-            formatted_tags.append(tag)
-        return formatted_tags
+        return [{"Key": k, "Value": v} for k, v in tags.items()]
 
     def _update_arn(self):
         self.revision += 1
--- a/moto/batch/models.py
+++ b/moto/batch/models.py
@@ -16,7 +16,7 @@ from moto.ec2 import ec2_backends
 from moto.ecs import ecs_backends
 from moto.logs import logs_backends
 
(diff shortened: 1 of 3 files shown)


moto/batch/models.py
```diff
@@ -198,14 +200,32 @@ def __init__(
...
+
+    def _format_tags(self, tags):
+        formatted_tags = []
+        for item in tags.items():
+            tag = {}
+            tag["Key"] = item[0]
+            tag["Value"] = item[1]
+            print(tag)
+            formatted_tags.append(tag)
+        return formatted_tags
```
bblommers, line 227:
```suggestion
        return [{"Key": k, "Value": v} for k, v in tags.items()]
```
This is more readable, IMO - and at the very least gets rid of the print-statement :)

bblommers (changes requested):

bblommers:
> Currently this affects register_job_definition() and describe_job_definitions() as, admittedly, this solves the issue I need solving, however if there are other endpoints you feel are essential to change or add around tagging job definitions before this can be accepted, then let me know.

This is absolutely fine like this @oakbramble - if other people require tagging for other endpoints, they are free to open a new PR :slightly_smiling_face:

bblommers (approved):
LGTM! Thanks for the PR

PR #4321, opened 2021-09-21
EC2 - instance lifecycle defaults to None when not using spot

steved:
Follow-up from spulec/moto#4260.

Using a real boto3 client:
```
>>> list(boto3.resource("ec2", region_name="us-west-2").instances.all())[0].instance_lifecycle is None
True
```


steved:
cc @bblommers

bblommers (approved):
Nice one - thanks for the fix @steved!

PR #4322, opened 2021-09-22
Implement EC2 describe_vpc_endpoint_services()

kbalk:


moto/config/models.py
```diff
@@ -854,6 +854,29 @@ def __init__(self):
         self.config_rules = {}
         self.config_schema = None
 
+    @staticmethod
+    def default_vpc_endpoint_service(region, zones, random_number_func):
+        """List of dicts representing default VPC endpoints for this service."""
+        return [
```
bblommers, line 860: Is there any value in moving this dict to `core/models.py`? I don't know what the best approach is to handle this, mind you, it just seems that duplicating this in every service is not very DRY. 

As a potential alternative: maybe each service could call `BaseModel.default_vpc_endpoint_service(region, zones, func, service_name, service_types)`, which in turn returns this dictionary, properly formatted for that service.

moto/ec2/models.py
```diff
@@ -3831,19 +3831,173 @@ def describe_vpc_endpoints(self, vpc_end_point_ids, filters=None):
...
+        # function will be passed as an argument.
+        def vpce_random_number():
+            """Create a 17 digit number of lowercase hex."""
+            return "".join([random.choice(string.hexdigits.lower()) for i in range(17)])
+
+        from moto import backends  # pylint: disable=import-outside-toplevel
+
+        default_services = []
+        for _backends in backends.unique_backends():
+            for region_name, backend in _backends.items():
```
bblommers, line 3854: Potential performance improvement:
Instead of iterating over all regions, we could just do with two ifs.
```
if region_name in _backends:
    ....
if "global" in _backends:
    ...
```

moto/config/models.py
kbalk: Each service may or may not have it's own vpc endpoint or endpoints, so I was keeping the endpoint info a part of the the service by making it a part of the service's model.  

I could just build a master list, but every time someone adds a service, they would need to update that master list.  Having a method in the model serves as a reminder that they need to determine if the service has endpoint(s) and they need to add them.

Re: `BaseModel.default_vpc_endpoints_service(region, zones, func, service_name, service_types)`, the EC2 `describe_vpc_endpoint_services` already has all the logic to filter a list of default services, but it expects to have a master list of services for the appropriate region.  A function that generated a master list just needs the region and zone.  It can define the func itself and the service_name and service_type is used by the EC2 API.

Not all services have endpoints, not all have both a service type of interface and gateway.  Some services can have more than one set of entries, such as for autoscaling and autoscaling-plans (I have no idea yet if both are supported).

In the original issue, I only showed the config endpoints, but that was because the call to the API was filtering by service name and by filters.

I don't know the best answer either and sometimes two heads are better than one.  You also know the code better than I.  I can either do a master list or a method in the model.

moto/ec2/models.py
kbalk: Doh!  Yes, I can use the region as a key.  I struggled with that function and get a bit lost.

moto/config/models.py
kbalk: I've been finishing up the unit tests involving the next token and realized that the random number generator is being invoked for each invocation of `describe_vpc_endpoint_services`.  The ServiceIds must be generated only once.  So however you decide you want this API implemented, I will need to consider how I generate those IDs a single time.

kbalk: I think that's what you were trying to tell me.  Once I create the list, I can check the list's existence in subsequent calls and not re-create it.

kbalk: I went back to your comments on the issue and you said there would be a problem with generating the list once due to context managers?  So I'm back to the issue of figuring out how to generate the random numbers just once.

I'm creating the list of endpoint services right now, but otherwise I'm stuck until I hear back from you with regards to creating the list in one place or per service model.

bblommers: > I don't know the best answer either and sometimes two heads are better than one. You also know the code better than I. I can either do a master list or a method in the model.
>


The method in the model definitely seems the best approach. I was just hoping to reduce the duplication in each model. 
So for instance, S3 could call: 
`return BaseModel.default_vpc_endpoint_service(region, "s3", service_types=["Gateway", "Interface"])`

Autoscaling could call: `return BaseModel.default_vpc_endpoint_service(region, "autoscaling", ..) + BaseModel.default_vpc_endpoint_service(region, "autoscaling-plans", ..)`

Etc.

The nitty gritty of what the exact response looks like can be handled in BaseModel.

kbalk: Do you mean `BaseBackend` rather than `BaseModel`?  I ask because there's no S3 model.

Now I see what you're saying about duplication.  So we're still modifying the relevant backends to add the call to `default_vpc_endpoint_service()`, but the template is in the parent class?

I'll have to add an argument for the creation of privateDNSNames ... not all services have them.  Hopefully that will cover all the bases.

kbalk: Let me know if the latest modifications will work.

kbalk: I will have to create special entries for aws.sagemaker.region.notebook and aws.sagemaker.region.studio by creating them within the sagemaker's `default_vpc_endpoint_service()` and returning them along with the standard entry.

bblommers: > Do you mean BaseBackend rather than BaseModel? I ask because there's no S3 model.
> 

Yes, you're right, that's what I meant - sorry for the confusion. Looks good so far 👍

kbalk:
@bblommers I generate the privateDnsNames (plural) info, but it's not being returned from the API.  It goes into the botocore libraries and gets removed.  This is a field that was added last year: https://awsapichanges.info/archive/changes/d2b896-ec2.html

I don't know how you manage the botocore libraries and "shapes", but the "shape" for this API would need to be updated.

bblommers:
> I don't know how you manage the botocore libraries and "shapes", but the "shape" for this API would need to be updated.

Hey @kbalk, that's managed directly by boto3, and taken directly from the `botocore` dependency - so as long as the dependencies are up to date, it should work

kbalk:
@bblommers I just remembered there is a elbv2 and a rds2 ... I added default VPC endpoint services to elb and rds.  How are the version 2's handled?  I don't want duplicate sets of endpoint services.

bblommers (changes requested):

moto/dynamodb/models.py
```diff
@@ -317,6 +317,19 @@ class DynamoDBBackend(BaseBackend):
     def __init__(self):
         self.tables = OrderedDict()
 
+    @staticmethod
+    def default_vpc_endpoint_service(service_region, zones):
```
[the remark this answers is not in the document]
kbalk: ok

bblommers (approved):
LGTM @kbalk. This was a massive piece - thank you so much for putting in the work!

Issue #4312, opened 2021-09-20
kms.list_resource_tags does not support KeyArn parameter


rtoma:
Hi,

Expected behaviour: call `kms.list_resource_tags(keyARN)` and get the tags back (if the key was created before).

Actual behaviour:
```
botocore.errorfactory.NotFoundException: An error occurred (NotFoundException) when calling
  the ListResourceTags operation: The request was rejected because the specified entity or resource
  could not be found.
```

Boto3 docs https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/kms.html#KMS.Client.list_resource_tags says you can pass KeyID or KeyARN as parameter.

I traced the `moto.kms.models.list_resource_tags` method and I see it does a dict.get on a dict keyed by KeyID. So, it makes sense a lookup by KeyARN fails.

Thanks again for developing moto!

bblommers:
Hi @rtoma,thanks for raising this! Marking it as a bug, as we should support both options.

bblommers:
This should be fixed as of moto >= 2.2.8.dev23 - let us know whether we've missed anything.

PR #4323, opened 2021-09-22
Fixes issue #4312

stromp:
- Adds ARN support to the tagging calls
- Adds test to test ARN support


bblommers (approved):
LGTM @stromp - thanks for the PR!

PR #4334, opened 2021-09-23
Fix list_job_excecutions_for_thing with status parameter

gofabian:
If the methods `list_job_executions_for_thing` and `list_job_executions_for_job` are used with an explicit `status` parameter then it will always run into a `KeyError "status"`.

[log.txt](https://github.com/spulec/moto/files/7218426/log.txt)

This PR fixes a bug where wrong fields of job executions are accessed when filtering by status.

### DIFF
--- a/moto/iot/models.py
+++ b/moto/iot/models.py
@@ -1338,7 +1338,7 @@ class IoTBackend(BaseBackend):
         if status is not None:
             job_executions = list(
                 filter(
-                    lambda elem: status in elem["status"] and elem["status"] == status,
+                    lambda elem: "status" in elem["jobExecutionSummary"] and elem["jobExecutionSummary"]["status"] == status,
                     job_executions,
                 )
             )
--- a/moto/iot/models.py
+++ b/moto/iot/models.py
@@ -1338,7 +1338,7 @@ class IoTBackend(BaseBackend):
         if status is not None:
             job_executions = list(
                 filter(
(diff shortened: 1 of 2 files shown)


moto/iot/models.py
```diff
@@ -1338,7 +1338,7 @@ def list_job_executions_for_thing(
         if status is not None:
             job_executions = list(
                 filter(
-                    lambda elem: status in elem["status"] and elem["status"] == status,
+                    lambda elem: "status" in elem["jobExecutionSummary"] and elem["jobExecutionSummary"]["status"] == status,
```
bblommers, line 1341:
```suggestion
                    lambda elem: elem["jobExecutionSummary"].get("status") == status
```

bblommers: This maybe a  more readable alternative - logically it does the same as far as I can see

gofabian:
> tests/test_s3/test_s3.py::test_missing_key_request_boto3 FAILED

@bblommers  Any idea what is the problem here? I have no clue why my changes should have influenced that test?

bblommers:
> 
> 
> > tests/test_s3/test_s3.py::test_missing_key_request_boto3 FAILED
> 
> @bblommers Any idea what is the problem here? I have no clue why my changes should have influenced that test?

Hey @gofabian , that looks like an issue with our CI - every now and then the Docker container decides not to respond. I've restarted the tests - should be fine the second time 🤞

bblommers (approved):
LGTM! Thanks for the fix @gofabian

PR #4343, opened 2021-09-24
Fix validation of ConfigService PutConfigRule InputParameters

kbalk:
Valid InputParameters were being flagged as invalid, causing an InvalidParameterValueException to be raised.


bblommers (approved):
LGTM!

Issue #4341, opened 2021-09-24
ListUserTags output is incorrect


farrellit:
ListUserTags should have a result `member` under `Tags`, currently there is `item` instead.

https://docs.aws.amazon.com/IAM/latest/APIReference/API_ListUserTags.html

Discovered while using localstack with Go SDK.   Expected the SDK to successfully list user tags, but since output doesn't match the spec, it doesn't find the tags in the response. 

Installed moto via docker localstack.

bblommers:
Thanks for raising this @farrellit - the Python SDK is a bit more forgiving with the exact XML format, that's why we don't tend to notice mistakes like this.

PR #4350, opened 2021-09-26
Use proper xml tag for ListUserTagsResult's Tags

farrellit:
boto3 is forgiving on this front but other SDKS are not.  Since moto is
used behind localstack it causes some SDKs (namely Go SDK) to fail to
properly look up tags on localstack.

I made this change on my localstack container and it did indeed fix the Go SDK behavior.

https://docs.aws.amazon.com/IAM/latest/APIReference/API_ListUserTags.html

This should resolve https://github.com/spulec/moto/issues/4341


bblommers (approved):
Thanks for the fix @farrellit!

Issue #4157, opened 2021-08-09
MotoServer: Cloudformation Update Stack should fail with 400 ValidationError instead of returning 200 success when using the same input


modamod:
I am using motoserver to test a package I am building in golang,

During testing I noticed that the update stack action returns 200 success instead of throwing a 400 ValidationError when using the same template input.

Below is roughly the code in python along with the output I got from Motoserver and the expected output

Code
```python
import boto3
import time


client = boto3.Session(region_name="us-east-1").client("cloudformation", endpoint_url='http://localhost:5000')
paramters = [{
    "ParameterKey": "TopicName",
    "ParameterValue": "ParameterValue"
}]
template_body = open("fixtures/templates/valid_template.yaml").read()
stack_name = "TestStack"
# client.create_stack(StackName=stack_name, TemplateBody=template_body, Parameters=paramters)
# time.sleep(1)

# Should fail since I am updating the stack with the same paramters and template body
result = client.update_stack(StackName=stack_name, TemplateBody=template_body, Parameters=paramters)
print(result)
```

MotoServer Response
```
{'StackId': 'arn:aws:cloudformation:us-east-1:123456789:stack/TestStack/b8366e49-805b-416e-958d-8b21baa27ec1', 'ResponseMetadata': {'RequestId': 'b9b4b068-3a41-11e5-94eb-example', 'HTTPStatusCode': 200, 'HTTPHeaders': {'server': 'amazon.com', 'content-type': 'text/html; charset=utf-8', 'content-length': '347', 'access-control-allow-origin': '*', 'date': 'Mon, 09 Aug 2021 22:54:51 GMT'}, 'RetryAttempts': 0}}
```

AWS API Response
```
Traceback (most recent call last):
  File "C:\Users\extreme pc\git\modamod\ami-inspector\sample.py", line 12, in <module>
    client.create_stack(StackName=stack_name, TemplateBody=template_body, Parameters=paramters)
  File "C:\Users\extreme pc\venv\lib\site-packages\botocore\client.py", line 386, in _api_call
    return self._make_api_call(operation_name, kwargs)
  File "C:\Users\extreme pc\venv\lib\site-packages\botocore\client.py", line 705, in _make_api_call
    raise error_class(parsed_response, operation_name)
botocore.errorfactory.AlreadyExistsException: An error occurred (AlreadyExistsException) when calling the CreateStack operation: Stack [TestStack] already exists
```

bblommers:
[...]

PR #4351, opened 2021-09-26
feat: Adding validation to Cloudformation to reject same template_body

teddylear:
feat: Adding validation to Cloudformation to reject same template_body and parameters

First PR to moto, please let me know if I'm missing anything or if anything should be changed. 

Addresses #4157

### DIFF
--- a/tests/test_cloudformation/test_cloudformation_stack_crud_boto3.py
+++ b/tests/test_cloudformation/test_cloudformation_stack_crud_boto3.py
@@ -1044,10 +1044,18 @@ def test_boto3_update_stack_fail_update_same_template_body():
     exp_err = exp.value.response.get("Error")
     exp_metadata = exp.value.response.get("ResponseMetadata")
 
-    exp_err.get("Code").should.match(r"ValidationError")
-    print(f"Message recieved: {exp_err.get('Message')}")
-    assert exp_err.get("Message") == f"Stack [{name}] already exists"
+    exp_err.get("Code").should.equal("ValidationError")
+    exp_err.get("Message").should.equal(f"Stack [{name}] already exists")
     exp_metadata.get("HTTPStatusCode").should.equal(400)
+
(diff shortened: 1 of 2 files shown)


tests/test_cloudformation/test_cloudformation_stack_crud_boto3.py
```diff
@@ -1023,6 +1021,35 @@ def test_boto3_update_stack_fail_missing_new_parameter():
...
+            StackName=name,
+            TemplateBody=dummy_template_yaml_with_ref,
+            Parameters=params,
+        )
+    exp_err = exp.value.response.get("Error")
+    exp_metadata = exp.value.response.get("ResponseMetadata")
+
+    exp_err.get("Code").should.match(r"ValidationError")
+    print(f"Message recieved: {exp_err.get('Message')}")
+    assert exp_err.get("Message") == f"Stack [{name}] already exists"
```
bblommers, line 1049:
```suggestion
    exp_err.get("Code").should.equal("ValidationError")
    exp_err.get("Message").should.equal(f"Stack [{name}] already exists")
```
This is more readable IMO

bblommers (changes requested):
Hi @teddylear, thanks for the PR!
Would you mind extending the test, and call the `update_stack`-method with updated parameters at the end?
There's no need to verify that the update-method itself works as expected, that hasn't changed - just to verify that the error will not be thrown with different params.

tests/test_cloudformation/test_cloudformation_stack_crud_boto3.py
teddylear: Agreed, updated to match this.

teddylear:
@bblommers 
Updated the test to call update stack with new parameters. Please let me know if there's anything else I should update.

bblommers (approved):
LGTM! Thanks for the PR @teddylear

PR #4355, opened 2021-09-27
Deregister batch job definition by 'name:revision' 

### LINKED ISSUE #4354: Bug when deregistering job definition by name/ revision
When deregistering a batch job definition with `deregister_job_definition()` the required parameter is as follows:
```
    jobDefinition (string) --

    [REQUIRED]

    The name and revision (name:revision ) or full Amazon Resource Name (ARN) of the job definition to deregister.
```
Currently when trying to deregister with `name:revision` it is not working as the `name:revision` string is split up and `revision` ends up type `str`. Due to this, when it is compared to the definition to be deleted it does not match as `revision` is stored in the `jobDefinition` object as an `int`. This isn't caught in the current testing as the test requests deletion by ARN.

I have made a fix to correct this and will attach the pull request.

oakbramble:
closes #4354


bblommers (approved):
LGTM! Thanks for the fix @oakbramble

PR #1024, opened 2017-07-26
Add add, remove, list endpoints for SSM tags

jpb:
Adds the following SSM endpoints:

- [add_tags_to_resource](http://boto3.readthedocs.io/en/latest/reference/services/ssm.html#SSM.Client.add_tags_to_resource)
- [list_tags_for_resource](http://boto3.readthedocs.io/en/latest/reference/services/ssm.html#SSM.Client.list_tags_for_resource)
- [remove_tags_from_resource](http://boto3.readthedocs.io/en/latest/reference/services/ssm.html#SSM.Client.remove_tags_from_resource)

PR #4361, opened 2021-09-29
[issue-4360] Fix InvalidResource Id,Type responses for ssm tagging methods

### LINKED ISSUE #4360: SSM tag methods should error with InvalidResourceId if not there
In #1024 changes were made to emulate tagging methods in the ssm subsystem.  Unfortunately they do not behave like AWS does.  When a resource is not available, AWS returns an `InvalidResourceId` error.  Here is an example I recorded with vcrpy:

```
  request:
    method: POST
    uri: https://ssm.us-west-2.amazonaws.com/
    body: '{"ResourceType": "Parameter", "ResourceId": "special-not-there-4"}'
  response:
    body:
      string: '{"__type":"InvalidResourceId"}'
    status:
      code: 400
      message: Bad Request
```

Here is the response from moto, which is incorrect:

```
(Pdb) pp client.list_tags_for_resource(ResourceType="Parameter", ResourceId="special-not-there-4")
{'ResponseMetadata': {'HTTPHeaders': {'server': 'amazon.com'},
                      'HTTPStatusCode': 200,
                      'RetryAttempts': 0},
 'TagList': []}
```

jeking3:
In #1024 changes were made to emulate tagging methods in the ssm subsystem.  Unfortunately they do not behave like AWS does.  When a resource is not available, AWS returns an `InvalidResourceId` error.  Here is an example I recorded with vcrpy:

```
  request:
    method: POST
    uri: https://ssm.us-west-2.amazonaws.com/
    body: '{"ResourceType": "Parameter", "ResourceId": "special-not-there-4"}'
  response:
    body:
      string: '{"__type":"InvalidResourceId"}'
    status:
      code: 400
      message: Bad Request
```

Here is the response from moto, which is incorrect:

```
(Pdb) pp client.list_tags_for_resource(ResourceType="Parameter", ResourceId="special-not-there-4")
{'ResponseMetadata': {'HTTPHeaders': {'server': 'amazon.com'},
                      'HTTPStatusCode': 200,
                      'RetryAttempts': 0},
 'TagList': []}
```

I also fixed the case where a Parameter or Document is deleted, but the tags were not cleaned up.

This closes #4360

### DIFF
--- a/moto/ssm/exceptions.py
+++ b/moto/ssm/exceptions.py
@@ -23,6 +23,24 @@ class InvalidFilterValue(JsonRESTError):
         super(InvalidFilterValue, self).__init__("InvalidFilterValue", message)
 
 
+class InvalidResourceId(JsonRESTError):
+    code = 400
+
+    def __init__(self):
+        super(InvalidResourceId, self).__init__(
+            "InvalidResourceId", "Invalid Resource Id"
+        )
+
+
+class InvalidResourceType(JsonRESTError):
+    code = 400
+
+    def __init__(self):
+        super(InvalidResourceType, self).__init__(
+            "InvalidResourceType", "Invalid Resource Type"
+        )
+
+
 class ParameterNotFound(JsonRESTError):
     code = 400
 
--- a/moto/ssm/models.py
+++ b/moto/ssm/models.py
@@ -38,6 +38,8 @@ from .exceptions import (
     ParameterMaxVersionLimitExceeded,
     DocumentPermissionLimit,
     InvalidPermissionType,
+    InvalidResourceId,
+    InvalidResourceType,
 )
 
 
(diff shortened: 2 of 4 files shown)


tox.ini
```diff
@@ -15,5 +15,5 @@ commands =
     pytest -v {posargs}
 
 [flake8]
-ignore = W503,W605,E128,E501,E203,E266,E501,E231
+ignore = W503,W605,E128,E501,E203,E266,E501,E231,SFS101,SFS201,SFS301
```
bblommers, line 18: We don't use the SFS plugin, so these ignore-statements are not necessary

bblommers (changes requested):
Hi @jeking3, thanks for the PR! 
Just got a single comment about the flake8-config change - otherwise bit looks fine

tox.ini
jeking3: They were necessary in my shared sandbox.  I will remove them.

bblommers (approved):
LGTM! Thanks for the PR @jeking3

PR #4362, opened 2021-09-30
[issue-4283] ssm put_parameter with tags and overwrite should error

### LINKED ISSUE #4283: Tagging SSM parameter when overwriting should throw validation error
# MWE

```
import boto3
from moto import mock_ssm

mock_ssm().start()

boto3.client('ssm').put_parameter(
    Name='test-param',
    Description=f'Delete after 14/9/21',
    Value='abc',
    Type='String',
    Overwrite=True,
    Tags=[
        {
            'Key': 'mytag',
            'Value': 'myval'
        }
    ],
)
```

## Behavior with moto

The script runs without throwing an exception

## behavior without moto

```
  File "param.py", line 4, in <module>
    boto3.client('ssm').put_parameter(
  File "/home/ec2-user/.pyenv/versions/3.8.11/lib/python3.8/site-packages/botocore/client.py", line 386, in _api_call
  File "/home/ec2-user/.pyenv/versions/3.8.11/lib/python3.8/site-packages/botocore/client.py", line 705, in _make_api_call
botocore.exceptions.ClientError: An error occurred (ValidationException) when calling the PutParameter operation: Invalid request: tags and overwrite can't be used together. To create a parameter with tags, please remove overwrite flag. To update tags for an existing parameter, please use AddTagsToResource or RemoveTagsFromResource.
```

jeking3:
This closes #4283

### DIFF
--- a/moto/ssm/models.py
+++ b/moto/ssm/models.py
@@ -1547,6 +1547,12 @@ class SimpleSystemManagerBackend(BaseBackend):
                 "1 validation error detected: Value '' at 'value' failed to satisfy"
                 " constraint: Member must have length greater than or equal to 1."
             )
+        if overwrite and tags:
+            raise ValidationException(
+                "Invalid request: tags and overwrite can't be used together. To create a "
+                "parameter with tags, please remove overwrite flag. To update tags for an "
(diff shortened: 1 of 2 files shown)


bblommers (approved):
LGTM! Thanks @jeking3

PR #4363, opened 2021-09-30
Require at least one destination config for create_delivery_stream()

kbalk:


bblommers (approved):
LGTM - thanks for the fix @kbalk!

PR #4366, opened 2021-10-01
Update CONTRIBUTING.md

### LINKED ISSUE #4275: Possible additions to CONTRIBUTING.md
Some suggestions for `CONTRIBUTING.md` to help contributors to moto:

- Promote the use of TaggingService.
- Describe how to run in test server mode and the environment variable `TEST_SERVER_MODE`.  You can work it out by looking at the github workflow file, but it might be nice to spell out the steps.  It's useful to know if your tests fail in server mode.  Better yet, add a rule to the Makefile to invoke it.   
- Within a unit test - don't test an exception to indicate something is not yet implemented.  If an exception is raised that isn't defined in an `exceptions.py` file and a unit test attempts to check that scenario, the test will fail when run in server mode.  Mention the use of `SkipTest`.
- It would be nice to encourage the use of a "NotImplementedError" exception where it makes sense, but it's not something that can be tested in server mode and that lowers your test coverage a little bit.  Maybe a generic error that can be defined with an understanding of `TEST_SERVER_MODE`?
- Provide some general guidelines on how to organize tests - I see monolithic test files and sometime test functions that handle a lot of tests.  I'm used to writing one test per function, but to be consistent, I've been combining multiple tests per API.  It all works, but it would be nice to see test files that aren't so large and test functions that are smaller.
- When importing the backend from another service, declare the import within the block of code where it's needed and not at the top of the file.  I don't know if recent changes have negated that need to improve performance.

These are some of the gotchas I've discovered the hard way or learned though code review comments.

bblommers:
Fixes #4275 

@kbalk Some improvements to the Contributing document, following your suggestions.  Hopefully this makes onboarding easier for new contributors. Let me know what you think.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,3 +1,5 @@
+# Table of Contents
+
 - [Contributing code](#contributing-code)
   * [Running the tests locally](#running-the-tests-locally)
   * [Linting](#linting)
@@ -39,9 +41,10 @@ TEST_SERVER_MODE=true pytest -sv tests/test_service/..
  - When writing tests, one test should only verify a single feature/method. I.e., one test for `create_resource()`, another for `update_resource()`, etc.
  - When writing negative tests, try to use the following format:
  ```
- with pytest.raises(botocore.exceptions.ClientError) as ex:
+ with pytest.raises(botocore.exceptions.ClientError) as exc:
      client.failing_call(..)
- err = ex.value.response["Error"]
+ err = exc.value.response["Error"]
+ # These assertions use the 'sure' library, but any assertion style is accepted
  err["Code"].should.equal(..)
  err["Message"].should.equal(..)
  ```
@@ -70,7 +73,7 @@ How to teach Moto to support a new AWS endpoint:
 
 Implementing a new service from scratch is more work, but still quite straightforward. All the code that intercepts network requests to `*.amazonaws.com` is already handled for you in `moto/core` - all that's necessary for new services to be recognized is to create a new decorator and determine which URLs should be intercepted.
 
-See this PR for an example of what's involved in creating a new service: https://github.com/spulec/moto/pull/2409/files
+See this PR for an example of what's involved in creating a new service: https://github.com/spulec/moto/pull/4076/files
 
 Note the `urls.py` that redirects all incoming URL requests to a generic `dispatch` method, which in turn will call the appropriate method in `responses.py`. 
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,3 +1,13 @@
+- [Contributing code](#contributing-code)
+  * [Running the tests locally](#running-the-tests-locally)
+  * [Linting](#linting)
+  * [General notes](#general-notes)
+- [Missing features](#missing-features)
+- [Missing services](#missing-services)
+    + [Generating template code of services.](#generating-template-code-of-services)
+    + [URL Indexing](#url-indexing)
+- [Maintainers](#maintainers)
+
 # Contributing code
 
 Moto has a [Code of Conduct](https://github.com/spulec/moto/blob/master/CODE_OF_CONDUCT.md), you can expect to be treated with respect at all times when interacting with this project.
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -1,3 +1,13 @@
+- [Contributing code](#contributing-code)
```
kbalk, line 1: Could you add a header before this text, e.g., "Table of Contents"?  It looks like something is missing without a title at the top.  

Or maybe an organization like this (without the bullets):

Contributing Code
- Table of Contents

General Guidelines for Contributing
-   Running the tests locally
-   Linting
-   General notes

Missing features ...

CONTRIBUTING.md
```diff
@@ -11,14 +21,39 @@ You should be able to run `make init` to install the dependencies and then `make
...
+```
+ - When writing tests, one test should only verify a single feature/method. I.e., one test for `create_resource()`, another for `update_resource()`, etc.
+ - When writing negative tests, try to use the following format:
+ ```
+ with pytest.raises(botocore.exceptions.ClientError) as ex:
+     client.failing_call(..)
+ err = ex.value.response["Error"]
+ err["Code"].should.equal(..)
+ err["Message"].should.equal(..)
+ ```
```
kbalk, line 47: pylint expects variables to be at least three characters long, so could you change "ex" to "exc" in the example above?

kbalk: Would you want to allow the possibility of using assert() as well as `sure`?  

It might be worth noting that the above code is using methods from the `sure` library ... I hadn't seen or used that library before this and if I saw that code, I might be a little confused by it.

CONTRIBUTING.md
```diff
@@ -31,12 +66,7 @@ How to teach Moto to support a new AWS endpoint:
 * Implementing the feature itself can be done by creating a method called `import_certificate` in `moto/acm/responses.py`. It's considered good practice to deal with input/output formatting and validation in `responses.py`, and create a method `import_certificate` in `moto/acm/models.py` that handles the actual import logic.
 * If you can also implement the code that gets that test passing then great! If not, just ask the community for a hand and somebody will assist you.
 
-## Before pushing changes to GitHub
-
-1. Run `black moto/ tests/` over your code to ensure that it is properly formatted
-1. Run `make test` to ensure your tests are passing
-
-## Missing services
+# Missing services
```
kbalk, line 69: The problem with referencing the PRs in this section is that those PRs import some Python 2 libraries, `urls_base` doesn't have the escape characters in the urls regex and the old `backend.py` is referenced.  

I looked at one of these PRs when I was trying to add a new service and thought it looked outdated, so then I tried the scaffold script.  I dunno.  

It took me a while to figure out how to interact with `click` of the scaffold script and I almost gave up in frustration.  I think the tab key expands the current selection and the arrow keys move up/down the lists, but I'm not sure I'm remembering correctly.

kbalk:
@bblommers Thank you for updating the document.  There was one point that didn't make it into the document and maybe that was a deliberate choice:

"_When importing the backend from another service, declare the import within the block of code where it's needed and not at the top of the file. I don't know if recent changes have negated that need to improve performance._"

CONTRIBUTING.md
bblommers: Done and done

CONTRIBUTING.md
bblommers: Good point - I've added the header

CONTRIBUTING.md
bblommers: Agreed - I've updated one of the PR's to a more recent one. 
There haven't been any new services since the updated urls/backend, unfortunately, but when one comes along, I'll update this doc again.

bblommers:
> @bblommers Thank you for updating the document. There was one point that didn't make it into the document and maybe that was a deliberate choice:
> 
> "_When importing the backend from another service, declare the import within the block of code where it's needed and not at the top of the file. I don't know if recent changes have negated that need to improve performance._"

Yeah, that was deliverate @kbalk. It's a bit of a balance: on one hand it's a useful point, but on the other, if there are too many pointers, users will just gloss over things. In this case, it does not happen often enough to warrant a dedicated bullet point, IMO. 

Thanks for the review!

kbalk:
@bblommers I appreciate you doing this.  Thank you!

I did remember one other thing ... I think you need to install the click library before you can run the scaffold script.  A user can figure it out, but it'll be one less frustration if they're warned in advance.  That and some advice on selecting options (use of the tab key) following the prompts!

PR #4367, opened 2021-10-03
fixed object locking mechanism in s3

macnev2013:
## Fixed
object locking mechanism in s3

## Added
S3_OBJECT_LEGAL_HOLD 

## Test Passing
TestAccAWSS3Bucket_forceDestroyWithObjectLockEnabled

## Note
This is a test PR. Please do not merge as of now.

### DIFF
--- a/moto/s3/responses.py
+++ b/moto/s3/responses.py
@@ -832,10 +832,10 @@ class ResponseObject(_TemplateEnvironmentMixin, ActionAuthenticatorMixin):
                     bucket_name, self._acl_from_headers(request.headers)
                 )
 
-            if request.headers.get("x-amz-bucket-object-lock-enabled", "") in [
-                "true",
-                "True",
-            ]:
+            if (
+                request.headers.get("x-amz-bucket-object-lock-enabled", "").lower()
+                == "true"
+            ):
                 new_bucket.object_lock_enabled = True
                 new_bucket.versioning_status = "Enabled"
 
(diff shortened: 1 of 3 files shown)


moto/s3/responses.py
```diff
@@ -832,7 +832,10 @@ def _bucket_response_put(
                     bucket_name, self._acl_from_headers(request.headers)
                 )
 
-            if request.headers.get("x-amz-bucket-object-lock-enabled", "") == "True":
+            if request.headers.get("x-amz-bucket-object-lock-enabled", "") in [
```
whummer, line 835: nit: The true strings could be put into a re-usable global constant like `TRUE_STRINGS = ["true", "True"]`. Perhaps such a constant already exists in the codebase @bblommers ?

whummer (approved):
LGTM, only two minor remarks - thanks for adding this @macnev2013 .

Waiting for review from @bblommers .. 🙏 Thanks

moto/s3/responses.py
```diff
@@ -1567,23 +1574,29 @@ def _key_response_head(self, bucket_name, query, key_name, headers):
...
         enabled = (
             parsed_xml["ObjectLockConfiguration"]["ObjectLockEnabled"] == "Enabled"
         )
+        response_dict["enabled"] = enabled
+
+        default_retention = parsed_xml.get("ObjectLockConfiguration").get("Rule")
+        if default_retention:
+            default_retention = default_retention.get("DefaultRetention")
+            mode = default_retention["Mode"]
+            days = int(default_retention["Days"]) if "Days" in default_retention else 0
```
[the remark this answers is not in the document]
bblommers: Agreed, that's much easier to read IMO

moto/s3/responses.py
bblommers: In other places we tend to solve this with `x.lower() == "true"`, so for consistency sake, would be good to have the same here

whummer:
> @whummer @bblommers, This PR will have some breaking effect on localstack. Some feature could break due to this. As a fix i have created a new PR in localstack here.

Thanks @macnev2013 - should be good to merge here. 👍 (we'll take care of merging / resolving on LocalStack side on our end then..). Thanks @bblommers

PR #4385, opened 2021-10-10
fixed default kms keys;fixed object locking exception

macnev2013:
## Added
- Default kms keys creation
- Exception added for if object locking config does not exists

## Fixed
- Default object encryption if bucket encryption is enabled
- Fixed bucket encryption storing mechanism

## Fixed TF Test
- TestAccAWSS3BucketObject_ (Few test from this are failing)

### DIFF
--- a/moto/kms/models.py
+++ b/moto/kms/models.py
@@ -158,11 +158,17 @@ class Key(CloudFormationModel):
 
 
 class KmsBackend(BaseBackend):
-    def __init__(self):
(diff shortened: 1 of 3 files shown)


bblommers:
Just played around with it @macnev2013, that's indeed what's happening.
Ideally Moto shows the same behaviour. List reserved aliases without any key on startup, but create them when the users calls `describe_key`.

PR #4404, opened 2021-10-12
EC2 - Bugfix where associate_public_ip was always True

### LINKED ISSUE #2762: AssociatePublicIpAddress = False on primary network interface is ignored
When creating an ec2 instance and a primary network interface with `"AssociatePublicIpAddress": False`, an instance is created with a public IP address.

I am expecting no public IP address to be created.

Example request:

```
kwargs = {
        "ImageId": ami,
        "MaxCount": 1,
        "MinCount": 1,
        "KeyName": "test-key",
        "InstanceType": "t2.medium",
        "EbsOptimized": False,
        "NetworkInterfaces": [
            {
                "DeviceIndex": 0,
                "Description": "Primary network interface",
                "DeleteOnTermination": True,
                "SubnetId": "subnet-317243a9",
                "Ipv6AddressCount": 0,
                "Groups": "default",
                "AssociatePublicIpAddress": False,
            }
        ]
    }
```

I'm using moto 1.3.14

bblommers:
Fixes #2762

The `public_ip_auto_assign`-attribute was persisted as a string, so `if"false"` would always evaluate to True

### DIFF
--- a/tests/test_ec2/test_instances.py
+++ b/tests/test_ec2/test_instances.py
@@ -3044,6 +3044,58 @@ def test_create_instance_with_launch_template_id_produces_no_warning(
     assert len(captured_warnings) == 0
 
 
+@mock_ec2
+def test_run_instance_and_associate_public_ip():
+    ec2 = boto3.resource("ec2", "us-west-1")
+    vpc = ec2.create_vpc(CidrBlock="10.0.0.0/16")
+    subnet = ec2.create_subnet(VpcId=vpc.id, CidrBlock="10.0.0.0/18")
+
+    # Do not pass AssociatePublicIpAddress-argument
+    instance = ec2.create_instances(
+        ImageId=EXAMPLE_AMI_ID,
+        MinCount=1,
+        MaxCount=1,
+        NetworkInterfaces=[{"DeviceIndex": 0}],
+        SubnetId=subnet.id,
+    )[0]
+    interfaces = instance.network_interfaces_attribute
+    addresses = interfaces[0]["PrivateIpAddresses"][0]
+    addresses.should.have.key("Primary").equal(True)
+    addresses.should.have.key("PrivateIpAddress")
+    addresses.shouldnt.have.key("Association")
+
+    # Pass AssociatePublicIpAddress=False
+    instance = ec2.create_instances(
+        ImageId=EXAMPLE_AMI_ID,
+        MinCount=1,
+        MaxCount=1,
+        NetworkInterfaces=[{"DeviceIndex": 0, "AssociatePublicIpAddress": False}],
+        SubnetId=subnet.id,
+    )[0]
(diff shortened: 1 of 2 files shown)


tests/test_ec2/test_instances.py
```diff
@@ -3044,6 +3044,58 @@ def test_create_instance_with_launch_template_id_produces_no_warning(
...
+        NetworkInterfaces=[{"DeviceIndex": 0}],
+        SubnetId=subnet.id,
+    )[0]
+    interfaces = instance.network_interfaces_attribute
+    addresses = interfaces[0]["PrivateIpAddresses"][0]
+    addresses.should.have.key("Primary").equal(True)
+    addresses.should.have.key("PrivateIpAddress")
+    addresses.shouldnt.have.key("Association")
+
+    # Pass AssociatePublicIpAddress=False
```
paikend, line 3067: Logic to check if public DNS is None seems to be necessary. 🙏

bblommers: Hi @paikend, please open a new issue if you're running into a bug, that makes it easier for  us to track

tests/test_ec2/test_instances.py
```diff
@@ -3044,6 +3044,58 @@ def test_create_instance_with_launch_template_id_produces_no_warning(
...
+        MinCount=1,
+        MaxCount=1,
+        NetworkInterfaces=[{"DeviceIndex": 0, "AssociatePublicIpAddress": False}],
+        SubnetId=subnet.id,
+    )[0]
+    interfaces = instance.network_interfaces_attribute
+    addresses = interfaces[0]["PrivateIpAddresses"][0]
+    addresses.should.have.key("Primary").equal(True)
+    addresses.should.have.key("PrivateIpAddress")
+    addresses.shouldnt.have.key("Association")
```
tekumara, line 3079: Shouldn't this check that there is no public IP address (or PublicDnsName)?

bblommers: I just verified this against AWS, and this is the response from the `describe_network_interfaces` call:
```
"PrivateIpAddresses": [{
    "Primary": true,
    "PrivateIpAddress": "10.0.33.93"
}]
```
AWS doesn't even have the 'Association'-key, let alone `Association.PublicIp`. So this assert looks correct to me.

Only when you pass `AssociatePublicIpAddress=True` does it return the `Association`-dict:
```
"PrivateIpAddresses": [{
    "Association": {
        "IpOwnerId": "amazon",
        "PublicDnsName": "",
        "PublicIp": "54.153.xx.yyy"
    },
    "Primary": true,
    "PrivateIpAddress": "10.0.62.61"
}]
```

tekumara: Ah gotcha thanks for the clarification!

PR #4412, opened 2021-10-14
Added admin_reset_user_password implementation

nluk:
Hi, I've added a pretty **basic** `admin_reset_user_password` implementation. Many corner cases in this area are still unexplored, and I will perhaps attempt to pursue them later on. In particular:

- Check user pool `AutoVerifiedAttributes`  impact on the unverified email/phone number exception
- Same for `AccountRecoverySetting`
- Check if, and how does `ConfirmSignUp` call affect `email_verified` and `phone_number_verified` (There's priority - cognito sends OTP to top priority channel, and then 'verifies' it once user sends it back maybe?)
- Check if these attributes show up at all in `get_user`, before OTP is sent - eg. after `admin_create_user`

I also have a few questions:

- Is there any particular reason for the user statuses to be defined as such a strange dictionary instead of an enum?
- Are there any plans to add some kind of a spy endpoint for moto server? For example you can check sent emails via importing SES backend in python code, but there's no way to accomplish that in server mode I think?
- How about adding a simple Cognito Hosted UI mock?

### DIFF
--- a/moto/cognitoidp/models.py
+++ b/moto/cognitoidp/models.py
@@ -27,6 +27,7 @@ UserStatus = {
     "FORCE_CHANGE_PASSWORD": "FORCE_CHANGE_PASSWORD",
     "CONFIRMED": "CONFIRMED",
     "UNCONFIRMED": "UNCONFIRMED",
+    "RESET_REQUIRED": "RESET_REQUIRED",
 }
 
 
@@ -267,6 +268,7 @@ class CognitoIdpUser(BaseModel):
         self.status = status
         self.enabled = True
         self.attributes = attributes
+        self.attribute_lookup = self._flatten_attributes(attributes)
         self.create_date = datetime.datetime.utcnow()
         self.last_modified_date = datetime.datetime.utcnow()
         self.sms_mfa_enabled = False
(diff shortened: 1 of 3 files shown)


bblommers:
Awesome - thanks for the PR @nluk!

>     * Is there any particular reason for the user statuses to be defined as such a strange dictionary instead of an enum?
> 

Not that I can see - an enum should do the job as well

>     * Are there any plans to add some kind of a spy endpoint for moto server? For example you can check sent emails via importing SES backend in python code, but there's no way to accomplish that in server mode I think?
> 
>     * How about adding a simple Cognito Hosted UI mock?

There is a dashboard for the MotoServer (http://localhost:5000/moto-api/) that shows the current data/models that are stored, and serves as rudimentary debugging tool.
The scope of Moto (mocking every boto3-service) is already quite large as it is, so I'm hesitant to expand that scope even more by introducing dedicated spy endpoints or hosted UI's. It's definitely useful, just too much for us to maintain.

If you are already using ServerMode, it may be worth checking out LocalStack. They add a lot of features on top of Moto, and I wouldn't be surprised if they have an option to intercept emails for example.

Issue #416, opened 2015-09-14
Unexpected behavior with @mock_s3 for a test suite


eliasdorneles:
Hey folks!

I was bitten by a weird behavior of moto today, I decided to take some time and build a test case for the behavior I'd expect:

```
import boto
import unittest
from moto import mock_s3


@mock_s3
class TestWithSetup(unittest.TestCase):
    def setUp(self):
        conn = boto.connect_s3()
        conn.create_bucket('mybucket')

    def test_should_find_bucket(self):
        conn = boto.connect_s3()
        self.assertIsNotNone(conn.get_bucket('mybucket'))

    def test_should_not_find_unknown_bucket(self):
        conn = boto.connect_s3()
        with self.assertRaises(Exception):
            conn.get_bucket('unknown_bucket')


@mock_s3
class TestWithPublicMethod(unittest.TestCase):
    def ensure_bucket_exists(self):
        conn = boto.connect_s3()
        conn.create_bucket('mybucket')

    def test_should_find_bucket(self):
        self.ensure_bucket_exists()

        conn = boto.connect_s3()
        self.assertIsNotNone(conn.get_bucket('mybucket'))

    def test_should_not_find_bucket(self):
        conn = boto.connect_s3()
        self.assertIsNone(conn.get_bucket('mybucket'))


@mock_s3
class TestWithPseudoPrivateMethod(unittest.TestCase):
    def _ensure_bucket_exists(self):
        conn = boto.connect_s3()
        conn.create_bucket('mybucket')

    def test_should_find_bucket(self):
        self._ensure_bucket_exists()
        conn = boto.connect_s3()
        self.assertIsNotNone(conn.get_bucket('mybucket'))

    def test_should_not_find_bucket(self):
        conn = boto.connect_s3()
        self.assertIsNone(conn.get_bucket('mybucket'))
```

This is related to commit https://github.com/spulec/moto/commit/3ed9428cb01cc0af157dfaf50fc4550e1a6d3ab5 for the issue https://github.com/spulec/moto/issues/363 -- I think this fix is not the best way, because it seems to be making the contents of the bucket global to the test suite, while the expected IMO would be to make it local to each test method.

Before that change, it seems that the contents of the bucket was indeed local to every method, but the result was unexpected because it was isolating also setUp and other non-test methods.

I'm not sure how to fix it, I hope the test case helps at least to decide which is the expected behavior.

eliasdorneles:
[...]

PR #4419, opened 2021-10-15
Refactor Class-decorator logic to reset per test

bblommers:
## DO NOT MERGE yet - this is a major change, and should only be merged as part of 3.x (if approved..)

This PR aims to change the way class-decorators work. At the moment, the class-decorator keeps a global state that persists over the duration of the class. In other words, resources created in one test are available other tests.

As pointed out in the conversation in #416, it might make more sense to always keep a local state. In other words: resources created in one test case should never show up in another test, regardless of the type of decorator is used (class, method, context, etc.).

To achieve this, this PR reworks the logic that decides when to call the reset-API. 

### Arguments against this change:
The current behaviour has been the undocumented standard for at least 5 years.
Do we really want to mess with users' code/test setup, just to make the behaviour normal/intuitive?

An alternative solution could be to just improve the documentation, and prescribe the 'expected' behaviour of the class-decorator.

-------
As an aside, as part of this PR I've also introduced a new CI workflow to validate this behaviour against older dependency versions that are known to have caused issues in the past. This used to be part of the main workflow, but it's more sensible (and is easier to extend) when it has it's own workflow IMO.


spulec (approved):

bblommers:
Closes #416

PR #4437, opened 2021-10-19
Route53 query-logging-config APIs

kbalk:
Addition of the Route53 APIs:

- `create_query_logging_config`
- `get_query_logging_config`
- `list_query_logging_config`
- `delete_query_logging_config`

Note:  `create_query_logging_config()` does not create a log stream for an edge location.  That's supposed to happen the first time that the edge location responds to DNS queries for the specified hosted zone.

### DIFF
--- a/moto/route53/responses.py
+++ b/moto/route53/responses.py
@@ -241,22 +241,13 @@ class Route53(BaseResponse):
             )
 
         elif request.method == "GET":
-            parsed_url = urlparse(full_url)
-            query_params = parse_qs(parsed_url.query)
-
-            # For some reason the params are returned as a list.
-            param_list = query_params.get("hostedzoneid")
-            hosted_zone_id = param_list[0] if param_list else None
-
-            param_list = query_params.get("nexttoken")
-            next_token = param_list[0] if param_list else None
-
-            param_list = query_params.get("maxresults")
-            max_results = int(param_list[0]) if param_list else None
-
+            hosted_zone_id = self._get_param("hostedzoneid")
+            next_token = self._get_param("nexttoken")
+            max_results = self._get_int_param("maxresults")
             try:
                 try:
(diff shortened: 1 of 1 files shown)


bblommers:
> 
> I made the changes you suggested, but I'm still missing something. 

No.. I missed something.
Moto deals with exceptions automatically _if_ the service redirects all requests to the `dispatch`-method (a catch-all method) in `urls.py`, but Route53 calls it's methods directly. 
That means Route53 has to parse the incoming request (hence all the calls to `self.setup_class(..)`), and also means it has to deal with the response formatting directly.

Try this:
```
try:
    query_logging_config = route53_backend.create_query_logging_config(
        self.region, hosted_zone_id, log_group_arn
    )
except Route53ClientError as r53error:
    return r53error.code, {}, r53error.description
```

Not as nice as I promised, but still better than having to deal with the XML response manually

kbalk:
@bblommers I like the pagination decorator and wished I had it for the stuff I worked on before!  I do have a question, though:  if I want my own invalid token exception other than the InvalidToken exception produced by the paginator, do I have to trap that exception in `responses.py` when I invoke the list function and raise my custom exception, or is there an alternate way?

kbalk:
> > @bblommers I like the pagination decorator and wished I had it for the stuff I worked on before! I do have a question, though: if I want my own invalid token exception other than the InvalidToken exception produced by the paginator, do I have to trap that exception in `responses.py` when I invoke the list function and raise my custom exception, or is there an alternate way?
> 
> Hmm.. good question. The best way is to probably set `"fail_on_invalid_token": False` in the configuration, and then throw your custom exception. I don't remember too much about the details I'm afraid, as the actual development was quite a while ago

Setting "fail_on_invalid_token" just returns "None".  I'll try trapping and re-raising the error.

PR #4444, opened 2021-10-19
Fix : SNS get attributes on FIFO topic

KrishnanRanjithkumar:
Fix : SNS get attributes on FIFO topic

localstack/localstack#4712

Added test cases for both FIFO topic and normal topic


KrishnanRanjithkumar:
Hi @bblommers , thanks for the comment. No problem in adding the requested test case, will do it and push the changes soon

PR #4481, opened 2021-10-26
Fix security group filters 2

jl2501:
Fixed the regex for AWS API globbing to match all unescaped stars and question marks as described [here](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/Using_Filtering.html) by using a negative lookbehind lookaround (thank you https://www.regular-expressions.info/lookaround.html).

It seems that actually none of the filtering operations that I tested against the AWS API do globbing unless you specifically pass a wildcard. I recall it being different some years ago, but my memory may be incorrect, and this seems to match the current state of the AWS API semantics regardless of what it may / may not used to have been.

This should fix the issue from my previous PR where filtering by things like description leads to having unexpected extra instances returned. It will also fix several other issues stemming from the previous PR assuming that the globbing was automatically added in cases where testing the AWS API today shows that it does not do that.


tests/test_ec2/test_security_groups.py
```diff
@@ -2332,3 +2379,39 @@ def test_get_groups_by_ippermissions_group_id_filter_across_vpcs():
...
+    """
+    this filter is an exact match, not a glob
+    """
+    ec2r = boto3.resource("ec2", region_name="us-west-1")
+    vpc = ec2r.create_vpc(CidrBlock="10.250.1.0/16")
+
+    sg1 = vpc.create_security_group(
+        Description="A Described Description Descriptor", GroupName="test-1"
+    )
+    sg2 = vpc.create_security_group(
```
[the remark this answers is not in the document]
jl2501: done. thx!

tests/test_ec2/test_security_groups.py
```diff
@@ -2332,3 +2379,39 @@ def test_get_groups_by_ippermissions_group_id_filter_across_vpcs():
...
+@mock_ec2
+def test_filter_group_name():
+    """
+    this filter is an exact match, not a glob
+    """
+    ec2r = boto3.resource("ec2", region_name="us-west-1")
+    vpc = ec2r.create_vpc(CidrBlock="10.250.1.0/16")
+
+    sg1 = vpc.create_security_group(
+        Description="A Described Description Descriptor", GroupName="test-1"
```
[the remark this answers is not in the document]
jl2501: added unique prefix as suggested/ described. thx for that!

jl2501:
not sure what I can do to get the docs/readthedocs.org:moto test to pass as I don't think I touched that __init__.py file and it seems to possibly be complaining about the f-string syntax b/c it looks to be maybe running on an older python version for that test.

I don't think that's something I caused, but I'm happy to fix it if there's something I can do there

tests/test_ec2/test_security_groups.py
```diff
@@ -2332,3 +2379,40 @@ def test_get_groups_by_ippermissions_group_id_filter_across_vpcs():
...
+
+    filter_to_match_group_1 = {
+        "Name": "group-name",
+        "Values": [sg1.group_name],
+    }
+    security_groups = ec2r.security_groups.filter(Filters=[filter_to_match_group_1])
+
+    security_groups = list(security_groups)
+    assert len(security_groups) == 1
+    assert security_groups[0].group_name == "test-1"
```
bblommers, line 2418:
```suggestion
    assert security_groups[0].group_name == sg1.group_name
```

bblommers (changes requested):

tests/test_ec2/test_security_groups.py
jl2501: derp.

jl2501: appreciate that @bblommers! apologies for missing that there.

bblommers:
> not sure what I can do to get the docs/readthedocs.org:moto test to pass as I don't think I touched that **init**.py file and it seems to possibly be complaining about the f-string syntax b/c it looks to be maybe running on an older python version for that test.
> 
> I don't think that's something I caused, but I'm happy to fix it if there's something I can do there

That's all good - the RTD build was fixed in master a while back, so it will pass for this PR once we merge :+1:

bblommers (approved):
LGTM! Thanks again for the PR @jl2501

PR #4500, opened 2021-10-29
Add verbose option to pull_down_aws_managed_rules.py

kbalk:
The `pull_down_aws_managed_rules.py` script takes a while to run because it's pulling down quite a few markdown files from the AWS website.  I added a "verbose" switch `-v` or `--verbose` so the invoker can have a sense of the script's progress and not worry that the script has stalled somewhere along the line.

Also updated some comments and cleaned up a pylint error in the script.


bblommers (approved):
👍

Issue #4499, opened 2021-10-29
[ECR] mocked describe_images returns timezone unaware datetime for imagePushedAt field when AWS return timezone aware datetime


dotboris:
There's a behaviour difference between moto and the real AWS when using `describe_images` in ECR. 

When moto is active (mocking with `@mock_ecr` decorator) the `imagePushedAt` field contains a `datetime` object with no `tzinfo` configured.

When calling AWS directly without moto, the `imagePushedAt` field contains a `datetime` object that has a `tzinfo` field set.

Here's a code snippet that shows the issue:

```python
import boto3
from moto import mock_ecr


def image_from_boto():
    ecr = boto3.client("ecr", region_name="us-east-1")
    return ecr.describe_images(repositoryName="coveo/python")["imageDetails"][0]


@mock_ecr
def image_from_moto():
    ecr = boto3.client("ecr", region_name="us-east-1")

    ecr.create_repository(repositoryName="coveo/python")
    ecr.put_image(repositoryName="coveo/python", imageManifest="{}")

    return ecr.describe_images(repositoryName="coveo/python")["imageDetails"][0]


print("boto3:")
image = image_from_boto()
print(image["imagePushedAt"])
print(image["imagePushedAt"].tzinfo)

print("moto3:")
image = image_from_moto()
print(image["imagePushedAt"])
print(image["imagePushedAt"].tzinfo)

```

The output:

```sh
$ python example.py
boto3:
2017-05-10 17:28:56-04:00
tzlocal()
moto3:
2021-10-29 16:04:33.118095
None
```

The real repository and image referenced here is specific to my AWS account. If you want to run this yourself, you'll need to specify another image.

I expect both my `boto3` and `moto` images to have a `imagePushedAt` field with a timezone in it. Currently this is not the case as the `moto` one does not have a timezone.

Here are the versions that I'm running:

```
boto3              1.19.6
botocore           1.22.6
moto               2.2.11
```

<details>
<summary>Full list of dependencies</summary>

```sh
$ poetry show --no-dev
boto3              1.19.6    The AWS SDK for Python
botocore           1.22.6    Low-level, data-driven core of boto 3.
certifi            2021.10.8 Python package for providing Mozilla's CA Bundle.
cffi               1.15.0    Foreign Function Interface for Python calling C code.
charset-normalizer 2.0.7     The Real First Universal Charset Detector. Open, modern and actively mai...
cryptography       35.0.0    cryptography is a package which provides cryptographic recipes and primi...
[...]

PR #4501, opened 2021-10-29
Ensure that imagePushedAt fields from ECR is timezone aware

dotboris:
Fix: #4499 

This PR ensures that the `imagePushedAt` field returned by `describe_images` has a timezone set. This matches the behaviour or `boto3` when not mocked.


bblommers (approved):
Welcome to Moto, @dotboris! LGTM - thanks for raising this and fixing it

PR #4502, opened 2021-10-29
Fix range headers: use response_headers parameters instead of creating a new dict

### LINKED ISSUE #4203: S3 .get(Range=...) doesn't return ETag
# Summary

When specifying a `Range` parameter for an S3 `.get()`, the response doesn't include metadata like `ETag`.
(It works without the `Range` parameter.)

# Steps to reproduce

```
import boto3
from moto import mock_s3


with mock_s3():

    bucket = 'fake-bucket'
    key = 'my-file'

    boto3.client('s3').create_bucket(
        Bucket=bucket,
        CreateBucketConfiguration={
            'LocationConstraint': 'ap-southeast-2'
        },
    )

    obj = boto3.resource('s3').Object(bucket, key)

    obj.put(Body=b"1234")

    resp = obj.get()
    assert resp['ETag'] == obj.e_tag

    resp = obj.get(Range="bytes=0-1")
    assert resp['ETag'] == obj.e_tag

```

# Expected behavior

Response includes the following top-level keys:

* ResponseMetadata
* AcceptRanges
* LastModified
* ContentLength
* ETag
* ContentRange
* ContentType
* ServerSideEncryption
* Metadata
* Body

# actual behavior

Response includes the following top level keys:

* ResponseMetadata
* ContentLength
* ContentRange
* Metadata
* Body

So the final line in the script throws a KeyError.

```
  File "mwe.py", line 26, in <module>
    assert resp['ETag'] == obj.e_tag
KeyError: 'ETag'
```

# Installation

I `pip install`ed moto.

Python 3.8.10

boto                   2.49.0              
boto3                  1.18.26             
botocore               1.21.26             
moto                   2.2.3

vbarbaresi:
Fixes #4203

`headers` parameter was unused in `_handle_range_header()` method, causing some headers to be lost

This looked like a typo as the method is invoked using `response_headers` as argument name:
https://github.com/spulec/moto/blob/5771dcf73bef22adb2e7c54ee4787ea6eebceebd/moto/s3/responses.py#L1063-L1065
I simply changed the argument back to `response_headers` and it fixed the issue.

### DIFF
--- a/moto/s3/responses.py
+++ b/moto/s3/responses.py
@@ -984,8 +984,7 @@ class ResponseObject(_TemplateEnvironmentMixin, ActionAuthenticatorMixin):
             template.render(deleted=deleted_objects, delete_errors=error_names),
         )
 
-    def _handle_range_header(self, request, headers, response_content):
-        response_headers = {}
+    def _handle_range_header(self, request, response_headers, response_content):
         length = len(response_content)
         last = length - 1
         _, rspec = request.headers.get("range").split("=")
--- a/tests/test_s3/test_s3.py
+++ b/tests/test_s3/test_s3.py
@@ -6486,3 +6486,25 @@ def test_head_object_should_return_default_content_type():
     s3.Object("testbucket", "testobject").content_type.should.equal(
         "binary/octet-stream"
     )
+
+@mock_s3
+def test_request_partial_content_should_contain_all_metadata():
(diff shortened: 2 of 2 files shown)


bblommers (approved):
Nice catch @vbarbaresi - thanks for the PR!

PR #4504, opened 2021-10-30
EMR: Rename get_cluster to describe_cluster

### LINKED ISSUE #4503: EMR: Add describe_cluster to implementation_coverage.md
I believe `emr.describe_cluster` is already [implemented](https://github.com/spulec/moto/blob/master/moto/emr/responses.py#L129) but currently does not show up under [implementation_coverage.md](https://github.com/spulec/moto/blob/master/IMPLEMENTATION_COVERAGE.md?plain=1#L1806) because the underlying method on the backend is called `get_cluster`.

Can create a quick PR with the rename since most of the changes are already in place.

anaynayak:
Fixes #4503

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,3 +25,4 @@ tests/file.tmp
 .venv/
 htmlcov/
 .~c9_*
+.coverage*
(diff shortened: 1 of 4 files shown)


bblommers (approved):
Nice one - thanks @anaynayak!

PR #4507, opened 2021-10-30
DirectoryService: Add support for other directories

kbalk:
Adds support for the APIs:

- `create_microsoft_ad()`
- `connect_directory()`
- `create_alias()`
- `enable_sso()`
- `disable_sso()`

### DIFF
--- a/IMPLEMENTATION_COVERAGE.md
+++ b/IMPLEMENTATION_COVERAGE.md
@@ -916,19 +916,19 @@
 
 ## ds
 <details>
-<summary>10% implemented</summary>
+<summary>19% implemented</summary>
 - [ ] accept_shared_directory
 - [ ] add_ip_routes
 - [ ] add_region
 - [X] add_tags_to_resource
 - [ ] cancel_schema_extension
-- [ ] connect_directory
-- [ ] create_alias
+- [X] connect_directory
+- [X] create_alias
 - [ ] create_computer
 - [ ] create_conditional_forwarder
 - [X] create_directory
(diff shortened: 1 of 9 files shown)


kbalk:
@bblommers I have a lot of validation steps in this code.  The steps and logic is not organized that well and the logic takes up a lot of space in `models.py`.  I could make the validation functions individual validations using static methods in the Directory class.   Or I could create a bunch of separate functions and move them to a new file, say `validation.py` or `utils.py`.  Or create a separate Validations class in a separate file?  Do you have a preference or an alternative suggestion?  

It's an issue I've come up against before, so it'd be nice to have a better approach, otherwise `models.py` can get pretty large.

bblommers:
@kbalk No preference here! Having the validation code separated out from `models.py` would indeed by good, but I trust your judgement in how that's done exactly.

kbalk:
I've got a Terraform script that creates an AD Connector and the script fails with my moto code.  I'm debugging the problem and it's taking me some time.  The other thing I have left to do is reorganize the validation.  So it may be a day or two before this code is ready for review.

kbalk:
@bblommers The rearrangement of the validation handling is not ideal, but it feels a little less messy.  

I discovered I was not setting the DnsIpAddrs field, so that was causing the terraform script to fail.  My calculation of those IPs is not great, but it was the best I could do with the knowledge I have.  If someone has a better solution, I hope they change the code.

bblommers (approved):
LGTM! Thanks for the PR @kbalk

PR #4510, opened 2021-10-31
feat: add describe_receipt_rule_set to ses mock

### LINKED ISSUE #4509: `describe_receipt_rule_set` for `ses` not implemented
It looks like `describe_receipt_rule_set` is not implemented [here](https://github.com/spulec/moto/blob/master/IMPLEMENTATION_COVERAGE.md#ses), I'll make a PR to add this feature shortly as my team also may need to use this function.

shermanhui:
This PR aims to resolve #4509

### DIFF
--- a/moto/ses/responses.py
+++ b/moto/ses/responses.py
@@ -239,6 +239,23 @@ class EmailResponse(BaseResponse):
         template = self.response_template(CREATE_RECEIPT_RULE)
         return template.render()
 
+    def describe_receipt_rule_set(self):
+        rule_set_name = self._get_param("RuleSetName")
+
+        rule_set = ses_backend.describe_receipt_rule_set(rule_set_name)
+
+        for i, rule in enumerate(rule_set):
+            formatted_rule = {}
+
+            for k, v in rule.items():
+                self._parse_param(k, v, formatted_rule)
+
+            rule_set[i] = formatted_rule
+
+        template = self.response_template(DESCRIBE_RECEIPT_RULE_SET)
+
+        return template.render(rule_set=rule_set, rule_set_name=rule_set_name)
+
     def describe_receipt_rule(self):
         rule_set_name = self._get_param("RuleSetName")
         rule_name = self._get_param("RuleName")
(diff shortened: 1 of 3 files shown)


moto/ses/responses.py
```diff
@@ -507,6 +524,57 @@ def describe_receipt_rule(self):
   </ResponseMetadata>
 </CreateReceiptRuleResponse>"""
 
+DESCRIBE_RECEIPT_RULE_SET = """<DescribeReceiptRuleSetResponse xmlns="http://ses.amazonaws.com/doc/2010-12-01/">
+  <DescribeReceiptRuleSetResult>
```
shermanhui, line 528: I got the shape of the XML response by inferring it from [here](https://github.com/boto/botocore/blob/develop/botocore/data/ses/2010-12-01/service-2.json#L2032) and the [SES docs](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/ses.html#SES.Client.describe_receipt_rule_set)

@bblommers I recall there being a suggestion to set up a MITM proxy to see the XML response directly, unfortunately, I don't have experience with this and I'm not sure where to start - could you point me in the right direction? That being said, I'm pretty confident that this produces the expected response.

moto/ses/responses.py
```diff
@@ -507,6 +524,57 @@ def describe_receipt_rule(self):
...
+        </Actions>
+        <TlsPolicy>{{rule["tls_policy"]}}</TlsPolicy>
+        <ScanEnabled>{{rule["scan_enabled"]}}</ScanEnabled>
+        <Enabled>{{rule["enabled"]}}</Enabled>
+      </member>
+      {% endfor %}
+    </Rules>
+    <Metadata>
+      <Name>{{rule_set_name}}</Name>
+      <CreatedTimestamp>2021-10-31</CreatedTimestamp>
```
shermanhui, line 569: I'm not sure how to populate this value here.

We'd probably have to update the `[create_receipt_rule_set](https://github.com/spulec/moto/blob/1d21cec0004afae584e9c429238d402b888ae78c/moto/ses/models.py#L415)` method to save the datetime when the ruleset is created, but I'm not sure where we'd add the created datetime since the current data structure is just a dict with the rule set name as the key to a list of rules. i.e.

```
{
   "testRuleSet":[
      {
         "name":"testRule",
         "enabled":"true",
         ...,
      }
   ]
}
```

Is there any preference in how we implement this?

moto/ses/responses.py
bblommers: Well, the tests pass, so boto3 seems to think it's correct - that's the most important part!

For proxying, I've always used [MITMProxy](https://mitmproxy.org/). 
On Linux (/Ubuntu) the installation/setup is as easy as:
```
sudo apt install mitmproxy
mitmweb
```
YMMV of course - it will depend on what OS you're using. 
MITMWeb will start a browser window where you can easily see the requests that have gone through the proxy.

After that, the easiest way of proxying requests to AWS is probably via the CLI.
```
export HTTP_PROXY=http://localhost:8080
export HTTPS_PROXY=http://localhost:8080
aws ses describe-rule-set --no-verify-ssl  # Do not verify SSL requests, as the proxy does not have SSL certificates set-up
```
It is possible to use a proxy using boto3 as well - as long as both the proxy is configured, and boto3 is configured to not verify SSL certificates.

Note that this only works for requests to AWS itself - Moto will intercept requests before they are send to the proxy.

moto/ses/responses.py
bblommers: Ideally the rule sets are stored as they're own class (class FakeRuleSet(BaseModel)), to be in-line with other data stores.

If this PR is ready, I'll merge it for now then - having proper metadata can be part of a next PR 👍

bblommers (approved):
LGTM! Thanks @shermanhui

PR #4512, opened 2021-10-31
Feature: Custom cloudformation resources

bblommers:
#### Background reading:
https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/template-custom-resources.html

Custom resources are typically resources that are created in an external system.
The application developer provides a Lambda that is responsible for creating/updating/deleting the external resource - CloudFormation is responsible for calling the Lambda with the appropriate details.

AWS CF works in asyncronously. All resource creation is kicked off in the background, and each resource reports back to CF when the resource has been created successfully (or not).
A custom resource works the same way. AWS provides a specific module, called `cfnresponse`, that is responsible for calling CF to let the parent stack know whether the creation of the custom resource was successful.

------------

The Moto implementation follows the same design:
 - App developer provides the lambda responsible for creating the custom resource
 - Moto CloudFormation creates all resources
 - Moto CloudFormation calls the custom lambda
   A Moto-specific callback-URL is provided to indicate resource creation success/failure
 - The custom Lambda responsible for reading this callback-URL, and providing it to the `cfnresponse` module
 - The `cfnresponse`-module makes a HTTP-request to the callback URL
 - Moto CloudFormation receives this URL, and marks custom resource creation as a success/failure

-----------------------------

#### Important implementation details:
 - As the cfnresponse-module needs to call a URL to indicate success/failure, this feature only works in MotoServer.
 - Moto CF-resources can now be created asyncronously. This is necessary for custom resources, as we need to wait until the callback-request comes in, before we can decide whether creation was successful.

I'll add comments to this PR to explain why some of the other changes were necessary


bblommers:
@kbalk Are you able to verify whether this branch works for your usecase? 

The example TF-stack worked for me, after making one change:
The ec2-client in `lambda.py` needed to explicitly connect to the MotoServer, as mocking does not work inside a separate Docker container:
```
log = logging.getLogger(__name__)

ec2 = boto3.client("ec2", endpoint_url="http://host.docker.internal:5000")

def build_search_body(event):
    ....
```

moto/apigateway/models.py
```diff
@@ -72,7 +72,7 @@ def cloudformation_type():
 
     @classmethod
     def create_from_cloudformation_json(
-        cls, resource_name, cloudformation_json, region_name
+        cls, resource_name, cloudformation_json, region_name, **kwargs
```
bblommers, line 75: Custom resources also need to know the LogicalResourceId/StackId/ResourceType. Passing in kwargs was the least invasive signature change that leaves the door open for further extensions

moto/awslambda/models.py
```diff
@@ -735,7 +741,12 @@ def update_from_cloudformation_json(
     def _create_zipfile_from_plaintext_code(code):
         zip_output = io.BytesIO()
         zip_file = zipfile.ZipFile(zip_output, "w", zipfile.ZIP_DEFLATED)
-        zip_file.writestr("lambda_function.zip", code)
+        zip_file.writestr("index.py", code)
```
bblommers, line 744: Existing bug, exposed in `test_cloudformation_stack_integration.py::test_lambda_function`.
The test created a Lambda-function, and provided the code as part of the CF template. Execution of this function was not possible before this change though, as Lambda looks for source code in a file called `index.py` by default.

moto/awslambda/models.py
```diff
@@ -735,7 +741,12 @@ def update_from_cloudformation_json(
...
+        # This should really be part of the 'lambci' docker image
+        from moto.packages.cfnresponse import cfnresponse
```
bblommers, line 746: Same scenario: this comes into play when creating a Lambda function via CF with the code provided inline in the CF template.

From https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/cfn-lambda-function-code-cfnresponsemodule.html:

>
> The cfn-response module is available only when you use the ZipFile property to write your source code. It isn't available for source code that's stored in Amazon S3 buckets. For code in buckets, you must write your own functions to send responses. 
>

moto/cloudformation/custom_model.py
```diff
@@ -0,0 +1,112 @@
...
+        def invoke_function():
+            try:
+                fn.invoke(json.dumps(event), {}, {})
+            except Exception as e:
+                print(e)
+
+        print("Invoking function")
+        # p = multiprocessing.Process(target=invoke_function)
+        # p.start()
+        invoke_thread = threading.Thread(target=invoke_function, args=())
```
bblommers, line 86: The custom Lambda needs to be invoked async, as we're still in the `create_stack`-call, and we don't want to wait 10+ seconds for Docker to be invoked

moto/cloudformation/models.py
```diff
@@ -254,9 +258,7 @@ def _create_resource_map(self):
         return resource_map
 
     def _create_output_map(self):
-        output_map = OutputMap(self.resource_map, self.template_dict, self.stack_id)
-        output_map.create()
```
bblommers, line 258: The full workflow looked like this:
  - Create FakeStack:
      - Initialize resource map
      - Create output map
 - Persist a reference to the FakeStack
 - create resources as defined in resource map

However:
When the OutputMap contained a reference to a resource (using `fn:getatt`), creating the OutputMap would also create the resource - as the resource needs to be created before we can retrieve the attribute.
This could lead to the scenario where a resource is created, but creation of the OutputMap ultimately fails because we're trying to `fn:getatt` an unknown attribute.. As the OutputMap was created in the FakeStack-constructor, the creation of the FakeStack also fails, and the resource itself is dangling.

moto/cloudformation/parsing.py
```diff
@@ -581,11 +591,27 @@ def load_conditions(self):
         for condition_name in self.lazy_condition_map:
             self.lazy_condition_map[condition_name]
 
+    def validate_outputs(self):
```
bblommers, line 594: Separate method to validate whether the OutputMap is valid, i.e. whether all attributes in `fn:;getatt` are known/valid.

This is the reason why all resource models now have a new static method called `has_cfn_attr` - we don't want to create the resource before finding out that the attribute is not valid.

moto/cloudformation/models.py
```diff
@@ -325,11 +327,27 @@ def stack_outputs(self):
...
 
+    def add_custom_resource(self, custom_resource):
+        self.custom_resources[custom_resource.logical_id] = custom_resource
+
+    def get_custom_resource(self, custom_resource):
+        return self.custom_resources[custom_resource]
+
     def create_resources(self):
-        self.resource_map.create(self.template_dict)
+        self.status = "CREATE_IN_PROGRESS"
```
bblommers, line 337: This makes the Stack-creation async (-ish). It's marked as create-in-progress - unless all of the resources are created immediately, then we can set the status as 'create-complete'

moto/cloudformation/responses.py
```diff
@@ -87,6 +93,19 @@ def _get_param_values(self, parameters_list, existing_params):
                 raise MissingParameterError(parameter["parameter_key"])
         return result
 
+    def process_cfn_response(self):
```
bblommers, line 96: This method ultimately deals with the HTTP-request made by the `cfnresponse`-module.

 - Find the stack
 - Find the custom resource that called this URL
 - Persist any outputs that are returned by the custom resource
 - Update the stack-status to mark it as `create_complete`, if applicable

moto/cloudwatch/models.py
```diff
@@ -677,33 +676,6 @@ def _get_paginated(self, metrics):
             return None, metrics
 
 
-class LogGroup(CloudFormationModel):
```
bblommers, line 680: Unrelated bug that only popped up after adding the 'has_cfn_attribute`-methods.
This class would be invoked by CF, but it returns another class (moto/logs/LogGroup) that did not extend CloudFormationModel - so responsibility for the creation was split between the two.

I've reworked this so that the `moto/logs/LogGroup`-class is completely responsible for CF creation.

moto/packages/cfnresponse/cfnresponse.py
```diff
@@ -0,0 +1,124 @@
+"""AWS cfnresponse Lambda script."""
+
+# Modified version of the official AWS cfnresponse-module
```
bblommers, line 3: TODO: We can probably use the official AWS version, without modifications?
https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/cfn-lambda-function-code-cfnresponsemodule.html

tests/test_cloudformation/test_stack_parsing.py
```diff
@@ -312,10 +314,17 @@ def test_parse_stack_with_get_availability_zones():
     output.value.should.equal(["us-east-1a", "us-east-1b", "us-east-1c", "us-east-1d"])
 
 
-def test_parse_stack_with_bad_get_attribute_outputs():
-    FakeStack.when.called_with(
-        "test_id", "test_stack", bad_output_template_json, {}, "us-west-1"
-    ).should.throw(ValidationError)
+@mock_sqs
+@mock_cloudformation
+def test_parse_stack_with_bad_get_attribute_outputs_using_boto3():
```
bblommers, line 319: Test verified against AWS.

tests/test_cloudformation/test_cloudformation_custom_resources.py
```diff
@@ -0,0 +1,121 @@
...
+        resource_id=f"CustomResource{str(uuid4())[0:6]}"
+    )
+    return pfunc
+
+
+@mock_cloudformation
+@mock_lambda
+@mock_logs
+@mock_s3
+def test_create_custom_lambda_resource():
```
bblommers, line 43: This test runs against AWS - I've just reduced the wait times, as AWS sometimes takes 1 minute to even get started with stack creation

tests/test_cloudformation/test_cloudformation_custom_resources.py
```diff
@@ -0,0 +1,121 @@
...
+def lambda_handler(event, context):
+    print(event)
+    response = dict()
+    response["Status"] = "SUCCESS"
+    response["StackId"] = event["StackId"]
+    response["RequestId"] = event["RequestId"]
+    response["LogicalResourceId"] = event["LogicalResourceId"]
+    response["PhysicalResourceId"] = "{resource_id}"
+    response_data = dict()
+    response_data["info_value"] = "special value"
```
bblommers, line 25: This is the meat of what we're actually testing here.
Can this value be persisted via the `cfnresponse`-module, and picked up via boto3's `describe_stacks`-request?

(Assume another call to a third-party was made to create a resource, and this `special value` was returned by the third-party.)

moto/cloudformation/custom_model.py
```diff
@@ -0,0 +1,112 @@
...
+
+    @staticmethod
+    def cloudformation_type():
+        return "?"
+
+    @classmethod
+    def create_from_cloudformation_json(
+        cls, resource_name, cloudformation_json, region_name, **kwargs
+    ):
+        print("create_from_cloudformation")
```
kbalk, line 38: I noticed some print statements when running my test -- I don't know if they are intentional or whether you're still doing more work and would be cleaning up later:

```
create_from_cloudformation
{'RequestType': 'Create', 'ServiceToken': 'arn:aws:lambda:us-east-1:123456789012:function:test-cfn-lookup-ami-ids-0b', 
'ResponseURL': 'http://host.docker.internal:4615/cloudformation_us-east-1/cfnresponse?stack=arn:aws:cloudformation:us-east-1:123456789:stack/ami-lookup-failure-test/e559eb5e-f43d-4477-8791-72315d54722f', 
'StackId': 'arn:aws:cloudformation:us-east-1:123456789:stack/ami-lookup-failure-test/e559eb5e-f43d-4477-8791-72315d54722f', 
'RequestId': '8d30ba47-f16c-4e79-adef-0e43480b3d4e', 'LogicalResourceId': 'AmiIdLookup', 
'ResourceType': 'Custom::AmiIdLookup', '
ResourceProperties': {'Filters': '[{"Name":"name","Values":["amzn-ami-hvm-*"]}]', 
'Owners': '["foo"]', 'Region': 'us-east-1', 
'ServiceToken': 'arn:aws:lambda:us-east-1:123456789012:function:test-cfn-lookup-ami-ids-0b'}}
Invoking function
started
127.0.0.1 - - [31/Oct/2021 11:08:12] "POST / HTTP/1.1" 200 -
get_cfn_attribute(Id)
{}
```

kbalk:
> @kbalk Are you able to verify whether this branch works for your usecase?
> 
> The example TF-stack worked for me, after making one change: The ec2-client in `lambda.py` needed to explicitly connect to the MotoServer, as mocking does not work inside a separate Docker container:
> 
> ```
> log = logging.getLogger(__name__)
> 
> ec2 = boto3.client("ec2", endpoint_url="http://host.docker.internal:5000")
> 
> def build_search_body(event):
>     ....
> ```

@bblommers I can confirm that your changes will work with the test I provided you.  I am so glad you took on this work, as this was a tricky, complicated problem!

I also ran our official test setup (which I pared down to derive the test I gave you) against your changes and I get the error shown below.  In our official test scenario, we run localstack and use `initaws.d` to kick off moto.  I use parameters to control which services use localstack or moto ports.  For this test, I set it to use moto ports for all services.

```
Error: error waiting for CloudFormation Stack creation: timeout while waiting for state to become 'CREATE_COMPLETE, CREATE_FAILED, DELETE_COMPLETE, DELETE_FAILED, ROLLBACK_COMPLETE, ROLLBACK_FAILED' (last state: 'CREATE_IN_PROGRESS', timeout: 1m0s)

  with module.failure.aws_cloudformation_stack.cfn,
  on ../modules/cfn/main.tf line 1, in resource "aws_cloudformation_stack" "cfn":
   1: resource "aws_cloudformation_stack" "cfn" {
```

bblommers:
> 
> ```
> Error: error waiting for CloudFormation Stack creation: timeout while waiting for state to become 
> 'CREATE_COMPLETE, CREATE_FAILED, DELETE_COMPLETE, DELETE_FAILED, 
> ROLLBACK_COMPLETE, ROLLBACK_FAILED' (last state: 'CREATE_IN_PROGRESS', timeout: 1m0s)
> 

That sounds like the CustomResource was not able to complete, or at least not able to notify CloudFormation that it was complete.

Can you see anything relevant in the Cloudwatch Logs, on whether there are connection problems with the cfnmodule? (http://localhost:5000/moto-api -> logs)

moto/cloudformation/custom_model.py
bblommers: Yeah, it needs a clean-up, and more tests. Just wanted to put it out here for initial testing!

kbalk:
> That sounds like the CustomResource was not able to complete, or at least not able to notify CloudFormation that it was complete.
> 
> Can you see anything relevant in the Cloudwatch Logs, on whether there are connection problems with the cfnmodule? (http://localhost:5000/moto-api -> logs)

Oh ... I'm so sorry ... we use a different moto port when running our automated tests, but I edited `lambda.py` in that test to use the default port of 5000.  When I changed the port to match what was running, the tests succeeded.  So it's looking very good!

bblommers:
> 
>  So it's looking very good!

Awesome - that's what I like to hear. I'll need to spend time on this PR to clean up, so it might take a few more days before it will be merged, but I'm happy to hear it works.

> 
> The logs showed an endpoint error, which is how I realized my error ... I don't use that website to check output, but it looks like I should.

I don't use it much either, but it came in very handy during development this time!

PR #4517, opened 2021-11-01
More improvements for sagemaker.

bogdangi:
We are mocking sagemaker we corrected some behaviour


bblommers (approved):
LGTM! Thanks for the PR @bogdangi

PR #4518, opened 2021-11-02
CognitoIdp fixes

mianowski:
* TokenType added to "AuthenticationResult"
* Username corrected in ChalengeParameters


bblommers (changes requested):
Hi @mianowski, welcome to Moto!
Would you mind adding some assertions for this? Probably easiest to just add to [one of the existing tests here](https://github.com/spulec/moto/blob/master/tests/test_cognitoidp/test_cognitoidp.py), considering they're quite small changes.

Looks like the linting is failing on our CI as well - you can run `make lint` to verify this locally.
If `black` has an issue with the code, make sure that you have the right version (`black==19.10b0`), as different versions will have different formatting rules.

bblommers (changes requested):
Thanks for making the changes @mianowski. Happy to merge if we can get the CI to pass

PR #4522, opened 2021-11-02
[BUGFIX] Can't perform GET/PUT operations against server from inside docker container 

eschkufz:
Whitelists `host.docker.internal` in `subdomain_based_buckets`. Fixes issue 4521.

Signed-off-by: eric schkufza <eric.schkufza@gmail.com>


bblommers (approved):
Good catch @eschkufz - thanks for the fix!

PR #4525, opened 2021-11-04
Add cognito-identity list identities method

acapagarcia:
Add cognito identity pool list_identities method and integrate it with get_id


bblommers (approved):
LGTM. Thanks for the PR @acapagarcia, and welcome to Moto!

PR #4497, opened 2021-10-29
 S3: Implement CORS headers in OPTIONS requests 

vbarbaresi:
Hacktoberfest contribution 🎃 

Fixes #4220 

The default headers from CORS plugin was not enough to mock S3 CORS Rules behavior
We need to manually define response headers depending on the bucket rules

- Fixed typos:
  -  `ExposedHeader` --> `ExposeHeader` (defined as such in AWS documentation)
  - `MaxAgeSecond` --> `MaxAgeSeconds`(defined as such in AWS documentation)
- I implemented OPTIONS method instead of relying on the defaults from Flask

- Iterated through CORS rule to return some CORS headers
This is not a perfectly accurate strategy: I didn't check but according to the docs, when multiple CORS Rules are defined, S3 returns the most relevant (I suppose the most permissive one, depending on the request)
In the strategy I implemented, the last rule takes precedence over the other rules.
It's not ideal but I consider that defining multiple rules is an edge case

- Removed a test about `Allow` headers: these headers were automatically set by flask-cors but they are not technically CORS?
I didn't add these headers back because they are not required and from my testing on a real S3 bucket, I didn't receive any `Allow` header, only `Access-Control-Allow-Methods`

PR #4528, opened 2021-11-04
add back OPTIONS key response after implementing CORS in #4497

vbarbaresi:
In #4497 I forgot to implement OPTIONS on key response, I only added unit tests for bucket responses (queries on `/`)

The OPTIONS key response on a real S3 endpoint is essentially the same as the bucket response.
If the bucket doesn't exist, return 403. If the bucket exists, return 200 with the bucket CORS headers, disregarding if the key exists or not.
So I used the same method for both key and bucket responses.


bblommers (approved):
LGTM! Thanks for fixing this @vbarbaresi :+1:

PR #4541, opened 2021-11-08
Directory Service: Handle IPV6 addresses and no tags

kbalk:
Issues fixed:

- `ipaddress.ip_network()` should be used instead of `ipaddress.IPv4Network()` in order to handle both IPv4 and IPv6 addresses.
- 
- There are checks for the maximum number of tags, but those checks don't account for no tags being present.  Set the default value for tags to an empty list.

### DIFF
--- a/moto/ds/models.py
+++ b/moto/ds/models.py
@@ -107,7 +107,7 @@ class Directory(BaseModel):  # pylint: disable=too-many-instance-attributes
         """
         ip_addrs = []
         for subnet in subnets:
-            ips = ipaddress.IPv4Network(subnet.cidr_block)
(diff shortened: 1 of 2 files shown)


kbalk:
My initial PR concerned DNS IP addresses.  Since we use Localstack and moto, we are dependent on the version of moto that Localstack pulls in.  The Directory Service PR that handles DNS IP addresses was committed the same day as the Localstack release, but after the release.  So the code wasn't there to handle the DNS IP addresses.  It's still dodgey logic and I hope someone can come up with cleaner code, but it's not in play yet with the latest release of LocalStack.

bblommers (approved):
LGTM! Thanks for the fix @kbalk

PR #4542, opened 2021-11-08
Add support for Label filter in get_parameters_by_path

dominikschubert:
See: https://docs.aws.amazon.com/systems-manager/latest/APIReference/API_GetParametersByPath.html

Adds the option to filter SSM parameters by their labels.

Example usage: `aws ssm get-parameters-by-path --path "/foo" --parameter-filters Key=Label,Values=label1`


whummer (approved):
Great, thanks for this fix @dominikschubert . 👍 Awaiting review from @bblommers ..   Thanks

bblommers (approved):
LGTM! Thanks for the PR @dominikschubert

Issue #4544, opened 2021-11-09
Cloudtrail `get_trail_status()` should also accept ARN in the `Name` parameter


kevinpark1217:
In the `boto3` documentation [here](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/cloudtrail.html#CloudTrail.Client.get_trail_status) for CloudTrail `get_trail_status()`, it specifies that the `Name` parameter can either be Cloudtrail Trail name or ARN. 

But `moto` only currently accepts name, not ARN. So, when ARN is passed, it throws `TrailNotFoundException` despite the trail existing.

Where modification is needed in `moto`:
[1](https://github.com/spulec/moto/blob/f4c4b349c7fbbf3568aa84fa6f246bdeedf1e68a/moto/cloudtrail/responses.py#L46-L49)
[2 (backend)](https://github.com/spulec/moto/blob/f4c4b349c7fbbf3568aa84fa6f246bdeedf1e68a/moto/cloudtrail/models.py#L215-L223)

PR #4547, opened 2021-11-09
Cloudtrail `get_trail_status()` also accepts ARN in `Name` parameter

kevinpark1217:
Fix for the issue #4544

### DIFF
--- a/moto/cloudtrail/models.py
+++ b/moto/cloudtrail/models.py
@@ -215,11 +215,19 @@ class CloudTrailBackend(BaseBackend):
     def get_trail_status(self, name):
         if len(name) < 3:
             raise TrailNameTooShort(actual_length=len(name))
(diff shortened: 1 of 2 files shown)


bblommers (changes requested):
Hi @kevinpark1217, thanks for the PR! Would you mind adding a new test for this scenario?

Existing tests can be found here, so you can copy/paste as required: https://github.com/spulec/moto/blob/master/tests/test_cloudtrail/test_cloudtrail.py#L256

kevinpark1217:
@bblommers Hey! thanks for the quick reply. I just added a unit test to the same commit. And here is the output below.

Screen Shot 2021-11-09 at 11 13 15 AM

bblommers (approved):
LGTM! Thanks again @kevinpark1217

Issue #4546, opened 2021-11-09
Search cognito ids while IoT Core attach policy


acapagarcia:
Hello,

I am trying to test iot `attach_policy` with cognito identity instead of iot core certificate but it raises an exception.
Doing a research I have found that during attach_policy uses the next method in order to found the target:
```
def _get_principal(self, principal_arn):
        """
        raise ResourceNotFoundException
        """
        if ":cert/" in principal_arn:
            certs = [_ for _ in self.certificates.values() if _.arn == principal_arn]
            if len(certs) == 0:
                raise ResourceNotFoundException()
            principal = certs[0]
            return principal
        else:
            # TODO: search for cognito_ids
            pass
        raise ResourceNotFoundException()
```

I am trying to implement this TODO section but previously created identity pool and user identity are not accessible from this `_get_principal` method.

I need to know if you have thought of any way to query cognito identities from other mocked services.

Any suggestion?

Thanks!

bblommers:
Hi @acapagarcia! You can import other service backends, and access all data stored in that service.
For performance reasons, the import should happen as late as possible.

So for this usecase, you can use:
```
def _get_principal(self, principal_arn):
        """
        raise ResourceNotFoundException
        """
        if ":cert/" in principal_arn:
            ...
        else:
            from moto.cognitoidp import cognitoidp_backends
            cognito = cognitoidp_backends["whatever-region-you're-in"]
            pools = cognito.list_user_pools()
            users = cognito.list_users()
```

I'm not too familiar with the IOT service, but let me know if that's helpful.

acapagarcia:
> Hi @acapagarcia! You can import other service backends, and access all data stored in that service.
> For performance reasons, the import should happen as late as possible.
> 
> So for this usecase, you can use:
> ```
> def _get_principal(self, principal_arn):
>         """
>         raise ResourceNotFoundException
>         """
>         if ":cert/" in principal_arn:
>             ...
>         else:
>             from moto.cognitoidp import cognitoidp_backends
>             cognito = cognitoidp_backends["whatever-region-you're-in"]
>             pools = cognito.list_user_pools()
>             users = cognito.list_users()
> ```
> 
> I'm not too familiar with the IOT service, but let me know if that's helpful.

Hi @bblommers! 
Ok. Tomorrow I will try what you say.
If everything works fine I will create a PR.

acapagarcia:
I have created new PR with changes https://github.com/spulec/moto/pull/4554

PR #4555, opened 2021-11-10
Search cognito identities from iot moto backend

acapagarcia:
Search cognito identities from iot core moto backend


acapagarcia:
Related to issue https://github.com/spulec/moto/issues/4546

moto/iot/models.py
```diff
@@ -911,9 +911,17 @@ def _get_principal(self, principal_arn):
...
+
+        cognito = cognitoidentity_backends[self.region_name]
+        identities = []
+        for identity_pool in cognito.identity_pools:
+            pool_identities = cognito.pools_identities.get(identity_pool, None)
+            identities.extend(
+                [pi["IdentityId"] for pi in pool_identities.get("Identities", [])]
+            )
+        if principal_arn in identities:
+            return {"IdentityId": principal_arn}
```
bblommers, line 924:
```suggestion
            identities = [pi["IdentityId"] for pi in pool_identities.get("Identities", [])]
            if principal_arn in identities:
                return {"IdentityId": principal_arn}
```
If I'm not mistaken, this would be a better solution in terms of performance. No need to cycle through all identity pools if the ID exists in the first one.

bblommers (changes requested):

moto/iot/models.py
acapagarcia: I have already made the changes

bblommers (approved):
LGTM! Thanks for the PR @acapagarcia

PR #4559, opened 2021-11-11
Route53Resolver: Add resolver endpoint-related APIs

kbalk:
Add support for the following APIs:

NOTE:  This does not include support for IPv6 addresses.

- `create_resolver_endpoint`
- `delete_resolver_endpoint`
- `get_resolver_endpoint`
- `list_resolver_endpoint_ip_addresses`  (filters will be handled in a different PR)
- `list_resolver_endpoints`
- `list_tags_for_resource`
- `tag_resource`
- `untag_resource`
- `update_resolver_endpoint`


kbalk:
@bblommers When I run the unit tests, I'm getting an intermittent failure and could use some help to make sense of it:

```
E   botocore.exceptions.ClientError: An error occurred (404) when calling the CreateResolverEndpoint operation: 
<!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 3.2 Final//EN">
E   <title>404 Not Found</title>
E   <h1>Not Found</h1>
E   <p>The requested URL was not found on the server. If you entered the URL manually please check your spelling and try again.</p>
```

Please ignore some of my ugly code and loose ends ... I'm still sorting out a bunch of issues.

bblommers:
> 
> Not sure of the best way to fix it and also if I can get away with only modifying `backend_index.py` or if I need to modify route53's `urls.py` as well.

Modifying the Route53 URL's would probably be better, to make sure that Route53 only intercepts calls to:
`route53.amazonaws.com`
and
`route53.region.amazonaws.com`

Out the top of my head, something like this could work:
`route53(\..+)?\.amazonaws\.com`

kbalk:
> Modifying the Route53 URL's would probably be better, to make sure that Route53 only intercepts calls to: `route53.amazonaws.com` and `route53.region.amazonaws.com`
> 
> Out the top of my head, something like this could work: `route53(\..+)?\.amazonaws\.com`

That sounds right ... I'll give it a try later (got a cold).  I think I saw other urls that were a bit "generous" in their matching, but it doesn't make sense for me to mess with them.  I'll need to change route53's `urls.py` and `backend_index.py`.  

Thank you for your help and suggestion, Bert!

kbalk:
@bblommers Worked beautifully, Bert.  The route53 unit tests still pass with the change.

kbalk:
I tried a terraform script that uses Route53Resolver and it can't find the service:

```
Unable to find appropriate backend for Route53Resolver.us-east-1.amazonaws.com.Remember to add the URL to urls.py, and run script/update_backend_index.py to index it.
```

kbalk:
@bblommers Found an anomaly with route53resolver.  `moto/server.py` was unable to find the backend because the service had capital R's in the auth string:

```
'AWS4-HMAC-SHA256 Credential=testing/20211115/us-east-1/Route53Resolver/aws4_request, 
SignedHeaders=content-length;content-type;host;x-amz-date;x-amz-security-token;x-amz-target, 
Signature=5fb111a1e7b2da980d38227d81a35f327f2cae3501f7e6340569c88a560e415a'
```

I can fix it in `infer_service_region_host()` either by forcing all services to lower case like this:

```
  credential_scope = auth.split(",")[0].split()[1]
  _, _, region, service, _ = credential_scope.split("/")
  service = SIGNING_ALIASES.get(service.lower(), service)
  service = service.lower().   <-----------------------------------
```

or add a special case similar to ones like this:

```
  elif service == "timestream":     # add a test for "Route53Resolver and create host str
     host = "ingest.{service}.{region}.amazonaws.com".format(
        service=service, region=region
     )
```

Do you have a preference?  I made the service lowercase for now, but I can change it.

bblommers:
> 
> Do you have a preference? I made the service lowercase for now, but I can change it.

Gah. Of course this service had to have something special... I'd prefer to lowercase the service @kbalk , so it will still work for others that decide to use uppercase

bblommers (approved):
Nice @kbalk - LGTM!

PR #4560, opened 2021-11-11
EC2 - fix missing AZ on auto-generated volumes for instances

dominikschubert:
Using `aws ec2 run-instance ...` automatically creates a Volume. This volume had an empty AvailabilityZone field when not deployed in us-east-1 since this was hardcoded previously. Creating Volumes manually did set the AZ properly, so no changes where necessary in this case.

### DIFF
--- a/tests/test_ec2/test_elastic_block_store.py
+++ b/tests/test_ec2/test_elastic_block_store.py
@@ -561,6 +561,11 @@ def test_volume_attach_and_detach_boto3():
     ec2 = boto3.resource("ec2", region_name="us-east-1")
     reservation = client.run_instances(ImageId=EXAMPLE_AMI_ID, MinCount=1, MaxCount=1)
     instance = reservation["Instances"][0]
+
+    volumes = client.describe_volumes()['Volumes']
+    volumes.should.have.length_of(1)
+    volumes[0]['AvailabilityZone'].should.equal("us-east-1a")
+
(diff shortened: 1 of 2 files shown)


tests/test_ec2/test_elastic_block_store.py
```diff
@@ -561,6 +561,11 @@ def test_volume_attach_and_detach_boto3():
     ec2 = boto3.resource("ec2", region_name="us-east-1")
     reservation = client.run_instances(ImageId=EXAMPLE_AMI_ID, MinCount=1, MaxCount=1)
     instance = reservation["Instances"][0]
+
+    volumes = client.describe_volumes()["Volumes"]
```
bblommers, line 565: EC2 tests run in parallel in ServerMode, so this call could also return volumes from other tests.
You can use `describe_volumes(VolumeIds=[..])` to only get the volume that you need.

(See http://docs.getmoto.org/en/latest/docs/contributing/development_tips/tests.html for more info)

bblommers (changes requested):

tests/test_ec2/test_elastic_block_store.py
dominikschubert: Good catch there. 

This didn't work in my case because I don't have a volume id to filter by (run_instances doesn't return a volume id).

Fortunately `describe_volumes` also allows Filters that support matching by instance id :+1:

bblommers (approved):
LGTM! Thanks for the fix @dominikschubert

PR #4562, opened 2021-11-11
Feature: Allow custom endpoints for S3

### LINKED ISSUE #2993: S3: Allow custom urls / Add hint that custom urls will not work
In our project we are using a ceph s3 storage (hosted on prem) and we wanted to mock it with mock_s3.

Problem is our url looks like:
https://s3.*some-internal-naming-convention*.de

Using this internal URL always raised the following issue:

```BASH
botocore.exceptions.ClientError: An error occurred (InvalidAccessKeyId) when calling the CreateBucket operation: Unknown
```

Which led me in the completely wrong direction for looking for an error fix :smile: 

After using the endpoint_url: https://s3.amazonaws.com everything works as expected.

If you do not plan to allow urls like the one we use, that is completely okay, but if someone faces the issue in the future he will at least see this issue and will be faster in solving it than I was :smiley: 

Appart from that adding a hint in the Readme would maybe help some people

bblommers:
Fixes #2993 

This extends Moto's functionality to also work when developing against a S3-compatible service like Ceph, Digital Ocean, etc.


moto/core/models.py
```diff
@@ -702,7 +703,7 @@ def urls(self):
         A dictionary of the urls to be mocked with this service and the handlers
         that should be called in their place
         """
-        url_bases = self._url_module.url_bases
+        url_bases = self.url_bases
```
bpandola, line 706: Curious why this change was needed...

scripts/implementation_coverage.py
```diff
@@ -34,11 +34,21 @@ def get_moto_implementation(service_name):
         return backends[0], mock_name
 
 
+def get_module_name(o):
```
bpandola, line 37: Are the changes to this script required as a result of the S3 changes?

bpandola:
@bblommers Implementation makes sense to me.  I posted a couple questions inline, just for my own understanding.

moto/core/models.py
bblommers: It wasn't needed, just a simplification, as `self.url_bases` points to `self._url_module.url_bases`

scripts/implementation_coverage.py
bblommers: Ah, sorry, I should've left a comment there to explain why that's included. 

Again not necessary, but it improves the look/layout of the documentation. You can see a preview here: https://moto--4562.org.readthedocs.build/en/4562/docs/services/s3.html

With this change, the documentation writing should be a lot simpler:

1.  The developer adds documentation as a doc-string
2.  he implementation-coverage script hoovers up all the doc-strings, and adds it to the required RTD pages

The implementation coverage is updated before every release, so this workflow ensures that the RTD-pages are also updated automatically before a release.

bpandola (approved):

PR #4574, opened 2021-11-14
Set version_id to null without bucket versioning

jmsanders:
The real S3 API returns `"null"` for `versionId` if bucket versioning
is disabled; not `None`. Moto returns `None`.

This causes behavior for the added test case to diverge between the real
API and the mock one: if you enable bucket configuration after objects
already exist in the bucket, it becomes impossible to empty the bucket
with moto.


bblommers (approved):
LGTM! Thanks for the fix @jmsanders

Issue #4573, opened 2021-11-14
Can drop more-itertools from setup.py?


hajapy:
I cannot find any reference of more-itertools [in the repository](https://github.com/spulec/moto/search?q=more_itertools&type=code). Is this dependency still needed, and if not, can it be removed from install_requires?

https://github.com/spulec/moto/blob/60d9667da26e91c69232d3366039b0e8da823339/setup.py#L42

PR #4576, opened 2021-11-14
Remove more-itertools from install_requires

hajapy:
I don't believe more-itertools is in use. 

Closes https://github.com/spulec/moto/issues/4573


bblommers (approved):
Good catch @hajapy.  :+1: 

One of our dependencies needed it, back when we still supported Py2, and we had to pin it to a version that still supported Py2.
After we dropped Py2 support, we simply unpinned the version - and never realized we didn't use it in the first place.

Issue #4575, opened 2021-11-14
six still in use despite being dropped in 2.2.0


hajapy:
There remains an import of six.moves here: https://github.com/spulec/moto/blob/2.2.15/moto/utilities/paginator.py#L4

despite the requirement being dropped and most other uses removed https://github.com/spulec/moto/commit/7693d7733345c29da0d007392d70b1acdfac2b62#diff-60f61ab7a8d1910d86d9fda2261620314edcae5894d5aaa236b821c7256badd7

Should this just use `functools.reduce` now?

PR #4577, opened 2021-11-14
Remove six import

hajapy:
I don't believe six should be used anymore. 

Closes https://github.com/spulec/moto/issues/4575


bpandola:
Good catch, @hajapy.  (The reason this import wasn't failing is because downstream packages still import `six`.)

PR #4587, opened 2021-11-17
Core - Use json module for error_json exception type rendering instead of jinja2

dominikschubert:
Encountered an issue with the json serialization/parsing of exception messages because they were simply substituted in the jinja2 template instead of being properly encoded. Since there didn't seem to be any reason to keep using jinja2 for this manual json rendering, I've replaced this in favor of using the built-in json module.

### DIFF
--- a/moto/core/exceptions.py
+++ b/moto/core/exceptions.py
@@ -53,22 +53,22 @@ class RESTError(HTTPException):
         "single_error": SINGLE_ERROR_RESPONSE,
         "wrapped_single_error": WRAPPED_SINGLE_ERROR_RESPONSE,
         "error": ERROR_RESPONSE,
-        "error_json": ERROR_JSON_RESPONSE,
     }
 
     def __init__(self, error_type, message, template="error", **kwargs):
         super(RESTError, self).__init__()
-        env = Environment(loader=DictLoader(self.templates))
         self.error_type = error_type
         self.message = message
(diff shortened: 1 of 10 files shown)


whummer (approved):
Awesome, nice catch and great fixes, thanks @dominikschubert 👍 . Awaiting feedback and review from @bblommers .. 🙏 Thanks!

moto/core/exceptions.py
```diff
@@ -53,22 +53,22 @@ class RESTError(HTTPException):
         "single_error": SINGLE_ERROR_RESPONSE,
         "wrapped_single_error": WRAPPED_SINGLE_ERROR_RESPONSE,
         "error": ERROR_RESPONSE,
-        "error_json": ERROR_JSON_RESPONSE,
```
bblommers, line 56: Can we remove the `ERROR_JSON_RESPONSE`-global, or is that still in use somewhere else?

bblommers (changes requested):

moto/core/exceptions.py
dominikschubert: Yeah actually thought I had deleted this already. :thinking:  Removed this now and I've also changed the exception rendering for managedblockchain in the same way.

bblommers (approved):
LGTM! Thanks for the fix @dominikschubert

PR #4588, opened 2021-11-17
DirectoryService: Create security group and ENI

kbalk:
When creating a directory, a security group should be created, which is then used to create a network interface.  The code currently contains some poor code to determine IP addresses, when it should be using the created ENI IP addresses.

Also, clean up the validation handling a bit.


kbalk:
@bblommers  I forgot to mark this PR as draft initially.  It's ready for review now.

moto/ds/exceptions.py
```diff
@@ -22,7 +22,7 @@ def __init__(self, error_tuples):
         )
         msgs = []
         for arg_name, arg_value, constraint in error_tuples:
-            value = "at" if arg_name == "password" else f"'{arg_value}' at"
+            value = "at" if "assword" in arg_name else f"'{arg_value}' at"
```
kbalk, line 25: This is rather unfortunate ...

bblommers (approved):
LGTM! Nice that we now have a proper EC2 integration

PR #4593, opened 2021-11-18
[s3] Allow KMS SSE parameters in copy_object

FloatingGhost:
AWS S3 allows [KMS key parameters to be passed in a copy_object request](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3.html#S3.Client.copy_object), however moto did not pass these on

This PR simply passes the parameters through

There is also one supplementary change of test_s3.py wherein a create_bucket call was failing with an invalid region


bblommers (approved):
LGTM! Thanks for the fix @FloatingGhost

PR #4596, opened 2021-11-18
Extend mock_all() to work as a context manager

jmsanders:


moto/__init__.py
```diff
@@ -171,22 +172,24 @@ def f(*args, **kwargs):
...
-    ]
+class mock_all(ContextDecorator):
+    def __init__(self):
+        self.mocks = [
+            globals()[mock]()
+            for mock in dir(sys.modules["moto"])
+            if mock.startswith("mock_")
+            and not mock.endswith("_deprecated")
+            and not mock == "mock_all"
+            and not mock == "mock_xray_client"
```
jmsanders, line 183: I'm not entirely sure why this is a lazyloaded at the top level. I figure it's probably safe - albeit a little dirty - to ignore it given that `mock_xray` still gets included but I'll defer to your judgement here.

moto/__init__.py
```diff
@@ -1,5 +1,6 @@
 import importlib
 import sys
+from contextlib import ContextDecorator
```
jmsanders, line 3: First available in Python 3.2 - what's the minimum supported moto version? Judging from setup.py, it's 3.5, but I'm not sure if there's an effort to remain further backwards compatible.

Makefile
```diff
@@ -3,7 +3,7 @@ SHELL := /bin/bash
 ifeq ($(TEST_SERVER_MODE), true)
 	# exclude test_kinesisvideoarchivedmedia
 	# because testing with moto_server is difficult with data-endpoint
-	TEST_EXCLUDE := -k 'not (test_kinesisvideoarchivedmedia or test_awslambda or test_batch or test_ec2 or test_sqs)'
+	TEST_EXCLUDE := -k 'not (test_kinesisvideoarchivedmedia or test_awslambda or test_batch or test_ec2 or test_sqs or test_context_manager)'
```
jmsanders, line 6: I think I'm understanding the reason for this failure correctly - are you comfortable skipping the test in server mode?

https://github.com/spulec/moto/pull/4596/commits/4d87f3c786500fa777630da79c8c1873486e9e41

> The test tries to list SQS queues outside of the scope of the context
> manager to prove that the context manager exited correctly.
> 
> But when running in server mode, this call still succeeds because the
> server is still running.

moto/__init__.py
bblommers: It's 3.5 indeed - although that's been EOL for a while now, so we'll probably drop that soon. There's no need to make things backwards compatible, so using this import is fine

tests/test_core/test_mock_all.py
```diff
@@ -18,3 +19,22 @@ def test_multiple_services():
...
+        lmbda = boto3.client("lambda", region_name=rgn)
+        r = lmbda.list_event_source_mappings()
+        r["ResponseMetadata"]["HTTPStatusCode"].should.equal(200)
+
+        ddb = boto3.client("dynamodb", region_name=rgn)
+        r = ddb.list_tables()
+        r["ResponseMetadata"]["HTTPStatusCode"].should.equal(200)
+
+    with pytest.raises(Exception):
+        sqs.list_queues()
```
bblommers, line 40: ServerMode will patch any clients inside a decorator by setting the `endpoind_url`-attribute.
So when calling this method here, it uses the client that still has the `endpoint_url` set - regardless of whether the mock is active or not.

As clients are cached within a session, the best option here is probably to create a new session, and let that session create a 'clean' client (without the endpoint_url):
```
with mock_all():
    ....

unpatched_sqs = boto3.Session().client("sqs", region_name=rgn)

with pytest.raises(Exception):
    unpatched_sqs.list_queues()
```

That passes in both the unit tests and ServerMode tests

bblommers (changes requested):
Hi @jmsanders, thanks for the PR! 
It does make sense to have `mock_all` as a context manager, but we do need to find a solution where `mock_xray_client` is included.

The XRay client is a mock that only works as a function-decorator, not as a context decorator.

As I see it, there are two options here:
 - Create a mock_all that invokes all mocks when used as a decorator, but only invokes mocks that support context management when used as a context manager
 - Rewrite the XRay client to also work as a context manager

As far as I can see, both options are possible - not sure what's the easiest/cleanest option is though. Let me know if you want me to look into this a little more.

PR #4598, opened 2021-11-19
Route53Resolver:  Add support for filters for list_resolver_endpoints

kbalk:
I tried to write this so the filter validation functions can be used by the other Route53Resolver list functions.  

Note that the allowed filter names can be in the old style (all caps, snake case) or in the new style (camel case), which complicates things.  And of course, there is an exception to the filter name rules!


bblommers (approved):
There's always an exception :slightly_smiling_face: Thanks for the PR!

PR #4599, opened 2021-11-19
SQS - Fix message attributes disappearing on multiple receives

dominikschubert:
See the test for the behavior this PR is trying to fix.

Because the filtering in `receive_message` was removing items from the attributes on the actual message item in the backend, the next (less restricted) call would not be able to get the previously filtered out attributes anymore.

### DIFF
--- a/moto/sqs/models.py
+++ b/moto/sqs/models.py
@@ -867,12 +867,14 @@ class SQSBackend(BaseBackend):
 
                 queue.pending_messages.add(message)
                 message.mark_received(visibility_timeout=visibility_timeout)
-                _filter_message_attributes(message, message_attribute_names)
+                # Create deepcopy to not mutate the message state when filtering for attributes
+                message_copy = deepcopy(message)
+                _filter_message_attributes(message_copy, message_attribute_names)
                 if not self.is_message_valid_based_on_retention_period(
                     queue_name, message
                 ):
                     break
-                result.append(message)
+                result.append(message_copy)
                 if len(result) >= count:
                     break
 
--- a/moto/sqs/models.py
+++ b/moto/sqs/models.py
@@ -534,7 +534,12 @@ class Queue(CloudFormationModel):
                 [backend.delete_message(self.name, m.receipt_handle) for m in messages]
             else:
                 # Make messages visible again
-                [m.change_visibility(visibility_timeout=0) for m in messages]
+                [
+                    backend.change_message_visibility(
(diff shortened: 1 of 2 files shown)


moto/sqs/models.py
```diff
@@ -534,7 +534,12 @@ def add_message(self, message):
                 [backend.delete_message(self.name, m.receipt_handle) for m in messages]
             else:
                 # Make messages visible again
-                [m.change_visibility(visibility_timeout=0) for m in messages]
+                [
```
bblommers, line 537: What's the reason for this change?

tests/test_sqs/test_sqs.py
```diff
@@ -3315,3 +3315,43 @@ def test_message_attributes_contains_trace_header():
...
+            "Custom1": {"StringValue": "Custom_Value_1", "DataType": "String"},
+            "Custom2": {"StringValue": "Custom_Value_2", "DataType": "String"},
+        },
+    )
+
+    first_messages = conn.receive_message(
+        QueueUrl=queue.url,
+        MaxNumberOfMessages=2,
+        MessageAttributeNames=["Custom1"],
+        VisibilityTimeout=1,
```
bblommers, line 3341: If the VisibilityTimeout is 0, we shouldn't need the `sleep`-statement anymore - would help a lot in speeding up the tests

bblommers (changes requested):

moto/sqs/models.py
dominikschubert: Otherwise it could now modify the returned message copy (the one with possibly filtered attributes) and the original message in the backend wouldn't be touched.

By using `backend.change_message_visibility` this can be avoided since it only references the receipt handle of the message and will do a lookup of the message in its internal `_messages` field. Previously this worked because these were always the same message objects.


This change was now required for this test to pass after the changes of this PR: 

https://github.com/spulec/moto/blob/be197caba6282260ca53112f1f5c7ab4a6675ca9/tests/test_awslambda/test_lambda_eventsourcemapping.py#L213-L274

tests/test_sqs/test_sqs.py
dominikschubert: Good point, I've updated the test :+1:

moto/sqs/models.py
bblommers: Gotcha, that makes sense

bblommers (approved):
LGTM! Thanks for the PR @dominikschubert

PR #4603, opened 2021-11-19
Route53Resolver: Add resolver rule-related APIs

kbalk:
Adds support for the following APIs:

- `create_resolver_rule`
- `get_resolver_rule`
- `delete_resolver_rule`
- `list_resolver_rules`

### DIFF
--- a/tests/test_route53resolver/test_route53resolver_rule.py
+++ b/tests/test_route53resolver/test_route53resolver_rule.py
@@ -12,7 +12,7 @@ from moto.core import ACCOUNT_ID
 from moto.core.utils import get_random_hex
 from moto.ec2 import mock_ec2
 
-from .test_route53resolver_endpoint import TEST_REGION, create_test_endpoint
+from .test_route53resolver_endpoint import create_test_endpoint
 
 TEST_REGION = "us-east-1" if settings.TEST_SERVER_MODE else "us-west-2"
 
--- a/tests/test_route53resolver/test_route53resolver_rule.py
+++ b/tests/test_route53resolver/test_route53resolver_rule.py
@@ -7,7 +7,6 @@ from botocore.exceptions import ClientError
 import pytest
 
 from moto import mock_route53resolver
-from moto import settings
 from moto.core import ACCOUNT_ID
 from moto.core.utils import get_random_hex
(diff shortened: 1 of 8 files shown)


tests/test_route53resolver/test_route53resolver_rule.py
```diff
@@ -0,0 +1,535 @@
...
+
+from moto import mock_route53resolver
+from moto import settings
+from moto.core import ACCOUNT_ID
+from moto.core.utils import get_random_hex
+from moto.ec2 import mock_ec2
+
+from .test_route53resolver_endpoint import TEST_REGION, create_test_endpoint
+
+TEST_REGION = "us-east-1" if settings.TEST_SERVER_MODE else "us-west-2"
```
bblommers, line 17: This was just imported, so we don't need to replicate it.

This format pops up in quite a few tests, and I'm not sure why. I'd be happy with either standardizing to one region, or using different regions in different test-methods (to verify that Moto can read/intercept the region correctly for this service)

bblommers (changes requested):
Thanks @kbalk - just got a minor nitpick in the tests

tests/test_route53resolver/test_route53resolver_rule.py
kbalk: Good eye!  I think back when I worked on the Config service, I discovered that the regions were different.  A test was failing that had a hard-coded region in it and it needed to be a different region.  So since then, I've been making the distinction.  I was curious why the different regions. 

I think having one TEST_REGION defined somewhere would be good.  I wish we didn't have to define it in the first place.

kbalk: Realized that I forgot to push my second change for this issue.  It should be good to go now unless you would like it done differently.

bblommers: > I think having one TEST_REGION defined somewhere would be good. I wish we didn't have to define it in the first place.
> 

It can be useful as a sanity check, to verify that ServerMode is able to read the region correctly. In most cases it's part of the standard Auth-header, butthere are always exceptions...

bblommers (approved):
LGTM @kbalk 👍

PR #4607, opened 2021-11-21
Rework mock_all to work as a context manager

bblommers:
At the moment `@mock_all` works as a decorator only, but it should also be usable as a context manager (`with mock_all():`).

In order for this to work, the `mock_xray_client` also needed to be changed to work as a context manager.

Tests have been added to verify the behaviour of both.

Based on and supercedes #4596 - does this look OK to you @jmsanders?

### DIFF
--- a/moto/__init__.py
+++ b/moto/__init__.py
@@ -174,14 +174,17 @@ mock_sdb = lazy_load(".sdb", "mock_sdb", boto3_name="sdb")
 
 class mock_all(ContextDecorator):
     def __init__(self):
-        self.mocks = [
-            globals()[mock]()
-            for mock in dir(sys.modules["moto"])
-            if mock.startswith("mock_")
-            and not mock.endswith("_deprecated")
-            and not mock == "mock_all"
-            and not mock == "mock_xray_client"
-        ]
+        self.mocks = []
+        for mock in dir(sys.modules["moto"]):
+            if (
+                mock.startswith("mock_")
+                and not mock.endswith("_deprecated")
+                and not mock == ("mock_all")
+            ):
+                try:
+                    self.mocks.append(globals()[mock]())
+                except:
+                    continue
 
     def __enter__(self):
         for mock in self.mocks:
--- a/moto/__init__.py
+++ b/moto/__init__.py
@@ -1,5 +1,6 @@
 import importlib
 import sys
+from contextlib import ContextDecorator
 
 
 def lazy_load(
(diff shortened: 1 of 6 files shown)


tests/test_xray/test_xray_client.py
```diff
@@ -45,6 +45,37 @@ def test_xray_dynamo_request_id():
     setattr(requests.Session, "prepare_request", original_session_prep_request)
 
 
+def test_xray_dynamo_request_id_with_context_mgr():
+    with mock_xray_client():
```
jmsanders, line 49: This test ensures that the context manager starts correctly, but I think it’s be a good idea to also add an assertion showing that it stops too.

moto/__init__.py
```diff
@@ -171,22 +172,24 @@ def f(*args, **kwargs):
...
-        d
-        for d in dir(sys.modules["moto"])
-        if d.startswith("mock_")
-        and not d.endswith("_deprecated")
-        and not d == "mock_all"
-    ]
+class mock_all(ContextDecorator):
+    def __init__(self):
+        self.mocks = []
+        for mock in dir(sys.modules["moto"]):
```
jmsanders, line 178: Nitpick, but now that we’re not doing a try/except, we could switch back to list comprehension instead of a for loop. Similar to how it was written in the original commit (minus the X-ray client exclusion):

https://github.com/spulec/moto/blob/a7ae379a7d5c6934ad0ca8ea5257dce130ad59d4/moto/__init__.py#L177-L184

I’m fine with either - up to you what you find more readable.

jmsanders (commented):
Awesome! Thanks - just a few minor bits of feedback. I’ll close https://github.com/spulec/moto/pull/4596

moto/__init__.py
bblommers: The for-loop is more readable IMO - the indentation helps understand better what's going on

tests/test_xray/test_xray_client.py
bblommers: Good point - added

moto/__init__.py
```diff
@@ -171,22 +172,24 @@ def f(*args, **kwargs):
...
 
-def mock_all():
-    dec_names = [
-        d
-        for d in dir(sys.modules["moto"])
-        if d.startswith("mock_")
-        and not d.endswith("_deprecated")
-        and not d == "mock_all"
-    ]
+class mock_all(ContextDecorator):
```
dedanaronson, line 175: nitpick from my side as well 😶‍🌫️ 
WDYT about:
```python 
class MockAll(ContextDecorator):
    ...

#in the end of the file:
mock_all = MockAll()
```

bblommers: Good catch - that is a better (more pythonic) naming scheme :+1:

bpandola (approved):

PR #4610, opened 2021-11-21
Implement list jobs

dedanaronson:


moto/glue/models.py
```diff
@@ -183,6 +183,32 @@ def create_job(
         )
         return name
 
+    def list_jobs(self, next_token, max_results, tags):
+        next_token = int(next_token) if next_token else 0
```
bblommers, line 187: What is the value/format that AWS uses for `next_token`?
If AWS uses a randomly generated string, it would be easier to use the `paginator`-decorator for these two methods.
That way we don't have to deal with the pagination logic ourselves.

See this bit in the docs with some guidelines on how to implement that: http://docs.getmoto.org/en/latest/docs/contributing/development_tips/utilities.html#paginator

(If AWS does use a simple number for the `next_token`, and we want to replicate that exact format, we can keep this implementation as is.)

moto/glue/models.py
```diff
@@ -183,6 +183,32 @@ def create_job(
...
+        total_number_of_jobs = len(self.jobs)
+        max_results = max_results or total_number_of_jobs
+        last_token = next_token + max_results
+        job_names = self.filter_jobs_by_tags(tags, next_token, last_token)
```
bblommers, line 191: Naming is wrong, as we are returning the entire job here - not just the names

moto/glue/models.py
```diff
@@ -183,6 +183,32 @@ def create_job(
...
+        return (
+            job_names,
+            last_token if last_token <= total_number_of_jobs else total_number_of_jobs,
```
bblommers, line 194: AWS does not return a value for `next_token` if there are no more results (or rather: `next_token=None`)

If a developer wants to retrieve all jobs, they would create a loop like this:
```
while next_token:
    jobs, next_token = list_jobs(next_token)
```
The current implementation would create a never-ending loop, if we always return a value for `next_token`

moto/glue/responses.py
```diff
@@ -360,3 +360,14 @@ def create_job(self):
...
+
+    def list_jobs(self):
+        next_token = self._get_param("NextToken")
+        max_results = self._get_int_param("MaxResults")
+        tags = self._get_param("Tags")
+        job_names, next_token = self.glue_backend.list_jobs(
+            next_token=next_token, max_results=max_results, tags=tags,
+        )
+        return json.dumps(
+            dict(JobNames=[job.as_dict() for job in job_names], NextToken=next_token)
```
bblommers, line 372: boto3 only returns the JobNames, not the entire dict. Looks like we don't need the `as_dict`-method yet, and this can simply say `JobNames=[job.get_name() for job in jobs]`

tests/test_glue/test_glue.py
```diff
@@ -9,19 +12,82 @@
...
+
+
+@mock_glue
+def test_list_jobs():
+    client = create_glue_client()
+    expected_jobs = randint(1, 15)
+    for _ in range(expected_jobs):
+        create_test_job(client)
+    response = client.list_jobs()
+    response["JobNames"].should.have.length_of(expected_jobs)
```
bblommers, line 41: Following from an earlier comment - we should check here that the NextToken is not returned if all jobs are returned.
`response.shouldnt.have.key("NextToken")`

bblommers (changes requested):

moto/glue/models.py
dedanaronson: Changed accordingly 😄

moto/glue/models.py
dedanaronson: Added `test_next_token_logic_does_not_create_infinite_loop` just to be on the safe side 😄

moto/glue/responses.py
dedanaronson: Sorry for not noticing that...

bblommers (approved):
LGTM! Thanks for the PR @dedanaronson

PR #4611, opened 2021-11-21
WPI: Route53Resolver: Add resolver rule association-related APIs

kbalk:
Adds support for the following APIs:

- `associate_resolver_rule`
- `disassociate_resolver_rule`
- `get_resolver_rule_association`
- `list_resolver_rule_associations`

### DIFF
--- a/IMPLEMENTATION_COVERAGE.md
+++ b/IMPLEMENTATION_COVERAGE.md
@@ -3675,12 +3675,12 @@
 
 ## route53resolver
 <details>
-<summary>14% implemented</summary>
+<summary>27% implemented</summary>
 
 - [ ] associate_firewall_rule_group
 - [ ] associate_resolver_endpoint_ip_address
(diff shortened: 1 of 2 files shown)


kbalk:
@bblommers I merged my code with the latest and now I'm seeing the following:

```
(Pdb) ec2_client.create_vpc(CidrBlock="10.0.0.0/16")
*** botocore.exceptions.ClientError: An error occurred (AuthFailure) when calling the CreateVpc operation: AWS was not able to validate the provided access credentials
```

It doesn't make sense.  I used this call in my other PRs without incident.

kbalk:
I was cleaning up my AWS account and realized that you can't:

- delete a rule unless the VPC's are disassociated
- delete an endpoint unless all the rules are removed

I'll see if I can add that logic.

bblommers (approved):
LGTM! Thanks @kbalk

PR #4621, opened 2021-11-22
Fix incorrect DnsEntries XML for describe_vpc_endpoints

phildamore-phdata:
The XML generated for describe_vpc_endpoints did not have the expected schema for DnsEntries, so creating an interface endpoint and then describing it gave you an empty DnsEntries array.  I fixed that and I updated the tests to test both gateway and interface endpoints.


bblommers (approved):
Nice one - thanks for the fix @phildamore-phdata!

PR #4626, opened 2021-11-23
Bump python-jose to support newer ecdsa versions

Dunedan:
As moto doesn't support Python 2.7 anymore, the minimum supported
version of python-jose doesn't need to support Python 2.7 either. This
allows updating the minimum supported version of python-jose to a
version with fixed version contraints on its ecdsa dependency, making
the workaround of pinning ecdsa to specific versions unnecessary. That
in turn then allows using newer ecdsa versions as well, as the version
constraint python-jose uses is ecdsa!=0.15 instead of ecdsa<0.15
currently used by moto.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -45,7 +45,7 @@ install_requires = [
 _dep_PyYAML = "PyYAML>=5.1"
 _dep_python_jose = "python-jose[cryptography]>=3.1.0,<4.0.0"
 _dep_python_jose_ecdsa_pin = (
-    "ecdsa<0.15"  # https://github.com/spulec/moto/pull/3263#discussion_r477404984
+    "ecdsa!=0.15"  # https://github.com/spulec/moto/pull/3263#discussion_r477404984
 )
 _dep_dataclasses = "dataclasses; python_version < '3.7'"
 _dep_docker = "docker>=2.5.1"
--- a/setup.py
+++ b/setup.py
@@ -43,10 +43,7 @@ install_requires = [
 ]
 
 _dep_PyYAML = "PyYAML>=5.1"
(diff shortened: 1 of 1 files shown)


bblommers (commented):
Hi @Dunedan! We try to keep the range of dependencies as wide as possible, just in case people are not able to upgrade `python-jose` due to some other dependency conflict in their setup.

Is there a specific bug/feature that would be solved by this PR?

Dunedan:
I agree with the intention to keep the range of possible versions as wide as possible, however the current dependencies prevent projects from using `ecdsa>=0.16` if they use `moto`.

While we have no specific feature we need from newer `ecdsa` versions we want to avoid the case that we're blocked by `moto` to apply a potential critical security fix for `ecdsa`.

An alternative solution to bumping `python-jose` would be to change the version constraint for `ecdsa` in `moto` from `ecdsa<0.15` to `ecdsa!=0.15` as `python-jose` does it as well.

bblommers:
> 
> An alternative solution to bumping `python-jose` would be to change the version constraint for `ecdsa` in `moto` from `ecdsa<0.15` to `ecdsa!=0.15` as `python-jose` does it as well.

Yup, that makes sense, I'd be happy with that solution.

Dunedan:
I just pushed commits for the alternative solution. If desired I can also do a force-push with squashed commit history.

bblommers:
> I just pushed commits for the alternative solution. If desired I can also do a force-push with squashed commit history.

That's OK, I'll just squash and merge. Thanks for the PR @Dunedan!

bblommers (approved):

PR #4627, opened 2021-11-23
Added restore table to point in time DynamoDB method

yardbirdsax:
I added the `restore_table_to_point_in_time` DyanmoDB method to the library. I _think_ I got adequate coverage for the unique use case I have (which is restore a table's most recent PIT to another table), but please let me know if you think I've missed anything.

**Please note**: There are 4 failing tests when I run the tests locally, which also failed when I checked out the `master` branch. They are in code I haven't touched so either it's a local issue or something pre-existing from what I can tell.

Thanks for your consideration!

### DIFF
--- a/moto/dynamodb2/models/__init__.py
+++ b/moto/dynamodb2/models/__init__.py
@@ -1009,6 +1009,7 @@ class RestoredTable(Table):
         }
         return result
 
+
 class RestoredPITTable(Table):
     def __init__(self, name, source):
         params = self._parse_params_from_table(source)
@@ -1025,10 +1026,10 @@ class RestoredPITTable(Table):
         params = {
             "schema": copy.deepcopy(table.schema),
             "attr": copy.deepcopy(table.attr),
-            "throughput": copy.deepcopy(table.throughput)
+            "throughput": copy.deepcopy(table.throughput),
         }
         return params
-    
+
     def describe(self, base_key="TableDescription"):
         result = super(RestoredPITTable, self).describe(base_key=base_key)
         result[base_key]["RestoreSummary"] = {
@@ -1037,6 +1038,7 @@ class RestoredPITTable(Table):
             "RestoreInProgress": False,
         }
         return result
+
 
 class Backup(object):
     def __init__(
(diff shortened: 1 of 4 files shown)


moto/dynamodb2/models/__init__.py
```diff
@@ -1690,6 +1720,17 @@ def restore_table_from_backup(self, target_table_name, backup_arn):
         self.tables[target_table_name] = new_table
         return new_table
 
+    def restore_table_to_point_in_time(self, target_table_name, source_table_name):
```
bblommers, line 1723: Would you mind adding a doc string here to document the current limitations? Can be as simple as
```
"""
Currently only takes the source/target table arguments, and copies all items from the source
"""
```

moto/dynamodb2/models/__init__.py
```diff
@@ -1690,6 +1720,17 @@ def restore_table_from_backup(self, target_table_name, backup_arn):
...
+        source = self.get_table(source_table_name)
+        if source is None:
+            raise KeyError()
```
bblommers, line 1726: There should be a separate test to verify this error is raised correctly. See this test for an example on how we verify error conditions:
https://github.com/spulec/moto/blob/master/tests/test_dynamodb2/test_dynamodb.py#L135:L141

moto/dynamodb2/models/__init__.py
```diff
@@ -1690,6 +1720,17 @@ def restore_table_from_backup(self, target_table_name, backup_arn):
...
+        existing_table = self.get_table(target_table_name)
+        if existing_table is not None:
+            raise ValueError()
```
bblommers, line 1729: Same here

bblommers (changes requested):
Hi @yardbirdsax, thanks for the PR! Looks good overall - just have some minor comments

moto/dynamodb2/models/__init__.py
yardbirdsax: Done!

moto/dynamodb2/models/__init__.py
yardbirdsax: Done, let me know if the test looks adequate.

moto/dynamodb2/models/__init__.py
yardbirdsax: Done, let me know if the test looks adequate.

yardbirdsax:
My bad on the failing linter. 🤦 It's fixed now!

bblommers (approved):
LGTM! Thanks @yardbirdsax

PR #4629, opened 2021-11-24
Directory Service, Route53Resolver:  Delete network interfaces created during initialization

kbalk:
Directory Service creates a security group and network interfaces, but those weren't being deleted when a directory was deleted.  Also, the Security Group Id should be displayed in the `describe_directory` output.

Route53Resolver creates network interfaces for the endpoint and those should be deleted when the endpoint is deleted.

### DIFF
--- a/moto/ds/models.py
+++ b/moto/ds/models.py
@@ -79,26 +79,31 @@ class Directory(BaseModel):  # pylint: disable=too-many-instance-attributes
         self.stage = "Active"
         self.launch_time = datetime.now(timezone.utc).isoformat()
         self.stage_last_updated_date_time = datetime.now(timezone.utc).isoformat()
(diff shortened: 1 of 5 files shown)


bblommers (changes requested):
Hey @kbalk, would you mind adding some assertions to the existing deletion-tests to verify that the ENI's/SG's have been deleted?

kbalk:
> Hey @kbalk, would you mind adding some assertions to the existing deletion-tests to verify that the ENI's/SG's have been deleted?

I knew you were going to make that suggestion, but got lazy!

kbalk:
@bblommers Made the additions ...

bblommers (approved):
Thanks @kbalk

PR #4630, opened 2021-11-24
emr auto scaling role

jdeepe:
Added AutoScalingRole='string' in describe emr clusters.


bpandola:
@jdeepe Please update the following test to pass in the new arg and assert its value on describe:
https://github.com/spulec/moto/blob/d64fd52b577aa8cea49399f743936c89385708c1/tests/test_emr/test_emr_boto3.py#L118

bpandola (approved):

PR #4633, opened 2021-11-24
added support for update IAM groups

macnev2013:
# Added
- added supoort for UpdateGroup API for IAM

### DIFF
--- a/moto/iam/models.py
+++ b/moto/iam/models.py
@@ -1574,6 +1574,8 @@ class IAMBackend(BaseBackend):
             policy = arns[policy_arn]
         except KeyError:
             raise IAMNotFoundException("Policy {0} was not found.".format(policy_arn))
+        if policy.arn in self.get_group(group_name).managed_policies.keys():
+            return
         policy.attach_to(self.get_group(group_name))
 
     def detach_group_policy(self, policy_arn, group_name):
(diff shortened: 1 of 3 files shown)


moto/iam/models.py
```diff
@@ -2120,6 +2122,29 @@ def delete_group(self, group_name):
...
+        existing_policies = group.managed_policies.copy()
+        for policy_arn in existing_policies:
+            self.detach_group_policy(policy_arn, group_name)
+        if new_path:
+            group.path = new_path
+        if new_group_name:
+            group.name = new_group_name
+            self.groups[new_group_name] = self.groups.pop(group_name)
+        for policy_arn in existing_policies:
+            self.attach_group_policy(policy_arn, new_group_name)
```
bblommers, line 2146: `new_group_name` could be None, so this call could fail. 

From what I can see the whole re-attaching business is only necessary when updating the name, so this can be rewritten as:
```
if new_group_name:
    existing_policies = group.managed_policies.copy()
    # detach policies...
    # update name...
    # attach policies...
```

bblommers (changes requested):

moto/iam/models.py
macnev2013: Thanks @bblommers.. Good point. Actually the whole logic is only necessary only if `new_group_name` exists..

moto/iam/models.py
```diff
@@ -2120,6 +2122,30 @@ def delete_group(self, group_name):
...
+                )
+
+            existing_policies = group.managed_policies.copy()
+            for policy_arn in existing_policies:
+                self.detach_group_policy(policy_arn, group_name)
+            if new_path:
+                group.path = new_path
+            if new_group_name:
+                group.name = new_group_name
+                self.groups[new_group_name] = self.groups.pop(group_name)
```
bblommers, line 2145:
```suggestion
            group.name = new_group_name
            self.groups[new_group_name] = self.groups.pop(group_name)
```

bblommers (changes requested):

bblommers (approved):
Thanks @macnev2013!

PR #4637, opened 2021-11-24
fix: STS xml was malformed

chrisevett:
Found from this issue in localstack https://github.com/localstack/localstack/issues/4807 we were getting \n characters in the xml for the GetFederationToken call. Thinking this will fix it


bblommers (approved):
Looks like the right fix indeed. Thanks for the PR @chrisevett!

PR #4639, opened 2021-11-25
Correct status code for a versioned key search in not versioned bucket.

pinzon:
This PR addresses issue localstack/localstack#4624.

Changes: 
- Before searching for a versioned key, now the S3 Backend checks if the bucket is versioned, if it isn't, it responds with the correct status code.


bblommers (changes requested):
Hi @pinzon, would you mind adding a test for this scenario?

pinzon:
@bblommers  done 👍

bblommers (approved):
LGTM! Thanks @pinzon

PR #4640, opened 2021-11-27
Feature - CloudFront

### LINKED ISSUE #2408: AWS CloudFront support
Hi there!

Any plans on implementing support for mocking AWS CloudFront?

Thanks!

bblommers:
Initial implementation of the CloudFront -service

Closes #2408

Methods implemented with this PR:
 - create_distribution
 - delete_distribution
 - get_distribution
 - list_distributions

The number of possible input-parameters for the `create_distribution`-method is endless. This PR only adds support for a small subset of possible parameters, focusing on a simple S3-backed distribution.

Custom distributions, backed by ELB/EC2/non-AWS domains have not been validated. They may work, they may not work.

The distribution-endpoint itself is not mocked.

This PR also extends the scaffolding-script to add support for the "rest-xml" protocol. The implementation is not perfect yet, but it's does provide a response-template that can be used as a starting point.

### DIFF
--- a/IMPLEMENTATION_COVERAGE.md
+++ b/IMPLEMENTATION_COVERAGE.md
@@ -398,6 +398,104 @@
 - [X] validate_template
 </details>
 
+## cloudfront
+<details>
+<summary>4% implemented</summary>
+
+- [ ] associate_alias
+- [ ] create_cache_policy
+- [ ] create_cloud_front_origin_access_identity
+- [X] create_distribution
+- [ ] create_distribution_with_tags
+- [ ] create_field_level_encryption_config
+- [ ] create_field_level_encryption_profile
+- [ ] create_function
+- [ ] create_invalidation
+- [ ] create_key_group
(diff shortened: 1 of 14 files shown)


kbalk:
Some possible future ideas for scaffold (I could create an issue if you'd like):

In the user documentation for the scaffold, I don't think it mentions that you can keep re-running the scaffold to add more API/operation templates to a new or existing service.  It'd be a nice feature if you could loop adding operations without having to exit, then restart scaffold to do so.  I like what you've done to better handle selection errors, which is why I was thinking about this.

You might already plan on doing this, but the file docstring doesn't indicate rest-xml as an operation.

It'd be nice if the scaffold did more for templating unit tests, but that's not as important as extracting the parameters for `responses.py`, which is does do.

However, I think scaffold is using something other than `_get_params` for certain types of parameters (lists and dicts, I think?) and I found myself having to correct it.  It was on the back of my mind to correct that and then I forgot.

PR #4641, opened 2021-11-28
Scaffold-script improvements

bblommers:
New features:

 - Adds support for the `rest-xml`-protocol (Mainly used by global services such as S3/Cloudfront/Route53)
 - Adds operation-specific test cases

Fixes:
 - In Responses, use the generic `self._get_param()` method, instead of type-specific methods
   `_get_param()` should work for any type - String/List/Dict/etc

UX fixes:

 - Allows the user to try again when providing an unknown service
 - Allow the user to try again when providing an unknown operation
 - Allow the user to try again when providing an operation already implemented
 - Allow the user to implement multiple operations in one session

@kbalk Thanks for the feedback earlier - you raised some good points on how to simplify/improve the usage of this script.
I've separated all the scaffold-changes out, so the other PR just contains the CloudFront-changes.

This should cover everything you mentioned, but let me know if I'm missing anything.

### DIFF
--- a/scripts/scaffold.py
+++ b/scripts/scaffold.py
@@ -10,7 +10,7 @@ This script will look at the botocore's definition file for the selected
 service and operation, then auto-generate the code and responses.
 
 Almost all services are supported, as long as the service's protocol is
-`query`, `json` or `rest-json`.  Even if aws adds new services, this script
+`query`, `json`, `rest-xml` or `rest-json`.  Even if aws adds new services, this script
 will work if the protocol is known.
 
 TODO:
@@ -20,6 +20,7 @@ TODO:
     should happen, please create an issue for the problem.
 """
 import os
+import random
 import re
 import inspect
 import importlib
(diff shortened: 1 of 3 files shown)


kbalk:
> This should cover everything you mentioned, but let me know if I'm missing anything.

This is really great ... it just keeps getting better!

I think it was `urls.py`, but I'd get a black error because the template put an extra blank line in the file.  If you run black and pylint on the files (I don't know about the rest-xml version), you'll see the error.

Regarding the tests -- I don't know if it makes sense or is worth it, but should there be two functions for each operation, one for good tests, one for bad tests, e.g. test_service_good_operation, test_service_bad_operation?   And maybe add the `get_random_hex()` or some other random number generator as an import to remind people to randomize ids in their tests?

bblommers:
> This is really great ... it just keeps getting better!
> 

All that's left now is automating the implementation, and we can go home! 

Developers will also have to run black/pylint to validate their own changes, so I'm not too worried about the fact that the file doesn't pass as is.

I'll add a link to the docs in the test template, as a gentle reminder to ~~RTFM~~ follow best practices. Based on that, people can make their own decision on whether to add good/bad tests, hex functions, etc. :+1:

kbalk:
Should the test name have the service name in it, i.e., test_service_operation?  I think you have it as test_operation?

bblommers:
Yup, that follows convention, I believe? `test_service/test_service.py::test_operation_action`

kbalk:
> Yup, that follows convention, I believe? `test_service/test_service.py::test_operation_action`

I always included the service name in the test function, but I copied what they were doing for one service and then just continued to do so.

PR #4648, opened 2021-12-02
Add support for IoT Domain Configuration

llawford:
Summary: 
Added support for the IoT DomainConfiguration object, and the create/update/describe/list/delete operations.
Added tests covering each operation plus validation. 

Checklist:
- [x] Feature is added
- [x] The linter is happy
- [x] Tests are added      
- [x] All tests pass, existing and new

### DIFF
--- a/moto/iot/models.py
+++ b/moto/iot/models.py
@@ -484,6 +484,74 @@ class FakeRule(BaseModel):
         }
 
 
+class FakeDomainConfiguration(BaseModel):
+    def __init__(
+        self,
+        region_name,
+        domain_configuration_name,
+        domain_name,
+        server_certificate_arns,
+        domain_configuration_status,
+        service_type,
+        authorizer_config,
+        domain_type,
+    ):
+        if domain_configuration_status not in ["ENABLED", "DISABLED"]:
+            raise InvalidRequestException(
+                " An error occurred (InvalidRequestException) when calling the DescribeDomainConfiguration "
+                "operation: Domain configuration status %s not recognized."
+                % domain_configuration_status
+            )
+        if domain_type not in ["ENDPOINT", "AWS_MANAGED", "CUSTOMER_MANAGED"]:
+            raise InvalidRequestException(
+                " An error occurred (InvalidRequestException) when calling the DescribeDomainConfiguration "
+                "operation: Domain type %s not recognized." % domain_type
+            )
+        # Handle None service type (is it possible)?
+        if service_type and service_type not in ["DATA", "CREDENTIAL_PROVIDER", "JOBS"]:
+            raise InvalidRequestException(
+                " An error occurred (InvalidRequestException) when calling the DescribeDomainConfiguration "
+                "operation: Service type %s not recognized." % service_type
+            )
+        self.domain_configuration_name = domain_configuration_name
+        self.domain_configuration_arn = "arn:aws:iot:%s:1:domainconfiguration/%s/%s" % (
+            region_name,
+            domain_configuration_name,
+            random_string(5),
+        )
+        self.domain_name = domain_name
+        self.server_certificates = []
+        if server_certificate_arns:
+            for sc in server_certificate_arns:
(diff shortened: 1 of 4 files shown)


moto/iot/models.py
```diff
@@ -484,6 +484,74 @@ def to_dict(self):
...
+        region_name,
+        domain_configuration_name,
+        domain_name,
+        server_certificate_arns,
+        domain_configuration_status,
+        service_type,
+        authorizer_config,
+        domain_type,
+    ):
+        if domain_configuration_status not in ["ENABLED", "DISABLED"]:
```
bblommers, line 499: This branch doesn't seem to be reachable, as we're always passing in a hardcoded `ENABLED` when creating an instance

moto/iot/models.py
```diff
@@ -484,6 +484,74 @@ def to_dict(self):
...
+            raise InvalidRequestException(
+                " An error occurred (InvalidRequestException) when calling the DescribeDomainConfiguration "
+                "operation: Domain configuration status %s not recognized."
+                % domain_configuration_status
+            )
+        if domain_type not in ["ENDPOINT", "AWS_MANAGED", "CUSTOMER_MANAGED"]:
+            raise InvalidRequestException(
```
bblommers, line 506: Same here

moto/iot/models.py
```diff
@@ -484,6 +484,74 @@ def to_dict(self):
...
+                " An error occurred (InvalidRequestException) when calling the DescribeDomainConfiguration "
+                "operation: Domain type %s not recognized." % domain_type
+            )
+        # Handle None service type (is it possible)?
+        if service_type and service_type not in ["DATA", "CREDENTIAL_PROVIDER", "JOBS"]:
```
bblommers, line 511: There should be a test for this branch

tests/test_iot/test_iot.py
```diff
@@ -2346,3 +2346,115 @@ def test_update_thing_with_deprecated_type(self):
...
+
+    # Creation with only the name should succeed, creating same name fails
+    domain_config_1 = client.create_domain_configuration(
+        domainConfigurationName="testConfig1"
+    )
+    domain_config_1.should.have.key("domainConfigurationName").which.should.equal(
+        "testConfig1"
+    )
+    domain_config_1.should.have.key("domainConfigurationArn").which.should_not.be.none
+    with pytest.raises(client.exceptions.ResourceAlreadyExistsException):
```
bblommers, line 2363: With negative tests, we always have to verify the format is correct:
```
with pytest.raises(..) as exc:
    ...
err["Code"].should.equal("..")
err["Message"].should.equal("..")
```

bblommers (changes requested):
Hi @llawford, thanks for the PR!
Would you mind splitting  up the tests in multiple methods? Ideally one scenario per test, i.e. `test_create..`, `test_list..`, `test_create_with_invalid_params`, etc..

moto/iot/models.py
llawford: You're right; domain configurations seem to always start as ENABLED and the domain type can't be set by the user, AWS sets it automatically based on the domain, which is why I hardcoded them. I'll remove these two checks, they can be added in the future if these parameters are allowed to be set somehow.

llawford:
@bblommers Sorry for the multiple fix commits. Let me know if I'm okay to squash these commits at some point, not sure if the merge process does that automatically here.

bblommers:
> @bblommers Sorry for the multiple fix commits. Let me know if I'm okay to squash these commits at some point, not sure if the merge process does that automatically here.

All good - we'll squash them anyway when merging!

Happy to merge once the CI passes - it looks like black is stil failing atm.

I see you already did some linting - keep in mind that different versions of black will give different results. We're using `black==19.10b0`, so only linting with that specific version guarantees that the build will pass.

bblommers (approved):
LGTM! Thanks again @llawford

PR #4649, opened 2021-12-03
ECS - Fix tag/untag for with multiple services in cluster

dominikschubert:
Due to a incorrectly indented return statement, the loop only ever ran once, so if there are multiple services in an ECS cluster, only the first one was ever considered when adding/removing tags.


bblommers (approved):
Nice catch @dominikschubert - thanks for the PR!

PR #4650, opened 2021-12-03
Paginator - make development easier + docs

### LINKED ISSUE #4584: pagination_decorator not passing kwargs to function
I might be interpreting this incorrectly or maybe I missing something:

`pagination_decorator` (`moto/utilities/paginator.py`) pops the kwargs (used for the pagination model) for the function that it is decorating before invoking the function.  This means that the values for those named arguments cannot be validated within the function because those arguments will no longer have a value.

The named arguments could be validated in `responses.py` before the decorator and function are invoked.  But if the other functions are processing parameters in `models.py`, then it's inconsistent.

bblommers:
Improvements to the `@paginate`-decorator:

 - Improve documentation
 - Rename `page_ending_range_keys` to `unique_attribute`, as that is more readable
 - Allow option to pass a string as a `unique_attribute` (instead of always having a list)
 - Allow passing an Exception on `fail_on_invalid_token`, so that custom exception will be thrown
 - Make `max_results` and `next_token` parameters optional in the model
 - Fixes a bug where model-arguments could not have a list as default arg (`def model(arg=[])`)

Closes #4584 

FYI @kbalk - this does not yet `check for a limit that exceeds the allowable value for limit`. I couldn't find an example of that use case, and I didn't want to pollute this PR with adding additional service-specific logic.

### DIFF
--- a/docs/docs/contributing/development_tips/utilities.rst
+++ b/docs/docs/contributing/development_tips/utilities.rst
@@ -59,10 +59,21 @@ See the below example how it works in practice:
                 # The default limit of the above parameter is not provided
                 "limit_default": 100,
                 #
-                # The collection of keys/attributes that should guarantee uniqueness for a given resource.
+                # One or more attributes  that guarantee uniqueness for a given resource.
                 # For most resources it will just be an ID, or ARN, which is always unique.
-                # In order to know what is the last resource that we're sending back, an encoded version of these attributes is used as the NextToken.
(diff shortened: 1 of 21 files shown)


kbalk:
I think it was InvalidParameterException if MaxResults is greater than the allowed value.  I think some APIs ignore a value outside of the range and some report it.  And I'm not sure if it's always that particular exception.

Thank you for making these changes ... it certainly complicates the paginator code, but the caller's code is cleaner.

PR #4651, opened 2021-12-03
Add lambda capability for resourcegroupstaggingapi

Eun:
This pull request adds lambda capability for the resourcegroupstaggingapi.

Notice that unlike other aws entities, lambda uses another tagging method, therefore its required to transform the tags before applying the filters.

all tests pass & no linter complaints


bblommers (approved):
LGTM! Thanks for the PR @Eun

PR #4657, opened 2021-12-05
Techdebt - remove dependency on deprecated distutils

bblommers:
Distutils is deprecated in Python3.10, and will be removed in Python3.12.

The recommendation is tto replace this with either `packaging` or `pkg_resources`, but considering we only need the one class (`LooseVersion`), I reckon it's easier to just package it with Moto for now.

This version is used in one place, to check the version of the `responses`-module and patch it accordingly.

A future improvement would be to:
 - Set a lower limit on the `responses`-dependency, so that we only need to support the latest version
 - Register a custom matcher with the `responses`-module - the moment they have a registry implemented. (They are currently working on something like that, and I can't wait to use it: https://github.com/getsentry/responses/pull/416)

### DIFF
--- a/moto/core/models.py
+++ b/moto/core/models.py
@@ -19,7 +19,6 @@ except ImportError:
 from botocore.config import Config
 from botocore.handlers import BUILTIN_HANDLERS
 from botocore.awsrequest import AWSResponse
-from distutils.version import LooseVersion
 from http.client import responses as http_responses
 from urllib.parse import urlparse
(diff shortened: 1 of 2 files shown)


beliaev-maksim:
@bblommers 
we have released `responses` 0.17 with custom registry :)

check all changes here: https://github.com/getsentry/responses/releases/tag/0.17.0

bblommers:
Awesome - thanks for letting me know @beliaev-maksim!

Issue #1108, opened 2017-09-05
switch server mode to proxies


thehesiod:
botocore just added support for specifying proxy servers.  It seems like it would be better to switch to using a single proxy, then specifying the endpoint_url, as it would require fewer changes on the client, and could support multiple services from a single endpoint.  Could also solve the url parsing issue with aws region(s).

thoughts?

JackDanger:
@thehesiod This sounds like a great idea to me.

thehesiod:
based on research in https://github.com/spulec/moto/issues/1317 it seems like one could either support this on the client side via Config.proxies, or `http[s]_proxy` environment variables.

terrycain:
Hmm, I "think" if we do https_proxy, we'd need to serve moto on https, and also trust whatever CA we're using in lambdas. If we only listen on http, we'll need to patch whatever library in use to default use_ssl=False

jacobisaliveandwell:
If moto exposes the ability to invoke a lambda locally, it should automatically proxy/patch the container. It's pretty much a guarantee that the user will be attempting to contact mocked services from the lambda code.

bblommers:
moto >= 4.2.5.dev12 now contains a proxy!

It contains a SSL certificate generator, but for SDK's to trust these certificates, they need to be configured to trust the certificate bundle that comes with Moto.
Alternatively, most SDK's support some version of `verify=False`.

Documentation on how to get started and configure this all:
http://docs.getmoto.org/en/latest/docs/proxy_mode.html

b-guerra:
@bblommers , that's great news!

I am trying to invoke a Lambda function to access my mock dynamoDB, so in order to do so, I am using the moto_proxy
It is starting without any problems:

```
$ moto_proxy -H 127.0.0.1
Call `moto_proxy -h` for example invocations
Serving HTTP Proxy on 127.0.0.1:5005 ..

```

And whenever I call my mock tests, it creates the dynamoDB and lambda function without issues, however, whenever I try to reach the dynamoDB via the proxy server, it just keeps hanging and hanging, and nothing ever reaches my server. Code snippet below:

```
session = boto3.session.Session(
    aws_access_key_id='mock',
    aws_secret_access_key='mock',
    aws_session_token='us-east-1',
)

url = "http://127.0.0.1:5005"
config = Config(proxies={"https": url})
dynamodb = session.resource('dynamodb', config=config, verify=False)
table = dynamodb.Table('MyTable')

```

[...]

PR #4658, opened 2021-12-05
Propagate MotoHost via env vars to Lambda

bblommers:
FYI @kbalk - this adds the env vars to the Lambda Docker container per your suggestion.

Marking it as draft for now, as the tests will probably fail..

### DIFF
--- a/.github/workflows/dockertests.yml
+++ b/.github/workflows/dockertests.yml
@@ -1,0 +1,93 @@
+name: DockerTests
+
+on: [push, pull_request]
+
+jobs:
+  cache:
+    name: Caching
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: [ 3.9 ]
+
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v2
+        with:
+          python-version: ${{ matrix.python-version }}
+      - name: Get pip cache dir
+        id: pip-cache-dir
+        run: |
+          echo "::set-output name=dir::$(pip cache dir)"
+      - name: pip cache
+        id: pip-cache
+        uses: actions/cache@v2
+        with:
+          path: ${{ steps.pip-cache-dir.outputs.dir }}
+          key: pip-${{ matrix.python-version }}-${{ hashFiles('**/setup.py') }}
+      - name: Update pip
+        if: ${{ steps.pip-cache.outputs.cache-hit != 'true' }}
+        run: |
+          python -m pip install --upgrade pip
+      - name: Install project dependencies
+        if: ${{ steps.pip-cache.outputs.cache-hit != 'true' }}
+        run: |
+          pip install -r requirements-dev.txt
+
+  test:
+    name: Test Docker configurations
+    runs-on: ubuntu-latest
+    needs: cache
+    strategy:
+      matrix:
+        python-version: [3.9]
+
+    steps:
+    - uses: actions/checkout@v2
+      with:
+        fetch-depth: 0
+    - name: Set up Python ${{ matrix.python-version }}
+      uses: actions/setup-python@v2
+      with:
+        python-version: ${{ matrix.python-version }}
+    - name: Start MotoServer on an unusual port
+      run: |
+        python setup.py sdist
+        docker run --rm -t --name motoserver -e TEST_SERVER_MODE=true -e AWS_SECRET_ACCESS_KEY=server_secret -e MOTO_PORT=4555 -e AWS_ACCESS_KEY_ID=server_key -v `pwd`:/moto -p 4555:5000 -v /var/run/docker.sock:/var/run/docker.sock python:3.7-buster /moto/scripts/ci_moto_server.sh &
+        python scripts/ci_wait_for_server.py
+    - name: Get pip cache dir
+      id: pip-cache
+      run: |
+        echo "::set-output name=dir::$(pip cache dir)"
+    - name: pip cache
+      uses: actions/cache@v2
+      with:
+        path: ${{ steps.pip-cache.outputs.dir }}
+        key: pip-${{ matrix.python-version }}-${{ hashFiles('**/setup.py') }}
+    - name: Update pip
+      run: |
+        python -m pip install --upgrade pip
+    - name: Install project dependencies
+      run: |
+        pip install -r requirements-dev.txt
+    - name: Test
+      env:
+        TEST_SERVER_MODE: ${{ true }}
+      run: |
+        pytest -sv tests/test_awslambda tests/test_cloudformation
+    - name: "Stop MotoServer"
+      if: always()
+      run: |
+        mkdir serverlogs
+        pwd
+        ls -la
+        cp server_output.log serverlogs/server_output.log
+        docker stop motoserver
+    - name: Archive TF logs
+      if: always()
+      uses: actions/upload-artifact@v2
+      with:
+        name: motoserver-${{ matrix.python-version }}
+        path: |
+          serverlogs/*
--- a/moto/awslambda/models.py
+++ b/moto/awslambda/models.py
@@ -568,6 +568,8 @@ class LambdaFunction(CloudFormationModel, DockerModel):
             }
 
             env_vars.update(self.environment_vars)
+            env_vars["MOTO_HOST"] = settings.moto_server_host()
+            env_vars["MOTO_PORT"] = settings.moto_server_port()
 
             container = exit_code = None
             log_config = docker.types.LogConfig(type=docker.types.LogConfig.types.JSON)
--- a/moto/settings.py
+++ b/moto/settings.py
@@ -51,12 +51,11 @@ def moto_server_port():
 
 
 def moto_server_host():
-    _port = moto_server_port()
     if is_docker():
         host = get_docker_host()
     else:
         host = "http://host.docker.internal"
-    return f"{host}:{_port}"
+    return f"{host}"
 
 
 def is_docker():
--- a/tests/test_awslambda/test_lambda_invoke.py
+++ b/tests/test_awslambda/test_lambda_invoke.py
@@ -13,12 +13,14 @@ from moto import (
     settings,
 )
 from uuid import uuid4
+from unittest import SkipTest
 from .utilities import (
     get_role_name,
     get_test_zip_file_error,
     get_test_zip_file1,
     get_zip_with_multiple_files,
     get_test_zip_file2,
+    get_lambda_using_environment_port,
 )
 
 _lambda_region = "us-west-2"
@@ -143,6 +145,32 @@ def test_invoke_event_function():
     )
     success_result["StatusCode"].should.equal(202)
     json.loads(success_result["Payload"].read().decode("utf-8")).should.equal(in_data)
+
+
+@pytest.mark.network
+@mock_lambda
+def test_invoke_lambda_using_environment_port():
+    if not settings.TEST_SERVER_MODE:
+        raise SkipTest("Can only test environment variables in server mode")
+    conn = boto3.client("lambda", _lambda_region)
+    function_name = str(uuid4())[0:6]
+    conn.create_function(
+        FunctionName=function_name,
+        Runtime="python3.7",
+        Role=get_role_name(),
+        Handler="lambda_function.lambda_handler",
+        Code={"ZipFile": get_lambda_using_environment_port()},
+    )
+
+    success_result = conn.invoke(
+        FunctionName=function_name, InvocationType="Event", Payload="{}"
+    )
+
+    success_result["StatusCode"].should.equal(202)
+    response = success_result["Payload"].read()
+    functions = json.loads(response.decode("utf-8"))["response"]
+    functions.should.have.length_of(1)
+    functions[0]["FunctionName"].should.equal(function_name)
 
 
 @pytest.mark.network
--- a/tests/test_awslambda/utilities.py
+++ b/tests/test_awslambda/utilities.py
@@ -48,6 +48,24 @@ def lambda_handler(event, context):
     return _process_lambda(func_str)
 
 
+def get_lambda_using_environment_port():
+    func_str = """
+import boto3
+import os
+
+def lambda_handler(event, context):
+    base_url = os.environ.get("MOTO_HOST")
+    _port = os.environ.get("MOTO_PORT")
+    url = base_url + ":" + _port
+    conn = boto3.client('lambda', region_name='us-west-2', endpoint_url=url)
+
+    functions = conn.list_functions()["Functions"]
+
+    return {'response': functions}
+"""
+    return _process_lambda(func_str)
+
+
 def get_test_zip_file3():
     pfunc = """
 def lambda_handler(event, context):


kbalk:
♥️ You're my hero, Bert!  

There's still documentation needed so people can use this and I still need to debug the issue of running that lambda within another docker container.  I don't know if it's because the docker container doesn't support a nested container or what.  

Also, the issue https://github.com/spulec/moto/issues/1108 might not be resolved by this PR, but it might be useful to link this PR to it because it's related.

bblommers:
> 
> I still need to debug the issue of running that lambda within another docker container. I don't know if it's because the docker container doesn't support a nested container or what.
> 

I just ran into an issue where I could only execute a Lambda as long as the host container was called `motoserver` - something to do with the links-argument here: https://github.com/spulec/moto/blob/master/moto/awslambda/models.py#L578

kbalk:
> I just ran into an issue where I could only execute a Lambda as long as the host container was called `motoserver` - something to do with the links-argument here: https://github.com/spulec/moto/blob/master/moto/awslambda/models.py#L578

Wow, you must have to dig to find that!   I got a "volume busy" error, which is why I was thinking it was a nesting issue.  I'm also concerned that once I get the lambda running, I might need something like LocalStack's DOCKER_BRIDGE_IP as the hostname to reach the moto server.

I've got to put in some regular work hours today, so I won't be able to debug until much later tonight.

kbalk:
I don't get the "volume busy" message and the lambda actually tries to execute.  I don't really have the right testing environment because I'm not going into the docker container to replace the version of moto installed by LocalStack.  However, I think what I'm seeing correlates with what you were saying about the container name?

When I run my test, the container name is "localstack", which is the value we assign to LocalStack's environment variable `MAIN_CONTAINER_NAME` in the docker compose file:

```
services:
  localstack:
    container_name: localstack
    image: localstack/localstack
    ports:
      - "4566:4566"
      - "4615:4615"
      - "4571:4571"
    environment:
      - MAIN_CONTAINER_NAME=localstack.   <---------------------------------
      - LAMBDA_DOCKER_NETWORK=terraform_pytest_default
    volumes:
      - "/var/run/docker.sock:/var/run/docker.sock"
      - "${TERRAFORM_PYTEST_DIR}/initaws.d:/docker-entrypoint-initaws.d"
```

LocalStack's README defines that as:

```MAIN_CONTAINER_NAME: Specify the main docker container name (default: `localstack_main`).```

We use that same name to set the host value in the provider file.  So the lambda endpoint would be http://localstack:4615, where 4615 is the moto port.

When the container is spun up and the Terraform test is executed within that container, the test fails with the Terraform message:

```
module.success.aws_cloudformation_stack.cfn: Refreshing state... [id=arn:aws:cloudformation:us-east-1:123456789:stack/ami-lookup-test-complex-search/aadc7ef3-c402-4fd4-97e0-4a11bb2b5ecf]
 
Error: RequestError: send request failed
caused by: Post "http://localstack:4615/": dial tcp: lookup localstack on 127.0.0.11:53: no such host

  with module.success.aws_cloudformation_stack.cfn,
  on ../modules/cfn/main.tf line 1, in resource "aws_cloudformation_stack" "cfn":
   1: resource "aws_cloudformation_stack" "cfn" {
```

The moto log events for the test show a status of SUCCESS from the lambda, then:

```
event_id: 13 	ingestion_time: 1638761914517 	message: send(..) failed executing requests.put(..): Invalid URL 'http://:4615/cloudformation_us-east-1/cfnresponse?stack=arn:aws:cloudformation:us-east-1:123456789:stack/ami-lookup-test-complex-search/aadc7ef3-c402-4fd4-97e0-4a11bb2b5ecf': No host supplied 	timestamp: 1638761914517.552
event_id: 14 	ingestion_time: 1638761914517 	message: 2021-12-06T03:38:33.997Z [lambda][INFO ]: found ami: ami-760aaa0f 	timestamp: 1638761914517.553
event_id: 15 	ingestion_time: 1638761914517 	message: [ERROR] InvalidURL: Invalid URL 'http://:4615/cloudformation_us-east-1/cfnresponse?stack=arn:aws:cloudformation:us-east-1:123456789:stack/ami-lookup-test-complex-search/aadc7ef3-c402-4fd4-97e0-4a11bb2b5ecf': No host supplied 	timestamp: 1638761914517.554
event_id: 16 	ingestion_time: 1638761914517 	message: Traceback (most recent call last): 	timestamp: 1638761914517.555
event_id: 17 	ingestion_time: 1638761914517 	message:   File "/var/task/lambda.py", line 105, in handler
```

Is there a way we could pass the container name to moto?  We pass the port number to the server, could we pass a container name?

Also, I just realized that the test for "motoserver" you referenced is just a few lines down from where the env vars was set.  So not so hard to find, but I didn't recognize the significance of that test.

**Comment by bblommers** [comment]
> Is there a way we could pass the container name to moto? We pass the port number to the server, could we pass a container name?
>

Yeah, that makes sense, using the same mechanism. I'm not sure why it wouldn't have a host though, it should always default to `host.docker.internal` if nothing else

**Comment by kbalk** [comment]
> Yeah, that makes sense, using the same mechanism. I'm not sure why it wouldn't have a host though, it should always default to `host.docker.internal` if nothing else

I'll try to debug it tonight or see what more I can figure out.

**Comment by kbalk** [comment]
I should have provided more of the logs in my previous comment.  Here's a snippet that's before the snippet I previously sent you:

```
event_id: 7 | ingestion_time: 1638857984012 | message:  http://:4615/cloudformation_us-east-1/cfnresponse?stack=arn:aws:cloudformation:us-east-1:123456789:stack/ami-lookup-test-complex-search/ef811864-a198-49d5-a810-0b20d48d729c | timestamp: 1638857984011.893
-- | -- | -- | --
event_id: 8 | ingestion_time: 1638857984012 | message: Response body: | timestamp: 1638857984011.894
event_id: 9 | ingestion_time: 1638857984012 | message: {"Status": "SUCCESS", "Reason":  "See the details in CloudWatch Log Stream:  2021/12/07/[$LATEST]c88f33bc6ba4ee9795e5c18c17531c38",  "PhysicalResourceId":  "2021/12/07/[$LATEST]c88f33bc6ba4ee9795e5c18c17531c38", "StackId":  "arn:aws:cloudformation:us-east-1:123456789:stack/ami-lookup-test-complex-search/ef811864-a198-49d5-a810-0b20d48d729c",  "RequestId": "0eda33f8-d72b-4709-acfb-8aee1f962000",  "LogicalResourceId": "AmiIdLookup", "NoEcho": false, "Data": {"Data":  {"Filters":  "[{\"Name\":\"name\",\"Values\":[\"amzn-ami-hvm-*\"]},{\"Name\":\"description\",\"Values\":[\"Amazon  Linux  AMI*\"]},{\"Name\":\"architecture\",\"Values\":[\"x86_64\"]},{\"Name\":\"root-device-type\",\"Values\":[\"ebs\"]}]",  "Owners": "[\"amazon\"]", "Region": "us-east-1", "ServiceToken":  "arn:aws:lambda:us-east-1:123456789012:function:test-cfn-lookup-ami-ids-49"},  "Id": "ami-760aaa0f"}} | timestamp: 1638857984011.896
```

The above snippet is coming from the print statements in `moto/packages/cfnresponse/cfnresponse.py` and it shows that the `send()` function was passed a bad url in the event.  

The only place where "cfnresponse" would be put in a URL is in `moto/cloudformation/custom_model.py` in the function `create_from_cloudformation_json()`:

```
        host = f"{settings.moto_server_host()}:{settings.moto_server_port()}"
        response_url = (
            f"{host}/cloudformation_{region_name}/cfnresponse?stack={stack_id}"
        )
```

I added print statements to show the output of `settings.moto_server_host()` and `settings.get_docker_host()`, but because moto is running in a docker container, I can't see the print output.  The modified moto is only being pulled in when the lambda runs.

So I added a new entry to the event that's passed to `send()`.  That new entry contains the output of `settings.get_docker_host()` and I print that from `send()`, but it's not showing up on the dashboard.  It should, so I don't know what I'm doing wrong.

**Comment by bblommers** [comment]
FYI @lorengordon I've added the option to change `network_mode`, per the discussion in #4638
https://github.com/spulec/moto/pull/4658/files#diff-908971deae8555292db0b80849241c2820db44529d71374f90212ed027c54f87R1133-R1144

**Comment by lorengordon** [comment]
I see, thanks @bblommers! I'm not entirely sure if setting `network_mode=host` will work for our use case, with a user-defined network bridge, but I suppose it might. Hadn't thought of that. I was more thinking that we would pass `network=<user-defined-bridge>` to connect the lambda container directly to the network bridge...

**Comment by kbalk** [comment]
This is a ridiculously hard thing to debug.  I'll give it one last try tomorrow.  I tried "updating" the container running LocalStack and moto, but the test then hung.

In `moto/awslambda/models.py` (arround line 1137) there is no longer a mention of MOTO_PORT. Instead the example shows a hard coded value for the port. Is there a reason you did that?

**Comment by bblommers** [comment]
> This is a ridiculously hard thing to debug. I'll give it one last try tomorrow. I tried "updating" the container running LocalStack and moto, but the test then hung.
> 

Just a suggestion - is it easier to create testreleases? You could checkout the branch, change the projectname to 'myfirstmotoproject' and push your own release to PyPi, and then incorporate that release as a normal dependency. It could speed up the debug/release/test cycle.

> In `moto/awslambda/models.py` (arround line 1137) there is no longer a mention of MOTO_PORT. Instead the example shows a hard coded value for the port. Is there a reason you did that?

The first example still uses MOTO_PORT, only the second example uses a hardcoded value, to remind people that this option always exists

**Comment by kbalk** [comment]
> Just a suggestion - is it easier to create testreleases? You could checkout the branch, change the projectname to 'myfirstmotoproject' and push your own release to PyPi, and then incorporate that release as a normal dependency. It could speed up the debug/release/test cycle.

The problem is LocalStack running within the docker container.  It's pulling in a specific version of moto, which is not the version I want it to use.  I'm going to have to look at how it starts up and see if I can insert the 6 files you changed (one needs a modification for a file it imports) and restart moto.  When I run LocalStack and moto from the command line I can usually restart moto without restarting LocalStack and it works fine.

It sounds terrible (as in there should be a better way), but it might be easier to wait until LocalStack uses a version of moto with this change and then see if I can get it to do what I need it to do.

> The first example still uses MOTO_PORT, only the second example uses a hardcoded value, to remind people that this option always exists

Ah ... I was looking at the individual commit, not the differences in their entirety.

**Comment by lorengordon** [comment]
>The problem is LocalStack running within the docker container. It's pulling in a specific version of moto, which is not the version I want it to use.

@kbalk perhaps we can update the init script that starts the moto server to optionally install moto? If you fork localstack/moto, merge these changes, then you could install using `python -m pip install git+https://github.com/kbalk/<forkedrepo>.git@<branch>`

**Comment by kbalk** [comment]
> > The problem is LocalStack running within the docker container. It's pulling in a specific version of moto, which is not the version I want it to use.
> 
> @kbalk perhaps we can update the init script that starts the moto server to optionally install moto? If you fork localstack/moto, merge these changes, then you could install using `python -m pip install git+https://github.com/kbalk/<forkedrepo>.git@<branch>`

LocalStack is using their own fork of moto and I noticed a change to an import when I was diffing the files I wanted to add to the docker container.  So maybe they pull in the latest released version of moto, yet make changes to it?   Since I'm not really using LocalStack except to hook up the endpoints, I suppose I could just drop in a new version of moto like I would from a command line?  I'll do a little digging and experimenting.

**Comment by kbalk** [comment]
I stopped worrying about what LocalStack was doing with moto and that made things much simpler.  I modified the `initaws.d/start.sh` script as @lorengordon suggested, but I commented out the line to start moto.  I copied the moto wheel to the LocalStack docker container, installed flask, flask-cors and the docker python module, then manually started up moto_server.  So now I can run moto_server from the command line in the docker container.

moto is crashing in the code I mentioned before (`moto/cloudformation/custom_model.py`).  `moto_server_host()` is calling `get_docker_host()` and the ip value is null.  The IPAddress we want is nested under the info for the docker network bridge we named "terraform_pytest_default"

```
root@920c40d6de2f:/opt/code/localstack# python
Python 3.8.12 (default, Nov 17 2021, 17:26:17) 
[GCC 8.3.0] on linux
>>> import os
>>> import json
>>> cmd = "curl -s --unix-socket /run/docker.sock http://docker/containers/$HOSTNAME/json"
>>> container_info = os.popen(cmd).read()
>>> _ip = json.loads(container_info)["NetworkSettings"]["IPAddress"]
>>> _ip
''
>>> from pprint import pprint
>>> pprint(json.loads(container_info)["NetworkSettings"])
{'Bridge': '',
 'EndpointID': '',
 'Gateway': '',
 'GlobalIPv6Address': '',
 'GlobalIPv6PrefixLen': 0,
 'HairpinMode': False,
 'IPAddress': '',
 'IPPrefixLen': 0,
 'IPv6Gateway': '',
 'LinkLocalIPv6Address': '',
 'LinkLocalIPv6PrefixLen': 0,
 'MacAddress': '',
 'Networks': {'terraform_pytest_default': {'Aliases': ['localstack',
                                                       '920c40d6de2f'],
                                           'DriverOpts': None,
                                           'EndpointID': '2bda0e66f8dd0f06c3e3e4357e805123fc1324459c9759466a54233d8e75f230',
                                           'Gateway': '172.21.0.1',
                                           'GlobalIPv6Address': '',
                                           'GlobalIPv6PrefixLen': 0,
                                           'IPAMConfig': None,
                                           'IPAddress': '172.21.0.2',
                                           'IPPrefixLen': 16,
                                           'IPv6Gateway': '',
                                           'Links': None,
                                           'MacAddress': '02:42:ac:15:00:02',
                                           'NetworkID': '4ae61ec093c553c58d2af51989898d2220af9b5c6832695e3fd715b1eb4152c6'}},
 'Ports': {'4566/tcp': [{'HostIp': '0.0.0.0', 'HostPort': '4566'},
                        {'HostIp': '::', 'HostPort': '4566'}],
           '4571/tcp': [{'HostIp': '0.0.0.0', 'HostPort': '4571'},
                        {'HostIp': '::', 'HostPort': '4571'}],
           '4615/tcp': [{'HostIp': '0.0.0.0', 'HostPort': '4615'},
                        {'HostIp': '::', 'HostPort': '4615'}],
           '5678/tcp': None},
 'SandboxID': '5563bbac505f055e76872f8ccdf24b969815effc22b567a05510d1e5d4dd41cd',
 'SandboxKey': '/var/run/docker/netns/5563bbac505f',
 'SecondaryIPAddresses': None,
 'SecondaryIPv6Addresses': None}
>>> json.loads(container_info)["NetworkSettings"]["Networks"]["terraform_pytest_default"]["IPAddress"]
'172.21.0.2'
```

**Comment by kbalk** [comment]
@bblommers I tested the following code change, but if you write it differently, I can retest it for you.  I'm not sure how to the handle the case where they give you a bad network name.  It might be better to raise an exception if the network name can't be found in the networks versus ignore it.  I'd wouldn't want to have to debug a misspelled network name!

moto is still crashing in `moto/cloudformation/parsing.py` due to bad json.  I'm debugging that now.

```
 def get_docker_host():
     try:
         cmd = "curl -s --unix-socket /run/docker.sock http://docker/containers/$HOSTNAME/json"
         container_info = os.popen(cmd).read()
-        _ip = json.loads(container_info)["NetworkSettings"]["IPAddress"]
+        network_settings = json.loads(container_info)["NetworkSettings"]
+        network_name = moto_network_mode()
+        if network_name and network_name in network_settings["Networks"]:
+            _ip = network_settings["Networks"][network_name]["IPAddress"]
+        else:
+            _ip = network_settings["IPAddress"]
         return f"http://{_ip}"
     except:  # noqa
         return "http://host.docker.internal"
```

**Comment by kbalk** [comment]
The json problem is timing dependent.  I set a breakpoint and stepped through the code and it didn't crash.  I removed the breakpoint, added more print statements and it didn't crash.  It's related to the AmiIdLookup.Id value not being available, I think.  Here's the output of the last set of print statements when everything is working:

```
Resource json: 
{'Description': 'The target AMI Id',
 'Value': {'Fn::If': ['UseAmiLookup',
                      {'Fn::Sub': '${AmiIdLookup.Id}'},
                      'AmiId']}}
Resource json: 
'The target AMI Id'
Resource json: 
{'Fn::If': ['UseAmiLookup', {'Fn::Sub': '${AmiIdLookup.Id}'}, 'AmiId']}
Resource json: 
{'Fn::Sub': '${AmiIdLookup.Id}'}
Resource json: 
'${AmiIdLookup.Id}'
Resource json: 
{'Fn::GetAtt': ['AmiIdLookup', 'Id']}
Fn::GetAtt case
['AmiIdLookup', 'Id']
Resource json: 
'ami-760aaa0f'
```

The second problem that remains is that while the IP is now being correctly determined in the LocalStack/moto container, a call to `moto_network_mode()` in the lambda container returns an empty string, so I can't use it to determine the host for the endpoint_url.

**Comment by bblommers** [comment]
> @bblommers I tested the following code change, but if you write it differently, I can retest it for you. I'm not sure how to the handle the case where they give you a bad network name. It might be better to raise an exception if the network name can't be found in the networks versus ignore it. I'd wouldn't want to have to debug a misspelled network name!
> 

Thanks @kbalk - I pushed your change, with a print-statement for now to indicate if a network name cannot be found.

Note that there are two different env variables now: `MOTO_DOCKER_NETWORK_NAME` and `MOTO_DOCKER_NETWORK_MODE`. (I didn't realize they were different configuration options until I dived in a little deeper..)

It sounds like you'll need to set `MOTO_DOCKER_NETWORK_NAME`, and can ignore the network_mode.

**Comment by kbalk** [comment]
> It sounds like you'll need to set `MOTO_DOCKER_NETWORK_NAME`, and can ignore the network_mode.

Yup.  I was a bit confused about the mode option as I needed the IP address, which I could access via the network name.  I think I have what I need for the three types of environments we test in (no moto, moto from the command line, moto within a docker container).

I'm not sure what to do about the timing problem in moto/cloudformation/parsing.py.  Should it be a separate issue?   I'm not all that familiar with that code.

**Comment by bblommers** [comment]
> 
> I'm not sure what to do about the timing problem in moto/cloudformation/parsing.py. Should it be a separate issue? I'm not all that familiar with that code.

Yeah, let's make it a separate issue. It could very well be related, but it will be easier to track.

In a nutshell, this is the flow when calling create_stack with a custom resource:

1. Custom resource is created
2. Custom Lambda is invoked asyncronously
3. create_stack finishes, with a 'CREATE_IN_PROGRESS' status
4. The Lambda calls MotoServer with AmiIdLookup.Id value
5. Moto persists the value against the Custom Resource
6. Moto sets the stack_status to 'COMPLETE'

So when testing, we do have to wait until the stack status is complete, that's the only timing issue that I can see.
(I don't quite remember.. but I hope I pointed it out during development? Usually the stack is complete before the create_stack finishes, so this is unusual behaviour for Moto.)

**Comment by bblommers** [comment]
@bpandola Would you mind reviewing this?

TLDR: This makes it easier for people to control how any Docker Lambda's can connect to an existing MotoServer.
The new documentation in the LambdaBackend-class has a summary of the options that users now have.

**Comment by kbalk** [comment]
I was thinking about this some more and it would help if the network name was retrieved in `moto/awslambda/models.py` at the same time you pull in moto port and server, i.e.,:

```
 env_vars["MOTO_HOST"] = settings.moto_server_host()
 env_vars["MOTO_PORT"] = settings.moto_server_port()
 env_vars["MOTO_NETWORK_NAME"] = settings.moto_network_name()
```

**Comment by kbalk** [comment]
@bblommers Thinking about this even more, the reason I need the network name is because in a docker environment, instead of using the host name, I need the network name to calculate the host ip address.  So moto_host as it is now isn't reflective of the environment the lambda is running in.

**Comment by bblommers** [comment]
@kbalk I'm open to changing either/both. What specifically do you want to change in the logic?

**Comment by kbalk** [comment]
> @kbalk I'm open to changing either/both. What specifically do you want to change in the logic?

I'm so sorry, Bert.  I was looking at the logs I printed in the lambda and I was looking at the output of `moto_server_host()`, which prints out host.docker.internal.  I was not printing out the value of MOTO_HOST.  I wasn't thinking.  I apologize.

I didn't finish my tests as there was a power outage and by the time power was restored, my laptop battery had been drained so I need to set up my test environment again.  That's kind of the way things have been going the last couple of days.  This evening I'll give it another go, but I was able to verify that I was getting the expected value in MOTO_HOST when running the lambda in a docker container, so please ignore my comments about MOTO_NETWORK_NAME.  

Again, my apologies for suggesting I needed other changes when that is not the case.

**Comment by bblommers** [comment]
No worries @kbalk - hopefully debugging is going more smoothly now :crossed_fingers:

**Comment by kbalk** [comment]
@bblommers Our terraform tests pass in all three scenarios (with AWS, with `moto_server` running on the command line and with `moto_server` and the terraform test running in different docker containers).  

I'm really happy with this PR as the lambda doesn't require terraform variables to indicate the type of endpoint it must use for boto3 calls.  If MOTO_PORT is defined, the lambda knows how to build `endpoint_url` using MOTO_PORT and MOTO_HOST.  The lambda doesn't have to import moto.  Our docker compose file that runs LocalStack and moto now sets MOTO_DOCKER_NETWORK_NAME, so that value is set only when it needs to be.  It's a nice, clean solution.  I think others will find it useful.

Regarding the intermittent timing error related to cfnresponse - I saw it once when running `moto_server` from the command line, so it doesn't occur just when the test is running within a docker container.  If I refresh the dashboard around the time moto is interacting with the lambda, I can usually cause the timing error, but that's not a normal situation.  Otherwise, I can't reliably reproduce the problem and I don't have a neatly packaged test case for you.  I also don't have the time to work on it.  If it's okay with @lorengordon , I'd like to ignore the timing problem for now.  If you wish, I can still go ahead an create an issue, but as I said, I don't have a test case for you.

**Comment by lorengordon** [comment]
The timing issue with `cfnresponse` doesn't appear related to this feature, so I'd say defer that one to later also (not that I presume to have any influence over the moto project). Code here looks great. Thanks @bblommers!

**Comment by bblommers** [comment]
Awesome @kbalk, happy to hear that. If I dare say so - it sounds like you have an unusually complex setup, so if it works for you, it probably works for 90% of users that have a simpler setup. :slightly_smiling_face: 

I haven't been able to reproduce the timing error, so I'd be happy to ignore it. Do let us know if it pops up more often though, and i'll dig into it a little deeper.

**Comment by kbalk** [comment]
@bblommers This change looks like it's been forgotten ...

**Comment by bblommers** [comment]
@bpandola Would you mind reviewing this? 
The documentation that's part of this PR (in awslambda/models.py) is a decent summary of what this PR aims to achieve

**Comment by lorengordon** [comment]
One thing that kinda occurs to me, is that we could simplify the boilerplate for users, and just expose an env, `MOTO_HTTP_ENDPOINT` (name chosen to account for adding HTTPS separately at some point)... Effectively:

```
env_vars["MOTO_HTTP_ENDPOINT"] = f'{env_vars["MOTO_HOST"]}:env_vars["MOTO_PORT"]}'
```

And then the user's lambda setup is just:

```
ec2 = boto3.client('ec2', region_name='us-west-2', endpoint_url=os.environ.get("MOTO_HTTP_ENDPOINT")
```

bblommers:
> One thing that kinda occurs to me, is that we could simplify the boilerplate for users, and just expose an env, `MOTO_HTTP_ENDPOINT` (name chosen to account for adding HTTPS separately at some point)... Effectively:
> 
> ```
> env_vars["MOTO_HTTP_ENDPOINT"] = f'{env_vars["MOTO_HOST"]}:env_vars["MOTO_PORT"]}'
> ```
> 

Yes, that's a good point - I've added that env var as well.

Thank you both for your reviews and contributions here, @lorengordon and @kbalk!

PR #4666, opened 2021-12-06
S3 - copy ContentType and Metadata from source object

naspeh:
I have created a test case for https://github.com/localstack/localstack/issues/5068 and run it against AWS S3, `Metadata` and `ContentType` are copied over to a new object, so updating it in `moto` as well

### DIFF
--- a/moto/s3/models.py
+++ b/moto/s3/models.py
@@ -2067,6 +2067,7 @@ class S3Backend(BaseBackend, CloudWatchMetricProvider):
             lock_until=key.lock_until,
         )
         self.tagger.copy_tags(key.arn, new_key.arn)
+        new_key.set_metadata(key.metadata)
 
         if acl is not None:
             new_key.set_acl(acl)
--- a/tests/test_s3/test_s3.py
+++ b/tests/test_s3/test_s3.py
@@ -1043,6 +1043,27 @@ def test_copy_key_replace_metadata():
 
 
 @mock_s3
+def test_copy_key_with_metadata_boto3():
+    s3 = boto3.resource("s3", region_name=DEFAULT_REGION_NAME)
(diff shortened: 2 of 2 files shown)


bblommers (commented):
Hi @naspeh, looks like there are two use cases:
1) Metadata set on the original key is not copied across
2) Metadata as part of the copy-command is not correctly set

Moto currently doesn't do either.

This PR solves 1), but if I'm readying the original issue correctly, they are complaining about use case 2):
> The copied object should respect parameters set in the copy object command, such as content-type, metadata, and metadata-directive. For example, if the copy object command specifies that the object should have content-type "application/json", the resulting object should not have a content-type of "binary/octet-stream".
>

Personally I'm happy with this PR as is, as it's already an improvement!, but do you want to fix the second use case here as well?

naspeh:
I think moto supports `MetadataDirective="REPLACE"` https://github.com/spulec/moto/blob/805fbef4fadc3ec0c08bc080d99640422bbaaa25/tests/test_s3/test_s3.py#L1080-L1080

I will test the second scenario against AWS tomorrow and I will adjust this PR and let you know.

Thank you for the quick response.

naspeh:
@bblommers, I've just checked the second case against real AWS once again, and AWS will copy metadata from the original object. If you want to replace metadata, you should specify `--metadata-directive=REPLACE`. Also confirmed this with the docs: https://docs.aws.amazon.com/cli/latest/reference/s3/cp.html
> --metadata-directive
> ...
> Valid values are COPY and REPLACE. If this parameter is not specified, COPY will be used by default. 
> ...

It means my changes from this PR are enough.

bblommers:
Awesome @naspeh - thanks for checking!

bblommers (approved):

PR #4670, opened 2021-12-07
S3 - restore_object() should error on non archived class

interimadd:
While writing a test to restore an S3 object, I noticed that restore_object() of moto3 does not error on standard class objects.  
For example, bellow code end with InvalidObjectState error message in the case of using real S3.

```py
import boto3

BUCKET_NAME = "SOME_BUCKET"
KEY = "STANDAR_CLASS_FILE"

client = boto3.client('s3')
response = client.restore_object(Bucket=BUCKET_NAME, Key=KEY)
```

```bash
$ python restore_test.py
  File "restore_test.py", line 7, in <module>
    response = client.restore_object(Bucket=BUCKET_NAME, Key=KEY)
  File "C:\Users\TakahisaIshikawa\AppData\Local\Programs\Python\Python37\lib\site-packages\botocore\client.py", line 386, in _api_call
  File "C:\Users\TakahisaIshikawa\AppData\Local\Programs\Python\Python37\lib\site-packages\botocore\client.py", line 705, in _make_api_call
botocore.errorfactory.InvalidObjectState: An error occurred (InvalidObjectState) when calling the RestoreObject operation: Restore is not allowed for the object's current storage class
```

Although [the boto3 document](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3.html#S3.Client.restore_object) dose not say clearly, the correct behavior is to get an error when attempting to restore any class object other than Glacier or DeepArchive class.

### DIFF
--- a/moto/s3/responses.py
+++ b/moto/s3/responses.py
@@ -2085,6 +2085,8 @@ class ResponseObject(_TemplateEnvironmentMixin, ActionAuthenticatorMixin):
             es = minidom.parseString(body).getElementsByTagName("Days")
             days = es[0].childNodes[0].wholeText
             key = self.backend.get_object(bucket_name, key_name)
+            if key.storage_class not in ["GLACIER", "DEEP_ARCHIVE"]:
+                raise InvalidObjectState(storage_class=key.storage_class)
             r = 202
             if key.expiry_date is not None:
                 r = 200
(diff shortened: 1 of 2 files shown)


bblommers (approved):
LGTM! Thanks for the PR @interimadd

PR #4681, opened 2021-12-12
ELBV2 - Fix elbv2 modify_rule expects Conditions (#4677)

### LINKED ISSUE #4677: elbv2 modify_rule expects 'Conditions' key to be present in params
## Summary

elbv2 [modify_rule](https://github.com/spulec/moto/blob/master/moto/elbv2/responses.py#L358) wrongly expects to receive `Conditions` key in the params, failing if it isn't supplied.

in [aws boto3 documentation](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/elbv2.html#ElasticLoadBalancingv2.Client.modify_rule)  we have that the only required argument is the `RuleArn` and that `Any properties that you do not specify are unchanged.`

### Exception

```
self = <moto.elbv2.responses.ELBV2Response object at 0x1113b6d60>

    @amzn_request_id
    def modify_rule(self):
        rule_arn = self._get_param("RuleArn")
        params = self._get_params()
>       conditions = params["Conditions"]
E       KeyError: 'Conditions'

moto/elbv2/responses.py:358: KeyError
```

## Root cause

in [responses.py](https://github.com/spulec/moto/blob/master/moto/elbv2/responses.py#L358) we have 
```
    def modify_rule(self):
        rule_arn = self._get_param("RuleArn")
        params = self._get_params()
        conditions = params["Conditions"] # <- this will imply 'Conditions' key to be present
        actions = params.get("Actions", [])
        rules = self.elbv2_backend.modify_rule(
            rule_arn=rule_arn, conditions=conditions, actions=actions
        )
        ...
```

This fails when trying to modify a rule without passing any conditions to it with `elbv2_client.modify_rule(RuleArn="arn", Actions=[action1, action2])`

We can see in [models.py](https://github.com/spulec/moto/blob/master/moto/elbv2/models.py#L1166-L1167) that it's actually respecting the documentation and already skips conditions if they are empty:

```
    def modify_rule(self, rule_arn, conditions, actions):
        ...
        if conditions:
            rule.conditions = conditions
        if actions:
            rule.actions = actions
        return rule

```

### Suggested Fix
in [responses.py](https://github.com/spulec/moto/blob/master/moto/elbv2/responses.py#L358) do instead:
```
    def modify_rule(self):
        rule_arn = self._get_param("RuleArn")
        params = self._get_params()
        conditions = params.get("Conditions", []) # suggested change here
        actions = params.get("Actions", [])
        rules = self.elbv2_backend.modify_rule(
            rule_arn=rule_arn, conditions=conditions, actions=actions
        )
        ...
```

Happy to open a PR for it

hudelgado:
Fixes #4677

### DIFF
--- a/moto/elbv2/responses.py
+++ b/moto/elbv2/responses.py
@@ -355,7 +355,7 @@ class ELBV2Response(BaseResponse):
     def modify_rule(self):
         rule_arn = self._get_param("RuleArn")
         params = self._get_params()
-        conditions = params["Conditions"]
+        conditions = params.get("Conditions", [])
         actions = params.get("Actions", [])
         rules = self.elbv2_backend.modify_rule(
             rule_arn=rule_arn, conditions=conditions, actions=actions
--- a/tests/test_elbv2/test_elbv2.py
+++ b/tests/test_elbv2/test_elbv2.py
@@ -634,6 +634,64 @@ def test_create_rule_priority_in_use():
 
 @mock_elbv2
 @mock_ec2
+def test_modify_rule_conditions():
+    response, _, _, _, _, elbv2 = create_load_balancer()
+    load_balancer_arn = response.get("LoadBalancers")[0].get("LoadBalancerArn")
+
+    action = {
+        "Type": "redirect",
+        "RedirectConfig": {
+            "Protocol": "HTTPS",
+            "Port": "443",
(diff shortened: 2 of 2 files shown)


bblommers (approved):
Nice one - thanks for the PR @hudelgado!

PR #4686, opened 2021-12-13
Add pagination for list experiments/trials/components

bogdangi:


bblommers (changes requested):
Hi @bogdangi! Looks like this PR removes a bit more than it should - maybe a dodgy merge?

For implementing pagination, Moto already has a service-agnostic decorator that you should be able to reuse: see http://docs.getmoto.org/en/latest/docs/contributing/development_tips/utilities.html#paginator

bogdangi:
hi @bblommers thank you for your comment. I refactor to use paginator.
it looks better ;)

Issue #4691, opened 2021-12-15
Dedicated moto-logger?


Wurstnase:
Hi,

I'm trying to disable the logging for other modules outside my project. Unfortunately the expressions module for dynamodb2 is using the root logger. 

https://github.com/spulec/moto/blob/de9aa9a8e3b998902381a565e25b5da17960c7f7/moto/dynamodb2/parsing/expressions.py#L64

PR #4692, opened 2021-12-15
use dedicated logger for dynamodb2

Wurstnase:
fixes: #4691


bblommers (approved):
LGTM! Thanks for the fix @Wurstnase

Issue #4694, opened 2021-12-15
`IoTDataPlaneResponse.publish` fails to correctly pass payload to `IotDataPlaneBackend.publish` method


iggyzap:
Hi,

When trying to implement a unit test that validates that MQTT message was sent via `boto3.client('iot-data')` client to specific topic, I found out that `IoTDataPlaneResponse.publish` fails to pass down `payload` correctly to `IotDataPlaneBackend.publish` method.

I was able to mock IotDataPlaneBackend backend for interaction in interest and even record publish calls, but payload was empty.

This is due to the fact that aws IoT client posts payload as http post body, but `IoTDataPlaneResponse.publish` expects payload to be passed as parameter (code below):
```python
class IoTDataPlaneResponse(BaseResponse):
...
    def publish(self):
        topic = self._get_param("topic")
        qos = self._get_int_param("qos")
        payload = self._get_param("payload")
        self.iotdata_backend.publish(topic=topic, qos=qos, payload=payload)
        return json.dumps(dict())
```
Image from debugger:
<img width="773" alt="image" src="https://user-images.githubusercontent.com/4525023/146213115-d3d82245-1f82-405b-ba36-7d0cdb4e9cd4.png">

I would expect when payload was posted to an MQTT topic then it will become available and passed down as payload to `IoTDataPlaneResponse.publish` method.

Following version of moto is used:
```json
"moto": {
            "hashes": [
                "sha256:2b3fa22778504b45715868cad95ad458fdea7227f9005b12e522fc9c2ae0cabc",
                "sha256:79aeaeed1592a24d3c488840065a3fcb3f4fa7ba40259e112482454c0e48a03a"
            ],
            "index": "pypi",
            "version": "==1.3.14"
        }

```

bblommers:
Hi @iggyzap, thanks for raising this! Will mark it as a bug.

PR #4697, opened 2021-12-16
A first work on the fix for iot-data payload not being passed into the iot-data backend.

iggyzap:
https://github.com/spulec/moto/issues/4694

### DIFF
--- a/moto/iotdata/responses.py
+++ b/moto/iotdata/responses.py
@@ -43,6 +43,5 @@ class IoTDataPlaneResponse(BaseResponse):
     def publish(self):
         topic = self._get_param("topic")
         qos = self._get_int_param("qos")
-        payload = self._get_param("payload")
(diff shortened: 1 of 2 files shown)


bblommers (changes requested):
Hi @iggyzap, it would probably be better to actually store the data in the `iotdata_backend`.
```
def publish(self, topic, qos, payload):
    iot_backends[self.region_name].published_data.append(payload)
    return None
```

That way we can test the data is persisted, without having to rely on the `unittest.mock` module:
```
from moto import settings

def test_publish(mock_backend):
    ...
        mock_backend.published_data.should.equal([..])
```
Let me know if that makes sense

iggyzap:
I'll make a change.

bblommers:
This has mow been superceded by #4848

PR #4699, opened 2021-12-17
Feature: Mock region

### LINKED ISSUE #4252: Mock region
Can you allow mock regions in the mocks? 
I have this set
`os.environ['AWS_DEFAULT_REGION'] = 'any-region'`

and error it produces:
```
E       KeyError: 'any-region'
../../../.local/lib/python3.8/site-packages/moto/dynamodb2/responses.py:79: KeyError
```

bblommers:
Feature to allow a user to specify a non-existing region. Having a non-existent region can be useful as an additional layer of protection, to ensure the user does not modify real AWS architecture.

 This feature is disabled by default, but can be enabled with an environment variable. (`MOTO_ALLOW_NONEXISTENT_REGION`)

In order to make this work, all backends are now generated on-request, rather then on start-up.
This results in:
 - A single place where we can decide whether a region is valid/allowed
 - Better performance on start-up (we don't need to generate Backends for 27 regions, just the one required)
 - Better performance on reset (as we only need to reset a single Backend, instead of all 27.)

Closes #4252


bpandola (approved):
I don't have the bandwidth to review all the changes across all 98 files, but I understand the problem this PR aims to solve and I think the implementation makes sense.

The added tests do seem to cover the new functionality, so if all the other tests still pass this LGTM.

bblommers:
Understandable! Thanks @bpandola

PR #4701, opened 2021-12-17
Added new functionality for EC2

SimonToftegaard:
Added new functionality for EC2: 
- disable_ebs_encryption_by_default, 
- enable_ebs_encryption_by_default, 
- get_ebs_encryption_by_default

### DIFF
--- a/IMPLEMENTATION_COVERAGE.md
+++ b/IMPLEMENTATION_COVERAGE.md
@@ -1185,7 +1185,7 @@
 
 ## ec2
 <details>
-<summary>33% implemented</summary>
+<summary>34% implemented</summary>
 
 - [ ] accept_reserved_instances_exchange_quote
 - [ ] accept_transit_gateway_multicast_domain_associations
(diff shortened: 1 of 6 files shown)


SimonToftegaard:
The linting is complaining that line 1487 is missing an empty line:
`    def __init__(self):
        self.ebs_encryption_by_default = False <- line 1487
        super().__init__()`

So im a bit unsure where the extra empty line should be and when I look at other classes mine is setup the same way and then I run flake8 locally I don't get the same linting error.

bblommers:
Hey @SimonToftegaard! Github has a different line numbering going on: https://github.com/spulec/moto/pull/4701/files#diff-8e52fedb6e57d93299accfbfb74557e5c58f6790190f7b0c6304d864ec9c41e2

So it refers to the class-definition itself, there should be two empty lines between each class:
```
    self.keypairs[key_name] = keypair
    return keypair
# 1
# 2
class SettingsBackend(object):
```

bblommers (approved):
LGTM - thanks for the PR @SimonToftegaard!

PR #4709, opened 2021-12-21
Update xray client to forward args

jhaenchen:
Mocked xray client should handle args and kwargs gracefully.


bblommers (approved):
Yes, it should - thanks for the fix @jhaenchen!

PR #4710, opened 2021-12-22
Batch - ensure that logs are posted in chronological order

morsecodist:
I am testing Batch jobs in moto and many of my batch jobs fail. It appears that some jobs are posting log events that are not in chronological order. I believe this is because here:

```python
                for line in logs_stdout + logs_stderr:
                    date, line = line.split(" ", 1)
                    date_obj = (
                        dateutil.parser.parse(date)
                        .astimezone(datetime.timezone.utc)
                        .replace(tzinfo=None)
                    )
                    date = unix_time_millis(date_obj)
                    logs.append({"timestamp": date, "message": line.strip()})
```

stdout and stderr are appended together but some stderr events may come before some stdout events. I considered a few things to fix this but ultimately I think sorting the logs after the fact is the best way to fix this. If the logs are out of order for any reason and you want them to stay that way this removes that.

I also considered weaving the two together and I tested this and it does fix the problem. However it makes the code a lot more complicated for minimal benefit.

I also considered adding the previous timestamp if the timestamp is earlier than the previous, though this distorts the timestamps.

### DIFF
--- a/moto/batch/models.py
+++ b/moto/batch/models.py
@@ -609,7 +609,6 @@ class Job(threading.Thread, BaseModel, DockerModel):
                 logs_stdout = [x for x in logs_stdout if len(x) > 0]
                 logs_stderr = [x for x in logs_stderr if len(x) > 0]
                 logs = []
-                prev_date = None
                 for line in logs_stdout + logs_stderr:
                     date, line = line.split(" ", 1)
                     date_obj = (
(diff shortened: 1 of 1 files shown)


bblommers (approved):
LGTM - makes sense to do this at the end. Thanks for the PR @morsecodist!

Issue #4708, opened 2021-12-21
respond_decision_task_completed API call not functionally complete, despite implementation_coverage claim


JimNero009:
I believe I have stumbled across an issue with the SWF implementation you provide -- at least, in terms of the documented support.

The implementation docs suggest that the 'respond_decision_task_completed' API call is supported. However, if you call it with a decision that is not implemented (see https://github.com/spulec/moto/blob/master/moto/swf/models/workflow_execution.py#L410-L419), then that call will fail and it will raise a NotImplementedError.

Sample code to reproduce:
```
import boto3
from moto import mock_swf

@mock_swf
def test_moto_directly():
    client = boto3.client("swf")
    client.register_domain(name="testdomain", workflowExecutionRetentionPeriodInDays="1")
    client.register_workflow_type(domain="testdomain", name="test", version="1", defaultTaskStartToCloseTimeout="60", defaultExecutionStartToCloseTimeout="60")
    client.start_workflow_execution(domain="testdomain", workflowId="foobar", workflowType={"name": "test", "version": "1"}, taskList={"name": "testtasklist"}, childPolicy="TERMINATE")
    task_token = client.poll_for_decision_task(domain="testdomain", taskList={"name": "testtasklist"})["taskToken"]
    decisions = [{'decisionType': 'RecordMarker', 'recordMarkerDecisionAttributes': {'details': '{"timestamp":sometime}', 'markerName': '_workflow_event_workflowExecutionStarted'}}]
    client.respond_decision_task_completed(taskToken=task_token, decisions=decisions)
```

Produces a stack trace like:
```
Traceback (most recent call last):
  File "recreate_swf_issue.py", line 30, in <module>
    test_moto_directly()
  File "/Users/jimcockburn/<redacted>/<redacted>/.venv/lib/python3.5/site-packages/moto/core/models.py", line 100, in wrapper
    result = func(*args, **kwargs)
  File "recreate_swf_issue.py", line 27, in test_moto_directly
    client.respond_decision_task_completed(taskToken=task_token, decisions=decisions)
  File "/Users/jimcockburn/<redacted>/<redacted>/.venv/lib/python3.5/site-packages/botocore/client.py", line 357, in _api_call
    return self._make_api_call(operation_name, kwargs)
  File "/Users/jimcockburn/<redacted>/<redacted>/.venv/lib/python3.5/site-packages/botocore/client.py", line 648, in _make_api_call
    operation_model, request_dict, request_context)
[...]

PR #4713, opened 2021-12-22
Implement "RecordMarker", "StartTimer", "CancelTimer" and "CancelWorkflowExecution" decisions

JimNero009:
This PR somewhat addresses https://github.com/spulec/moto/issues/4708, in that more of the unimplemented decision types now have logic backing them.

I was interested also in the StartChildWorkflowExecution decision, but felt the PR was getting big enough and I can live without that for a little while longer.

I have also tried to implement a couple of 'common' failure modes for these decisions, but they are not completed. I have marked TODOs in the code where other failure states should be considered.

### DIFF
--- a/tests/test_swf/models/test_workflow_execution.py
+++ b/tests/test_swf/models/test_workflow_execution.py
@@ -1,7 +1,16 @@
+from threading import Timer as ThreadingTimer
+
 from freezegun import freeze_time
+from unittest.mock import Mock, patch
 import sure  # noqa # pylint: disable=unused-import
 
-from moto.swf.models import ActivityType, Timeout, WorkflowType, WorkflowExecution
+from moto.swf.models import (
+    ActivityType,
+    Timeout,
+    Timer,
+    WorkflowType,
+    WorkflowExecution,
+)
 from moto.swf.exceptions import SWFDefaultUndefinedFault
 from ..utils import (
     auto_start_decision_tasks,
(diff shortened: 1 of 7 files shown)


tests/test_swf/models/test_workflow_execution.py
```diff
@@ -580,8 +579,52 @@ def test_create_timer():
         last_event.event_attributes["decisionTaskCompletedEventId"].should.equal(123)
 
 
+def test_start_timer_correctly_fires_timer_later():
+    wfe = make_workflow_execution()
+    START_TIMER_EVENT_ATTRIBUTES = {"startToFireTimeout": "60", "timerId": "abc123"}
+
+    class MockEvent:
```
JimNero009, line 586: Note to self -- probably only actually need to mock the 'wait' function.

bblommers (commented):
LGTM so far @JimNero009! It would be good to have an integration test as well, to verify the actual behaviour of the timer - so without any Mocks involved.

There are a few tests already in `responses/test_workflow_executions.py` that you can borrow from. (Note that you'd only need to write a test for `mock_swf` - no need to verify the behaviour under the deprecated decorator.)

If a 'perfect' integration test means adding some `time.sleep()` statements, that's OK as well.

JimNero009:
Hi Bert, thanks for the suggestion -- I was thinking myself that essentially inserting a full workflow definition to follow that includes these events would be great, so awesome to see you have a suite for that.

Thanks for the eyes, good to know I'm going in generally the right direction, will keep at it!

JimNero009:
I don't _think_ that test failure is caused by my change?

bblommers:
> I don't _think_ that test failure is caused by my change?

No, that was an intermittent failure - I've restarted the build and it all looks OK now

bblommers (approved):
Approved - thanks for the PR @JimNero009!

Issue #4707, opened 2021-12-21
S3Control api in moto


SimonToftegaard:
I needed to use the s3control api in moto and according to https://github.com/spulec/moto/blob/master/IMPLEMENTATION_COVERAGE.md it was not yet implemented.
So I started working on the implementation, but then I noticed that part of the s3control api was actually already implemented under the s3 api in moto.
So before continuing the work I would really like to be sure what is the right way to continue?
Do a full separate implementation under s3control or something else?

Also the current implementation only seem to work with the builtin account:
`from moto.s3.models import ACCOUNT_ID`

bblommers:
Hi @SimonToftegaard, good catch! Ideally everything related to s3control would be separated out, but that would be a breaking change at this point.
How many features are you looking to implement? To ensure backwards compatibility, it might be better to make it part of the s3 api (`mock_s3`) for now. Once we get around to doing a major release, we can yoink everything out and make it available under `mock_s3control`.

SimonToftegaard:
My plan was to initially implement: 
get_public_access_block
put_public_access_block
delete_public_access_block
And I have made an urls.py for s3control like this:
`from .responses import S3ControlResponse

url_bases = [
    r"https?://(.+)\.s3-control\.(.+)\.amazonaws\.com",
]

url_paths = {
    "{0}/v20180820/configuration/publicAccessBlock$": S3ControlResponse.dispatch
}`

Could I just put something similar into the urls py of s3? The main problem for me right now it seems like the current implementation only works for the builtin accountid and I need to setup additional configuration for the account for my test, including an organization structure and that does not seem possible with the builtin one.

Another solution would just to implement the new s3control and leave the old implementation in s3. That way anyone could use the old implementation with mock_s3 and the new implementation could be used with mock_s3control.
Let me know what you think.

bblommers:
The `*_public_access_block`-features are already supported, as far as I know? At least, we have tests for them: 

https://github.com/spulec/moto/blob/ce9fb906635d266610abe2acd1f47c750bb5257d/tests/test_s3/test_s3.py#L2326-L2688

The fact that we only support the builtin accountid is definitely an issue, but that's Moto-wide - every service is single-account/single-user based. I don't see an easy way of implementing support for this, without losing the current functionality (accepting one account, rejecting unknown accounts).

The builtin accountid is centralized though - can you mock it during the test?
```
moto.core.ACCOUNT_ID = "myotheraccountid"
```

[...]

PR #4714, opened 2021-12-22
Initial work for split out of s3control

SimonToftegaard:
As discussed in #4707 here is the initial separate implementation of s3control

### DIFF
--- a/IMPLEMENTATION_COVERAGE.md
+++ b/IMPLEMENTATION_COVERAGE.md
@@ -4215,6 +4215,69 @@
 - [ ] write_get_object_response
 </details>
 
+## s3control
+<details>
+<summary>5% implemented</summary>
+
+- [ ] create_access_point
+- [ ] create_access_point_for_object_lambda
+- [ ] create_bucket
+- [ ] create_job
+- [ ] create_multi_region_access_point
+- [ ] delete_access_point
+- [ ] delete_access_point_for_object_lambda
+- [ ] delete_access_point_policy
+- [ ] delete_access_point_policy_for_object_lambda
+- [ ] delete_bucket
+- [ ] delete_bucket_lifecycle_configuration
+- [ ] delete_bucket_policy
+- [ ] delete_bucket_tagging
+- [ ] delete_job_tagging
+- [ ] delete_multi_region_access_point
+- [X] delete_public_access_block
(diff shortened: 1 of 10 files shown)


bblommers:
Thanks @SimonToftegaard

SimonToftegaard:
Seems like the unit tests in server mode do not get intercepted correctly by the mock_s3control, not sure how to fix this.

bblommers:
One potential cause could be that the request is picked up by the S3-module, instead of S3control, because the regex for the S3-urls will still catch requests to `s3control`-urls. That's just a hunch though, I'd have look a little deeper into it to verify whether that's the case

bblommers:
OK, found the issue - ServerMode tests will always fail for S3Control.  
From the original tests:
```
# All tests for s3-control cannot be run under the server without a modification of the
# hosts file on your system. This is due to the fact that the URL to the host is in the form of:
# ACCOUNT_ID.s3-control.amazonaws.com <-- That Account ID part is the problem. If you want to
# make use of the moto server, update your hosts file for `THE_ACCOUNT_ID_FOR_MOTO.localhost`
# and this will work fine.
```

#4745 contains a clean implementation of this change, without any changes to the logic or tests. We will merge that one first, to ensure we're not breaking anything else.
After that we can have a look at extending the logic to allow other account ID's.

bblommers:
OK, found the issue - ServerMode tests will always fail for S3Control.  
From the original tests:
```
# All tests for s3-control cannot be run under the server without a modification of the
# hosts file on your system. This is due to the fact that the URL to the host is in the form of:
# ACCOUNT_ID.s3-control.amazonaws.com <-- That Account ID part is the problem. If you want to
# make use of the moto server, update your hosts file for `THE_ACCOUNT_ID_FOR_MOTO.localhost`
# and this will work fine.
```

#4745 contains a clean implementation of this change, without any changes to the logic or tests. We will merge that one first, to ensure we're not breaking anything else.
After that we can have a look at extending the logic to allow other account ID's.

bpandola:
Closing this out as it's gotten too out-of-sync with the `master` branch, which has had lots of work done on the `s3control` backend in the ensuing years.  If there are bits in here that are still not implemented in Moto, feel free to clean up and resubmit.

Issue #4418, opened 2021-10-15
Allow mock_s3 to use the standalone moto server


jbvsmo:
Long story short, I need to run the standalone moto server for pyspark to be able to access files, but I would like to use the same server for the internal calls happening on the python side using boto3.

The solution for changing the `endpoint_url` per every Session created isn't feasible for me and having to add a `.boto` config file is also complicated as I don't want to keep writing and erasing files and potentially have a broken env.

If boto were to implement something like this: https://github.com/boto/boto3/pull/2746
it would be very simple to setup the environment variable, but seems like they aren't willing to.

Then another solution would seem the best which is allow mock_s3 (or any service mock) decorator to get a parameter and point to original mock server such as

    @mock_s3(url='http://localhost:5000')
    def test_foo(self):
        s3 = boto3.client('s3')
        s3.download_file(....)

Background info:
https://github.com/spulec/moto/issues/1543
https://github.com/boto/boto3/issues/1375

jbvsmo:
Seems like you have many hidden secrets! (Or I can't read docs)

It appears to be working for the initial thing I'm trying. I will do further testing.
Thank you

PR #4723, opened 2021-12-27
Allow mock_s3 to use the standalone moto server running in alternative endpoint by adding TEST_SERVER_MODE_ENDPOINT environment variable

ikkala:
Before this, the ServerMode (TEST_SERVER_MODE=true) endpoint has been hardcoded to `http://localhost:5000`. If MotoServer runs in different endpoint than `http://localhost:5000`, for example in case when the it runs in separate Docker container, the tests are failing.

This pull request makes the previously hardcoded value as default value for new `TEST_SERVER_MODE_ENDPOINT` environment variable.  If MotoServer runs in different endpoint than `http://localhost:5000` the endpoint can be overridden by using `TEST_SERVER_MODE_ENDPOINT` environment variable.

Example usage in docker-compose.yml:

```
  moto-server:
    image: motoserver/moto:2.2.11

  test:
    build:
      context: packages/tests/
      dockerfile: Dockerfile_tests
    depends_on:
      - moto-server
    environment:
      TEST_SERVER_MODE: 'true'
      TEST_SERVER_MODE_ENDPOINT: http://moto-server:5000
      AWS_ENDPOINT_URL: http://moto-server:5000
```

In addition, this pull requests fixes error in documentation: It was saying erroneously that `MOTO_CALL_RESET_API=true` keeps the state in between tests.


ikkala:
This pull requested was inspired by discussion in https://github.com/spulec/moto/issues/4418 ("Allow mock_s3 to use the standalone moto server").

bblommers:
Hi @ikkala, this makes a lot of sense, to make this endpoint configurable. There is already a PR outstanding with some changes in this area, #4658. I'll have another look here once that is merged.

ikkala:
Ok, excellent!  Nice to hear that there's related work going on. 

I tried to skim through similar issues and pull requests but didn't find anything that changes also the host name.  For example, looks like that the #4658 only makes the port variable: https://github.com/spulec/moto/pull/4658/files#diff-3ed8c52a492672972d143cfc55b9d3a6024396b00ab554538b72c01b46948430R522
So, I decided to make this pull request.

This pull request change is already in use in our integration tests that have been put together by using Docker Compose, as described in "Example usage in docker-compose.yml" above, so it works. But, it's only for testing of S3 bucket handling, so it might of course miss something that should be added for make it work in all scenarios.

ikkala:
Hi @bblommers again, looks like you have now merged the #4658 changes -- you make lots of work for moto, thank you!  :+1: 

I adapted this PR changes to the latest master trunk. Would it be ready for merge in near future..?

bblommers (approved):
LGTM! Thanks again for the PR @ikkala

PR #4724, opened 2021-12-27
fix(s3/listparts): check parts length

dlsrb6342:
* With Multipart Uploads, If there is no parts yet, `is_truncated` value is set `[]`.
  * `parts and self.backend.is_truncated(...)` result is always  `[]` if `parts` is empty list.
* Check `len(parts) != 0` not `parts`.


bblommers (approved):
LGTM! Thanks for the fix @dlsrb6342

PR #4729, opened 2021-12-29
Escape windows return character

pinzon:
This PR addresses issue localstack/localstack#5172 where is described that the SQS service removes the "\r" character from the body messages when receiving the messages from a queue.

Changes:
 - replacement of the `\r` for its unicode representation `&#xd;`. 
 - replacement of the `"` for the html entity `&quot;`. This was not necessary but it makes the responses more in pair with the AWS responses.
 - Test added to validate character is not being removed.

### DIFF
--- a/moto/sqs/models.py
+++ b/moto/sqs/models.py
@@ -150,7 +150,7 @@ class Message(BaseModel):
 
     @property
     def body(self):
-        return escape(self._body)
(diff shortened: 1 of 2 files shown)


bblommers (approved):
LGTM! Thanks for the fix @pinzon

Issue #4423, opened 2021-10-16
AWS Batch - cancel job has no opportunity to catch jobs for cancellation


dazza-codes:
Based on the implementation, it should be possible to cancel an AWS Batch job before it is `STARTING`, i.e.

https://github.com/spulec/moto/blob/master/moto/batch/models.py#L1378-L1382

Contains:

```python
    def cancel_job(self, job_id, reason):
        job = self.get_job_by_id(job_id)
        if job.job_state in ["SUBMITTED", "PENDING", "RUNNABLE"]:
            job.terminate(reason)
        # No-Op for jobs that have already started - user has to explicitly terminate those
```

In short, moto 1.x provided jobs with status in that list, while moto 2.x does not.

The actual effective implementation for batch jobs seems to have changed between moto 1.x and 2.x, where the latter seems to create jobs that enter a `STARTING` state immediately, or nearly immediately.  This is based on experience with moto 1.x and 2.x from tests in https://github.com/dazza-codes/aio-aws

## moto 1.x

https://github.com/dazza-codes/aio-aws/blob/main/tests/test_aio_aws_batch.py

It might be too much to ask to go clone that repo/branch and run it, but here's the summary and some log snippets:
```
$ pytest -s tests/test_aio_aws_batch.py -k cancel

2021-10-16T21:23:33.290Z | INFO | aio_aws.aio_aws_batch:aio_batch_job_submit:264 | AWS batch-submit-job (sleep-5-job:0f6f1963-e5e0-428c-8b4f-303cf6b968a8) try: 1 of 4
2021-10-16T21:23:33.296Z | INFO | aio_aws.aio_aws_batch:aio_batch_job_cancel:487 | AWS Batch job to cancel: 0f6f1963-e5e0-428c-8b4f-303cf6b968a8, test-job-cancel
2021-10-16T21:23:33.318Z | INFO | aio_aws.aio_aws_batch:aio_batch_job_status:594 | AWS Batch job (sleep-5-job:0f6f1963-e5e0-428c-8b4f-303cf6b968a8) status: PENDING
2021-10-16T21:23:33.690Z | INFO | aio_aws.aio_aws_batch:aio_batch_job_status:594 | AWS Batch job (sleep-5-job:0f6f1963-e5e0-428c-8b4f-303cf6b968a8) status: PENDING
[...]

PR #4734, opened 2022-01-02
Feature: Resource State Transition

bblommers:
This PR exposes an API that can artificially delay state transitions. 

At the moment, resources are usually available immediately. This API would make it possible to configure, for example, at which point an EC2 instance has the status 'ready' - after 3 seconds, after 30 seconds, .. whatever makes sense for your test.

Link to documentation: https://moto--4734.org.readthedocs.build/en/4734/docs/configuration/state_transition.html

Related issues:
#380
#4423 

Unrelated changes that are part of this PR:
 - Added documentation for environment variables
 - Move the MotoAPIBackend to a dedicated `moto/moto_api` directory

TODO:
 - [x] Implement this feature for Support (cases)
 - [ ] ~~Implement this feature for Support (checks)~~ This actually involves a big rewrite of the implementation, so I'll leave this for now
 - [x] CloudFront::Distribution
 - [x] Transcribe::Vocabulary
 - [x] Transcribe::MedicalVocabulary
 - [x] Transcribe::TranscriptionJob
 - [x] Transcribe::MedicalTranscriptionJob
 - [ ] Test/verify this feature covers the use cases in #380
 - [x] Implement/test this feature for Batch, to ensure it covers #4423 
 - [x] Write development documentation


bblommers:
@spulec @bpandola This is very much WIP, but it would be good to get some early feedback to ensure if I'm on the right track here. 
This is quite different from what Moto does normally, so if we go ahead with this approach, we need to really make sure it's sensible/usable/understandable for developers.

PR #4736, opened 2022-01-04
Fix response data types for kinesis operations

timleslie:
This PR fixes a bug in kinesis operations returning the fields `"SequenceNumber"`, `"StartingSequenceNumber"`, and `"EndingSequenceNumber"`. This includes `put_record()`, `put_records()`, and `list_shards()`.

Each of these fields now returns a string rather than an int, as per the spec:

Reference: https://docs.aws.amazon.com/kinesis/latest/APIReference/kinesis-api.pdf


bblommers (approved):
LGTM! Thanks for the fix @timleslie

PR #4741, opened 2022-01-06
SNS: publish_batch()

### LINKED ISSUE #4612: Support for sns publish_batch
AWS Recently announce support for publishing messages in a batch to a SNS topic:
https://aws.amazon.com/blogs/compute/publishing-messages-in-batch-to-amazon-sns-topics/

This is supported by boto:
https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/sns.html#SNS.Client.publish_batch

 But as this was just published few days ago, as of `moto==2.2.16` this is not yet supported by moto:
 
 ```python
 import os

import boto3
from moto import mock_sns


@mock_sns
def main():
    os.environ["AWS_DEFAULT_REGION"] = "us-east-1"
    sns = boto3.client("sns")

    sns.publish_batch(
        TopicArn="arn:aws:sns:us-east-1:123456789012:mystack-mytopic-NZJ5JSMVGFIE",
        PublishBatchRequestEntries=[
            {
                "Id": "id1",
                "Message": "Message content 1",
            },
            {
                "Id": "id2",
                "Message": "Message content 2",
            },
        ]
    )


if __name__ == "__main__":
    main()  # Will throw `NotImplementedError: The publish_batch action has not been implemented`
 ```

I tested that the above works to a real topic with `boto3==1.20.10` without `mock_sns` decorator.

bblommers:
Fixes #4612 

Negative tests have been verified against AWS

### DIFF
--- a/moto/core/responses.py
+++ b/moto/core/responses.py
@@ -482,14 +482,8 @@ class BaseResponse(_TemplateEnvironmentMixin, ActionAuthenticatorMixin):
             tracked_prefixes or set()
         )  # prefixes which have already been processed
 
-        def is_tracked(name_param):
-            for prefix_loop in tracked_prefixes:
-                if name_param == prefix_loop:
-                    return True
-            return False
-
         for name, value in self.querystring.items():
-            if is_tracked(name) or not name.startswith(param_prefix):
+            if not name.startswith(param_prefix):
                 continue
 
             if len(name) > len(param_prefix) and not name[
--- a/moto/core/responses.py
+++ b/moto/core/responses.py
@@ -484,7 +484,7 @@ class BaseResponse(_TemplateEnvironmentMixin, ActionAuthenticatorMixin):
 
         def is_tracked(name_param):
             for prefix_loop in tracked_prefixes:
-                if name_param.startswith(prefix_loop):
+                if name_param == prefix_loop:
                     return True
             return False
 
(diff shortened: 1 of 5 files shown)


moto/core/responses.py
```diff
@@ -482,14 +482,8 @@ def _get_multi_param_helper(
...
         )  # prefixes which have already been processed
 
-        def is_tracked(name_param):
-            for prefix_loop in tracked_prefixes:
-                if name_param.startswith(prefix_loop):
-                    return True
-            return False
-
         for name, value in self.querystring.items():
-            if is_tracked(name) or not name.startswith(param_prefix):
```
bblommers, line 492: This was originally introduced for CloudWatch. SNS parameters are in a slightly different format (of course), so it could not be reused. It's not clear to me why this was introduced in the first place, but as far as I can tell, it doesn't do much - removing it ensures that both CW and SNS parameters are handled correctly

vaibsgharge:
Sorry, I can't find the publish_batch() inside boto3 is it only added to moto? am I missing something?

bblommers:
Hi @vaibsgharge, it is part of boto3, see the documentation here: https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/sns.html#SNS.Client.publish_batch

This feature was introduced last November, IIRC, so older versions of `boto3` may not support it - but version `1.21.34` should absolutely do.

PR #4745, opened 2022-01-08
Separate S3Control methods from S3

bblommers:
The S3-module contained three methods that were technically part of `S3Control`: 
 - delete_public_access_block()
 - get_public_access_block()
 - put_public_access_block

Because they were part of the S3-module, it needed the S3-mock:
```
@mock_s3
def test():
    client = boto3.client("s3control")
    ...
```

This PR extracts the logic into a dedicated module, making it more discoverable and obvious that we support these methods.  
This means users now have to use the dedicated `mock_s3control`:
```
@mock_s3control
def test():
    client = boto3.client("s3control")
    ...
```
--------------

This is a breaking change, so it will only be merged as part of the next major release.

Note that none of the existing logic or tests have been changed as part of this PR - only moved around.


bblommers:
@mikegrima I believe you implemented this feature originally - any objections to this change from your side?

mikegrima:
Hi @bblommers! As long as all of the previous functionality will work properly then I'm OK with the change.  Is the breaking change that instead of `mock_s3`, its `mock_s3control`?

bblommers:
Correct @mikegrima - it's just a matter of using the new decorator from 3.x. None of the logic has changed.

mikegrima:
LGTM

bblommers:
Thanks @mikegrima!

PR #4747, opened 2022-01-11
feat: add support for filtering by tags in describe_managed_prefix_lists

Neki:
This PR adds support for filtering by tags (instead of ignoring the tag filter and returning all results) during calls to `describe_managed_prefix_lists` (EC2 client).

The test added in this PR doesn't pass without the associated change.

I made the minimal change that would allow me to use moto to test a project I'm working on. Please let met know if that approach is OK!

### DIFF
--- a/moto/ec2/models.py
+++ b/moto/ec2/models.py
@@ -5610,6 +5610,7 @@ class ManagedPrefixListBackend(object):
         result = managed_prefix_lists
         if filters:
             result = filter_resources(managed_prefix_lists, filters, attr_pairs)
+            result = describe_tag_filter(filters, managed_prefix_lists)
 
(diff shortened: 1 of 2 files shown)


Neki:
The test failures seem unrelated. Please let me know if there's something I need to do to make them pass.

bblommers (changes requested):
Hi @Neki, thanks for the PR! Would you mind merging with master? That should fix the CI - happy to merge once all the tests pass

Neki:
Thanks, I rebased this MR against the latest master, tests are green now :heavy_check_mark:

bblommers (approved):
Nice one - thanks again @Neki!

PR #4748, opened 2022-01-11
add internal host to batch to mirror sfn behavior

morsecodist:
Add the equivalent behavior to batch for this PR: https://github.com/spulec/moto/pull/4408.


bblommers (changes requested):
Hi @morsecodist, looks like the linter still fails because of the unused imports.

I expect some of the tests to fail due to unrelated bugs, so don't worry if that happens

bblommers (approved):
LGTM! Thanks for the PR @morsecodist

PR #4752, opened 2022-01-12
Fix interoperability with responses 0.17.x

bblommers:
The `_is_string` method has been removed from `responses`, and will break compatibility with future releases


bblommers:
Closing in favour of #4753

PR #4754, opened 2022-01-13
docker networks support

morsecodist:
This adds a new environment variable `MOTO_DOCKER_NETWORK` that is the name of a docker network to bind docker containers ran by moto to. This is for usage of the moto container with docker compose. By creating an overlay network in your docker compose file you can add that network name in this environment variable. This allows you to communicate with other mock services while mocking your application in docker compose.

### DIFF
--- a/moto/awslambda/models.py
+++ b/moto/awslambda/models.py
@@ -584,6 +584,8 @@ class LambdaFunction(CloudFormationModel, DockerModel):
                             "host.docker.internal": "host-gateway"
                         }
 
+                    networks = os.environ["MOTO_DOCKER_NETWORKS"].split() if "MOTO_DOCKER_NETWORKS" in os.environ else []
+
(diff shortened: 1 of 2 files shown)


bblommers:
This makes a lot of senes, @morsecodist. Adding this variable in AWSLambda is already part of #4658, so let's wait until that one gets merged.
Afterwards we can rebase/repurpose this PR to also add the network name to Docker containers in Batch.

morsecodist:
@bblommers makes sense, thanks sorry I missed that other PR. I will get this in batch once that's done but I will close this for now. Let me know if there is any way I can assist. Also maybe we should create some sort of docker wrapper so we don't need to add functionality in two places? It may also be relevant for ECS if it is added.

bblommers:
> Also maybe we should create some sort of docker wrapper so we don't need to add functionality in two places? 

Agreed, some DRY code would be nice, with all the configuration options that we have now!

PR #4765, opened 2022-01-15
SQS - Improve delete_message_batch() error responses

sjakthol:
Currently, if the input to SQS.delete_message_batch() includes an invalid receipt handle, moto returns an error:

```
  File "test.py", line 26, in <module>
    sqs.delete_message_batch(
  File "<...>/botocore/client.py", line 391, in _api_call
  File "<...>/botocore/client.py", line 719, in _make_api_call
botocore.errorfactory.ReceiptHandleIsInvalid: An error occurred (ReceiptHandleIsInvalid) when calling the DeleteMessageBatch operation: The input receipt handle is invalid.
```

However, if the input to Amazon SQS DeleteMessageBatch API includes an invalid receipt handle, the Amazon SQS API returns the error in Failed field of an 200 OK response:

```
{
  "Successful": [...],
  "Failed": [{
    "Id": "invalid",
    "SenderFault": True,
    "Code": "ReceiptHandleIsInvalid",
    "Message": 'The input receipt handle "invalid" is not a valid receipt handle.',
  }],
  "ResponseMetadata": {...},
}
```

This commit updates moto delete_message_batch() method implementation to return ReceiptHandleIsInvalid errors in the "Failed" field of a 200 OK response instead of failing the entire request.

The implementation follows the implementation of similar functionality in the SQS change_message_visibility_batch() method. Test case added to check output in case of a partial failure.

### DIFF
--- a/moto/sqs/responses.py
+++ b/moto/sqs/responses.py
@@ -351,14 +351,26 @@ class SQSResponse(BaseResponse):
                 raise BatchEntryIdsNotDistinct(receipt_and_id["msg_user_id"])
             receipt_seen.add(receipt)
 
+        success = []
+        errors = []
         for receipt_and_id in receipts:
-            self.sqs_backend.delete_message(
-                queue_name, receipt_and_id["receipt_handle"]
-            )
+            try:
+                self.sqs_backend.delete_message(
+                    queue_name, receipt_and_id["receipt_handle"]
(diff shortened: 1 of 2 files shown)


bblommers (approved):
LGTM! Thanks for the fix @sjakthol

PR #4769, opened 2022-01-16
feat: SES add (minimal) address validation

kefi550:
Adding validation of source and destination addresses as I need a mock related to the following specification.

https://docs.aws.amazon.com/ses/latest/APIReference/API_SendRawEmail.html
> If you send a single message to more than one recipient address, and one of the recipient addresses isn't in a valid format (that is, it's not in the format UserName@[SubDomain.]Domain.TopLevelDomain), Amazon SES rejects the entire message, even if the other addresses are valid.

The validation is minimal and only checks if the domain part of the email address exists.
Full validation is not implemented because the exact specification is unknown.

### DIFF
--- a/tests/test_ses/test_ses_boto3.py
+++ b/tests/test_ses/test_ses_boto3.py
@@ -140,6 +140,29 @@ def test_send_unverified_email_with_chevrons():
 
 
 @mock_ses
+def test_send_email_invalid_address():
+    conn = boto3.client("ses", region_name="us-east-1")
+    conn.verify_domain_identity(Domain="example.com")
+
+    with pytest.raises(ClientError) as ex:
+        conn.send_email(
+            Source="test@example.com",
+            Destination={
+                "ToAddresses": ["test_to@example.com", "invalid_address"],
+                "CcAddresses": [],
+                "BccAddresses": [],
+            },
+            Message={
+                "Subject": {"Data": "test subject"},
+                "Body": {"Text": {"Data": "test body"}},
+            },
+        )
+    err = ex.value.response["Error"]
+    err["Code"].should.equal("InvalidParameterValue")
+    err["Message"].should.equal("Missing domain")
+
+
+@mock_ses
 def test_send_templated_email():
     conn = boto3.client("ses", region_name="us-east-1")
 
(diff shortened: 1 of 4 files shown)


tests/test_ses/test_ses_boto3.py
```diff
@@ -79,6 +79,11 @@ def test_send_email():
         **dict(kwargs, Destination={"ToAddresses": too_many_addresses})
     ).should.throw(ClientError)
 
+    invalid_addresses = ["test_to@example.com", "invalid_address"]
```
bblommers, line 82: This should be in a separate method, as it is a standalone test case.

Ideally it also uses the with pytest.raises format:
```
    ....
err["Message"].should.equal("specific error message")
```
Following this format ensures that the correct error is thrown.

tests/test_ses/test_ses_boto3.py
```diff
@@ -179,6 +184,11 @@ def test_send_templated_email():
...
```
bblommers, line 187: Same here (separate method + pytest.raises)

tests/test_ses/test_ses_boto3.py
```diff
@@ -243,6 +253,23 @@ def test_send_raw_email_validate_domain():
...
+    del message["To"]
+    conn.verify_domain_identity(Domain="example.com")
+
+    invalid_addresses = ["test@example.com", "invalid_address"]
+    kwargs = dict(
+        Source=message["From"],
+        Destinations=invalid_addresses,
+        RawMessage={"Data": message.as_string()},
+    )
+    conn.send_raw_email.when.called_with(**kwargs).should.throw(ClientError)
```
bblommers, line 270: This needs the `pytest.raises` approach as well

bblommers (changes requested):

tests/test_ses/test_ses_boto3.py
kefi550: Fixed to (separate method + pytest.raises)

tests/test_ses/test_ses_boto3.py
kefi550: Same as above

tests/test_ses/test_ses_boto3.py
kefi550: Fixed to `pytest.raises` format

bblommers (approved):
LGTM! Thanks for making the changes @kefi550

PR #4770, opened 2022-01-17
Add CreatedDate and LastChangedDate in secretsmanager responses

dfangl:
Currently, these dates are missing, but should be present.

As an example, here is a request for list-secrets against AWS:
```
$ aws secretsmanager list-secrets
{
    "SecretList": [
        {
            "ARN": "arn:aws:secretsmanager:...",
            "Name": "testsecret",
            "Description": "some desc",
            "LastChangedDate": "2022-01-17T10:41:35.163000+01:00",
            "SecretVersionsToStages": {
                "e75a49b6-e07a-4bd9-acd7-510557dfc9b1": [
                    "AWSCURRENT"
                ]
            },
            "CreatedDate": "2022-01-14T19:43:03.127000+01:00"
        }
    ]
}
```

This PR introduces this behavior in moto. As their names say, CreatedDate will only be set on creation of the secret, LastChangedDate also on update operation.

The PR also includes the corresponding tests, both in existing test cases (where fitting, for additional attributes this seemed appropriate) and adds a new test specifically testing the CreatedDate / LastChangedDate behavior on object updates.

### DIFF
--- a/moto/secretsmanager/models.py
+++ b/moto/secretsmanager/models.py
@@ -61,6 +61,8 @@ class FakeSecret:
         kms_key_id=None,
         version_id=None,
         version_stages=None,
+        last_changed_date=None,
+        created_date=None,
     ):
         self.secret_id = secret_id
         self.name = secret_id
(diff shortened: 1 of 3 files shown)


bblommers (approved):
LGTM! Thanks for the PR @dfangl  :+1:

PR #4771, opened 2022-01-18
Implement list_hosted_zones_by_vpc() functionality

pll:
Hi,

This is my first attempt at implementing something for moto. Please tell me if I'm even on the right track with this one :)

-Paul

### DIFF
--- a/moto/route53/models.py
+++ b/moto/route53/models.py
@@ -487,8 +487,8 @@ class Route53Backend(BaseBackend):
                             "HostedZoneId": this_id,
                             "Name": zone.name,
                             "Owner": {"OwningAccount": ACCOUNT_ID},
-                            }
-                        )
+                        }
+                    )
 
         return zone_list
 
--- a/moto/route53/models.py
+++ b/moto/route53/models.py
@@ -15,7 +15,7 @@ from moto.route53.exceptions import (
     NoSuchQueryLoggingConfig,
     QueryLoggingConfigAlreadyExists,
 )
-from moto.core import BaseBackend, BaseModel, CloudFormationModel
+from moto.core import BaseBackend, BaseModel, CloudFormationModel, ACCOUNT_ID
 from moto.utilities.paginator import paginate
 from .utils import PAGINATION_MODEL
 
@@ -474,11 +474,15 @@ class Route53Backend(BaseBackend):
             zones = sorted(zones, key=sort_key)
         return dnsname, zones
 
-    def list_hosted_zones_by_vpc(self, VpcId, VPCRegion):
+    def list_hosted_zones_by_vpc(self, VPCId, VPCRegion, MaxItems=None, NextToken=None):
+
         zone_list = []
         for zone in self.list_hosted_zones():
-            if zone["Config"]["Private"]:
-                this_zone = self.get_hosted_zone(zone["Id"])
+            import pytest
+            pytest.set_trace()
+
+            if zone.private_zone:
+                this_zone = self.get_hosted_zone(zone.id)
                 for vpc in this_zone["VPCs"]:
                     if vpc["VPCId"] == VpcId:
                         this_id = zone["Id"].replace("/hostedzone/", "")
@@ -486,7 +490,7 @@ class Route53Backend(BaseBackend):
                             {
                                 "HostedZoneId": this_id,
                                 "Name": zone["Name"],
-                                "Owner": {"OwningAccount": "XXXXXXXXXXXX"},
+                                "Owner": {"OwningAccount": ACCOUNT_ID},
                             }
                         )
 
--- a/moto/route53/models.py
+++ b/moto/route53/models.py
@@ -478,19 +478,15 @@ class Route53Backend(BaseBackend):
 
         zone_list = []
         for zone in self.list_hosted_zones():
-            import pytest
-            pytest.set_trace()
-
-            if zone.private_zone:
+            if zone.private_zone == "true":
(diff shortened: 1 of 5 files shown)


moto/route53/models.py
```diff
@@ -459,6 +459,24 @@ def sort_key(zone):
...
+            if zone["Config"]["Private"]:
+                this_zone = self.get_hosted_zone(zone["Id"])
+                for vpc in this_zone["VPCs"]:
+                    if vpc["VPCId"] == VpcId:
+                        this_id = zone["Id"].replace("/hostedzone/", "")
+                        zone_list.append(
+                            {
+                                "HostedZoneId": this_id,
+                                "Name": zone["Name"],
+                                "Owner": {"OwningAccount": "XXXXXXXXXXXX"},
```
bblommers, line 474: Moto uses a static account ID across all services that we can use here: `from moto.core import ACCOUNT_ID`

bblommers (changes requested):
Looks good so far @pll! Would you mind adding a test case for this as well? Anywhere in `tests/test_route53` is OK - there are plenty of other examples

moto/route53/models.py
pll: Hi @bblommers,

Thanks for the info on the ACCOUNT_ID, I'll add that in.

I'm trying to write the test now, and that's clearly indicating I have no idea what I'm doing :)

In `route53.models` I was assuming those methods returned similar data structures to what boto3 actually returns (i.e. various types of lists and dicts, etc. as shown in the boto3 docs). That is apparently not the case. e.g. in my new model where I call `self.list_hosted_zones()`, I expected a return structure like this:

```
{
    'HostedZones': [
        {
            'Id': 'string',
            'Name': 'string',
            'CallerReference': 'string',
            'Config': {
                'Comment': 'string',
                'PrivateZone': True|False
            },
            'ResourceRecordSetCount': 123,
            'LinkedService': {
                'ServicePrincipal': 'string',
                'Description': 'string'
            }
        },
    ],
    'Marker': 'string',
    'IsTruncated': True|False,
    'NextMarker': 'string',
    'MaxItems': 'string'
}
```
as documented [here](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/route53.html#Route53.Client.list_hosted_zones)

But instead I get a `<class 'moto.route53.models.FakeZone'>`.  Same with `self.get_hosted_zone()`, which should return a dict structure listing zones and the associated VPC IDs.

I'm happy to keep poking away at this if you can point me in the right direction.  I've figured out I can get at the fake zone id by referencing the object's `.id` property. But have no idea how to get at the other underlying pieces of the expected data structures normally returned by boto3. 

Instead of calling self.list_hosted_zones(), should I be calling something else inside `route53.models` ?

Thanks!
Paul

pll: I think the CREATE_HOSTED_ZONE_RESPONSE in `route53.responses` is missing some stuff.

According to the [AWS Route53 API docs ](https://docs.aws.amazon.com/Route53/latest/APIReference/API_GetHostedZone.html) the response for the `get_hosted_zone()` should be:
```
HTTP/1.1 200
<?xml version="1.0" encoding="UTF-8"?>
<GetHostedZoneResponse>
   <DelegationSet>
      <CallerReference>string</CallerReference>
      <Id>string</Id>
      <NameServers>
         <NameServer>string</NameServer>
      </NameServers>
   </DelegationSet>
   <HostedZone>
      <CallerReference>string</CallerReference>
      <Config>
         <Comment>string</Comment>
         <PrivateZone>boolean</PrivateZone>
      </Config>
      <Id>string</Id>
      <LinkedService>
         <Description>string</Description>
         <ServicePrincipal>string</ServicePrincipal>
      </LinkedService>
      <Name>string</Name>
      <ResourceRecordSetCount>long</ResourceRecordSetCount>
   </HostedZone>
   <VPCs>
      <VPC>
         <VPCId>string</VPCId>
         <VPCRegion>string</VPCRegion>
      </VPC>
   </VPCs>
</GetHostedZoneResponse>
```

But moto only has:
```GET_HOSTED_ZONE_RESPONSE = """<GetHostedZoneResponse xmlns="https://route53.amazonaws.com/doc/2012-12-12/">
   <HostedZone>
      <Id>/hostedzone/{{ zone.id }}</Id>
      <Name>{{ zone.name }}</Name>
      <ResourceRecordSetCount>{{ zone.rrsets|count }}</ResourceRecordSetCount>
      <Config>
        {% if zone.comment %}
            <Comment>{{ zone.comment }}</Comment>
        {% endif %}
        <PrivateZone>{{ zone.private_zone }}</PrivateZone>
      </Config>
   </HostedZone>
   <DelegationSet>
      <NameServers>
         <NameServer>moto.test.com</NameServer>
      </NameServers>
   </DelegationSet>
</GetHostedZoneResponse>"""
```

And looking at `CREATE_HOSTED_ZONE_RESPONSE`, it appears it too, is lacking a bunch of fields in the XML. So it stands to reason if moto isn't capturing the VPC-related data, it won't return it when asked.

I wonder if this is going to turn into one of those yak-shaving rabbit holes I'll never return from :)

bblommers: Fixing the format sounds like the right direction - that's always tricky and error-prone with XML responses.

Feel free to push what you have so far, if you get stuck, happy to have a closer look.

bblommers (approved):
LGTM - thank you @pll!

pll:
@bblommers - Thanks! Happy to help!

PR #4773, opened 2022-01-18
Implement remove_account_from_organization, disallow organization deletion if accounts exist

dfangl:
As seen in the specification of DeleteOrganization (https://docs.aws.amazon.com/organizations/latest/APIReference/API_DeleteOrganization.html), organizations cannot be deleted when accounts are still assigned to it (except for the master account).
This PR implements this behavior, and implements a rudimentary remove_account_from_organization (so testing is easier).
I hope I did not miss any reference onto this account when deleting it though.

Both behaviors are tested with new test cases.


bblommers (approved):
LGTM! Thanks for the PR @dfangl

PR #4775, opened 2022-01-18
Fix : Exception type in cloudwatch get_metric_statistics

KrishnanRanjithkumar:
This PR addresses localstack/localstack#5134. The issue is that exception message thrown currently is bit misleading. 

Changes

1. Changed the `NotImplementedError` exception to `InvalidParameterCombination` exception.
2. Added test case to verify the behaviour.


bblommers (approved):
That TODO looks familiar.. it might have been mine. Thanks for fixing my bug @KrishnanRanjithkumar! 🙂

PR #4782, opened 2022-01-21
set region before executing the script

felipe1982:
region must be selected or the script terminates with an error

I was unable to run the script until I set the default region. I chose us-east-1 arbitrarily.


bblommers (approved):
LGTM! Thanks for the fix @felipe1982

PR #4783, opened 2022-01-21
CloudWatchLogs: create_export_task()

felipe1982:
I am a novice Python programmer, and avid user of AWS, and fan of Moto.

I have need to mock the "create_export_task()" method from the CloudWatch Logs AWS service, as I am testing the functionality of some of my python code which automates this process for me.

I have executed the scaffolding scripts, and generated the boilerplate code, and began to get my hands dirty, when I encountered a `SyntaxError`.

the "`from`" argument to the create_export_task() function is causing issues with
Python as "`from`" is one of the reserved words.

Any suggestions on how I can get over this hurdle ?


bblommers:
Hi @felipe1982, welcome to Moto!

Looking at the docs, it looks like this param should be called `from_time`: https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/logs.html#CloudWatchLogs.Client.create_export_task? 

If `from` was generated by the scaffolding script, there might be a bug there, not sure why that would happen.

In general, as long as the attribute name is descriptive of what it needs to do, it doesn't really matter what it's called: `from_time`, `from_start`, `from_..`.. 
Just make sure that the attribute-name is read correctly (`params.get("from")`) - that name is determine by boto3, so we can not deviate from that.

felipe1982:
I was finally successful in emitting a [`task_id`](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/logs.html#CloudWatchLogs.Client.create_export_task) 🎉 . Is there anything more that I should be doing other than emitted a task id? Should I (or **must** I) also provide functionality for [`describe_export_task`](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/logs.html#CloudWatchLogs.Client.describe_export_tasks) ?

As a novice python developer, I spent many days on trying to make this work! I encountered a few problems which, I believe, should have been avoided. I _think_ the problem may have stemmed from the scaffolding script.

A few problems that I encountered.

### Scaffolding

the `from` problem as stated above does indeed from the scaffolding script. Renaming `from` to `fromTime` (as matching the boto3 _AND_ botocore documentation) solves the problem. I did however keep `params.get("from")` _see next item_

### `params.get()`

`params.get("something")` always returns `None`. This was causing all arguments to have a `None` value. This was provided by the scaffolding script. Other instance methods are using `self._get_param("something")` (note the lack of **`s`** suffix). This is working well now.

### Missing Bucket Exception

In real life, when create\_export\_task() is called on a bucket that does not exist, an exception from botocore is raised. I was trying to simulate similar, and was gathering inspiration from [cloudtrail](./moto/cloudtrail/models.py) and [ec2](./moto/ec2/models.py). I was not able to raise a `MissingBucket` exception. I get a generic `ClientError` from botocore instead.

### uuid4() replaces uuid1()

The Scaffolding provided an example `task_id` using `uuid.uuid1()`, but I can only see 1 other reference to `uuid1()`. However I can see many references to `uuid4()`

bblommers:
> Is there anything more that I should be doing other than emitted a task id? Should I (or **must** I) also provide functionality for [`describe_export_task`](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/logs.html#CloudWatchLogs.Client.describe_export_tasks) ?
> 
No, that's not necessary at all. If someone ever needs the `describe` feature, they can add it in a different PR, but if all you need is the `create` version, than there's no need to go beyond that.

> As a novice python developer, I spent many days on trying to make this work! I encountered a few problems which, I believe, should have been avoided. I _think_ the problem may have stemmed from the scaffolding script.
> 
> ...

Thanks for sharing the feedback @felipe1982. The script is useful to get started, but it definitely needs work in some areas.

> ### Missing Bucket Exception
> 
> In real life, when create_export_task() is called on a bucket that does not exist, an exception from botocore is raised. I was trying to simulate similar, and was gathering inspiration from [cloudtrail](/spulec/moto/blob/master/moto/cloudtrail/models.py#L124). I was not able to raise a `MissingBucket` exception. I get a generic `ClientError` from botocore instead.

Do you mean the internal `moto.s3.exceptions.MissingBucket`? This is as expected. Moto only intercepts the HTTP request that botocore makes. When we want to mock an error scenario, we'll return the XML or JSON in a format that botocore understands and can convert into either a generic ClientError, or a specific subclass.
So our internal naming has no bearing on what error botocore does - we just have to make sure that the XML/JSON response is the same as what AWS HTTP API would return.


> ### `make test_server` failing
> 
I couldn't reproduce this (unfortunately?). The CI seems to be happy with the ServerMode tests as well now :+1:

felipe_test.py
```diff
@@ -0,0 +1,47 @@
+import boto3
```
bblommers, line 1: This file is not necessary, if all the tests are already covered in `tests/test_logs`

tests/test_logs/test_server.py
```diff
@@ -0,0 +1,13 @@
+"""Test different server responses."""
+import sure  # noqa # pylint: disable=unused-import
+
+import moto.server as server
+
+
+def test_logs_list():
```
bblommers, line 7: The scaffolding-script creates this automatically, but it's not required.

It can be useful if you want to have fine-grained control over the HTTP request/response, but if everything you're testing is covered by the unit test already, there's no point in adding it.

bblommers (changes requested):

bblommers (approved):
LGTM! Thanks for the PR @felipe1982, much appreciated 👍

Issue #4785, opened 2022-01-22
moto/route53 does not correctly handle private hosted zones


pll:
moto/route53 does not correctly handle private hosted zones with `create_hosted_zone()` such that when calling `list_hosted_zones()` or `get_hosted_zone()` with a specific zone ID, the associated VPC information is not available.

Further, calling `create_hosted_zone()` and requesting a private hosted zone is allowed without passing in an associated VPC configuration block, which is incorrect behavior.

pll:
I've attempted to fix the above here: https://github.com/spulec/moto/pull/4786

PR #4786, opened 2022-01-22
moto/route53 does not correctly handle private hosted zones #4785

pll:
models.py:
- Route53Backend.create_hosted_zone now receive vpcid, vpcregion
- Passes into FakeZone.__init___

responses.py:
- Class Route53 now handles the existence of a VPC key in the returned "HOSTEDZONECONFIG
- Cleaned up Route53.list_or_create_hostzone_response() for better readability
- If a CreateHostedZoneRequest comes in without a VPC block we now throw a KeyError
- Added VPC block to GET_HOSTED_ZONE_RESPONSE

test_route53.py:
- import mock_ec2 to create mock VPCs
- test_hosted_zone_private_zone_preserved_boto3()
  - now creates a VPC and a CreateHostedZoneRequest with a VPC block
  - return is now tested:
    - for VPCs section
    - that VPCId and VPCRegion are non-empty
    - that VPCId and VPCRegion match what we created the VPC with and what was sent to the initial CreateHostedZoneRequest
- test_list_hosted_zones_by_name()
  - still creates 3 zones,
  - but 1 is a true Private zone with VPC block
  - others are Public zone by virtue of Private=False
  - return for Private=True zone tested same as test_hosted_zone_private_zone_preserved_boto3()
  - other zones with Private=False set are tested for that.


moto/route53/responses.py
```diff
@@ -30,32 +30,47 @@ class Route53(BaseResponse):
...
+            zone_request = elements["CreateHostedZoneRequest"]
+            if "HostedZoneConfig" in zone_request:
+                zone_config = zone_request["HostedZoneConfig"]
+                comment = zone_config["Comment"]
                 try:
                     # in boto3, this field is set directly in the xml
-                    private_zone = elements["CreateHostedZoneRequest"][
-                        "HostedZoneConfig"
-                    ]["PrivateZone"]
+                    private_zone = zone_config["PrivateZone"]
```
bblommers, line 45:
```suggestion
                    private_zone = zone_config["PrivateZone"].lower() == "true"
```
Ideally this returns a boolean - as the `except`-clause also returns a boolean.

moto/route53/responses.py
```diff
@@ -30,32 +30,47 @@ class Route53(BaseResponse):
...
+                    # if a VPC subsection is only included in xmls
+                    # params when private_zone=True, see boto:
+                    # boto/route53/connection.py
+                    private_zone = "VPC" in zone_request
             else:
                 comment = None
                 private_zone = False
 
-            name = elements["CreateHostedZoneRequest"]["Name"]
+            if private_zone == "true":
```
bblommers, line 55:
```suggestion
            if private_zone:
```
Which means this can be simplified

tests/test_route53/test_route53.py
```diff
@@ -390,6 +403,21 @@ def test_hosted_zone_private_zone_preserved_boto3():
     len(hosted_zones["HostedZones"]).should.equal(1)
     hosted_zones["HostedZones"][0]["Config"]["PrivateZone"].should.equal(True)
 
+    # create_hosted_zone statements with  PrivateZone=True,
+    # but without a _valid_ vpc-id should fail.
+    conn = boto3.client("route53", region_name=region)
+    with pytest.raises(KeyError) as e:
```
bblommers, line 409: Moto in ServerMode will only throw exceptions of type `ClientError`, so we should catch that instead.
The assertion then becomes something like `e.value.response["Error"]["Message"] == expected_msg`

Changing this should get the tests to pass

bblommers (changes requested):
Thanks for the PR @pll - please see my comments

moto/route53/responses.py
pll: Hi Bert,

I actually had this as `if private_zone:` originally. However, it doesn't actually return a boolean. `private_zone` was set to `"false"`, which was still evaluating as `True`, which is why I ended up changing it to the expanded comparison using strings.  I'll double check it and post the output from the debugger.

Thanks!
Paul

tests/test_route53/test_route53.py
pll: Oh, okay.  I'm not sure how to test that locally on my laptop. I'll give it a whirl. Do I just fire up the moto_server and then just run the server tests ?

bblommers: Firing up moto_server to start it, yes. Then you can use `TEST_SERVER_MODE=true pytest tests/test_route53` to run the same tests against that server.

moto/route53/responses.py
pll: @bblommers, here's the trace from the debugger:

```> /Users/pll/dev/github/moto/moto/route53/responses.py(47)list_or_create_hostzone_response()
-> private_zone = zone_config["PrivateZone"]
(Pdb) zone_config["PrivateZone"]
'true'
(Pdb) type(zone_config["PrivateZone"])
<class 'str'>
```

So something somewhere is setting this to a string, not a boolean, unfortunately...

tests/test_route53/test_route53.py
pll: Awesome, that's what I suspected. Looks like the Makefile has a `test_server` target as well, so I'm assuming I just need to fire it up on port 5000 (assuming the developer FAQ is current).

I'll see what I can do tonight.

moto/route53/responses.py
pll: @bblommers 

Any thoughts on the above ? Good enough to leave as a string comparison? Or should do something else?

bblommers: Fair enough - a string comparison makes sense then :+1:

bblommers (approved):
LGTM - thank you for your contribution @pll!

pll:
@bblommers Thanks! Quite a learning curve, but glad I can help out a little!  Thanks for all the help!

Now back to fixing what I originally intended  :)

TTYL!

PR #4788, opened 2022-01-24
test: add missing test for delete_log_stream

felipe1982:
This PR tests the CW logs method `delete_log_stream()` which was missing tests. This checks the happy path:

* create log group
* create log stream
* describe log streams
* assert one exists
* delete log stream
* assert zero exist


bblommers (approved):
LGTM! Thanks for adding this test @felipe1982

PR #4791, opened 2022-01-25
When a timer fires, it should also schedule a decision

JimNero009:
Hello again! A bug fix for the implementation of the timers in https://github.com/spulec/moto/pull/4713

As not explicitly pointed out in the API docs, but mentioned instead [here](https://docs.aws.amazon.com/amazonswf/latest/developerguide/swf-dg-timers.html), when a timer has fired, a decision task should be scheduled. If this doesn't happen, then the workflow will not progress, as there is nothing notifying it that it should do something.


bblommers (approved):
LGTM! Thanks for the fix @JimNero009 :+1:

PR #4796, opened 2022-01-26
Missing VersionID for versioned buckets in API

rrader:
Regarding the API docs, https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3.html#S3.Client.get_object_tagging `get_object_tagging` method (as well as some others get_object related methods) should return `VersionId` in response.

The current implementation of moto does not return this value which makes it incompatible with real API for this use case.

This PR makes a small change by adding `x-amz-version-id` header to the response in case version exists.


bblommers (approved):
LGTM! Thanks for the PR @rrader

PR #4801, opened 2022-01-27
Add host to batch

morsecodist:
Ports https://github.com/spulec/moto/pull/4658 to batch

### DIFF
--- a/tests/test_batch/test_batch_jobs.py
+++ b/tests/test_batch/test_batch_jobs.py
@@ -166,7 +166,7 @@ def test_list_jobs():
         job.should.have.key("createdAt")
         job.should.have.key("jobDefinition")
         job.should.have.key("jobName")
-        job.should.have.key("status").equals("RUNNABLE")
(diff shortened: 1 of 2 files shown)


tests/test_batch/test_batch_jobs.py
```diff
@@ -166,7 +166,7 @@ def test_list_jobs():
         job.should.have.key("createdAt")
         job.should.have.key("jobDefinition")
         job.should.have.key("jobName")
-        job.should.have.key("status").equals("STARTING")
+        job.should.have.key("status").equals("RUNNABLE")
```
morsecodist, line 169: It seems like my change introduces a slight lag in starting the jobs. This test seems a little brittle since it is effected by timing but I see no reason why `RUNNABLE` is invalid here

bblommers (approved):
LGTM! Thanks for the PR (again) @morsecodist :slightly_smiling_face:

PR #4802, opened 2022-01-27
Add TEST_SERVER_MODE_ENDPOINT env var documentation

ikkala:
Add missing TEST_SERVER_MODE_ENDPOINT env var documentation


bblommers (approved):
Makes sense - appreciate the follow-up :+1:

Issue #3909, opened 2021-05-06
mock_dynamodb2 - Inconsistent query result between the mock and dynamodb in terms of missing LastEvaluatedKey


Nivmorabin:
Hi,

I experience an inconsistent query result between the mock_dynamodb2 and the actual live dynamodb query result returned from dynamodb in terms of missing LastEvaluatedKey.

I suspect that the limit doesn't apply before the FilterExpression as mentioned in the [documentation](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Query.html)

Please run the following code sample, the query result indeed return 0 items but when I'm running this same query against the actual live dynamodb I receive 0 items with LastEvaluatedKey containing the first item scanned (notice the Limit 1).

Expected result (the first Item I've added in table.put_item in the sample below)
```
{ Items:[], LastEvaluatedKey: {"pk": "TEST", "sk": "ENV#TEST_ENV#FILE#file1", "depth": 4, "file_path": "D/dir1/dir2/dir3/file1", "driveId": "D", "myId": "file1"} }
```
Actual result - the result is correct, but I expect to query with the LastEvaluatedKey a few iterations until I receive LastEvaluatedKey: None due to the Limit 1 (and that's also what returned from dynamodb).
```
{ Items:[], LastEvaluatedKey: None } 
```

moto version 2.0.4

```
import boto3
from moto import mock_dynamodb2
from boto3.dynamodb.conditions import Attr, And, Key
@mock_dynamodb2
def test():
    dynamo_db_resource = boto3.resource("dynamodb", region_name="us-west-2")
    dynamo_db_resource.create_table(TableName="TestTable", KeySchema=[{"AttributeName": "pk", "KeyType": "HASH"},
                                                               {"AttributeName": "sk", "KeyType": "RANGE"}],
                                    AttributeDefinitions=[{"AttributeName": "pk", "AttributeType": "S"},
                                                          {"AttributeName": "sk", "AttributeType": "S"}],
                                    ProvisionedThroughput={"ReadCapacityUnits": 5, "WriteCapacityUnits": 5},
                                    )
    table = dynamo_db_resource.Table("TestTable")

    table.put_item(Item={"pk": "TEST", "sk": "ENV#T_E#FILE#file1", "depth": 4, "file_path": "D/dir1/dir2/dir3/file1", "driveId": "D", "myId": "file1"})
[...]

PR #4805, opened 2022-01-28
dynamodb: filter expression should applied after query/scan. fixes spulec/moto#3909

dpretty:
As discussed in spulec/moto#3909 and as per the AWS documentation:
- https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Query.html#Query.FilterExpression
- https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Scan.html#Scan.FilterExpression

A DynamoDB filter expression should be applied after the query or scan completes.

This PR includes a test case which failed prior to the changes made.

### DIFF
--- a/moto/dynamodb2/models/__init__.py
+++ b/moto/dynamodb2/models/__init__.py
@@ -839,9 +839,6 @@ class Table(CloudFormationModel):
             index = self.get_index(index_name)
             for result in results:
                 index.project(result)
-        if projection_expression:
-            for result in results:
-                result.filter(projection_expression)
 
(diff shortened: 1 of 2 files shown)


bblommers (changes requested):
Hi @dpretty, thanks for the PR! That's a very clean fix. 
Happy to merge once we get the build to pass

dpretty:
Thanks @bblommers - hopefully good to go now. I had forgotten to run black after that last commit.

bblommers (approved):
Thanks @dpretty!

PR #4808, opened 2022-01-28
ELBv2 improvements

### LINKED ISSUE #4711: Feature Request ELBv2 Add and Remove Listener Certificates
Please can I request the following features are added to the ELBv2 service.
1. Replace a listener default certificate using the modify_listener function.

```
response = self.client.modify_listener(
  Certificates=[
          {
              'CertificateArn': 'string'
          },
      ],
  ListenerArn='string'
)
```

bblommers:
Closes #4711 

Improvements:
 - Add ProtocolVersion to the response (defaults to HTTP1)
 - Expand the number of valid protocols inline with what AWS considers valid atm
 - Move modify_target_group_attributes to `models.py` - no logic changes
 - Fixes a bug so that certificates are always persisted after a `modify_target_group`. (Before, they were only persisted if the protocol was changed to `HTTPS` at the same time

### DIFF
--- a/moto/elbv2/models.py
+++ b/moto/elbv2/models.py
@@ -951,8 +951,6 @@ Member must satisfy regular expression pattern: {}".format(
         return target_group
 
     def modify_target_group_attributes(self, target_group_arn, attributes):
-        if "matcher" in attributes:
-            pass
         target_group = self.target_groups.get(target_group_arn)
         if not target_group:
             raise TargetGroupNotFoundError()
--- a/moto/elbv2/models.py
+++ b/moto/elbv2/models.py
@@ -62,6 +62,7 @@ class FakeTargetGroup(CloudFormationModel):
         vpc_id,
         protocol,
         port,
+        protocol_version=None,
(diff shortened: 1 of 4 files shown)


rickbowden:
Hi, even after upgrading to version 3.0.3.dev4 I still get an exception when attempting to attach a certificate using the ModifyListener operation.

botocore.errorfactory.CertificateNotFoundException: An error occurred (CertificateNotFound) when calling the ModifyListener operation:

The certificate I am attempting to attach is an IAM server certificate, and not an ACM certificate.

bblommers:
Thanks for letting us know @rickbowden - we'll change the logic here to also accept IAM certificates

bblommers:
This should now be fixed as of moto >= 3.0.3.dev7 @rickbowden - with #4823 merged we now accept IAM certificates as well.
Can you verify this fixes your problem?

rickbowden:
@bblommers I've tested this and it works now. Thanks very much for sorting this.

PR #4810, opened 2022-01-29
Route53 improvements

### LINKED ISSUE #3879: Route53 not truncating long list of records
When there are a lot of Route53 records, `boto3` [`list_resource_record_sets`](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/route53.html#Route53.Client.list_resource_record_sets) returns up to 300 resource records and sets the `Truncate` attribute to `True`. 

The only implementation of `Truncate` was done in the PR #2661 to fix #2660. Which is a patch for the cases where there are not enough records to truncate them.

I need to test some code that processes that truncation, so it would be awesome if `moto` mimics AWS behaviour.

```python
@pytest.fixture(name="route53")
def route53_(_aws_credentials: None) -> Any:
    """Configure the boto3 Route53 client."""
    with mock_route53():
        yield boto3.client("route53")

def test_update_creates_route53_instances_when_there_are_a_lot(route53: Any) -> None:
    hosted_zone = route53.create_hosted_zone(
        Name="example.com", CallerReference="Test"
    )["HostedZone"]
    hosted_zone_id = re.sub(".hostedzone.", "", hosted_zone["Id"])
    for ip in range(0, 500):
        route53.change_resource_record_sets(
            ChangeBatch={
                "Changes": [
                    {
                        "Action": "CREATE",
                        "ResourceRecordSet": {
                            "Name": "example.com",
                            "ResourceRecords": [
                                {
                                    "Value": f"192.0.2.{ip}",
                                },
                            ],
                            "TTL": 60,
                            "Type": "A",
                        },
                    },
                ],
                "Comment": "Web server for example.com",
            },
            HostedZoneId=hosted_zone_id,
        )
        records = route53.list_resource_record_sets(
                HostedZoneId=hosted_zone["Id"],
        )

    instances = records["ResourceRecordSets"]
    assert len(instances) == 300
```

In this example, the last assertion would fail, as `len(instances) == 500`.

Thanks!

### LINKED ISSUE #3897: support route53 "create-reusable-delegation-set" API
As far as I can tell, this API is not yet supported.

Thanks.

bblommers:
Various Route53 features, + fixes to make the behaviour more in-line with what AWS expects.

Also fixes some ELB bugs to get the Terraform tests to work

Closes #3897
Closes #3879


bblommers:
Going to close this, as it's far from ready yet

PR #4818, opened 2022-01-31
Test sns publish_batch with raw message delivery

simonschmidt:
Add failing test as an example for reported issue

### DIFF
--- a/tests/test_sns/test_publish_batch.py
+++ b/tests/test_sns/test_publish_batch.py
@@ -152,3 +152,42 @@ def test_publish_batch_to_sqs():
             ],
         }
     )
+
+
+# @pytest.mark.xfail(reason="MessageAttributes not transformed for sqs subscription")
+@mock_sqs
+@mock_sns
+def test_publish_batch_to_sqs_raw():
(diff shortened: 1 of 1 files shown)


tests/test_sns/test_publish_batch.py
```diff
@@ -152,3 +152,41 @@ def test_publish_batch_to_sqs():
...
+        {"Id": "1", "Message": "foo",},
+        {
+            "Id": "2",
+            "Message": "bar",
+            "MessageAttributes": {"a": {"DataType": "String", "StringValue": "v"}},
+        },
+    ]
+    resp = client.publish_batch(TopicArn=topic_arn, PublishBatchRequestEntries=entries,)
+
+    resp.should.have.key("Successful").length_of(2)
```
simonschmidt, line 184: This is where it fails as the message with Id=2 is forgotten about completely

bblommers:
Thanks for adding this test case @simonschmidt, that's really useful :+1:

bblommers:
A fix for this issue is now part of moto >= 3.0.5.dev16

PR #4820, opened 2022-02-02
FIX: Cognito-idp

alastairmccormack:
Minor fixes to cognito-idp:

* Added missing `TokenType` to  all `AuthenticationResults`
* `REFRESH_TOKEN_AUTH` flow is an alias for `REFRESH_TOKEN` flow, so can be handled by the same logic
* During token refresh, access_token and id_token were incorrectly switched

### DIFF
--- a/moto/cognitoidp/models.py
+++ b/moto/cognitoidp/models.py
@@ -1509,7 +1509,7 @@ class CognitoIdpBackend(BaseBackend):
                     "TokenType": "Bearer",
                 }
             }
-        elif auth_flow in ["REFRESH_TOKEN", "REFRESH_TOKEN_AUTH"]:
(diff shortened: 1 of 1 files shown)


alastairmccormack:
Unit tests for 3.6 fail in one place - I'm not able to recreate locally. Can this test be re-run?

bblommers:
Restarting the build resulted in even more failures - looks like the CI is a bit unstable ATM.
Will try to restart it again later.

bblommers (changes requested):
Would you mind merging in master @alastairmccormack? The latest changes should fix the build, I'd be happy to merge once that happens

bblommers (approved):
LGTM! Thanks @alastairmccormack

PR #4824, opened 2022-02-03
Addition of bucket name when creating a key object

pinzon:
This PR addresses issue localstack/localstack#5121 where the user describes that having the same key_name in different bucket triggers the objects to share the same tags.

Changes:
- Addition of bucket name parameter during key object creation.
- Test creation that asserts that the issue is resolved.


bblommers (changes requested):
Nice find @pinzon. Could you merge the latest master in this PR? That should fix the build

bblommers (approved):
LGTM!

PR #4826, opened 2022-02-04
SES: Fix GetSendStatistics response template

viren-nadkarni:
This PR makes following fixes to the response for GetSendStatistics call:

- Change XML tags of statistics items to comply with upstream AWS response.
  Closes https://github.com/localstack/localstack/issues/5074
  AWS uses the XML tag `<member>` to represent each statictics object, not `<item>`

- Change timestamp representation to ISO 8601 format.
  Closes https://github.com/localstack/localstack/issues/3645.
  Currently the str representation of Python datetime is used, which causes various clients to fail timestamp parsing since they expect an ISO timestamp.

### DIFF
--- a/tests/test_ses/test_server.py
+++ b/tests/test_ses/test_server.py
@@ -1,3 +1,6 @@
+from datetime import datetime
+import re
+
 import sure  # noqa # pylint: disable=unused-import
 
 import moto.server as server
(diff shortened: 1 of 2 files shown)


bblommers (changes requested):
Hi @viren-nadkarni, would you mind merging in the latest fixes from master? That should fix the build - happy to merge once it all passes

tests/test_ses/test_server.py
```diff
@@ -13,3 +16,16 @@ def test_ses_list_identities():
...
+    backend = server.create_backend_app("ses")
+    test_client = backend.test_client()
+
+    res = test_client.get("/?Action=GetSendStatistics")
+    res.data.should.contain(b"GetSendStatisticsResponse")
+
+    # Timestamps must be in ISO 8601 format
+    groups = re.search("<Timestamp>(.*)</Timestamp>", res.data.decode("utf-8"))
+    timestamp = groups.groups()[0]
+    datetime.fromisoformat(timestamp)
```
bblommers, line 31: Looks like the tests are still failing because the `fromisoformat` is only available as of Python 3.7, and we're still supporting 3.6 - you'll have to find an alternative for this method.

bblommers (changes requested):

tests/test_ses/test_server.py
viren-nadkarni: Switched to 3.6 compatible `strptime` in [0591514](https://github.com/spulec/moto/pull/4826/commits/05915141b7d619873448afdda7087793838ba776)

bblommers (approved):
LGTM! Thanks for the fix @viren-nadkarni

PR #4827, opened 2022-02-04
Increase Batch timeouts, in case Docker takes a while to start

bblommers:
Github Actions started failing on batch tests, exposing:
 - A timing issue where failing a job too late would not actually fail it
 - Tests that were too dependent on Docker being fast - the moment Docker start-up takes longer than the usual 0.5 seconds, tests would fail.

Playing around with the `seconds_to_wait`-parameter ensures that the tests now wait a long time, in case Docker is slow.

This also fixes some SQS tests that would fail if too many queues were created in parallel.

### DIFF
--- a/moto/settings.py
+++ b/moto/settings.py
@@ -1,6 +1,9 @@
 import json
 import os
 
+from functools import lru_cache
+
+
 TEST_SERVER_MODE = os.environ.get("TEST_SERVER_MODE", "0").lower() == "true"
 INITIAL_NO_AUTH_ACTION_COUNT = float(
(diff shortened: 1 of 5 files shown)


moto/settings.py
```diff
@@ -55,6 +58,7 @@ def moto_server_port():
     return os.environ.get("MOTO_PORT") or "5000"
 
 
+@lru_cache()
```
bblommers, line 61: GithubActions would sometimes take 15 seconds to make the CURL call to find the docker host.
Caching this at least ensures it only has to happen once per system

PR #4828, opened 2022-02-04
ec2: Add GpuInfo to describe_instance_types

edisongustavo:
## Testing

I have inspected the return result and saw that the instance types that should have Gpus do have the correct values.

For example, looking at the entry for `g4ad.8xlarge`:

```
{'GpuInfo': {'Gpus': [{'Count': 2,
                       'Manufacturer': 'AMD',
                       'MemoryInfo': {'SizeInMiB': 8192},
                       'Name': 'Radeon Pro V520'}],
             'TotalGpuMemoryInMiB': 16384},
 'InstanceStorageInfo': {'TotalSizeInGB': 1200},
 'InstanceType': 'g4ad.8xlarge',
 'MemoryInfo': {'SizeInMiB': 131072},
 'ProcessorInfo': {'SupportedArchitectures': ['x86_64']},
 'VCpuInfo': {'DefaultCores': 16,
              'DefaultThreadsPerCore': 2,
              'DefaultVCpus': 32}}
```

### DIFF
--- a/moto/ec2/responses/instances.py
+++ b/moto/ec2/responses/instances.py
@@ -871,6 +871,23 @@ EC2_DESCRIBE_INSTANCE_TYPES = """<?xml version="1.0" encoding="UTF-8"?>
                     {% endfor %}
                 </supportedArchitectures>
             </processorInfo>
+            {% if instance_type.get('GpuInfo', {})|length > 0 %}
+            <gpuInfo>
+                <gpus>
+                    {% for gpu in instance_type.get('GpuInfo').get('Gpus') %}
+                    <item>
+                        <count>{{ gpu['Count']|int }}</count>
+                        <manufacturer>{{ gpu['Manufacturer'] }}</manufacturer>
+                        <memoryInfo>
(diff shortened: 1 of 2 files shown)


bblommers (changes requested):
Hi @edisongustavo, thanks for the PR! Would you mind adding a test case for this, to verify that the values are send through correctly? The test cases in `tests/test_ec2/test_instance_types.py` should be a good starting point

edisongustavo:
Hey @bblommers , sure I will work on that right on! Thanks for taking a look at the PR.

edisongustavo:
@bblommers , I have added a unit test where I `assert` on the entire result of `GpuInfo`, this way anything that gets added will have to be reflected in the test so that if people in the future improve the template, they also have to change the test.

If you think I shouldn't go this route and use the `.should...` statements, then I'll gladly rewrite the test.

bblommers (approved):
LGTM! Thanks for adding the test, @edisongustavo 👍

PR #4830, opened 2022-02-04
Allow "True" in SQS Queue boolean attributes

pinzon:
This PR addresses issue localstack/localstack#5330. CDK sends the boolean parameters with a capitalized string but the behaviour should be the same as if it send a lowercase string.

### DIFF
--- a/moto/sqs/models.py
+++ b/moto/sqs/models.py
@@ -327,7 +327,7 @@ class Queue(CloudFormationModel):
             if key in integer_fields:
                 value = int(value)
             if key in bool_fields:
-                value = value == "true" or value == "True"
+                value = value.lower() == "true"
 
             if key in ["Policy", "RedrivePolicy"] and value is not None:
                 continue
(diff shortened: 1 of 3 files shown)


moto/sqs/models.py
```diff
@@ -327,7 +327,7 @@ def _set_attributes(self, attributes, now=None):
             if key in integer_fields:
                 value = int(value)
             if key in bool_fields:
-                value = value == "true"
+                value = value == "true" or value == "True"
```
bblommers, line 330:
```suggestion
                value = value.lower() == "true"
```
This might be a better approach, to catch all possibilities - assuming value is always a string?

bblommers (changes requested):
Thanks for the PR @pinzon. Looks like one of our existing tests was also using `True`, and is now failing because of this validation: `test_create_queue_with_different_attributes_fail`

As with the other PR, you'll also have to merge in master to get the build to pass - apologies for that

bblommers (approved):
LGTM! Thanks @pinzon

PR #4832, opened 2022-02-04
Add redshift-data: cancel_statement, describe_statement, execute_statement, get_statement_result

vincbeck:
Add following actions for redshift-data:
- cancel_statement
- describe_statement
- execute_statement
- get_statement_result


moto/redshiftdata/models.py
```diff
@@ -0,0 +1,192 @@
...
+        cluster_identifier,
+        database,
+        db_user,
+        query_parameters,
+        query_string,
+        secret_arn,
+    ):
+        now = iso_8601_datetime_without_milliseconds(datetime.now())
+
+        self.Id = str(uuid.uuid4())
```
bblommers, line 23: Variables and attribute-names should be `snake_case` everywhere, to be consistent

moto/redshiftdata/models.py
```diff
@@ -0,0 +1,192 @@
...
+            database=database,
+            db_user=db_user,
+            query_parameters=parameters,
+            query_string=sql,
+            secret_arn=secret_arn
+        )
+        self.statements[statement.Id] = statement
+        return statement
+
+    def get_statement_result(self, statement_id):
```
bblommers, line 145: Can we put an explanation on what this method here, as part of the docstring?
```
def get_statement_result(..)
"""
Return static statement result, because ..

"""
```
Setting the docstring ensures that the explanation automatically becomes part of the docs when calling `update_implementation_coverage`

moto/redshiftdata/models.py
```diff
@@ -0,0 +1,192 @@
...
+        )
+
+
+def _validate_uuid(uuid):
+    match = re.search(r'^[a-z0-9]{8}(-[a-z0-9]{4}){3}-[a-z0-9]{12}(:\d+)?$', uuid)
+    if not match:
+        raise ValidationException("id must satisfy regex pattern: ^[a-z0-9]{8}(-[a-z0-9]{4}){3}-[a-z0-9]{12}(:\\d+)?$")
+
+
+# For unknown reasons I cannot use the service name "redshift-data" as I should
```
bblommers, line 186: Yes - that is a known issue, there are several services that have the same problem. Using the `redshift`-service is a good workaround until boto3 fixes this

moto/redshiftdata/models.py
```diff
@@ -0,0 +1,192 @@
...
+    def execute_statement(
+        self,
+        cluster_identifier,
+        database,
+        db_user,
+        parameters,
+        secret_arn,
+        sql,
+    ):
+        statement = Statement(
```
bblommers, line 134: Ideally we have a docstring here as well, to explain that there is no `redshift`-integration (yet). So the database-parameter does not have validation of any kind.

tests/test_redshiftdata/test_redshiftdata.py
```diff
@@ -0,0 +1,82 @@
+import boto3
```
bblommers, line 1: Woul dyou midn extending the tests here to also include positive tests, i.e. using the boto3-client to cancel/describe/execute a statement?
I know the functionality is covered in `test_server.py`, but the majority of the users will use the boto3-client to call these methods - so we should have test cases to verify it works with boto3.

moto/redshiftdata/urls.py
```diff
@@ -0,0 +1,10 @@
+from .responses import RedshiftDataAPIServiceResponse
+
+url_bases = [
+    r"https?://redshift-data\.(.+)\.amazonaws\.com",
```
bblommers, line 4: Whenever the URLs change, you should also call `scripts/update_backend_index.py`.

There is a hardcoded index of all these URL's to speed things up, and ensure Moto doesn't have to load every module to get the list of URL's - this script will automatically update this index.

bblommers (changes requested):
Hi @vincbeck, thanks for the PR! Please see my comments

moto/redshiftdata/urls.py
vincbeck: That's good to know! Thank you

moto/redshiftdata/models.py
vincbeck: Sure. Done

moto/redshiftdata/models.py
vincbeck: Good point. I added the docstring

moto/redshiftdata/models.py
vincbeck: Done

tests/test_redshiftdata/test_redshiftdata.py
vincbeck: Sure, I added some positive tests as well.

moto/__init__.py
```diff
@@ -145,6 +145,11 @@ def f(*args, **kwargs):
 mock_xray = lazy_load(".xray", "mock_xray")
 mock_xray_client = lazy_load(".xray", "mock_xray_client")
 mock_wafv2 = lazy_load(".wafv2", "mock_wafv2")
+mock_sdb = lazy_load(".sdb", "mock_sdb")
```
bblommers, line 148: This looks like a dodgy merge - I changed the ordering earlier to be alphabetical, so SDB and ElastiCache are already defined above

vincbeck: My bad for `backend_index,py`. It should be fine now. I also fixed `__init__.py`

bblommers (approved):
LGTM! Thanks for the PR @vincbeck, much appreciated

Issue #4623, opened 2021-11-23
The cognito-idp initiate_auth method raises a TypeError


JamieMcKernanKaizen:
### Dependencies
This is with Python 3.9 running on Ubuntu 20.04.3 LTS 64-bit using the Python mocks. I'm testing with PyTest.

My dependencies (using a poetry environment):
```
boto3         1.20.11
botocore      1.23.11
moto          2.2.16
pytest        6.2.5
python-jose   3.3.0
```

### My reproducible code example
`conftest.py`
```python
import os

import pytest
import boto3
from moto import mock_cognitoidp


COGNITO_USER_USERNAME = "pytest"
COGNITO_USER_PASSWORD = "testPassword"


@pytest.fixture
def aws_credentials():
    """Mocked AWS Credentials for moto."""
    os.environ["AWS_ACCESS_KEY_ID"] = "testing"
    os.environ["AWS_SECRET_ACCESS_KEY"] = "testing"
    os.environ["AWS_SECURITY_TOKEN"] = "testing"
    os.environ["AWS_SESSION_TOKEN"] = "testing"


@pytest.fixture
def cognito_idp_client(aws_credentials):
    with mock_cognitoidp():
        client = boto3.client("cognito-idp")
        yield client


@pytest.fixture
def create_user_pool(cognito_idp_client):
    user_pool_response = cognito_idp_client.create_user_pool(PoolName="test-user-pool")
    pool_id = user_pool_response["UserPool"]["Id"]

    pool_client_response = cognito_idp_client.create_user_pool_client(
        UserPoolId=pool_id,
        ClientName="pytest",
    )

    pool_app_client_id = pool_client_response["UserPoolClient"]["ClientId"]

    yield {
        "user_pool_id": pool_id,
        "pool_app_client_id": pool_app_client_id,
    }


@pytest.fixture
def cognito_user(cognito_idp_client, create_user_pool):
    cognito_idp_client.admin_create_user(
        UserPoolId=create_user_pool["user_pool_id"],
        Username=COGNITO_USER_USERNAME,
    )

    cognito_idp_client.admin_set_user_password(
        UserPoolId=create_user_pool["user_pool_id"],
        Username=COGNITO_USER_USERNAME,
        Password=COGNITO_USER_PASSWORD,
        Permanent=True,
    )

    yield {
        "pool_app_client_id": create_user_pool["pool_app_client_id"],
    }


@pytest.fixture
[...]

PR #4836, opened 2022-02-07
Restrict initiate_auth and admin_initiate_auth AuthFlow

alastairmccormack:
The API documentation for `initiate_auth` and `admin_initiate_auth` Cognito IDP methods are ambiguous about the supported AuthFlow parameters.

As stated in #4623, AWS have confirmed via email what the valid attributes are. I have also validated the behaviour through testing.

This PR restricts the AuthFlow parameters to the known values for initiate_auth and admin_initiate_auth methods and raises `InvalidParameterException` for invalid or inappropriate auth_flow values.

In the test case raised by #4623, moto will properly emulate AWS by raising an `InvalidParameterException('Initiate Auth method not supported')` exception to the user for using `ADMIN_USER_PASSWORD_AUTH` with the regular user `initiate_auth` endpoint.

### DIFF
--- a/moto/cognitoidp/models.py
+++ b/moto/cognitoidp/models.py
@@ -1126,7 +1126,9 @@ class CognitoIdpBackend(BaseBackend):
             }
         }
 
-    def _validate_auth_flow(self, auth_flow: str, valid_flows: typing.List[AuthFlow]) -> AuthFlow:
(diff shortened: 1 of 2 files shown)


bblommers (approved):
LGTM! Thanks for the investigation and fix, @alastairmccormack

PR #4837, opened 2022-02-07
Correctly use the filter for describe_..._prefix_lists

dfangl:
This change affects both ec2s describe_managed_prefix_lists and describe_prefix_lists, as they boil down to the same method.

As you can see, in the current filter method, the result of filter_resources is (incorrectly) discarded, and replaced by describe_tag_filter (since both operations act on copies of the original object).

This PR fixes this behavior by calculating both operations on the currently valid result (so describe_tag_filter gets the already filtered result).

Also added a test for this behavior.

### DIFF
--- a/moto/ec2/models.py
+++ b/moto/ec2/models.py
@@ -5566,8 +5566,8 @@ class ManagedPrefixListBackend(object):
 
         result = managed_prefix_lists
         if filters:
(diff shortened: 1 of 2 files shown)


bblommers (approved):
LGTM! Thanks for the fix @dfangl

PR #4851, opened 2022-02-11
[LocalStack] Fixes to secretsmanager's PutSecretValue, CreateSecret, DeleteSecret

MEPalma:
LocalStack recently received some issue reports about the handling of PutSecretValue requests to secretsmanager. After validating the behaviour of AWS's secretsmanager with respect to PutSecretValue, CreateSecret and DeleteSecret requests, I propose the following fixes.
-- Please note that all tests introduced, or updated, were verified agains AWS before the opening of the pull request.

#### Response

I updated moto's implementation of `PutSecretValue` to return the version of the secret resulting from the execution of the request itself. This is in contrast to the original logic, which returned the current/default version of the secret.
The fix consists in handling the derivation of the secret version's `VersionId` (either from validation of the provided `ClientRequestToken` value, or generating a new one) in the `PutSecretValue` function. Hence, passing this `VersionId` to `__add_secret` (as expected), and later retrieving the version to compute the response. The response is compiled as before through the `to_short_dict` function, which now accepts an optional `version_id` parameter, defaulting to the current version if unspecified.

#### VersionStages

Moto's implementation of `PutSecretValue` to prevent setting as new default/current version of the secret the same version that resulted in the introduction of a new `VersionStages` collection. This was achieved by updating `__add_secret` function to only update its default version iff the submitted `VersionStages` contain the `AWSCURRENT` value.
This behaviour can be verified by forwarding to AWS's secretsmanager a `GetSecretValue` by `SecretId`, following any number of `PutSecretValue` requests updating the secret with any valid set of `VersionStages` not containing `AWSCURRENT` (`AWSPENDING`, in this context behaves no differently to any other valid non-current `VersionStage` value).

#### Tests

To test the two items above, I introduced the test cases: `test_put_secret_value_version_stages_response `, `test_put_secret_value_version_stages_pending_response `, `test_after_put_secret_value_version_stages_can_get_current `, and `test_after_put_secret_value_version_stages_pending_can_get_current` in `tests/test_secretsmanager/test_secretsmanager.py`.
Please note the coupling of the two fixes applied.
Updating of the server test `test_update_secret_version_stage` was also required.

#### ClientRequestToken
I updates moto's implementation of `CreateSecret` to accept the optional `ClientRequestToken` value, as AWS this is allowed by AWS itself.

#### Tests
To validate `CreateSecret`'s handing of  `ClientRequestToken` I have introduced the test case `test_create_secret_with_client_request_token` in `tests/test_secretsmanager/test_secretsmanager.py`.

#### Force Deletion of Inexistent Secrets
I updated moto's implementation to not raise an `InvalidRequestException` exception when the deletion requests demands to delete an inexistent `SecretId` with `force_delete_without_recovery`. This behaves as if the secret was in fact deleted, matching the AWS's behaviour observed.

#### Parameters Validation Sequence
I updated moto's implementation of `DeleteSecret` to validate the parameters of the request in the same, obsoverd, order of AWS.

#### Tests
To test the two items above, I introduced the test cases: `test_delete_secret_force_no_such_secret`, `test_delete_secret_force_no_such_secret_with_invalid_recovery_window`, and updated `test_delete_secret_that_does_not_exist` to match the parameter validation sequence.

### DIFF
--- a/moto/secretsmanager/models.py
+++ b/moto/secretsmanager/models.py
@@ -129,7 +129,7 @@ class FakeSecret:
             "VersionId": version_id,
         }
         if include_version_stages:
-            dct["VersionStages"] = self.versions[version_id]['version_stages']
+            dct["VersionStages"] = self.versions[version_id]["version_stages"]
         return json.dumps(dct)
 
     def to_dict(self):
@@ -319,7 +319,7 @@ class SecretsManagerBackend(BaseBackend):
         description=None,
         tags=None,
         kms_key_id=None,
-        client_request_token=None
+        client_request_token=None,
     ):
 
         # error if secret exists
@@ -335,7 +335,7 @@ class SecretsManagerBackend(BaseBackend):
             description=description,
             tags=tags,
             kms_key_id=kms_key_id,
-            version_id=client_request_token
+            version_id=client_request_token,
         )
 
         return secret.to_short_dict()
@@ -374,7 +374,7 @@ class SecretsManagerBackend(BaseBackend):
 
             secret.update(description, tags, kms_key_id, last_changed_date=update_time)
 
-            if 'AWSCURRENT' in version_stages:
+            if "AWSCURRENT" in version_stages:
                 secret.reset_default_version(secret_version, version_id)
             else:
                 secret.versions[version_id] = secret_version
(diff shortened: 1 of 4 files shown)


bblommers (approved):
LGTM! Thanks for the PR @MEPalma, and welcome to Moto!

Issue #4863, opened 2022-02-15
Add support for AWS Textract


ecumene:
[Add support for AWS Textract](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/textract.html
)

I only really need...
[get_document_text_detection()](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/textract.html#Textract.Client.get_document_text_detection)
[start_document_text_detection()](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/textract.html#Textract.Client.start_document_text_detection)

I've read the CONTRIBUTE.md and it doesn't seem like it would be too much effort for me to add this. Anyone want to weigh in?

bblommers:
Hi @ecumene, welcome to Moto! Our [scaffolding-script](https://docs.getmoto.org/en/latest/docs/contributing/new_feature.html) should help a lot with the plumbing - it's not guaranteed to get everything right, but at least it will be in the right direction.

If you do get stuck, you can always open a draft PR - I'd be happy to have a look to see what's missing.

PR #4869, opened 2022-02-18
Feature: Textract

ecumene:
#4863


bblommers:
Continuing the conversation here @ecumene, as it's easier to have everything in one place.

URL's are matched here: https://github.com/spulec/moto/blob/876c783a243a34f36bcd067e9f97a7afa5d87297/moto/core/models.py#L341
This method takes all the combinations of urls and paths (as defined in `urls.py`) and creates a mock response for them. So the actual matching is done by the `responses` library.

Running the test with `pytest ..` shouldn't matter, but that did give me an idea:
Do you have an older version of Moto installed in your virtual environment? Maybe pytest is running these tests against an old version, instead of the current source code.

The draft PR is currently failing because `make lint` is failing: https://github.com/spulec/moto/runs/5258367887?check_suite_focus=true
If we fix that first, the CI will run the tests and _hopefully_ confirm that the matching logic is indeed correct as is!

ecumene:
> Do you have an older version of Moto installed in your virtual environment? Maybe pytest is running these tests against an old version, instead of the current source code.

Well done... That was it! Making a new `pyenv` did the trick.

PR is updated and linted 😎

ecumene:
Lmk if you want me to squash my commits, btw

bblommers:
> > Do you have an older version of Moto installed in your virtual environment? Maybe pytest is running these tests against an old version, instead of the current source code.
> 
> Well done... That was it! Making a new `pyenv` did the trick.
> 

Awesome! Glad to hear that's solved.

Looks like black is still failing though - make sure that you're using the right version of black, as different versions will yield different formats: `black==19.10b0`

tests/test_textract/test_server.py
```diff
@@ -0,0 +1,13 @@
...
+import moto.server as server
+
+
+def test_textract_list():
+    backend = server.create_backend_app("textract")
+    test_client = backend.test_client()
+
+    resp = test_client.get("/")
+    resp.status_code.should.equal(200)
+    str(resp.data).should.contain("?")
```
bblommers, line 13: This test is automatically generated - the idea behind it is that we send a raw HTTP request to the server, to verify that Moto understands it and responds accordingly.
The other tests verify that Moto  understands requests created by boto3 - this one really acts as a belts and braces.

A simple example on how to create a HTTP request:
https://github.com/spulec/moto/blob/master/tests/test_cloudtrail/test_server.py

As Textract does not have any straightforward `list_...` methods (yet). So the simplest test case would probably be to verify that `get_document_text_detection` throws an error message if no JobId is send through.
The appropriate header for Textract looks like this: `X-Amz-Target=Textract.GetDocumentTextDetection`

Apologies for the lack of documentation here - that is still a work in progress :slightly_smiling_face:

bblommers (changes requested):

ecumene:
It appears in server mode the `TextractBackend` class doesn't relay the info I set. Is that because it's not setting those values on the server? I wonder if there's a way to do that. Maybe an HTTP API for moto that sets values on the server? That's overkill for this PR though.

Branching the test with an if statement would be pretty bad. Is there a way to disable this test for server mode? Then I could add something to the docs stating that this sort of mocking doesn't work.

bblommers:
> It appears in server mode the `TextractBackend` class doesn't relay the info I set. Is that because it's not setting those values on the server? I wonder if there's a way to do that. Maybe an HTTP API for moto that sets values on the server? That's overkill for this PR though.
> 
Correct -the MotoServer runs inside a Docker container, so there's no way to change those values (yet)

> Branching the test with an if statement would be pretty bad. Is there a way to disable this test for server mode? Then I could add something to the docs stating that this sort of mocking doesn't work.

Branching with an if-statement is pretty common in our tests actually - there are quite a few scenarios where this is necessary, so it wouldn't be that bad.

Out the top of my head, the check would look something like this:
```
from moto import settings
from unittest import SkipTest
if settings.TEST_SERVER_MODE:
    raise SkipTest("..reason")
```

bblommers (approved):
LGTM! Thanks for the PR @ecumene :+1:

PR #4873, opened 2022-02-18
Test responses compatibility

bblommers:
Example PR to show that `responses` works OK-ish with Moto.


bblommers:
This was fixed with #4896

PR #4876, opened 2022-02-20
If value is a string, it's not a list

thedrow:
This is a quick cleanup of the code. The extra condition doesn't make sense to me.


bblommers:
They're two different attributes: `value` (the actual value) vs the `filter_value` (the expected value)

Considering the number of failed tests, it looks like the if-statement is necessary. :slightly_smiling_face:

thedrow:
🤦

bblommers:
> 🤦

Thanks for the PR anyway @thedrow, I appreciate the intent!

PR #4877, opened 2022-02-20
Image names can now be glob patterns

thedrow:
The following terraform doesn't work:
```hcl
data "aws_ami" "aml2" {
  most_recent = true

  filter {
    name   = "name"
    values = ["amzn2-ami-hvm-*-x86_64-ebs"]
  }
  owners = ["amazon"]
}
```
I figured this could do the trick.

### DIFF
--- a/moto/ec2/models.py
+++ b/moto/ec2/models.py
@@ -1783,6 +1783,10 @@ class AmiBackend(object):
         self, ami_ids=(), filters=None, exec_users=None, owners=None, context=None
     ):
         images = self.amis.copy().values()
+        
+        if 'name' in filters and any(True for c in filters['name'] if c in ("*", "?")):
+            pattern = filters['name']
+            filters['name'] = [fnmatch.fnmatch(image, pattern) for image in images]
 
         if len(ami_ids):
             # boto3 seems to default to just searching based on ami ids if that parameter is passed


thedrow:
@bblommers How can I add this TF snippet as a test?

bblommers:
Hi @thedrow, the easiest way would be to convert it to a boto3-test. We do not have a mechanism to test/verify random TF-samples, but plenty of boto3 tests that you can draw inspiration from.

This test comes closest to what you're trying to achive, if I'm reading the TF snippet correctly (?)
https://github.com/spulec/moto/blob/master/tests/test_ec2/test_amis.py#L954

thedrow:
@bblommers I was referring to the terraform tests that are [running in CI](https://github.com/spulec/moto/runs/5263910774?check_suite_focus=true). Are they out of our control?

bblommers:
> @bblommers I was referring to the terraform tests that are [running in CI](https://github.com/spulec/moto/runs/5263910774?check_suite_focus=true). Are they out of our control?

Yes, they are. We run (a small subset of the) tests created by Hashicorp,  as defined here:
https://github.com/hashicorp/terraform-provider-aws/tree/b063e4f577ef4626d66662bd6f752e696460c8ed/aws

These tests are not meant as a replacement for boto3-style unit tests though, and they are not finegrained enough to be a replacement either. They act more as an additional validation step, to ensure that our unit tests are roughly correct.

PR #4879, opened 2022-02-21
RDS modify-db-instance fix KeyError

steffyP:
This PR fixes `KeyError` exception which arises if `modify-db-instance` is called with a DBParameterGroupName that does not exist.
In that case it will raise a `DBParameterGroupNotFoundError`


bblommers (approved):
LGTM! Thanks for the PR @steffyP

PR #4882, opened 2022-02-22
RDS: enable add and remove tag for resources

steffyP:
This PR enables adding and removing tags from EventSubscription and DBParameterGroup.

Added testcase where DBParameterGroup + EventSubscription tags are modified.


bblommers (approved):
LGTM! Thanks for the fix @steffyP

PR #4883, opened 2022-02-23
Fix copying S3 objects, when using MetadataDirective='REPLACE'

Amwam:
This relates to the change in https://github.com/spulec/moto/pull/4838.

When using the MetadataDirective, the copy_object call is still allowed, even if no other changes take place. The current version of moto 3.0.4 doesn't allow this.

Example using boto3:

```python
import boto3
key_name = "my_file.txt"
source_bucket = "bucket.bucket.com"

s3_client = boto3.client('s3')

# Write a file for us to use
s3 = boto3.resource("s3", region_name='eu-west-1')
key = s3.Object(source_bucket, key_name)
key.put(Body=b"some value")

# attempt to copy object to itself
# This would fail with a ClientError (Invalid Request) because the object is already in the bucket and nothing is being changed
s3_client.copy_object(
    Bucket=source_bucket,
    CopySource=f"{source_bucket}/{key_name}",
    Key=key_name,
)


# This is allowed by boto3, but fails when using moto
s3_client.copy_object(
    Bucket=source_bucket,
    CopySource=f"{source_bucket}/{key_name}",
    Key=key_name,
    MetadataDirective="REPLACE",
)
```

I'm not sure if I've done the correct thing here by adding a new property to the copy_object function in the model, so happy to receive some input on that.

Also added a test which fails on the latest `master` branch.

### DIFF
--- a/moto/s3/models.py
+++ b/moto/s3/models.py
@@ -2016,6 +2016,7 @@ class S3Backend(BaseBackend, CloudWatchMetricProvider):
         encryption=None,
         kms_key_id=None,
         bucket_key_enabled=False,
+        mdirective=None,
     ):
         if (
             src_key.name == dest_key_name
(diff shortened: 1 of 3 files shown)


bblommers (approved):
LGTM! Thanks for fixing this @Amwam

PR #4884, opened 2022-02-23
Add implementation for cognitoidp admin_set_user_mfa_preference

linuxlewis:
* Update set_user_mfa_preference to handle PreferredMFA option
* Add tests

### DIFF
--- a/moto/cognitoidp/models.py
+++ b/moto/cognitoidp/models.py
@@ -1637,7 +1637,9 @@ class CognitoIdpBackend(BaseBackend):
                 _, username = user_pool.access_tokens[access_token]
                 user = self.admin_get_user(user_pool.id, username)
 
-                if software_token_mfa_settings["Enabled"]:
+                if software_token_mfa_settings and software_token_mfa_settings.get(
+                    "Enabled"
+                ):
                     if user.token_verified:
                         user.software_token_mfa_enabled = True
                     else:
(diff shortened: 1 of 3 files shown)


moto/cognitoidp/models.py
```diff
@@ -1643,13 +1645,39 @@ def set_user_mfa_preference(
...
                 return None
         else:
             raise NotAuthorizedError(access_token)
 
+    def admin_set_user_mfa_preference(
+        self, user_pool_id, username, software_token_mfa_settings, sms_mfa_settings
+    ):
+        user = self.admin_get_user(user_pool_id, username)
+
+        if software_token_mfa_settings["Enabled"]:
```
bblommers, line 1664: The docs do not say whether this is required, so it's probably safer to assume this field may not be set:
```
if software_token_mfa_settings.get("Enabled", False):
```

moto/cognitoidp/models.py
```diff
@@ -1643,13 +1645,39 @@ def set_user_mfa_preference(
...
+            if user.token_verified:
+                user.software_token_mfa_enabled = True
+            else:
+                raise InvalidParameterException(
+                    "User has not verified software token mfa"
+                )
+
+            if software_token_mfa_settings["PreferredMfa"]:
```
bblommers, line 1672: Same here

moto/cognitoidp/models.py
```diff
@@ -1643,13 +1645,39 @@ def set_user_mfa_preference(
...
+                user.preferred_mfa_setting = "SOFTWARE_TOKEN_MFA"
+        elif sms_mfa_settings["Enabled"]:
```
bblommers, line 1674: Same here

moto/cognitoidp/models.py
```diff
@@ -1643,13 +1645,39 @@ def set_user_mfa_preference(
...
+            user.sms_mfa_enabled = True
+
+            if sms_mfa_settings["PreferredMfa"]:
```
bblommers, line 1677: Same here

bblommers (changes requested):
Hi @linuxlewis, would you mind adding a few more tests?
Ideally there's  a unit test for every possible path, especially a test where `SoftwareTokenMfaSettings` is set and another where we verify that the error is raised correctly.

The docs have a small example on what a negative test should look like:
http://docs.getmoto.org/en/latest/docs/contributing/development_tips/tests.html

linuxlewis:
@bblommers Thanks for the review. Made updates to access parameters safely and added a negative test case. Please let me know if there's anything else to add

bblommers (approved):
LGTM! Thanks for the PR @linuxlewis

PR #4885, opened 2022-02-23
Addition of s3control support in server mode

pinzon:
This PR addresses localstack/localstack#5168. With this patch s3control works fine in server mode.

### DIFF
--- a/moto/server.py
+++ b/moto/server.py
@@ -147,7 +147,7 @@ class DomainDispatcherApplication(object):
             )
         elif service == "s3":
             path = environ.get("PATH_INFO", "")
-            if "v20180820" in path or "s3-control" in environ["HTTP_HOST"]:
+            if path.startswith("/v20180820/") or "s3-control" in environ["HTTP_HOST"]:
                 host = "s3control"
         else:
             host = "{service}.{region}.amazonaws.com".format(
--- a/moto/server.py
+++ b/moto/server.py
@@ -145,6 +145,10 @@ class DomainDispatcherApplication(object):
             host = "ingest.{service}.{region}.amazonaws.com".format(
                 service=service, region=region
             )
+        elif service == "s3":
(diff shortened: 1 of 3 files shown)


moto/server.py
```diff
@@ -145,6 +145,10 @@ def infer_service_region_host(self, body, environ):
             host = "ingest.{service}.{region}.amazonaws.com".format(
                 service=service, region=region
             )
+        elif service == "s3":
+            path = environ.get("PATH_INFO", "")
+            if "v20180820" in path or "s3-control" in environ["HTTP_HOST"]:
```
bblommers, line 150: This is something I was actually looking into  myself a couple of days yesterday, and the biggest worry I had:

What will happen if a user has a bucket or key that contains the phrase `v20180820`? That would redirect the user to S3Control as well. Ideally we tighten this down as much as possible, i.e. only redirect to S3Control if we're *really* sure that's what we want.

Do you know how many possible paths there are? I wonder if it's feasible to simply list all possible S3Control paths:
```
if path in ["/20180820/accesspoint", "/20180820/"publicaccessblock", "/20180820/other-s3control-method", ..]
```

pinzon: It could be a very long list. What get I from a quick read of this [doc](https://docs.aws.amazon.com/AmazonS3/latest/API/API_Operations_AWS_S3_Control.html), every action of the s3control service starts with "v20180820" at the beginning of the path. And also the service s3 allows for a bucket called "v20180820"

bblommers: Hmm. It's a long list indeed, and a maintenance nightmare. 
If it always starts the same way, let's just use that as a check instead - `if path.startswith("..")`. At least that reduces the chance of unintended redirects.

bblommers (approved):
Awesome - thanks for the PR @pinzon!

PR #4889, opened 2022-02-24
Added support for custom AWS Lambda images.

ngnpope:
~~Also changed the default to `mlupin/docker-lambda`.~~

Currently the image used for executing Lambda functions is hardcoded:

https://github.com/spulec/moto/blob/e84cc20abe5b92e3bb77bcddd9240d4233a9524e/moto/awslambda/models.py#L613

Unfortunately it seems that development has also stalled, the last release at the time of writing being over one year ago:

- https://github.com/lambci/docker-lambda
- https://hub.docker.com/r/lambci/lambda/tags

Because of this, there is no image built for Python 3.9 which means it is not possible to use this with moto.

~~Instead, we should~~ With this change it is possible to switch to using a supported fork:

- https://github.com/mLupine/docker-lambda
- https://hub.docker.com/r/mlupin/docker-lambda/tags

~~In addition, it is now possible to set~~ This is achieved by setting the `MOTO_DOCKER_LAMBDA_IMAGE` environment variable to override this as required.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,11 +1,12 @@
 Moto Changelog
-===================
+==============
 
 3.0.5
 -----
 Docker Digest for 3.0.5: <autopopulateddigest>
 
-    TBD
+    Miscellaneous:
+        * AWSLambda: Made the docker image repository selectable via the `MOTO_DOCKER_LAMBDA_IMAGE` environment variable.
 
 3.0.4
 -----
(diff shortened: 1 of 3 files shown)


bblommers (changes requested):
Hi @ngnpope, thanks for the PR!
It makes a lot of sense to make this configurable, but I'm not really comfortable with making the `mlupin`-repo the default. AWSLambda still supports Python 3.6 and 3.7, and it doesn't look like that is supported by this repo, making this a breaking change.

I'd suggest we keep `lambci/lambda` as the default for now - with the next major release we can look into switching to `mlupin`.

ngnpope:
Hi @bblommers. Thanks for pointing out about the older Python versions - I'd overlooked that. I've tweaked this to keep the original default as requested.

bblommers (approved):
LGTM!

PR #4891, opened 2022-02-24
Addition of Entities Ids when listing

pinzon:
This PR addresses localstack/localstack#5296. where the user describes that the IAM service is missing the Resource Id (GroupId, UserId and RoleId) when calling **ListEntitiesForPolicy**. I compared the Moto response to AWS and it's true.

Changes:
- Addition of GroupId's, RoleId's and UserId's to the **ListEntitiesForPolicy** response.
- Extension of Iam tests to validate the existanse of the attribute in the response.


bblommers (approved):
LGTM - thanks for the PR @pinzon!

PR #4898, opened 2022-03-01
This adds a new service "databrew", mostly around managing databrew

shield007:
This adds a new service "databrew" and supports some methods for managing databrew recipes.

New Services:
  * databrew:
      * create_recipe()
      * list_recipes()
      * get_recipe()


moto/databrew/__init__.py
```diff
@@ -0,0 +1,4 @@
+from .models import databrew_backend
+
+databrew_backends = {"global": databrew_backend}
+mock_databrew = databrew_backend.decorator
```
bblommers, line 4: Using the `global`-region in the backends is only used in global services such as S3, and as far as I can tell `databrew` is regional. 
See how the CloudTrail-service is setup as an example: https://github.com/spulec/moto/blob/master/moto/cloudtrail/__init__.py

moto/databrew/models.py
```diff
@@ -0,0 +1,62 @@
...
+        return {
+            "Name": self.name,
+            "Steps": self.steps,
+            "Description": self.description,
+            "CreateTime": self.created_time.isoformat(),
+            "Tags": self.tags or dict(),
+        }
+
+
+databrew_backend = DataBrewBackend()
```
bblommers, line 62: You can use the BackendDict to automatically create a separate backend for all regions - see Cloudtrail again as an example:
https://github.com/spulec/moto/blob/206590acf5215abef67684a0122dc668750f71fa/moto/cloudtrail/models.py#L410

moto/databrew/models.py
```diff
@@ -0,0 +1,62 @@
...
+        },
+    }
+
+    def __init__(self):
+        self.recipes = OrderedDict()
+
+    def create_recipe(self, recipe_name, recipe_description, recipe_steps, tags):
+        # https://docs.aws.amazon.com/databrew/latest/dg/API_CreateRecipe.html
+        if recipe_name in self.recipes:
+            raise RecipeAlreadyExistsException()
```
bblommers, line 28: We should have a separate test for this scenario

bblommers (changes requested):
Hi @shield007, thanks for the PR! Please see my comments.

The ServerMode-tests are still failing - that my be fixed by running  `scripts/update_backend_index.py`.
(All AWS URL’s that need intercepting are indexed - this scripts ensures the index is up-to-date, and should be run whenever a new URL/`urls.py` is added.

moto/databrew/__init__.py
shield007: Ah great. will push a fix for that shortly.

moto/databrew/models.py
shield007: yeah, that makes sense. will add one.

shield007:
Hi, Just pushed changes that address most the issue you mentioned. @bblommers . Just waiting to see if the tests pass.

shield007:
I see the unit tests for python  > 3.6 are still failing. Is this a known problem? or something caused by my changes? Stange that 3.6 python passed and the others did not. Server mode looks to have passed now.

bblommers:
@shield007 The failures are due to a new `responses`-release, so unrelated to this change. When that is fixed, I'll merge this PR. :+1:

bblommers (approved):

Issue #4899, opened 2022-03-01
secretsmanager tag_resource creates duplicate tag key's if the key already exists


stromp:
secretsmanager tag_resource method add a tag instead of overwrites a tags if the key name is the same. Resulting in keys with duplicated keys. 

Normal AWS behavior:

```
❯ aws secretsmanager create-secret --name 'test-sjoerd'
{
    "ARN": "arn:aws:secretsmanager:eu-west-1:319518030215:secret:test-sjoerd-JOj6kh",
    "Name": "test-sjoerd"
}
❯ aws secretsmanager tag-resource --secret-id 'test-sjoerd' --tags "[{'Key': 'a', 'Value': 'b'}]"

❯ aws secretsmanager describe-secret --secret-id 'test-sjoerd'
{
    "ARN": "arn:aws:secretsmanager:eu-west-1:319518030215:secret:test-sjoerd-JOj6kh",
    "Name": "test-sjoerd",
    "LastChangedDate": "2022-03-01T16:39:02.553000+01:00",
    "Tags": [
        {
            "Key": "a",
            "Value": "b"
        }
    ],
    "CreatedDate": "2022-03-01T16:37:05.377000+01:00"
}
❯ aws secretsmanager tag-resource --secret-id 'test-sjoerd' --tags '[{"Key": "a", "Value": "c"}]'
❯ aws secretsmanager describe-secret --secret-id 'test-sjoerd'
{
    "ARN": "arn:aws:secretsmanager:eu-west-1:319518030215:secret:test-sjoerd-JOj6kh",
    "Name": "test-sjoerd",
    "LastChangedDate": "2022-03-01T16:39:29.306000+01:00",
    "Tags": [
        {
            "Key": "a",
            "Value": "c"
        }
    ],
    "CreatedDate": "2022-03-01T16:37:05.377000+01:00"
}
```

however in moto you get:

```
from moto import mock_secretsmanager
import boto3


def tags():
    mock_secretsmanager().start()
    boto_client = boto3.client('secretsmanager')
    boto_client.create_secret(Name='test',
                              Tags=[{'Key': 'a', 'Value': 'b'}])
    boto_client.tag_resource(SecretId='test',
                             Tags=[{'Key': 'a', 'Value': 'c'}])
    print(boto_client.describe_secret(SecretId='test')['Tags']
```
resulting in:

```
tags()
[{'Key': 'a', 'Value': 'b'}, {'Key': 'a', 'Value': 'c'}]
```

PR #4900, opened 2022-03-01
fix for duplicate tag keys in secretsmanager

stromp:
This should fix issue #4899


bblommers (approved):
LGTM! Thanks for the fix @stromp

Issue #4901, opened 2022-03-02
[sns] sns_backend should default to env AWS_DEFAULT_REGION / DEFAULT_REGION_NAME


Lilja:
I'm checking the documentation to see if it's possible to get the notifications of published SNS messages.

There is a like here that says you can access it through:
`from moto.sns import sns_backend`. After 30 minutes of debugging why `sns_backend.topics` was empty when I just created the topic using boto3, I found the following line.

https://github.com/spulec/moto/blob/7e88a901d9ed7612d92bdeb9dd67056c99bf3126/moto/sns/__init__.py#L4

We have this line in our tests: 
```python
os.environ["AWS_DEFAULT_REGION"] = moto.s3.responses.DEFAULT_REGION_NAME = AWS_REGION = "eu-west-1"
```
Shouldn't this try to get the backend based on the default region?

Lilja:
My workaround right now:
```
from moto.sns.models import sns_backends
sns_backend = sns_backends["eu-west-1"]
```

bblommers:
Hi @Lilja, thanks for raising this! In this case it is the documentation that is wrong, actually - your workaround is the recommended fix.
```
from moto.sns.models import sns_backends
sns_backend = sns_backends["eu-west-1"]
```

Apologies for the confusion - I'll update the documentation to make this more clear.

PR #4902, opened 2022-03-02
[sns] get default region by env var

Lilja:
https://github.com/spulec/moto/issues/4901


bblommers:
Thanks for creating a PR for this! As I said in the issue itself, the `sns_backends` variable would be the correct way to access this information.
In addition, the `AWS_DEFAULT_REGION` is not guaranteed to be set, so this is not the right solution for everyone.

I'll raise a PR shortly to fix the docs, and I'll remove the `sns_backend`-variable as well to remove any confusion.

PR #4905, opened 2022-03-02
Admin - change XML2Dict namespace separator

bblommers:
The default separator, ':',  broke the underlying parser after a security update.

Source: https://bugs.launchpad.net/ubuntu/+source/python-xmltodict/+bug/1961800
Proposed fix to xmltodict: https://github.com/martinblech/xmltodict/pull/290

As far as I can tell, this is the only place where it affects Moto, as no other services use XML namespaces.
This PR should remove the need to wait for any upstream/downstream fixes that may or may not happen.

@thrau FYI, as it looks like were bitten by this first here: https://github.com/localstack/localstack/pull/5585

### DIFF
--- a/moto/sts/models.py
+++ b/moto/sts/models.py
@@ -102,22 +102,23 @@ class STSBackend(BaseBackend):
             force_cdata=True,
             process_namespaces=True,
             namespaces=namespaces,
+            namespace_separator="|",
         )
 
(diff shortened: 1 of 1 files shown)


thrau:
Yes, we already went through the obligatory rants about XML!

PR #4918, opened 2022-03-08
Fix deleting secrets by ARN

### LINKED ISSUE #4917: Deleting secrets not possible with full ARN
Deleting secrets in AWS Secrets Manager isn't possible when using the full ARN as secrets id and not specifying `ForceDeleteWithoutRecovery=True`:

```python
import boto3
from moto import mock_secretsmanager

@mock_secretsmanager
def test_delete():
    client = boto3.client("secretsmanager")
    secret = client.create_secret(Name="test")

    client.delete_secret(SecretId=secret["ARN"])  # doesn't work
    client.delete_secret(SecretId=secret["ARN"], ForceDeleteWithoutRecovery=True)  # works
    client.delete_secret(SecretId=secret["Name"])  # works
```

The reason for that is that during the deletion `SecretsStore.get()` is getting called:
https://github.com/spulec/moto/blob/1d7440914e4387661b0f200d16cd85298fd116e9/moto/secretsmanager/models.py#L716

`SecretsStore.get()` however isn't overwritten to resolve ARNs to secret names, therefore accessing the item fails and results in a `ResourceNotFoundException` even though the secret does exist.

Dunedan:
This fixes an issue with deleting secrets in AWS Secrets Manager not
being possible when using the full ARN as secret id, because SecretsStore
didn't do resolve from ARNs to secret names when using get().

Fixes #4917

### DIFF
--- a/moto/secretsmanager/models.py
+++ b/moto/secretsmanager/models.py
@@ -180,6 +180,10 @@ class SecretsStore(dict):
         new_key = get_secret_name_from_arn(key)
         return dict.__contains__(self, new_key)
 
+    def get(self, key, *args, **kwargs):
+        new_key = get_secret_name_from_arn(key)
+        return super().get(new_key, *args, **kwargs)
+
     def pop(self, key, *args, **kwargs):
         new_key = get_secret_name_from_arn(key)
         return super().pop(new_key, *args, **kwargs)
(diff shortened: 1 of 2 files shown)


bblommers (approved):
LGTM - Thanks for the fix @Dunedan!

PR #4930, opened 2022-03-11
Fix RDS incorrect statuses

kazanzhy:
Fixed RDS incorrect statuses that were added by me in previous PRs


bblommers (approved):
Thanks for the fix @kazanzhy

PR #4932, opened 2022-03-11
fix route53 private zone value

macnev2013:
## Fixed
- `PrivateZone` value type in response
- default behaviour for `PrivateZone` when `vpc` is present


bblommers:
Hi @macnev2013! I've just enabled a bunch of stricter linting rules, so you'll have to merge/rebase with master if you want to verify this change passes the new linter.
Note that we switched to a new version of Black as well, so formatting with an older version may produce invalid results. We now  use `black==22.1.0`.

macnev2013:
Hey @bblommers , Thanks for pointing out. Updating PR shortly.

PR #4935, opened 2022-03-11
fixed route53 private zone value

macnev2013:
## Fixed
- `PrivateZone` value type in response
- default behaviour for `PrivateZone` when `vpc` is present


bblommers (approved):
LGTM! Thanks @macnev2013

PR #4936, opened 2022-03-12
Simplify error handling

bblommers:
Various services had different ways of catching and dealing with errors. This PR streamlines the exception handling across all services - every error that extends either RESTError or AWSError is now caught by Moto, and the appropriate response format is determined uniformly.

### DIFF
--- a/moto/acm/responses.py
+++ b/moto/acm/responses.py
@@ -2,7 +2,7 @@ import json
 import base64
 
 from moto.core.responses import BaseResponse
-from .models import acm_backends, AWSError, AWSValidationException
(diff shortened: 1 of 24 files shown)


bpandola:
Always great to see a PR that deletes more code than it adds.  Nice work.  I think there are plenty more opportunities like this in the code base--I hope to get some time to work on some myself this year.

bblommers:
> Always great to see a PR that deletes more code than it adds.

Agreed - it's very satisfying. The first time I came across the error_handler decorator I thought it was quite an elegant solution, but after copy/pasting it 5 times across to other services... not so much :slightly_smiling_face:

PR #4948, opened 2022-03-18
Feature/add glue get job

malekhnovich:
* Adding boto3 get_job functionality (first time contributor so please go easy on me)

### DIFF
--- a/moto/glue/models.py
+++ b/moto/glue/models.py
@@ -15,6 +15,7 @@ from .exceptions import (
     PartitionAlreadyExistsException,
     PartitionNotFoundException,
     VersionNotFoundException,
+    JobNotFoundException,
 )
 from ..utilities.paginator import paginate
 
@@ -197,6 +198,12 @@ class GlueBackend(BaseBackend):
             worker_type,
         )
         return name
+
+    def get_job(self, name):
+        try:
+            return self.jobs[name]
+        except KeyError:
+            raise JobNotFoundException(name)
 
     @paginate(pagination_model=PAGINATION_MODEL)
     def list_jobs(self):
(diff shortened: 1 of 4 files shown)


moto/glue/models.py
```diff
@@ -470,5 +477,29 @@ def __init__(
...
+            "Connections": self.connections,
+            "MaxRetries": self.max_retries,
+            "AllocatedCapacity": self.allocated_capacity,
+            "Timeout": self.timeout,
+            "MaxCapacity": self.max_capacity,
+            "WorkerType": self.worker_type,
+            "NumberOfWorkers": self.number_of_workers,
+            "SecurityConfiguration": self.security_configuration,
+            "NotificationProperty": self.notification_property,
+            "GlueVesion": self.glue_version,
```
bblommers, line 501: Typo - should probably be `GlueVersion`

tests/test_glue/test_glue.py
```diff
@@ -41,6 +42,26 @@ def test_list_jobs():
...
+    exc.value.response["Error"]["Code"].should.equal("EntityNotFoundException")
+    exc.value.response["Error"]["Message"].should.match("Job my_job_name not found")
+
+
+@mock_glue
+def test_get_job_exists():
+    client = create_glue_client()
+    job_name = create_test_job(client)
+    response = client.get_job(JobName=job_name)
+    assert response["Job"]["Name"] == job_name
```
bblommers, line 62: Can we get a few more assertions here, just to avoid the chance of any other typo's?

bblommers (changes requested):
Hi @malekhnovich, thanks for the PR! Looks good overall, just got two minor comments

moto/glue/models.py
malekhnovich: 👍

tests/test_glue/test_glue.py
malekhnovich: yes on the way

bblommers (approved):
Awesome - thanks @malekhnovich!

PR #4954, opened 2022-03-20
Fix test failures on systems with 32-bit time_t

mgorny:
Skip tests if OverflowError is raised when boto3 is processing
timestamps.  This is a known limitation of boto3 on 32-bit platforms
(https://github.com/boto/botocore/issues/2355).

Catching OverflowError is the best option here since some 32-bit
platforms (e.g. NetBSD) use 64-bit time_t, and others are working
on providing a switch to the 64-bit type (e.g. glibc).

### DIFF
--- a/tests/test_acm/test_acm.py
+++ b/tests/test_acm/test_acm.py
@@ -160,7 +160,10 @@ def test_describe_certificate():
     client = boto3.client("acm", region_name="eu-central-1")
     arn = _import_cert(client)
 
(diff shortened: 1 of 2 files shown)


bblommers (approved):
This is not something I can test, but the solution LGTM.
Thanks for the PR @mgorny!

mgorny:
> This is not something I can test, but the solution LGTM.

Yeah, I had to set up an i386 container to reproduce it ;-).

> Thanks for the PR @mgorny!

Thanks!

Issue #4071, opened 2021-07-14
AWS Rekognition support


Robmosh91:
HI,

first, thanks for this awesome project which helped me a lot.
Currently I'm looking for a way to mock the AWS Rekognition Service and would like to contribute to its implementation.

Can you give any guidance on how to start? 
If anybody else would like to contribute, I'm happy to co-work on this.

bblommers:
Hi @Robmosh91, that sounds good - new services are always welcome!
There's some documentation how to add new services here: 
https://github.com/spulec/moto/blob/master/CONTRIBUTING.md#missing-services

Hopefully that's useful. Feel free to ask any questions if you get stuck, or just open a draft PR if there's something specific that doesn't work.

bblommers:
With Moto 3.1.4.dev3 we now have an initial Rekognition implementation, with support for the features `start_text_detection` and `get_text_detection`.

I'll close this, but feel free to open a new issue or PR if you'd like to see any other features!

Issue #4955, opened 2022-03-20
Unable to contribute / ERROR: Permission to spulec/moto.git denied


AlanDonohoe:
Hello 👋

I am working on adding support for [Rekognition](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/rekognition.html).

But when I try and push my branch to moto's repo, I get this error:
```
ERROR: Permission to spulec/moto.git denied to MY HANDLE...
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
```

Settings seem to be OK on my side:

```
git remote -v
origin  git@github.com:spulec/moto.git (fetch)
origin  git@github.com:spulec/moto.git (push)
```

Am I missing something?

Any ideas?

Thank you.

bblommers:
Hi @AlanDonohoe! This can be solved  by creating a copy of Moto's repo, using the `Fork`-button in the topright corner of https://github.com/spulec/moto.

After checking our your own copy/fork, you can create feature-branches there - and then create a PR from your branch into `spulec/moto/master`.

Let me know if you run into any issues with that.

AlanDonohoe:
Hi @bblommers Fantastic, thank you!

PR #4957, opened 2022-03-21
Add Rekognition support.

AlanDonohoe:
##### Description
Adds Rekognition support.

##### Background Context
At the moment just adding support for start_text_detection and get_text_detection.

refs:
https://docs.aws.amazon.com/rekognition/latest/dg/what-is.html
https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/rekognition.html
https://github.com/spulec/moto/issues/4071


AlanDonohoe:
Hi @bblommers following on from #4955 - I have had a crack at adding Rekognition support to moto.

I am sure that I have got a few bits wrong/ missed something, as this is the first time working in this codebase.

If anyone can review the code as is and I can then amend.

Thank you.

moto/rekognition/responses.py
```diff
@@ -0,0 +1,63 @@
...
+
+    def start_text_detection(self):
+        params = self._get_params()
+        video = params.get("Video")
+        client_request_token = params.get("ClientRequestToken")
+        notification_channel = params.get("NotificationChannel")
+        job_tag = params.get("JobTag")
+        filters = params.get("Filters")
+        job_id = self.rekognition_backend.start_text_detection(
+            Video=video,
```
bblommers, line 24: These are probably not necessary, considering they are unused in the backend.
If do you want to keep them, they should be in `snake_case`-format

moto/rekognition/responses.py
```diff
@@ -0,0 +1,63 @@
...
+        next_token = params.get("NextToken")
+        (
+            job_status,
+            status_message,
+            video_metadata,
+            text_detections,
+            next_token,
+            text_model_version,
+        ) = self.rekognition_backend.get_text_detection(
+            JobId=job_id,
```
bblommers, line 46: Same here

tests/test_rekognition/test_rekognition.py
```diff
@@ -0,0 +1,37 @@
...
+    resp["ResponseMetadata"]["HTTPStatusCode"].should.equal(200)
+
+
+@mock_rekognition
+def test_get_text_detection():
+    client = boto3.client("rekognition", region_name="us-east-2")
+    job_id = "".join(
+        random.choice(string.ascii_uppercase + string.digits) for _ in range(64)
+    )
+    resp = client.get_text_detection(JobId=job_id)
```
bblommers, line 35: This shoul dhave some assertions as well, to verify that the data is returned in the correct format

tests/test_rekognition/test_server.py
```diff
@@ -0,0 +1,13 @@
+"""Test different server responses."""
+import sure  # noqa # pylint: disable=unused-import
+
+import moto.server as server
+
+
+def test_rekognition_list():
```
bblommers, line 7: Let's remove this test (and the entire file), it doesn't really add anything

bblommers (changes requested):
Hi @AlanDonohoe, thanks for the PR! Please see my comments.

If you want to run the linter locally, to verify it passes, you can run `make lint`.

moto/rekognition/responses.py
bblommers: Let me give a bit of background in how Moto works:

1. The user calls the boto3 API
2. Boto3 performs some initial validation, such as validating all mandatory parameters are included
3. Boto3 turns the API request into an HTTP request, to send this off to the AWS HTTP API
4. Moto only then steps in, and intercepts this request
5. Moto returns a HTTP response, in the same format as the AWS HTTP API would
6. Boto then transforms the HTTP response into the expected boto3 API response

This means two things:
 - There is no guarantee that the parameters Moto receives, the parameters from the HTTP request, are the same name/format as what was send from the boto3 API.
 - Users will rarely interact with the Moto API,  only with the boto3 API.

In other words, we have free range as to what Moto does internally - as long as we use the correct naming/format for the HTTP request parameters (such as params.get("`JobTag`")) and the response. 

The reason that we happen to use `snake_case` is simply consistency, making it easier for developers and maintainers to read the code.

bblommers: BTW, the `params.get("..."`) is deceivingly simple. As the HTTP request parsing logic is the same across all services, we just expose a few simple functions to retrieve the user-supplied parameters. But there is a lot happening behind the scenes there, to figure out whether that parameter comes from the request body, the querystring, as a path-variable, etc.

tests/test_rekognition/test_rekognition.py
```diff
@@ -0,0 +1,37 @@
...
+def test_start_text_detection():
+    client = boto3.client("rekognition", region_name="ap-southeast-1")
+    video = {
+        "S3Object": {
+            "Bucket": "bucket",
+            "Name": "key",
+        }
+    }
+
+    resp = client.start_text_detection(Video=video)
```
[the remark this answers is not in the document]
AlanDonohoe: I added a test for this ([below](https://github.com/spulec/moto/pull/4957/files#diff-28b0c6cfb603fc90786f3370f2d3ea6fd1be61a9650b4e0bdc864fa6f0663a8aR28)) but for some reason, while the call to get_text_detection does return the expected data, this call to start_text_detection, which should return a JobId, but it just returns:

```
{'ResponseMetadata': {'HTTPStatusCode': 200, 'HTTPHeaders': {'server': 'amazon.com'}, 'RetryAttempts': 0}}
```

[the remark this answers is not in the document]
AlanDonohoe: Any ideas what that might be about?

I have been playing about with it but a bit stumped...

moto/rekognition/responses.py
AlanDonohoe: Thank you so much for your detailed response @bblommers Have amended that now.

tests/test_rekognition/test_rekognition.py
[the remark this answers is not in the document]
bblommers: What does the response-body of the AWS HTTP API return for this call?
If the format that we return `json.dumps(dict(JobId=job_id))` is different from what boto3 expects, it's not able to parse the response, and that's usually what happens when boto3's response doesn't include any data.

The quickest way to figure out the response-body from AWS itself is usually to make a real call to AWS, and use a proxy to intercept the request:
http://docs.getmoto.org/en/latest/docs/contributing/development_tips/urls.html#intercepting-aws-requests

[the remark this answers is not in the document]
AlanDonohoe: I tried the MITMProxy but kept getting issues connecting.. 

So I went back down to the AWS CLI, to make a request with debug info, like so:

```
aws rekognition start-text-detection --video "{\"S3Object\":{\"Bucket\":\"foo\",\"Name\":\"bah.mp4\"}}" --debug
```

And among all the logging this was present:
```
botocore.parsers - DEBUG - Response headers: {'x-amzn-RequestId': '5015215e-e8f6-4c06-bb9a-f7d3c0194498', 'Content-Type': 'application/x-amz-json-1.1', 'Content-Length': '76', 'Date': 'Sun, 27 Mar 2022 14:50:44 GMT'}
botocore.parsers - DEBUG - Response body:
b'{"JobId":"eb81c0201c8541e9835539449925fdfcc3884b216cdca7bef576b1ce85ce32b5"}'
```
So AWS API returns bytes, which is what boto3 is expecting.

I have amended now and this is test covered. 👍

AlanDonohoe:
OK, @bblommers I have made a number of amendments based on your feedback and the code is ready for a re-review please.

Thank you.

bblommers:
LGTM so far @AlanDonohoe - happy to merge once the linter is happy (and assuming the build passes, obviously)

AlanDonohoe:
> LGTM so far @AlanDonohoe - happy to merge once the linter is happy (and assuming the build passes, obviously)

OK cool...

There was one linter issue, which I have since resolved.

Have pushed that up now, so once all the linter checks and tests are good, you'll be happy to merge in!

And I can use moto's new Rekognition support back in my original project!

I'm going to resolve all the conversations up to this point so it's clear what amendments have been subsequently fixed up.

bblommers:
@AlanDonohoe There are still some test failures, but they should be resolved after you merge this PR that I raised: https://github.com/LiniusAustPty/moto/pull/1

AlanDonohoe:
> @AlanDonohoe There are still some test failures, but they should be resolved after you merge this PR that I raised: [LiniusAustPty#1](https://github.com/LiniusAustPty/moto/pull/1)

@bblommers Thanks for that... I have merged in.

Do you want me to do some rebasing now on this branch (or are the commits OK as is?)

bblommers:
> Do you want me to do some rebasing now on this branch (or are the commits OK as is?)

No, that's not necessary, I'll just squash and merge.

Thank you for all your work on this @AlanDonohoe!

bblommers (approved):

AlanDonohoe:
> > Do you want me to do some rebasing now on this branch (or are the commits OK as is?)
> 
> No, that's not necessary, I'll just squash and merge.
> 
> Thank you for all your work on this @AlanDonohoe!

Amazing. V happy this will be in the next release.

Thanks for all the feedback and assistance @bblommers

PR #4962, opened 2022-03-22
bugfix: apply cloudwatch subscription filters to newly created streams

SwatsonCodes:
This PR fixes a bug where existing CloudWatch subscription filters were not applied to newly created streams. This should also fix a downstream bug in LocalStack: https://github.com/localstack/localstack/issues/5715


bblommers:
Hi @SwatsonCodes, make sure you use the correct version of black - we recently upgraded to `black==22.1.0`, so formatting using an older versions may fail on our version.

bblommers (approved):
LGTM! Thanks for the PR @SwatsonCodes

PR #4978, opened 2022-03-27
Create & Describe file system APIs return Name property based on tags

asherf:
see: https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/efs.html#EFS.Client.create_file_system
and https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/efs.html#EFS.Client.describe_file_systems


bblommers (approved):
LGTM! Thanks for the PR @asherf

PR #4980, opened 2022-03-28
Remove apostrophe

symroe:
This is an example of a "greengrocers' apostrophe" https://en.wikipedia.org/wiki/Apostrophe#Greengrocers'_apostrophes and is generally considered incorrect


bblommers (approved):
Thanks for the fix @symroe 👍

PR #4983, opened 2022-03-29
Extract region from auth headers for s3 bucket creation

pinzon:
Found out an issue while triaging localstack/localstack#5748, that while creating a bucket, the region from the aws client is not respected. From the comments, I get that in server mode the url is only `localhost:5000` so the url doesn't contain the region. But the region can be also extracted from the request authorization header.

During my testing with I found out that region extracted from the url takes precedence from the header region. So the my patch considers that order before setting the default region.

### DIFF
--- a/moto/core/utils.py
+++ b/moto/core/utils.py
@@ -400,6 +400,14 @@ def aws_api_matches(pattern, string):
         return False
 
 
+def extract_region_from_aws_authorization(string):
+    auth = string or ""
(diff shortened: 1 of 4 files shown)


bblommers (changes requested):
Hey @pinzon, the approach looks good to me - we'd just need to fix the linter and tests.
Note that we had some dependency issues with `black` yesterday - if the linter still fails because of that you'll have to merge in the latest master to fix that

bblommers (approved):
LGTM!

PR #4991, opened 2022-03-30
IOTData - use available services from 'IOT'

bblommers:
Botocore 1.24.29 still released a list of regions in `get_available_regions`.

Botocore 1.24.30 now returns an empty list for `get_available_regions("iot-data").

`IOT` should have the same regions, so we can use that instead.

[Related issue: https://github.com/boto/botocore/issues/2631](https://github.com/aws/aws-sdk/issues/206)


bblommers:
Closing in favor of #4992

PR #4992, opened 2022-03-30
IOTData - intercept new URLs in botocore 1.24.30

bblommers:
Botocore 1.24.29 still released a list of regions in get_available_regions.

Botocore 1.24.30 now returns an empty list for `get_available_regions("iot-data").

IOT should have the same regions, so we can use that instead.

AWS also changed the IOTdata endpoint, so we now have to intercept both URL's.

https://github.com/aws/aws-sdk/issues/206


bblommers:
This is fixed as of moto >= 3.1.4.dev13

PR #4995, opened 2022-03-31
fix cloudformation arn account id

pirakleous:
Currently the account id on the CloudFormation  stack arn and cloudformation change arn is hardcoded to be 123456789, that is not a valid account id since the length is not 12 digits and also it does not use the same account id as the rest of the resources.

The ACCOUNT_ID was introduced on [this](https://github.com/spulec/moto/pull/2641) PR

### DIFF
--- a/moto/cloudformation/utils.py
+++ b/moto/cloudformation/utils.py
@@ -7,7 +7,7 @@ import string
 from moto.core import ACCOUNT_ID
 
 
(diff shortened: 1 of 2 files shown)


pirakleous:
@bblommers Some tests are failing but it looks like they are not related to the changes i made

bblommers (approved):
LGTM! Thanks for the fix @pirakleous

PR #4996, opened 2022-04-01
This adds some new new methods to the databrew service for recipes and rulesets.

shield007:
This adds some new new methods to the databrew service for recipes and rulesets.

New methods:

- databrew:
  -  update_recipe()
  -  create_ruleset()
  -  delete_ruleset()
  -  update_ruleset()
  -   list_rulesets()

### DIFF
--- a/moto/databrew/models.py
+++ b/moto/databrew/models.py
@@ -5,6 +5,7 @@ from moto.core import BaseBackend, BaseModel
 from moto.core.utils import BackendDict
 from moto.utilities.paginator import paginate
 from .exceptions import RecipeAlreadyExistsException, RecipeNotFoundException
+from .exceptions import RulesetAlreadyExistsException, RulesetNotFoundException
 
 
 class DataBrewBackend(BaseBackend):
@@ -15,11 +16,18 @@ class DataBrewBackend(BaseBackend):
             "limit_default": 100,
             "unique_attribute": "name",
         },
+        "list_rulesets": {
+            "input_token": "next_token",
+            "limit_key": "max_results",
+            "limit_default": 100,
+            "unique_attribute": "name",
+        },
     }
 
     def __init__(self, region_name):
         self.region_name = region_name
         self.recipes = OrderedDict()
+        self.rulesets = OrderedDict()
 
     def reset(self):
         """Re-initialize all attributes for this instance."""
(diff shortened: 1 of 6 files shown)


moto/databrew/models.py
```diff
@@ -49,6 +71,52 @@ def get_recipe(self, recipe_name):
...
+        return ruleset
+
+    def get_ruleset(self, ruleset_name):
+        if ruleset_name not in self.rulesets:
+            raise RulesetNotFoundException(ruleset_name)
+        return self.rulesets[ruleset_name]
+
+    @paginate(pagination_model=PAGINATION_MODEL)
+    def list_rulesets(self):
+        return [self.rulesets[key] for key in self.rulesets] if self.rulesets else []
```
bblommers, line 112:
```suggestion
        return list(self.rulesets.values())
```
Slightly easier to read, IMO, and it should do the same thing.

moto/databrew/models.py
```diff
@@ -37,6 +45,20 @@ def create_recipe(self, recipe_name, recipe_description, recipe_steps, tags):
         self.recipes[recipe_name] = recipe
         return recipe
 
+    def update_recipe(self, recipe_name, recipe_description, recipe_steps, tags):
+        if recipe_name not in self.recipes:
+            raise RecipeNotFoundException(recipe_name)
+
+        recipe = self.recipes[recipe_name]
+        if recipe_description:
```
bblommers, line 53: What happens if we want to empty the description? I.e., if a user calls:
```
client.update_recipe(.., Description="", ..)
```
For most services this is valid, and means that the description should now be set to an empty string.
So it may be better to explicitly check for a None-value:
```
if recipe_description is not None:
    ...
```

moto/databrew/models.py
```diff
@@ -49,6 +71,52 @@ def get_recipe(self, recipe_name):
...
+        )
+        self.rulesets[ruleset_name] = ruleset
+        return ruleset
+
+    def update_ruleset(self, ruleset_name, ruleset_description, ruleset_rules, tags):
+        if ruleset_name not in self.rulesets:
+            raise RulesetNotFoundException(ruleset_name)
+
+        ruleset = self.rulesets[ruleset_name]
+        if ruleset_description:
```
bblommers, line 96: Same here re: the null-check

moto/databrew/responses.py
```diff
@@ -50,12 +50,98 @@ def list_recipes(self):
...
+    def delete_ruleset_response(self, ruleset_name):
+        self.databrew_backend.delete_ruleset(ruleset_name)
+        return 204, {}, ""
+
+    @amzn_request_id
+    def ruleset_response(self, request, full_url, headers):
+        self.setup_class(request, full_url, headers)
+        parsed_url = urlparse(full_url)
+
+        ruleset_name = parsed_url.path.rstrip("/").rsplit("/", 1)[1]
```
bblommers, line 120:
```suggestion
        ruleset_name = parsed_url.path.split("/")[-1]
```
Assuming that the ruleset_name is always last in the URL, this would be easier to read

bblommers (changes requested):
Hi @shield007, thanks for the PR! Please see my comments

moto/databrew/models.py
shield007: Ah yes I agree, will change that. Thanks.

moto/databrew/models.py
shield007: Good point, yes will do as you suggested.

moto/databrew/responses.py
shield007: Agreeed, though I copied that code from else where in moto. Though maybe it's use case is different. So will use as suggested.

bblommers (approved):
LGTM! Thanks again for the PR and for making the improvements, @shield007, much appreciated.

Issue #4997, opened 2022-04-01
Missing implementation of network attachment describe functionality


glungley:
`boto3.client('ec2').describe_network_interfaces()` and `boto3.client('ec2').describe_network_interface_attribute()` have not implemented the attachment parts.

PR #4998, opened 2022-04-01
#4997 implement network attachments

glungley:


moto/ec2/models.py
```diff
@@ -1047,6 +1049,7 @@ def attach_eni(self, eni, device_index):
         # This is used upon associate/disassociate public IP.
         eni.instance = self
         eni.attachment_id = random_eni_attach_id()
+        eni.attach_time = utc_date_and_time()
         eni.device_index = device_index
```
pablo1a, line 1054: it will be nice to update `eni.status = "in-use"` here as well, to reflect the ENI is in use by an instance once it's attached.

glungley: Good idea, added

moto/ec2/responses/elastic_network_interfaces.py
```diff
@@ -1,3 +1,4 @@
+from moto.awslambda.exceptions import InvalidParameterValueException
```
bblommers, line 1: This should come from EC2's exceptions (`from .exceptions import ..`)

I don't know if there is an exception already, but if not, we can just recreate it there

moto/ec2/responses/elastic_network_interfaces.py
```diff
@@ -39,16 +40,43 @@ def delete_network_interface(self):
...
+                    DESCRIBE_NETWORK_INTERFACE_ATTRIBUTE_RESPONSE_SOURCEDESTCHECK
+                )
+                return template.render(eni=eni)
+            elif attribute == "attachment":
+                template = self.response_template(
+                    DESCRIBE_NETWORK_INTERFACE_ATTRIBUTE_RESPONSE_ATTACHMENT
+                )
+                return template.render(eni=eni)
+            else:
+                raise InvalidParameterValueException(
```
bblommers, line 69: This case should be tested as well, to verify that the exception is returned in the correct format

bblommers (changes requested):
Hi @glungley, thanks for the PR! Just two minor comments - the rest looks good to me

moto/ec2/responses/elastic_network_interfaces.py
```diff
@@ -39,16 +40,43 @@ def delete_network_interface(self):
...
+        eni_id = self._get_param("NetworkInterfaceId")
+        attribute = self._get_param("Attribute")
+        if self.is_not_dryrun("DescribeNetworkInterfaceAttribute"):
+            eni = self.ec2_backend.get_all_network_interfaces([eni_id])[0]
+
+            if attribute == "description":
+                template = self.response_template(
+                    DESCRIBE_NETWORK_INTERFACE_ATTRIBUTE_RESPONSE_DESCRIPTION
+                )
+                return template.render(eni=eni)
```
bblommers, line 52: Let's move the return-statement to the bottom of the method, considering it is the same for all if/elif clauses

glungley:
@bblommers thanks, I've addressed your comments and squashed the commits

bblommers (approved):
LGTM - thanks again @glungley!

PR #4999, opened 2022-04-01
allow string types for vcpus and memory in parameter validation

coldfire79:
Since botocore requires a `str` type for both `memory` and `vcpus` (containerProperties.resourceRequirements) when trying register_job_definition, `memory` and `vcpus` should be allowed to be a string.

### DIFF
--- a/moto/batch/models.py
+++ b/moto/batch/models.py
@@ -313,13 +313,13 @@ class JobDefinition(CloudFormationModel):
         memory = self._get_resource_requirement("memory")
         if memory is None:
             raise ClientException("containerProperties must contain memory")
-        if memory < 4:
+        if float(memory) < 4:
             raise ClientException("container memory limit must be greater than 4")
 
         vcpus = self._get_resource_requirement("vcpus")
         if vcpus is None:
             raise ClientException("containerProperties must contain vcpus")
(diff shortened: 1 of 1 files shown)


bblommers:
Hi @coldfire79, the conversion to floats happens right before in `_get_resource_requirements`, so this should not be necessary:

https://github.com/spulec/moto/blob/3718cde444b3e0117072c29b087237e1787c3a66/moto/batch/models.py#L295:L296

https://github.com/spulec/moto/blob/3718cde444b3e0117072c29b087237e1787c3a66/moto/batch/models.py#L293:L294

Is there a specific use-case where you're seeing errors without this change?

coldfire79:
> Hi @coldfire79, the conversion to floats happens right before in `_get_resource_requirements`, so this should not be necessary:
> 
> https://github.com/spulec/moto/blob/3718cde444b3e0117072c29b087237e1787c3a66/moto/batch/models.py#L296
> 
> https://github.com/spulec/moto/blob/3718cde444b3e0117072c29b087237e1787c3a66/moto/batch/models.py#L294
> 
> Is there a specific use-case where you're seeing errors without this change?

Thank you. Maybe because I was using the previous version v2.2.19. Let me close this.

PR #5004, opened 2022-04-05
Enhance IAM credentials report to dynamic mfa_active and password_last_used

kentnsw:
Hi,

There're 2 changes in this PR. Please review. Thanks.
- add user property password_last_used to enable the possibility of dynamic it in the credentials report
- dynamic mfa_active in credentials report

Regarding password_last_used, a developer can use the following code to dynamic it before get_credential_report().
```
from moto.backends import get_backend
motoIam = get_backend("iam")["global"]
motoIam.users[userName].password_last_used = datetime.utcnow()
```


bblommers (changes requested):
Hi @kentnsw, can we have a test case for this, to verify that this works correctly?
There is an existing test that can be extended here: 
https://github.com/spulec/moto/blob/2aa3788174c475da08ada8e925428a69c107c509/tests/test_iam/test_iam.py#L1759

Note that calling the internal API is only possible when using decorators, so we should guard it to make sure we don't try to test it in ServerMode. The existing test already has an example of how this should be done:
```
    iam_backend = get_backend("iam")["global"]
    iam_backend.users[userName].password_last_used = datetime.utcnow()

# get report
report.passwordlastused.equals(..)
```

kentnsw:
Hi @bblommers, thanks for your review and the great tips.

PR #5005, opened 2022-04-05
Make docs on AWS_... env vars consistent, cross reference FAQ

huonw:
This applies two tweaks to the discussion of the `AWS_...` environment variables in the documentation:

- set the same variables in the first block as in the following section (about `pytest`)
- reference the FAQ entry that discusses how to use a region name that AWS doesn't support

(Thanks for moto!)


bblommers (approved):
LGTM - documentation improvements are always welcome. Thanks for the PR @huonw!

PR #5006, opened 2022-04-05
Use correct service name for "ecr" in `BackendDict`

garyd203:
I am pretty sure this should be "ecr" (the AWS name for the ECR service), not "ec2".

I presume the bug arises from a copy-paste error.

The service name appears to only be used to lookup supported regions, which is probably why this bug hasn't had any impact on users.


bblommers (approved):
LGTM - thanks for the fix, @garyd203!

PR #5013, opened 2022-04-08
Support Expression in Glue's get_partitions

robert-schmidtke:
I am using `moto` a lot and it's a great help for me, so thanks a lot for putting in all the work!
There's a feature missing I'd like to have, hence this PR.

This PR adds (experimental) support for [Glue GetPartitions](https://docs.aws.amazon.com/glue/latest/webapi/API_GetPartitions.html) `Expression`s.
It does so by mimicking Glue's partition filtering as closely as possible using a `pyparsing` implementation of a very small subset of SQL (parsed by JSQLParser according to the docs).
The version of `pyparsing` required is 3.0.0, but only because that used PEP8 compliant naming. I do not believe there is an actual dependency on that version, so relaxing the requirement there should be possible, at the cost of switching between `camelCase` for pre-PEP8 and `snake_case` as it is now.
I figured since `pyparsing` is pulled in anyway, this is not too big of a dependency to add.

I am happy about any feedback regarding usage of `pyparsing`, as this is honestly my first time implementing a parser.
Also with respect to code style, abstraction etc, I am happy to make changes as necessary to keep the codebase consistent.

The test cases included produce 100% test coverage, and are based on documentation and pure trial and error.
On a side note, I wasn't able to get the terraform tests to succeed in my fork, any idea what might be going on there?

# Known partial implementations
I have added a `NOTE` and/or `warning` in the code for each of these.
- AWS allows date and timestamp filtering including the 31st of every month. I do not know how they do it, but my guess is they do regex based matching and then clamp. I have not added this "feature", so the implementation below will fail for invalid dates and timestamps, emitting a warning to make the user aware.
- I have not found a way to match an underscore literally (escaping using `[_]` does not work on AWS). Hence I think there may be some subtleties in how I convert the SQL LIKE clause to a Python regex in order to match it.
- Some error messages may not match 100% what AWS returns, do you consider that a big issue?
- There is no support for nanoseconds in `datetime`, hence the implementation below rounds to `microseconds` when comparing timestamps.
- The column names currently follow Athena best practices, instead of the [Column Name](https://docs.aws.amazon.com/glue/latest/webapi/API_Column.html) regex. I think I remember stumbling across a regex pattern way more complex than this, and I wasn't able to get it to work properly. I can take another stab at it though.


bblommers:
Hi @robert-schmidtke, thanks for the PR! Good to hear that Moto is useful for you.

Considering the size of this PR, it may be a while before this is reviewed in full, so bear with us. Some initial thoughts in the meantime:

> The version of `pyparsing` required is 3.0.0

That's fine with me

> On a side note, I wasn't able to get the terraform tests to succeed in my fork, any idea what might be going on there?

Yeah, that's a known issue - don't worry about it. It does pass when running against this repo, so that's all good.

> I have added a `NOTE` and/or `warning` in the code for each of these.

Could you also add a docstring to `def get_partitions(self, expression):` with (a summary of a) documentation what is supported, and what isn't?
All docstrings will show up on `docs.getmoto.org`, so it's an easy way for users to see what's supported without having to try it out.

>     * Some error messages may not match 100% what AWS returns, do you consider that a big issue?

No, that's not a big issue. Ideally it's a 100% match obviously, but if that's not possible, as long as we return the correct error code, that's OK as well.

The failing test is because the warnings will not be omitted in ServerMode tests. ServerMode means we're running Moto in a separate process/container, so there is no way to catch warnings.
The usual way around this is to guard that particular check to only run when not in ServerMode:
```
    # catch warnings
```
This structure is used in a few places in the tests. An example can be found here:
https://github.com/spulec/moto/blob/a0eb48d588d8a5b18b1ffd9fa6c48615505e3c1d/tests/test_acm/test_acm.py#L591

moto/glue/utils.py
```diff
@@ -0,0 +1,378 @@
...
+
+
+class _Ident(_Expr):
+    def __init__(self, tokens: ParseResults):
+        self.ident: str = tokens[0]
+
+    def eval(self, part_keys: List[Dict[str, str]], part_input: Dict[str, Any]) -> Any:
+        for key, value in zip(part_keys, part_input["Values"]):
+            if self.ident == key["Name"]:
+                try:
```
bblommers, line 110: There seems to be some duplication between `eval` and `leval` - is there a way to reduce this? Maybe by replacing this try-except with a call to `leval` instead?

moto/glue/utils.py
```diff
@@ -0,0 +1,378 @@
...
+                microseconds = (int(nanoseconds) * 10**i) / 1000
+                if round(microseconds) == 0 and microseconds > 0:
+                    warnings.warn("Nanoseconds not supported, rounding to microseconds")
+
+                timestamp += timedelta(microseconds=round(microseconds))
+
+        return timestamp
+
+    raise InvalidInputException(
+        "An error occurred (InvalidInputException) when calling the"
```
bblommers, line 92: The first part of the error message looks to be the same for all messages - we should extract that into the `InvalidInputException`-class instead

moto/glue/utils.py
```diff
@@ -0,0 +1,378 @@
...
+
+        nanos = match.group("nanos")
+        if nanos is not None:
+            # strip leading dot, reverse and left pad with zeros to nanoseconds
+            nanos = "".join(reversed(nanos[1:])).zfill(9)
+            for i, nanoseconds in enumerate(nanos):
+                # NOTE precision loss here, as nanoseconds are not supported in datetime
+                microseconds = (int(nanoseconds) * 10**i) / 1000
+                if round(microseconds) == 0 and microseconds > 0:
+                    warnings.warn("Nanoseconds not supported, rounding to microseconds")
```
bblommers, line 85: Considering Moto will never deal with nanoseconds, I don't think this warning is necessary. Let's just treat it as an implementation detail. It should be part of the documentation in the docstring of `get_partition`, but we don't have to warn the user everytime.

moto/glue/utils.py
```diff
@@ -0,0 +1,378 @@
...
+        }[self.bin_op](ident, rhs)
+
+
+class _Like(_Expr):
+    def __init__(self, tokens: ParseResults):
+        self.ident: _Ident = tokens[0]
+        self.literal: str = tokens[2]
+
+    def eval(self, part_keys: List[Dict[str, str]], part_input: Dict[str, Any]) -> bool:
+        warnings.warn("LIKE expression conversion to regex is experimental")
```
bblommers, line 193: Same here w.r.t. the warning - this should be part of the documentation, but we don't have to warn people every time. Either it works correctly, or it doesn't.

moto/glue/utils.py
```diff
@@ -0,0 +1,378 @@
...
+
+    if type_ in ("char", "string", "varchar"):
+        return value  # no length is enforced
+
+    if type_ == "date":
+        try:
+            return datetime.strptime(value, "%Y-%m-%d").date()
+        except ValueError:
+            # NOTE AWS accepts up until 31st of every month, e.g. '2022-02-31'
+            warnings.warn("Date filtering beyond last of month not supported")
```
bblommers, line 52: Is this a relevant warning here? Looks to me like this warning will only be thrown if the date format is unknown, but we never get to the point where the date is valid but unsupported

bblommers (changes requested):

moto/glue/utils.py
robert-schmidtke: Yes, you're right. I wanted to make the user aware of the fact that their date may be wrong but actually supported by AWS, e.g. February 31st. In that case their production code would work, but the tests using moto would not. Hence I wanted the warning out there to point them to this potential source of the error.

But I agree, we can remove the warning here, as the ValueError will show the faulty date.

moto/glue/utils.py
robert-schmidtke: Agree

moto/glue/utils.py
robert-schmidtke: I thought about the same actually, but note how `eval` raises an `InvalidStateException`, while `leval` raises an `InvalidInputException`. This is because `eval` applies to existing partition values, and when an existing value cannot be cast to the schema (possibly because the schema changed after the initial value was written), an `InvalidStateException` is raised.

Correspondingly, when the user supplies a value that cannot be cast to the target partition column type, this is an `InvalidInputException`.

Maybe I can extract an `_eval` method and catch the `ValueError` separately.

bblommers: Yeah, I saw there were some differences, that's why I raised it as a question, instead of a request. If it doesn't work, or if the resulting code looks worse, than don't worry about it!

robert-schmidtke:
@bblommers I hope I addressed your concerns appropriately. Waiting to hear back from you :)

robert-schmidtke:
@bblommers thanks for kicking this off, I'll take a look at the failing 3.6 test!

bblommers:
> @bblommers thanks for kicking this off, I'll take a look at the failing 3.6 test!

I'm expecting this to be an unrelated failure - I've restarted it, it should pass on the secondrun

robert-schmidtke:
> > @bblommers thanks for kicking this off, I'll take a look at the failing 3.6 test!
> 
> I'm expecting this to be an unrelated failure - I've restarted it, it should pass on the secondrun

It's actually got to do with how regex escaping changed from Python 3.6 to 3.7, wrt. the percentage sign %. I pushed a fix just now.

bblommers (approved):
LGTM! Thank you for supporting us and for implementing this feature, @robert-schmidtke, much appreciated!

PR #5023, opened 2022-04-13
rds - support final snapshot when delete a cluster

kentnsw:
create a final snapshot if FinalDBSnapshotIdentifier is given while deleting a rds cluster

### DIFF
--- a/tests/test_rds/test_rds_clusters.py
+++ b/tests/test_rds/test_rds_clusters.py
@@ -211,6 +211,25 @@ def test_delete_db_cluster():
 
 
 @mock_rds
+def test_delete_db_cluster_do_snapshot():
+    client = boto3.client("rds", region_name="eu-north-1")
+
+    client.create_db_cluster(
+        DBClusterIdentifier="cluster-id",
+        Engine="aurora",
+        MasterUsername="root",
+        MasterUserPassword="hunter2_",
+    )
+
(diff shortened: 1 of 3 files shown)


tests/test_rds/test_rds_clusters.py
```diff
@@ -210,6 +210,25 @@ def test_delete_db_cluster():
...
+        MasterUsername="root",
+        MasterUserPassword="hunter2_",
+    )
+
+    client.delete_db_cluster(
+        DBClusterIdentifier="cluster-id", FinalDBSnapshotIdentifier="final-ss"
+    )
+    client.describe_db_clusters()["DBClusters"].should.have.length_of(0)
+    snapshots = client.describe_db_cluster_snapshots()["DBClusterSnapshots"]
+    snapshots[0]["DBClusterIdentifier"] == "cluster-id"
```
bblommers, line 229: Can we get another assertion here on the snapshot name, just to double check the `final-ss` is stored correctly?

bblommers (changes requested):

tests/test_rds/test_rds_clusters.py
kentnsw: sure thing

bblommers (approved):
LGTM! Thanks for adding this feature @kentnsw

PR #5024, opened 2022-04-14
rds - modify_db_instance does not support tags

kentnsw:
This pr is to remove the `tags` param in the modify_db_instance implementation as the boto3 API does not support that, otherwise, the existing tags will be cleared.

NOTE: I haven't checked other language SDKs.


bblommers (approved):
LGTM! Thanks for the fix @kentnsw

PR #5032, opened 2022-04-16
ELB: Fix response templates

viren-nadkarni:
### Summary
This PR fixes response template for `SetRulePriorities` operations. The template now correctly renders the list of [Rule objects](https://docs.aws.amazon.com/elasticloadbalancing/latest/APIReference/API_SetRulePriorities.html#API_SetRulePriorities_ResponseElements).

The list of valid [Load Balancer attributes](https://docs.aws.amazon.com/elasticloadbalancing/latest/application/application-load-balancers.html#load-balancer-attributes) is also updated.


bblommers (approved):
LGTM! Thanks for the PR @viren-nadkarni

PR #5038, opened 2022-04-18
Refer to correct black version

iainelder:
The correct black version is the one in requirements-dev.txt.

Instead of keeping the docs and Makefile in sync they now point to the source of truth.


bblommers (approved):
Makes sense - thanks @iainelder  :+1:

PR #5039, opened 2022-04-18
fix(swf_tests): compare tz-aware timestamps

iainelder:
My computer's local time zone is CEST. Without modifications I could make the
tests pass only by setting the environment variable `TZ=Etc/UTC`.


bblommers (approved):
Thanks for the PR @iainelder! This did fail for other people as well, but we never got around to fixing this.

PR #5041, opened 2022-04-20
cloudwatch: filter 'dimensions' for get_metric_data

steffyP:
This PR enhances the `get_metric_data` request of Cloudwatch.

If "dimensions" are part of the metric-queries for `get_metric_data`, the logs will be filtered accordingly.

[The docs state](https://aws.amazon.com/premiumsupport/knowledge-center/cloudwatch-getmetricstatistics-data/):
> If the metric is created with multiple dimensions, then the data points for that metric can be retrieved only by specifying all the configured dimensions.


bblommers (approved):
LGTM! Thanks for the PR @steffyP

PR #5050, opened 2022-04-22
Autoscaling - create custom BlockDevice in addition to default image block device

bblommers:
- BlockDeviceMappings defined in a launch config/launch template should be created in addition to the default block device that will be created for an instance.
   (This is a follow-up to #5044, where the user-defined mappings replaces the default device)
 - Adds the parameters ClassicLinkVPCId, ClassicLinkVPCSecurityGroups, MetadataOptions to the `create_launch_configuration` method
 - Fix the ARN creation for launch configurations
 - Persist the Throughput/Encrypted/NoDevice parameters when configuring BlockDeviceMappings

EC2 improvements:
 - Don't throw an error when describing an image that has been deleted recently (behaviour confirmed against AWS)
 - Deregistering the same image twice throws a custom error (behaviour confirmed against AWS)

### DIFF
--- a/moto/autoscaling/models.py
+++ b/moto/autoscaling/models.py
@@ -131,6 +131,10 @@ class FakeLaunchConfiguration(CloudFormationModel):
         ebs_optimized,
         associate_public_ip_address,
         block_device_mapping_dict,
+        region_name,
+        metadata_options,
+        classic_link_vpc_id,
+        classic_link_vpc_security_groups,
     ):
         self.name = name
         self.image_id = image_id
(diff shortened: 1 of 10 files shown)


bblommers:
>     * Don't throw an error when describing an image that has been deleted recently (behaviour confirmed against AWS)
> 

To follow-up on this - the command still succeeds  48 hours after deletion.
(ami-aaaaaaa776 is the AMI of the deleted image)

```bash

$ aws ec2 describe-images --image-id ami-aaaaaaa776 --region us-east-1
{
    "Images": []
}

$ aws ec2 describe-images --image-id ami-aaaaaaa777 --region us-east-1

An error occurred (InvalidAMIID.Malformed) when calling the DescribeImages operation: Invalid id: "ami-aaaaaaa777"

$ aws ec2 describe-images --image-id ami-aaaaaaa766 --region us-east-1

An error occurred (InvalidAMIID.NotFound) when calling the DescribeImages operation: The image id '[ami-aaaaaaa766]' does not exist
```
Looks like AWS keeps quite a long record of recently deleted AMI's.

PR #5056, opened 2022-04-24
events: delete_rule implementation update

rangerthegood:
events, delete_rule should throw a ValidationException when attempting to delete a rule which has targets. This PR adds an additional logic, one test case, and modifies two others tests.

The boto3 docs don't really outline this specific `ValidationException`, but they say "Before you can delete the rule, you must remove all targets, using RemoveTargets."

Through testing using my live account I found that it the API does raise a ValidationException with the string "Rule can't be deleted since it has targets."

### DIFF
--- a/moto/events/models.py
+++ b/moto/events/models.py
@@ -1045,7 +1045,11 @@ class EventsBackend(BaseBackend):
         return rule
 
     def delete_rule(self, name):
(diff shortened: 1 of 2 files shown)


bblommers (approved):
LGTM - thank you for contributing this fix @rangerthegood!

PR #5060, opened 2022-04-25
Fix `ResourceWarning` in `moto.settings.is_docker()`.

ngnpope:
The `/proc/self/cgroup` file is not closed properly after reading which causes a `ResourceWarning` to be raised.


bblommers (approved):
Nice catch @ngnpope - thank you for contributing a fix!

PR #5070, opened 2022-04-27
mediaconnect: fix flow and source arn format and update tests to use …

nickphbrown:
…valid name format

### DIFF
--- a/tests/test_mediaconnect/test_mediaconnect.py
+++ b/tests/test_mediaconnect/test_mediaconnect.py
@@ -1,9 +1,12 @@
+from uuid import UUID
+
 import boto3
 import pytest
 import sure  # noqa # pylint: disable=unused-import
 from botocore.exceptions import ClientError
 
 from moto import mock_mediaconnect
+from moto.core import ACCOUNT_ID
 
 region = "eu-west-1"
 
@@ -18,7 +21,7 @@ def _create_flow_config(name, **kwargs):
                 "Description": "An entitlement",
                 "Encryption": {"Algorithm": "aes256", "RoleArn": "some:role"},
                 "EntitlementStatus": "ENABLED",
-                "Name": "Entitlement A",
+                "Name": "Entitlement-A",
                 "Subscribers": [],
             }
         ],
@@ -26,9 +29,9 @@ def _create_flow_config(name, **kwargs):
     outputs = kwargs.get(
         "outputs",
         [
-            {"Name": "Output 1", "Protocol": "zixi-push"},
-            {"Name": "Output 2", "Protocol": "zixi-pull"},
-            {"Name": "Output 3", "Protocol": "srt-listener"},
+            {"Name": "Output-1", "Protocol": "zixi-push"},
+            {"Name": "Output-2", "Protocol": "zixi-pull"},
+            {"Name": "Output-3", "Protocol": "srt-listener"},
         ],
     )
     source = kwargs.get(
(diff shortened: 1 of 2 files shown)


nickphbrown:
I can't find any easy to read evidence of flow/source ARN format. However, it can be worked out from the example at the bottom of this page https://docs.aws.amazon.com/mediaconnect/latest/ug/flows-create-standard-source.html

where the relevant details in the example are:
- flow arn: "arn:aws:mediaconnect:us-east-1:111122223333:flow:1-23aBC45dEF67hiJ8-12AbC34DE5fG:AwardsShow"
implying the format for flow arn is "arn:aws:mediaconnect:{region}:{account_id}:flow:{id}:{flow_name}"

- source arn: "arn:aws:mediaconnect:us-east-1:111122223333:source:3-4aBC56dEF78hiJ90-4de5fG6Hi78Jk:AwardsShowSource"
implying the format for source arn is "arn:aws:mediaconnect:{region}:{account_id}:source:{id}:{source_name}"

Weirdly the id values in the example differ for the flow and the source ARNs, however from experience I've only ever seen the ids for the flow and its sources be the same. That observation may not always be true but I haven't seen an exception

tests/test_mediaconnect/test_mediaconnect.py
```diff
@@ -209,13 +223,15 @@ def test_add_flow_vpc_interfaces_fails():
         client.add_flow_vpc_interfaces(FlowArn=flow_arn, VpcInterfaces=[])
     err = err.value.response["Error"]
     err["Code"].should.equal("NotFoundException")
-    err["Message"].should.equal("flow with arn=unknown-flow not found")
+    err["Message"].should.equal(
```
bblommers, line 226: This looks like a dodgy merge - this was changed as part of our flake8-upgrade, but this PR reverses it again

tests/test_mediaconnect/test_mediaconnect.py
```diff
@@ -254,13 +270,15 @@ def test_remove_flow_vpc_interface_fails():
         )
     err = err.value.response["Error"]
     err["Code"].should.equal("NotFoundException")
-    err["Message"].should.equal("flow with arn=unknown-flow not found")
+    err["Message"].should.equal(
```
bblommers, line 273: Same

tests/test_mediaconnect/test_mediaconnect.py
```diff
@@ -289,7 +307,9 @@ def test_add_flow_outputs_fails():
         client.add_flow_outputs(FlowArn=flow_arn, Outputs=[])
     err = err.value.response["Error"]
     err["Code"].should.equal("NotFoundException")
-    err["Message"].should.equal("flow with arn=unknown-flow not found")
+    err["Message"].should.equal(
```
bblommers, line 310: Same

tests/test_mediaconnect/test_mediaconnect.py
```diff
@@ -301,13 +321,15 @@ def test_remove_flow_output_fails():
         client.remove_flow_output(FlowArn=flow_arn, OutputArn=output_arn)
     err = err.value.response["Error"]
     err["Code"].should.equal("NotFoundException")
-    err["Message"].should.equal("flow with arn=unknown-flow not found")
+    err["Message"].should.equal(
```
bblommers, line 324: Same

bblommers (changes requested):
Yesterday we upgraded flake8 to 4.x - looks like this PR tries to reverse some of the changes, which causes the linter to fail.
https://github.com/spulec/moto/runs/6200952774?check_suite_focus=true

The change itself looks fine - we'll just need to the build to pass

tests/test_mediaconnect/test_mediaconnect.py
nickphbrown: woops, sorry, got myself a bit confused trying to get the testing/linting working locally. Thanks for pointing out the issue

bblommers (approved):
LGTM! Thanks for the fix @nickphbrown

PR #5077, opened 2022-04-29
upgrade black and unpin click

huang06:
The upstream issue <https://github.com/pallets/click/issues/2232> has been solved.

We can upgrade `black` to the latest version and do not need to pin `click`'s version.

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -3,6 +3,7 @@
 
 black==22.3.0
 flake8==4.0.1
+click
 inflection
 lxml
 packaging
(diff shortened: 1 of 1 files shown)


requirements-dev.txt
```diff
@@ -1,10 +1,8 @@
 -e .[all,server]
 -r requirements-tests.txt
 
-black==22.1.0
+black==22.3.0
 flake8==4.0.1
-# https://github.com/pallets/click/issues/2232
-click<8.1.0
```
bblommers, line 7: We do need to keep `click` as a requirement though :)

bblommers (changes requested):

requirements-dev.txt
huang06: indeed `click` is required by scripts/scaffold.py

bblommers (approved):
Thanks @huang06!

PR #5078, opened 2022-04-29
Support SES get_identity_verification_attributes

Amwam:
Adds the get_identity_verification_attributes API

I used the scripts to generate the code, the took the XML from a debug output of awscli.

I'm not too sure if this is all correct, however the test passes. 
I wasn't sure if there was a way of setting a status _other_ than Success, although I feel that might mean we would need to explicitly record/track other status separately from the core boto3 API?

### DIFF
--- a/moto/ses/models.py
+++ b/moto/ses/models.py
@@ -523,5 +523,16 @@ class SESBackend(BaseBackend):
 
         return attributes_by_identity
 
+    def get_identity_verification_attributes(self, identities=None):
+        if identities is None:
+            identities = []
+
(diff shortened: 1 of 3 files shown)


bblommers:
> 
> I wasn't sure if there was a way of setting a status _other_ than Success, although I feel that might mean we would need to explicitly record/track other status separately from the core boto3 API?

Well, this looks like a good start - if people need the ability to track other statuses we can always refine it. 

Thanks for the PR @Amwam!

bblommers (approved):

PR #5080, opened 2022-04-29
Support all Action types in ELBv2 create_rule and set_rule_priorities responses

stephaniemanning:
This PR adds support for all Action types in ELBv2 create_rule and set_rule_priorities responses.


bblommers (approved):
Good idea to use the `to_xml`-method here @stephaniemanning - that makes things much more readable.
Thanks for the PR!

PR #5082, opened 2022-04-30
Sagemaker: add support for update_endpoint_weights_and_capacities 

Killian168:
This PR adds support for the `update_endpoint_weights_and_capacities` route for sagemaker. Also fixes a small bug in the way `production_variants` were being stored in FakeEndpoint which was discovered when writing the tests using `describe_endpoint`.

Boto3 documentation: [here](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/sagemaker.html#SageMaker.Client.update_endpoint_weights_and_capacities)


moto/sagemaker/models.py
```diff
@@ -1890,6 +1928,38 @@ def list_training_jobs(
...
+
+            for index, variant in enumerate(endpoint.production_variants):
+                if variant.get("VariantName") == name:
+                    variant["DesiredWeight"] = desired_weight
+                    variant["CurrentWeight"] = desired_weight
+                    variant["DesiredInstanceCount"] = desired_instance_count
+                    variant["CurrentInstanceCount"] = desired_instance_count
+                    break
+                elif index == len(endpoint.production_variants) - 1:
+                    raise AWSValidationException(
```
bblommers, line 1956: If there are two variants, and the user wants to update 3 variants where the second one doesn't exist, this would update the first, but throw an error on the second - and completely ignore the third. Am I understanding that correctly? Is that how AWS behaves?

Normally a check like that happens in the beginning, but it wouldn't surprise me if this is how AWS implements it..

bblommers, line 1956: Second point - do we need to reset the status here? Or does AWS also keep the status as `Updating` if anything goes wrong?

bblommers (changes requested):

Killian168:
Hello @bblommers, thanks for the review. You were correct in your questions. I've fixed the implementation and added some more test cases to cover some that I found when testing how boto3/sagemaker reacts on my personal AWS account.

Please have another look and see if there is anything that I should have another look at or change.

bblommers (approved):
LGTM! Thanks for investigating and fixing this @Killian168 :+1:

PR #5093, opened 2022-05-04
Bugfix/mediaconnect create flow include source ingest ip

nickphbrown:
mediaconnect:
- Adding "ingestIp" to non entitlement flow sources on "create_flow"
- moving flow outputs "listenerAddress" update so that it doesn't error when "outputs" argument isn't passed to "create_flow" method
- moving all "create_flow" flow updates to their own function to keep them all in one place


bblommers (approved):
LGTM! Thanks for the fixes @nickphbrown

PR #5098, opened 2022-05-05
add indirection to access account id

thrau:
This change unifies the access to the global `ACCOUNT_ID` constant, and introduces a layer of indirection to make it easier for third-party systems to modify how the current account id is resolved at runtime.

The major change for the moto code base moving forward with this change, is that moto modules should now use `get_account_id` rather than importing `moto.core.ACCOUNT_ID` directly.
As briefly discussed with @bblommers, the change should be backwards compatible.
To demonstrate, I left the access to `ACCOUNT_ID` in the tests unchanged (just made a couple of import fixes).
Existing code will therefore not break (both `moto.core.models.ACCOUNT_ID` and `moto.core.ACCOUNT_ID` will be resolved once to the environment variable `MOTO_ACCOUNT_ID`.

Here's how the indirection can be used:
```python
import os

os.environ['MOTO_ACCOUNT_ID'] = 'hello'

import moto.core.models
from moto.core import get_account_id


def main():
    print(moto.core.ACCOUNT_ID)  # prints "hello"
    print(get_account_id())  # prints "hello"

    moto.core.models.ACCOUNT_ID = "overwritten"
    print(get_account_id())  # print "overwritten"

    def _custom_account_id_resolver():
        return "my-account-id"

    moto.core.models.account_id_resolver = _custom_account_id_resolver
    print(get_account_id())  # prints "my-account-id"

if __name__ == '__main__':
    main()
```

one open task would be to eliminate static rendering of the account id in all templates, like here:
https://github.com/spulec/moto/blob/c2727a7c205d183034f1fc14ac09234134c9135e/moto/ec2/responses/instances.py#L382-L388

a quick search seems to indicate that there are only a few of them left though


moto/core/utils.py
```diff
@@ -411,6 +411,7 @@ def extract_region_from_aws_authorization(string):
 class BackendDict(dict):
     def __init__(self, fn, service_name):
         self.fn = fn
+        self.service_name = service_name
```
bblommers, line 414: What's the reason behind this change?

thrau: seems this one slipped in. we're patching the backend dict to return account specific instances, to be able to do that we need to remember which service the originally patched backend belongs to. it's basically just to help extensibility.

bblommers (approved):
LGTM! Thanks for the PR @thrau

thrau:
Thanks for merging so quickly Bert!

PR #5126, opened 2022-05-12
Fix dashboard not to show JSON as '[object Object]'

shibukawa:
Original dashboard can't show JSON (it results `[object Object]`). My fix shows JSON content.

Before 
スクリーンショット 2022-05-12 10 30 37

After
スクリーンショット 2022-05-12 23 58 14


bblommers (approved):
Thanks for the PR @shibukawa - this makes things much more readable!

PR #5127, opened 2022-05-13
Feature/implement glue tags methods

joshuaghezzi:
Implemented 3 glue client methods
- glue get_tags()
- glue tag_resource()
- glue untag_resource()

### DIFF
--- a/moto/glue/models.py
+++ b/moto/glue/models.py
@@ -3,6 +3,7 @@ from collections import OrderedDict
 from datetime import datetime
 
 from moto.core import BaseBackend, BaseModel
+from moto.core.models import get_account_id
 from moto.glue.exceptions import CrawlerRunningException, CrawlerNotRunningException
 from .exceptions import (
     JsonRESTError,
@@ -20,6 +21,7 @@ from .exceptions import (
 )
 from .utils import PartitionFilter
 from ..utilities.paginator import paginate
+from ..utilities.tagging_service import TaggingService
 
 
 class GlueBackend(BaseBackend):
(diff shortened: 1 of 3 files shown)


moto/glue/models.py
```diff
@@ -502,13 +522,16 @@ def __init__(
...
         self.notification_property = notification_property
         self.glue_version = glue_version
         self.number_of_workers = number_of_workers
         self.worker_type = worker_type
         self.created_on = datetime.utcnow()
         self.last_modified_on = datetime.utcnow()
+        self.arn = f"arn:aws:glue:us-east-1:{get_account_id()}:job/{self.name}"
+        self.backend = backend
+        self.backend.tag_resource(self.arn, tags)
+        self.tags = tags
```
bblommers, line 534: Do we still need this assignment?
As far as I can tell, this was only used in responses.py:;filter_jobs_by_tags - and that method has refactored to use `get_tags`

moto/glue/responses.py
```diff
@@ -328,7 +328,7 @@ def filter_crawlers_by_tags(self, crawlers, tags):
         return [
             crawler.get_name()
             for crawler in crawlers
-            if self.is_tags_match(crawler.tags, tags)
+            if self.is_tags_match(self.glue_backend.get_tags(crawler.arn), tags)
```
bblommers, line 331:
```suggestion
            if self.is_tags_match(crawler.arn, tags)
```
This would be easier to read. It does mean we have to move the `get_tags()`-call to inside the `is_tags_match`-method, and remove the duplication in `filter_jobs_by_tags`

bblommers (changes requested):

moto/glue/responses.py
joshuaghezzi: Agreed. I've moved that get_tags() call into the is_tags_match method.

moto/glue/models.py
joshuaghezzi: No, that assignment is no longer needed. Have removed now.

bblommers (approved):
LGTM - thanks for the PR and for making the changes @joshuaghezzi!

Issue #5128, opened 2022-05-13
Index error in moto server when attempting to create Instance with availability zone not matching the region


szopen321:
**Repro steps:**
    1. Start moto server.
    2. Prepare terraform file that will:
        ◦ connect to the started moto server ( add: endpoints {ec2 = "http://127.0.0.1:5000/"} in provider block)
        ◦ contain defined region (eg.: us-east-1) 
        ◦ contain availability zone from a different region (eg.: us-west-2b)
    3. Run terraform init; terraform apply
    4. Type yes and press enter when prompted.

**Description:**
When attempting to create an instance with availability zone not matching region on running moto server one of the functions (get_default_subnet) will throe an IndexError: list index out of range.

**Observed results:**
Index error in moto server when attempting to create Instance with availability zone not matching the region

**Expected results:**
Informative error with status code 400 - matching true AWS behaviour when attempting to create Instance with availability zone not matching the region.

szopen321:
PR with a proposed fix

PR #5129, opened 2022-05-13
Add availability zone check on add instance

szopen321:
Fix to:
https://github.com/spulec/moto/issues/5128

### DIFF
--- a/moto/ec2/models/instances.py
+++ b/moto/ec2/models/instances.py
@@ -544,11 +544,11 @@ class InstanceBackend(object):
     def add_instances(self, image_id, count, user_data, security_group_names, **kwargs):
         location_type = (
             "availability-zone"
-            if "placement" in kwargs and kwargs["placement"]
+            if kwargs.get("placement")
             else "region"
         )
         valid_instance_types = INSTANCE_TYPE_OFFERINGS[location_type]
-        if "placement" in kwargs and "region_name" in kwargs and kwargs["placement"]:
+        if "region_name" in kwargs and kwargs.get("placement"):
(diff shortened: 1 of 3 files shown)


moto/ec2/models/instances.py
```diff
@@ -538,6 +535,12 @@ def get_instance(self, instance_id):
         raise InvalidInstanceIdError(instance_id)
 
     def add_instances(self, image_id, count, user_data, security_group_names, **kwargs):
+        location_type = "availability-zone" if "placement" in kwargs and kwargs["placement"] else "region"
```
bblommers, line 538:
```suggestion
        location_type = "availability-zone" if kwargs.get("placement") else "region"
```
This is logically equivalent, I believe, and more readable

moto/ec2/models/instances.py
```diff
@@ -538,6 +535,12 @@ def get_instance(self, instance_id):
...
+        valid_instance_types = INSTANCE_TYPE_OFFERINGS[location_type]
+        if "placement" in kwargs and "region_name" in kwargs and kwargs["placement"]:
```
bblommers, line 540:
```suggestion
        if "region_name" in kwargs and kwargs.get("placement"):
```
Same

bblommers (changes requested):
Hi @szopen321, thanks for the PR! Just got two minor comments. Looks like the linter is failing as well: https://github.com/spulec/moto/runs/6430459096?check_suite_focus=true

bblommers (approved):
LGTM! Thanks again @szopen321

Issue #5131, opened 2022-05-13
No instance type validation on run instance


szopen321:
Repro steps:
    1. Start moto server.
    2. Prepare terraform file that will contain invalid instance type
    3. Run terraform init; terraform apply

Description:
When attempting to create an instance of inexistent instance type, moto doesn't verify that and creates an instance

Observed results:
It is possible to create an instance of nonexistent type

expected results:
Informative error with status code 400 - matching true AWS behaviour when attempting to create Instance with an inexistent instance type

szopen321:
PR with proposed solution https://github.com/spulec/moto/pull/5132

PR #5132, opened 2022-05-13
Add instance type validation on add instance method

szopen321:
Fix to https://github.com/spulec/moto/issues/5131


szopen321:
It's WIP, I am aware some tests won't pass - working on it.

bblommers (commented):
The overall approach looks good to me - I'll review this properly once the other PR is merged, then we can rebase this branch and make sure it only contains the changes for the instance type validation.

bblommers (changes requested):
I've given this a bit more thought, and I'm worried about the implications. Moto is not guaranteed to have an up-to-date list of all instance types, and there is no way to extend that list, so we could lock people out by accident.

Are you ok with putting this check behind a feature flag? That way people can use the validation if they want to, but it will not make Moto unusable if you want to use a new instance type that Moto does not know yet.

As an example:
```
if settings.ec2_enable_instance_type_validation():
    raise ...
```
This flag can be added to `moto/settings.py` by reading an environment variable.

Note that this should default to False, to not break existing behavior for users.
On a next major release we can think about turning this on be default, and announce it as a breaking change.

bblommers:
Apologies, I should have mentioned this before - it's not possible to set the environment variable in ServerMode, so we can skip the test in ServerMode:
```
if settings.TEST_SERVER_MODE:
  raise Skip test("")
```

PR #5134, opened 2022-05-13
Fix `Exists` pattern matching with None values

### LINKED ISSUE #5125: EventBridge rule doesn't match null value with "exists" filter
### What I expected to happen

I expected moto to match null values with `{"exists": true}` as this is the way AWS behaves.

### What actually happens

Moto doesn't seem to match null values with `{"exists": true}`.

### How to reproduce the issue

Using `python 3.9.7`, run the following tests using `boto3==1.22.4`, `botocore==1.25.4` and `moto==3.1.8`. The first test passes but the second one throws 
```
AssertionError: Lists differ: [{'foo': '123', 'bar': '123'}] != [{'foo': '123', 'bar': '123'}, {'foo': None, 'bar': '123'}]

Second list contains 1 additional elements.
First extra element 1:
{'foo': None, 'bar': '123'}

- [{'bar': '123', 'foo': '123'}]
+ [{'bar': '123', 'foo': '123'}, {'bar': '123', 'foo': None}]
```

```python
import json
import unittest

import boto3
from moto import mock_events, mock_logs
from moto.core import ACCOUNT_ID


class TestPattern(unittest.TestCase):
    def setUp(self):
        self.pattern = {
            "source": ["test-source"],
            "detail-type": ["test-detail-type"],
            "detail": {
                "foo": [{"exists": True}],
                "bar": [{"exists": True}],
            },
        }

    def test_aws_matches_none_with_exists_filter(self):
        events_client = boto3.client("events")
        response = events_client.test_event_pattern(
            EventPattern=json.dumps(self.pattern),
            Event=json.dumps(
                {
                    "id": "123",
                    "account": ACCOUNT_ID,
                    "time": "2022-05-12T11:40:30.872473Z",
                    "region": "eu-west-1",
                    "resources": [],
                    "source": "test-source",
                    "detail-type": "test-detail-type",
                    "detail": {"foo": None, "bar": "123"},
                }
            ),
        )
        self.assertTrue(response["Result"])

    @mock_events
    @mock_logs
    def test_moto_matches_none_with_exists_filter(self):
        logs_client = boto3.client("logs")
        log_group_name = "test-log-group"
        logs_client.create_log_group(logGroupName=log_group_name)

        events_client = boto3.client("events")
        event_bus_name = "test-event-bus"
        events_client.create_event_bus(Name=event_bus_name)

        rule_name = "test-event-rule"
        events_client.put_rule(
            Name=rule_name,
            State="ENABLED",
            EventPattern=json.dumps(self.pattern),
            EventBusName=event_bus_name,
        )

        events_client.put_targets(
            Rule=rule_name,
            EventBusName=event_bus_name,
            Targets=[
                {
                    "Id": "123",
                    "Arn": f"arn:aws:logs:eu-west-1:{ACCOUNT_ID}:log-group:{log_group_name}",
                }
            ],
        )

        events_client.put_events(
            Entries=[
                {
                    "EventBusName": event_bus_name,
                    "Source": "test-source",
                    "DetailType": "test-detail-type",
                    "Detail": json.dumps({"foo": "123", "bar": "123"}),
                },
                {
                    "EventBusName": event_bus_name,
                    "Source": "test-source",
                    "DetailType": "test-detail-type",
                    "Detail": json.dumps({"foo": None, "bar": "123"}),
                },
            ]
        )

        events = sorted(
            logs_client.filter_log_events(logGroupName=log_group_name)["events"],
            key=lambda x: x["eventId"],
        )
        event_details = [json.loads(x["message"])["detail"] for x in events]

        self.assertEqual(
            event_details,
            [
                {"foo": "123", "bar": "123"},
                # Moto doesn't match this but it should
                {"foo": None, "bar": "123"},
            ],
        )

```

bpandola:
Exists matching works on the presence or absence of a field in the JSON of the event.
The value of the field is not considered when evaluating a match.[1]

* Add sentinel to distinguish between missing fields and existing fields with None values
* Update event pattern matching tests to cover this specific use-case.
* Add integrated test to cover this scenario within a larger context.

[1]: https://docs.aws.amazon.com/eventbridge/latest/userguide/eb-event-patterns-content-based-filtering.html

Closes #5125

### DIFF
--- a/moto/events/models.py
+++ b/moto/events/models.py
@@ -33,6 +33,9 @@ from uuid import uuid4
 
 from .utils import PAGINATION_MODEL
 
+# Sentinel to signal the absence of a field for `Exists` pattern matching
+UNDEFINED = object()
+
 
 class Rule(CloudFormationModel):
     Arn = namedtuple("Arn", ["service", "resource_type", "resource_id"])
@@ -827,7 +830,7 @@ class EventPattern:
         return self._does_event_match(event, self._pattern)
 
     def _does_event_match(self, event, pattern):
-        items_and_filters = [(event.get(k), v) for k, v in pattern.items()]
+        items_and_filters = [(event.get(k, UNDEFINED), v) for k, v in pattern.items()]
         nested_filter_matches = [
             self._does_event_match(item, nested_filter)
             for item, nested_filter in items_and_filters
@@ -856,7 +859,7 @@ class EventPattern:
         filter_name, filter_value = list(pattern.items())[0]
         if filter_name == "exists":
             is_leaf_node = not isinstance(item, dict)
-            leaf_exists = is_leaf_node and item is not None
+            leaf_exists = is_leaf_node and item is not UNDEFINED
             should_exist = filter_value
             return leaf_exists if should_exist else not leaf_exists
         if filter_name == "prefix":
--- a/tests/test_events/test_event_pattern.py
+++ b/tests/test_events/test_event_pattern.py
@@ -23,6 +23,7 @@ def test_event_pattern_with_nested_event_filter():
 def test_event_pattern_with_exists_event_filter():
     foo_exists = EventPattern.load(json.dumps({"detail": {"foo": [{"exists": True}]}}))
     assert foo_exists.matches_event({"detail": {"foo": "bar"}})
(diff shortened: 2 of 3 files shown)


bblommers:
Nice one - thanks @bpandola!

PR #5138, opened 2022-05-14
Add ability to list ECS services by launch type

alexandera5:
Add optional `launchType` parameter to the ECS service according to the latest [boto3](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/ecs.html#ECS.Client.list_services) version and AWS API


bpandola (changes requested):
This looks good.  Can you write a simple test case to cover it?

bpandola (approved):

PR #5139, opened 2022-05-16
Update getting_started.rst

acabra85:
* wrap the modules in single quotes as to support zshell [see](https://github.com/napari/napari/issues/2081)


bblommers (approved):
Makes sense. IIRC correctly there is a note in our documentation related to this, but it makes more sense to just use quotes everywhere. Thanks for the PR @acabra85!

PR #5142, opened 2022-05-17
add passthroughBehaviour to the integration

calvernaz:
Minor addition to the integration model object. The property `passthroughBehavior` is used in the template logic.

### DIFF
--- a/moto/apigateway/models.py
+++ b/moto/apigateway/models.py
@@ -126,6 +126,8 @@ class Integration(BaseModel, dict):
         uri,
         http_method,
         request_templates=None,
+        passthrough_behavior="WHEN_NO_MATCH",
+        cache_key_parameters=None,
         tls_config=None,
         cache_namespace=None,
         timeout_in_millis=None,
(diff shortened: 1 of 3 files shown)


moto/apigateway/models.py
```diff
@@ -126,14 +126,18 @@ def __init__(
         uri,
         http_method,
         request_templates=None,
+        passthrough_behavior="WHEN_NO_MATCH",
+        cache_key_parameters=None,
```
bblommers, line 130: This value is never read, so this always defaults to the empty list. Was that on purpose?

Looks like the test also fails because of this field.

bblommers (changes requested):

moto/apigateway/models.py
calvernaz: The cache_key_parameters defaults to empty list, at least the boto response documentation says so and by observation. The tests are failing because I missed to include the field in the assertion.

bblommers (approved):
LGTM! Thanks for the fix @calvernaz

PR #5144, opened 2022-05-17
add authorizerPayloadFormatVersion to the authorizer

calvernaz:
Adds the authoriser payload format version to the authoriser. This changes should only affect apigateway v2

### DIFF
--- a/moto/apigateway/models.py
+++ b/moto/apigateway/models.py
@@ -415,6 +415,7 @@ class Authorizer(BaseModel, dict):
                 "identity_validation_expression"
             )
         self["authorizerResultTtlInSeconds"] = kwargs.get("authorizer_result_ttl")
+        self["authorizerPayloadFormatVersion"] = kwargs.get("authorizer_payload_format_version")
 
     def apply_operations(self, patch_operations):
         for op in patch_operations:
@@ -437,6 +438,8 @@ class Authorizer(BaseModel, dict):
                 raise Exception('Patch operation for "%s" not implemented' % op["path"])
             elif "/type" in op["path"]:
                 self["type"] = op["value"]
+            elif "/authorizerPayloadFormatVersion" in op["path"]:
+                self["authorizerPayloadFormatVersion"] = op["value"]
             else:
                 raise Exception('Patch operation "%s" not implemented' % op["op"])
         return self
@@ -967,6 +970,7 @@ class RestAPI(CloudFormationModel):
         identity_source=None,
         identiy_validation_expression=None,
         authorizer_result_ttl=None,
+        authorizer_payload_format_version=None
     ):
         authorizer = Authorizer(
             authorizer_id=authorizer_id,
@@ -979,6 +983,7 @@ class RestAPI(CloudFormationModel):
             identity_source=identity_source,
             identiy_validation_expression=identiy_validation_expression,
             authorizer_result_ttl=authorizer_result_ttl,
+            authorizer_payload_format_version=authorizer_payload_format_version
         )
         self.authorizers[authorizer_id] = authorizer
         return authorizer
--- a/moto/apigateway/models.py
+++ b/moto/apigateway/models.py
@@ -415,7 +415,6 @@ class Authorizer(BaseModel, dict):
                 "identity_validation_expression"
             )
         self["authorizerResultTtlInSeconds"] = kwargs.get("authorizer_result_ttl")
-        self["authorizerPayloadFormatVersion"] = kwargs.get("authorizer_payload_format_version")
 
     def apply_operations(self, patch_operations):
         for op in patch_operations:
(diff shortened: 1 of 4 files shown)


bblommers (changes requested):
Would you mind adding a test for this? I know it's a small change, but it's always good to verify that it works correctly

bblommers (approved):
LGTM - thanks @calvernaz!

bblommers:
Hi @calvernaz, do you remember the usecase for this PR? I'm not able to reproduce this, i.e. creating a authorizer against AWS without version does no automatically assign one.
Does AWS add the version only for a specific parameter combination, maybe?

The request that I used:
```
{
    "authorizerType": "JWT",
    "identitySource": [
        "$request.header.Authorization"
    ],
    "jwtConfiguration": {
        "audience": [
            "sth"
        ],
        "issuer": "..."
    },
    "name": "testauth"
}
```

And response from AWS:
```
{
    "authorizerId": "..",
    "authorizerType": "JWT",
    "identitySource": [
        "$request.header.Authorization"
    ],
    "jwtConfiguration": {
        "audience": [
            "sth"
        ],
        "issuer": "..."
    },
    "name": "testauth"
}
```

The Terraform tests for `TestAccAPIGatewayV2Authorizer` also fail because of this change - they pass when I revert it.

calvernaz:
Hi @bblommers, so it should come from [here](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/apigatewayv2_authorizer#authorizer_payload_format_version), so it should be only available for v2 APIs. It's also available [here](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-apigatewayv2-authorizer.html) for CF.

I can't speak for why the terraform tests fail, but the setting is definitely there. 

The use case comes from the lambda authorizers input and output format. (https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-lambda-authorizer.html), 

> Hi @calvernaz, do you remember the usecase for this PR? I'm not able to reproduce this, i.e. creating a authorizer against AWS without version does no automatically assign one. Does AWS add the version only for a specific parameter combination, maybe?
> 
> The request that I used:
> 
> ```
> {
>     "authorizerType": "JWT",
>     "identitySource": [
>         "$request.header.Authorization"
>     ],
>     "jwtConfiguration": {
>         "audience": [
>             "sth"
>         ],
>         "issuer": "..."
>     },
>     "name": "testauth"
> }
> ```
> 
> And response from AWS:
> 
> ```
> {
>     "authorizerId": "..",
>     "authorizerType": "JWT",
>     "identitySource": [
>         "$request.header.Authorization"
>     ],
>     "jwtConfiguration": {
>         "audience": [
>             "sth"
>         ],
>         "issuer": "..."
>     },
>     "name": "testauth"
> }
> ```
> 
> The Terraform tests for `TestAccAPIGatewayV2Authorizer` also fail because of this change - they pass when I revert it.

bblommers:
The setting is definitely there, but I'm just curious at under which conditions it defaults to `2.0`. It doesn't always take that default value.

calvernaz:
> The setting is definitely there, but I'm just curious at under which conditions it defaults to `2.0`. It doesn't always take that default value.

My understanding is that it defaults to 2.0, for HTTP APIs, when there is no value. The documentation doesn't state that, but it says "if you need REST compatibility use 1.0". I think that's what serverless defaults to.

bblommers:
Not specifying this value for type=REQUEST actually throws an error. It's not stated in the documentation, but running this command:
```
aws apigatewayv2 create-authorizer --api-id ... --authorizer-type "REQUEST" --identity-source "" --name testauth2 --authorizer-uri "arn:aws:apigateway:us-east-1:lambda:path/2015-03-31/functions/arn:aws:lambda:us-east-1:0000:function:test/invocations"
```
returns this error from AWS:
```
An error occurred (BadRequestException) when calling the CreateAuthorizer operation: AuthorizerPayloadFormatVersion is a required parameter for REQUEST authorizer
```

So it will never default to anything. I'll try to create a PR shortly to mimic this error in Moto, and reverse the default value.

calvernaz:
> Not specifying this value for type=REQUEST actually throws an error. It's not stated in the documentation, but running this command:
> 
> ```
> aws apigatewayv2 create-authorizer --api-id ... --authorizer-type "REQUEST" --identity-source "" --name testauth2 --authorizer-uri "arn:aws:apigateway:us-east-1:lambda:path/2015-03-31/functions/arn:aws:lambda:us-east-1:0000:function:test/invocations"
> ```
> 
> returns this error from AWS:
> 
> ```
> An error occurred (BadRequestException) when calling the CreateAuthorizer operation: AuthorizerPayloadFormatVersion is a required parameter for REQUEST authorizer
> ```
> 
> So it will never default to anything. I'll try to create a PR shortly to mimic this error in Moto, and reverse the default value.

Thanks for getting to the bottom of this, clearly an oversight from my side.

PR #5147, opened 2022-05-17
fix template variable typo

radam9:
The template variable name for the elbv2 healthcheck protocol is incorrect, which causes the healthcheck protocol in any api call response to always be an empty string.

### DIFF
--- a/moto/elbv2/responses.py
+++ b/moto/elbv2/responses.py
@@ -804,7 +804,7 @@ CREATE_TARGET_GROUP_TEMPLATE = """<CreateTargetGroupResponse xmlns="http://elast
         {% if target_group.vpc_id %}
         <VpcId>{{ target_group.vpc_id }}</VpcId>
         {% endif %}
-        <HealthCheckProtocol>{{ target_group.health_check_protocol }}</HealthCheckProtocol>
+        <HealthCheckProtocol>{{ target_group.healthcheck_protocol }}</HealthCheckProtocol>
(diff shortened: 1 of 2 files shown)


bblommers (changes requested):
Hi @radam9, would you mind adding a check for this in our tests?
Looks like we already have a few tests where we verify the creation works, so we only need to add a single assertion to verify that the protocol is returned correctly:
https://github.com/spulec/moto/blob/6ae0aa52728311158c4556ac38caffc717af21dc/tests/test_elbv2/test_elbv2_target_groups.py#L80

radam9:
> Hi @radam9, would you mind adding a check for this in our tests? Looks like we already have a few tests where we verify the creation works, so we only need to add a single assertion to verify that the protocol is returned correctly:
> 
> https://github.com/spulec/moto/blob/6ae0aa52728311158c4556ac38caffc717af21dc/tests/test_elbv2/test_elbv2_target_groups.py#L80

Hey @bblommers,
I have added the test, but after committing I thought I would check to see if the test would fail if I revert my change, but that was not the case.
So in the tests `test_create_target_group_and_listeners()` and `test_modify_target_group()` for example, the `target_group` has the proper `HealthCheckProtocol` attribute even if I don't pass it as a parameter to the `create_target_group()` method.

bblommers:
@radam9 The `describe_target_groups`-function returns the value correctly, yes.
But the `create_target_group`-function does not return the value without your change.

radam9:
That explains why the check didn't make any difference :)
I moved the check to the proper location in the test.

bblommers (approved):
LGTM! Thanks for the fix @radam9

PR #5149, opened 2022-05-19
Set `pyparsing >= 3.0.7` requirement

dominusmi:
Currently, the moto requirement is `>= 3.0.0`, however, at this line:
https://github.com/spulec/moto/blob/5d84eec79e213c0f71cc3bd7bd94ef55e1d474e4/moto/glue/utils.py#L278
it uses the function `delimited_list` with kwarg `min=1`. 
This keyword only appears starting `3.0.7` in `pyparsing`, causing an exception if you have a lower `pyparsing` version installed.

Link to pyparsing file:
3.0.6
https://github.com/pyparsing/pyparsing/blob/pyparsing_3.0.6/pyparsing/helpers.py
3.0.7
https://github.com/pyparsing/pyparsing/blob/pyparsing_3.0.7/pyparsing/helpers.py

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -55,7 +55,7 @@ _dep_idna = "idna<4,>=2.5"
 _dep_cfn_lint = "cfn-lint>=0.4.0"
 _dep_sshpubkeys = "sshpubkeys>=3.1.0"
 _dep_openapi = "openapi-spec-validator>=0.2.8"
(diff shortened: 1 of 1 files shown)


bblommers (approved):
LGTM! Thanks for the PR @dominusmi

Issue #5151, opened 2022-05-20
No keypair validation on run instance


szopen321:
**Repro steps:**
1. Start moto server.
2. Prepare terraform file that will contain an instance with declared key_name that is not assigned to any existing aws_key_pair
3. Run terraform init; terraform apply

**Description:**
When attempting to create an instance with key_name that is not assigned to any existing aws_key_pair moto doesn't verify that and creates an instance.

**Observed results:**
It is possible to create an instance with nonexisting key_pair

**expected results:**
Informative error with status code 400 - matching true AWS behaviour when attempting to create Instance with key_name that is not assigned to any existing aws_key_pair

szopen321:
PR with proposed fix https://github.com/spulec/moto/pull/5152

PR #5152, opened 2022-05-20
Add keypair validation

szopen321:
proposed fix for https://github.com/spulec/moto/issues/5151

### DIFF
--- a/moto/ec2/models/instances.py
+++ b/moto/ec2/models/instances.py
@@ -550,6 +550,8 @@ class InstanceBackend(object):
     def add_instances(self, image_id, count, user_data, security_group_names, **kwargs):
         location_type = "availability-zone" if kwargs.get("placement") else "region"
         default_region = "us-east-1"
+        describe_key_pairs = self.describe_key_pairs
+        describe_key_pairs(key_names=[kwargs.get("key_name")])
         valid_instance_types = INSTANCE_TYPE_OFFERINGS[location_type]
         if "region_name" in kwargs and kwargs.get("placement"):
             valid_availability_zones = {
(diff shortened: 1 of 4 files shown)


szopen321:
This PR is a WIP. 
I'm creating it to show my aproach and possibly change it according to your suggestions.

tests/test_ec2/test_ec2_integration.py
```diff
@@ -11,13 +11,16 @@ def test_run_instance_with_encrypted_ebs():
     kms = boto3.client("kms", region_name="us-east-1")
     resp = kms.create_key(Description="my key", KeyUsage="ENCRYPT_DECRYPT")
     key_id = resp["KeyMetadata"]["Arn"]
+    ec2_client = boto3.client("ec2", region_name="us-east-1")
```
bblommers, line 14: This client is already initialized two lines down

moto/ec2/models/instances.py
```diff
@@ -550,6 +550,8 @@ def get_instance(self, instance_id):
     def add_instances(self, image_id, count, user_data, security_group_names, **kwargs):
         location_type = "availability-zone" if kwargs.get("placement") else "region"
         default_region = "us-east-1"
+        describe_key_pairs = self.describe_key_pairs
+        describe_key_pairs(key_names=[kwargs.get("key_name")])
```
bblommers, line 554:
```suggestion
        self.describe_key_pairs(key_names=[kwargs.get("key_name")])
```
Minor nitpick - I don't think reassigning the method does anything for readability

bblommers (changes requested):
Thanks for the PR @szopen321! The approach looks good, just left a few comments

szopen321:
I have put keypair validation behind the flag as it can cause problems potentially.

bblommers:
> I have put keypair validation behind the flag as it can cause problems potentially.

That actually makes a lot of sense @szopen321.

Now that this validation is optional, we wouldn't need to change everything - existing tests can continue to use an invalid keypair. Would you mind changing the PR to only include he logic and the one new test, that validates that it breaks with the validation enabled?
I realize that it's a bit of extra work - feel free to just open a new branch/PR with only the required changes if that's easier.

bblommers (approved):
Awesome - thanks @szopen321!

PR #5028, opened 2022-04-15
add failying test that shows bug

bsneider:
#5027

PR #5156, opened 2022-05-22
Referring to the environment variable MOTO_ACCOUNT_ID directly in get_account_id()

### LINKED ISSUE #5027: ACCOUNT_ID Override is not working
Moto version's tried master branch and 3.1.4

```python
@mock_sts
def test_sts_account_override(monkeypatch):
    dummy_account_id = "MEMEMEMEME"
    monkeypatch.setenv("MOTO_ACCOUNT_ID", dummy_account_id)
    os_account_id = os.environ.get("MOTO_ACCOUNT_ID")
    os_account_id.should.equal(dummy_account_id)
    client = boto3.client("sts", region_name="us-west-2")
    resp = client.get_caller_identity()
    resp["Account"].should.equal(dummy_account_id)
```
this fails. 

Expected behavior would be resp["Account"].should.equal("MEMEMEMEME")

This test is being run with pytest. Several other ways to patch env variables have been tried. As seen in the test though the env variable is properly set. That is know bc `os_account_id.should.equal(dummy_account_id)` passes. 

## Reporting Bugs

Please be aware of the following things when filing bug reports:

1. Avoid raising duplicate issues. *Please* use the GitHub issue search feature
   to check whether your bug report or feature request has been mentioned in
   the past.
2. When filing bug reports about exceptions or tracebacks, please include the
   *complete* traceback. Partial tracebacks, or just the exception text, are
   not helpful.
3. Make sure you provide a suitable amount of information to work with. This
   means you should provide:

   - Guidance on **how to reproduce the issue**. Ideally, this should be a
     *small* code sample that can be run immediately by the maintainers.
     Failing that, let us know what you're doing, how often it happens, what
     environment you're using, etc. Be thorough: it prevents us needing to ask
     further questions.
   - Tell us **what you expected to happen**. When we run your example code,
     what are we expecting to happen? What does "success" look like for your
     code?
   - Tell us **what actually happens**. It's not helpful for you to say "it
     doesn't work" or "it fails". Tell us *how* it fails: do you get an
     exception? A hang? How was the actual result different from your expected
     result?
   - Tell us **what version of Moto you're using**, and
     **how you installed it**. Tell us whether you're using standalone server
     mode or the Python mocks. If you are using the Python mocks, include the
     version of boto/boto3/botocore.

   If you do not provide all of these things, it will take us much longer to
   fix your problem.

sarjun2493:
Resolves #5027.

Test case taken from #5028

### DIFF
--- a/moto/core/models.py
+++ b/moto/core/models.py
@@ -26,15 +26,10 @@ from .utils import convert_flask_to_responses_response
 ACCOUNT_ID = os.environ.get("MOTO_ACCOUNT_ID", "123456789012")
 
 
-def _get_default_account_id():
-    return ACCOUNT_ID
-
-
-account_id_resolver = _get_default_account_id
-
-
 def get_account_id():
-    return account_id_resolver()
+    # Those referring to the global var ACCOUNT_ID directly can continue to do so, but this will support the cases when
+    # an environment variable is overridden after moto is loaded/imported
+    return os.environ.get("MOTO_ACCOUNT_ID", "123456789012")
 
 
 class BaseMockAWS:
--- a/tests/test_sts/test_sts.py
+++ b/tests/test_sts/test_sts.py
@@ -1,5 +1,6 @@
 from base64 import b64encode
 import json
+import os
 
 import boto3
 from botocore.client import ClientError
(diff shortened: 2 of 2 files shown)


bblommers:
Hi @sarjun2493, thanks for the PR!

A lot of people currently override `ACCOUNT_ID` directly, i.e. by setting `moto.core.ACCOUNT_ID = ".."`, so this change would break their workflows.
FWIW, we are planning to tackle this in a similar way, as part of a bigger overhaul in how Moto deals with multiple accounts. But considering this is a breaking change, it will only be part of our next major release.

PR #5170, opened 2022-05-25
More param defaults for update password policy 

pinzon:
This PR addresses localstack/localstack#6119. If HardExpiry is not set, Moto returns 'none' as value, which makes terraform crash when parsing the XML response.


bblommers (approved):
LGTM! Thanks @pinzon

PR #5173, opened 2022-05-26
add kafka broker-storage VolumeSize to application autoscaling dimensions

thrau:
This PR adds an application-autoscaling dimension for MSK. Currently `broker-storage`is the only resource dimension for Kafka: https://docs.aws.amazon.com/autoscaling/application/APIReference/API_ScalableTarget.html

This example request was taken from the docs here: https://docs.aws.amazon.com/autoscaling/application/userguide/services-that-can-integrate-msk.html

```sh
aws application-autoscaling register-scalable-target \
   --service-namespace kafka \
   --scalable-dimension kafka:broker-storage:VolumeSize \
   --resource-id arn:aws:kafka:us-east-1:123456789012:cluster/demo-cluster-1/6357e0b2-0e6a-4b86-a0b4-70df934c2e31-5 \
   --min-capacity 100 \
   --max-capacity 800
```

I know that moto does not have MSK support yet, so I cannot add a proper test for it. I tested it against LocalStack.

### DIFF
--- a/moto/applicationautoscaling/models.py
+++ b/moto/applicationautoscaling/models.py
@@ -11,6 +11,8 @@ import uuid
 @unique
 class ResourceTypeExceptionValueSet(Enum):
     RESOURCE_TYPE = "ResourceType"
+    # MSK currently only has the "broker-storage" resource type which is not part of the resource_id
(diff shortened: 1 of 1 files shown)


bblommers (approved):
LGTM!

PR #5174, opened 2022-05-26
add ec2 network-acl egress filter

thrau:
This PR adds the EC2 Network ACL filter for the `egress` flag. 

This is not in the [official documentation](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_DescribeNetworkAcls.html#API_DescribeNetworkAcls_RequestParameters
), for network acl filters, but AWS allows it (and it is sometimes queried by Terraform)
```
% aws ec2 describe-network-acls --filters Name=entry.egress,Values=false 
{
    "NetworkAcls": [
        {
            "Associations": [
                {
                    "NetworkAclAssociationId": "aclassoc-bc6a1111",
                    ...
```

### DIFF
--- a/tests/test_ec2/test_network_acls.py
+++ b/tests/test_ec2/test_network_acls.py
@@ -389,13 +389,13 @@ def test_describe_network_acls():
     resp4 = conn.describe_network_acls(
         Filters=[{"Name": "entry.egress", "Values": ["false"]}]
     )
-    resp4["NetworkAcls"].should.have.length_of(3)
+    [entry["NetworkAclId"] for entry in resp4["NetworkAcls"]].should.contain(
+        network_acl_id
+    )
     # the ACL with network_acl_id contains no entries with Egress=True
-    resp4["NetworkAcls"][2]["NetworkAclId"].should.equal(network_acl_id)
     resp4 = conn.describe_network_acls(
         Filters=[{"Name": "entry.egress", "Values": ["true"]}]
     )
-    resp4["NetworkAcls"].should.have.length_of(2)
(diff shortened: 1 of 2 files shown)


bblommers (changes requested):
Would you mind adding a test for this? We could just add a few lines to this test, where we verify the other filters as well: https://github.com/spulec/moto/blob/76a094b6fb4dc6a3a1f31f56017720cd7164be04/tests/test_ec2/test_network_acls.py#L314

thrau:
@bblommers i added the test, sorry for the sloppiness ;-)

tests/test_ec2/test_network_acls.py
```diff
@@ -385,6 +385,21 @@ def test_describe_network_acls():
     )
     resp4["NetworkAcls"].should.have.length_of(0)
 
+    # Ensure filtering by egress flag
+    resp4 = conn.describe_network_acls(
+        Filters=[{"Name": "entry.egress", "Values": ["false"]}]
+    )
+    resp4["NetworkAcls"].should.have.length_of(3)
```
bblommers, line 392: These tests are executed in parallel in ServerMode, so Moto is also returning NetworkACL's created in tests. That's why the tests are failing, it's returning much more than 3 ACL's.

The best way to deal with this IMO, is to only assert whether the response (does not) contains a specific ACL - the specific length is irrelevant.

thrau: thanks for the clarification! updated the asserts

bblommers:
> @bblommers i added the test, sorry for the sloppiness ;-)

All good - I was tempted to just let it go :) Thanks for the PR!

bblommers (approved):

PR #5176, opened 2022-05-27
Pin versions of werkzeug

bchen1116:
`Werkzeug<0.5` fails installation on Moto with py38. Pinning to `>=0.5` allows the install to successfully happen.

(installing `werkzeug==0.4` after `moto`)
image

(installing `werkzeug==0.5` after `moto`)
image


bblommers (approved):
Makes sense @bchen1116 - thanks for the PR!

PR #5181, opened 2022-05-31
key_name encoding for ListBucketResult

pinzon:
This PR addresses issue localstack/localstack#6098, where is shown that ListBucketResult should url encode the names of the keys. Also for `list_objects_v2` the EncodingType attribute must be readed and returned in the xml response (v1 does the same but doesn't affect boto3 behavior).

Changes:
- key_name is url encoded for ListBucketResult and ListBucketResultV2.
- EncodingType is read from the query string and returned accordingly.
- Test that validates the correct handling of key_names with weird characters.

### DIFF
--- a/moto/s3/models.py
+++ b/moto/s3/models.py
@@ -14,6 +14,7 @@ import pytz
 import sys
 import time
 import uuid
+import urllib.parse
 
 from bisect import insort
 from importlib import reload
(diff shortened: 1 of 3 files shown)


pinzon:
@tsiq-rob thanks for the comment, I'm going to make a new PR that makes the Key name being encoded only if the `encoding-type=url`

PR #5182, opened 2022-06-01
Bug fix for selecting KMS key via alias ARN

BlueDragon23:
When using Localstack with moto as the KMS provider, I encountered issues with selecting a key via the alias ARN. I believe this change fixes the issue by ensuring the `alias/` prefix is included when the name is extracted from the ARN. I've added tests to the KMS model to verify that, and that encryption works via all the other supported mechanisms

### DIFF
--- a/tests/test_kms/test_model.py
+++ b/tests/test_kms/test_model.py
@@ -1,70 +1,52 @@
+import pytest
 import sure  # noqa # pylint: disable=unused-import
 
 from moto.kms.models import KmsBackend
 
+PLAINTEXT = b"text"
+REGION = "us-east-1"
 
-def test_encrypt_key_id():
-    region = "us-east-1"
-    backend = KmsBackend(region)
-    # Create key
-    key = backend.create_key(
-        None, "ENCRYPT_DECRYPT", "SYMMETRIC_DEFAULT", "Test key", None, region
+
+@pytest.fixture
+def backend():
+    return KmsBackend(REGION)
+
+
+@pytest.fixture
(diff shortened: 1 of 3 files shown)


tests/test_kms/test_model.py
```diff
@@ -0,0 +1,52 @@
...
+    return backend.create_key(
+        None, "ENCRYPT_DECRYPT", "SYMMETRIC_DEFAULT", "Test key", None, REGION
+    )
+
+
+def test_encrypt_key_id(backend, key):
+    ciphertext, arn = backend.encrypt(key.id, PLAINTEXT, {})
+
+    assert ciphertext is not None
+    assert arn is not None
```
bblommers, line 25: These assertions should be more specific. What is the format of the ARN that we're expecting?

(Same goes for all other tests)

bblommers (changes requested):
Hi @BlueDragon23, could you add another test using the boto3 SDK, to show what scenario is fixed with this change?

There are quite a few tests already in `tests/test_kms/test_kms_boto3`, so feel free to reuse or extend one of them.

BlueDragon23:
Thanks for the feedback! I've updated the model tests to be more specific. It turns out there _was_ an existing test in `test_kms_boto3` that covered this use case, but it was always using the `key_id` instead of the `target_id`. I've changed that now, so it verifies the case of generating a data key using the alias ARN

bblommers (approved):
LGTM - thanks for the PR @BlueDragon23!

PR #5184, opened 2022-06-01
s3 key_name encoding should be variable

pinzon:
This PR corrects the last problem commented in #5181 by @tsiq-rob.

### DIFF
--- a/moto/s3/models.py
+++ b/moto/s3/models.py
@@ -154,8 +154,8 @@ class FakeKey(BaseModel):
 
         self.s3_backend = s3_backend
 
-    def safe_name(self, encoding_type = None):
-        if encoding_type == 'url':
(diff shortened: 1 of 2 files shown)


bblommers (approved):

bblommers:
Does this release fix the problem for you, @tslq-rob?

tsiq-rob:
> Does this release fix the problem for you, @tslq-rob?

Thanks for the fix. I will investigate soon and let you know whether it resolves the issue we saw.

tsiq-rob:
Yes, the issue seems to be resolved. Thanks again!

bblommers:
Thanks for confirming @tsiq-rob, and thanks again for the fixing it @pinzon!

PR #5194, opened 2022-06-07
Stable cognito user pool

c0va23:
Hi!

We create a Cogito UserPool before running integration tests, and random IDs complicate the process of finding a pool.

I added an environment variable `MOTO_COGNITO_IDP_USER_POOL_ID_STRATEGY=HASH`, setting which makes the ID of Cognito Userpool reproducible. I think this change will be useful to other moto users as well.

### DIFF
--- a/docs/docs/services/cognito-idp.rst
+++ b/docs/docs/services/cognito-idp.rst
@@ -145,3 +145,10 @@ cognito-idp
 
 - [ ] verify_user_attribute
 
+
+|start-h3| Stable Cognito User Pool Id |end-h3|
+
+In some cases, you need to have reproducible IDs for the user pool.
+For example, a single initialization before the start of integration tests.
+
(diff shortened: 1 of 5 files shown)


bblommers (changes requested):
Hi @c0va23, thanks for the PR! This makes sense - I've just got some comments on the approach.

Would you mind adding a new test for this as well? Existing tests for `create_user_pool` can be found here: 
https://github.com/spulec/moto/blob/master/tests/test_cognitoidp/test_cognitoidp.py

See this test for an example on how to set environment variables for a single test:
https://github.com/spulec/moto/blob/master/tests/test_core/test_mock_regions.py#L27

Note the `if TEST_SERVER_MODE: skip` construct there - that is always necessary for tests that modify environment variables, as we cannot control them in ServerMode.

c0va23:
@bblommers, I will correct all comments and suggestions. I will try to do it within the next couple of weeks.

c0va23:
@bblommers , i fix all comments and add tests.

bblommers (approved):
Awesome - thanks again @c0va23!

PR #5197, opened 2022-06-07
fixed route53 rrset terraform suit

macnev2013:
## Fixed TF Test
- TestAccRoute53Record_basiac
- TestAccRoute53Record_underscored
- TestAccRoute53Record_disappears
- TestAccRoute53Record_Disappears_multipleRecords
- TestAccRoute53Record_Basic_fqdn
- TestAccRoute53Record_Basic_trailingPeriodAndZoneID
- TestAccRoute53Record_txtSupport
- TestAccRoute53Record_spfSupport
- TestAccRoute53Record_caaSupport
- TestAccRoute53Record_dsSupport
- TestAccRoute53Record_generatesSuffix
- TestAccRoute53Record_wildcard
- TestAccRoute53Record_failover
- TestAccRoute53Record_Weighted_basic
- TestAccRoute53Record_WeightedToSimple_basic
- TestAccRoute53Record_Alias_s3
- TestAccRoute53Record_Geolocation_basic
- TestAccRoute53Record_HealthCheckID_setIdentifierChange
- TestAccRoute53Record_HealthCheckID_typeChange
- TestAccRoute53Record_Latency_basic
- TestAccRoute53Record_typeChange
- TestAccRoute53Record_nameChange
- TestAccRoute53Record_setIdentifierChange
- TestAccRoute53Record_empty
- TestAccRoute53Record_longTXTrecord
- TestAccRoute53Record_doNotAllowOverwrite
- TestAccRoute53Record_allowOverwrite

## Fixes
- Added static getdnssec response
- updated error from `single_error` to `wrapped_error`
- added exception `HostedZoneNotEmpty` when trying to delete hosted zone with record
- updated response templates to be more inclined with AWS

### DIFF
--- a/moto/route53/responses.py
+++ b/moto/route53/responses.py
@@ -115,6 +115,20 @@ class Route53(BaseResponse):
             route53_backend.delete_hosted_zone(zoneid)
             return 200, headers, DELETE_HOSTED_ZONE_RESPONSE
 
+    def get_dnssec_response(self, request, full_url, headers):
+        # returns static response
+        # TODO: implement enable/disable dnssec apis
+        self.setup_class(request, full_url, headers)
+
+        parsed_url = urlparse(full_url)
+        method = request.method
+
+        zoneid = parsed_url.path.rstrip("/").rsplit("/", 2)[1]
+
+        if method == "GET":
+            route53_backend.get_dnssec(zoneid)
+            return 200, headers, GET_DNSSEC
+
     def rrset_response(self, request, full_url, headers):
         self.setup_class(request, full_url, headers)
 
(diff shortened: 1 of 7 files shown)


bblommers (changes requested):
Hi @macnev2013, would you mind adding the new tests to the list?
https://github.com/spulec/moto/blob/master/tests/terraformtests/terraform-tests.success.txt

Note that we do not support a `failure`-list anymore, so you may have to list all these tests individually.

moto/route53/responses.py
```diff
@@ -526,17 +546,26 @@ def reusable_delegation_set(self, request, full_url, headers):
...
+      {% endif %}
    </HostedZone>
+   {% if not zone.private_zone %}
    <DelegationSet>
       <Id>{{ zone.delegation_set.id }}</Id>
       <NameServers>
          {% for name in zone.delegation_set.name_servers %}<NameServer>{{ name }}</NameServer>{% endfor %}
       </NameServers>
    </DelegationSet>
-   <VPC>
```
bblommers, line 536: As far as I can tell, the VPC-config was in the correct location here. The docs say it's supposed to be at the top-level, not inside the `<HostedZone>`.
https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/route53.html#Route53.Client.create_hosted_zone

Changing that might fix a few unit tests, but it would be good to get the rest passing as well.

macnev2013: Yep, my bad. Thanks for pointing out.

macnev2013:
Hey @bblommers 

When using `conn.delete_health_check(HealthCheckId=check["Id"])`
- Moto in server mode is calling `get_or_delete_health_check_response` function (correct one)
- Moto in unit test mode is calling `health_check_response` function (which is incorrect)

Why is it behaving differently on two modes?

bblommers (approved):
LGTM! Thanks @macnev2013

PR #5198, opened 2022-06-08
Greengrass Implement create_device_definition 

cm-iwata:
Implement these APIs

- create_device_definition
- create_device_definition_version


bblommers (approved):
LGTM! Thanks @cm-iwata

PR #5209, opened 2022-06-09
create_vpc should never create a default vpc

jbmchuck:
The create_vpc call never creates a default VPC in AWS - there is a
separate create_default_vpc call. This removes a behavior by which
previously moto would create a default vpc if no other vpcs exist.


bblommers:
Thanks for this @jbmchuck! This should only be done after we've added the `create_default_vpc`-function is in place - we don't want to be in a situation where it's impossible to create a default VPC.
I should have something ready for this by tomorrow.

Issue #5203, opened 2022-06-08
Ability to tag and remove a tag for an EKS resource


mmerrill3:
This is a feature request to leverage moto for mocking tag functions in the EKS client, including
tag_resource()
untag_resource()
list_tags_for_resource()

According to the latest published dev docs, this is not implemented yet at http://docs.getmoto.org/en/latest/docs/services/eks.html

bblommers:
Thanks for raising this @mmerrill3 - will mark it as an enhancement to add these methods.

If this is something you'd like to tackle yourself, PR's are always welcome!

mmerrill3:
Yeap, I was going to give it a whirl.  You can assign to me.

mmerrill3:
@bblommers could your approve the workflow since its my first contribution to this project?

PR #5215, opened 2022-06-10
feature: tag functions in eks(#5203)

mmerrill3:
Implementation for EKS functions tag_resource, untag_resource, and list_tags_for_resources

Implements #5203 

Signed-off-by: mmerrill3 <jjpaacks@gmail.com>

### DIFF
--- a/moto/eks/models.py
+++ b/moto/eks/models.py
@@ -631,6 +631,78 @@ class EKSBackend(BaseBackend):
         cluster.nodegroup_count -= 1
         return result
 
+    def tag_resource(self, resource_arn, tags):
+        """
+        This function currently will tag an EKS cluster only.  It does not tag a managed node group
+        """
+
+        try:
+            cluster = next(
+                self.clusters[x]
+                for x in self.clusters
+                if self.clusters[x].arn == resource_arn
+            )
+        except StopIteration:
+            # Cluster does not exist.
+            raise ResourceNotFoundException(
+                clusterName=None,
+                nodegroupName=None,
+                fargateProfileName=None,
+                addonName=None,
+                message="An error occurred (NotFoundException) when calling the TagResource operation: Resource was not found",
+            )
+        cluster.tags.update(tags)
+        return ""
+
+    def untag_resource(self, resource_arn, tag_keys):
+        """
+        This function currently will remove tags on an EKS cluster only.  It does not remove tags from a managed node group
+        """
+        if not isinstance(tag_keys, list):
+            tag_keys = [tag_keys]
+
+        try:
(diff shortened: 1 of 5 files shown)


moto/eks/models.py
```diff
@@ -631,6 +631,57 @@ def delete_nodegroup(self, cluster_name, nodegroup_name):
         cluster.nodegroup_count -= 1
         return result
 
+    def tag_resource(self, resource_arn, tags):
+        try:
```
bblommers, line 635: Can we have a docstring here to say that we only support tagging Clusters, and not managed node groups?

moto/eks/models.py
```diff
@@ -631,6 +631,57 @@ def delete_nodegroup(self, cluster_name, nodegroup_name):
...
+                message="An error occurred (NotFoundException) when calling the UntagResource operation: Resource was not found",
+            )
+        for name in tag_keys:
+            if name in cluster.tags:
+                del cluster.tags[name]
+        return ""
+
+    def list_tags_for_resource(self, resource_arn):
+        try:
+            for x in self.clusters:
```
bblommers, line 671: Print-statements should not be in Moto

bblommers (changes requested):
Looks like the linter is failing on a few things as well:
https://github.com/spulec/moto/runs/6838049781?check_suite_focus=true

moto/eks/models.py
mmerrill3: yes, will update, thanks

moto/eks/models.py
mmerrill3: I missed this, will update, thanks

mmerrill3:
@bblommers forgot to install linter on VSC on new mac.  Fixed that.

mmerrill3:
@bblommers ready for review again, thank you for your feedback.

bblommers (changes requested):
Thanks for making the changes @mmerrill3. Looks like it's just the ServerMode tests that are failing now. Looking at the code, I can't immediately see why though - let me know if you want me to do some debugging on this.

mmerrill3:
Hi @bblommers, the server tests are running in the "docker" mode, using the latest tag of the motoserver docker image.  That doesn't have the new URLs from eks/urls.py for tagging.  What normally is the process to update that docker image?  This looks like a chicken/egg problem.

mmerrill3:
@bblommers I updated the urls to point to a specific tags function in the EKSResponse object, like other services have done (mq is one).

bblommers (approved):
Awesome - thanks again for all your work on this @mmerrill3!

PR #5217, opened 2022-06-11
Fix InvalidObjectState HTTP status code

luisgabrielroldan:
Hello!

This PR fixes the HTTP status code for the InvalidObjectState error.

The expected HTTP code ([AWS docs](https://docs.aws.amazon.com/AmazonS3/latest/API/ErrorResponses.html)) is 403.


bblommers (approved):
Thanks for the fix @luisgabrielroldan!

PR #5221, opened 2022-06-13
Greengrass Implement create_resource_definition

cm-iwata:
Implement these APIs

- create_resource_definition
- create_resource_definition_version

### DIFF
--- a/moto/greengrass/models.py
+++ b/moto/greengrass/models.py
@@ -117,6 +117,55 @@ class FakeDeviceDefinitionVersion(BaseModel):
         return obj
 
 
+class FakeResourceDefinition(BaseModel):
+    def __init__(self, region_name, name, initial_version):
+        self.region_name = region_name
+        self.id = str(uuid.uuid4())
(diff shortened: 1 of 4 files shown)


cm-iwata:
I don't know why SQS test fails

```
FAILED tests/test_sqs/test_sqs.py::test_change_message_visibility_on_visible_message
```

I guess VisibilityTimeout may too short for this testcase,  and sometimes testcase will failue.


https://github.com/spulec/moto/blob/9640ec20d125248ac91243591c7db50daabfd135/tests/test_sqs/test_sqs.py#L1638



I tried to extend VisibilityTimeout and the test ended normally.
https://github.com/cm-iwata/moto/runs/6872692585?check_suite_focus=true

@bblommers 
Can I add a commit that extends the Visibility Timeout as a trial?

bblommers (approved):
This looks good to me - I'll just merge on the assumption that the SQS tests will now pass. Thanks again for the PR @cm-iwata!

PR #5223, opened 2022-06-13
fixed tf suite: Route53HostedZone

macnev2013:
## Added
**APIs**
- associate_vpc
- disassociate_vpc

## Updated
- Mechanism to store VPC in FakeZone

## TF Tests
- TestAccRoute53Zone_basic
- TestAccRoute53Zone_disappears
- TestAccRoute53Zone_multiple
- TestAccRoute53Zone_comment
- TestAccRoute53Zone_delegationSetID
- TestAccRoute53Zone_forceDestroy
- TestAccRoute53Zone_ForceDestroy_trailingPeriod
- TestAccRoute53Zone_tags
- TestAccRoute53Zone_VPC_single
- TestAccRoute53Zone_VPC_multiple
- TestAccRoute53Zone_VPC_updates

Note:
- The test `test_hosted_zone_private_zone_preserved` was expecting to create a vpc without having to add VPC info. In `create_hosted_zone` response VPCs section was having one entry with empty string so it was expecting len(vpc) == 1. (Which was changed to contain empty vpc list::: len(vpc) == 0)
- Future Work: Need to make sure we can create a vpc with params as : `private_zone = True` and without `VPC` config

### DIFF
--- a/moto/route53/exceptions.py
+++ b/moto/route53/exceptions.py
@@ -89,6 +89,28 @@ class HostedZoneNotEmpty(Route53ClientError):
         self.content_type = "text/xml"
 
 
+class PublicZoneVPCAssociation(Route53ClientError):
+    """Public hosted zone can't be associated."""
+
+    code = 400
+
+    def __init__(self):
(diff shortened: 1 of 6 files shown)


bblommers (approved):
LGTM! Thanks @macnev2013

PR #5238, opened 2022-06-16
Update get_secret_value to properly check versions and stages match

bsch150:
Howdy folks, just started using moto, it's been a great help with testing some lambdas I've got. 

I noticed that the get_secret_value operation has a minor issue where we weren't checking that the version_stage and version_id match each other. the AWS cli and boto3 will throw an error in real life.

### DIFF
--- a/moto/secretsmanager/exceptions.py
+++ b/moto/secretsmanager/exceptions.py
@@ -30,6 +30,15 @@ class SecretHasNoValueException(SecretsManagerClientError):
         )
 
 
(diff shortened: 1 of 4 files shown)


bsch150:
Some of these test cases appear to be inaccurate. [Here is a gist](https://gist.github.com/bsch150/340ba9e0bef66f3b38b34fa02a89ce86) with steps to provide evidence using the AWS cli. I'm working on updating the tests!

bblommers (approved):
Awesome - thanks for the PR @bsch150!

PR #5241, opened 2022-06-17
fixed tf tests: TestAccRoute53HealthCheck

macnev2013:
## TF Tests

- TestAccRoute53HealthCheck_basic
- TestAccRoute53HealthCheck_tags
- TestAccRoute53HealthCheck_withSearchString
- TestAccRoute53HealthCheck_withChildHealthChecks
- TestAccRoute53HealthCheck_withHealthCheckRegions
- TestAccRoute53HealthCheck_ip
- TestAccRoute53HealthCheck_ipv6
- TestAccRoute53HealthCheck_cloudWatchAlarmCheck
- TestAccRoute53HealthCheck_withSNI
- TestAccRoute53HealthCheck_disabled
- TestAccRoute53HealthCheck_withRoutingControlARN
- TestAccRoute53HealthCheck_disappears

## Fixed
- CreateHealthCheck

## Added
- UpdateHealthCheck


moto/route53/models.py
```diff
@@ -64,10 +64,26 @@ def __init__(self, health_check_id, caller_reference, health_check_args):
         self.measure_latency = health_check_args.get("measure_latency") or False
         self.inverted = health_check_args.get("inverted") or False
         self.disabled = health_check_args.get("disabled") or False
-        self.enable_sni = health_check_args.get("enable_sni") or False
-        self.children = health_check_args.get("children") or None
+        self.enable_sni = health_check_args.get("enable_sni") or True
         self.caller_reference = caller_reference
 
+    def add_children(self, children):
```
bblommers, line 70: The name `set_children` would be more accurate, as it doesn't add anything to the existing list of children

moto/route53/models.py
```diff
@@ -64,10 +64,26 @@ def __init__(self, health_check_id, caller_reference, health_check_args):
...
+        if children and isinstance(children, list):
+            self.children = children
+        elif children and isinstance(children, str):
+            self.children = [children]
+        else:
+            self.children = None
+        return self.children
```
bblommers, line 77: The return value is not used anywhere, as far as I can tell, so we can remove this

moto/route53/models.py
```diff
@@ -64,10 +64,26 @@ def __init__(self, health_check_id, caller_reference, health_check_args):
...
+        self.enable_sni = health_check_args.get("enable_sni") or True
         self.caller_reference = caller_reference
 
+    def add_children(self, children):
+        if children and isinstance(children, list):
+            self.children = children
+        elif children and isinstance(children, str):
+            self.children = [children]
+        else:
+            self.children = None
```
bblommers, line 76: Is this desired behaviour?
When calling `update_healthcheck` without the `children`-argument, this removes any existing children. It seems more intuitive to keep the existing children, if the user doesn't explicitly say they want to change/update them.

moto/route53/models.py
```diff
@@ -64,10 +64,26 @@ def __init__(self, health_check_id, caller_reference, health_check_args):
...
+        return self.children
+
+    def add_regions(self, regions):
```
bblommers, line 79: Same for the name

moto/route53/models.py
```diff
@@ -64,10 +64,26 @@ def __init__(self, health_check_id, caller_reference, health_check_args):
...
+        if regions and isinstance(regions, list):
+            self.regions = regions
+        elif regions and isinstance(regions, str):
+            self.regions = [regions]
+        else:
+            self.regions = None
```
bblommers, line 85: And same question about the behaviour

bblommers (changes requested):

moto/route53/models.py
macnev2013: Agreed. Updated

moto/route53/models.py
macnev2013: Yep. Missed it. Thanks for pointing out.

moto/route53/models.py
macnev2013: Correct. if calling `update_healthcheck` without `--child-health-checks` argument - its not removing any children. Fixed this.

But it will overwrite all children if `--child-health-checks` argument is present with valid children.

moto/route53/models.py
macnev2013: updated

moto/route53/models.py
macnev2013: Updated

bblommers (approved):
Thanks @macnev2013!

PR #5243, opened 2022-06-20
Add KMS Sign/Verify

josecoelho:
This is adding support for KMS [sign](https://docs.aws.amazon.com/kms/latest/APIReference/API_Sign.html) and [verify](https://docs.aws.amazon.com/kms/latest/APIReference/API_Sign.html).

I added support for `KeySpec`  keeping support for the deprecated deprecated `CustomerMasterKeySpec`. I can move that change to another PR if it makes things easier.

This is not the full implementation for sign and verify, as the algorithm and private key type are hardcoded. But it's enough to use `sign` and `verify` within the application. Would that be enough for now? It is good enough for our use case.

`GrantTokens` and `MessageType` `DIGEST` are not supported as well.

### DIFF
--- a/moto/kms/models.py
+++ b/moto/kms/models.py
@@ -2,6 +2,11 @@ import json
 import os
 from collections import defaultdict
 from datetime import datetime, timedelta
+from cryptography.exceptions import InvalidSignature
+from cryptography.hazmat.primitives import hashes
+
+from cryptography.hazmat.primitives.asymmetric import padding
+from moto.apigateway.exceptions import ValidationException
 
 from moto.core import get_account_id, BaseBackend, BaseModel, CloudFormationModel
 from moto.core.utils import get_random_hex, unix_time, BackendDict
@@ -14,6 +19,7 @@ from .utils import (
     encrypt,
     generate_key_id,
     generate_master_key,
+    generate_private_key,
 )
 
 
@@ -64,6 +70,7 @@ class Key(CloudFormationModel):
         self.key_rotation_status = False
         self.deletion_date = None
         self.key_material = generate_master_key()
+        self.private_key = generate_private_key()
         self.origin = "AWS_KMS"
         self.key_manager = "CUSTOMER"
         self.customer_master_key_spec = customer_master_key_spec or "SYMMETRIC_DEFAULT"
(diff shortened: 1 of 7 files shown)


tests/test_kms/test_kms_boto3.py
```diff
@@ -1106,3 +1117,101 @@ def test_key_tag_added_arn_based_happy():
...
+def test_sign_invalid_signing_algorithm():
+    client = boto3.client("kms", region_name="us-west-2")
+
+    key = client.create_key(Description="sign-key", KeyUsage="SIGN_VERIFY")
+    key_id = key["KeyMetadata"]["KeyId"]
+
+    message = "My message"
+    signing_algorithm = "INVALID"
+
+    try:
```
bblommers, line 1151: Can we use the `pytest.raises` here instead - that makes it easy to verify that the correct error message is raised, not just that it fails.

See http://docs.getmoto.org/en/latest/docs/contributing/development_tips/tests.html#negative-tests

tests/test_kms/test_kms_boto3.py
```diff
@@ -1106,3 +1117,101 @@ def test_key_tag_added_arn_based_happy():
...
+def test_sign_invalid_key_usage():
+    client = boto3.client("kms", region_name="us-west-2")
+
+    key = client.create_key(Description="sign-key", KeyUsage="ENCRYPT_DECRYPT")
+    key_id = key["KeyMetadata"]["KeyId"]
+
+    message = "My message"
+    signing_algorithm = "RSASSA_PSS_SHA_256"
+
+    try:
```
bblommers, line 1168: Same

bblommers (changes requested):
Hi @josecoelho, thanks for the PR! Any implementation is welcome, it doesn't have to support everything at the same time.

Please see my comments - it looks like some other tests are failing as well due to the removal of `CustomerMasterKeySpec`.

moto/kms/models.py
```diff
@@ -516,5 +520,94 @@ def retire_grant(self, key_id, grant_id, grant_token) -> None:
...
+
+        if message_type == "DIGEST":
+            raise NotImplementedError(
+                "MessageType DIGEST not supported by kms_verify mock yet"
+            )
+
+    def sign(self, key_id, message, message_type, grant_tokens, signing_algorithm):
+        key = self.describe_key(key_id)
+
+        self.__ensure_implemented_sign_and_verify_args(grant_tokens, message_type)
```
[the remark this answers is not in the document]
josecoelho: I moved to warnings [here](https://github.com/spulec/moto/pull/5243/files#diff-8bf627e2436b027528752b0b1818a087eb16cf58b490f3e9adb26a2de6609aa8R670)

moto/kms/utils.py
```diff
@@ -58,6 +59,18 @@ def generate_master_key():
     return generate_data_key(MASTER_KEY_LEN)
 
 
+def generate_private_key():
+    """Generate a private key to be used on asymmetric sign/verify.
+
+    NOTE: KeySpec is not taken into consideration and the key is always RSA_2048
```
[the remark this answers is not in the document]
josecoelho: Thanks for the advice, I added more info to [`sign`](https://github.com/spulec/moto/pull/5243/files#diff-0514df96f93d76619859abfba375a4eb8ae5113221b9feb9c57ddc648255c59eR545) and [`verify`](https://github.com/spulec/moto/pull/5243/files#diff-0514df96f93d76619859abfba375a4eb8ae5113221b9feb9c57ddc648255c59eR572)

tests/test_kms/test_kms_boto3.py
josecoelho: Fixed

tests/test_kms/test_kms_boto3.py
josecoelho: Fixed

tests/test_kms/test_kms_boto3.py
```diff
@@ -1106,3 +1117,101 @@ def test_key_tag_added_arn_based_happy():
...
+def test_sign_invalid_message():
+    client = boto3.client("kms", region_name="us-west-2")
+
+    key = client.create_key(Description="sign-key", KeyUsage="SIGN_VERIFY")
+    key_id = key["KeyMetadata"]["KeyId"]
+
+    message = ""
+    signing_algorithm = "RSASSA_PSS_SHA_256"
+
+    try:
```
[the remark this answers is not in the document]
josecoelho: Fixed

josecoelho:
@bblommers I still have some issues with [one of the tests](https://github.com/josecoelho/moto/blame/feature/kms-sign-verify/tests/test_kms/test_kms_boto3.py#L915).

It is raising `ExpiredTokenException` but I'm having a hard time to identify the reason. Would it be possible to help me out?
```python
botocore.exceptions.ClientError: An error occurred (ExpiredTokenException) when calling the CreateKey operation: The security token included in the request is expired
```

josecoelho:
@bblommers fixed the failing test. I missed `@mock_kms` before [a test](https://github.com/josecoelho/moto/blob/ca862a2e31d0e7d56e90e59e9caf4f3cba17a28b/tests/test_kms/test_kms_boto3.py#L122).

moto/kms/models.py
```diff
@@ -516,5 +520,93 @@ def retire_grant(self, key_id, grant_id, grant_token) -> None:
             key = self.describe_key(key_id)
             key.retire_grant(grant_id)
 
+    def __ensure_valid_sing_and_verify_key(self, key: Key):
```
lxop, line 523:
```suggestion
    def __ensure_valid_sign_and_verify_key(self, key: Key):
```

bblommers (approved):
LGTM - thanks again @josecoelho!

PR #5247, opened 2022-06-20
Deletion of endpoints after platform app

pinzon:
While making some changes for localstack/localstack#6256 in came to light that AWS deletes the endpoints of a deleted platform application.

Changes:
- deletion of endpoints when deleting the platform app.
- test  that validates that the endpoint doesn't exist after it's app was deleted.

Note: in AWS it takes like 5 seconds to delete the endpoint after the deletion of the app.


bblommers (approved):
Thanks @pinzon!

PR #5253, opened 2022-06-22
EC2: create_customer_gateway - support PublicIp and IpAddress

bblommers:
The EC2 API was changed on 21st of June - see https://github.com/boto/botocore/commit/86202c8698cf77fb6ecccfdbc05bbc218e861d14#diff-c51449716bfc26c1eac92ec403b470827d2dcba1126cf303567074b872d5c592.


bnavigator:
This means that the test suite fails with  botocore < 1.27.14

PR #5261, opened 2022-06-24
Fix: S3 Key safe_name should avoid encoding "/"

pinzon:
This is another fix for another issue that came from encoding the key's name. Mentioned in localstack/localstack#6323


bblommers (approved):
LGTM :+1:

PR #5281, opened 2022-07-01
Greengrass Implement Deployments APIs

cm-iwata:
mplement these APIs

- create_deployment
- list_deployments
- get_deployment_status
- reset_deployments

### DIFF
--- a/tests/test_greengrass/test_greengrass_deployment.py
+++ b/tests/test_greengrass/test_greengrass_deployment.py
@@ -259,3 +259,99 @@ def test_list_deployments():
 
     if not TEST_SERVER_MODE:
         deployments.should.have.key("CreatedAt").equal("2022-06-01T12:00:00.000Z")
+
+
+@freezegun.freeze_time("2022-06-01 12:00:00")
+@mock_greengrass
+def test_get_deployment_status():
+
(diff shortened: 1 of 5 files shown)


tests/test_greengrass/test_greengrass_deployment.py
```diff
@@ -0,0 +1,496 @@
...
+        client.get_deployment_status(
+            GroupId="7b0bdeae-54c7-47cf-9f93-561e672efd9c", DeploymentId=deployment_id
+        )
+    ex.value.response["Error"]["Message"].should.equal(
+        f"Deployment '{deployment_id}' does not exist."
+    )
+    ex.value.response["Error"]["Code"].should.equal("InvalidInputException")
+
+
+@pytest.mark.skipif(
```
cm-iwata, line 360: ResetDeployments API contains $reset in path.
https://docs.aws.amazon.com/greengrass/v1/apireference/-greengrass-groups-groupid-deployments-reset.html

And in servermode we can't  use path that contains $.
Because $ will replace to empty in here.
https://github.com/spulec/moto/blob/6282e5124d726f806a54fdad9645fc236f9baf90/moto/core/utils.py#L96

bblommers (approved):
LGTM! Thanks @cm-iwata

PR #5282, opened 2022-07-01
Add support for codebuild

b0bu:
Adding base support for codebuild and for the following methods. I've only added the methods that I have a need for right now but if there's nothing wrong with the approach I'll likely add more over time. There's also some defaults around environment and source that could do with some further validation / general checking but I don't have any need for that right now. 

- batch_get_builds
- delete_project
- list_builds
- list_builds_for_project
- start_build
- stop_build


moto/codebuild/models.py
```diff
@@ -0,0 +1,265 @@
...
+        self.project_metadata["artifacts"] = artifacts
+        self.project_metadata["source"] = project_source
+        self.project_metadata["cache"] = dict()
+        self.project_metadata["cache"]["type"] = "NO_CACHE"
+        self.project_metadata["timeoutInMinutes"] = ""
+        self.project_metadata["queuedTimeoutInMinutes"] = ""
+
+
+class CodeBuildBackend(BaseBackend):
+    """find which functions I used in the dev branch stuff"""
```
bblommers, line 126: Just as an FYI: Docstrings like this will end up on `docs.getmoto.org` as an explanation, so they can be useful if you want to explain what is/isn't supported. This comment looks more like a reminder to yourself though :)

moto/codebuild/responses.py
```diff
@@ -0,0 +1,222 @@
...
+    ResourceNotFoundException,
+)
+from moto.core import get_account_id
+import json
+import re
+
+
+def _validate_required_params_source(source):
+    try:
+        assert source["type"] in [
```
bblommers, line 15: Let's use regular checks here, to keep in line with the code style everywhere else:
```
if x not in y:
    raise Exception(..)
```
There's a few others, I won't point hem all out, but they should all follow the same pattern

tests/test_codebuild/test_codebuild.py
```diff
@@ -0,0 +1,704 @@
...
+        environment=environment,
+        serviceRole=service_role,
+    )
+    client.start_build(projectName="project-2")
+
+    response = client.list_builds()
+    response["ids"].should_not.be.empty
+
+    for build_id in response["ids"]:
+        try:
```
bblommers, line 582: Similar comment here - to keep the same style, let's use `build_id.should.be.within([..])`

tests/test_codebuild/test_codebuild.py
```diff
@@ -0,0 +1,704 @@
...
+    err.value.response["Error"]["Code"].should.equal("InvalidInputException")
+
+
+@mock_codebuild
+def test_codebuild_batch_get_builds_empty_build_id():
+    client = boto3.client("codebuild", region_name="eu-central-1")
+
+    with pytest.raises(ParamValidationError) as err:
+        client.batch_get_builds(ids=[])
+        err.response["Error"]["Code"].should.equal("ParamValidationError")
```
bblommers, line 610:
```suggestion
    err.response["Error"]["Code"].should.equal("ParamValidationError")
```
Wrong indentation - this line doesn't actually get executed if `client.batch_get_builds` throws an error

tests/test_codebuild/test_codebuild.py
```diff
@@ -0,0 +1,704 @@
...
+    client.start_build(projectName=name)
+
+    response = client.list_builds_for_project(projectName=name)
+    response["ids"].should_not.be.empty
+
+    client.delete_project(name=name)
+
+    with pytest.raises(ClientError) as err:
+        client.list_builds_for_project(projectName=name)
+        err.response["Error"]["Code"].should.equal("ResourceNotFoundException")
```
bblommers, line 650: Same here

bblommers (changes requested):
Hi @b0bu, thanks for the PR! Please see my comments

bblommers:
> I ran `make implementation_coverage` doesn't seem to have been run recently though as there's a few other changes.

Yup, that's fine. We tend to run the script right before a release to hover up all the changes, so all the other changes must not've been released yet.

tests/test_codebuild/test_codebuild.py
b0bu: Nice catch thanks! From what I can see `ParamValidation` err object doesn't even have a response property like the others and its `value` is the return message "Invalid length for parameter ids, value: 0, valid min length: 1". I tried it and it seems like this would be better for checking the error type:
```python
err.typename.should.equal("ParamValidationError")
```

b0bu:
> Hi @b0bu, thanks for the PR! Please see my comments

Awesome feedback thanks! And great project, happy to contribute 😀

tests/test_codebuild/test_codebuild.py
bblommers: Hmm.. `err.value.response["Error"]` should always give you a dict with the `Code` and a `Message`. You could also try to use `with pytest.raises(ClientError)`, instead of ParamValidationError, as that as the overarching exception-class, and should always behave the same

b0bu: Tried a few things. `ClientError` isn't catching ParamValidationError 😬. I can see that `typename` has been used throughout the code to prove out the return of catching a ParamValidationError. There's a few cases where I can see something like `assert err["Code"] == "ParamValidationError"` but that pattern isn't working for codebuild, which like you say I'd imagine it would.

```python
# passes
with pytest.raises(ParamValidationError) as err:
    client.batch_get_builds(ids=[])
err.typename.should.equal("ParamValidationError")

# fail with botocore.exceptions.ParamValidationError
    client.batch_get_builds(ids=[])
err.value.response["Error"]["Code"].should.equal("ParamValidationError")
```

bblommers (approved):
LGTM! Thanks again for the PR @b0bu

PR #5288, opened 2022-07-04
Adding support for multi-region keys in KMS and implementing replicate_key API.

taras-kobernyk-localstack:
The current version of moto/kms only supports single-region keys. This change adds some very basic support for multi-regional keys. For details about such keys see https://docs.aws.amazon.com/kms/latest/developerguide/multi-region-keys-overview.html

The added implementation is far from thorough. There is a whole infrastructure on AWS behind multi-regional keys - they get synced with each other automatically if certain fields get changed. Also not all the fields are supposed to be copied to a replica key from its master key. But it is at least something.

### DIFF
--- a/moto/kms/models.py
+++ b/moto/kms/models.py
@@ -1,6 +1,7 @@
 import json
 import os
 from collections import defaultdict
+from copy import copy
 from datetime import datetime, timedelta
 from cryptography.exceptions import InvalidSignature
 from cryptography.hazmat.primitives import hashes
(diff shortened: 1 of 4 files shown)


whummer:
Hi @bblommers , would be great if you could approve the workflow here, so that we can check whether the CI tests are all passing.. Thanks!

bblommers (changes requested):
Apologies for the delay here @taras-kobernyk-localstack.
Could you merge in the latest master? That should fix the build

PR #5297, opened 2022-07-11
check for non-default policy versions before deleting

MrGreenTea:
According [to the docs](https://docs.aws.amazon.com/iot/latest/apireference/API_DeletePolicy.html) all non-default policy versions have to be deleted before deleting a policy itself.


bblommers (approved):
LGTM! Thanks @MrGreenTea

PR #5303, opened 2022-07-13
Allow subscription role attribute to SNS.Subscription

pinzon:
With this PR it is now possible to set the attribute SubscriptionRoleArn for a Subscription. Related to localstack/localstack#6426


bblommers (approved):
LGTM! Thanks for the fix @pinzon

PR #5305, opened 2022-07-15
AWS::AutoScaling::ScheduledAction add more features

sahilshah6196:
Adding support for:
1. describe_scheduled_actions: https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/autoscaling.html#AutoScaling.Client.describe_scheduled_actions
2. delete_scheduled_action: https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/autoscaling.html#AutoScaling.Client.delete_scheduled_action

## Issue
https://github.com/spulec/moto/issues/5335


moto/autoscaling/models.py
```diff
@@ -855,6 +855,25 @@ def put_scheduled_update_group_action(
...
         return scheduled_action
 
+    def describe_scheduled_actions(
+        self, autoscaling_group_name=None, scheduled_action_names=None
+    ):
+        scheduled_actions = []
+        for scheduled_action in self.scheduled_actions.values():
+            if scheduled_action.scheduled_action_name in scheduled_action_names:
+                scheduled_actions.append(scheduled_action)
+            elif not scheduled_action_names and autoscaling_group_name:
```
bblommers, line 865: This doesn't look right - it should only return actions where the group name matches the supplied name.

Additionally, we should have tests to verify this:
 - what happens when providing a non-existent group name? It should list 0 actions
 - what happens when calling `describe` without arguments? Presumably it should return all actions across all groups

bblommers (changes requested):
Hi @sahilshah6196, the index.html file doesn't seem necessary

sahilshah6196:
@bblommers What do you think of this PR?
Thanks!

moto/autoscaling/models.py
sahilshah6196: @bblommers Oh yeah. My bad on that one. Fixed in: [9d8bb61](https://github.com/spulec/moto/pull/5305/commits/9d8bb6113d323c409577f387f338869f0c99c022)

bblommers (approved):
LGTM! Thanks for adding these features @sahilshah6196

PR #5308, opened 2022-07-17
Cache Jinja environment by response type

karolinepauls:
`DynamicDictLoader` supports caching but the loader and the environment would be created on each construction, rather than shared, enabling caching. Here we cache the Jinja environment (and loader) by response type.

In terms of real performance, this gives is a couple of ms (measurements made with the motoserver):

After this change:
```
In [106]: timeit for i in range(10): sqs_client.create_queue(QueueName=str(i))
85.5 ms ± 3.21 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [107]: timeit for i in range(10): sqs_client.create_queue(QueueName=str(i))
68.3 ms ± 3.32 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

Master:
```
In [108]: timeit for i in range(10): sqs_client.create_queue(QueueName=str(i))
99.9 ms ± 3.78 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [109]: timeit for i in range(10): sqs_client.create_queue(QueueName=str(i))
80.5 ms ± 2.98 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

### DIFF
--- a/moto/core/responses.py
+++ b/moto/core/responses.py
@@ -25,6 +25,8 @@ from moto import settings
 
 log = logging.getLogger(__name__)
 
+JINJA_ENVS = {}
+
 
 def _decode_dict(d):
     decoded = OrderedDict()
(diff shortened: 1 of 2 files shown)


bblommers (approved):
Performance improvements are always welcome. Thanks for the PR @karolinepauls!

PR #5309, opened 2022-07-18
add list_targets_for_policy action to IoTBackend

MrGreenTea:
according to the [docs of `ListPolicyPrincipals`](https://docs.aws.amazon.com/iot/latest/apireference/API_ListPolicyPrincipals.html) it is deprecated in favour of [`ListTargetsForPolicy`](https://docs.aws.amazon.com/iot/latest/apireference/API_ListTargetsForPolicy.html). That action is missing from the `IoTBackend` so I added it.

The behaviour is different to `ListPolicyPrincipals` when the policy does not exist:

```
❯ aws iot list-policy-principals --policy-name I_DO_NOT_EXIST
{
    "principals": []
}

❯ aws iot list-targets-for-policy --policy-name I_DO_NOT_EXIST

An error occurred (ResourceNotFoundException) when calling the ListTargetsForPolicy operation: Policy not found
```

### DIFF
--- a/IMPLEMENTATION_COVERAGE.md
+++ b/IMPLEMENTATION_COVERAGE.md
@@ -3483,7 +3483,7 @@
 - [ ] list_security_profiles_for_target
 - [ ] list_streams
 - [ ] list_tags_for_resource
-- [ ] list_targets_for_policy
+- [x] list_targets_for_policy
 - [ ] list_targets_for_security_profile
 - [X] list_thing_groups
 - [X] list_thing_groups_for_thing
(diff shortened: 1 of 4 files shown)


bblommers (changes requested):
Same here actually @MrGreenTea - happy to merge once the conflicts are resolved

MrGreenTea:
@bblommers So finally got around to this last one 👍

bblommers (approved):
Thanks @MrGreenTea! 👍

Issue #5310, opened 2022-07-19
RDS aurora-mysql DB instance engine type


martijnvdp:
## Reporting Bugs

I cant seem to create a rds db instance with instance type aurora-mysql with the latest version

error:
```hcl engine = 'aurora-mysql', storage_type = 'gp2'

    @staticmethod
    def default_allocated_storage(engine, storage_type):
>       return {
            "aurora": {"gp2": 0, "io1": 0, "standard": 0},
            "mysql": {"gp2": 20, "io1": 100, "standard": 5},
            "mariadb": {"gp2": 20, "io1": 100, "standard": 5},
            "postgres": {"gp2": 20, "io1": 100, "standard": 5},
            "oracle-ee": {"gp2": 20, "io1": 100, "standard": 10},
            "oracle-se2": {"gp2": 20, "io1": 100, "standard": 10},
            "oracle-se1": {"gp2": 20, "io1": 100, "standard": 10},
            "oracle-se": {"gp2": 20, "io1": 100, "standard": 10},
            "sqlserver-ee": {"gp2": 200, "io1": 200, "standard": 200},
            "sqlserver-ex": {"gp2": 20, "io1": 100, "standard": 20},
            "sqlserver-se": {"gp2": 200, "io1": 200, "standard": 200},
            "sqlserver-web": {"gp2": 20, "io1": 100, "standard": 20},
        }[engine][storage_type]
E       KeyError: 'aurora-mysql
```

to reproduce:
```hcl
@pytest.fixture
def mysql_db_instance(rds):
    rds_cluster_instance = rds.create_db_instance(
        DBClusterIdentifier="test",
        DBInstanceIdentifier="test-01",
        DBInstanceClass="db.r5.large",
        Engine="aurora-mysql",
    )
    yield
```

The valid values should be as on the aws site https://docs.aws.amazon.com/cli/latest/reference/rds/create-db-instance.html:

```hcl
aurora (for MySQL 5.6-compatible Aurora)
aurora-mysql (for MySQL 5.7-compatible and MySQL 8.0-compatible Aurora)
aurora-postgresql
custom-oracle-ee (for RDS Custom for Oracle instances)
custom-sqlserver-ee (for RDS Custom for SQL Server instances)
custom-sqlserver-se (for RDS Custom for SQL Server instances)
custom-sqlserver-web (for RDS Custom for SQL Server instances)
mariadb
mysql
oracle-ee
oracle-ee-cdb
oracle-se2
oracle-se2-cdb
[...]

PR #5311, opened 2022-07-19
add aurora-mysql and aurora-postgresql to model

martijnvdp:
PR to add aurora-mysql and aurora-postgresql to the db instance module to fix
https://github.com/spulec/moto/issues/5310


bblommers (changes requested):
Hi @martijnvdp, would you mind merging the latest master here? That should fix the build

bblommers (approved):
Awesome, thanks for the fix @martijnvdp!

PR #5312, opened 2022-07-20
fix rds cluster tests failing on multi-az params

martijnvdp:


bblommers (commented):
Can you add some context to this @martijnvdp? I'm not seeing the relationship between the changes and the title, or why the changes are beneficial

bblommers:
The RDS tests have been passing fine for us. If there's an error around missing parameters, that is probably due to an older version of botocore that doesn't yet know about these parameters.

martijnvdp:
> The RDS tests have been passing fine for us. If there's an error around missing parameters, that is probably due to an older version of botocore that doesn't yet know about these parameters.

ok, i also get the errors when i specify these params in phyton code with the db_create_cluster function

PR #5315, opened 2022-07-21
Fix mediaconnect remove_flow_output.

donfiguerres:
remove_flow_output argument OutputArn expects an output name but should expect an output ARN.


bblommers (commented):
Hi @donfiguerres, how does AWS behave when you pass in a name, instead of an ARN?

I know the docs say that it should be an ARN, but there are quite a few services that will happily accept both, so I'm curious whether that's the case here as well.

donfiguerres:
Hi @bblommers , let me check that and get back to you.

PR #5318, opened 2022-07-22
fix attribute addition when updating thing groups

MrGreenTea:
This fixes a bug when updating a thing group without attributes.

To reproduce it:

```py

client = boto3.client("iot")
# important: no thingGroupProperties here
client.create_thing_group(thingGroupName="example")
# will raise KeyError for "attributePayload"
client.update_thing_group(
    thingGroupName=name,
    thingGroupProperties={
        "attributePayload": {
            "attributes": {"example-attribute": "example-value"},
            "merge": False,
        }
    },
)
```

I also fixed a minor difference in behaviour. When a thing group is updated with an empty object for `"attributes"` AWS will simply ignore it and not add it to the properties.

### DIFF
--- a/moto/iot/models.py
+++ b/moto/iot/models.py
@@ -1301,14 +1301,17 @@ class IoTBackend(BaseBackend):
         if attribute_payload is not None and "attributes" in attribute_payload:
             do_merge = attribute_payload.get("merge", False)
             attributes = attribute_payload["attributes"]
-            if not do_merge:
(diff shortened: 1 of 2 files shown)


bblommers (approved):

bblommers (changes requested):
Same here actually @MrGreenTea - happy to merge once the conflicts are resolved

MrGreenTea:
done

bblommers (approved):
Nice one 👍

PR #5323, opened 2022-07-23
Fix/KeyError on get_spot_fleet_request

felipemarinho97:
Prevent KeyError from happening on get_spot_fleet_request and change describe_spot_fleet_instances to behave the same as AWS.


bblommers (changes requested):
Hi @felipemarinho97, would you mind merging in the latest master? That should fix the build

felipemarinho97:
Done @bblommers !

bblommers (approved):
Thanks @felipemarinho97!

PR #5324, opened 2022-07-23
bugfix codepipeline handling iam trust policies and absence of tags.

pvbouwel:
Add a test where a valid trust policy is provided but where tests will fail and add a test where no tags are provided which will fail with the current code.
A commit will be added afterwards to make the tests green.

### DIFF
--- a/moto/codepipeline/models.py
+++ b/moto/codepipeline/models.py
@@ -92,10 +92,13 @@ class CodePipelineBackend(BaseBackend):
 
         try:
             role = self.iam_backend.get_role_by_arn(pipeline["roleArn"])
-            service_principal = json.loads(role.assume_role_policy_document)[
+            trust_policy_statements = json.loads(role.assume_role_policy_document)[
                 "Statement"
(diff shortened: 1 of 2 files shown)


pvbouwel:
For some reason moto server doesn't come up in the test environment. Locally I can start it without an issue, so it doesn't immediately seem related to my PR.

The test_codepipeline folder didn't had an __init__.py and seemed that it was disabled but generally since that is test code that shouldn't influence moto server's capability to get up and running as it is only test related code in there.

bblommers:
The failures are unrelated @pvbouwel - the `werkzeug` dependency released a new version that broke our builds.

Would you mind merging in the latest master? That should fix everything

pvbouwel:
Thanks for the info

Since I don't have dependencies on this yet I just rebased my changes to keep commit history clean. 
I see server starts now so :crossed_fingers: for a green build.

bblommers (approved):
Nice one - thanks for the PR @pvbouwel!

PR #5330, opened 2022-07-25
Downgrade werkzeug

bblommers:
This needs to be fixed properly down the line, but downgrading is good enough for now.

This PR also changes some TF timeouts to ensure the tests have a lower chance of timing out

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -32,7 +32,7 @@ install_requires = [
     "cryptography>=3.3.1",
     "requests>=2.5",
     "xmltodict",
-    "werkzeug>=0.5",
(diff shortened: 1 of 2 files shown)


derlikh-smart:
@bblommers 
What was the reason to downgrade `werkzeug`? I didn't find related issue.

I found this PR during the dependency upgrade. Latest `flask` requires 2.2.2.
It is done in this [commit](https://github.com/pallets/flask/commit/de16718b39e3544688130339a0813997e3790c48).

jakirkham:
Maybe this comment helps ( https://github.com/spulec/moto/issues/5341#issuecomment-1206995825 )?

PR #5334, opened 2022-07-26
IAM: Issues with ListMFADevices and CreateVirtualMFADevice

viren-nadkarni:
- Add a missing attribute `EnableDate` to the `ListMFADevices` response.
- Save base64 encoded attribute `QRCodePNG` as plaintext instead of bytes for `CreateVirtualMFADevice`

Closes https://github.com/localstack/localstack/issues/6451, https://github.com/localstack/localstack/issues/6450


bblommers (approved):
LGTM! Thanks for the fixes @viren-nadkarni

PR #5337, opened 2022-07-26
IAM Access Key limit per user

pinzon:
According to this [doc](https://docs.aws.amazon.com/IAM/latest/UserGuide/id_credentials_access-keys.html) an IAM user can only have a maximum of 2 access keys.

Related to:  https://github.com/localstack/localstack/pull/6488

### DIFF
--- a/moto/iam/models.py
+++ b/moto/iam/models.py
@@ -2427,7 +2427,7 @@ class IAMBackend(BaseBackend):
     def create_access_key(self, user_name=None, status="Active"):
         user = self.get_user(user_name)
         keys = self.list_access_keys(user_name)
-        if len(keys) == LIMIT_KEYS_PER_USER:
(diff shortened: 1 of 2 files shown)


moto/iam/models.py
```diff
@@ -2423,6 +2426,12 @@ def delete_policy(self, policy_arn):
 
     def create_access_key(self, user_name=None, status="Active"):
         user = self.get_user(user_name)
+        keys = self.list_access_keys(user_name)
+        if len(keys) == LIMIT_KEYS_PER_USER:
```
MrGreenTea, line 2430: Just to make this more secure against weird states it would be better to make this check a `>=` in my opinion.

```suggestion
        if len(keys) >= LIMIT_KEYS_PER_USER:
```

bblommers (approved):
LGTM! Thanks @pinzon

Issue #3571, opened 2020-12-30
Return value of GlueClient.get_database does not return full AWS API Database object.


aaronfowles-emis:
I could not find an existing issue so I'm creating this. Please let me know if there's anything else I can do 😄 .

## Description

The problem is that `boto3.Session().client("glue").get_database(Name="...")` does not return a payload with the correct [schema](https://docs.aws.amazon.com/glue/latest/webapi/API_GetDatabase.html).

We would also need to allow `boto3.Session().client("glue").create_database(CatalogId="...", DatabaseInput="...")` to fully use all fields of [DatabaseInput](https://docs.aws.amazon.com/glue/latest/webapi/API_CreateDatabase.html#API_CreateDatabase_RequestSyntax).

## Current behaviour
```python
import moto

with moto.mock_glue():
  glue = boto3.Session().client("glue")
  glue.create_database(
      CatalogId=glue_catalog_id,
      DatabaseInput={
          "Name": "myspecialdatabase",
          "Description": "...",
          "LocationUri": "...",
          "CreateTableDefaultPermissions": [],
          "TargetDatabase": {
              "CatalogId": "...",
              "DatabaseName": "myspecialdatabase",
          },
      },
  )
  print(glue.get_database(Name="myspecialdatabase"))

... 
{
  "Database": {
    "Name": "myspecialdatabase"
  }
}
```

## Desired Behaviour
```python
import moto

with moto.mock_glue():
  glue = boto3.Session().client("glue")
  glue.create_database(
      CatalogId=glue_catalog_id,
      DatabaseInput={
          "Name": "myspecialdatabase",
          "Description": "...",
          "LocationUri": "...",
          "CreateTableDefaultPermissions": [],
          "TargetDatabase": {
              "CatalogId": "...",
              "DatabaseName": "myspecialdatabase",
          },
      },
  )
  print(glue.get_database(Name="myspecialdatabase"))

... 
{
   "Database": { 
      "CatalogId": "...",
      "CreateTableDefaultPermissions": [ ],
      "CreateTime": ...,
      "Description": "...",
      "LocationUri": "...",
      "Name": "myspecialdatabase",
      "Parameters": {},
      "TargetDatabase": { 
         "CatalogId": "...",
         "DatabaseName": ""myspecialdatabase""
      }
   }
}
```

bblommers:
This is now part of moto >= 1.3.16.dev196

timforby:
[...]

PR #5339, opened 2022-07-27
Support CatalogId in create_database

timforby:
Related issue: https://github.com/spulec/moto/issues/3571

> This is still not fully solved, the CatalogId is not correctly handled. As seen in the Boto3 docs for [create_database](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/glue.html#Glue.Client.create_database) the call is made with CatalogId and DatabaseInput.
> 
> This means that in the [glue/responses.py](https://github.com/spulec/moto/blob/004c971a27d00b7a76b42f88cf3b095f672fff17/moto/glue/responses.py#L21) the CatalagId should be placed in the DatabaseInput or handled like database_name.
> 
> My suggestion would be the following:
> 
> ```python
>     def create_database(self):
>         database_input = self.parameters.get("DatabaseInput")
>         database_name = database_input.get("Name")
>         if 'CatalogId' in self.parameters:
>             database_input["CatalogId"] = self.parameters.get("CatalogId")
>         self.glue_backend.create_database(database_name, database_input)
>         return ""
> ```

### DIFF
--- a/moto/glue/responses.py
+++ b/moto/glue/responses.py
@@ -21,6 +21,8 @@ class GlueResponse(BaseResponse):
     def create_database(self):
         database_input = self.parameters.get("DatabaseInput")
         database_name = database_input.get("Name")
+        if "CatalogId" in self.parameters:
+            database_input["CatalogId"] = self.parameters.get("CatalogId")
         self.glue_backend.create_database(database_name, database_input)
         return ""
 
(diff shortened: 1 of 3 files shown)


bblommers (approved):
LGTM - thanks for fixing this @timforby!

PR #5340, opened 2022-07-27
(#5317) Implement IAM back end part of update_assume_role_policy to v…

### LINKED ISSUE #5317: Feature request: Implement IAM back end part of update_assume_role_policy to validate policies
Hello,
The current implementation of update_assume_role_policy for IAM can validate the policy passed in.
I would like to validate this data, so that invalid policies will receive an exception

mmerrill3:
Fixes #5317 to implement a shallow validation of the assume role policy.

Signed-off-by: mmerril8 <mmerril8@its.jnj.com>

### DIFF
--- a/IMPLEMENTATION_COVERAGE.md
+++ b/IMPLEMENTATION_COVERAGE.md
@@ -3290,7 +3290,7 @@
 - [X] untag_user
 - [X] update_access_key
 - [X] update_account_password_policy
-- [ ] update_assume_role_policy
+- [X] update_assume_role_policy
 - [X] update_group
 - [X] update_login_profile
 - [X] update_open_id_connect_provider_thumbprint
--- a/docs/docs/services/iam.rst
+++ b/docs/docs/services/iam.rst
@@ -192,7 +192,7 @@ iam
 - [X] untag_user
 - [X] update_access_key
 - [X] update_account_password_policy
-- [ ] update_assume_role_policy
+- [X] update_assume_role_policy
 - [X] update_group
 - [X] update_login_profile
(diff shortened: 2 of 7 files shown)


bblommers:
The TF test fails with this change - one Cognito-test tries to update a (valid) policy that Moto rejects.
I don't have time to debug this properly right now, unfortunately. I don't know whether our validator is too strict, or whether there our different rules for different policies.

mmerrill3:
> The TF test fails with this change - one Cognito-test tries to update a (valid) policy that Moto rejects. I don't have time to debug this properly right now, unfortunately. I don't know whether our validator is too strict, or whether there our different rules for different policies.

I'll check it out

bblommers:
Thanks @mmerrill3. I'd be happy to look into it as well, but that would take a few weeks l, as I don't have any availability soon

mmerrill3:
@bblommers , for trust policies, we don't need to enforce that the resource is there.  The source trust policy that creates the issue is here: [sts:AssumeRole](https://github.com/hashicorp/terraform-provider-aws/blob/632cff7679cb6e1e14076b9aac3e68b73f584b70/internal/service/cognitoidp/user_pool_test.go#L1490)

I'll work on updating the PR so this can work for trust policies attached to a role.  It kind of makes sense, since the resource would be the role itself, right?

mmerrill3:
@bblommers ready for review.  I refactored the policy validator to reuse some base functionality.  For trust policies, we shouldn't be setting a resource, and the actions are limited to sts:assumeRole and a couple others.

bblommers (approved):
Nice one - thanks for fixing this @mmerrill3!

Issue #5146, opened 2022-05-17
Add support for AWS Glue Schema Registry


blacktooth:
Support for AWS Glue doesn't include the schema registry APIs like `CreateSchema`, `RegisterSchemaVersion`, `GetSchemaVersion` etc. Schema Registry APIs were added to Glue in 2020. Is this feature on the roadmap of moto? Are you open to contributions?

bblommers:
Hi @blacktooth! There is no roadmap as such - most of our PR's come from contributors who need a specific feature.
So if you want to create a PR for this, that would be very welcome! 

We have a [Contributor Section](http://docs.getmoto.org/en/latest/docs/contributing/index.html) in our documentation that should get you started. Feel free to ask any questions if you get stuck!

hiimanii:
Hi @bblommers, can you review the [pull request](https://github.com/spulec/moto/pull/5301). Do you have any comments?

bblommers:
Hi @blacktooth, @hiimanii, were there any remaining Glue features/methods that you would like to see in Moto? Or did the PR's cover everything you need?

blacktooth:
The PRs covered everything we need for now! Thanks for your support on reviews! We will open a new issue in case we want to add more APIs.

bblommers:
Sounds good to me @blacktooth - thanks!

PR #5351, opened 2022-08-02
Feature: Glue Schema Registry APIs: GetSchemaVersion, GetSchemaByDefinition, DeleteSchema, PutSchemaVersionMetadata

hiimanii:
Feature Added: GetSchemaVersion API from Glue Schema Registry

Feature: Implemented GetSchemaByDefinition API featues from Glue Schema Registry.

Feature: Implemented PutSchemaVersionMetadata API featues from Glue Schema Registry.

Bug Fix: CreateSchema with default-registry

Feature: DeleteSchema API for Glue Schema Registry

This is for #5146

### DIFF
--- a/moto/glue/glue_schema_registry_utils.py
+++ b/moto/glue/glue_schema_registry_utils.py
@@ -20,6 +20,15 @@ from .glue_schema_registry_constants import (
     MAX_SCHEMAS_ALLOWED,
     MAX_SCHEMA_VERSIONS_ALLOWED,
     MAX_REGISTRIES_ALLOWED,
+    SCHEMA_VERSION_ID_PATTERN,
+    SCHEMA_VERSION_ID,
+    VERSION_NUMBER,
+    LATEST_VERSION,
+    METADATA_VALUE,
+    METADATA_KEY,
+    MAX_SCHEMA_VERSION_METADATA_LENGTH,
+    SCHEMA_VERSION_METADATA_PATTERN,
+    MAX_SCHEMA_VERSION_METADATA_ALLOWED,
 )
 
 from .exceptions import (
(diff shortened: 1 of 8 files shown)


tests/test_glue/test_schema_registry.py
```diff
@@ -0,0 +1,1341 @@
...
+    TEST_INVALID_SCHEMA_ID_SCHEMA_DOES_NOT_EXIST,
+    TEST_INVALID_SCHEMA_ID_REGISTRY_DOES_NOT_EXIST,
+    TEST_METADATA_KEY_VALUE,
+    TEST_METADATA_KEY,
+    TEST_METADATA_VALUE,
+    TEST_DELETING_STATUS,
+)
+
+
+def create_glue_client():
```
MrGreenTea, line 45: Why not make this a `pytest` fixture?

moto/glue/glue_schema_registry_utils.py
```diff
@@ -0,0 +1,411 @@
...
+    InvalidSchemaVersionNumberBothParamsProvidedException,
+    InvalidSchemaVersionIdProvidedWithOtherParamsException,
+    InvalidSchemaVersionNumberNotProvidedException,
+    SchemaVersionMetadataLimitExceededException,
+)
+
+
+def validate_registry_name_pattern_and_length(param_value):
+    param_name = "registryName"
+    max_name_length = MAX_REGISTRY_NAME_LENGTH
```
MrGreenTea, line 58: Why assign these constants to a local variable? Can't they be inlined into the function call that immediately follows?

So in this case it would become this instead:

```py
validate_param_pattern_and_length(
  param_value, 
  param_name = "registryName", 
  max_name_length=MAX_REGISTRY_NAME_LENGTH, 
  pattern=RESOURCE_NAME_PATTERN
)
```

hiimanii: Thanks for pointing that out.

tests/test_glue/test_schema_registry.py
hiimanii: Are suggesting to move it to the file: `glue_schema_registry_utils.py`?
That makes sense.

MrGreenTea: Not really, I suggest changing it to this:
```suggestion
@pytest.fixture
def glue_client():  # or just client()
```

and then instead of the current way of calling 
```py
def test_XXX():
    client = create_glue_client
    ...
```

You would do 
```py
def test_XXX(glue_client):  # or just test_XXX(client)
  ...
```

hiimanii: Okay, got it.

bblommers:
@hiimanii  Now that the first PR is merged, I'll review this once it's been rebased, and the merge conflicts are resolved

moto/glue/glue_schema_registry_utils.py
hiimanii: Doing this would be Redefining name `glue_client` or `client` from outer scope. And this is not advisable.

tests/test_glue/test_schema_registry.py
hiimanii: Doing this would be Redefining name `glue_client` or `client` from outer scope. And this is not advisable.

MrGreenTea: It is how you do and use pytest fixtures and therefore recommended (and even necessary).

hiimanii: okay:)

hiimanii:
@bblommers I have rebased and resolved the merge conflicts. This PR can be merged now. Thanks:)

tests/test_glue/test_schema_registry.py
```diff
@@ -9,57 +9,65 @@
...
+    TEST_METADATA_KEY,
+    TEST_METADATA_VALUE,
+    TEST_DELETING_STATUS,
 )
 
 
-def create_glue_client():
+@pytest.fixture
+def client():
     return boto3.client("glue", region_name="us-east-1")
```
MrGreenTea, line 47: With this you could move all the `mock_glue` decorators in this fixture:
```suggestion
    with mock_glue():
        yield boto3.client("glue", region_name="us-east-1")
```

This will make sure that all tests that use the client through this fixture have mocking applied. We could then remove the `@mock_glue` from all tests and can't forget it in future tests :)

MrGreenTea (changes requested):

moto/glue/models.py
```diff
@@ -288,7 +304,14 @@ def create_schema(
         tags=None,
     ):
         """CrateSchema API"""
-        """Validate Registry Id"""
+        warnings.warn(
```
bblommers, line 307: Implementation details (or non-implementations, in this case) sh ould not be exposed as a warning.
This should be part of the docstring instead, and it will show up in the documentation.
```
"""
The following parameters/features are not yet implemented:
....
"""
```

moto/glue/models.py
```diff
@@ -375,8 +406,186 @@ def register_schema_version(self, schema_id, schema_definition):
...
+        ) = validate_schema_version_params(
+            self.registries, schema_id, schema_version_id, schema_version_number
+        )
+
+        # PutSchemaVersionMetadata using SchemaVersionId
+        if schema_version_id:
+            for registry in self.registries.values():
+                for schema in registry.schemas.values():
+                    if schema.schema_versions.get(schema_version_id, None):
+                        validate_number_of_schema_version_metadata_allowed(
```
bblommers, line 508: This part is a bit hard to read, with all the newlines.
Maybe we can extract `schema.schema_versions[schema_version_id].metadata` into a temporary variable? That may help a lot in improving the readability

tests/test_glue/test_schema_registry.py
bblommers: Agreed - with so many tests, it is always nice to reduce repetition

bblommers (changes requested):

tests/test_glue/test_schema_registry.py
hiimanii: Got it:)

moto/glue/models.py
hiimanii: Yeah. That really improves the readability.

moto/glue/models.py
```diff
@@ -375,8 +403,173 @@ def register_schema_version(self, schema_id, schema_definition):
...
+                        validate_number_of_schema_version_metadata_allowed(metadata)
+
+                        if metadata_key in metadata:
+                            if metadata_value in metadata[metadata_key]:
+                                raise SchemaVersionMetadataAlreadyExistsException(
+                                    schema_version_id, metadata_key, metadata_value
+                                )
+                            metadata[metadata_key].append(metadata_value)
+
+                        metadata[metadata_key] = [metadata_value]
```
bblommers, line 515:
```suggestion
                            metadata[metadata_key].append(metadata_value)
                        else:
                            metadata[metadata_key] = [metadata_value]
```
Now that it's more readable - it's suddenly looks like there's an `else` missing :)

moto/glue/models.py
```diff
@@ -375,8 +403,173 @@ def register_schema_version(self, schema_id, schema_definition):
...
+                if metadata_key in schema_version.metadata:
+                    if metadata_value in schema_version.metadata[metadata_key]:
+                        raise SchemaVersionMetadataAlreadyExistsException(
+                            schema_version.schema_version_id,
+                            metadata_key,
+                            metadata_value,
+                        )
+                    schema_version.metadata[metadata_key].append(metadata_value)
+
+                schema_version.metadata[metadata_key] = [metadata_value]
```
bblommers, line 541: Same here

bblommers (approved):
Thanks @hiimanii!

PR #5352, opened 2022-08-02
check if policy name already taken

MrGreenTea:
```
❯ aws iot create-policy --policy-name example --policy-document '{"Version": "2012-10-17", "Statement": [{"Effect": "Allow", "Acti
on": "*", "Resource": "*"}]}'

An error occurred (ResourceAlreadyExistsException) when calling the CreatePolicy operation: Policy cannot be created - name already exists (name=example)
```

Right now the policy is just overwritten in `moto`


bblommers (approved):
Looks good to me! Thanks @MrGreenTea

Issue #5360, opened 2022-08-05
Add SES SendBulkTemplatedEmail enhancement


deepaksrgm2010:
send_bulk_templated_email is not yet implemented in ses, so I would like to implement this.

PR #5361, opened 2022-08-05
SES: Implement SendBulkTemplatedEmail

deepaksrgm2010:
Feature: Implemented SendBulkTemplatedEmail API feature from SES.

This is [#5360](https://github.com/spulec/moto/issues/5360)

### DIFF
--- a/moto/ses/responses.py
+++ b/moto/ses/responses.py
@@ -94,6 +94,35 @@ class EmailResponse(BaseResponse):
         template = self.response_template(SEND_TEMPLATED_EMAIL_RESPONSE)
         return template.render(message=message)
 
+    def send_bulk_templated_email(self):
+        source = self.querystring.get("Source")[0]
+        template = self.querystring.get("Template")
+        template_data = self.querystring.get("DefaultTemplateData")
+
+        destinations = []
+        for i in range(1, 52):
+            destination_field = "Destinations.member.%s.Destination.ToAddresses.member.1" % (i)
+            if self.querystring.get(destination_field) is None:
+                break
+            destination = {"ToAddresses": [], "CcAddresses": [], "BccAddresses": []}
+            for dest_type in destination:
+                # consume up to 51 to allow exception
+                for j in range(1, 52):
+                    field = "Destinations.member.%s.Destination.%s.member.%s" % (i, dest_type, j)
+                    address = self.querystring.get(field)
+                    if address is None:
+                        break
+                    destination[dest_type].append(address[0])
+            destinations.append({"Destination": destination})
+
+
+        message = self.backend.send_bulk_templated_email(
+            source, template, template_data, destinations, self.region
+        )
+        template = self.response_template(SEND_BULK_TEMPLATED_EMAIL_RESPONSE)
+        result = template.render(message=message)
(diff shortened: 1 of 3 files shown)


moto/ses/responses.py
```diff
@@ -94,6 +94,35 @@ def send_templated_email(self):
...
+            destination = {"ToAddresses": [], "CcAddresses": [], "BccAddresses": []}
+            for dest_type in destination:
+                # consume up to 51 to allow exception
+                for j in range(1, 52):
+                    field = "Destinations.member.%s.Destination.%s.member.%s" % (i, dest_type, j)
+                    address = self.querystring.get(field)
+                    if address is None:
+                        break
+                    destination[dest_type].append(address[0])
+            destinations.append({"Destination": destination})
```
bblommers, line 116: Can we retrieve these parameters using one of the existing helper methods, such as `self._get_params()` or `self._get_multi_param_dict`?

tests/test_ses/test_ses_boto3.py
```diff
@@ -174,6 +174,68 @@ def test_send_email_invalid_address():
...
+        }
+    )
+
+    conn.send_bulk_templated_email(**kwargs)
+
+    too_many_destinations = list({"Destination": {"ToAddresses": ["to%s@example.com" %i ], "CcAddresses": [],
+                                                  "BccAddresses": [] }} for i in range(51))
+    conn.send_bulk_templated_email.when.called_with(
+        **dict(kwargs, Destinations=too_many_destinations)
+    ).should.throw(ClientError)
```
bblommers, line 225: We should verify that the correct error was thrown, like you did above:
```
with pytest.raises() as ex:
    ....
ex... should.equal(...)
```

tests/test_ses/test_ses_boto3.py
```diff
@@ -174,6 +174,68 @@ def test_send_email_invalid_address():
...
+
+    too_many_destinations = list("to%s@example.com" % i for i in range(51))
+
+    conn.send_bulk_templated_email.when.called_with(
+         **dict(kwargs, Destinations=[{"Destination": {"ToAddresses": too_many_destinations,
+                                                       "CcAddresses": [], "BccAddresses": []}}])
+    ).should.throw(ClientError)
```
bblommers, line 232: Same

bblommers (changes requested):
Hi @deepaksrgm2010, thanks for the PR! Please see my comments.

Note that the linter is currently failing as well, that should also be fixed

moto/ses/responses.py
deepaksrgm2010: @bblommers
-  I tried using self._get_params(). There is `index out of bound` error when its used decode query string whose memebers are out of order (I know becuase my tests failed when i used them. but only on specific cases).
Note: I plan have look at this helper method later and fix it if possible
- Rest of the methods in this module uses the same convention I have. So for now can we keep it this way.

deepaksrgm2010:
Apologies for the lint error. Some error in my lint set up. Corrected it.

bblommers (approved):
LGTM - thanks @deepaksrgm2010!

PR #5365, opened 2022-08-06
fix(ec2): handle non-existent template for describe-launch-templates

zetamorph:
Currently, when calling `ec2 describe-launch-templates` with a template name
that does not (yet) exist, an unhandled exception occurs and HTTP 500 is
returned.

This commit adds a check for the existence of the passed launch template name.
If a template by this name does not exist, an `InvalidLaunchTemplateNameNotFoundError` is raised.
It also renames the current `InvalidLaunchTemplateNameError` to `InvalidLaunchTemplateNameAlreadyExistsError`,
so that both cases are representable.

### DIFF
--- a/moto/ec2/exceptions.py
+++ b/moto/ec2/exceptions.py
@@ -702,11 +702,19 @@ class OperationNotPermitted4(EC2ClientError):
         )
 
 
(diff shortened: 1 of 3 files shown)


bblommers (approved):
LGTM! Thanks for fixing this @zetamorph

PR #5367, opened 2022-08-08
EC2: Add ec2 fleets basic api

felipemarinho97:
This MR adds:
 - EC2 CreateFleet
 - EC2 DescribeFleets
 - EC2 DescribeFleetInstances
 - EC2 DeleteFleets

The main reason is because I need they for another project that I'm working on. I'm not a 24/7 python developer, but most of the implementation was based on ~stolen from~ the `spot_requests` files.

### DIFF
--- a/moto/ec2/models/fleets.py
+++ b/moto/ec2/models/fleets.py
@@ -1,0 +1,311 @@
+from collections import defaultdict
+
+from moto.ec2.models.spot_requests import SpotFleetLaunchSpec
+
+from .core import TaggedEC2Resource
+from ..utils import (
+    random_fleet_id,
+    convert_tag_spec,
+)
+
+
+class Fleet(TaggedEC2Resource):
+    def __init__(
+        self,
+        ec2_backend,
+        fleet_id,
+        on_demand_options,
+        spot_options,
+        target_capacity_specification,
+        launch_template_configs,
+        excess_capacity_termination_policy,
+        replace_unhealthy_instances,
+        terminate_instances_with_expiration,
+        fleet_type,
+        valid_from,
+        valid_until,
+        tag_specifications,
+    ):
+
+        self.ec2_backend = ec2_backend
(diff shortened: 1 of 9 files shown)


moto/ec2/models/fleets.py
```diff
@@ -0,0 +1,311 @@
...
+                break
+
+            instance_ids.append(instance.id)
+            new_fulfilled_capacity -= 1
+
+        self.spot_requests = [
+            req for req in self.spot_requests if req.instance.id not in instance_ids
+        ]
+        self.on_demand_instances = [
+            req for req in self.on_demand_instances if req["id"] not in instance_ids
```
bblommers, line 242: What is the reason for using `req["id"]` here, which refers to the reservation ID? Why not `req["instance"].id`?

moto/ec2/models/instances.py
```diff
@@ -367,18 +368,30 @@ def terminate(self):
...
+            fleet.spot_requests = [
+                req for req in fleet.spot_requests if req.instance != self
             ]
+            try:
+                fleet.on_demand_instances = [
+                    inst
+                    for inst in fleet.on_demand_instances
+                    if inst["instance"] != self
+                ]
+            except AttributeError:
```
bblommers, line 393: This doesn't look right to me - in what instance would this assignment throw an AttributeError?
And why can we just ignore that?

bblommers (changes requested):

moto/ec2/models/instances.py
felipemarinho97: This is because the attribute `on_demand_instances` is only present on a `fleet` resource, `spot_fleet_request` resource doesn't have on demand instances. Yea, it doesn't look right.. I can change to a _if statement_ checking if the `id` starts with `fleet` if it's better..

bblommers (approved):
LGTM! Thanks @felipemarinho97

PR #5368, opened 2022-08-08
Enhancement/docker dev

lidia-k:
Currently the development installation guide only explains about setting moto up in a virtual environment, not with Docker. 
There is already a dockerfile but with the current dockerfile, it requires a few more manual steps to complete running moto such as getting into a docker container, installing make, running a make command to install requirements, etc. 
I updated the dockerfile to implement the initial steps as a part of the docker build and created a new make command `make docker-dev` to build, run and volumize a docker container all in one command.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,8 +5,10 @@ ENV PYTHONUNBUFFERED 1
 
 WORKDIR /moto/
 RUN  pip3 --no-cache-dir install --upgrade pip setuptools && \
-     pip3 --no-cache-dir install ".[server]"
+     pip3 --no-cache-dir install ".[server]" && \
+     python3 setup.py develop && \
+     pip3 --no-cache-dir install -r requirements-dev.txt
 
-ENTRYPOINT ["/usr/local/bin/moto_server", "-H", "0.0.0.0"]
+CMD ["python3", "/moto/moto/server.py", "-H", "0.0.0.0"]
 
 EXPOSE 5000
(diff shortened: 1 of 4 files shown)


bblommers (commented):
Thanks for the PR @lidia-k - makes sense to extend the Dockerfile to include this.
If we're going add this however, I would like to see the Make-command as part of our CI as well to verify it works, and some documentation to let developers know that this feature exists.

lidia-k:
Thanks for the feedback @bblommers. As per your request, I included the new make command for docker as part of the CI and successfully tested it locally. I also updated the installation doc on how to use docker to set up a development server. Let me know if you have any other comments. Thank you.

Dockerfile
```diff
@@ -5,8 +5,11 @@ ENV PYTHONUNBUFFERED 1
 
 WORKDIR /moto/
 RUN  pip3 --no-cache-dir install --upgrade pip setuptools && \
-     pip3 --no-cache-dir install ".[server]"
+     pip3 --no-cache-dir install ".[server]" && \
+     python3 setup.py develop && \
```
bblommers, line 9: The changes here are a bit confusing to me - do we need them?
`pip install .` already install all the runtime dependencies, and we shouldn't need the dev-dependencies at this point.
The make-command is not used inside Docker/while running `moto/server.py`, so that's also unnecessary.

On a general note: This Docker-file is mainly used by the maintainers to create our releases, so I'd like to keep this intact if possible. If it does need changing to use this as a development server, maybe we can have a separate Dockerfile for that purpose? (I'm not too familiar with Docker, so if there is a better way to ensure we can serve both use-cases, I'm all ears.)

PR #5370, opened 2022-08-09
Key pairs must have unique IDs

viren-nadkarni:
### Summary

Makes all key pairs have a `KeyPairId`.

### Related

Closes: https://github.com/localstack/localstack/issues/6614


bblommers (approved):
LGTM - thanks @viren-nadkarni

Issue #5356, opened 2022-08-03
Error listing Lambda layers after layer deletion


jhw:
Hi

`lambda.list_layers()` fails if you create a layer and then subsequently delete it

Here's an example illustrating the problem -

```
from moto import mock_lambda, mock_s3

import boto3, unittest, zlib

MyBucket={"name": "my-bucket"}

PyyamlLayer={"name": "pyyaml",
             "bucket": "my-bucket"}

@mock_lambda
@mock_s3
class MockLayerDeleteTest(unittest.TestCase):

    def setup_s3(self, buckets=[MyBucket], layers=[PyyamlLayer]):
        self.s3=boto3.client("s3")
        for bucket in buckets:
            self.s3.create_bucket(Bucket=bucket["name"],
                                  CreateBucketConfiguration={'LocationConstraint': 'EU'})
        for layer in layers:
            s3key="layer-%s.zip" % layer["name"]
            payload=zlib.compress("how now brown cow".encode('utf-8'))
            self.s3.put_object(Bucket=layer["bucket"],
                               Key=s3key,
                               Body=payload,
                               ContentType="application/gzip")
            
    def setUp(self):
        self.setup_s3()
        self.lamda=boto3.client("lambda")
            
    def test_create_delete_list_layer(self, bucket=MyBucket, layer=PyyamlLayer):
        s3key="layer-%s.zip" % layer["name"]
        print ("--- creating layer ---")
        resp=self.lamda.publish_layer_version(LayerName=layer["name"],
                                              Content={"S3Bucket": bucket["name"],
                                                       "S3Key": s3key})
        layerarn=resp["LayerVersionArn"]
        layerversion=int(layerarn.split(":")[-1])
        print (self.lamda.list_layers()) # <- WORKS FINE HERE
        print ("--- deleting layer ---")
        self.lamda.delete_layer_version(LayerName=layer["name"],
                                        VersionNumber=layerversion)
        print (self.lamda.list_layers()) # <- FAILS HERE
    
    def teardown_s3(self, buckets=[MyBucket]):
        def empty_bucket(s3, bucket):            
            struct=s3.list_objects(Bucket=bucket["name"])
            if "Contents" in struct:
                for obj in struct["Contents"]:
                    # print ("deleting %s" % obj["Key"])
                    s3.delete_object(Bucket=bucket["name"],
                                     Key=obj["Key"])
        def delete_bucket(s3, bucket):
            s3.delete_bucket(Bucket=bucket["name"])
[...]

PR #5376, opened 2022-08-12
fix: Error listing Lambda layers after layer deletion

hajali-amine:
- [ ] #5356

After reading the ticket above, I decided to try and fix the bug 😄 

## Context

The bug happened when a user created a layer, which resulted in a new layer version, and then deleted that same layer.

The issue here is that at first, there is no layer. When we create one, we add a new `Layer` to the `LayerStorage._layers`. Now, that same `Layer` we are going to add to its dict attribute `layer_versions` the first version. 

So now, `Layer` has `_latest_version = 1` and `layer_versions = {'1': layer_v1}`

Now when we delete that layer version, we don't reset `_latest_version` since we want to keep track of our versions, but we do remove element in the dictionary. I think _that's bug number one_

We also keep the layer instance in the Layer storage, and we retrieve all Layers, we used to retrieve all even the ones that have no versions in store. I think _that's bug number two_

## Fix

For _bug number one_, we can query for the latest version instead of using the `_latest_version` attribute since we can only increment it.

``` python
    def to_dict(self):
        if not self.layer_versions:
            return {}

        last_key = sorted(self.layer_versions.keys(), key=lambda version: int(version))[-1]
        return {
            "LayerName": self.name,
            "LayerArn": self.layer_arn,
            "LatestMatchingVersion": self.layer_versions[
                last_key
            ].get_layer_version(),
        }
```

For _bug number two_, we can get only Layers that have available version from the Layer storage.

``` python
    def list_layers(self):
        return [layer.to_dict() for layer in self._layers.values() if layer.layer_versions]
```

## Conclusion

This is actually my first PR ever in an open source project! So thank you!
Can't wait for the feedback 😄

### DIFF
--- a/moto/awslambda/models.py
+++ b/moto/awslambda/models.py
@@ -364,11 +364,15 @@ class Layer(object):
         self.layer_versions.pop(str(layer_version), None)
 
     def to_dict(self):
+        if not self.layer_versions:
+            return {}
+            
+        last_key = sorted(self.layer_versions.keys(), key=lambda version: int(version))[-1]
         return {
             "LayerName": self.name,
             "LayerArn": self.layer_arn,
             "LatestMatchingVersion": self.layer_versions[
-                str(self._latest_version)
+                last_key
             ].get_layer_version(),
         }
 
@@ -1242,7 +1246,7 @@ class LayerStorage(object):
         self._layers[layer_version.name].attach_version(layer_version)
 
     def list_layers(self):
-        return [layer.to_dict() for layer in self._layers.values()]
+        return [layer.to_dict() for layer in self._layers.values() if layer.layer_versions]
 
     def delete_layer_version(self, layer_name, layer_version):
         self._layers[layer_name].delete_version(layer_version)
--- a/moto/awslambda/models.py
+++ b/moto/awslambda/models.py
@@ -366,14 +366,14 @@ class Layer(object):
     def to_dict(self):
         if not self.layer_versions:
             return {}
-            
-        last_key = sorted(self.layer_versions.keys(), key=lambda version: int(version))[-1]
+
+        last_key = sorted(self.layer_versions.keys(), key=lambda version: int(version))[
+            -1
+        ]
         return {
             "LayerName": self.name,
             "LayerArn": self.layer_arn,
-            "LatestMatchingVersion": self.layer_versions[
-                last_key
-            ].get_layer_version(),
+            "LatestMatchingVersion": self.layer_versions[last_key].get_layer_version(),
         }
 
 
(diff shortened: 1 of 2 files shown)


hajali-amine:
Something that I am not sure of;
When we delete all of the versions from a layer, should we remove the Layer from the storage?

bblommers:
Hi @hajali-amine, thank you for the PR, and welcome to Moto and open source in general!

> Something that I am not sure of; When we delete all of the versions from a layer, should we remove the Layer from the storage?

That is a good question - and we should figure out what AWS does in this scenario. If they delete the Layer itself, we can do the same. If, when listing a layer without any versions, they just show the LayerName and LayerARN and nothing else, we can follow that behaviour.

bblommers:
We should have a unit test that verifies this behaviour, btw. The existing tests around layers can be found in `tests/test_awslambda/test_lambda_layers`, so it would be good to add another test there.

hajali-amine:
@bblommers will start on the test tonight 😄
As for the layer, I will test it on AWS also tonight by adding a Layer and removing it then adding an another one with the same name and check if the version increments or not. If not, that means it's deleted from the storage.

hajali-amine:
**Update:**

- AWS still stores layers and keeps track of the version increment even after all of the versions are deleted.
- Added a UT for the scenario

moto/awslambda/models.py
```diff
@@ -1242,7 +1246,9 @@ def put_layer_version(self, layer_version):
         self._layers[layer_version.name].attach_version(layer_version)
 
     def list_layers(self):
-        return [layer.to_dict() for layer in self._layers.values()]
+        return [
+            layer.to_dict() for layer in self._layers.values() if layer.layer_versions
```
bblommers, line 1250: Just to double check..

> Actually, the current list-layer is right, even AWS returns LayerName, LayerARN and version.
> 

So we can reverse this? Or did you mean `the updated implementation for list-layer is right`?

bblommers:
I should have mentioned this earlier, but the tests for AWSLambda run in parallel. So that means a call to `list_layers` may also return layers that were created in other tests. So make sure that any assertions account for that - it looks like that's the reason our CI is currently failing.

http://docs.getmoto.org/en/latest/docs/contributing/development_tips/tests.html#parallel-tests

moto/awslambda/models.py
hajali-amine: ## What happens in AWS

Let's say we create a new Layer. The LayerStorage has a new Layer with a new LayerVersion (v1).
When we delete that Layer, we can not actually retrieve it with `list-layers` since the Layer was deleted.

However, I believe that it's still saved in LayerStorage. Why? Well, because when I create a new Layer with the same name, I will get a Layer with LayerVersion (_**v2**_). The LayerStorage keeps the Layer saved, even if it has no LayerVersions, just to keep track of the _version counter_. I think AWS soft-deletes layers.

## Moto prior to my changes

When I add a Layer and delete its LayerVersion, the call to `list-layers` raises an error since we are calling `layer.to_dict()` to a layer with no layer_versions. (the list layer_versions is empty)

## Moto after my changes

`list-layers` only returns layers with layer_versions.

hajali-amine: The thing is, layer.to_dict() should never be called to a Layer with no LayerVersions 😄

bblommers: Makes sense!

bblommers (approved):
LGTM - thanks for investigating and fixing this @hajali-amine!

Issue #5270, opened 2022-06-27
DynamoDB - No validation on reserved keywords for get_item


JorisLimousin:
Hi,

Just raising this issue as a potential enhancement to add validation to the `get_item` function for reserved keywords being present in the `ProjectionExpression`. Using the function like this: `result = db_table.get_item(Key={...},ProjectionExpression="name")` should throw the following exception: `botocore.exceptions.ClientError: An error occurred (ValidationException) when calling the GetItem operation: Invalid ProjectionExpression: Attribute name is a reserved keyword; reserved keyword: name`.

PR #5380, opened 2022-08-12
enhancement: add DynamoDB Projection Expression constraints check

hajali-amine:
- [ ] #5270

## Context

When doing `scan`, `get_item` and `query` operations, we can specify which attributes to get instead of getting all of them. For this case, AWS provided __projection expression__.

However, the value of __projection expression__ have some restrictions.

As specified in the [AWS documentation](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Expressions.ExpressionAttributeNames.html):

> If an attribute name begins with a number, contains a space or contains a reserved word, you must use an expression attribute name to replace that attribute's name in the expression.

## Bug

Currently, we do not really check on the __projection expression__ value at all. Thus, as specified in the issue, when we do the following;

``` python
client.query(
            TableName=table_name,
            KeyConditionExpression="a",
            ProjectionExpression="name",
        )
```

It won't raise an error, even though `name` is a reserved keyword.

## Fix

When we treat the projection expressions and attributes names, we check on the validity of the projection expressions.

``` python
def _adjust_projection_expression(self, projection_expression, expr_attr_names):
    def _adjust(expression):
        return (
            expr_attr_names[expression]
            if expression in expr_attr_names
            else expression
        )

    if projection_expression:
        expressions = [x.strip() for x in projection_expression.split(",")]
        for expression in expressions:
            check_projection_expression(expression) # We check here
        if expr_attr_names:
            return ",".join(
                [
                    ".".join([_adjust(expr) for expr in nested_expr.split(".")])
                    for nested_expr in expressions
                ]
            )

    return projection_expression
```

Now, how do we check? Simply like this!

``` python
def check_projection_expression(expression):
    if expression.upper() in ReservedKeywords.get_reserved_keywords():
        raise MockValidationException(
            f"ProjectionExpression: Attribute name is a reserved keyword; reserved keyword: {expression}"
        )
    if expression[0].isnumeric():
        raise MockValidationException(
            "ProjectionExpression: Attribute name starts with a number"
        )
    if " " in expression:
        raise MockValidationException(
            "ProjectionExpression: Attribute name contains white space"
        )
```

## Conclusion

I will be awaiting the feedback and review! 
Thank you 😄

### DIFF
--- a/moto/dynamodb/models/__init__.py
+++ b/moto/dynamodb/models/__init__.py
@@ -1485,13 +1485,6 @@ class DynamoDBBackend(BaseBackend):
             filter_expression, expr_names, expr_values
         )
 
-        projection_expression = ",".join(
-            [
-                expr_names.get(attr, attr)
-                for attr in projection_expression.replace(" ", "").split(",")
-            ]
-        )
-
         return table.scan(
             scan_filters,
             limit,
--- a/moto/dynamodb/responses.py
+++ b/moto/dynamodb/responses.py
@@ -7,6 +7,7 @@ from functools import wraps
 
 from moto.core.responses import BaseResponse
 from moto.core.utils import camelcase_to_underscores, amz_crc32, amzn_request_id
+from moto.dynamodb.parsing.reserved_keywords import ReservedKeywords
 from .exceptions import (
     MockValidationException,
     ResourceNotFoundException,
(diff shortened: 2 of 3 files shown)


bblommers (approved):
LGTM! Thanks @hajali-amine

PR #5381, opened 2022-08-13
Fix/launch template tags

felipemarinho97:
Currently, launch templates in moto only support TagSpecifications from inside the LaunchTemplateData (for tagging instances and volumes). This MR adds support for tagging of the launch-template resource itself.


bblommers (approved):
LGTM @felipemarinho97 - thanks for the PR. 

We're currently in a code freeze while we're releasing Moto v4, but this will be merged after that's all done.

PR #5382, opened 2022-08-13
Feat/describe instance types filtering

felipemarinho97:
Added a small subset of filters as documented here: https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_DescribeInstanceTypes.html

### DIFF
--- a/moto/ec2/models/instance_types.py
+++ b/moto/ec2/models/instance_types.py
@@ -50,7 +50,7 @@ class InstanceType(dict):
         elif filter_name in ("current-generation"):
             return str(self.get("CurrentGeneration")).lower()
         else:
-            return super().get_filter_value(filter_name, "DescribeSubnets")
+            return super().get_filter_value(filter_name, "DescribeInstanceTypes")
 
 
 class InstanceTypeBackend:
--- a/moto/ec2/models/instance_types.py
+++ b/moto/ec2/models/instance_types.py
@@ -22,7 +22,7 @@ for location_type in listdir(root / offerings_path):
 
 
 class InstanceTypeBackend:
-    def describe_instance_types(self, instance_types=None):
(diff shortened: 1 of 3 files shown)


bblommers (changes requested):
A lot of the filtering logic already exists, and is part of the `moto.ec2.utils.generic_filter()` method.

A good example of this can be found in how Subnets are filtered by.
The Subnet-class itself contains the logic on what filter-key corresponds to which value:

https://github.com/spulec/moto/blob/78b89684157857d6f288de280d8f1e94e46a1bc7/moto/ec2/models/subnets.py#L125:L155

And in the backend, in `describe_instance_types`, you can just call `generic_filter`:

https://github.com/spulec/moto/blob/78b89684157857d6f288de280d8f1e94e46a1bc7/moto/ec2/models/subnets.py#L358:L359

felipemarinho97:
Good suggestion! I've updated the code, the only problem is that I needed to transform the instance dictionary into an object

> A lot of the filtering logic already exists, and is part of the `moto.ec2.utils.generic_filter()` method.
> 
> A good example of this can be found in how Subnets are filtered by. The Subnet-class itself contains the `get_filter_value` method with logic on what filter-key corresponds to which value:
> 
> https://github.com/spulec/moto/blob/78b89684157857d6f288de280d8f1e94e46a1bc7/moto/ec2/models/subnets.py#L125
> 
> And in the backend, in `describe_instance_types`, you can just call `generic_filter`, which will take care of the rest (like how it's done in `SubnetBackend.get_all_subnets`).


Good suggestion! I've updated the code, the only catch is that I needed to transform the instance dictionary into an object to be able to implement `get_filter_value`

bblommers (approved):
LGTM! Thanks for adding this feature @felipemarinho97

PR #5393, opened 2022-08-15
Store and return ServerSideEncryption and KMS Key Id for S3 Multiparts

pinzon:
This PR addresses localstack/localstack#6604. Now the S3 Multipart stores the 'SSE' and `KmsKeyId` values and uses them when creating the key parts and the complete key.

### DIFF
--- a/tests/test_s3/test_s3_multipart.py
+++ b/tests/test_s3/test_s3_multipart.py
@@ -10,7 +10,7 @@ from io import BytesIO
 from moto.s3.responses import DEFAULT_REGION_NAME
 
 
-from moto import settings, mock_s3, mock_kms
+from moto import settings, mock_s3
 import moto.s3.models as s3model
 from moto.settings import get_s3_default_key_buffer_size, S3_UPLOAD_PART_MIN_SIZE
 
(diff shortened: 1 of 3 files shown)


tests/test_s3/test_s3_multipart.py
```diff
@@ -885,3 +885,45 @@ def test_complete_multipart_with_empty_partlist():
...
+    s3_client.create_bucket(Bucket=bucket_name)
+
+    kms_key = kms_client.create_key()["KeyMetadata"]
+    key_name = "test-file.txt"
+
+    create_multipart_response = s3_client.create_multipart_upload(
+        Bucket=bucket_name,
+        Key=key_name,
+        ServerSideEncryption="aws:kms",
+        SSEKMSKeyId=kms_key["KeyId"],
```
bblommers, line 906: Can we simplify this to `SSEKMSKeyId="some key"`, without the entire KMS integration? 
It would make the test easier to read, faster, and also make it clear that there is no validation yet

bblommers (changes requested):

tests/test_s3/test_s3_multipart.py
pinzon: done! :+1:

tests/test_s3/test_s3_multipart.py
```diff
@@ -10,7 +10,7 @@
 from moto.s3.responses import DEFAULT_REGION_NAME
 
 
-from moto import settings, mock_s3
+from moto import settings, mock_s3, mock_kms
```
bblommers, line 13: This is no longer necessary - I'm assuming that's why the linter fails

pinzon: Thanks for catching this.

bblommers (approved):
LGTM - thanks @pinzon!

PR #5394, opened 2022-08-16
align batch cancel_job and terminate_job

rav-evax:
I was testing AWS batch via a mock environment and identified some issues with `cancel_job`. Specifically, it raises a `AttributeError` when cancelling a non-existing job id. I would expect it to behave like `terminate_job`and raise a `ClientException("Job not found")`. Therefore, I have aligned `cancel_job` to behave like `terminate_job`.

### DIFF
--- a/moto/batch/models.py
+++ b/moto/batch/models.py
@@ -1585,7 +1585,14 @@ class BatchBackend(BaseBackend):
         return jobs
 
     def cancel_job(self, job_id, reason):
+        if job_id is None:
+            raise ClientException("Job ID does not exist")
+        if reason is None:
+            raise ClientException("Reason does not exist")
+
         job = self.get_job_by_id(job_id)
+        if job is None:
+            raise ClientException("Job not found")
         if job.status in ["SUBMITTED", "PENDING", "RUNNABLE"]:
             job.terminate(reason)
         # No-Op for jobs that have already started - user has to explicitly terminate those
--- a/moto/batch/models.py
+++ b/moto/batch/models.py
@@ -1585,11 +1585,15 @@ class BatchBackend(BaseBackend):
         return jobs
 
     def cancel_job(self, job_id, reason):
-        if job_id == '':
-            raise ClientException("'reason' is a required field (cannot be an empty string)")
-        if reason == '':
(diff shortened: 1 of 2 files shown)


moto/batch/models.py
```diff
@@ -1585,7 +1585,14 @@ def list_jobs(self, job_queue, job_status=None):
         return jobs
 
     def cancel_job(self, job_id, reason):
+        if job_id is None:
```
bblommers, line 1588: [Both parameters are required](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/batch.html#Batch.Client.cancel_job) when calling the SDK, so the first two checks shouldn't be necessary

moto/batch/models.py
```diff
@@ -1585,7 +1585,14 @@ def list_jobs(self, job_queue, job_status=None):
...
+            raise ClientException("Job ID does not exist")
+        if reason is None:
+            raise ClientException("Reason does not exist")
+
         job = self.get_job_by_id(job_id)
+        if job is None:
+            raise ClientException("Job not found")
```
bblommers, line 1595: I've just tested this against the AWS CLI with a random job-id, and it looks like it just returns an empty `200 OK` response. 
So it doesn't actually throw an error, it just always succeeds.

Copying that behaviour would still be a lot better than our current `AttributeError`. Would you mind adding a test case for this as well? The existing tests are in `tests/test_batch/test_jobs.py`, you can just add a small test that calls cancel_job with a random ID and assert that the response has a HTTP status code of 200.

bblommers (changes requested):

moto/batch/models.py
rav-evax: You are right, they not needed. I have looked further into the behavior of the SDK and if  you call it with empty strings it raises a `ClientException`. I have updated `cancel_job` and `temrinate_job` to mimic this behaviour.

moto/batch/models.py
rav-evax: No problem. I have added tests to accommodate the changes I have made to `cancel_job` and `terminate_job`.

bblommers (approved):
LGTM! Thanks for fixing this @rav-evax

PR #5397, opened 2022-08-16
fix apigateway rest API status codes

calvernaz:
Fixes response status codes according to https://docs.aws.amazon.com/apigateway/latest/api/API_Operations.html

/cc @bblommers the server mode tests are failing, however I don't see the connection with my PR


bblommers (approved):
Looks like the tests just needed to be restarted. Thanks for going through this and fixing it @calvernaz!

PR #5410, opened 2022-08-20
KMS: revoke grant raises NotFound

xmooner:
Revoke grant raises `NotFoundException` when the grant does not exist.
This aligns moto behavior with boto3/aws.

Also removed `test_revoke_grant_by_token` as it was just a copy of `test_retire_grant_by_token`.
There is no `revoke_grant_by_token` functionality in boto3's kms client. 
It must have have been a copy-paste leftover.

Issue [5411](https://github.com/spulec/moto/issues/5411)

### DIFF
--- a/tests/test_kms/test_kms_grants.py
+++ b/tests/test_kms/test_kms_grants.py
@@ -1,5 +1,6 @@
 import boto3
 import sure  # noqa # pylint: disable=unused-import
+import pytest
 
 from moto import mock_kms
 from moto.core import DEFAULT_ACCOUNT_ID as ACCOUNT_ID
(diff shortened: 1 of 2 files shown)


bblommers (approved):
LGTM - thanks for fixing this @xmooner, and welcome to Moto!

tests/test_kms/test_kms_grants.py
```diff
@@ -137,23 +138,18 @@ def test_revoke_grant():
 
 
 @mock_kms
-def test_revoke_grant_by_token():
-
+def test_revoke_grant_raises_when_grant_does_not_exist():
```
bpandola, line 141: Why are you modifying/removing an existing test here?  Why not just add a new one to cover your changes?

xmooner: As mentioned in the PR description. The `test_revoke_grant_by_token` code was a 100% duplicate of `test_retire_grant_by_token`. It had the same exact code, so it wasn't testing revoking a grant by grant id. 

That would have been actually impossible, because in the [KMS API](https://docs.aws.amazon.com/kms/latest/APIReference/API_RevokeGrant.html) you cannot revoke a grant by token, you can only retire a grant by token.
As a matter of fact, boto3 [revoke_grant](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/kms.html#KMS.Client.revoke_grant) has no functionality for revoking a grant by passing a token id.
So the test `test_revoke_grant_by_token` was:
 1. inaccurate, at least in the naming
 2. a duplicate of `test_retire_grant_by_token`, as the code inside the test was 100% the same.

Hence my decision to remove it.

xmooner:
> LGTM - thanks for fixing this @xmooner, and welcome to Moto!

Thank you @bblommers !

Issue #5354, opened 2022-08-02
Dynamodb incorrectly allows a mix of "expression" and "non-expression" parameters


paulmdavies:
# Versions

`moto`: 3.1.16
`boto3`: 1.21.46

# Code
```
table.get_item(
    Key={
        'key': 'the-key'
    },
    AttributesToGet=[
        'foo'
    ],
    ProjectionExpression='foo'
)
```

# Expectation

An error. This is what is produced by boto3 when calling this against a real table:
```
botocore.exceptions.ClientError: An error occurred (ValidationException) when calling the GetItem operation: Can not use both expression and non-expression parameters in the same request: Non-expression parameters: {AttributesToGet} Expression parameters: {ProjectionExpression}
```

# Reality

No error.

bblommers:
Thanks for raising this @paulmdavies - marking it as an enhancement to add this validation.

If this is something you want to implement yourself, PR's are always welcome! Let me know if you need help getting started.

PR #5412, opened 2022-08-20
Dynamodb - ensure getItem does not use both expression and non-expression

queue-tip:
Addresses #5354 by raising an exception.


bblommers (approved):
LGTM - thanks for the PR @queue-tip!

PR #5192, opened 2022-06-04
MultiAccount support

bblommers:
TODO:

 - [x] Documentation: https://moto--5192.org.readthedocs.build/en/5192/docs/multi_account.html
 - [x] Determine account ID based on provided aws_access_key
 - [x] Use account ID to select the correct backend
 - [x] Use account ID everywhere, instead of `get_account_id`
 - [x] Add tests for STS usecase
 - [x] Add tests for Environment Variable usecase
 - [x] change scaffold to use backend[account_id][region]
 - [x] Fix server tests - needs to pass an account ID
 - [x] Add more tests - verify that requests to different accounts are separated 
 - [x] fix ServerMode tests - needs to find this account ID 
 - [x] Update docs with the option to add a request-header
 - [x] Look into performance issues

PR #5414, opened 2022-08-22
Fix bug. route53 backend should be global

skinitimski:
There is [exactly one backend defined for route53](https://github.com/spulec/moto/blob/master/moto/route53/models.py#L815), and it is the "global" one:

```python
route53_backends = BackendDict(
    Route53Backend, "route53", use_boto3_regions=False, additional_regions=["global"]
)
```

But, this one method is using the given region, presumably the one of the containing CloudFormation stack?

I am pretty sure this was just an oversight in https://github.com/spulec/moto/pull/5192. The right region is used [15 or so lines above](https://github.com/spulec/moto/blob/master/moto/route53/models.py#L235):

```python
        backend = route53_backends[account_id]["global"]
```

### DIFF
--- a/moto/route53/models.py
+++ b/moto/route53/models.py
@@ -249,7 +249,7 @@ class RecordSet(CloudFormationModel):
 
     def delete(self, account_id, region):  # pylint: disable=unused-argument
         """Not exposed as part of the Route 53 API - used for CloudFormation"""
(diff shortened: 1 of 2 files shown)


bblommers (approved):
Good catch @skinitimski - that was indeed an oversight. Thanks for the fix and adding a test case for this!

PR #5416, opened 2022-08-24
add guard against no integration when adding a new response

calvernaz:
This PR adds a guard to cases where moto tries to add an integration response when there is no integration (yet).

The result is this exception.
```
File ".venv/lib/python3.10/site-packages/moto/apigateway/responses.py", line 497, in integration_responses
    integration_response = self.backend.put_integration_response(
  File ".venv/lib/python3.10/site-packages/moto/apigateway/models.py", line 1640, in put_integration_response
    integration_response = integration.create_integration_response(
AttributeError: 'NoneType' object has no attribute 'create_integration_response'
```

### DIFF
--- a/moto/apigateway/models.py
+++ b/moto/apigateway/models.py
@@ -1641,7 +1641,8 @@ class APIGatewayBackend(BaseBackend):
         response_templates,
         content_handling,
     ):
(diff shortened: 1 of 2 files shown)


bblommers (approved):
LGTM! Thanks for fixing this @calvernaz

PR #5431, opened 2022-08-29
MotoServer - ensure HTTP_HOST-header is optional

### LINKED ISSUE #2539: Wrong selection of backend when running moto server on EC2
I'm running moto on an EC2 VM. I got another system running also as an EC2 VM  which calls moto. when trying to create a new stack (CloudFormation) i'm getting a NotImplemented Error.

After plenty of debugging - i found that moto actually infers the backend from the host name.
Reference: moto/server.py line: 121

Is there a reason for this? I'm not really sure this is true to the normal behavior of AWS

Suggestion: completely remove this check. i'd be happy to do create a PR if that's agreed.

bblommers:
Fixes #2539

### DIFF
--- a/moto/moto_server/werkzeug_app.py
+++ b/moto/moto_server/werkzeug_app.py
@@ -167,7 +167,7 @@ class DomainDispatcherApplication(object):
         elif path_info.startswith("/latest/meta-data/"):
             host = "instance_metadata"
         else:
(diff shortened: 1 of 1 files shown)


bblommers:
Closing - this doesn't actually fix anything.

PR #5435, opened 2022-08-30
fix(acm): use .utcnow() instead of .now()

blimmer:
The `cryptography` library returns information about certificates using the UTC timezone (e.g., for the [`not_valid_before` property](https://cryptography.io/en/latest/x509/reference/?highlight=not_before#cryptography.x509.Certificate.not_valid_before)). However, `moto` uses `datetime.now()`, which returns the time for the user's current timezone.

I generated a self-signed certificate for testing some ACM behaviors, and I encountered this error:

> botocore.errorfactory.ValidationException: An error occurred (ValidationException) when calling the ImportCertificate operation: The certificate chain is not in effect yet, is not valid.

I tracked that error down to this code in `moto`: https://github.com/spulec/moto/blob/2c98b7574b82f803b6f333eeaf21653856603a98/moto/acm/models.py#L309-L318

In my case, `now` was `2022-08-30 13:23:36.363011` and the certificate `not_valid_before` value was `2022-08-30 19:17:00`, causing the error.

Because `cryptography` reports these values in UTC, we should use `.utcnow()` instead of `.now()`. As part of this PR, I made all calls to `.now()` use `.utcnow()`. I think that should be safe and will help avoid these errors throughout.

## Workaround

In CI, your machines are probably already using UTC as the timezone. However, this issue usually happens when running locally. To be safe, you can set your global Python timezone to UTC like this:

```py
os.environ["TZ"] = "Etc/UTC"
```

### DIFF
--- a/moto/acm/models.py
+++ b/moto/acm/models.py
@@ -130,7 +130,7 @@ class CertBundle(BaseModel):
         cert_type="IMPORTED",
         cert_status="ISSUED",
     ):
-        self.created_at = datetime.datetime.now()
+        self.created_at = datetime.datetime.utcnow()
         self.cert = certificate
         self._cert = None
         self.common_name = None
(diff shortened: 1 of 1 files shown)


bblommers (approved):
LGTM - thanks for fixing this @blimmer!

PR #5445, opened 2022-09-02
Add support for Serverless Endpoint Configurations

### LINKED ISSUE #5443: create_endpoint_config doesn't support Serverless deployment
It is currently not possible to use moto to mock `create_endpoint_config` if the endpoint config involves using serverless deployment.

To reproduce:
```
from moto import mock_sagemaker
import boto3

with mock_sagemaker():
    client = boto3.client("sagemaker")
    endpoint_conf_name = 'test'
    
    client.create_endpoint_config(
        EndpointConfigName=endpoint_conf_name,
        ProductionVariants=[
            {'VariantName': 'dummy',
                'ModelName': 'dummy',
                'ServerlessConfig': {
                'MemorySizeInMB': 1024,
                'MaxConcurrency': 20
            }}
        ],
    )
```

Stacktrace:
```
self = <moto.sagemaker.models.FakeEndpointConfig object at 0x7f90f8934850>, production_variants = [{'ModelName': 'test', 'ServerlessConfig': {'MaxConcurrency': 20, 'MemorySizeInMB': 6144}, 'VariantName': 'default-variant-test'}]

    def validate_production_variants(self, production_variants):
        for production_variant in production_variants:
>           self.validate_instance_type(production_variant["InstanceType"])
E           KeyError: 'InstanceType'

ml_framework/venv/lib/python3.7/site-packages/moto/sagemaker/models.py:401: KeyError
```

The error is triggered when moto creates a new `FakeEndpointConfig`: https://github.com/spulec/moto/blob/930c58bd1366b073e8630a82fa7f34624babd867/moto/sagemaker/models.py#L399-L401

When deploying using `ServerlessConfig`, there is no `InstanceType` present in the ProductionVariant (see documentation: https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/sagemaker.html#SageMaker.Client.create_endpoint_config)

Ideally, we should condition based on whether we have `InstanceType` or `ServerlessConfig` in the ProductionVariant, and add some checks on the Serverless Config parameters.

Happy to get a PR out to fix this :)

arnaudstiegler:
Closes #5443 

Adds support for Serverless Config in ProductionVariants, and updates tests to cover this use case

### DIFF
--- a/tests/test_sagemaker/test_sagemaker_endpoint.py
+++ b/tests/test_sagemaker/test_sagemaker_endpoint.py
@@ -47,7 +47,6 @@ def sagemaker_client():
     return boto3.client("sagemaker", region_name=TEST_REGION_NAME)
 
 
-@mock_sagemaker
 def create_endpoint_config_helper(sagemaker_client, production_variants):
     _create_model(sagemaker_client, TEST_MODEL_NAME)
 
--- a/tests/test_sagemaker/test_sagemaker_endpoint.py
+++ b/tests/test_sagemaker/test_sagemaker_endpoint.py
@@ -20,12 +20,24 @@ TEST_ENDPOINT_NAME = "MyEndpoint"
 TEST_ENDPOINT_CONFIG_NAME = "MyEndpointConfig"
 TEST_VARIANT_NAME = "MyProductionVariant"
 TEST_INSTANCE_TYPE = "ml.t2.medium"
+TEST_MEMORY_SIZE = 1024
+TEST_CONCURRENCY = 10
 TEST_PRODUCTION_VARIANTS = [
     {
         "VariantName": TEST_VARIANT_NAME,
         "ModelName": TEST_MODEL_NAME,
         "InitialInstanceCount": 1,
         "InstanceType": TEST_INSTANCE_TYPE,
+    },
+]
+TEST_SERVERLESS_PRODUCTION_VARIANTS = [
+    {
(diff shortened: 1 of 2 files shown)


tests/test_sagemaker/test_sagemaker_endpoint.py
```diff
@@ -36,18 +48,18 @@ def sagemaker_client():
 
 
 @mock_sagemaker
-def test_create_endpoint_config(sagemaker_client):
+def create_endpoint_config(sagemaker_client, production_variants):
```
arnaudstiegler, line 51: Not sure whether we need the moto decorator for that helper function

arnaudstiegler:
Not sure about the coverage test, should I be updating something? I did revert some changes to coverage.xml that were automatically done by one of the `make ...` command

tests/test_sagemaker/test_sagemaker_endpoint.py
bblommers: It shouldn't be necessary on the helper-function. From what I can see, the decorator are active on all the tests - so it will still be active when this method is invoked.
If you can just remove that, everything else looks good to me!

arnaudstiegler: Removed the decorator!

bblommers (approved):
Awesome - thank you for fixing this @arnaudstiegler!

Issue #5451, opened 2022-09-07
REFRESH_TOKEN_AUTH is not working


daka83:
When calling `initiate_auth` with `REFRESH_TOKEN_AUTH` I am getting next exception

```
Traceback (most recent call last):
  File "/usr/local/lib/python3.8/site-packages/werkzeug/serving.py", line 335, in run_wsgi
    execute(self.server.app)
  File "/usr/local/lib/python3.8/site-packages/werkzeug/serving.py", line 322, in execute
    application_iter = app(environ, start_response)
  File "/usr/local/lib/python3.8/site-packages/moto/moto_server/werkzeug_app.py", line 241, in __call__
    return backend_app(environ, start_response)
  File "/usr/local/lib/python3.8/site-packages/flask/app.py", line 2091, in __call__
    return self.wsgi_app(environ, start_response)
  File "/usr/local/lib/python3.8/site-packages/flask/app.py", line 2076, in wsgi_app
    response = self.handle_exception(e)
  File "/usr/local/lib/python3.8/site-packages/flask_cors/extension.py", line 165, in wrapped_function
    return cors_after_request(app.make_response(f(*args, **kwargs)))
  File "/usr/local/lib/python3.8/site-packages/flask/app.py", line 2073, in wsgi_app
    response = self.full_dispatch_request()
  File "/usr/local/lib/python3.8/site-packages/flask/app.py", line 1519, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "/usr/local/lib/python3.8/site-packages/flask_cors/extension.py", line 165, in wrapped_function
    return cors_after_request(app.make_response(f(*args, **kwargs)))
  File "/usr/local/lib/python3.8/site-packages/flask/app.py", line 1517, in full_dispatch_request
    rv = self.dispatch_request()
  File "/usr/local/lib/python3.8/site-packages/flask/app.py", line 1503, in dispatch_request
    return self.ensure_sync(self.view_functions[rule.endpoint])(**req.view_args)
  File "/usr/local/lib/python3.8/site-packages/moto/core/utils.py", line 129, in __call__
    result = self.callback(request, request.url, dict(request.headers))
  File "/usr/local/lib/python3.8/site-packages/moto/core/responses.py", line 217, in dispatch
    return cls()._dispatch(*args, **kwargs)
  File "/usr/local/lib/python3.8/site-packages/moto/core/responses.py", line 356, in _dispatch
    return self.call_action()
[...]

PR #5453, opened 2022-09-07
fixed initiate_auth when refresing token

daka83:


daka83:
Fixes https://github.com/spulec/moto/issues/5451

bblommers (approved):
Nice catch. Thanks for the fix @daka83!

PR #5459, opened 2022-09-09
Test backward-incompatible changes in `openapi-spec-validator` package

bpandola:
Changes introduced after version 0.4.0

Ref: https://github.com/p1c2u/openapi-spec-validator/commit/70325f638001c615909978381bb48ad3d924dafd


bpandola:
Was just trying to test this new workflow because the original PR (#5458) was running under the old workflow for some reason...

PR #5467, opened 2022-09-13
Bug distribution with tags

eloymg:
Solve a bug with distribution with tags error if only exists a one tag and update coverage document adding create_distribution_with_tags


bblommers (approved):
LGTM - thanks for fixing this @eloymg!

PR #5470, opened 2022-09-13
add rds enable_http_endpoint attribute

marshall7m:
The current implementation of mocking the  RDS `CreateDBCluster` sets HttpEndpointEnabled to False by default without regard to an enable_http_endpoint parameter value. 

The PR introduces a `self.enable_http_endpoint` attribute within the `Cluster` class and simulates the AWS API's conditional constraints for enabling HTTP endpoints within serverless V1 clusters. Specifically, the conditions relate to the desired engine, engine version, and engine_mode configurations for the cluster to be created.

### DIFF
--- a/moto/rds/models.py
+++ b/moto/rds/models.py
@@ -114,6 +114,17 @@ class Cluster:
         self.enabled_cloudwatch_logs_exports = (
             kwargs.get("enable_cloudwatch_logs_exports") or []
         )
+        self.enable_http_endpoint = False
+        # instead of raising an error on aws rds create-db-cluster commands with
+        # incompatible configurations with enable_http_endpoint 
+        # (e.g. engine_mode is not set to "serverless"), the API
+        # automatically sets the enable_http_endpoint parameter to False
+        if kwargs.get("enable_http_endpoint"):
+            if self.engine_mode == "serverless":
+                if self.engine == "aurora-sql" and self.engine_version in ["5.6.10a", "5.6.1", "2.07.1", "5.7.2"]:
+                    self.enable_http_endpoint = kwargs.get("enable_http_endpoint", False)
+                elif self.engine == "aurora-postgresql" and self.engine_version in ["10.12", "10.14", "10.18", "11.13"]:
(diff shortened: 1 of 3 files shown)


bblommers (changes requested):
Hi @marshall7m, thanks for the PR! Would you mind adding a few tests to verify this works correctly?

The existing tests reside in `tests/test_rds`, so you should be able to just extend those. Let me know if you need any help with that.

marshall7m:
> Hi @marshall7m, thanks for the PR! Would you mind adding a few tests to verify this works correctly?
> 
> The existing tests reside in `tests/test_rds`, so you should be able to just extend those. Let me know if you need any help with that.

Hi @bblommers, thanks for the quick response. I added the appropriate tests. Let me know if the input Engine, EngineMode, and EngineVersion should be parametrized within the tests for more coverage or if the current input is sufficient. I'll be glad to make the changes if needed.

moto/rds/models.py
```diff
@@ -1740,6 +1751,7 @@ def modify_db_parameter_group(
     def create_db_cluster(self, kwargs):
         cluster_id = kwargs["db_cluster_identifier"]
         kwargs["account_id"] = self.account_id
+        print(kwargs)
```
bblommers, line 1754: Please remove the print-statement

bblommers (changes requested):
Looks like the linter is failing @marshall7m - if we can remove the print-statement and get the build to pass I'd be happy to merge

moto/rds/models.py
marshall7m: ooops, thanks!

bblommers:
The linter is still failing @marshall7m - see the CI here: https://github.com/spulec/moto/actions/runs/3078428258

You can run `make lint` locally to check as well.

bblommers (approved):
Nice one - thanks again for fixing this @marshall7m!

marshall7m:
> Nice one - thanks again for fixing this @marshall7m!

@bblommers No problem, thanks for the guidance along the way!

PR #5477, opened 2022-09-16
Add CloudFormation support to EC2 Launch Templates

skinitimski:
## Purpose

* Implement CloudFormation CRUD model for EC2 LaunchTemplate resources.
* Fix a bug (discovered while writing unit tests) in `describe-launch-template-versions`: an error should be raised when the specified template does not exist:

> aws ec2 describe-launch-template-versions --launch-template-name timski
> An error occurred (InvalidLaunchTemplateName.NotFoundException) when calling the DescribeLaunchTemplateVersions operation: The specified launch template, with template name timski, does not exist.

### DIFF
--- a/moto/ec2/exceptions.py
+++ b/moto/ec2/exceptions.py
@@ -718,6 +718,14 @@ class InvalidLaunchTemplateNameNotFoundError(EC2ClientError):
         )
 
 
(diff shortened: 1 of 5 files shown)


bblommers (approved):
LGTM! Thanks for adding this feature @skinitimski

Issue #5479, opened 2022-09-16
When trying to create the subnet group get error: Error listing tags for RDS DB Subnet Group


szopen321:
**Repro steps:**

1. Start moto server.
2. Prepare terraform file that will create rds db subnet group.
3. Run terraform init; terraform apply

**Description:**
When attempting to create a subnet group terraform errors out with error: "Error listing tags for RDS DB Subnet Group: InvalidParameterValue: Invalid resource name:"

**Observed results:**
An error on attempted creation of rds db subnet group

**expected results:**
It is possible to create an rds db subnet group without issues

szopen321:
PR with proposed solution https://github.com/spulec/moto/pull/5480

PR #5480, opened 2022-09-16
Fix error listing tags for rds db subnet group

szopen321:
Suggested solution to https://github.com/spulec/moto/issues/5479

### DIFF
--- a/moto/rds/models.py
+++ b/moto/rds/models.py
@@ -1492,8 +1492,12 @@ class RDSBackend(BaseBackend):
         security_group.authorize_cidr(cidr_ip)
         return security_group
 
-    def create_subnet_group(self, subnet_name, description, subnets, tags, region, account_id):
-        subnet_group = SubnetGroup(subnet_name, description, subnets, tags, region, account_id)
+    def create_subnet_group(
+        self, subnet_name, description, subnets, tags, region, account_id
+    ):
+        subnet_group = SubnetGroup(
+            subnet_name, description, subnets, tags, region, account_id
+        )
         self.subnet_groups[subnet_name] = subnet_group
         return subnet_group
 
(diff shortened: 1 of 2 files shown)


moto/rds/models.py
```diff
@@ -1485,8 +1492,12 @@ def authorize_security_group(self, security_group_name, cidr_ip):
         security_group.authorize_cidr(cidr_ip)
         return security_group
 
-    def create_subnet_group(self, subnet_name, description, subnets, tags):
-        subnet_group = SubnetGroup(subnet_name, description, subnets, tags)
+    def create_subnet_group(
+        self, subnet_name, description, subnets, tags, region, account_id
```
bblommers, line 1496: The region and account are already available in the backend, so we don't have to pass them explicitly.
`self.region_name` and `self.account_id` out the top of my head

bblommers (changes requested):
Would you mind adding a test for this @szopen321? As it stands I don't immediately see what error it would fix

Edit: I just realized you made it clear what's wrong in the original issue. Verification that this value is now returned properly would still be nice though :)

szopen321:
Oh, yes, my apologies! I've been meaning to mention aws docs to make the changes clearer.

This docs have information about arn missing:
https://docs.aws.amazon.com/AmazonRDS/latest/APIReference/API_DescribeDBSubnetGroups.html

While below it is included:
https://docs.aws.amazon.com/cli/latest/reference/rds/describe-db-subnet-groups.html
https://docs.aws.amazon.com/AmazonRDS/latest/APIReference/API_DBSubnetGroup.html

I used their feedback form in hopes they will complete the docs.

bblommers (approved):
Nice one - thanks for the fix @szopen321!

PR #5492, opened 2022-09-25
Feature: Ability to seed Moto, and make random ID's deterministic

### LINKED ISSUE #5299: Add possibility to record and replay state of moto

bblommers:
Follow-up to #5481 

Moto now uses a single source of random-ness, to simplify the process of making ID's deterministic. (Hence the size of this PR..)

WIP documentation:
https://moto--5492.org.readthedocs.build/en/5492/docs/configuration/recorder/index.html#deterministic-identifiers

An alternative API, in pure Python-mode,  could look something like this:
```
from moto import random
random.seed(x)
```
This would eliminate the need for a HTTP-request.
However, it would have made the implementation much more difficult to ensure it plays nice with the Recorder, having to record and replay API calls in addition to HTTP requests

Closes #5299 
Supercedes and closes #5300

### DIFF
--- a/moto/moto_api/_internal/moto_random.py
+++ b/moto/moto_api/_internal/moto_random.py
@@ -7,26 +7,24 @@ HEX_CHARS = list(range(10)) + ["a", "b", "c", "d", "e", "f"]
 
 
 class MotoRandom(Random):
-    def __init__(self):
-        self._rnd = Random()
-
-    def __getattribute__(self, name):
-        if name in ["_rnd", "uuid1", "uuid4", "get_random_hex", "get_random_string"]:
-            return object.__getattribute__(self, name)
-        return object.__getattribute__(self._rnd, name)
+    """
+    Class used for all sources of random-ness in Moto.
+    Used as a singleton, which is exposed in `moto/moto_api/_internal`.
+    This Singleton can be seeded to make identifiers deterministic.
+    """
 
     def uuid1(self):
-        return uuid.UUID(int=self._rnd.getrandbits(128), version=1)
+        return uuid.UUID(int=self.getrandbits(128), version=1)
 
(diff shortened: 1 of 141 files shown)


moto/moto_api/_internal/moto_random.py
```diff
@@ -0,0 +1,32 @@
+from random import Random
+import string
+import uuid
+
+
+HEX_CHARS = list(range(10)) + ["a", "b", "c", "d", "e", "f"]
+
+
+class MotoRandom(Random):
```
bblommers, line 9: There should be a docstring with the rationale + usecase for this class

moto/moto_api/_internal/moto_random.py
```diff
@@ -0,0 +1,32 @@
...
+    def __init__(self):
+        self._rnd = Random()
```
bblommers, line 11: This (and __getattribute__) should not be necessary, as we already extend Random - just call `super()` if necessary

borlafu:
This breaks custom extensions (service implementations) that make use of `amzn_request_id`, since it was moved to another module.

It's not the end of the world, but something I was surprised to see in a minor version update...

bblommers:
@borlafu `amzn_request_id` is supposed to be an internal API, that's why we didn't feel the need for a major release.

Just out of interest, what kind of custom extensions are you using, that use/extend this decorator?

borlafu:
It's a (partial) FSX implementation, that I actually planned to contribute back to moto. I guess I'm not the only one in that situation, but I understand the blast radius is minimal. As a general practice I like putting internal APIs under an "internal" or "private" package, or at least adding a comment to be explicit about it.

PR #5495, opened 2022-09-26
Add expression param to existing get_tables glue api

rafcio19:
Currently [boto3 implements](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/glue.html#Glue.Client.get_tables) `glue.get_tables` with additional `Expression` parameter:

> Expression (string) -- A regular expression pattern. If present, only those tables whose names match the pattern are returned.

This is the implementation.

### DIFF
--- a/moto/glue/models.py
+++ b/moto/glue/models.py
@@ -139,11 +139,14 @@ class GlueBackend(BaseBackend):
 
     def get_tables(self, database_name, expression):
         database = self.get_database(database_name)
-        return [
-            table
-            for table_name, table in database.tables.items()
-            if bool(re.match(expression, table_name))
-        ]
+        if expression:
+            return [
+                table
+                for table_name, table in database.tables.items()
+                if re.match(expression, table_name)
+            ]
+        else:
+            return [table for table_name, table in database.tables.items()]
 
     def delete_table(self, database_name, table_name):
         database = self.get_database(database_name)
--- a/moto/glue/models.py
+++ b/moto/glue/models.py
@@ -137,7 +137,7 @@ class GlueBackend(BaseBackend):
         except KeyError:
             raise TableNotFoundException(table_name)
 
-    def get_tables(self, database_name, expression=""):
+    def get_tables(self, database_name, expression):
         database = self.get_database(database_name)
         return [
             table
--- a/moto/glue/models.py
+++ b/moto/glue/models.py
@@ -1,6 +1,7 @@
 import time
 from collections import OrderedDict
 from datetime import datetime
+import re
 from typing import List
 from uuid import uuid4
 
@@ -136,9 +137,13 @@ class GlueBackend(BaseBackend):
         except KeyError:
             raise TableNotFoundException(table_name)
 
-    def get_tables(self, database_name):
+    def get_tables(self, database_name, expression=""):
         database = self.get_database(database_name)
-        return [table for table_name, table in database.tables.items()]
+        return [
+            table
+            for table_name, table in database.tables.items()
+            if bool(re.match(expression, table_name))
+        ]
 
     def delete_table(self, database_name, table_name):
         database = self.get_database(database_name)
--- a/tests/test_glue/helpers.py
+++ b/tests/test_glue/helpers.py
@@ -66,8 +66,11 @@ def get_table(client, database_name, table_name):
     return client.get_table(DatabaseName=database_name, Name=table_name)
 
 
-def get_tables(client, database_name, expression=""):
-    return client.get_tables(DatabaseName=database_name, Expression=expression)
+def get_tables(client, database_name, expression=None):
+    if expression:
+        return client.get_tables(DatabaseName=database_name, Expression=expression)
+    else:
+        return client.get_tables(DatabaseName=database_name)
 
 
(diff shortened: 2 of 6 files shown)


rafcio19:
Hi @bblommers, thanks for approving the workflows. However, some of the TestTerraform ones fail and they look unrelated to my changes 🤷 
All I see in the logs is `Error:`
Are you able to see more?

bblommers:
Don't worry about the Terraform tests @rafcio19 - their 1.3.0 release broke quite a few things. Our tests should be fine again in 1.3.1, so we can ignore those failures for now

tests/test_glue/helpers.py
```diff
@@ -66,8 +66,8 @@ def get_table(client, database_name, table_name):
     return client.get_table(DatabaseName=database_name, Name=table_name)
 
 
-def get_tables(client, database_name):
-    return client.get_tables(DatabaseName=database_name)
+def get_tables(client, database_name, expression=""):
+    return client.get_tables(DatabaseName=database_name, Expression=expression)
```
bblommers, line 70:
```suggestion
def get_tables(client, database_name, expression=None):
    if expression:
        return client.get_tables(DatabaseName=database_name, Expression=expression)
    else:
        return client.get_tables(DatabaseName=database_name)
```
Most of the tests do not provide an expression - and we should verify that it still works, even if an expression is not provided.

You may want to test this out locally, I expect some tests to fail with this change

moto/glue/models.py
```diff
@@ -136,9 +137,13 @@ def get_table(self, database_name, table_name):
...
             raise TableNotFoundException(table_name)
 
-    def get_tables(self, database_name):
+    def get_tables(self, database_name, expression=""):
         database = self.get_database(database_name)
-        return [table for table_name, table in database.tables.items()]
+        return [
+            table
+            for table_name, table in database.tables.items()
+            if bool(re.match(expression, table_name))
```
bblommers, line 145: With the earlier change, the expression will sometimes be None. So we should make sure that we handle that appropriately.

Also: I don't think the `bool()` conversion is necessary here. `re.match()` will return a None-object if there is no match, so we can test against that

moto/glue/models.py
```diff
@@ -136,9 +137,13 @@ def get_table(self, database_name, table_name):
         except KeyError:
             raise TableNotFoundException(table_name)
 
-    def get_tables(self, database_name):
+    def get_tables(self, database_name, expression=""):
```
bblommers, line 140:
```suggestion
    def get_tables(self, database_name, expression):
```
The `expression`-parameter is always provided by `responses.py` - we just have to ensure we can deal with a None-value

bblommers (changes requested):

bblommers (approved):
Awesome - thanks for fixing this @rafcio19!

robert-schmidtke:
I am not sure this is the correct implementation. Glue's `get_tables` does accept a simple wildcard, e.g. `*`:

```python
>>> tables = glue.get_tables(DatabaseName="my_test_database", Expression="*")
>>> len(tables)
2
```

This does not work with the Python `re` backed implementation here.

rafcio19:
Do you have examples @robert-schmidtke? I've only been able to verify `*` as a wildcard:

```
    import boto3
    glue = boto3.client('glue')
    glue.create_database(DatabaseInput={"Name": 'test'})
    glue.create_table(TableInput={"Name": "test123_table"}, DatabaseName='test')
    o = glue.get_tables(DatabaseName='test', Expression='test123_table')
    print([x['Name'] for x in o['TableList']])

    $ ['test123_table']

    o = glue.get_tables(DatabaseName='test', Expression='test1*table')
    print([x['Name'] for x in o['TableList']])

    $ ['test123_table']
```

but others like`?` for single character matching don't work:

```
    o = glue.get_tables(DatabaseName='test', Expression='test12?_table')
    print([x['Name'] for x in o['TableList']])

    $ []
```

robert-schmidtke:
Hey @rafcio19, thanks for taking the time to produce examples here. I think I can only make `?` work as part of a normal regular expression, not a glob pattern. You're right on this one. So maybe the `*` wildcard is indeed special. So maybe the only idea for a change that I have is, if the `Expression` `startswith` `*`, prepend a `.`? What do you think?

robert-schmidtke:
I actually found a case where it does not work:

```python
# no dot before last asterisk works
>>> o = glue.get_tables(DatabaseName='test', Expression='.*table*')
>>> print([x['Name'] for x in o['TableList']])
['test123_table']

# dot before last asterisk does not work
>>> o = glue.get_tables(DatabaseName='test', Expression='.*table.*')
>>> print([x['Name'] for x in o['TableList']])
[]
```

rafcio19:
Hmm.. that last call `'.*table.*'` looks like a glue bug since the regex is valid. Prepending `*` with a `.`, if `*` is not the last character in the expression seems like a pragmatic way to tackle this, rather than trying to divine all the possible edge cases.

robert-schmidtke:
Would you be willing to open another PR for this? I would much appreciate it for sure :)

bblommers:
Thanks for raising this and investigating the AWS behaviour @robert-schmidtke!

PR #5497, opened 2022-09-26
setting default FeatureSet to ALL in create_organization

ackdav:
### Issue
Within the service `Organizations` for the operation `create_organization`, the docs ([doc1](https://docs.aws.amazon.com/cli/latest/reference/organizations/create-organization.html) and [doc2](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/organizations.html#Organizations.Client.create_organization)) state the following: 
> By default (or if you set the FeatureSet parameter to ALL ), the new organization is created with all features enabled and service control policies automatically enabled in the root.  

As of now, not setting the FeatureSet, and therefore relying on the default value, results in a KeyError.  

### Summary

This PR solves this issue by setting "ALL" as the default if none is present.

### DIFF
--- a/moto/organizations/models.py
+++ b/moto/organizations/models.py
@@ -355,7 +355,7 @@ class OrganizationsBackend(BaseBackend):
         return root
 
     def create_organization(self, **kwargs):
-        self.org = FakeOrganization(self.account_id, kwargs["FeatureSet"])
+        self.org = FakeOrganization(self.account_id, kwargs.get("FeatureSet") or "ALL")
         root_ou = FakeRoot(self.org)
         self.ou.append(root_ou)
         master_account = FakeAccount(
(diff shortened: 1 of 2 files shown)


bblommers:
Hi @ackdav, thanks for the PR. Would you mind adding a test case for this, to verify it works as expected?
Existign tests can be found here: https://github.com/spulec/moto/blob/master/tests/test_organizations/test_organizations_boto3.py

ackdav:
Hello @bblommers - sure, I added a little test. Wasn't quite sure how to test the proposed changes in isolation... Happy to adjust it..

By the way, `make lint` results in countless warnings and ends in an Error. Is that just on my machine, or is that currently expected?

bblommers:
> 
> By the way, `make lint` results in countless warnings and ends in an Error. Is that just on my machine, or is that currently expected?

That is probably an issue with your setup, maybe an issue with the versions of black/flake8/pylint that you have installed?
The linter passes in our CI: see https://github.com/spulec/moto/actions/runs/3133548981/jobs/5099910561

bblommers (approved):
LGTM! Thanks again for fixing this @ackdav

PR #5501, opened 2022-09-28
Make assumed role user ids dependent on role

dfangl:
## Problem description
Currently, when assuming roles, the User Id returned (and also returned by sts get-caller-identity) is not identical to the supposed format `role_id:session_name`, but regenerated freshly every time.

## Solution in this PR
* if a matching role exists, it will take the role id of that role. If not, the behavior will be unchanged (to prevent breaking behavior)
* Both access-key-ids and role ids are now, in accordance with AWS behavior, based on the account id. Basically, the first 9 characters after the 4 char prefix are (more or less) dependent on the account id. This generation is based on the blog post https://awsteele.com/blog/2020/09/26/aws-access-key-format.html, and is verified using `aws sts get-access-key-info --access-key-id ...`. The code is basically shifting the account id by an offset, and base32 encoding the big endian byte representation, with the next char after the 8 digit base32 being dependent on the LSB of the account id.

### DIFF
--- a/moto/iam/models.py
+++ b/moto/iam/models.py
@@ -49,6 +49,8 @@ from .utils import (
     random_alphanumeric,
     random_resource_id,
     random_policy_id,
+    random_role_id,
+    generate_access_key_id_from_account_id,
 )
 from ..utilities.tagging_service import TaggingService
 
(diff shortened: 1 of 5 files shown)


dfangl:
I cannot really reproduce the one failing test, and it only happens in 3.9, so might be a flake?

bblommers:
> I cannot really reproduce the one failing test, and it only happens in 3.9, so might be a flake?

Yes, it was - restarting it fixed it. Thank you for the PR @dfangl!

bblommers (approved):

PR #5517, opened 2022-10-02
pin terraform version to 1.2.9 in terraform tests

thrau:
This PR is a workaround for the currently failing terraform tests.

See https://github.com/localstack/localstack/pull/6962 for an explanation.


bblommers (approved):
Nice one - thanks for fixing this @thrau

Issue #3688, opened 2021-02-14
ECS task with no ENI attachments


ltagliamonte:
It looks like that run_task and describe_tasks using `awsvpc` mode do not attach an ENI to the task.
the ENI info should be present under `attachments` -> `details` with value name `networkInterfaceId`
https://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_Attachment.html

bblommers:
Thanks for raising this @ltagliamonte - marking it as an enhancement!

Balasundaram:
Hi @spulec @bblommers I am also interested in this as I am trying to mock run_task with networkconfiguration. Can this is prioritized ?

bblommers:
Hi @Balasundaram, most of our features are actually contributed by the community, not necessarily by us.

So if this is something you really need, a PR is always welcome!
There is a [Contributing-section in our docs](http://docs.getmoto.org/en/latest/docs/contributing/index.html) - happy to help if you have any questions around this.

PR #5520, opened 2022-10-03
Issue 3688, add eni to an ecs task

rafcio19:
Resolves https://github.com/spulec/moto/issues/3688

### DIFF
--- a/moto/ecs/models.py
+++ b/moto/ecs/models.py
@@ -333,7 +333,7 @@ class Task(BaseObject):
 
             self.network_configuration = networking_configuration
             net_conf = networking_configuration["awsvpcConfiguration"]
-            ec2_backend = ec2_backends["123456789012"]["us-east-1"]
+            ec2_backend = ec2_backends[self._account_id][self.region_name]
 
             eni = ec2_backend.create_network_interface(
                 subnet=net_conf["subnets"][0],
@@ -346,7 +346,7 @@ class Task(BaseObject):
 
             self.attachments.append(
                 {
-                    "id": str(uuid.uuid4()),
+                    "id": str(mock_random.uuid4()),
                     "type": "ElasticNetworkInterface",
                     "status": "ATTACHED",
                     "details": [
--- a/moto/ecs/models.py
+++ b/moto/ecs/models.py
@@ -3,6 +3,7 @@ from copy import copy
 from datetime import datetime
 
 import pytz
+import uuid
 
 from moto import settings
 from moto.core import BaseBackend, BaseModel, CloudFormationModel
@@ -13,6 +14,8 @@ from moto.core.utils import (
     remap_nested_keys,
     BackendDict,
 )
+
+from ..ec2.utils import random_eni_attach_id, random_resource_id
 from moto.ec2 import ec2_backends
 from moto.moto_api._internal import mock_random
 from moto.utilities.tagging_service import TaggingService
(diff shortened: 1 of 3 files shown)


moto/ecs/models.py
```diff
@@ -318,6 +322,42 @@ def __init__(
...
+                private_ip_address=random_private_ip(),
+                group_ids=net_conf["securityGroups"],
+                description="moto ECS",
+            )
+            eni.status = "in-use"
+            eni.device_index = 0
+
+            self.attachments.append(
+                {
+                    "id": str(uuid.uuid4()),
```
bblommers, line 349:
```suggestion
                    "id": str(mock_random.uuid4()),
```
The `mock_random`-object was introduced a week ago or so, and gives us the option to make random identifiers (such as uuid's) deterministic; see http://docs.getmoto.org/en/latest/docs/configuration/recorder/index.html#deterministic-identifiers
Seeding Moto only works if all our identifiers are generated by the same `Random`-object, which is what happens within the `MockRandom`-object

moto/ecs/models.py
```diff
@@ -318,6 +322,42 @@ def __init__(
...
+        if task_definition.network_mode == "awsvpc":
+            if not networking_configuration:
+
+                raise InvalidParameterException(
+                    "Network Configuration must be provided when networkMode 'awsvpc' is specified."
+                )
+
+            self.network_configuration = networking_configuration
+            net_conf = networking_configuration["awsvpcConfiguration"]
+            ec2_backend = ec2_backends["123456789012"]["us-east-1"]
```
bblommers, line 336:
```suggestion
            ec2_backend = ec2_backends[self._account_id][self.region_name]
```

bblommers: This should work for all accounts/regions

tests/test_ecs/test_ecs_boto3.py
```diff
@@ -1705,6 +1705,109 @@ def test_run_task():
...
+
+    # Setup
+    client = boto3.client("ecs", region_name="us-east-1")
+    ec2 = boto3.resource("ec2", region_name="us-east-1")
+
+    # ECS setup
+    setup_ecs(client, ec2)
+
+    # Execute
+    with pytest.raises(ClientError):
```
bblommers, line 1801: Would you mind adding some assertions here, to verify that this throws the correct error?
```
    ...
err["Code"].equals("?")
err["Message"].equals("?")
```

tests/test_ecs/test_ecs_boto3.py
```diff
@@ -1705,6 +1705,109 @@ def test_run_task():
...
+        UUID(response["tasks"][0]["attachments"][0]["id"])
+    except ValueError:
+        assert False
+    response["tasks"][0]["attachments"][0]["status"].should.equal("ATTACHED")
+    response["tasks"][0]["attachments"][0]["type"].should.equal(
+        "ElasticNetworkInterface"
+    )
+    details = response["tasks"][0]["attachments"][0]["details"]
+
+    subnet_found = False
```
bblommers, line 1756: This entire section may be more readable if we use:
```
assert {"name": "subnetId", "value": setup_resources[0].id} in details
assert {"name": "networkInterfaceId", ..} in details
# etc
```
As far as I can tell, that would test the same thing

bblommers (changes requested):

tests/test_ecs/test_ecs_boto3.py
rafcio19: You're right, much cleaner and less lines 💪

bblommers (approved):
LGTM - thanks for fixing this @rafcio19! Much appreciated.

Issue #5496, opened 2022-09-26
Dynamodb mock, missing validation for transactions (transact_write_items)


LukaK:
Hello,
creating transaction with conflicting operations seem to pass locally with dynamodb_mock. For example creating Add and Delete operation of the same item will fail on the stack but pass locally.

Code to reproduce.
```
#!/usr/bin/env python
import boto3
import pytest
from moto import mock_dynamodb



@pytest.fixture()
def mock_table():
    with mock_dynamodb():

        dynamodb = boto3.resource("dynamodb", region_name="eu-west-1")

        # Create the table
        dynamodb.create_table(
            TableName="test_table",
            KeySchema=[
                {"AttributeName": "pk", "KeyType": "HASH"},
                {"AttributeName": "sk", "KeyType": "RANGE"},
            ],
            AttributeDefinitions=[
                {"AttributeName": "pk", "AttributeType": "S"},
                {"AttributeName": "sk", "AttributeType": "S"},
            ],
            BillingMode="PAY_PER_REQUEST",
        )
        yield dynamodb

def main():
    dc = boto3.client("dynamodb")
    dc.transact_write_items(
        TransactItems = [
            {
                "Put": {
                    "Item": {"pk": {"S": 'test'}, "sk": {"S": 'test'}},
                    "TableName": "test_table",
                },
                "Delete": {
                    "Key": {"pk": {"S": 'test'}, "sk": {"S": 'test'}},
                    "TableName": "test_table",
                }
            }
        ]
    )

def test__main(mock_table):
    main()

if __name__ == "__main__":
    main()

```

If you run `pytest` locally it will pass, but if you run it on real dynamodb table you will get
```
Traceback (most recent call last):
  File "./test.py", line 50, in <module>
    main()
  File "./test.py", line 40, in main
    "TableName": "test_table",
  File "/home/luka/.virtualenvs/tg-scheduler/lib/python3.7/site-packages/botocore/client.py", line 514, in _api_call
    return self._make_api_call(operation_name, kwargs)
  File "/home/luka/.virtualenvs/tg-scheduler/lib/python3.7/site-packages/botocore/client.py", line 938, in _make_api_call
    raise error_class(parsed_response, operation_name)
botocore.exceptions.ClientError: An error occurred (ValidationException) when calling the TransactWriteItems operation: TransactItems can only contain one of Check, Put
, Update or Delete
```
[...]

PR #5521, opened 2022-10-03
Improve DDB transaction validation

rafcio19:
Solution for https://github.com/spulec/moto/issues/5496

### DIFF
--- a/moto/dynamodb/exceptions.py
+++ b/moto/dynamodb/exceptions.py
@@ -323,3 +323,12 @@ class InvalidConversion(JsonRESTError):
     def __init__(self):
         er = "SerializationException"
         super().__init__(er, "NUMBER_VALUE cannot be converted to String")
+
+
+class TransactWriteSingleOpException(MockValidationException):
+    there_can_be_only_one = (
+        "TransactItems can only contain one of Check, Put, Update or Delete"
+    )
+
+    def __init__(self):
+        super().__init__(self.there_can_be_only_one)
--- a/moto/dynamodb/models/__init__.py
+++ b/moto/dynamodb/models/__init__.py
@@ -1656,16 +1656,11 @@ class DynamoDBBackend(BaseBackend):
             target_items.add(item)
 
         errors = []  # [(Code, Message, Item), ..]
-        ddb_verbs = ["ConditionCheck", "Put", "Delete", "Update"]
         for item in transact_items:
             # check transact writes are not performing multiple operations
             # in the same item
-            for ddb_verb in ddb_verbs:
-                if ddb_verb in item:
-                    verb1 = ddb_verb
-                    for verb2 in ddb_verbs:
-                        if verb2 in item and verb2 != verb1:
-                            raise TransactWriteSingleOpException
+            if len(list(item.keys())) > 1:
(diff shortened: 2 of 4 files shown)


tests/test_dynamodb/exceptions/test_dynamodb_exceptions.py
```diff
@@ -761,3 +761,42 @@ def test_query_begins_with_without_brackets():
...
+                }
+            ]
+        )
+        # Verify
+        err = exc.value.response["Error"]
+        assert err["Code"] == "ValidationException"
+        assert (
+            err["Message"]
+            == "TransactItems can only contain one of Check, Put, Update or Delete"
+        )
```
bobbyi, line 802:
```suggestion
    # Verify
    err = exc.value.response["Error"]
    assert err["Code"] == "ValidationException"
    assert (
        err["Message"]
        == "TransactItems can only contain one of Check, Put, Update or Delete"
)
```

As it stands, if `transact_write_items` throws an exception, the verification code is never run.

rafcio19: Good catch

moto/dynamodb/exceptions.py
```diff
@@ -323,3 +323,12 @@ class InvalidConversion(JsonRESTError):
     def __init__(self):
         er = "SerializationException"
         super().__init__(er, "NUMBER_VALUE cannot be converted to String")
+
+
+class TransactWriteSingleOpException(MockValidationException):
+    there_can_be_only_one = (
```
bblommers, line 329: Nice reference :)

moto/dynamodb/models/__init__.py
```diff
@@ -1655,7 +1656,17 @@ def check_unicity(table_name, key):
             target_items.add(item)
 
         errors = []  # [(Code, Message, Item), ..]
+        ddb_verbs = ["ConditionCheck", "Put", "Delete", "Update"]
         for item in transact_items:
+            # check transact writes are not performing multiple operations
+            # in the same item
+            for ddb_verb in ddb_verbs:
```
bblommers, line 1663: This logic doesn't seem quite right. It is possible to have different verbs appear, they just should not act on the same item.
An alternative solution that may work:
List the values in `transact_items`, so the items that are being PUT/DELETED/etc., and then verify there are no duplicates there.
The easiest solution to check for duplicates is `list(values) != set(values)` - the `set`-operation will eliminate any duplicates, so if the list contains more items - we know it's a dup.

bblommers (changes requested):

moto/dynamodb/models/__init__.py
bblommers: Ah, interesting. I completely misunderstood the issue, then. Thanks for the explanation.

Can we still use the `list(keys) == set(keys)`-approach, to verify there are no multiple verbs? That is more readable then iterating over the list twice, IMO

bblommers: Looks like our replies crossed - using `len(set(..))` would also work, yes :)

bblommers (approved):
Awesome - thanks for the fix @rafcio19!

PR #5522, opened 2022-10-03
More portable grep invocation in Makefile

bobbyi:
The Makefile is using grep -oP. However, the -P flag to grep (for "use Perl regex syntax") is GNU-specific and does not exist in BSD grep (e.g., on MacOS).

We can use sed to accomplish the same thing in a way that will work on any POSIX-compliant system.


bblommers (commented):
Thanks for fixing this @bobbyi!

Issue #5523, opened 2022-10-03
Private IP address specified for EC2 instance via NetworkInterfaces not always used


bobbyi:
I initially encountered this with moto 4.0.5 and botocore 1.27.84 using Python mocks.

I create an EC2 instance and specify a private IP address for its primary network interface by providing NetworkInterfaces. But moto's `Instance.prep_nics` method overrides my provided IP address with what it gets from `subnet.get_available_subnet_ip`.

I have verified that when creating a real EC2 instance via boto (without moto involved), the provided IP address from NetworkInterfaces is used.

Here is a failing test demonstrating the problem. The assertion on the last line should pass if moto is behaving like real EC2/ boto, but currently doesn't.

```
import boto3
from moto import mock_ec2

EXAMPLE_AMI_ID = "ami-12c6146b"

@mock_ec2
def test_run_instance_in_subnet_with_nic_private_ip():
    vpc_cidr_block = "10.26.0.0/16"
    subnet_cidr_block = "10.26.1.0/24"
    private_ip = "10.26.1.3"
    ec2 = boto3.resource("ec2", region_name="eu-west-1")
    vpc = ec2.create_vpc(CidrBlock=vpc_cidr_block)
    subnet = ec2.create_subnet(
        VpcId=vpc.id,
        CidrBlock=subnet_cidr_block,
    )
    my_interface = {
        "SubnetId": subnet.id,
        "DeviceIndex": 0,
        "PrivateIpAddress": private_ip,
    }
    [instance] = ec2.create_instances(
        ImageId=EXAMPLE_AMI_ID, NetworkInterfaces=[my_interface], MinCount=1, MaxCount=1
    )
    instance.private_ip_address.should.equal(private_ip)
```

PR #5524, opened 2022-10-03
Use private IP provided via NetworkInterfaces when creating EC2 instance (#5523)

bobbyi:
When creating an EC2 instance with a primary network specified by NetworkInterfaces, `Instance.prep_nics` should not clobber the provided private IP address with what it gets from `subnet.get_available_subnet_ip`.

This resolves issue #5523


bobbyi:
The coverage report above has a single line of my commit without coverage. I'm going to try to rework the logic so there is full coverage.

bobbyi:
I added one additional commit  because I learned after creating this that the `Primary` field is a string, not a bool

bblommers (approved):
Awesome - thanks for fixing this @bobbyi!

PR #5526, opened 2022-10-04
Support all (current) EC2 instance-type filters

EpicWink:


bblommers (approved):
LGTM! Thanks for adding these @EpicWink

PR #5528, opened 2022-10-04
Include actual previous state in EC2 instance state-change actions

EpicWink:
State-change actions include start, stop and terminate instances.


bblommers (approved):
LGTM! Thanks again @EpicWink

PR #5531, opened 2022-10-04
Prettify responses

szopen321:
Prettify responses from moto. I monkey patch jinja render to avoid adding the same piece of code manually in each place where template is rendered.

- [x] make monkey patch to prettify responses
- [x] add documentation
- [x] add test

Please note that AWS returns pretiffied responses

Solution to https://github.com/spulec/moto/issues/5537


bblommers:
What problem are you trying to solve here, @szopen321?

szopen321:
Hi @bblommers!

Thank you for your vigilance! Moto returns responses in form of a single line and I want to make it return prettified responses - with newlines and indents. Please note that AWS is returning prettified responses.

This is a work in progress, I will create an issue when  I will be ready.

bblommers:
> 
> Moto returns responses in form of a single line and I want to make it return prettified responses

That's the solution - not the problem. :) 

Are you trying to make debugging easier, by making the response human-readable? Or is there an SDK/library that connects to Moto, and expects pretty responses?

>
> Please note that AWS is returning prettified responses.

That's interesting - AWS used to return XML like we do now, that's why we explicitly removed any whitespace. (Original PR is here: https://github.com/spulec/moto/pull/2523)

szopen321:
> Are you trying to make debugging easier, by making the response human-readable? Or is there an SDK/library that connects to Moto, and expects pretty responses?

My motivation is to make debugging easier. 

> > Please note that AWS is returning prettified responses.
> 
> That's interesting - AWS used to return XML like we do now, that's why we explicitly removed any whitespace. (Original PR is here: #2523)

I was debugging other parts of my code and I was comparing responses from moto and from AWS. AWS returned prettified responses. At that time I was not calling AWS or moto directly but with terraform while having `TF_LOG=True` and selected `TF_LOG_PATH`. I am not sure but maybe it is possible that AWS is responding to terraform differently than to direct API calls?

In any case I will put this prettifying functionality as an option, behind env var.

szopen321:
Hi @bblommers! 

I'm trying to figure out why in the tests fail? I can see that they claim `ModuleNotFoundError: No module named 'jinja2' but i don't understand why is jinja2 not found. Thanks in advance!
`

szopen321:
Thank you kindly @bblommers for your response - it helped!

moto/core/responses.py
```diff
@@ -18,6 +18,7 @@
 from typing import Dict, List, Union, Any, Optional
 from urllib.parse import parse_qs, parse_qsl, urlparse
 from werkzeug.exceptions import HTTPException
+import moto.prettify_jinja_render  # noqa # pylint: disable=unused-import
```
bblommers, line 21: Sorry, I wasn't very clear earlier. I was hoping to have all the logic here, instead of in a separate file.
It is better to extend the existing method that stores the XML, IMO. In pseudocode:
```
def response_template(self, source):
    template_id = self._make_template_id(source)
        if not self.contains_template(template_id):
            if PRETTIFY:
                pretty_xml = parseXML(..)
            else:
                collapsed = re.sub(
                    self.RIGHT_PATTERN, ">", re.sub(self.LEFT_PATTERN, "<", source)
                )
            self.environment.loader.update({template_id: collapsed or pretty_xml})
        return self.environment.get_template(template_id)
```

 - easier to reason if the logic is in one place
 - it doesn't clutter the root folder
 - it prevents people from treating this behaviour as a public API, by calling `from moto import prettify_jinja_render` in their own code

bblommers (approved):
LGTM! Thanks again for adding this feature @szopen321

PR #5532, opened 2022-10-04
Fix invalid escape sequence, enable pylint check

bobbyi:
In commit c4626888460, `r"x-amz-checksum-\w+:(\w+={1,2})"` was changed to  `b"x-amz-checksum-\w+:(\w+={1,2})"` (same content, different prefix). In addition to changing the type from to str to bytes (which was the intention), this lost the 'r' prefix which means that `\w` is now treated as an escape sequence.

Since it isn't actually a valid escape sequence, Python falls back to assuming it's not an escape sequence at all and the backslash is literal. This is correct and gets the same result as the "raw" version would, but generates `DeprecationWarning: invalid escape sequence \w`

This PR re-adds the `r` prefix as well as enabling the pylint check that would have caught this.

### DIFF
--- a/moto/s3/responses.py
+++ b/moto/s3/responses.py
@@ -1394,7 +1394,7 @@ class S3Response(BaseResponse):
         checksum_value = request.headers.get(checksum_header)
         if not checksum_value and checksum_algorithm:
             # Extract the checksum-value from the body first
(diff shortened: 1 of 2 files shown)


bblommers (approved):
Nice catch @bobbyi - thanks!

PR #5533, opened 2022-10-05
Fix glue expression

rafcio19:
Fix for issue discussed in PR conversation on https://github.com/spulec/moto/pull/5495

Glue `get_tables` treats expressions with `*` as a glob-like wildcard.

FYI @robert-schmidtke


bblommers (approved):
Nice one - thanks for fixing this @rafcio19!

Issue #5508, opened 2022-09-30
moto's ec2:ReplaceRouteTableAssociation breaks for the main association


yoctozepto:
Calling ec2:ReplaceRouteTableAssociation with the main association results in InvalidAssociationIdError being raised (regardless of the placement of the main association id, i.e. the idempotency path is also failing).

This is fixed by an upcoming patch. This bug report is for issue tracking purposes only.

Issue #5507, opened 2022-09-30
moto's ec2 DescribeRouteTablesResponse claims all route tables have their main association


yoctozepto:
Calling ec2:DescribeRouteTables returns a main association for any route table even if it does not have a main association. This confuses external tools.

The issue has been partially introduced by patch #4726 for #1647 but it was wrong before as well (not all associations are main).

This is fixed by an upcoming patch. This bug report is for issue tracking purposes only.

Issue #5506, opened 2022-09-30
moto's ec2 ReplaceRouteTableAssociationResponse lacks associationState element


yoctozepto:
Calling ec2:ReplaceRouteTableAssociation returns an incomplete response, lacking associationState element, which may confuse some external tools.

This is fixed by an upcoming patch. This bug report is for issue tracking purposes only.

PR #5509, opened 2022-09-30
Fix 3 issues with route table associations

yoctozepto:
Fixes #5508 - replacing the main association raising an exception.
Fixes #5507 - showing all route tables as having the main association.
Fixes #5506 - the response for replace missing bits.

Existing unit tests fixed and a new one (testing replacing the main association) added.

PR #5534, opened 2022-10-05
Fix 3 issues with route table associations

szopen321:
Fixes https://github.com/spulec/moto/issues/5508 - replacing the main association raising an exception.
Fixes https://github.com/spulec/moto/issues/5507 - showing all route tables as having the main association.
Fixes https://github.com/spulec/moto/issues/5506 - the response for replace missing bits.

Existing unit tests fixed and a new one (testing replacing the main
association) added.


yoctozepto:
Already in via https://github.com/spulec/moto/pull/5509

PR #5539, opened 2022-10-06
Make available logStreamName key when a job is running

juanblo:
When you use describe_jobs method, it should return logStreamKey as part of response in container object when job is running (currently logStreamKey is returned only when job is stopped).


bblommers (approved):
LGTM! Thanks for the PR @juanblo

juanblo:
> LGTM! Thanks for the PR @juanblo

@bblommers thanks for accept the PR. Moto is awesome !

PR #5550, opened 2022-10-10
Add RDS modify_db_cluster()

marshall7m:
This introduces a Moto implementation for the AWS RDS ModifyDbCluster endpoint. To enforce the attribute constraints specified within `Cluster.__init__()` during the modification of the `Cluster` instance, the appropriate attributes are now wrapped within a Python `@property` decorators that implement the constraint within their respective property setters.

With the introduction of property values, the `Cluster` instance's `__dict__` values will not contain the same attributes that other backend methods may have relied on (e.g. `restore_db_cluster_from_snapshot()`). To combat this, there's a `get_cfg()` method that will return a modified version of `__dict__` but with hidden attribute names having their prefix `_` removed.

### DIFF
--- a/IMPLEMENTATION_COVERAGE.md
+++ b/IMPLEMENTATION_COVERAGE.md
@@ -4806,7 +4806,7 @@
 - [ ] modify_certificates
 - [ ] modify_current_db_cluster_capacity
 - [ ] modify_custom_db_engine_version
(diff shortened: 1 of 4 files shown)


bblommers (approved):
LGTM! Thanks for implementing this feature @marshall7m

PR #5556, opened 2022-10-11
S3: Close FileHandles on GC

bblommers:
Followup to (and improved implementation of) #5545. This PR closes any FileHandle (FH) that is created when a key is destroyed, 

####  Solution
This PR enhances the earlier solution to close FH's the moment keys/multiparts are deleted. Additionally, all are closed as part of the destructor, just in case there are any FH's created that we forgot to close.

### Unrelated changes

Also removes an method (`complete_multipart`) that is nowhere used or invoked, to remove any confusion.

@hannes-ucsc FYI: this is IMHO the best solution to ensure we're always closing file handles, regardless of where they come from/when they are no longer reachable. Is this acceptable for you?

### DIFF
--- a/moto/s3/models.py
+++ b/moto/s3/models.py
@@ -135,6 +135,7 @@ class FakeKey(BaseModel, ManagedState):
             max_buffer_size if max_buffer_size else get_s3_default_key_buffer_size()
         )
         self._value_buffer = tempfile.SpooledTemporaryFile(self._max_buffer_size)
+        self.disposed = False
         self.value = value
         self.lock = threading.Lock()
 
@@ -339,6 +340,22 @@ class FakeKey(BaseModel, ManagedState):
 
         return False
 
+    def dispose(self, garbage=False):
+        if garbage and not self.disposed:
+            import warnings
+
+            warnings.warn("S3 key was not disposed of in time", ResourceWarning)
+        try:
+            self._value_buffer.close()
+            if self.multipart:
+                self.multipart.dispose()
+        except:  # noqa: E722 Do not use bare except
+            pass
+        self.disposed = True
+
+    def __del__(self):
+        self.dispose(garbage=True)
+
 
 class FakeMultipart(BaseModel):
     def __init__(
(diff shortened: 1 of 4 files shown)


moto/s3/models.py
```diff
@@ -339,6 +339,16 @@ def is_locked(self):
...
+    def dispose(self):
+        if self.multipart is not None:
+            for part in self.multipart.parts.values():
+                part.dispose()
+        if self._value_buffer:
+            self._value_buffer.close()
+
+    def __del__(self):
+        self.dispose()
+
```
hannes-ucsc, line 351: I think it is a good idea to call `dispose()` in `__del__()` as a last resort. In the future, there may be additional clean-up for other per-key resources. It's worth noting that this is equivalent to what the Python runtime is doing: it's closing the file handle at collection time, which may be too late. Let's say we've missed one scenario in which we need to call `dispose()` explicitly, then many undisposed, garbage keys may have accumulated, which is still a leak. I would add a warning here so that we're alerted if we missed a situation in which an explicit call to `dispose()` is needed. Something along the lines of:

```suggestion
    def dispose(self, *, garbage=False):
        if garbage and not self.disposed:
            warnings.warn("S3 key was not disposed of in time", ResourceWarning)
        if self.multipart is not None:
            for part in self.multipart.parts.values():
                part.dispose(garbage=garbage)
        if self._value_buffer:
            self._value_buffer.close()
        self.disposed = True

    def __del__(self):
        self.dispose(garbage=True)

```

I know this is a high bar to pass, what do you think?

bblommers: Having a warning here is indeed useful, to notify us that we've missed a scenario. Implementing that does expose some techdebt in our code, unfortunately; it looks like we open FileHandles everywhere without closing them properly. TBC! I'll continue to see if there is a cleaner solution.

bblommers: As far as I can tell, all scenarios are now covered. I'll close this, but let me know if I've missed anything. Thanks again for raising this and reviewing this @hannes-ucsc!

hannes-ucsc: https://github.com/spulec/moto/issues/5584

PR #5585, opened 2022-10-20
Feature/improve SES documentation

bendalby82:
Brings SES documentation in line with SNS documentation:
http://docs.getmoto.org/en/latest/docs/services/sns.html

Based on Stack Overflow answer here:
https://stackoverflow.com/a/74044893/363784

### DIFF
--- a/docs/docs/services/ses.rst
+++ b/docs/docs/services/ses.rst
@@ -12,7 +12,7 @@
 ses
 ===
 
-.. autoclass:: moto.sns.models.SESBackend
+.. autoclass:: moto.ses.models.SESBackend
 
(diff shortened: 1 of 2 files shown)


bendalby82:
I think I've fixed the lint problems now - I'm not sure why the [sns check](https://github.com/spulec/moto/actions/runs/3289522855/jobs/5422652914) failed, it seems to be an ephemeral problem.

bblommers:
Resulting docs: https://moto--5585.org.readthedocs.build/en/5585/docs/services/ses.html

bblommers (approved):
LGTM! Thanks for adding this to the docs @bendalby82

PR #5590, opened 2022-10-23
TechDebt - Unpin TF

bblommers:
Originally raised in https://github.com/hashicorp/terraform-provider-aws/issues/27049

Upgrading to the latest `main`-branch removes the need to pin Terraform to an older version.

The latest branch tests some functionality that's still not covered by Moto, so we had to remove some of the tests that previously passed.


bblommers:
FYI @thrau

Issue #5605, opened 2022-10-26
create_vpc_endpoint_service_configuration does not include `Owner` key in resource


sergargar:
When using `create_vpc_endpoint_service_configuration` function, the created service does not have the `owner` key that AWS automatically associates with your account_id.
So, when I use the `describe_vpc_endpoint_services` function, the created service does not show any owner and I expect to be the mocked account id.

This is an example of an expected service:

> {'ServiceName': 'com.amazonaws.vpce.eu-west-1.vpce-svc-0e8175b5', 'ServiceId': 'vpce-svc-00e8175b5', 'ServiceType': [{'ServiceType': 'Interface'}], 'AvailabilityZones': ['eu-west-1a'], **_'Owner': '123456789012'_**, 'BaseEndpointDnsNames': ['vpce-svc-00e8175b5.eu-west-1.vpce.amazonaws.com'], 'VpcEndpointPolicySupported': False, 'AcceptanceRequired': True, 'ManagesVpcEndpoints': False, 'Tags': [], 'SupportedIpAddressTypes': ['ipv4']}

More info:
https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/ec2.html#EC2.Client.describe_vpc_endpoint_services

PR #5606, opened 2022-10-26
fix(add owner to service): set self.current_account as VPC Service Endpoint Owner

sergargar:
Regarding https://github.com/spulec/moto/issues/5605 since the owner in VPC Service Endpoints are always empty, set self.current_account as VPC Service Endpoint Owner.

### DIFF
--- a/moto/ec2/responses/vpcs.py
+++ b/moto/ec2/responses/vpcs.py
@@ -1,7 +1,7 @@
 from moto.core.utils import camelcase_to_underscores
 from moto.ec2.utils import add_tag_specification
 from ._base_response import EC2BaseResponse
-
+from moto.core import DEFAULT_ACCOUNT_ID
 
 class VPCs(EC2BaseResponse):
     def _get_doc_date(self):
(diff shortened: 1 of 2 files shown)


bblommers:
The default services will always have `owner=amazon`, so this solution is not correct.
To verify this, see the output for the cassandra service:
```
aws ec2 describe-vpc-endpoint-services --filters '[{"Name":"service-name","Values":["com.amazonaws.us-east-1.cassandra"]}]'
```

-----------------
From the issue you opened:
`So, when I use the describe_vpc_endpoint_services function, the created service does not show any owner and I expect to be the mocked account id.`
As far as I can tell, our `describe-`-endpoint doesn't yet return user-created (non-default) services? From the docs:
```
[X] describe_vpc_endpoint_services 
Return info on services to which you can create a VPC endpoint.

    Currently only the default endpoing services are returned.
```
http://docs.getmoto.org/en/latest/docs/services/ec2.html

So we would have to implement that first, to actually return user-created configurations, and then we can change the `Owner`-field to be either `amazon` or the account ID.

PR #5617, opened 2022-10-28
feat: link instance profile with ec2 instance

rafcio19:
Improvement on https://github.com/spulec/moto/pull/3064

Links up IAM instance profile with the ec2 backend.

### DIFF
--- a/moto/ec2/models/iam_instance_profile.py
+++ b/moto/ec2/models/iam_instance_profile.py
@@ -17,6 +17,11 @@ class IamInstanceProfileAssociation(CloudFormationModel):
         self.instance = instance
         self.iam_instance_profile = iam_instance_profile
         self.state = "associated"
+        ec2_backend.modify_instance_attribute(
+            instance.id,
+            "iam_instance_profile",
+            {"Arn": self.iam_instance_profile.arn, "Id": association_id},
+        )
 
 
 class IamInstanceProfileAssociationBackend:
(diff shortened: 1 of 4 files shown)


rafcio19:
Hey @bblommers, could you rerun the [failing Unit tests in Server Mode (3.6) job ](https://github.com/spulec/moto/actions/runs/3346472400/jobs/5543521741#step:9:6684)? Seems like a fluke, everything else passes 🤞 🤔

rafcio19:
That _ Unit tests in Server Mode (3.6)_ job passes now so I think this is ready to be reviewed.

bblommers:
Hi @rafcio19, thanks for the PR! 

The Terraform tests are still failing with this error:
`Error: error setting iam_instance_profile: error parsing ARN (): arn: invalid prefix`
Which is due to the fact that the ARN is now empty when no instance profile is provided.

It should be OK if we change this to only output this part of the XML if the instance profile exists:
```
{% if instance.iam_instance_profile %}
          <iamInstanceProfile>
            <arn>{{ instance.iam_instance_profile['Arn'] }}</arn>
            <id>{{ instance.iam_instance_profile['Id'] }}</id>
          </iamInstanceProfile>
{% endif %}
```
That should fix all the tests - happy to merge once they pass.

bblommers (approved):
LGTM - thanks for the PR @rafcio19!

PR #5618, opened 2022-10-29
Add EnableNetworkAddressUsageMetrics to VPC attribute

marshall7m:
Adds support for `enableNetworkAddressUsageMetrics` to be used in `modify_vpc_attribute()` and `describe_attribute()` calls to the Ec2 service.

### DIFF
--- a/moto/ec2/responses/vpcs.py
+++ b/moto/ec2/responses/vpcs.py
@@ -137,8 +137,7 @@ class VPCs(EC2BaseResponse):
 
     def modify_vpc_attribute(self):
         vpc_id = self._get_param("VpcId")
-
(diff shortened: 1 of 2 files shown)


bblommers (changes requested):
Hi @marshall7m! `botocore` has dropped support for Py3.6 a while back, which means that if we run the tests in python 3.6, it is using a older version of botocore that does not support this newer attribute.

You should be able to just skip these tests in 3.6 - see the below for an example on how to do so:

https://github.com/bblommers/moto/blob/906a597c1774d513e5800eb5aa83d5ca375f10b3/tests/test_amp/test_amp_logging_config.py#L10:L13

bblommers (approved):
LGTM! Thanks for the PR @marshall7m

Issue #5625, opened 2022-10-31
End time for Cloudwatch `get_metric_data` should be exclusive


christiankjaer:
Looking at the code for `get_metric_data` is looks like the metrics to be included are including the end point

https://github.com/spulec/moto/blob/master/moto/cloudwatch/models.py#L503

But looking at the docs

https://docs.aws.amazon.com/AmazonCloudWatch/latest/APIReference/API_GetMetricData.html

> EndTime
>   The time stamp indicating the latest data to be returned.
>   The value specified is exclusive; results include data points up to the specified time stamp.

It seems like the `end_time` should be excluded

PR #5627, opened 2022-11-01
End times for aggregations should be exclusive

christiankjaer:
Fixes https://github.com/spulec/moto/issues/5625

Data to be considered for aggregations should not include the `end_time`

More info on the ticket.


bblommers (approved):
LGTM! Thanks for reporting this and for providing the fix @christiankjaer

christiankjaer:
No problem. Thanks for the quick feedback.

PR #5644, opened 2022-11-06
Fix #5595 : batch_simple regression.

### LINKED ISSUE #5595: Regression when using mock_batch_simple since 4.0.7
Since moto 4.0.7 we noticed some regressions in our unit tests using `mock_batch_simple`. All the jobs container details are empty while they weren't in version 4.0.6.

For instance this code is now failing because there's no 'command' key in the dictionary.

```python
    result = batch.describe_jobs(jobs=job_ids)['jobs']
    commands = [job['container']['command'] for job in result]
```

We think this [change](https://github.com/spulec/moto/pull/5539) is the source of the issue but I'm not sure that changing the behavior of `mock_batch_simple` was intended here.

Can you confirm this, we can try to fix if needed ?

bmaisonn:
Fix  #5595 : batch_simple regression.

The container details weren't returned in the job description anymore.

        [ ] Feature is added
        [x] The linter is happy
        [x] Tests are added
        [x] All tests pass, existing and new

### DIFF
--- a/moto/batch/models.py
+++ b/moto/batch/models.py
@@ -474,9 +474,25 @@ class Job(threading.Thread, BaseModel, DockerModel, ManagedState):
         self._log_backend = log_backend
         self.log_stream_name: Optional[str] = None
 
-        self.container_details: Dict[str, Any] = {}
         self.attempts: List[Dict[str, Any]] = []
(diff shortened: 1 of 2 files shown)


bblommers (approved):
LGTM! Thanks for fixing this @bmaisonn

PR #5651, opened 2022-11-09
Update pytest and pytest-xdist to the latest versions

mikelane:
# Overview

This work updates the cryptography module to the latest version in the setup.py file. The current version was set to be >=3.3.1 and the latest is 38.0.3. In Sept 2021, the cryptography releases went from 3.4.8 to 35.0.0 which is why the version bump seems so huge.

# Rationale

The service I'm working on at work uses moto. The version of cryptography that gets installed triggers a dependabot warning: Vulnerable OpenSSL included in cryptography wheels. I noticed that the version in use in the moto dependencies was a ways off from the unaffected version, so I decided to make this PR to update it.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -29,7 +29,7 @@ def get_version():
 install_requires = [
     "boto3>=1.9.201",
     "botocore>=1.12.201",
-    "cryptography>=3.3.1",
+    "cryptography>=38.0.3",
     "requests>=2.5",
     "xmltodict",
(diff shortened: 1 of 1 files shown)


bblommers:
Hi @mikelane, thanks for the PR! As a library, we want to support the widest range of versions as possible. Other users may require an older version of `cryptography` for their project, and it is not be Moto's place to tell people to upgrade.

I'd suggest pinning the `cryptography` dependency manually in your project, if you want a more recent version.

PR #5652, opened 2022-11-10
Replace email with username in cogntio JWT tokens

mglowinski93:
According to [documentation](https://docs.aws.amazon.com/cognito/latest/developerguide/amazon-cognito-user-pools-using-the-access-token.html) JWT token must contain "username" and not `email`. In edge case, `username` can contain `sub` as value, but implementation of this package does not support that situation, so simply `email` was replaced with `username`.

### DIFF
--- a/moto/cognitoidp/models.py
+++ b/moto/cognitoidp/models.py
@@ -539,7 +539,7 @@ class CognitoIdpUserPool(BaseModel):
             "token_use": token_use,
             "auth_time": now,
             "exp": now + expires_in,
-            "email": flatten_attrs(self._get_user(username).attributes).get("email"),
+            "username": username,
         }
         payload.update(extra_data or {})
         headers = {"kid": "dummy"}  # KID as present in jwks-public.json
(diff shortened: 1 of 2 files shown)


bblommers (changes requested):
Hi @mglowinski93, thanks for the PR!

We already have a test case (`tests/test_congitoidp/test_cognitoidp.py::test_token_legitimacy`) where we verify that the jwt-token is correct - can you add an assertion there that the `username`-field is set, and contains the correct value?

mglowinski93:
> Hi @mglowinski93, thanks for the PR!
> 
> We already have a test case (`tests/test_congitoidp/test_cognitoidp.py::test_token_legitimacy`) where we verify that the jwt-token is correct - can you add an assertion there that the `username`-field is set, and contains the correct value?

Done.

bblommers (approved):
Awesome - thanks again @mglowinski93!

mglowinski93:
> Awesome - thanks again @mglowinski93!

Glad to help :)

Issue #5579, opened 2022-10-18
Style: formats can be done with f-strings


wonhyeongseo:
[x] Avoid raising duplicate issues.

- Guidance on **how to reproduce the issue**.
I am using VSCode on WSL and with a query on the search tool as below,
![image](https://user-images.githubusercontent.com/29195190/196421189-94b5e439-dc50-4f4e-8d4d-3216805761d4.png)
there are 2116 format strings that can be f-strings.

- What does "success" look like for your code?
Ideally all of the strings would be replaced by f-strings.

- Tell us **what version of Moto you're using**, and **how you installed it**.
I installed moto via localstack.
`pip list` shows me `moto-ext 4.0.7.dev0 /home/wseo/localstack-dev/moto`

bblommers:
Done! Thanks for raising this @wonhyeongseo

PR #5657, opened 2022-11-11
Techdebt: Replace string-format with f-strings (for b* dirs)

bblommers:
See #5579

### DIFF
--- a/moto/batch/models.py
+++ b/moto/batch/models.py
@@ -609,7 +609,7 @@ class Job(threading.Thread, BaseModel, DockerModel, ManagedState):
                 )
                 for m in self._get_container_property("mountPoints", [])
             ]
-            name = "{0}-{1}".format(self.job_name, self.job_id)
+            name = f"{self.job_name}-{self.job_id}"
 
             self.advance()
             while self.status == "PENDING":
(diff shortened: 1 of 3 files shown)


moto/batch/models.py
```diff
@@ -1045,34 +1035,30 @@ def create_compute_environment(
...
         except IAMNotFoundException:
             raise InvalidParameterValueException(
-                "Could not find IAM role {0}".format(service_role)
+                f"Could not find IAM role {service_role}"
             )
 
         if _type not in ("MANAGED", "UNMANAGED"):
             raise InvalidParameterValueException(
-                "type {0} must be one of MANAGED | UNMANAGED".format(service_role)
+                f"type {_type} must be one of MANAGED | UNMANAGED"
```
bblommers, line 1051: Bugfix - having 'service_role' in the error message was probably a copy/paste error

moto/batch/models.py
```diff
@@ -1593,7 +1573,7 @@ def list_jobs(
 
         job_queue = self.get_job_queue(job_queue_name)
         if job_queue is None:
-            raise ClientException("Job queue {0} does not exist".format(job_queue))
+            raise ClientException(f"Job queue {job_queue_name} does not exist")
```
bblommers, line 1576: Bugfix - we know that 'job_queue' is None, so returning the name instead makes more sense

PR #5658, opened 2022-11-11
add support for cloudwatch put-metric-data with values list

steffyP:
This PR fixes [an issue](https://github.com/localstack/localstack/issues/7154) reported for LocalStack: 

[According to the cloudwatch API ](https://docs.aws.amazon.com/AmazonCloudWatch/latest/APIReference/API_PutMetricData.html)`put-metric-data` could also accepts a list of `values` and optional `counts` for those values.

> You can publish either individual data points in the Value field, or arrays of values and the number of times each value occurred during the period by using the Values and Counts fields in the MetricDatum structure.

For example:
`awslocal cloudwatch put-metric-data --namespace test --metric-data MetricName=mymetric,Values=10,1,Counts=1,2`

In this PR:
* parse `values` and `counts` if set and add datapoints accordingly
* added tests

Also [prepared a PR in LocalStack](https://github.com/localstack/localstack/pull/7166) that adds a snapshot test (AWS verified) to verify the fix.

### DIFF
--- a/moto/cloudwatch/models.py
+++ b/moto/cloudwatch/models.py
@@ -473,22 +473,66 @@ class CloudWatchBackend(BaseBackend):
                 raise InvalidParameterValue(
                     f"The value NaN for parameter MetricData.member.{i + 1}.Value is invalid."
                 )
+            if metric.get("Values.member"):
+                if "Value" in metric:
+                    raise InvalidParameterValue(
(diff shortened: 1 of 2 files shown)


bblommers (approved):
LGTM! Thanks for adding this feature @steffyP

PR #5660, opened 2022-11-11
feat(CloudFront): Include get_distribution_config

jfagoagas:
#### Description

Include the `get_distribution_config` CloudFront API call.

https://docs.aws.amazon.com/cloudfront/latest/APIReference/API_GetDistributionConfig.html

### DIFF
--- a/moto/cloudfront/models.py
+++ b/moto/cloudfront/models.py
@@ -73,17 +73,20 @@ class DefaultCacheBehaviour:
         self.compress = config.get("Compress", "true").lower() == "true"
         self.lambda_function_associations: List[Any] = []
         self.function_associations: List[Any] = []
-        self.field_level_encryption_id = ""
+        self.field_level_encryption_id = config.get("FieldLevelEncryptionId") or ""
         self.forwarded_values = ForwardedValues(config.get("ForwardedValues", {}))
         self.min_ttl = config.get("MinTTL") or 0
         self.default_ttl = config.get("DefaultTTL") or 0
         self.max_ttl = config.get("MaxTTL") or 0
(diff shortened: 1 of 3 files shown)


moto/cloudfront/models.py
```diff
@@ -274,6 +278,13 @@ def get_distribution(self, distribution_id: str) -> Tuple[Distribution, str]:
         dist.advance()
         return dist, dist.etag
 
+    def get_distribution_config(self, distribution_id: str) -> Tuple[Distribution, str]:
+        if distribution_id not in self.distributions:
+            raise NoSuchDistribution
```
bblommers, line 283: Ideally this gets a separate test case, as it is distinct behaviour that we want to verify

bblommers (changes requested):

moto/cloudfront/models.py
jfagoagas: What do you mean? I've included a test case for this new `get_distribution_config`

bblommers: I meant the behaviour of calling the method with an unknown/missing identifier.

See `test_create_distribution_with_mismatched_originid` as an example - it verifies that Moto returns the correct error message when incorrect parameters are supplied. We should have a similar test case that verifies what happens when calling `get_distribution_config(Id="unknown")`

jfagoagas: Totally agree, I'll review it. Thanks!

jfagoagas: Included test case @bblommers.

bblommers (approved):
LGTM!

PR #5670, opened 2022-11-15
[SWF] runId should be optional

use-strict:
> If a runId isn't specified, then the WorkflowExecutionSignaled event is recorded in the history of the current open workflow with the matching workflowId in the domain.

See https://docs.aws.amazon.com/amazonswf/latest/apireference/API_SignalWorkflowExecution.html


bblommers (approved):
LGTM!

PR #5671, opened 2022-11-15
[SWF] Schedule a single task if another is started

use-strict:
> If something occurs that generates a decision task while a decider is processing another decision task, Amazon SWF queues the new task until the current task completes. After the current task completes, Amazon SWF makes the new decision task available. Also, decision tasks are batched in the sense that, if multiple activities complete while a decider is processing a decision task, Amazon SWF will create only a single new decision task to account for the multiple task completions. However, each task completion will receive an individual event in the workflow execution history.

See https://docs.aws.amazon.com/amazonswf/latest/developerguide/swf-dg-dev-deciders.html#swf-dg-deciders-launch

### DIFF
--- a/moto/swf/models/workflow_execution.py
+++ b/moto/swf/models/workflow_execution.py
@@ -263,6 +263,18 @@ class WorkflowExecution(BaseModel):
         self.schedule_decision_task()
 
     def _schedule_decision_task(self):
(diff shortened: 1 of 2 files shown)


bblommers (approved):
LGTM! Thanks for improving the logic here @use-strict

PR #5672, opened 2022-11-15
fix(add-secret): SecretString can be None

jfagoagas:
#### Description

As per Boto3 documentation the `SecretString` `create_secret` API parameter is not required. 

So if you try to enable rotation in a previously created secret without a `SecretString`, the moto `rotate_secret` fails with a `KeyError` due to the inexistence of `old_secret_version["secret_string"]`

```python
@mock_secretsmanager
    def test_create_and_rotate_secret(self):
        secretsmanager_client = client("secretsmanager", region_name="eu-west-1")
        # Create Secret
        resp = secretsmanager_client.create_secret(
            Name="test-secret"
        )
        secret_arn = resp["ARN"]
        secret_name = resp["Name"]
        # Enable Rotation
        secretsmanager_client.rotate_secret( # <-- this call fails with a KeyError
            SecretId=secret_arn,
        ) 
```

### DIFF
--- a/moto/secretsmanager/models.py
+++ b/moto/secretsmanager/models.py
@@ -522,9 +522,14 @@ class SecretsManagerBackend(BaseBackend):
         # We add the new secret version as "pending". The previous version remains
         # as "current" for now. Once we've passed the new secret through the lambda
         # rotation function (if provided) we can then update the status to "current".
+        old_secret_version_secret_string = (
+            old_secret_version["secret_string"]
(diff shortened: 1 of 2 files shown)


jfagoagas:
> Hi @jfagoagas, thanks for the PR! Would you mind adding a unit test for this?

Sure, done!

bblommers (approved):
LGTM!

PR #5673, opened 2022-11-15
Fix IfModifiedSince handling in S3 get_object/head_object

lalinsky:
I think the most common use case for IfModifiedSince is:

1. Call get_object
2. Remember the LastModified date
3. Call get_object again with IfModifiedSince=LastModified to only get the object if it has changed

The implementation in moto currently doesn't work for two reasons:

1. The check for returning 'NotModified should include the case when LastModified == IfModifiedSince.
2. The returned timestamps are truncated to seconds, but the internally stored ones include microseconds, so the directly comparing them doesn't work. I initially wanted to truncate the internal datetimes, but that changes a very large number of tests, because some hashes change.


bblommers (approved):
LGTM! Thanks for fixing this @lalinsky

PR #5674, opened 2022-11-16
Timestreamwrite: Add regions

shimo164:
Timestream is available in these regions.
-  Asia Pacific (Sydney) | ap-southeast-2
-  Asia Pacific (Tokyo) | ap-northeast-1


bblommers (approved):
LGTM! Thanks for adding these @shimo164

PR #5675, opened 2022-11-16
Add response parameters to integration response

calvernaz:
Adds response parameters to integration response model.

https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-swagger-extensions-integration-responseParameters.html

https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-swagger-extensions-integration-response.html


bblommers (approved):
LGTM! Thanks for the PR @calvernaz

PR #5693, opened 2022-11-21
Add two new properties to the integration model object

calvernaz:
Adds `contentHandling`, `credentials` and `tlsConfig` properties to the Integration model object.

https://docs.aws.amazon.com/apigateway/latest/api/API_PutIntegration.html

Reads `disableExecuteApiEndpoint` property

https://docs.aws.amazon.com/apigateway/latest/api/API_CreateRestApi.html


bblommers (approved):
LGTM!

Issue #5677, opened 2022-11-16
EC2 client describe_images returning default AMIs even with MOTO_AMIS_PATH set


dblanchette:
## Context

We are setting `MOTO_AMIS_PATH` in a pytest plugin. Our test uses `describe_images` on an EC2 client.

## Expected

We expect the images set in the file at the path set in the environment variables `MOTO_AMIS_PATH` to be the only ones returned.

## Actual behavior

The images from `MOTO_AMIS_PATH` are there, but there are also other images that are hard-coded in the library.

## Affected versions

Present in 4.0.9, was not present in 3.1.18. moto installed with poetry (reproduced with moto installed manually in a virtual environment)

Seems to have been introduced by #5427 

Code here: https://github.com/spulec/moto/blob/6ab2497a12d8f0a5fe9be91387bc6b589d78e295/moto/ec2/models/amis.py#L157

## Reproduction steps

Run the following script
```python
import json
import os
from pathlib import Path

test_ami = [
    {
        "ami_id": "ami-760aaa0f760aaa0fe",
        "name": "infra-eks-20211110080547-bionic",
        "description": "An AMI",
        "owner_id": "123456789012",
        "public": False,
        "virtualization_type": "hvm",
        "architecture": "x86_64",
        "state": "available",
        "platform": None,
        "image_type": "machine",
        "hypervisor": "xen",
        "root_device_name": "/dev/sda1",
        "root_device_type": "ebs",
        "sriov": "simple",
        "creation_date": "2021-11-10T08:13:01.000Z"
    }
]

test_ami_path = Path(__file__).parent / "test_ami.json"

with test_ami_path.open("w") as fp:
    json.dump(test_ami, fp)

os.environ["MOTO_AMIS_PATH"] = str(test_ami_path)

import moto
import boto3

with moto.mock_ec2():
    ec2_client = boto3.client("ec2")
    images = ec2_client.describe_images()["Images"]
    print(f"Found {len(images)} images!")
    assert len(images) == 1
```
With moto 3.1.18:
```
$ python ./reproduce_step.py 
Found 1 images!
```
With moto 4.0.9:
```
$ python ./reproduce_step.py 
Found 18 images!
Traceback (most recent call last):
  File "./reproduce_step.py", line 39, in <module>
    assert len(images) == 1
AssertionError
```

## Conclusion

A workaround is to use the old version for now. Thanks for your help resolving this issue!

bblommers:
[...]

PR #5697, opened 2022-11-21
Fix custom amis import

rafcio19:
If custom amis are specified via `MOTO_AMIS_PATH` env var, then do not import default amis.

Resolves https://github.com/spulec/moto/issues/5677


bblommers:
Some context, in case you hadn't seen it yet: all tests in `tests/test_ec2` are [run in parallel](http://docs.getmoto.org/en/latest/docs/contributing/development_tips/tests.html#parallel-tests), to speed things up. So testing this particular feature, and assert we have only one AMI available, is a bit tricky.

I haven't really sat down to think about how to do this - but maybe we can simply run this test from a separate folder, something like `tests/test_ec2_custom_ami/test.py`. That way we can continue running all 'regular' EC2 tests in parallel, and we can be sure that this particular test is always run on it's own, without interference from other tests.

Let me know if you want me to take a closer look.

rafcio19:
Thanks for looking @bblommers . I saw this behaviour when testing locally:

```
 pytest tests/test_ec2/test_*ami*
```

but running these fellas individually works:

```
pytest tests/test_ec2/test_amis.py
pytest tests/test_ec2/test_custom_amis.py
```

I will see if a bit of tinkering with the makefile will help.

rafcio19:
Hey @bblommers this is ready for a review, now lemme know what you think. I changed the `Unit tests in Server Mode` workflow to run the `test_server` make target and the tests pass this way.. 😬

Clearly this changes the CI a little bit but because the "regular" unit tests pass then perhaps it's ok.

bblommers (approved):
Thanks @rafcio19! I've refactored the test a bit - it now resets and re-imports the EC2-backend after setting the env variable. That way we can remove any existing AMI's, and only load our custom AMI, but without having to worry about running `pytest` multiple times.

Issue #5372, opened 2022-08-09
SQS allows sending of messages with delays of longer than 15 minutes


paulmdavies:
# Versions

`moto`: 3.1.16
`boto3`: 1.21.46

# Code
```
response = boto3.client('sqs').send_message_batch(
    QueueUrl='the-queue-url,
    Entries=[
        {
            'Id': '1'
            'MessageBody': 'foo',
            'DelaySeconds': 180
        },
        {
            'Id': '2'
            'MessageBody': 'foo',
            'DelaySeconds': 1800
        }
    ]
)
```

# Expectation

An response with a failure for the second message:
```
{'Successful': [{'Id': '180', 'MessageId': '...', 'MD5OfMessageBody': '...'}], 'Failed': [{'Id': '1800', 'SenderFault': True, 'Code': 'InvalidParameterValue', 'Message': 'Value 1800 for parameter DelaySeconds is invalid. Reason: DelaySeconds must be >= 0 and <= 900.'}], 'ResponseMetadata': {'RequestId': 'a59c20de-034d-5922-b608-7bb267435efc', 'HTTPStatusCode': 200, 'HTTPHeaders': {'x-amzn-requestid': 'a59c20de-034d-5922-b608-7bb267435efc', 'date': 'Tue, 09 Aug 2022 13:31:58 GMT', 'content-type': 'text/xml', 'content-length': '714'}, 'RetryAttempts': 0}}
```

# Reality
```
{'Successful': [{'Id': '180', 'MessageId': '...', 'MD5OfMessageBody': '...'}, {'Id': '1800', 'MessageId': '...', 'MD5OfMessageBody': '...'}], 'ResponseMetadata': {'RequestId': '5ZAAK7GFDHO7FZSZVFMFD36QVM5TDZWPYRKFXL26GTJME1WU82KV', 'HTTPStatusCode': 200, 'HTTPHeaders': {'server': 'amazon.com', 'x-amzn-requestid': '5ZAAK7GFDHO7FZSZVFMFD36QVM5TDZWPYRKFXL26GTJME1WU82KV', 'x-amz-crc32': '3107167176'}, 'RetryAttempts': 0}}
```

bblommers:
Thanks for raising this @paulmdavies - marking it as an enhancement to add this validation.

RahiBhensdadia:
I want to work on this issue. I found it interesting and I think I can fix this. It would be great if I get the approval from the owner.

bblommers:
[...]

PR #5703, opened 2022-11-23
added code to resolve issue #5372

RahiBhensdadia:
Dear sir,
I have written code to resolve the issue number #5372. I have made following changes for the same:
1. models.py file (moto/moto/sqs/models.py): 
   Line number 47: Micro defined as MAXIMUM_MESSAGE_DELAY = 900
   Line number 808: In send_message function, I have put condition to check if delay_seconds is greater than MAXIMUM_MESSAGE_DELAY. If it is greater then InvalidParameterValue exception is raised with message described in the issue

2. test_sqs.py file (moto/tests/test_sqs.py):
   Line number 3125: Written test to check if above approach works for 2 messages that are given in the issue.  One message is sent with delayseconds =180 which has to be sent successfully and another message is sent with delayseconds=1800 which should raise an exception. 
When I run the test for sqs module, test test_message_delay_is_more_than_15_minutes() is passed.
Kindly Let me know if any further changes are required in this issue.

Thank you,
Rahi Bhensdadia

### DIFF
--- a/moto/sqs/models.py
+++ b/moto/sqs/models.py
@@ -857,7 +857,7 @@ class SQSBackend(BaseBackend):
                 )
                 message.user_id = entry["Id"]
                 messages.append(message)
-            except InvalidParameterValue as e:
+            except InvalidParameterValue:
                 failedInvalidDelay.append(entry)
 
         return messages, failedInvalidDelay
--- a/moto/sqs/models.py
+++ b/moto/sqs/models.py
@@ -801,16 +801,19 @@ class SQSBackend(BaseBackend):
         if message_attributes:
             message.message_attributes = message_attributes
 
+        # UPDATED Line 804-811
         if delay_seconds > MAXIMUM_MESSAGE_DELAY:
-            invalidDelay = True
-        else:
-            invalidDelay = False
+            msg = (
+                f"Value {delay_seconds} for parameter DelaySeconds is invalid. "
+                "Reason: DelaySeconds must be >= 0 and <= 900."
+            )
+            raise InvalidParameterValue(msg)
 
         message.mark_sent(delay_seconds=delay_seconds)
 
         queue.add_message(message)
 
-        return message, invalidDelay
+        return message
 
     def send_message_batch(self, queue_name, entries):
         self.get_queue(queue_name)
@@ -841,23 +844,22 @@ class SQSBackend(BaseBackend):
             raise TooManyEntriesInBatchRequest(len(entries))
 
         messages = []
-
+        # UPDATED Line 850-851, 853-868
         failedInvalidDelay = []
         for entry in entries.values():
-            # Loop through looking for messages
-            message, invalidDelay = self.send_message(
-                queue_name,
-                entry["MessageBody"],
-                message_attributes=entry["MessageAttributes"],
-                delay_seconds=entry["DelaySeconds"],
-                group_id=entry.get("MessageGroupId"),
-                deduplication_id=entry.get("MessageDeduplicationId"),
-            )
-            message.user_id = entry["Id"]
-
-            if not invalidDelay:
+            try:
+                # Loop through looking for messages
+                message = self.send_message(
+                    queue_name,
+                    entry["MessageBody"],
+                    message_attributes=entry["MessageAttributes"],
+                    delay_seconds=entry["DelaySeconds"],
+                    group_id=entry.get("MessageGroupId"),
+                    deduplication_id=entry.get("MessageDeduplicationId"),
+                )
+                message.user_id = entry["Id"]
                 messages.append(message)
-            else:
+            except InvalidParameterValue as e:
                 failedInvalidDelay.append(entry)
 
         return messages, failedInvalidDelay
--- a/moto/sqs/models.py
+++ b/moto/sqs/models.py
@@ -137,9 +137,9 @@ class Message(BaseModel):
     def validate_attribute_name(name):
         if not ATTRIBUTE_NAME_PATTERN.match(name):
             raise MessageAttributesInvalid(
-                "The message attribute name '{0}' is invalid. "
+                f"The message attribute name '{name}' is invalid. "
                 "Attribute name can contain A-Z, a-z, 0-9, "
-                "underscore (_), hyphen (-), and period (.) characters.".format(name)
+                "underscore (_), hyphen (-), and period (.) characters."
             )
 
     @staticmethod
@@ -402,9 +402,7 @@ class Queue(CloudFormationModel):
         else:
             raise RESTError(
                 "AWS.SimpleQueueService.NonExistentQueue",
-                "Could not find DLQ for {0}".format(
-                    self.redrive_policy["deadLetterTargetArn"]
-                ),
+                f"Could not find DLQ for {self.redrive_policy['deadLetterTargetArn']}",
             )
 
     @staticmethod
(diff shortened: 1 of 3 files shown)


bblommers (changes requested):
Hi @RahiBhensdadia!

When providing a delay that is too long, the call should succeed - but the response should say which entries failed, and which entries succeeded.
See the expected response from the original issue:
```
{
 'Successful': [{'Id': '180', 'MessageId': '...', 'MD5OfMessageBody': '...'}],
 'Failed': [{'Id': '1800', 'SenderFault': True, 'Code': 'InvalidParameterValue', 'Message': 'Value 1800 for parameter DelaySeconds is invalid. Reason: DelaySeconds must be >= 0 and <= 900.'}], 
 'ResponseMetadata': ...}
```

Your approach is correct here IMO, to add validation inside the `send_message`. But the `send_message_batch`-method should catch any errors when sending a message. Then it can either add the message to the list of successful messages, or add it to the list of failed messages.
(At the moment the `batch`-method always assumes success, so there only is one list)

-----
On a separate note: Moto added some linting rules recently, and we switched from string-formats `"abcd".format(..)` to f-strings. That's why the build is currently failing for this PR: https://github.com/spulec/moto/actions/runs/3529637252/jobs/5924309745

RahiBhensdadia:
Ok got it sir. But can you please tell me where and how exactly we can see the responses in the output for specifically this test. Because when we run commands for test, it is testing sqs module as a whole. And in the output we can just see whether testcase is passed or not and no any other details.

RahiBhensdadia:
Dear sir, 
As per your suggestions, I have made following changes:
1. models.py file: There is a flag named invalidDelay which is set to true if delay_seconds>MAXIMUM_MESSAGE_DELAY in send_message() function. Then in send_message_batch() function, one list is created named as failedInvalidDelay. If invalidDelay is true then message will be appended to this list. Else append it to messages.
2. test_sqs.py file: created test test_message_delay_is_more_than_15_minutes() in which 2 messages are sent with 180 and 1800 delays respectively. 
3. responses.py file: Added code to print the error message in SEND_MESSAGE_BATCH_RESPONSE at line 682.

Using this approach I am able to print the expected response when I run the following command:
pytest -sv tests/test_sqs/test_sqs.py::test_message_delay_is_more_than_15_minutes

Test case is passed and 2 lists are created as per your suggestion. Kindly let me know if any further changes are required.

moto/sqs/models.py
```diff
@@ -834,9 +841,11 @@ def send_message_batch(self, queue_name, entries):
             raise TooManyEntriesInBatchRequest(len(entries))
 
         messages = []
+
+        failedInvalidDelay = []
         for entry in entries.values():
             # Loop through looking for messages
-            message = self.send_message(
+            message, invalidDelay = self.send_message(
```
bblommers, line 848: The `DelaySeconds`-parameter can also part of the `boto3.client("sqs").send_message(..)`-call, so IMO we should stick with the initial implementation. Raising the error in `def send_message()` ensures that the validation happens in both scenarios, when calling `send_message` and `send_message_batch`.

That just means that `send_message_batch` should catch the error, something like:
```
try:
    message = self.send_message(..)
    messages.append(message)
except ..:
    failedInvalidDelay.append(entry)
```

tests/test_sqs/test_sqs.py
```diff
@@ -3115,3 +3115,44 @@ def test_message_has_windows_return():
...
+            # {
+            #     "Id": "4",
+            #     "MessageBody": "foo",
+            #     "DelaySeconds": 2800,
+            #     "MessageGroupId": "message_group_id_1",
+            # },
+        ],
+    )
+
+    print(msgs)
```
bblommers, line 3158: Can we add some assertions here, that `msgs` contains the correct response?

bblommers (changes requested):

moto/sqs/models.py
RahiBhensdadia: Ok sir this part I understood and implemented it.

tests/test_sqs/test_sqs.py
RahiBhensdadia: for this part, what exact assertions you are expecting. Can you please elaborate

RahiBhensdadia: I have done something like this:

@mock_sqs
def test_message_delay_is_more_than_15_minutes():
    client = boto3.client("sqs", region_name="us-east-1")
    response = client.create_queue(
        QueueName=f"{str(uuid4())[0:6]}.fifo", Attributes={"FifoQueue": "true"}
    )
    queue_url = response["QueueUrl"]

    response = client.send_message_batch(
        QueueUrl=queue_url,
        Entries=[
            {
                "Id": "id_1",
                "MessageBody": "body_1",
                "DelaySeconds": 10,
                "MessageAttributes": {
                    "attribute_name_1": {
                        "StringValue": "attribute_value_1",
                        "DataType": "String",
                    }
                },
                "MessageGroupId": "message_group_id_1",
                "MessageDeduplicationId": "message_deduplication_id_1",
            },
            {
                "Id": "id_2",
                "MessageBody": "body_2",
                "DelaySeconds": 1800,
                "MessageAttributes": {
                    "attribute_name_2": {"StringValue": "123", "DataType": "Number"}
                },
                "MessageGroupId": "message_group_id_2",
                "MessageDeduplicationId": "message_deduplication_id_2",
            },
        ],
    )

    sorted([entry["Id"] for entry in response["Successful"]]).should.equal(["id_1"])

    sorted([entry["Id"] for entry in response["Failed"]]).should.equal(["id_2"])

    print(response)

    time.sleep(11)

    response = client.receive_message(
        QueueUrl=queue_url,
        MaxNumberOfMessages=10,
        MessageAttributeNames=["attribute_name_1", "attribute_name_2"],
        AttributeNames=["MessageDeduplicationId", "MessageGroupId"],
    )

    response["Messages"].should.have.length_of(1)
    response["Messages"][0]["Body"].should.equal("body_1")
    response["Messages"][0]["MessageAttributes"].should.equal(
        {"attribute_name_1": {"StringValue": "attribute_value_1", "DataType": "String"}}
    )
    response["Messages"][0]["Attributes"]["MessageGroupId"].should.equal(
        "message_group_id_1"
    )
    response["Messages"][0]["Attributes"]["MessageDeduplicationId"].should.equal(
        "message_deduplication_id_1"
    )

I am receiving the message using receive_message function and then checking its attributes. Here the problem is - if delay seconds is i then we have to wait for i+1 seconds to receive the message then only we can do some checking. Is it ok or you are expecting something else

bblommers: Yes, I think that's fine. 

To speed things up, we can just reduce the DelaySeconds-parameter - e.g., if we can set it to 2, so we only have to wait 3 seconds.

RahiBhensdadia: OK sir I have raised PR with required changes. Let me know if it works

bblommers:
Thanks @RahiBhensdadia, that looks good to me overall.

The linter does seem to be failing: https://github.com/spulec/moto/actions/runs/3558579840/jobs/5977461461

You can run `make lint` locally to check whether it passes.

RahiBhensdadia:
Lint errors are resolved for the code in which I have made updates. There was one print statement and one unused variable. I have removed them. Now when I run make lint, it is giving some error in moto/core/custom_responses_mock.py file which is not part of the module in which I am working. So I have not made any changes in it as it might be risky to make changes.

bblommers (approved):
LGTM - thank you so much for fixing this @RahiBhensdadia!

RahiBhensdadia:
Welcome! Thank you sir for your continuous guidance and support to complete the task..

PR #5705, opened 2022-11-23
CloudFormation: Fixed TypeError using Number Parameters with Fn::Sub

aiudirog:
I ran into an issue testing a CloudFormation template that used Number Paramters with the Sub function where the integer value is passed directly to `str.replace()` while performing the substitution, resulting a TypeError. 

This PR fixes that bug by converting all values to strings before replacing.

Please let me know if there's anything you would like me to change or do.


bblommers (approved):
LGTM - thanks for fixing this @aiudirog, and welcome to Moto!

PR #5720, opened 2022-11-29
Fix  #5704: Moto isn't checking validity of the Batch job names

### LINKED ISSUE #5704: Moto isn't checking validity of the Batch job names
This is more an enhancement request.

As of now moto accepts to submit batch jobs with characters that aren't accepted by the real infra like '.'

image

image

Would it possible to perform the same kind of check in moto ?

bmaisonn:
Fix  #5704: Moto isn't checking validity of the Batch job names

Add a regex to valid the job name when submitting a job to raise a ClientException

        [] Feature is added
        [x] The linter is happy
        [x] Tests are added
        [x] All tests pass, existing and new


bblommers (approved):
LGTM! Thanks for implementing this @bmaisonn

PR #5722, opened 2022-11-29
Add unit filter to cloudwatch query get_metric_data

bogdangi:


bblommers (approved):
LGTM! Thanks for adding this feature @bogdangi

PR #5728, opened 2022-12-02
Add custom label param to cloudwatch get_metric_data

mandisma:
This PR adds the Label parameter for the get_metric_data request and allows modifying the label returned to the response.


bblommers (approved):
LGTM! Thanks for adding this @mandisma, and welcome to Moto!

PR #5729, opened 2022-12-02
EC2: AllocateAddress fallback domain

viren-nadkarni:
The [EC2 AllocateAddress](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_AllocateAddress.html) operation takes an optional `domain` parameter. Valid values for this parameter are `standard`|`vpc`. AWS accepts invalid values for this field (eg. `''` or `'xyz'`) and falls back to the default, `vpc`. Also, Pulumi seems to send `''`.

This PR corrects the behaviour such that in case of missing or invalid value, `vpc` is used instead of raising an exception. Furthermore, since the exception (`InvalidDomainError`) is not defined in AWS, it has been removed.

Tested against AWS:

```
$ aws ec2 allocate-address --domain ""
{
    "PublicIp": "3.74.120.XXX",
    "AllocationId": "eipalloc-06f0e8671db58XXXX",
    "PublicIpv4Pool": "amazon",
    "NetworkBorderGroup": "eu-central-1",
    "Domain": "vpc"
}

$ aws ec2 allocate-address --domain "xyz"
{
    "PublicIp": "3.123.242.XXX",
    "AllocationId": "eipalloc-05814a850338dXXXX",
    "PublicIpv4Pool": "amazon",
    "NetworkBorderGroup": "eu-central-1",
    "Domain": "vpc"
}
```

See https://github.com/localstack/localstack/issues/7058

### DIFF
--- a/moto/ec2/models/elastic_ip_addresses.py
+++ b/moto/ec2/models/elastic_ip_addresses.py
@@ -2,7 +2,6 @@ from moto.core import CloudFormationModel
 from .core import TaggedEC2Resource
 from ..exceptions import (
     FilterNotImplementedError,
-    InvalidDomainError,
     InvalidAddressError,
     InvalidAllocationIdError,
     ResourceAlreadyAssociatedError,
@@ -113,7 +112,7 @@ class ElasticAddressBackend:
 
     def allocate_address(self, domain, address=None, tags=None):
         if domain not in ["standard", "vpc"]:
-            raise InvalidDomainError(domain)
+            domain = "vpc"
         if address:
             address = ElasticAddress(self, domain=domain, address=address, tags=tags)
         else:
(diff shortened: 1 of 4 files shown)


bblommers (changes requested):
LGTM @viren-nadkarni - happy to merge if we can get the tests to pass

moto/ec2/models/elastic_ip_addresses.py
```diff
@@ -113,7 +112,7 @@ def __init__(self):
 
     def allocate_address(self, domain, address=None, tags=None):
         if domain not in ["standard", "vpc"]:
-            raise InvalidDomainError(domain)
+            domain = "vpc"
```
viren-nadkarni, line 115: The default value `standard` was only used in EC2-Classic. EC2-Classic was retired in August 2022.

bblommers (approved):
Thanks @viren-nadkarni

PR #5737, opened 2022-12-06
CloudFront: Persist origin changes made via update-distribution calls

### LINKED ISSUE #5736: CloudFront: Persist changes to origins
### Current Behaviour

New or modified origins passed as part of an `update-distribution` API call are not persisted.

### Expected Behaviour

Changes to origins made via `update-distribution` calls are visible in subsequent `get-distribution-config` & `get-distribution` calls.

I will submit a PR for this.

g-dx:
Fixes #5736


bblommers (approved):
LGTM - thanks for fixing this @g-dx, and welcome to Moto!

PR #5747, opened 2022-12-09
Route53: Record overwrites

viren-nadkarni:
Changes introduced in #5725 were not tested properly because of wrong Terraform test-case names.

This PR corrects the test case names. 

Source: https://github.com/hashicorp/terraform-provider-aws/blob/163210e073650650adba645b636128fbc19244f4/internal/service/route53/record_test.go#L1322

This needs to be accompanied with the fix. @bblommers would you be able to take this over?


bblommers:
No worries @viren-nadkarni - I'll have a look. Thanks for letting me know.

PR #5749, opened 2022-12-09
Remove dependency on pytz

Dunedan:
This removes the dependency on pytz, which was used to provide the UTC offset information for timezone aware datetime objects. While that was helpful for Python versions up to 3.1, UTC offset information is included in the standard library since Python 3.2. As moto doesn't support Python versions <3.6 anymore, using pytz isn't necessary anymore either.

### DIFF
--- a/moto/core/responses.py
+++ b/moto/core/responses.py
@@ -6,7 +6,6 @@ import logging
 import os
 import re
 import requests
-import pytz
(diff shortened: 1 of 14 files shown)


bblommers (approved):
Nice one - thanks @Dunedan!

PR #5753, opened 2022-12-10
ECR Manifest List Support

bkeane:
# What?
- `imageManifestMediaType` now present in `describe_image()` response model.
- Deployment of lambdas using `{imageUri}@{digest}` now supported.
- Multi-arch manifest lists now supported.
- `imageByteSize` now attempts to present manifest information, if not present, defaults to previous static value. It also returns `None` in the case it's a manifest list, as manifest lists do not have this in their response model.

Largely this PR is to support that which is captured in this screenshot:

image

The above shows 3 image manifests, and one manifest list. The `0.0.1` image manifest stands on it's own. The `0.0.2` image manifest list represents the metadata which describes the untagged image manifests as being either `arm64` or `amd64`. `docker pull {uri}:0.0.2` traverses this to pull the correct image for your local machine arch.

### DIFF
--- a/moto/awslambda/models.py
+++ b/moto/awslambda/models.py
@@ -489,9 +489,16 @@ class LambdaFunction(CloudFormationModel, DockerModel):
                 ).hexdigest()
                 self.code_size = 0
             else:
-                uri, tag = self.code["ImageUri"].split(":")
+                if "@" in self.code["ImageUri"]:
+                    # deploying via digest
+                    uri, digest = self.code["ImageUri"].split("@")
+                    image_id = { "imageDigest": digest }
+                else:
+                    # deploying via tag
+                    uri, tag = self.code["ImageUri"].split(":")
+                    image_id = {"imageTag": tag}
+
                 repo_name = uri.split("/")[-1]
-                image_id = {"imageTag": tag}
                 ecr_backend = ecr_backends[self.account_id][self.region]
                 registry_id = ecr_backend.describe_registry()["registryId"]
                 images = ecr_backend.batch_get_image(
--- a/moto/awslambda/models.py
+++ b/moto/awslambda/models.py
@@ -492,7 +492,7 @@ class LambdaFunction(CloudFormationModel, DockerModel):
                 if "@" in self.code["ImageUri"]:
                     # deploying via digest
                     uri, digest = self.code["ImageUri"].split("@")
-                    image_id = { "imageDigest": digest }
+                    image_id = {"imageDigest": digest}
                 else:
                     # deploying via tag
                     uri, tag = self.code["ImageUri"].split(":")
--- a/moto/ecr/models.py
+++ b/moto/ecr/models.py
@@ -236,7 +236,14 @@ class Repository(BaseObject, CloudFormationModel):
 
 class Image(BaseObject):
     def __init__(
-        self, account_id, tag, manifest, repository, image_manifest_mediatype=None, digest=None, registry_id=None
+        self,
+        account_id,
+        tag,
+        manifest,
+        repository,
+        image_manifest_mediatype=None,
+        digest=None,
+        registry_id=None,
     ):
         self.image_tag = tag
         self.image_tags = [tag] if tag is not None else []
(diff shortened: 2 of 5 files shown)


moto/ecr/models.py
```diff
@@ -486,12 +510,31 @@ def describe_images(self, repository_name, registry_id=None, image_ids=None):
...
+
+        if image_manifest_mediatype:
+            parsed_image_manifest["imageManifest"] = image_manifest_mediatype
+        else:
+            if "mediaType" not in parsed_image_manifest:
+                raise Exception(
+                    "image manifest mediatype not provided in manifest or parameter"
+                )
+            else:
+                image_manifest_mediatype = parsed_image_manifest["mediaType"]
```
bkeane, line 536: This is set of changes approximates the behavior of AWS. The manifest media type is not a required parameter of the boto3 call, because AWS makes a strong assumption that the media type will be included within the image manifest, and will subsequently derive it from there. If it's not in the image manifest, the manifest is invalid.

I didn't go down the rabbit hole of edge cases surrounding this and simply implemented a coercion of given manifest media type to override. What does AWS do when you give a manifest list media type that differs from the one within the manifest? I do not know. It's a bunch of work to figure out for a case that represents an anti-pattern as far as I can see.  It's curious that such a scenario is even exposed as a possibility and I have a strong feeling this is an oddity of ECR.

moto/ecr/models.py
```diff
@@ -250,17 +257,33 @@ def __init__(
...
+    def get_image_size_in_bytes(self):
+        image_manifest = json.loads(self.image_manifest)
+        if "layers" in image_manifest:
+            try:
+                return image_manifest["config"]["size"]
+            except KeyError:
+                return 50 * 1024 * 1024
+        else:
+            return None
+
```
bkeane, line 286: instead of enumerating all the possible media types an image manifest may have (oci, docker, etc), I've opted here to detect if the manifest contains layer information as that implies it is an image manifest of some media type (and not a manifest list).

moto/awslambda/models.py
```diff
@@ -489,9 +489,16 @@ def __init__(
...
-                uri, tag = self.code["ImageUri"].split(":")
+                if "@" in self.code["ImageUri"]:
+                    # deploying via digest
+                    uri, digest = self.code["ImageUri"].split("@")
+                    image_id = {"imageDigest": digest}
+                else:
+                    # deploying via tag
+                    uri, tag = self.code["ImageUri"].split(":")
+                    image_id = {"imageTag": tag}
+
```
bkeane, line 500: When deploying multi-arch images it's usually pretty sane to use the `{imageUri}@{digest}` format. Also, it's just a supported method of doing things, and therefor it should be supported. Pretty simple and reliable fix methinks.

tests/test_ecr/test_ecr_helpers.py
```diff
@@ -34,7 +35,52 @@ def _create_image_manifest(image_digest=None):
...
+#             'registryId': '677771948337',
+#             'repositoryName': 'kaixo',
+#             'imageDigest': 'sha256:1b9176ea7c94fb3adf5b75770201d07c9f3c4659d745c53bd4b4f6090de2df85',
+#             'imageTags': ['0.0.1'],
+#             'imageSizeInBytes': 180225217,
+#             'imagePushedAt': datetime.datetime(2022, 12, 9, 10, 24, 12, tzinfo=tzlocal()),
+#             'imageManifestMediaType': 'application/vnd.docker.distribution.manifest.v2+json',
+#             'artifactMediaType': 'application/vnd.docker.container.image.v1+json',
+#             'lastRecordedPullTime': datetime.datetime(2022, 12, 9, 10, 24, 22, 519000, tzinfo=tzlocal())
+#         },
```
bkeane, line 86: oops

moto/awslambda/models.py
```diff
@@ -489,9 +489,16 @@ def __init__(
                 ).hexdigest()
                 self.code_size = 0
             else:
-                uri, tag = self.code["ImageUri"].split(":")
+                if "@" in self.code["ImageUri"]:
```
bblommers, line 492: Would you mind adding a test for this usecase? It seems pretty straightforward, but still good to verify that creating a function succeeds using this config.

bblommers (approved):
LGTM! Thank you for expanding Moto's functionality here, @bkeane  - much appreciated.

PR #5757, opened 2022-12-11
Validate Password function is added

aarushisoni:


moto/cognitoidp/models.py
```diff
@@ -39,6 +42,10 @@ class UserStatus(str, enum.Enum):
     UNCONFIRMED = "UNCONFIRMED"
     RESET_REQUIRED = "RESET_REQUIRED"
 
+class InvalidPasswordException(Exception):
```
bblommers, line 45: Please add this Exception to `exceptions.py`, having all exceptions in one place makes it much more readable. 

Also please check it follows the same standard as the other exceptions in that file: extend JSONRestError, passing a message that explains what's going wrong (`This password was invalid`), etc.

aarushisoni: Update exceptions.py #5758 
kindly review this.

bblommers:
It looks like the build is failing - see https://github.com/spulec/moto/actions/runs/3672930949/jobs/6212952807

You can run the following commands locally, to see if it would pass:


```
flake8 moto/cognitoidp
black --check moto/cognitoidp
pylint moto/cognitoidp
mypy moto/cognitoidp
pytest tests/test_cognitoidp
```

You could also run these commands for the entire codebase, but that would take a bit more time:

```
make test
```

bblommers (approved):
Awesome - thanks @aarushisoni!

PR #5758, opened 2022-12-12
Update exceptions.py

aarushisoni:
A class for invalid password exception has been added to this file.


bblommers:
This looks OK, but it needs to be part of #5757. If somebody wants to understand this change later on, and does not have any context, they should be able to see all the changes in one place/PR.

PR #5760, opened 2022-12-12
added condition to edit user details

ArpanShah2k:
Added email condition to update user attributes


bblommers (changes requested):
This change doesn't address the issue, as far as I can tell. The ticket (#5271) specifically talks about `[updating] the email address of a user to one that already exists in the user pool`

The change would probably be in `admin_update_user_attributes`, and would look something like:
```
for user in users:
    if user has new_email_address:
        raise
````

We'll need to have a test case for this as well, so a test where
 - a user is created with email1
 - user2 is created
 - calling `admin_update_user_attributes(user2, email1)` fails

bblommers:
Closing this in favour of #5876

Issue #5762, opened 2022-12-12
Route 53's create_hosted_zone mock does not align with AWS behaviour w.r.t. SOA record


codeocelot:
## Reporting Bugs

[According to AWS docs,](https://docs.aws.amazon.com/Route53/latest/DeveloperGuide/SOA-NSrecords.html) "For each public hosted zone that you create, Amazon Route 53 automatically creates a name server (NS) record and a start of authority (SOA) record".   Moto currently does not create an SOA record when `create_hosted_zone` is called.  This makes it awkward to use moto in places where one does expect an SOA record to be created when the hosted zone is created.

Luckily, this doesn't look like a hard thing to add to the project.  If this is something you'd like to have in moto, I'm happy to open a PR with this change.

bblommers:
A PR would be very welcome @codeocelot! We have some docs for new developers on how to get started: http://docs.getmoto.org/en/latest/docs/contributing/index.html

Let me know if you have any questions around this.

codeocelot:
No problem @bblommers.  Just submitted #5763.  And to you a heartfelt thank you for maintaining this project, it's a great bit of code.

PR #5763, opened 2022-12-13
Create SOA in create_hosted_zone

codeocelot:
According to AWS Route 53 docs, an SOA record will be automatically added to any hosted zone created with create_hosted_zone.

https://docs.aws.amazon.com/Route53/latest/DeveloperGuide/SOA-NSrecords.html

See issue #5762

### DIFF
--- a/moto/batch/models.py
+++ b/moto/batch/models.py
@@ -726,7 +726,7 @@ class Job(threading.Thread, BaseModel, DockerModel, ManagedState):
                     )
                     date = unix_time_millis(date_obj)
                     logs.append({"timestamp": date, "message": line.strip()})
-                logs = sorted(logs, key=lambda l: l["timestamp"])
+                logs = sorted(logs, key=lambda log: log["timestamp"])
 
                 # Send to cloudwatch
                 log_group = "/aws/batch/job"
--- a/moto/route53/models.py
+++ b/moto/route53/models.py
@@ -480,14 +480,28 @@ class Route53Backend(BaseBackend):
             comment=comment,
             delegation_set=delegation_set,
         )
+        # For each public hosted zone that you create, Amazon Route 53 automatically creates a name server (NS) record
+        # and a start of authority (SOA) record.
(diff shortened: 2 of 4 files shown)


moto/batch/models.py
```diff
@@ -726,7 +726,7 @@ def run(self) -> None:
                     )
                     date = unix_time_millis(date_obj)
                     logs.append({"timestamp": date, "message": line.strip()})
-                logs = sorted(logs, key=lambda l: l["timestamp"])
+                logs = sorted(logs, key=lambda log: log["timestamp"])
```
codeocelot, line 729: pylint told me to do this 😬

bblommers (commented):
Thanks for the PR, @codeocelot! 

This change looks good to me - we'll just have to figure out why the Terraform test suddenly fails.

Some quick debugging tells me that TF creating a few CNAME records, but then fails to delete them afterwards. Without adding the SOA record, TF deletes them just fine... 
Will continue to look into this later.

moto/route53/models.py
```diff
@@ -480,14 +480,28 @@ def create_hosted_zone(
             comment=comment,
             delegation_set=delegation_set,
         )
+        # For each public hosted zone that you create, Amazon Route 53 automatically creates a name server (NS) record
+        # and a start of authority (SOA) record.
+        # https://docs.aws.amazon.com/Route53/latest/DeveloperGuide/SOA-NSrecords.html
+        soa_record_set = {
+            "Name": f"{name}.",
```
bblommers, line 487:
```suggestion
            "Name": f"{name}" + ("" if name.endswith(".") else "."),
```
The name sometimes already ends with a dot, causing the record set to look like `<Name>hrkd1lvc.test..</Name>`. 
That in turn makes it impossible to delete the record, because TF will only look for `hrkd1lvc.test.`

bblommers (approved):
LGTM - thanks again for adding this feature @codeocelot!

PR #5765, opened 2022-12-13
S3: Return checksum header only if set

viren-nadkarni:
### Summary

The `x-amz-sdk-checksum-algorithm` header was being returned with an empty `''` value.

This PR includes a fix so that it is only returned if also set in the request.

### Related

Regression of #5735


bblommers (approved):
Good catch @viren-nadkarni - thanks for the fix!

PR #5767, opened 2022-12-13
DynamoDB: Allow removal of multiple listitems

### LINKED ISSUE #5748: [dynamodb] Incorrect handling of array indices when removing multiple items from a list attribute
## [dynamodb] Incorrect handling of array indices when removing multiple items from a list attribute

We have noticed an issue with the dynamodb mock implementation related to update-queries with a remove-expression, in particular when multiple entries are removed from the same array:
For a remove expression like `REMOVE RelatedItems[1], RelatedItems[2]`, the entries at indexes 1 and 2 should be removed from the array (see [AWS Docs](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Expressions.UpdateExpressions.html#Expressions.UpdateExpressions.SET), I have also verified the expected behavior against the DynamoDB API).

Instead of removing the entries at index 1 and 2, moto will first remove the entry with index 1, then shift the remaining array entries and then remove the entry at index 2 (which previously was at index 3).

I have been able to reproduce this with moto 4.0.11 an boto3 1.26.23 using the following script:
```python
import boto3

@mock_dynamodb
def test_multiple_remove_operations():
    session = boto3.Session()
    dynamodb = session.resource("dynamodb")
    table_name = "remove-test"

    dynamodb.create_table(
        TableName=table_name,
        KeySchema=[
            {"AttributeName": "key", "KeyType": "HASH"},
        ],
        AttributeDefinitions=[
            {"AttributeName": "key", "AttributeType": "S"},
        ],
    )

    table = dynamodb.Table(table_name)
    table.put_item(
        Item={
            "key": "woop",
            "bla": ["1", "2", "3", "4", "5"],
        },
    )

    table.update_item(Key={"key": "woop"}, UpdateExpression="REMOVE bla[0], bla[1], bla[2]")

    result = table.get_item(Key={"key": "woop"})
    item = result["Item"]
    assert item["bla"] == ["4", "5"]
```

This results in the following AssertionError:
```
E       AssertionError: assert ['2', '4'] == ['4', '5']
E         At index 0 diff: '2' != '4'
E         Full diff:
E         - ['4', '5']
E         + ['2', '4']
```

bblommers:
Fixes #5748

### DIFF
--- a/moto/dynamodb/models/__init__.py
+++ b/moto/dynamodb/models/__init__.py
@@ -1588,6 +1588,7 @@ class DynamoDBBackend(BaseBackend):
                 table=table,
             )
             validated_ast = validator.validate()
+            validated_ast.normalize()
             try:
                 UpdateExpressionExecutor(
                     validated_ast, item, expression_attribute_names
--- a/moto/dynamodb/parsing/ast_nodes.py
+++ b/moto/dynamodb/parsing/ast_nodes.py
@@ -23,10 +23,10 @@ class Node(metaclass=abc.ABCMeta):
 
     def validate(self):
         if self.type == "UpdateExpression":
-            nr_of_clauses = len(self.find_clauses(UpdateExpressionAddClause))
+            nr_of_clauses = len(self.find_clauses([UpdateExpressionAddClause]))
             if nr_of_clauses > 1:
                 raise TooManyAddClauses()
-            set_actions = self.find_clauses(UpdateExpressionSetAction)
+            set_actions = self.find_clauses([UpdateExpressionSetAction])
             # set_attributes = ["attr", "map.attr", attr.list[2], ..]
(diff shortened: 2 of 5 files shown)


bpandola:
@bblommers Just wanted to give you a virtual pat on the back for this one.  I looked at the issue (briefly) when it came in and couldn't really come up with a strategy that didn't involve re-writing a huge portion of the parser/visitor code.  I was trying to avoid walking the tree more than once--which your code does--but in thinking about it more I really feel like you made exactly the right trade-offs given the (pre)existing code.  Anyway, in short: Great job!

bblommers:
Thanks @bpandola! It took me some time to wrap my head around as well, but this feels like the cleanest solution (aside from a complete rewrite 🙂 ).

PR #5771, opened 2022-12-14
Feature for SageMaker: create_pipeline, list_pipelines

stiebels:
Added the following mocking functionality for `sagemaker`:
- `create_pipeline`
- `list_pipelines`

### DIFF
--- a/moto/sagemaker/models.py
+++ b/moto/sagemaker/models.py
@@ -1822,14 +1822,22 @@ class SageMakerModelBackend(BaseBackend):
                 pipelines_fetched,
             )
 
+        def format_time(x):
+            return (
+                x
+                if isinstance(x, str)
+                else datetime.fromtimestamp(x).strftime("%Y-%m-%d " "%H:%M:%S")
+            )
+
         if created_after is not None:
             pipelines_fetched = filter(
-                lambda x: x.creation_time > created_after, pipelines_fetched
+                lambda x: x.creation_time > format_time(created_after),
+                pipelines_fetched,
             )
 
         if created_before is not None:
             pipelines_fetched = filter(
-                lambda x: x.creation_time < created_before,
+                lambda x: x.creation_time < format_time(created_before),
                 pipelines_fetched,
             )
 
--- a/moto/sagemaker/models.py
+++ b/moto/sagemaker/models.py
@@ -105,17 +105,6 @@ class FakePipeline(BaseObject):
         self.last_modified_time = now_string
         self.last_execution_time = now_string
 
-    @property
-    def response_object(self):
-        response_object = self.gen_response_object()
-        return {
-            k: v for k, v in response_object.items() if v is not None and v != [None]
-        }
-
-    @property
-    def response_create(self):
-        return {"PipelineArn": self.pipeline_arn}
-
 
 class FakeProcessingJob(BaseObject):
     def __init__(
--- a/moto/sagemaker/models.py
+++ b/moto/sagemaker/models.py
@@ -74,6 +74,49 @@ class BaseObject(BaseModel):
         return self.gen_response_object()
 
 
+class FakePipeline(BaseObject):
+    def __init__(
+        self,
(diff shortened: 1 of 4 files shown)


moto/sagemaker/models.py
```diff
@@ -74,6 +74,49 @@ def response_object(self):
...
+        self.parallelism_configuration = parallelism_configuration
+        self.pipeline_definition_s3_location = pipeline_definition_s3_location
+
+        now_string = datetime.now().strftime("%Y-%m-%d %H:%M:%S")
+        self.creation_time = now_string
+        self.last_modified_time = now_string
+        self.last_execution_time = now_string
+
+    @property
+    def response_object(self):
```
bblommers, line 109: The `response_object` and `response_create` arent' used, as far as I can tell, so let's remove them

tests/test_sagemaker/test_sagemaker_pipeline.py
```diff
@@ -0,0 +1,176 @@
...
+    assert isinstance(response, dict)
+    assert response["PipelineArn"].should.equal(
+        arn_formatter("pipeline", fake_pipeline_name, ACCOUNT_ID, TEST_REGION_NAME)
+    )
+
+
+def test_list_pipelines_none(sagemaker_client):
+    response = sagemaker_client.list_pipelines()
+    assert isinstance(response, dict)
+    assert response["PipelineSummaries"].should.be.empty
```
bblommers, line 50:
```suggestion
    response["PipelineSummaries"].should.be.empty
```
If you use the `should`, you don't need to assert - asserting whatever it `should` be is automatic.

There are quite a few assertions like that - I won't point them all out, but ideally we remove all of them.

tests/test_sagemaker/test_sagemaker_pipeline.py
```diff
@@ -0,0 +1,176 @@
...
+
+    created_before_timestamp = datetime.timestamp(created_before_datetime)
+    response = sagemaker_client.list_pipelines(CreatedBefore=created_before_timestamp)
+    assert response["PipelineSummaries"].should.be.empty
+
+
+def test_list_pipelines_invalid_values(sagemaker_client):
+    with pytest.raises(botocore.exceptions.ClientError):
+        _ = sagemaker_client.list_pipelines(MaxResults=200)  # Must be <= 100
+        _ = sagemaker_client.list_pipelines(NextToken="some-invalid-next-token")
```
bblommers, line 174: This doesn't work unfortunately. After the first command fails, we immediately jump out of the context, so the next three statements are ignored.

So this should either follow:
```
with pytest.raises():
    ....
with pytest.raises():
    ...
# etc
```
or we use `pytest.mark.parametrize` to execute the same test multiple times. I'll leave it up to you what works best!

bblommers (changes requested):
Hi @stiebels, thanks for the PR! Looks good overall, just got a few minor nitpicks.

moto/sagemaker/models.py
stiebels: Thanks for the review and the comments! All change requests implemented.

tests/test_sagemaker/test_sagemaker_pipeline.py
stiebels: flake8 only gives me a pass when having an `assert` before `.should.be.empty` statements. For other `should.have` (e.g. `should.have.length_of`), no assert seems to be required.

See here the workflow run on my forked repo: https://github.com/stiebels/moto/actions/runs/3709067440/jobs/6287306243

Therefore, I've left the `assert` in for `should.be.empty` and removed it where it doesn't seem to be needed.

bblommers (approved):
LGTM - Thanks again for adding these feature @stiebels!

stiebels:
> LGTM - Thanks again for adding these feature @stiebels!

Thanks! Feel free to merge any time ("Only those with write access to this repository can merge pull requests.").

Issue #5769, opened 2022-12-14
sqs not able to receive messages with message attributes


rita005:
Hi.

When using:
    @aws-sdk for c#
    moto-server using the docker image

We are trying to mock sqs for integration testing in our api. We are able to create and connect to it, but when we try to receive messages with attributes, the ReceiveMessagesAsync hangs without giving any exception. We have tried to send messages with no message attributes and with different attributes than the requested ones and we are able to receive the messages.

We are creating our sqs via AWS cli:
```
aws sqs create-queue --queue-name test --endpoint-url http://localhost:8090/
```

Here is our API code to receive the messages:

```
var attributesList = new List<string> { "attribute1" };
var request = new ReceiveMessageRequest
            {
                QueueUrl = queueUrl,
                MaxNumberOfMessages = maxMessages,
                WaitTimeSeconds = waitTimeSeconds,
                MessageAttributeNames = attributesList
            };
var messageResponse = await _sqsClient.ReceiveMessageAsync(request);
```

Here is the message sent:

```
aws sqs send-message --queue-url http://localhost:8090/123456789012/test --message-body file://message-body.json --delay-seconds 10 --message-attributes file://message-attributes.json --endpoint-url http://localhost:8090/ 
```

Our message-body file:
```
{"Id": "81507a96-950c-41bb-967b-5811e1574e4f"}
```

And our message-atributes file:
```
{ "attribute1":{ "DataType":"String","StringValue":"att1"} }
```

It's important to notice that this workflow has the expected behavior when using aws sqs.

bblommers:
Thanks for raising this @rita005 - this typically happens when Moto does not return a response in the exact same format as AWS.  
The AWS CLI and boto3 are quite forgiving in that respect, but other SDK's not quite.

I'll do some debugging to see if I can find out what's happening.

rita005:
Tested the latest version with the c# aws sdk and it's working perfectly.
Thank you for your work!

PR #5773, opened 2022-12-14
SQS: Align receive_message() response with AWS

bblommers:
Potential fix for #5769 

`{{ value.string_value|e }}` results in HTML-encoded output: `StringValue>&lt;att1&gt;</StringValue>`, which is equivalent to what AWS sends.

Incidentally, this changes the computation of `MD5OfMessageAttributes`, which computes the wrong value when the `StringValue` has the additional `CDATA`-stuff. 
I presume this is what trips up the C#-SDK, as reported in the linked issue - the MD5 computation.

(The other changes - adding `<?xml..>` and the RequestID - are just for sake of completeness. I don't imagine C# cares much, but you never know.)

### DIFF
--- a/moto/sqs/responses.py
+++ b/moto/sqs/responses.py
@@ -583,7 +583,8 @@ SEND_MESSAGE_RESPONSE = """<SendMessageResponse>
     </ResponseMetadata>
 </SendMessageResponse>"""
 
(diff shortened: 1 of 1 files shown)


bblommers:
This is now part of moto >= 4.0.12.dev30

PR #5775, opened 2022-12-15
Techdebt: replace mock with unittest.mock

yan12125:
From https://pypi.org/project/mock/,

> mock is now part of the Python standard library, available as
> unittest.mock in Python 3.3 onwards.
>
> This package contains a rolling backport of the standard library mock
> code compatible with Python 3.6 and up.

As moto already requires Python >= 3.6 [1], usage of the backport library is not necessary and can be replaced with the one from the standard library.

[1] https://github.com/spulec/moto/blob/4.0.11/setup.py#L141

---

This is part of efforts on dropping python-mock package in Arch Linux: https://archlinux.org/todo/drop-python-mock-checkdepends/

### DIFF
--- a/tests/test_cognitoidentity/test_cognitoidentity.py
+++ b/tests/test_cognitoidentity/test_cognitoidentity.py
@@ -1,5 +1,5 @@
 import boto3
-import mock
+from unittest import mock
 import sure  # noqa # pylint: disable=unused-import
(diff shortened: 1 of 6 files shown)


bblommers (approved):
LGTM! Thanks for fixing this @yan12125

PR #5779, opened 2022-12-16
Fix some Batch error message typos

futurulus:
Trying to cancel a job with an empty `job_id` results in an error message complaining about `reason` (and vice versa).


bblommers (approved):
Nice catch @futurulus, thanks for fixing this!

PR #5788, opened 2022-12-19
fix(route53): Allow upsert for resource records that already exist.

pawelrubin:
From [AWS API Reference](https://docs.aws.amazon.com/Route53/latest/APIReference/API_ChangeResourceRecordSets.html) about the UPSERT action:

> `UPSERT`: If a resource set exists Route 53 updates it with the values in the request.

Thus, providing a resource record set that already exist to the UPSERT
action should succeed.

Additional changes in tests:
- Add check for error message in the existing unit test.
- Add unit test for creating a resource record with an UPSERT.

### DIFF
--- a/moto/route53/exceptions.py
+++ b/moto/route53/exceptions.py
@@ -174,11 +174,11 @@ class DnsNameInvalidForZone(Route53ClientError):
         super().__init__("InvalidChangeBatch", error_msg)
 
 
-class ChangeSetAlreadyExists(Route53ClientError):
+class ResourceRecordAlreadyExists(Route53ClientError):
     code = 400
 
-    def __init__(self, action: str, name: str, _type: str):
+    def __init__(self, name: str, _type: str):
         super().__init__(
             "InvalidChangeBatch",
-            f"Tried to {action.lower()} resource record set [name='{name}', type='{_type}'] but it already exists",
+            f"Tried to create resource record set [name='{name}', type='{_type}'] but it already exists",
         )
--- a/moto/route53/models.py
+++ b/moto/route53/models.py
@@ -19,7 +19,7 @@ from moto.route53.exceptions import (
     PublicZoneVPCAssociation,
     QueryLoggingConfigAlreadyExists,
     DnsNameInvalidForZone,
-    ChangeSetAlreadyExists,
+    ResourceRecordAlreadyExists,
     InvalidInput,
 )
 from moto.core import BaseBackend, BackendDict, BaseModel, CloudFormationModel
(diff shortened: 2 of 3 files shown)


bblommers:
The original issue requested validation for CREATE and DELETE requests, see InvalidChangeBatch.
And that makes sense to me - one wouldn't delete a record twice. Are you aware of any documentation on whether duplicate UPSERTS are allowed, @pawelrubin? I briefly checked, but I couldn't find anything, and I do not have the option to verify this with AWS atm.

Also: the new test only upserts once, so it does not verify the logic change.

pawelrubin:
Hi @bblommers,
thank you for the quick response. I've rephrased the commit message to better address the issue, please have another look.

Subsequent UPSERTs with the same changes are indeed allowed. Making the next call to `ChangeResourceRecordSets` with the same changes as the previous one is essentially a no-op.

bblommers (changes requested):
Thanks for making the change @pawelrubin. The only other thing is that DELETE's should also not be allowed. So `if value["Action"] == "CREATE"` should probably be `if value["Action"] != "UPSERT"`, or `if value["Action"] in ["CREATE", "DELETE"]`

pawelrubin:
Subsequent deletes are handled later on in the codebase in [here](https://github.com/spulec/moto/blob/bdef72d481fe1e80ec3031108070b29469f03086/moto/route53/models.py#L574) - trying to delete the same record set twice will result in an error.

The `ResourceRecordAlreadyExists` error makes sense only in the `CREATE` case.

bblommers:
> Subsequent deletes are handled later on in the codebase in [here](https://github.com/spulec/moto/blob/bdef72d481fe1e80ec3031108070b29469f03086/moto/route53/models.py#L574) - trying to delete the same record set twice will result in an error.
> 

I've just tested it out in your branch - it does not actually throw an error. The logic you pointed out checks whether our _change log_ contains a CREATE/UPSERT record, i.e. whether it has been created it any point. But the change log is append-only, so even if a record has been deleted, the changelog will still contain the original CREATE. 

However, the fact that we rely on our change log to determine whether a record has been created before is actually incorrect. This makes it impossible to recreate a previously deleted record, which (I assume) is supported by AWS.

I'm going to go ahead and merge this for now, as it does fix an existing bug. Let's refactor this later on, and then we can validate new changes against the records that are currently active, instead of looking at a change log that may contain outdated information.

Thank you again for the PR @pawelrubin!

pawelrubin:
Oh, you're right. I incorrectly assumed that the logic mentioned above was correct. 
Thank you for a thorough review.

pawelrubin:
@bblommers what does the release schedule look like? Can we expect the `4.0.13` version to be released anytime soon?

bblommers:
> @bblommers what does the release schedule look like? Can we expect the `4.0.13` version to be released anytime soon?

Hi @pawelrubin, I don't have a lot of time over Christmas, so the next release will probably happen in the first week of January.

pawelrubin:
Got it. Merry Christmas Bert! 🎅🏻

PR #5793, opened 2022-12-20
support statistic-values in put-metric-data

steffyP:
This PR adds support for the `statistic-values` parameter for the `put-metric-data` operation in `cloudwatch`. 

It has initially been reported a while ago in #1615, but there was no actual implementation provided (just a fix for the `KeyError`). Recently, an issue with `statistic-values` has been reported for [LocalStack](https://github.com/localstack/localstack/issues/7263) as well.

### DIFF
--- a/moto/cloudwatch/models.py
+++ b/moto/cloudwatch/models.py
@@ -1,5 +1,4 @@
 import json
-import statistics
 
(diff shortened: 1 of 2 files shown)


bblommers (approved):
This looks really good. It's quite the refactor, but storing the MetricDatum and MetricAggregrateDatum separately makes a lot of sense.

Thanks @steffyP!

PR #5797, opened 2022-12-21
Add sagemaker mock call: describe_pipeline

stiebels:
- Adding sagemaker mock call `describe_pipeline`. It's the only remaining API call directly related to a SageMaker Pipeline (`PipelineExecution` and related mocks are then still missing).
`boto3.client("sagemaker").describe_pipeline`
https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/sagemaker.html#SageMaker.Client.describe_pipeline

- Adding check that `PipelineDefinition` OR `PipelineDefinitionS3Location` is passed to `create_pipeline`. Although the docs do not say so, it is actually a mandatory parameter (one of `{PipelineDefinition, PipelineDefinitionS3Location}` is mandatory).
- Adding support for `PipelineDefinitionS3Location` and additional tests covering this case

### DIFF
--- a/moto/sagemaker/models.py
+++ b/moto/sagemaker/models.py
@@ -1,4 +1,3 @@
-import boto3
 import json
 import os
 from datetime import datetime
@@ -11,6 +10,7 @@ from .exceptions import (
     AWSValidationException,
     ResourceNotFound,
 )
+from .utils import load_pipeline_definition_from_s3
 
 
 PAGINATION_MODEL = {
@@ -1789,12 +1789,9 @@ class SageMakerModelBackend(BaseBackend):
             )
 
         if pipeline_definition_s3_location:
-            client = boto3.client("s3")
-            result = client.get_object(
-                Bucket=pipeline_definition_s3_location["Bucket"],
-                Key=pipeline_definition_s3_location["ObjectKey"],
+            pipeline_definition = load_pipeline_definition_from_s3(
+                pipeline_definition_s3_location
             )
-            pipeline_definition = json.loads(result["Body"].read())
 
         pipeline = FakePipeline(
             pipeline_name,
@@ -1836,24 +1833,24 @@ class SageMakerModelBackend(BaseBackend):
                 message=f"Could not find pipeline with name {pipeline_name}."
             )
 
-        provided_kwargs = set(kwargs.keys())
-        allowed_kwargs = {
-            "pipeline_display_name",
-            "pipeline_definition",
-            "pipeline_definition_s3_location",
-            "pipeline_description",
-            "role_arn",
-            "parallelism_configuration",
-        }
-        invalid_kwargs = provided_kwargs - allowed_kwargs
-
-        if invalid_kwargs:
-            raise TypeError(
-                f"update_pipeline got unexpected keyword arguments '{invalid_kwargs}'"
+        if all(
+            [
+                kwargs.get("pipeline_definition"),
+                kwargs.get("pipeline_definition_s3_location"),
+            ]
+        ):
+            raise ValidationError(
+                "An error occurred (ValidationException) when calling the UpdatePipeline operation: "
+                "Both Pipeline Definition and Pipeline Definition S3 Location shouldn't be present"
             )
 
         for attr_key, attr_value in kwargs.items():
             if attr_value:
+                if attr_key == "pipeline_definition_s3_location":
+                    self.pipelines[
+                        pipeline_name
+                    ].pipeline_definition = load_pipeline_definition_from_s3(attr_value)
+                    continue
                 setattr(self.pipelines[pipeline_name], attr_key, attr_value)
 
         return pipeline_arn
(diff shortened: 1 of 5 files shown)


moto/sagemaker/models.py
```diff
@@ -86,24 +87,42 @@ def __init__(
...
         region_name,
         parallelism_configuration,
-        pipeline_definition_s3_location,
     ):
         self.pipeline_name = pipeline_name
         self.pipeline_arn = arn_formatter(
             "pipeline", pipeline_name, account_id, region_name
         )
-        self.pipeline_display_name = pipeline_display_name
+        self.pipeline_display_name = pipeline_display_name or pipeline_name
```
stiebels, line 95: Default `pipeline_display_name` is `pipeline_name` (if no other specified)

moto/sagemaker/models.py
```diff
@@ -86,24 +87,42 @@ def __init__(
...
         self.parallelism_configuration = parallelism_configuration
-        self.pipeline_definition_s3_location = pipeline_definition_s3_location
 
         now_string = datetime.now().strftime("%Y-%m-%d %H:%M:%S")
         self.creation_time = now_string
         self.last_modified_time = now_string
-        self.last_execution_time = now_string
+        self.last_execution_time = None
+
+        self.pipeline_status = "Active"
```
stiebels, line 107: Default state is `Active` if pipeline has only been upserted but not started.

moto/sagemaker/utils.py
```diff
@@ -0,0 +1,16 @@
+import boto3
+import json
+
+
+def load_pipeline_definition_from_s3(pipeline_definition_s3_location):
+    client = boto3.client("s3")
```
bblommers, line 6: It will be quicker to use the internal API to retrieve data from other services:
```
from moto.s3.models import s3_backends
s3_backend = s3_backends[account_id]["global"]
s3_backend.get_object(...)
```

stiebels: Thanks!

stiebels: I've been testing this implementation variant. However, I keep running into retrieving a `FakeKey` object from `get_object` then, which I cannot retrieve the original content from (at least I haven't found out how yet). Therefore, I can't really mimick the original chain of events. I'm wondering what I am not seeing or doing right here when it comes to handling the retrieved FakeKey via `get_object`?

https://github.com/spulec/moto/actions/runs/3756496788/jobs/6382939745

bblommers: You can use the value-attribute: `json.loads(key.value)`

https://github.com/spulec/moto/blob/master/moto/s3/models.py#L162

We don't have type-annotations yet for S3, but this property should return the original content as `bytes`

stiebels: Thanks, now I got it working!

stiebels: Unfortunately, I can't seem to get the Server Mode tests working as the mocked S3 buckets can't be found... Do you perhaps have pointer here?

bblommers: Hi @stiebels, apologies for the delay. In ServerMode, Moto runs inside a separate Python runtime, making it impossible to reach the S3 data directly. You would have to use boto3 to download the pipeline definition (`boto3.client("s3").download_object(..)`)

stiebels: Thanks! Hope you've had a nice christmas break!

stiebels:
The one test that is failing is unrelated to this PR.

tests/test_sagemaker/test_sagemaker_pipeline.py
```diff
@@ -21,29 +39,90 @@ def fixture_sagemaker_client():
...
+    object_key = "some/object/key.json"
+    pipeline_definition = {"key": "value"}
+
+    with mock_s3():
+        with setup_s3_pipeline_definition(
+            bucket_name,
+            object_key,
+            pipeline_definition,
+        ):
+            observed_pipeline_definition = load_pipeline_definition_from_s3(
```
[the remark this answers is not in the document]
stiebels: Not sure if I understand your here. In the other comment you're advising to use `s3_backend` within `load_pipeline_definition_from_s3`. Here you're advising to use `boto3` directly when testing `load_pipeline_definition_from_s3`. However, if the implementation of that function uses `s3_backend` based on your advice, I'd have to re-implement the logic using `boto3` in the test here, which would result in not really testing the `load_pipeline_definition_from_s3` function.

Or did I get you wrong?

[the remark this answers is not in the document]
bblommers: Ah, my bad - I should read better before commenting!
You're interpretation is correct - `s3_backend` within Moto for speed, `boto3` within tests for compatibility. 

My advice was based on the (incorrect) assumption that this was an 'integration'-style test, where we create the pipeline-definition using boto3's `create_pipeline(PipelineDefinitionS3Location=..)`, and then we can describe it again using boto3.

If we only test the `load_pipeline_definition_from_s3`-function, we can keep it as is. We'll just have to skip this test in ServerMode, as we do not have access to s3_backend there:
```
if settings.TEST_SERVER_MODE:
    raise SkipTest("..")
```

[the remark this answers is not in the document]
stiebels: Clear now and done! Thanks

bblommers (approved):
Awesome - thank you @stiebels!

PR #5800, opened 2022-12-21
changing error message to match aws api (stack update attempt with no updates)

imayobrown:
When a stack exists and an update is attempted with the same parameters and template body the AWS api responds with a specific message which is injected into the `ClientError` generated by boto.

To reduce complexity we make some of our processes ignorant of whether an update constitutes changes to a stack by checking if this exact message is returned as part of a `ClientError`. Reflecting this message in moto will allow us test the components that use this in their processes while preserving the the updates provided in https://github.com/spulec/moto/pull/4351.

### DIFF
--- a/moto/cloudformation/responses.py
+++ b/moto/cloudformation/responses.py
@@ -367,7 +367,7 @@ class CloudFormationResponse(BaseResponse):
             new_params = self._get_param_values(incoming_params, old_stack.parameters)
             if old_stack.template == stack_body and old_stack.parameters == new_params:
                 raise ValidationError(
-                    old_stack.name, message=f"Stack [{old_stack.name}] already exists"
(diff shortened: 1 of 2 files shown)


bblommers:
It looks like one of our tests expects the wrong value, and is now failing:
https://github.com/spulec/moto/actions/runs/3753624731/jobs/6377473172

Happy to merge once the build passes.

imayobrown:
> It looks like one of our tests expects the wrong value, and is now failing: https://github.com/spulec/moto/actions/runs/3753624731/jobs/6377473172
> 
> Happy to merge once the build passes.

My apologies, I missed updating the tests with the expected value. The most recent commit should solve the problem.

bblommers (approved):
LGTM - thanks for fixing this @imayobrown!

PR #5802, opened 2022-12-22
chore(readme): Align code snippets formatting with project's rules.

pawelrubin:
- Use black formatting
- Since Python 2.* is no longer supported, don't inherit from `object`


bblommers (approved):
LGTM! Thanks for bringing this uptodate @pawelrubin

PR #5803, opened 2022-12-22
Add tags for users in the get-account-authorization-details response

dfangl:
## Problem

Currently the tags are missing for users, when running get-account-authorization-details.
It is present for roles, however.

Related localstack/localstack/issues/5230

## Solution
* Add tags to response
* Test tags in response

## Additional comments
It seems fine that an empty array is indeed returned without tags, as it is in parity with aws.
AWS response:
```
        {
            "Path": "/",
            "UserName": "test-user",
            "UserId": "AIDAXXXX",
            "Arn": "arn:aws:iam::XXXXXXXXXXXX:user/test-user",
            "CreateDate": "2022-12-22T16:30:32+00:00",
            "GroupList": [],
            "AttachedManagedPolicies": [],
            "Tags": []
        }
```

### DIFF
--- a/moto/iam/responses.py
+++ b/moto/iam/responses.py
@@ -849,6 +849,7 @@ class IamResponse(BaseResponse):
             groups=account_details["groups"],
             roles=account_details["roles"],
             get_groups_for_user=self.backend.get_groups_for_user,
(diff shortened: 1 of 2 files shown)


bblommers (approved):
LGTM!

PR #5804, opened 2022-12-27
docs: fix patching external client/resource docs

cooncesean:
## What 
The documentation sample code [currently in master](https://docs.getmoto.org/en/latest/docs/getting_started.html?highlight=caveats#patching-the-client-or-resource) runs into a "NameError: name `client` is not defined" exception if executed. 

This change updates the documentation to refer to the externally provided `client` as `outside_client` for example clarity + functional accuracy.


bblommers (approved):
Nice catch @cooncesean - thanks for fixing this!

cooncesean:
@bblommers no worries; happy to help. Thank YOU guys for everything y'all have been doing to build + maintain this library and provide so much value + support to the community using it. 🙏

PR #5805, opened 2022-12-28
Add TargetKeyId to default kms key alias response.

bkeane:
Small "fix" to add in TargetKeyIds to canonical aliased kms keys.

These keys are a curiosity to me. I can't seem to find a way to initialize them via boto3 in actual AWS land. If you setup, say, ssm and add a first key to it using the service kms key it auto-generates. \o/.

### DIFF
--- a/moto/kms/responses.py
+++ b/moto/kms/responses.py
@@ -2,11 +2,10 @@ import base64
 import json
 import os
 import re
-import uuid
 import warnings
 
 from moto.core.responses import BaseResponse
-from moto.kms.utils import RESERVED_ALIASES
+from moto.kms.utils import RESERVED_ALIASES, RESERVED_ALIASE_TARGET_KEY_IDS
 from .models import kms_backends, KmsBackend
 from .policy_validator import validate_policy
 from .exceptions import (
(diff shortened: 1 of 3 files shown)


moto/kms/responses.py
```diff
@@ -278,9 +279,11 @@ def list_aliases(self):
                 a for a in response_aliases if a["AliasName"] == reserved_alias
             ]
             if not exsisting:
+                arn = f"arn:aws:kms:{region}:{self.current_account}:{reserved_alias}"
                 response_aliases.append(
                     {
-                        "AliasArn": f"arn:aws:kms:{region}:{self.current_account}:{reserved_alias}",
+                        "TargetKeyId": str(uuid.UUID(int=abs(hash(arn)), version=4)),
```
bblommers, line 285: The `hash`-function is only guaranteed to give the same result within the same Python runtime. 
That probably explains the test failure - the hash as calculated during the test is in a different runtime than the Moto server, and therefore yields a different result.

Does it make sense to use a random UUID instead? Having a reproducible value is nice for our tests, but this is not guaranteed in AWS as far as I know.

bkeane: True story. Admittedly I addressed my thought in an overly complicated manner now that I'm looking back at it.

They will be different across accounts for sure, but what I wanted to avoid was getting a random uuid back per request. Within one account, the same id's ought to be returned.

I will go look at it and see if I can't come up with a simpler approach!

bblommers (approved):
Makes sense - thanks for the PR @bkeane!

Issue #5814, opened 2023-01-04
ecs.start_task and ecs.describe_tasks do not handle tags correctly


shanishiri:
Facing similar to this issues but for create_task() and describe_tasks() https://github.com/spulec/moto/issues/3138, https://github.com/spulec/moto/issues/3872. When staring ECS task in cluster with "tags" argument, the task starts with empty "tags" list, same thing happen when calling describe_tasks() with include=["TAGS"].

Create ECS task in cluster:
```
client.start_task(
        cluster="cluster_name",
        taskDefinition="task_definition",
        overrides={},
        containerInstances=["container_instance_id"],
        startedBy="moto",
        tags=[{
            'key': 'tagName',
            'value': 'TavValue'
        }]
    )
```

Actual return of start_task():
```
{"tasks": [{"id": "8f1d0248-6fb0-4e1b-bdd0-e377a900b341", "overrides": {}, "containers": [], "tags": [], "attachments": [], "clusterName": "test_ecs_cluster", "clusterArn": "arn:aws:ecs:us-east-1:123456789012:cluster/test_ecs_cluster", "containerInstanceArn": "arn:aws:ecs:us-east-1:123456789012:container-instance/test_ecs_cluster/8dcca21a-4752-42d4-950b-d083ab6d0f09", "lastStatus": "RUNNING", "desiredStatus": "RUNNING", "taskDefinitionArn": "arn:aws:ecs:us-east-1:123456789012:task-definition/test_ecs_task:1", "startedBy": "moto", "launchType": "", "stoppedReason": "", "resourceRequirements": {"CPU": 1024, "MEMORY": 400, "PORTS": [], "PORTS_UDP": []}, "regionName": "us-east-1", "taskArn": "arn:aws:ecs:us-east-1:123456789012:task/test_ecs_cluster/8f1d0248-6fb0-4e1b-bdd0-e377a900b341"}], "failures": []}
```

Expected tags in return of the describe_tasks() call with include=["TAGS"]:
```
tasks = client.describe_tasks(
                        cluster="cluster_id", 
                        tasks=["task_arn"], 
                        include=["TAGS"]
                    )
```

Actual return of describe_tasks():
```
[...]

PR #5817, opened 2023-01-04
Fix/support tags in ECS start_task and describe_tasks

shanishiri:
closes https://github.com/spulec/moto/issues/5814


bblommers (approved):
LGTM - thanks for adding this @shanishiri!

PR #5822, opened 2023-01-06
support $Latest and $Default when describing launch templates

Luukvdm:


bblommers (approved):
LGTM - thanks for adding this @Luukvdm!

Luukvdm:
I think the code coverage test is blocking the PR, but all the code that I added is being tested.

PR #5832, opened 2023-01-10
Remove redundant wheel dep from pyproject.toml

mgorny:
Remove the redundant `wheel` dependency, as it is added by the backend automatically.  Listing it explicitly in the documentation was a historical mistake and has been fixed since, see: https://github.com/pypa/setuptools/commit/f7d30a9529378cf69054b5176249e5457aaf640a


bblommers (approved):
Thanks for the fix and explanation @mgorny! I was not aware this wasn't necessary.

mgorny:
Thank you!

Issue #5838, opened 2023-01-12
Enhancement: Support PreferredMaintenanceWindow for DBInstance


mgshirali:
Currently there are no validation checks for PreferredMaintenanceWindow when working with RDS DBInstance. Add support for validating this parameter. Support various exceptions that are generated as part of validation process.

PR #5839, opened 2023-01-12
#5838: Support PreferredMaintenanceWindow for RDS DBInstance

mgshirali:
Handle `PreferredMaintenanceWindow` parameters for creation and modification of RDS DB instances. Generate the relevant AWS exceptions as part of handling the same.


bblommers (approved):
Awesome - thanks for adding this validation @mgshirali!

PR #5840, opened 2023-01-13
batch: add multinode support

d4l3k:
This adds multinode support to `mock_batch`. This will allow users to run jobs that have multiple different jobs configured.

Notes:
* describe_jobs doesn't return most container field
* describe_jobs doesn't respect the `#idx` suffix to job IDs
* all containers log to the same stream
* container exit logic is very simple and waits for them to finish in order so if replica 1 fails but 0 doesn't it may hang until timeout
* there's minimal/no validation on the job spec right now for nodeProperties

Test plan:

pyre

added multinode slow test

### DIFF
--- a/moto/batch/models.py
+++ b/moto/batch/models.py
@@ -21,6 +21,7 @@ from .exceptions import InvalidParameterValueException, ClientException, Validat
 from .utils import (
     make_arn_for_compute_env,
     make_arn_for_job_queue,
+    make_arn_for_job,
     make_arn_for_task_def,
     lowercase_first_key,
 )
@@ -221,6 +222,7 @@ class JobDefinition(CloudFormationModel):
         parameters: Optional[Dict[str, Any]],
         _type: str,
         container_properties: Dict[str, Any],
+        node_properties: Dict[str, Any],
         tags: Dict[str, str],
         retry_strategy: Dict[str, str],
         timeout: Dict[str, int],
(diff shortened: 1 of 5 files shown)


kiukchung:
this is pretty amazing! lets us test distributed AI workloads (that are submitted as batch's multi-node-parallel jobs)

moto/batch/models.py
```diff
@@ -282,6 +286,9 @@ def _get_resource_requirement(self, req_type: str, default: Any = None) -> Any:
         :return: The value of the resource requirement, or None.
         :rtype: any
         """
+        assert (
```
bblommers, line 289: If we want to validate this, we should throw the same exception that AWS throws. See the other `ClientExceptions` that are thrown in this file - that guarantees at least that the same format is in-line with AWS.

moto/batch/responses.py
```diff
@@ -215,8 +217,9 @@ def describejobs(self) -> str:
     def listjobs(self) -> str:
         job_queue = self._get_param("jobQueue")
         job_status = self._get_param("jobStatus")
+        filters = self._get_param("filters")
```
bblommers, line 220: We should have some tests here as well, to verify this parameter is handled correctly

bblommers (changes requested):
Hi @d4l3k, thanks for the PR! Please see my comments

moto/batch/models.py
d4l3k: removed, this was mostly for testing to make sure I didn't miss a spot

moto/batch/responses.py
d4l3k: done

bblommers (approved):
LGTM - thanks again for adding this feature @d4l3k!

PR #5841, opened 2023-01-13
Feature #5810: Implemented `describe_security_group_rules` for `ec2`

### LINKED ISSUE #5810: Implement Describe Security Group Rules for EC2
Currently there's no implementation of [Describe Security Group Rules](https://docs.aws.amazon.com/cli/latest/reference/ec2/describe-security-group-rules.html). I would love to take up this issue and implement it.

Here's my approch
create an `describe_security_group_rules` method in the `SecurityGroupBackend` class which resides in `moto/ec2/models/security_groups.py`

However i'm a bit uncertain on the contents on `describe_security_group_rules` method. Not sure what to return from the method.
Here is what i think

```py
def describe_security_group_rules(self, filters=None):
    all_groups = self.groups.copy()
    matches = itertools.chain(*[x.copy().values() for x in all_groups.values()])
    if filters:
        matches = [grp for grp in matches if grp.matches_filters(filters)]
    return matches
```

This gets all the security groups that matches the filters. Now i'm not sure how the get the ingress/egress rules. Would be great if someone could enlighten me on this.

Also do i need to make any additional changes so that the `describe_security_group_rules` method gets fired when the 
`aws ec2 describe-security-group-rules` cmd is run

uzaxirr:
Closes #5810 and https://github.com/localstack/localstack/issues/7245

### DIFF
--- a/moto/ec2/models/security_groups.py
+++ b/moto/ec2/models/security_groups.py
@@ -520,6 +520,9 @@ class SecurityGroupBackend:
 
         return matches
 
+    def describe_security_group_rules(self):
+        pass
+
     def _delete_security_group(self, vpc_id, group_id):
         vpc_id = vpc_id or self.default_vpc.id
         if self.groups[vpc_id][group_id].enis:
--- a/moto/ec2/models/security_groups.py
+++ b/moto/ec2/models/security_groups.py
@@ -520,22 +520,19 @@ class SecurityGroupBackend:
 
         return matches
 
-    def describe_security_group_rules(self, group_ids=None, groupnames=None, filters=None):
+    def describe_security_group_rules(self, group_ids=None, filters=None):
         matches = itertools.chain(*[x.copy().values() for x in self.groups.values()])
         if group_ids:
             matches = [grp for grp in matches if grp.id in group_ids]
             if len(group_ids) > len(matches):
                 unknown_ids = set(group_ids) - set(matches)
                 raise InvalidSecurityGroupNotFoundError(unknown_ids)
-        if groupnames:
-            matches = [grp for grp in matches if grp.name in groupnames]
-            if len(groupnames) > len(matches):
-                unknown_names = set(groupnames) - set(matches)
-                raise InvalidSecurityGroupNotFoundError(unknown_names)
         if filters:
             matches = [grp for grp in matches if grp.matches_filters(filters)]
             if not matches:
-                raise InvalidSecurityGroupNotFoundError("No security groups found matching the filters provided.")
+                raise InvalidSecurityGroupNotFoundError(
+                    "No security groups found matching the filters provided."
+                )
 
         rules = []
         for group in matches:
--- a/moto/ec2/models/security_groups.py
+++ b/moto/ec2/models/security_groups.py
@@ -533,7 +533,6 @@ class SecurityGroupBackend:
                 raise InvalidSecurityGroupNotFoundError(
                     "No security groups found matching the filters provided."
                 )
-
         rules = []
         for group in matches:
             rules.extend(group.ingress_rules)
--- a/moto/ec2/models/security_groups.py
+++ b/moto/ec2/models/security_groups.py
@@ -520,8 +520,29 @@ class SecurityGroupBackend:
 
         return matches
 
-    def describe_security_group_rules(self):
-        pass
+    def describe_security_group_rules(self, group_ids=None, groupnames=None, filters=None):
+        matches = itertools.chain(*[x.copy().values() for x in self.groups.values()])
+        if group_ids:
+            matches = [grp for grp in matches if grp.id in group_ids]
+            if len(group_ids) > len(matches):
+                unknown_ids = set(group_ids) - set(matches)
+                raise InvalidSecurityGroupNotFoundError(unknown_ids)
+        if groupnames:
+            matches = [grp for grp in matches if grp.name in groupnames]
+            if len(groupnames) > len(matches):
+                unknown_names = set(groupnames) - set(matches)
+                raise InvalidSecurityGroupNotFoundError(unknown_names)
+        if filters:
+            matches = [grp for grp in matches if grp.matches_filters(filters)]
+            if not matches:
+                raise InvalidSecurityGroupNotFoundError("No security groups found matching the filters provided.")
+
+        rules = []
+        for group in matches:
+            rules.extend(group.ingress_rules)
+            rules.extend(group.egress_rules)
+
+        return rules
 
     def _delete_security_group(self, vpc_id, group_id):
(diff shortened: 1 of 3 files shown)


uzaxirr:
Hey @ackdav @bblommers 
I've been able to figure out where to keep the view and model logic but need a bit of help here,
in `describe_security_group_rules` for  `moto/ec2/models/security_groups.py`
I'm not quite sure what to return from this method. Should i query all the group matching the filters and return all of their ingress/egress rules?

bblommers:
> Should i query all the group matching the filters and return all of their ingress/egress rules?

Yes, that makes sense. The exact format of the data doesn't matter too much, to be fair. As long as `responses.py` can take the output from `self.ec2_backend.describe_security_group_rules` and convert it into the appropriate XML response, it should be fine.

uzaxirr:
Hey @bblommers 
Thanks for the above info, I've been able to figure out the models and response logic now. But how do I proceed on figuring out the xml response which will be in `DESCRIBE_SECURITY_GROUP_RULES_RESPONSE` variable.

Also can u also take a quick look at the methods i've implemented in response and model class and let me know your thoughts

bblommers:
Looking at the documentation, it doesn't look like we can search by security group names, so I don't think that filter is necessary? But everything else looks good so far!

The easiest way to figure out the XML response is to see what AWS returns, and just copy/paste that. 
If you have security groups setup in your actual AWS account, you can call `describe_security_group_rules` against AWS and intercept the response using a proxy. Our docs describe the easiest way of setting up a proxy here: http://docs.getmoto.org/en/latest/docs/contributing/development_tips/urls.html#intercepting-aws-requests

Let me know if you run into any issues with that.

uzaxirr:
Thanks again @bblommers 
I was able to get the xml response from AWS, I ran mitproxy and made a call on describe_security_group_rules on my actual AWS account and copied the XML response as shown below

Screenshot 2023-01-15 at 9 26 37 AM
 
I get the below response when i run the below cmd against my local `moto_server`
`aws --endpoint=http://127.0.0.1:5000 ec2 describe-security-group-rules`

Screenshot 2023-01-15 at 9 28 27 AM


I think the PR is good to be reviewed now, please let me know if any further changes are required or if i've done something wrong.

moto/ec2/models/security_groups.py
```diff
@@ -520,6 +520,30 @@ def describe_security_groups(self, group_ids=None, groupnames=None, filters=None
...
         return matches
 
+    def describe_security_group_rules(self, group_ids=None, groupnames=None, filters=None):
+        matches = itertools.chain(*[x.copy().values() for x in self.groups.values()])
+        if group_ids:
+            matches = [grp for grp in matches if grp.id in group_ids]
+            if len(group_ids) > len(matches):
+                unknown_ids = set(group_ids) - set(matches)
+                raise InvalidSecurityGroupNotFoundError(unknown_ids)
+        if groupnames:
```
bblommers, line 530: We are never passing `groupnames` to this function, so that argument (and this if-block) can be removed

moto/ec2/responses/security_groups.py
```diff
@@ -239,6 +247,39 @@ def update_security_group_rule_descriptions_egress(self):
     </tagSet>
 </CreateSecurityGroupResponse>"""
 
+DESCRIBE_SECURITY_GROUP_RULES_RESPONSE = """
+<DescribeSecurityGroupRulesResponse xmlns="http://ec2.amazonaws.com/doc/2016-11-15/">
+  <requestId>2baaa089-3113-4af5-a1cc-c3af1abe7c6e</requestId>
+  <securityGroupRuleSet>
+    <item>
```
bblommers, line 254: You can add multiple rules to the XML using a for-loop. As an example:
```
{% for rule in rules %}
<item>
  <cidrIpv4>{{ rule.ipv4 }}</cidrIpv4>
  ...
</item>
{% endfor %}
```
See the `CREATE_SECURITY_GROUP_RESPONSE`-template as another example, that has a similar for-loop.

bblommers (changes requested):
Left a few comments @uzaxirr. Keep in mind that the linter is currently failing - see https://github.com/getmoto/moto/actions/runs/3921860236/jobs/6706086309

You can run `make lint` locally to verify whether the linter will pass. 

This page has some background info on how to run the linter only for a specific module, such as EC2, if  you want to speed things up: http://docs.getmoto.org/en/latest/docs/contributing/installation.html#installing-moto-locally

moto/ec2/models/security_groups.py
uzaxirr: Done via https://github.com/getmoto/moto/pull/5841/commits/dd324c3078725f9c1ccf33cf4a4837399aa5b3b5

uzaxirr:
> This looks good to me @uzaxirr - happy to merge this if we can get the test to pass

Yes working on the test cases

tests/test_ec2/test_security_groups.py
```diff
@@ -563,6 +563,48 @@ def test_authorize_all_protocols_with_no_port_specification():
...
+    sgr_id = sgr['SecurityGroupRules'][0]['SecurityGroupRuleId']
+    response = client.describe_security_group_rules(
+        Filters=[
+            {
+                'Name': 'ip-permission-id',
+                'Values': [sgr_id]
+            }
+        ]
+    )
+    ingress_rule = response["SecurityGroupRules"]
```
uzaxirr, line 596: the `describe_security_group_rules` API call returns a list of all the rules that match the filter criteria, not just the one that you are interested in.  So the variable `ingress_rule` contains 4 results. I can't figure out the reason why bcz i'm not creating any additional rules. But it also contains the one in which we are interested so I've iterated through the list of rules and check if the security group rule id matches the id we are expecting. Once we find the expected rule, Then assertions are performed on it.

Let me know if this approach dosen't sounds good to you. And feel free to suggest a better one if you have in mind

uzaxirr:
And how can i assert IP Ranges on the rule that is created if this is required?
I tried `assert rule["IpRanges"][0]["CidrIp"] == "1.2.3.4/32"` but getting a KeyError on `IpRanges`

uzaxirr:
How can i get the `cidrIpv4` for `SecurityRule` object?
I tried `rule.ipv4` and `rule.CidrIp` but in vain

Screenshot 2023-01-17 at 11 19 12 AM

uzaxirr:
I got all the test cases to pass except [this](https://github.com/getmoto/moto/actions/runs/3941498682/jobs/6744345310)

It is passing in my local machine. But not in the CI

Screenshot 2023-01-17 at 11 34 09 PM

uzaxirr:
It passed upon a re-run

bblommers (approved):
Awesome - thank you again for adding this feature @uzaxirr!

PR #5867, opened 2023-01-22
add connection type field to integration model

calvernaz:
Adds the connection type field to the integration model.

https://docs.aws.amazon.com/apigateway/latest/api/API_Integration.html


bblommers (approved):
LGTM - thanks for the PR @calvernaz!

PR #5872, opened 2023-01-23
CognitoIDP: fix REFRESH_TOKEN_AUTH flow in admin_initiate_auth()

opavlyuk:
Fix https://github.com/getmoto/moto/issues/5871


bblommers (approved):
LGTM - thanks for the fix @opavlyuk!

PR #5874, opened 2023-01-24
bugfix: add dollar to the end of mediaconnect, medialive, mediapackage url_paths

nickphbrown:
to prevent them being popped from matchers list

I'm struggling to recreate the issue I'm having. Do we need tests for this? Or are we happy to merge this given that all other url_paths end with a dollar sign and it hasn't broken any tests?


bblommers:
Are you running into any specific issues, @nickphbrown? I'm not against the PR as such, just trying to understand the reasoning behind it.

The URL's should no longer be popped from the matcher-list btw, with the latest version - see https://github.com/getmoto/moto/pull/5837 where this was fixed/removed

nickphbrown:
@bblommers ah wow, that's a wild coincidence that I was looking at the same piece of code at a similar time. 

The issue I have that because of the lack of line end signifier in the regex (`$`) the describe url path matches both the list path and the describe one. This meant that a describe call would delete pop the list path, and the list call wouldn't find the list path and so the call would go out to call real AWS

I think the issue will surface slighlty different now, where the describe call will match the list path first and just make the call to the list endpoint instead?

bblommers:
> I think the issue will surface slighlty different now, where the describe call will match the list path first and just make the call to the list endpoint instead?

Yup, sounds like it. Thanks for the PR to fix this!

bblommers (approved):

Issue #5881, opened 2023-01-29
ThreadPoolExecutor and IAM


mikegrima:
Hello Again!

I found an interesting issue when using a `ThreadPoolExecutor` and I wanted to document it here because it can be useful for others who encounter this.

Moto is not known to be thread safe, however, there are a number of tests that you can run safely with threads. I recently encountered the following when using code like this:

```python
def some_tasking_func(*args) -> Any:
    """do some boto thing here"""
    boto3.client("...")
    ...
    return ...

def some_func(loop: AbstractEventLoop, things_to_do: List) -> None:
    with ThreadPoolExecutor(max_workers=10) as executor:
        tasks = []

        for thing in things_to_do:
            tasks.append(loop.run_in_executor(executor, some_tasking_func, thing))

        try:
            await asyncio.gather(*tasks)
        except Exception as err:
            print("EXCEPTION ENCOUNTERED!!")
            print(err)
            raise     
```

You would get this stacktrace:
```
  File "/.../python3.10/site-packages/moto/iam/models.py", line 73, in get_account_id_from
    for account_id, account in iam_backends.items():
RuntimeError: dictionary changed size during iteration
Future exception was never retrieved
future: <Future finished exception=RuntimeError('dictionary changed size during iteration')>
```

Looking at the `get_account_id_from` function in `moto/iam/models.py`, I see that it's doing some iteration that would cause issues in threads. However, for the test that I am running, I don't really care about the account ID. So, I was able to solve this problem by doing the following in a pytest fixture:

```python
def mock_moto_accounts() -> Generator[None, None, None]:
    with mock.patch("moto.iam.models.get_account_id_from", lambda x: "123456789012"):
        yield

...
def test_some_func(mock_moto_accounts: None) -> None:
    loop = asyncio.new_event_loop()
    things_to_do = [1, 2, 3]
    try:
        loop.run_until_complete(some_func(loop, things_to_do))
    finally:
        loop.close()
```

^^ Using this fixture in my tests solved my issue. I don't think that we need to make any changes to moto, but I would like to have this documented as a workaround for the community just in case they need to do things with Threads and don't actually care about unique account IDs being returned in any of their tests.

I would be happy to add this to the docs, or just even leave this here as a GitHub issue so it can be searched for by someone later. Just let me know!

Thank You

bblommers:
Hey @mikegrima - thanks for raising this!

[...]

PR #5882, opened 2023-01-30
Fixed an issue with the moto IAM backend when used in Thread Pools

mikegrima:
- Fixes issue #5881


bblommers (approved):
Awesome - thanks for the fix @mikegrima!

PR #5887, opened 2023-01-31
EC2: Allow providing NIC template when creating several instances

Kimamisa:
AWS ec2 API allows providing a NIC "template" when creating instances, which does not contain any IP. ec2 assign an IP address automatically for each instance.

With moto, almost everything was made to reproduce the same behaviour, except that the dict containing the params was directly updated with the assigned IP address, and was producing an error the second time when checking if the IP was available, because it was checking it against the previously assigned IP.

The solution is very simple: duplicating the parameter to avoid updating the original entry.


bblommers (approved):
LGTM! Thank you for fixing this @Kimamisa, and welcome to Moto!

PR #5915, opened 2023-02-09
Glue: support nanosecond-precision timestamp partition filtering

robert-schmidtke:
This PR supports nanosecond-resolution timestamp partition filters for Glue. It does so by, instead of using a Python `datetime` to compare timestamps, using POSIX timestamps augmented to nanosecond resolution.

The validity of the timestamp string is still checked, but the POSIX representation is used internally for comparisons.


robert-schmidtke:
Will fix the import lint tomorrow!

bblommers (approved):
LGTM! Thanks for the PR @robert-schmidtke

PR #5918, opened 2023-02-10
Transcribe: add IdentifyMultipleLanguages and return LanguageCodes

benvandenberg:
The [start_transcription_job method](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/transcribe.html#TranscribeService.Client.start_transcription_job) has an IdentifyMultipleLanguages attribute, and returns LanguageCodes if more than one language is found in the recording.


bblommers (approved):
LGTM! Thank you for adding this feature @benvandenberg

Issue #5914, opened 2023-02-09
create_function() return wrong "LastModified" date format


shanishiri:
create_function method is creating instance of "LambdaFunction" class, which initialize last_modified with different datetime string format then the [official boto3 docs define  ](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/lambda.html#Lambda.Client.create_function)

moto:
self.last_modified = datetime.datetime.utcnow().strftime("%Y-%m-%d %H:%M:%S")

where the datetime format should be "%Y-%m-%dT%H:%M:%S.%f+0000"

```
@mock_lambda
def test_create_function():
  func = client.create_function(
          FunctionName="1",
          Runtime="nodejs12.x",
          Role=role_arn,
          Code={"ZipFile": b""},
          Handler="original_handler",
          Environment={"Variables": {"aa": "bb"}},
          Layers=[layer]
      )
```

for func['LastModified'] expected value is: `"2023-02-09T11:02:38.281+0000"`

actual value: `"2023-02-09 11:02:38"`

using moto==4.1.2

bblommers:
Thanks for raising this @shanishiri - marking it as a bug!

Do you want to raise a PR for this yourself?

PR #5921, opened 2023-02-12
fixed create_function() return wrong LastModified date format #5914

samirhseo:


bblommers (changes requested):
Hi @samirhseo, thanks for the PR! 
We have a utility-method that does this conversion: `moto/core/utils.py::iso_8601_datetime_with_nanoseconds()`, so we should be using that instead. 


Further down we update this attribute again, in the `set_version`-method, so ideally we refactor that as well to be formatted using the same utility-method.

samirhseo:
> Hi @samirhseo, thanks for the PR! We have a utility-method that does this conversion: `moto/core/utils.py::iso_8601_datetime_with_nanoseconds()`, so we should be using that instead.
> 
> Further down we update this attribute again, in the `set_version`-method, so ideally we refactor that as well to be formatted using the same utility-method.

thanks @bblommers I didn't spot that utility. I've ammended now. Is that what you were looking for?

bblommers (approved):
Beautiful. Thank you for fixing this @samirhseo, and welcome to Moto!

PR #5928, opened 2023-02-14
Add /moto-api/set-athena-result API

ikkala:
This pull request adds support for setting Athena query result via HTTP POST to moto server. 

This change might help people in needs like [in comment](https://github.com/getmoto/moto/issues/1524#issuecomment-812649597).

### DIFF
--- a/docs/docs/services/athena.rst
+++ b/docs/docs/services/athena.rst
@@ -87,7 +87,7 @@ athena
             # Make query execution ID generation deterministic
             # by calling first /moto-api/seed:
             requests.get(
-                "http://localhost:5000/moto-api/seed?a=42",
+                "http://motoapi.amazonaws.com:5000/moto-api/seed?a=42",
             )
 
             exex_id = "bdd640fb-0667-4ad1-9c80-317fa3b1799d"
@@ -114,7 +114,7 @@ athena
                 ],
             }
             resp = requests.post(
-                "http://localhost:5000/moto-api/set-athena-result",
+                "http://motoapi.amazonaws.com:5000/moto-api/static/athena/region/query_executions",
                 json=athena_result,
             )
             resp.status_code.should.equal(201)
--- a/docs/docs/services/athena.rst
+++ b/docs/docs/services/athena.rst
@@ -51,69 +51,50 @@ athena
 - [X] get_query_execution
 - [X] get_query_results
   
-        Queries are not executed, so this call will always return 0 rows by default.
+        Queries are not executed by Moto, so this call will always return 0 rows by default.
 
-        When using decorators, you can use the internal API to manually set results:
+        You can use a dedicated API to configure this. Moto has a queue that can be filled with the expected results.
 
-        .. sourcecode:: python
+        A request to `get_query_results` will take the first result from that queue, and assign it to the provided QueryExecutionId. Subsequent requests using the same QueryExecutionId will return the same result. Other requests using a different QueryExecutionId will take the next result from the queue, or return an empty result if the queue is empty.
 
-            from moto.athena.models import athena_backends, QueryResults
-            from moto.core import DEFAULT_ACCOUNT_ID
-
-            backend = athena_backends[DEFAULT_ACCOUNT_ID]["us-east-1"]
-            rows = [{'Data': [{'VarCharValue': '..'}]}]
-            column_info = [{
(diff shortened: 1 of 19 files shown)


bblommers:
Hi @ikkala, thanks for getting the ball rolling on this! This is definitely a useful feature, and something that should be part of Moto. 

Some initial thoughts:
 - This problem is not unique to Athena. There are other services that can benefit from the same approach, so the API should be extendible. Maybe we can have an endpoint like `moto-api/static/athena/region/query_executions`. It's a bit of a mouthful, but it will make it easier to add other services/features down the line
 - Having to rely on seeding Moto is a bit clunky and error prone. So I'd prefer to have a solution without, where the user doesn't have to worry about this part.
   Alternative solution: Let the user provide a list of possible results. When the `get_query_execution` is invoked, we can then pull the result from this list. In pseudo-code something like:
```
if exec_id not in results:
    if user_provided_results:
        results[exec_id] = user_provided_results.pop(0)
    else:
        raise UnknownQuery
```
 - Just a documentation/test note: it is possible to use this solution when using the decorators. The URL should just be updated to `motoapi.amazonaws.com` instead of localhost (for decorators)

Like I said, this feature is a really great addition to Moto, but because it is quite fundamental and has the potential for expansion, I really want to make sure this is as user-friendly and intuitive as possible.

Let me know what you think. Would the suggested solution (providing a list) work for your use-case?

ikkala:
@bblommers Do you happen to know how to make the get_query_execution request work in server mode in the end of the test case of this pull request -- so that we can verify that the new  /moto-api/set-athena-result really did set the result content?

I tried that originally and it failed big time, so [I removed the final step](https://github.com/getmoto/moto/pull/5928/commits/0c6771a1a432763d067badbb5e7b16d9bd95dd4c) from the PR eventually yesterday.

bblommers:
> @bblommers Do you happen to know how to make the get_query_execution request work in server mode in the end of the test case of this pull request -- so that we can verify that the new /moto-api/set-athena-result really did set the result content?
> 

It looks like it should work - I can't see anything obvious wrong with it. I'd have to do some debugging to figure out why it fails.

ikkala:
> This problem is not unique to Athena. There are other services that can benefit from the same approach, so the API should be extendible. Maybe we can have an endpoint like moto-api/static/athena/region/query_executions. It's a bit of a mouthful, but it will make it easier to add other services/features down the line

Indeed, there's good to have room for growing definitely! 

> Having to rely on seeding Moto is a bit clunky and error prone. So I'd prefer to have a solution without, where the user doesn't have to worry about this part.

I agree on this also. The reason I took this approach was that the already-existing mock for Athena that can be run without test server seems to have it same way: You kind of have to guess what will be the next Query Execution ID if your application logic that is under testing doesn't reveal the ID naturally. Should this be enhanced also there, in the non-test-server side..?  Hmm.. maybe I have just missed something.

> Alternative solution: Let the user provide a list of possible results. When the get_query_execution is invoked, we can then pull the result from this list.

This sounds actually good idea. As wondered above: Should this be changed also in the non-test-server side?

> Would the suggested solution (providing a list) work for your use-case?

In our case for example, there's REST service that encapsulates the whole "trigger Athena query, get execution ID from that, wait for query to be ready, get the result" chain and therefore it doesn't reveal the ID.

So yes, in our case the alternative solution would work indeed.

bblommers:
The suggested approach should work for both decorators and ServerMode, @ikkala. It's just that configuring it involves an HTTP-request to a different endpoint. 

BTW, I've updated the suggested endpoint in my earlier post to also include the `account`, as people may want to configure this for different/multiple accounts.

ikkala:
> BTW, I've updated the suggested endpoint in my earlier post to also include the account, as people may want to configure this for different/multiple accounts.

I was about to tinker these changes talked above but realized that the scope I originally had expanded -- you know, originally I had just idea to provide the same functionality to server  mode than there's already for decorator.
And, because I have very limited understand of the moto, the adding of account and region to API path is above it...
Can you @bblommers make them dynamic the way you mentioned?

bblommers:
> Can you @bblommers make them dynamic the way you mentioned?
>

No probs, I'll have a look later. Looks good to me so far!

bblommers:
I created a PR for you to review here @ikkala: https://github.com/ikkala/moto/pull/2

PR #5932, opened 2023-02-15
EC2: DestinationCidrBlock filter for DescribeRouteTables

viren-nadkarni:


tests/test_ec2/test_route_tables.py
```diff
@@ -147,6 +147,22 @@ def test_route_tables_filters_standard():
         [route["GatewayId"] == igw.id for table in resp for route in table["Routes"]]
     )
 
+    # Filter by route destination CIDR block
+    resp = client.describe_route_tables(
+        Filters=[
+            {"Name": "route.destination-cidr-block", "Values": ["10.0.0.4/24"]},
+        ]
+    )["RouteTables"]
+    resp.should.have.length_of(1)
```
bblommers, line 156: The EC2 tests run in parallel, so there can be multiple route tables in the system that match this query.

tests/test_ec2/test_route_tables.py
```diff
@@ -147,6 +147,22 @@ def test_route_tables_filters_standard():
...
+    resp[0]["RouteTableId"].should.equal(route_table2.id)
+    assert any(
```
bblommers, line 158:
```suggestion
    all_routes = itertools.chain(..)
    assert len(all_routes) >= 1
    assert all([route["DestinationCidrBlock"] == "10.0.0.4/24" for route in all_routes])
```
Just a suggestion. I don't recall out of the top of my head how to flatten the response using itertools, but this would make the test agnostic against other resources. We only need to know that we have at least one route, and that all returned routes match our specified CidrBlock

bblommers (changes requested):

tests/test_ec2/test_route_tables.py
viren-nadkarni: > all returned routes match our specified CidrBlock

I don't think that's the behaviour we want. As per [AWS](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_DescribeRouteTables.html):

> `route.destination-cidr-block` - The IPv4 CIDR range specified in a route in the table.

i.e. not all routes will have the specified CidrBlock

tests/test_ec2/test_route_tables.py
viren-nadkarni: Fixed with c9be51fcf273507baecb7a7d5e5fd244d73452fe

bblommers (approved):
LGTM! Thanks for the fix @viren-nadkarni

PR #5935, opened 2023-02-16
implement create_group

dana-orca:


moto/__init__.py
```diff
@@ -167,6 +167,8 @@ def f(*args, **kwargs):
 mock_xray_client = lazy_load(".xray", "mock_xray_client")
 mock_wafv2 = lazy_load(".wafv2", "mock_wafv2")
 mock_textract = lazy_load(".textract", "mock_textract")
+mock_identitystore = lazy_load(".identitystore", "mock_identitystore", boto3_name="identitystore")
```
bblommers, line 170: The boto3-name will be automatically deduced in most cases, so we don't have to explicitly define it
```suggestion
mock_identitystore = lazy_load(".identitystore", "mock_identitystore")
```
We also only have to define it once :slightly_smiling_face:

moto/identitystore/models.py
```diff
@@ -0,0 +1,28 @@
+"""IdentityStoreBackend class with methods for supported APIs."""
+import uuid
```
bblommers, line 2: Please use `moto.moto_api._internal.mock_random` instead. That will guarantee that UUID's are always deterministic when Moto is seeded.

bblommers (changes requested):
Thanks for the PR @dana-orca - please see my comments.

Looks like the linter is failing as well: https://github.com/getmoto/moto/actions/runs/4193736984/jobs/7276278721

moto/identitystore/models.py
dana-orca: Gr8

Changed accordingly

dana-orca:
> 

I ran `make format` this time.
hopefully that would do the trick 😄

dana-orca:
@ananhagyahya

dana-orca:
@bblommers server tests failed because I didnt update backeds index 🤦‍♀️

bblommers:
> @bblommers server tests failed because I didnt update backeds index woman_facepalming

Yeah, that's my bad - I should've mentioned that earlier. Thank you for this PR @dana-orca!

PR #5951, opened 2023-02-20
Enhancement - Add policy to update rest api for apigateway

JCharlot:
- Add the capacity to update ("replace" operation) the **policy** attribute in **update_rest_api** endpoint.
- Enhancement and updating the tests.

### DIFF
--- a/moto/apigateway/models.py
+++ b/moto/apigateway/models.py
@@ -1015,10 +1015,7 @@ class RestAPI(CloudFormationModel):
         self.binaryMediaTypes = kwargs.get(RestAPI.PROP_BINARY_MEDIA_TYPES) or []
         self.create_date = int(time.time())
         self.api_key_source = kwargs.get("api_key_source") or "HEADER"
-        self.policy = (
-            kwargs.get(RestAPI.PROP_POLICY)
-            or '{"Version": "2012-10-17", "Statement": []}'
-        )
+        self.policy = kwargs.get(RestAPI.PROP_POLICY) or None
         self.endpoint_configuration = kwargs.get("endpoint_configuration") or {
             "types": ["EDGE"]
         }
--- a/moto/apigateway/models.py
+++ b/moto/apigateway/models.py
@@ -1015,10 +1015,7 @@ class RestAPI(CloudFormationModel):
         self.binaryMediaTypes = kwargs.get(RestAPI.PROP_BINARY_MEDIA_TYPES) or []
         self.create_date = int(time.time())
         self.api_key_source = kwargs.get("api_key_source") or "HEADER"
-        self.policy = (
-            kwargs.get(RestAPI.PROP_POLICY)
-            or '{"Version": "2012-10-17", "Statement": []}'
-        )
+        self.policy = kwargs.get(RestAPI.PROP_POLICY) or None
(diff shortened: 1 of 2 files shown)


bblommers (changes requested):
Looks like Terraform doesn't like the new default value: https://github.com/getmoto/moto/actions/runs/4219125564/jobs/7341346193

Let me know if you want me to take a look at this.

I'm OK with this change, as long as we can get the tests to pass!

JCharlot:
> Looks like Terraform doesn't like the new default value: https://github.com/getmoto/moto/actions/runs/4219125564/jobs/7341346193
> 
> Let me know if you want me to take a look at this.
> 
> I'm OK with this change, as long as we can get the tests to pass!

Yep, maybe a leading space or something in the format of the JSON default value is getting TF unhappy.
If you can take a look at it, I would be grateful, thanks :)

bblommers (commented):
@JCharlot I can't actually reproduce this behaviour. If I provide only the name for a new REST api, AWS does not return a policy at all:
```
$ aws apigateway create-rest-api --name poltest --no-verify-ssl
{
    "id": "gjmlzg6hr2",
    "name": "poltest",
    "createdDate": "2023-02-21T18:47:01-01:00",
    "apiKeySource": "HEADER",
    "endpointConfiguration": {
        "types": [
            "EDGE"
        ]
    },
    "disableExecuteApiEndpoint": false
}


$ aws apigateway get-rest-api --rest-api-id gjmlzg6hr2 --no-verify-ssl

{
    "id": "gjmlzg6hr2",
    "name": "poltest",
    "createdDate": "2023-02-21T18:47:01-01:00",
    "apiKeySource": "HEADER",
    "endpointConfiguration": {
        "types": [
            "EDGE"
        ]
    },
    "tags": {},
    "disableExecuteApiEndpoint": false
}
```
How are you calling AWS, that you are seeing this behaviour? Are you using a specific set of parameters?

JCharlot:
Yep if you don't set a policy when creating the rest-api, it doesn't return a policy at all. But if you update the policy, it will be there.

Check out this example (btw that's my point, I'm not able to update a policy due to the missing if condition in moto, even if not set when creating the rest-api) :

```
@mock_apigateway
def test_update_rest_api_policy():
    client = boto3.client("apigateway", region_name="us-west-2")
    response = client.create_rest_api(name="my_api", description="this is my api")
    api_id = response["id"]
    patchOperations = [
        {"op": "replace", "path": "/name", "value": "new-name"},
        {"op": "replace", "path": "/description", "value": "new-description"},
        {"op": "replace", "path": "/policy", "value": json.dumps('{"Version": "2012-10-17", "Statement": []}')},
    ]

    response = client.update_rest_api(restApiId=api_id, patchOperations=patchOperations)

    response.should.equal(
        {
            "id": api_id,
            "name": "new-name",
            "description": "new-description",
            "policy": '"{\\"Version\\": \\"2012-10-17\\", \\"Statement\\": []}"',
            ...
        }
    )
```

Be careful about the policy which needs to be a stringified JSON document.

I'll fix the tests with no default value, should be fine then.

bblommers (approved):
Awesome - thanks @JCharlot!

PR #5953, opened 2023-02-20
Fix "local variable 'containers' referenced before assignment".

oxidase:
If "import docker" fails https://github.com/getmoto/moto/blob/aa332c4d/moto/batch/models.py#L610 `Job.run`  fails with an obscure error `UnboundLocalError` error as:

```
.../_pytest/threadexception.py:73: PytestUnhandledThreadExceptionWarning: Exception in thread MOTO-BATCH-e3d4d43a-b68e-4f64-b2e8-7c396c0f3ad2

    File ".../lib/python3.9/threading.py", line 973, in _bootstrap_inner
      self.run()
    File ".../moto/batch/models.py", line 854, in run
      for container in containers:
  UnboundLocalError: local variable 'containers' referenced before assignment

    warnings.warn(pytest.PytestUnhandledThreadExceptionWarning(msg))
```

The PR changes the scope of 'containers' variable to prevent the issue.

### DIFF
--- a/moto/batch/models.py
+++ b/moto/batch/models.py
@@ -606,7 +606,12 @@ class Job(threading.Thread, BaseModel, DockerModel, ManagedState):
         """
         try:
             import docker
+        except ImportError as err:
+            logger.error(f"Failed to run AWS Batch container {self.name}. Error {err}")
+            self._mark_stopped(success=False)
+            return
 
+        try:
             containers: List[docker.models.containers.Container] = []
 
             self.advance()


bblommers (changes requested):
It would be nice to keep the MyPy notation - which only works if Docker has been properly imported before.

The error occurs because we try to execute the `finally`-section - but that is pointless if Docker hasn't been imported.

So maybe we can just try-catch the docker import separately?

```
try:
    import docker
catch ImportError:
    logger.error(f"Failed to run AWS Batch container {self.name}. Error failed to import docker-module")
    self._mark_stopped(success=False)
    return
try:
    containers: List[docker.models.containers.Container] = []
    # existing logic...
```
The error message is adapted from the existing, generic, outer catch - I think this is explicit/clear enough? Improvements are welcome of-course.

oxidase:
@bblommers thanks for the suggestion! I have updated the commit and the error message will be now
```
ERROR:moto.batch.models:Failed to run AWS Batch container MOTO-BATCH-bb3b9367-5ead-4c03-9282-212799fc2217. Error No module named 'docker'
```

bblommers (approved):
LGTM! Thanks for the PR @oxidase

PR #5973, opened 2023-02-24
fix(acm): Include more info to the list_certificates

jfagoagas:
### Context

The purpose of this PR is to approach the behaviour of the ACM `list_certificates` API call including new fields. I don't know if this is a new feature of the ACM API but it would be helpful. 

Also, there are more fields present in the `list_certificates` call that are not covered by moto.

Rationale: https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/acm.html#ACM.Client.list_certificates

### Changes

Include the following fields in the `list_certificates` call:
- Status
- Type
- NotBefore
- NotAfter
- CreatedAt
- KeyAlgorithm
- RenewalEligibility
- ImportedAt
- IssuedAt

I don't know if in this case this is the best approach. I didn't make test for that reason, if so I'll do it.

### DIFF
--- a/moto/acm/responses.py
+++ b/moto/acm/responses.py
@@ -3,7 +3,7 @@ import base64
 
 from moto.core.responses import BaseResponse
 from typing import Dict, List, Tuple, Union
-from .models import acm_backends, AWSCertificateManagerBackend, datetime_to_epoch
+from .models import acm_backends, AWSCertificateManagerBackend
 from .exceptions import AWSValidationException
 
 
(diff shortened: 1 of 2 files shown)


bblommers (changes requested):
This looks right to me @jfagoagas - if you can add tests for some of these fields to ensure they are configured correctly that would indeed be welcome.

jfagoagas:
> This looks right to me @jfagoagas - if you can add tests for some of these fields to ensure they are configured correctly that would indeed be welcome.

I've included some changes in the `test_list_certificates` function but I'm having some issues freezing the time because I'm getting 1 hour difference between the test case and the `list_certificates` return value even when freezing all the time in UTC with `@freeze_time(datetime.utcnow())`.

bblommers (approved):
Thanks @jfagoagas!

PR #5974, opened 2023-02-24
Introduce glue.get_jobs call

danielward:
Implements `GetJobs` to the Moto Glue Client in accordance with the [get_jobs()](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/glue.html#Glue.Client.get_jobs) boto specification.

In a nutshell, `get_jobs()` isn't far dissimilar to `list_jobs()`, however it does not accept **any** filter parameters (only pagination parameters). In addition, whilst `list_jobs()` returns a list of the jobs' `Name` attributes, `get_jobs()` returns the Job object similarly (or identically) to the data returned from `create_job()`.

Testing has been introduced, drawing from @dana-orca's contribution for `list_jobs()` to maintain consistency.

As a final point, the `scaffold.py` script updated the `mock_glue` `lazy_load` call with the `boto3_name` parameter locally, so I've updated the original call with this change also.


danielward:
The coverage report has nicely highlighted that there's some redundant code I need to look into. Will take a look now.

danielward:
Resolved the issue and hoping to see an improvement in the coverage as a result also.

Also, on a slightly related topic, `list_jobs` supports pagination because the API suggests that's possible, however there actually isn't a paginator and I've raised a botocore issue about it here: https://github.com/boto/botocore/issues/2876

bblommers (approved):
Awesome, thanks for adding this feature @danielward!

PR #5977, opened 2023-02-24
fix: add cognito:groups to idToken

adriantomas:
This pull request solves https://github.com/getmoto/moto/issues/5976

I have simply taken the existing code in the `accessToken` and adapted it here.


bblommers (approved):
LGTM! Thank you for fixing this @adriantomas

PR #6002, opened 2023-02-28
Admin: Add Java tests

### LINKED ISSUE #5993: [DynamoDb] All Errors are deserialized as DynamoDbException in JAVA SDK
Using MotoServer to test Java based services. This Service creates the DynamoDb table if it doesn't exist, ie, on ResourceNotFoundException, table is created.
When using the MotoServer, We get response

```json
{"__type": "com.amazonaws.dynamodb.v20111205#ResourceNotFoundException", "message": "Requested resource not found"}
```
When this is deserialized in java sdk, as there is no error-code for  `com.amazonaws.dynamodb.v20111205#ResourceNotFoundException` hence, All the error is wrapped, and returned as `DynamoDbException`

With other services, similar issue is not occurring. The issue exists for version>=3.1.5, and seems to be introduced by this https://github.com/getmoto/moto/pull/5012
Tried using <=3.1.4 to overcome this problem, but ended up encountering another bug related to Glue, which is fixed in the recent version. So, unable to use the older versions as a workaround

Could you please share reason behind the error message? would it be possible to have error message compatible with Java SDKs too?

bblommers:
Some very simple tests using the Java SDK to verify Moto behaves correctly.

Initial tests include some basic S3 and DynamoDB operations.

Why this is necessary:
As a user, when I create a table that already exists, DynamoDB throws a ResourceInUseException.
The Java SDK expects a specific header: `"Content-Type": "application/x-amz-json-1.0"`. Without this header, the Java-call throws a generic DynamoDbException, instead of the dedicated exception that the user expects.

This PR also adds the required header to DynamoDB to ensure the Java-test succeeds.

This should fix #5993

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -70,6 +70,35 @@ jobs:
       run:
         make lint
 
+  javatest:
+    name: Test behaviour for Java SDK
+    runs-on: ubuntu-latest
+    needs: lint
+
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v4
+        with:
+          python-version: "3.8"
+      - name: Start MotoServer
+        run: |
+          pip install build
+          python -m build
+          docker run --rm -t --name motoserver -e TEST_SERVER_MODE=true -e AWS_SECRET_ACCESS_KEY=server_secret -e AWS_ACCESS_KEY_ID=server_key -v `pwd`:/moto -p 5000:5000 -v /var/run/docker.sock:/var/run/docker.sock python:3.7-buster /moto/scripts/ci_moto_server.sh &
+          python scripts/ci_wait_for_server.py
+      - uses: actions/setup-java@v3
+        with:
+          distribution: 'temurin'
+          java-version: '17'
+          cache: 'maven'
+      - name: Build with Maven
+        run: |
+          mkdir ~/.aws && touch ~/.aws/credentials && echo -e "[default]\naws_access_key_id = test\naws_secret_access_key = test" > ~/.aws/credentials
+          cd other_langs/tests_java && mvn test
+
   test:
     name: Unit test
     runs-on: ubuntu-latest
@@ -182,7 +211,7 @@ jobs:
   deploy:
     name: Deploy
     runs-on: ubuntu-latest
-    needs: [test, testserver ]
+    needs: [javatest, test, testserver ]
     if: ${{ github.event_name == 'push' && github.ref == 'refs/heads/master' && github.repository == 'getmoto/moto' }}
     strategy:
       matrix:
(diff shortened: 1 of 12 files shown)


bpandola:
@bblommers I think this is a super cool idea to have some tests using the AWS SDKs for other languages.  I fixed something long ago that was failing with the Go SDKs, and it would've been nice to actually have had that test captured as well.  Maybe it makes sense to put these into a separate GitHub workflow, though, instead of adding to the already-very-busy TestNDeploy workflow?

Also, looks like your Java tests are failing because the region isn't set.  But once you address that, I suspect that the tests will still fail for the DynamoDB exceptions.  I did have a quick look earlier and I *think* I know what moto is doing incorrectly here.  But anyway, I'll wait till you get these tests sorted out before getting into this further (because maybe you already know what's wrong).

bblommers:
Hey @bpandola, I agree about the busyness of TestNDeploy. It would be nice to keep using these tests as a gatekeeper for the Deploy (/release) step though - I'll see if it's possible to do that with multiple workflows.

The tests did pass for me locally, so I hope that it's just the region-issue that keeps it from passing. What did you spot, that you think will still cause it to fail?
Note that I only fixed a single exception, so other scenarios/exceptions are still likely to fail with the SDK - but I'd like to at least get this over the line, before extending/improving the coverage.

TLDR: Will look into it! Appreciate the feedback. :+1:

bpandola:
@bblommers Some (hopefully) helpful input:

Well... your tests passed despite my thinking they wouldn't, but not for the reason you think!  If you modify the `get_headers` method you added to the `ResourceInUseException` to just return an empty array, the tests will still pass.  But even forgetting that for a minute, you don't really want to add a `get_headers` method to every DynamoDB exception, do you?  

The real culprit is here:

https://github.com/getmoto/moto/blob/c5a91e6cc6f6c9c0ab207f2e9f36fcec85febfaa/moto/core/exceptions.py#L74

The Amazon specs say that error information should be passed in a body field named `__type` (which moto does) or a header name `X-Amzn-Errortype` (which moto also does).  It doesn't appear to be an issue that moto uses both, but DynamoDB is a legacy protocol that returns long-form error codes, e.g. `com.amazonaws.dynamodb.v20120810#ResourceInUseException`.  These are fine if they're in the `__type` field, but the `X-Amzn-Errortype` value should be the "relative" error type, disambiguated in the following way:

> If a # character is present, then take only the contents after the first # character in the value.

Adding a bit of code to `core/exceptions.py` to only serialize the relative error code in the header should fix this issue for *all* DynamoDB exceptions (and shouldn't break anything).

Hope that make sense (and that I'm actually correct about this 😄).

bblommers:
That does make sense @bpandola - that's a good find, and thanks for the explanation. Returning the 'relative' error-type is a much cleaner solution, and does solve the issue.

Just a nitpick: the spec does say that `All client-side implementations SHOULD support sanitizing the value [..]` - in my eyes, that means the SDK should resolve the relative error-type for  us. Ah, well..

PR #6004, opened 2023-03-01
Admin: Introduce reusable CI workflows

bblommers:
Extract CI steps into separate files to make the overall configuration file more readable.

Documentation: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onworkflow_call

Other possible improvements:
 - Reuse all the steps, so that the main workflow file only points to sub-workflows, without containing any logic
 - Look into reusing individual steps (such as creation of the MotoServer Docker image, which is reproduced everywhere)

Based on a suggestion made in #6002.

This PR also fixes some wrong cache-keys, which resulted in the cache being filled with duplicates.

FYI @bpandola - I'm not sure yet whether this is the best approach, but it does help readability. Suggestions are always welcome!

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -25,7 +25,7 @@ jobs:
         uses: actions/cache@v3
         with:
           path: ${{ steps.pip-cache-dir.outputs.dir }}
-          key: pip-${{ matrix.python-version }}-${{ hashFiles('**/setup.cfg') }}-4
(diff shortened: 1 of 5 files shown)


bpandola:
I was just thinking it might make sense to have the "other language" tests in their own workflow, e.g. `test_outdated_versions.yml`; but what you've done here also seems useful!

bblommers:
Yeah, I'll merge the Java tests first, and then I'll also extract those to a separate file (but still call it within the 'main' workflow to act as a gatekeeper).

PR #6023, opened 2023-03-07
Support CloudFormation and ECS ResourceTypeFilters in ResourceGroupsTaggingAPI get_resources()

stephaniemanning:
Add support for cloudformation:stack, ecs:service, ecs:cluster, and ecs:task ResourceTypeFilters for ResourceGroupsTaggingAPI get_resources.

### DIFF
--- a/moto/resourcegroupstaggingapi/models.py
+++ b/moto/resourcegroupstaggingapi/models.py
@@ -186,8 +186,9 @@ class ResourceGroupsTaggingAPIBackend(BaseBackend):
                     if not tag_filter(tags):
                         continue
                     yield {"ResourceARN": f"{stack.stack_id}", "Tags": tags}
+
             except Exception:
-                yield {}
+                pass
 
         # ECS
         if not resource_type_filters or "ecs:service" in resource_type_filters:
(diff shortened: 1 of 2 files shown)


moto/resourcegroupstaggingapi/models.py
```diff
@@ -106,6 +108,20 @@ def lambda_backend(self):
         """
         return lambda_backends[self.account_id][self.region_name]
 
+    @property
+    def cloudformation_backend(self):
+        """
+        :rtype: moto.cloudformation.models.CloudFormationBackend
+        """
+        return cloudformation_backends[self.account_id][self.region_name]
```
bblommers, line 116: The CF-module imports all other modules, which means that the user needs to have all dependencies installed - which may not be the case

Can we try-catch the entire thing, just for CF? So that if anything fails with importing, we can still continue to collect tags for other services.
```
def get_cf_stack_tags():
    try:
        from moto.cloudformation import cloudformation_backends

        backend = cloudformation_backends[self.account_id][self.region_name]

        return all_tags
    except:
        return []
```

bblommers (changes requested):

moto/resourcegroupstaggingapi/models.py
stephaniemanning: updated, let me know if that's what you had in mind

bblommers (approved):
LGTM! Thanks for adding this feature @stephaniemanning

PR #6024, opened 2023-03-07
Techdebt: allow test & lint to run for specific service

hans-d:
make test SERVICE_NAME=s3


Makefile
```diff
@@ -20,34 +25,35 @@ init:
...
 	$(eval black_version := $(shell grep "^black==" requirements-dev.txt | sed "s/black==//"))
 	@echo "(Make sure you have black-$(black_version) installed, as other versions will produce different results)"
-	black --check moto/ tests/
+	black --check moto/${SERVICE_NAME} tests/${TEST_SERVICE}
 	@echo "Running pylint..."
-	pylint -j 0 moto tests
+	pylint -j 0 moto/${SERVICE_NAME} tests/${TEST_SERVICE}
 	@echo "Running MyPy..."
-	mypy --install-types --non-interactive
+	mypy --install-types --non-interactive moto/${SERVICE_NAME}
```
bblommers, line 36: Not all services support typing at the moment - the ones that do are listed in `setup.cfg`. This is a very quick check, it only takes a few seconds, so I would suggest to just leave this as is.

hans-d: changed it a bit, but we can go this route as well.

setup.cfg
```diff
@@ -211,7 +211,130 @@ universal=1
 
 [tool:pytest]
 markers =
+    # special cases
     network: marks tests which require network connection
+    parallel: can be run in parallel
+    parallel_server_only: can be run in parallel (in server mode only)
+    skip_server: should not be run in server mode
+    # services
+    acm: acm service
```
bblommers, line 220: Can we get this feature without listing all services? This would be a pain to maintain, considering the list changes quite often

setup.cfg
```diff
@@ -229,7 +352,14 @@ disable = W,C,R,E
 enable = anomalous-backslash-in-string, arguments-renamed, dangerous-default-value, deprecated-module, function-redefined, import-self, redefined-builtin, redefined-outer-name, reimported, pointless-statement, super-with-arguments, unused-argument, unused-import, unused-variable, useless-else-on-loop, wildcard-import
 
 [mypy]
-files= moto/a*,moto/b*,moto/c*,moto/d*,moto/e*,moto/f*,moto/moto_api,moto/neptune
+files= moto/a*,moto/b*,moto/c*,moto/d*,moto/e*,moto/moto_api,moto/neptune
```
bblommers, line 355: This changes often as well, we're continuously adding typing support for new services, so I'd rather to just have a simple list. Having to understand and get the regex right isn't worth the effort

bblommers (changes requested):

setup.cfg
hans-d: good point, will revert this bit

bpandola:
While I appreciate the spirit of this PR, I think it's too big of a change and too much of a future maintenance hassle for the benefits.  Locally testing just the service you've worked on is already quite simple, e.g. `pytest -sv ./tests/test_redshift`.  (A simply note in the development docs could make this more clear.) 

`make lint` takes a lot longer than it used to, but it's still not that onerous, considering you really only need to run it once at the end of your local development.

Happy to consider better arguments for the need here, but I'm not currently in support of this.

hans-d:
The tagging of resources comes into play, eg when not just wanting to run all cloudformation tests in the tests_cloudformation folder, but also related test in other folders (or for sns, lambda, ....)

A similar list for the services already needs to be maintained for the dependencies. We can piggyback on that re checking for completeness  / use the an additional check that also validates that this list dependency list is complete (not re the dependencies added, but at least having an entry present).

bblommers:
@bpandola The first commit was simpler (https://github.com/getmoto/moto/pull/6024/commits/02b516c74b0da8f59e75fd9321a5e6ec87f03890). A change like that would be a nice QOL-improvement IMO, to run lint/test on a specific folder.

Agree that having another explicit list of services is too much of a hassle.

hans-d:
shall I revert to the first one?

bblommers:
Yes, let's, @hans-d, I think that's better.

PR #6026, opened 2023-03-07
Techdebt: skip tests when docker is not running

hans-d:
Taking a quick approach by assuming that all tests that mention lambda (function name, file name, or folder name) need a docker running, and thus skip them.

The `EXTRA_MARKERS_OTHER` naming is based on an upcoming change where more markers are used.


bblommers (changes requested):
I'm not opposed to adding an optional flag, but these tests should be executed by default IMO.

They should be marked manually though. There are quite a few Lambda tests that do not require Docker (because they only test CRUD-operations), and other tests (for Batch, SQS, maybe others) that do require Docker.

hans-d:
We can mark the individual tests with a requires_docker decorator (takes a bit more time to do that).

For the current situation, all tests will run if a running docker is detected. So there is no need to specify a flag. This only prevents failing tests (as in my case, I don't have docker running but podman)

bblommers:
`all tests will run if a running docker is detected.`
All tests should always run. Unless the user knows what they're doing, and specifically indicate that they don't want Docker-tests to run, everything should always run.

I don't want a scenario where we introduce a bug, but don't know about it because Docker happens to not be running.

hans-d:
Diving further into the tagged functions, to see how far we can run them before providing a setting backed bail out.

With the current setup:
- functions will be fully run
- in case of providing the given setting they will xfail instead of fail

hans-d:
@bblommers do you want these docker related test to run, or is fast fail by default also an option? (only for those tests that do need docker)
When docker is not running, these tests will fail anyway as the output is not wat is expected. As a side effect, these tests will take longer in most cases. 

Also for the fast fail scenario, some "skip these tests" via an option would be good (eg when working with pytest -x)

bblommers:
Skipping them makes more sense then trying, but ignoring the result, I'd say. When setting `TESTS_SKIP_DOCKER_REQUIRED`, you explicitly say that you do not care about these tests, so why not save some time.

Note that I don't have a lot of knowledge of the inner workings of pytest, so I don't know how feasible that is, to skip tests based on an environment variable? I do like the current setup: one env variable defined, and a simple marker above individual tests. Nothing too invasive, and easy to read.

tests/markers.py
```diff
@@ -0,0 +1,19 @@
+import docker
+import pytest
+from docker.errors import DockerException
+
+from moto import settings
+
+
+def _docker_is_available() -> bool:
+    try:
+        docker.from_env()
```
bblommers, line 10: Just a nitpick - I think we should import docker inside the try, and catch all exceptions. 
You may not even have the `docker`-module installed, if you're not going to run the tests anyway.

hans-d: Moved these more explicit bits to a dedicated test, so users are informed via a failing test that other tests may fail as well (and those tests are suppressed along with the switch).

hans-d:
Skipping based on the env variable is no problem.

moto/core/exceptions.py
```diff
@@ -178,3 +181,40 @@ class InvalidToken(AWSError):
 
     def __init__(self, message: str = "Invalid token"):
         super().__init__(f"Invalid Token: {message}", "InvalidToken")
+
+
+class MotoDockerException(RuntimeError):
```
bblommers, line 186: Having a custom exception is not a good solution IMO.
 - It couples the logic very closely to the tests
 - It doesn't work when testing ServerMode, where we only communicate with Moto over HTTP requests.

hans-d: "It couples the logic very closely to the tests" That's usually the case with unit tests, they are very clearbox oriented. 

As you may have noticed, these are not raised by default. This is more like a helper exception, so that the related tests will hard fail when something breaks that is not docker related, instead of softfailing. That exception is only used by the requires_docker marker/decorator. The tests themselves have no awareness.

From code perspective, it only knows about: in case of error raise, in addition to the current logic already in place to capture the error and providing some non-aws response. I only had to create some hoops for the run-in-separate-thread situation to have the exception keep bubbling up.

Alternative is to add various checks in the tests to see if some data returned contains some indication that docker is having issues, which I find really ugly and does couples the code and tests even closer (besides from the many additional non-dry code getting in the various tests).

It did help me uncover some corner cases that I did not detect before as likely/potentially impacted by not having docker running.

Most users will not use the xfails method, but it's a good one for sanity checks IMHO.

I'll have to run with servermode as well, and see how that works out / can be handled.

If you prefer, I can remove the exception related bits together with the xfail.

hans-d: instead of optionally raising this custom exception, we can make sure that when docker acts up, also a proper log message is generated.

moto/awslambda/models.py
```diff
@@ -830,6 +829,7 @@ def _invoke_lambda(self, event: Optional[str] = None) -> Tuple[str, bool, str]:
         except docker.errors.DockerException as e:
             # Docker itself is probably not running - there will be no Lambda-logs to handle
             msg = f"error running docker: {e}"
+            logger.error(msg)
```
bblommers, line 832: Looks like this change is no longer necessary

moto/batch/models.py
```diff
@@ -483,6 +483,7 @@ def __init__(
         self.job_stopped_at = datetime.datetime(1970, 1, 1)
         self.job_stopped = False
         self.job_stopped_reason: Optional[str] = None
+        self.raised_exception: Optional[Exception] = None
```
bblommers, line 486: Same for all changes in this file

moto/settings.py
```diff
@@ -32,6 +32,11 @@
 
 PRETTIFY_RESPONSES = bool(os.environ.get("MOTO_PRETTIFY_RESPONSES", False))
 
+# Fully skip test that require docker
+SKIP_REQUIRES_DOCKER = bool(os.environ.get("TESTS_SKIP_REQUIRES_DOCKER", False))
+# Run test that require docker, but allow failing when docker exceptions are raised
+RAISE_DOCKER_EXCEPTION = bool(os.environ.get("TESTS_XFAIL_REQUIRES_DOCKER", False))
```
bblommers, line 38: And this setting

bblommers (changes requested):

moto/awslambda/models.py
hans-d: I would keep the logging of Docker errors, as in case of fail situations this would be very useful to have. Eg a new testcase is added, where the `@requires_docker` is not used. In case of a failing test, it is not always obvious when looking at the pytest outcome why this was caused. Being able to use eg '-o log_cli=true' would provide help here.

moto/batch/models.py
hans-d: will check, I knew there was something I did miss...

moto/settings.py
hans-d: hmm, that should have been gone already. Will check.

bblommers (approved):
LGTM! Thanks @hans-d

PR #6041, opened 2023-03-08
Fix #6040 by using _filters_from_querystring() method

### LINKED ISSUE #6040: ec2.describe_security_group_rules does not use filter
## how to reproduce the issue
```
import boto3
from moto import mock_ec2

with mock_ec2():
    config = Config(
            region_name="eu-west-1"
            )
    client = boto3.client("ec2", config=config)

    # Create dummy sg
    response_sg = client.create_security_group(
            Description="description",
            GroupName="dummy-sg"
            )
    print("SG creation")
    print(response_sg)


    response_rules = client.describe_security_group_rules(
            Filters=[
                {
                    "Name":"group-id",
                    "Values": [
                        response_sg["GroupId"],
                    ],
                }
            ]
        )

    print()
    print("Response")
    print(response_rules)

    print()
    print("Rules:")
    for rule in response_rules["SecurityGroupRules"]:
        print(rule)

```
## what you expected to happen
Only the default egress rules should appear for this SG

## what actually happens
Rules for all SG appears (the default sg plus my dummy sg in this case)

## what version of Moto you're using
```
pip freeze | grep oto
boto3==1.26.86
botocore==1.29.86
moto==4.1.4
```

CiolFr:


bblommers (changes requested):
Hi @CiolFr, looks like a simple solution. Would you mind just adding a unit test for this scenario?

CiolFr:
Hi,

I've just finished rework the test case.
The previous one was working because filters does not applied but it shouldn't.
The new test case, create a security group. Get it back and test for the default rules.
Let me know if it's ok for you.

bblommers (approved):
Awesome - thanks again @CiolFr!

PR #6043, opened 2023-03-09
S3: Restored archived objects must be retrievable

viren-nadkarni:
This PR fixes a bug where S3 objects with deep archive storage class could not be retrieved after restoration was complete.


bblommers (approved):
Nice one - thanks @viren-nadkarni

PR #6045, opened 2023-03-09
[Glue] Add support to more fields in Glue job

szemek:
Add support to more fields in Glue job
- `CodeGenConfigurationNodes`
- `ExecutionClass`
- `SourceControlDetails`

### DIFF
--- a/tests/test_glue/test_glue.py
+++ b/tests/test_glue/test_glue.py
@@ -89,30 +89,30 @@ def test_get_job_exists():
     }
     job_name = create_test_job_w_all_attributes(client, **job_attributes)
     job = client.get_job(JobName=job_name)["Job"]
-    job.should.have.key("Name").equals(job_name)
-    job.should.have.key("Description")
-    job.should.have.key("LogUri")
-    job.should.have.key("Role")
-    job.should.have.key("ExecutionProperty").equals({"MaxConcurrentRuns": 123})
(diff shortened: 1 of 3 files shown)


szemek:
@bblommers if you find time, could you take a look? Thanks in advance!

tests/test_glue/test_glue.py
```diff
@@ -107,6 +110,9 @@ def test_get_job_exists():
     job.should.have.key("SecurityConfiguration")
     job.should.have.key("NotificationProperty")
     job.should.have.key("GlueVersion")
+    job.should.have.key("CodeGenConfigurationNodes")
```
bblommers, line 113: Minor nitpick - could you rewrite these to use traditional asserts? 
`assert x in job`

We've decided to move away from sure, the library that allows `.should.have`-notation, because it makes debugging more difficult than it needs to be. 
This will be a gradual process, but we're starting with ensuring that new PR's do not use it.

bblommers (changes requested):

tests/test_glue/test_glue.py
szemek: @bblommers sure, I changed to assert in the next commit. Please let me know if that's enough or I should change more in this file.

bblommers (approved):
LGTM! Thank you for the PR @szemek

PR #6054, opened 2023-03-11
Logs: add `describe_queries`

Killian168:
Hello,

This adds support for the `describe_queries` endpoint as described [here](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/logs.html)

This implementation just sets the status of a query to "Complete" which seems like the simplest solution for now. I did have an idea to try creating a background task when a query is started that runs the query through the supported states`'Scheduled'|'Running'|'Complete'|'Failed'|'Cancelled'|'Timeout'|'Unknown'` but I don't see anything else like this in Moto so just went with the straight "Complete" at creation. Some guidance/advice here would be appreciated.


bblommers:
Hi @Killian168, you can change the status using State Transitions: http://docs.getmoto.org/en/latest/docs/contributing/development_tips/new_state_transitions.html

If you're OK with just having the `Complete`-status, than feel free to leave it like this. We can always implement the Transitions later on, when people do need it.

bblommers:
This is now implemented by #6730

PR #6055, opened 2023-03-11
Some Athena fixes and additions

paulmartins:
* Add the 3 athena mocks below:
  * `list_named_queries()`  with pagination
  * `create_prepared_statement()`
  * `get_prepared_statement()`

* Enable default `primary` workgroup in athena_backend

* Ensure Named Queries are always part of a workgroup (default to primary)


bblommers (approved):
LGTM! Thanks for adding these features @paulmartins

PR #6063, opened 2023-03-12
fix: add typing to AMIS

### LINKED ISSUE #6062: Type error using AMIS[0]
```python
from moto.ec2.models.amis import AMIS

an_ami =  AMIS[0]["ami_id"]
```

pyright/pylance on line 3:

```
Argument of type "Literal[0]" cannot be assigned to parameter "__key" of type "str" in function "__getitem__"
  "Literal[0]" is incompatible with "str" Pylance[reportGeneralTypeIssues](https://github.com/microsoft/pyright/blob/main/docs/configuration.md#reportGeneralTypeIssues)
```

Because `AMIS` is `Any`.

moto 4.1.4

tekumara:
resolves #6062

### DIFF
--- a/moto/ec2/models/amis.py
+++ b/moto/ec2/models/amis.py
@@ -21,9 +21,9 @@ from ..utils import (
 
 if "MOTO_AMIS_PATH" in environ:
     with open(environ["MOTO_AMIS_PATH"], "r", encoding="utf-8") as f:
-        AMIS = cast(List[Dict[str, Any]], json.load(f))
+        AMIS: List[Dict[str, Any]] = json.load(f)
 else:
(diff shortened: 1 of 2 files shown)


bblommers (changes requested):
Can we make this implicit, instead of casting? I feel that this would be more readable, and more in-line with our typing approach so far:
```
AMIS: List[Dict[str, Any]] = json.load(f)
```

This does mean we also have to  change the return-type of `load_resource`. It currently returns `Dict[str, Any]`, which is incorrect. I'd be OK with with just making the return-type `Any`, to make it clear it's up to the caller to determine what format it expects.

tekumara:
Thanks for reviewing this @bblommers ... I prefer the type annotations over casting too, I've added that.

bblommers (approved):
Awesome - thanks @tekumara!

Issue #6035, opened 2023-03-08
use of sure package for asserts make pytest output hard to use


hans-d:
When running a test (which uses pytest), on a failing test and when using regular assert statements, various good info is shown already by pytest.

Using `sure`, this generates a lot of additional stack trace garbage, making it hard to discover the issue.
I've noticed that not all modules are using sure, so by removing `sure` we get more consistency in the code and more readability in case of a test failure.

Noticed when deep-diving into #6026. I'm ok to pick this up, when that one is wrapped.

bpandola:
🙋  Raising my hand here in support of dropping `sure`.  It served us well for a while, but I don't think the value is there anymore, especially after `moto` made the switch from `nose` to `pytest`.  The way `sure` hijacks the Python internals might have been magical at first blush, but these days it seems like a liability.  It adds numerous attributes to every object that hinders active debugging; it utterly confuses PyCharm, resulting in countless errors and warnings; and, as @hans-d points out, it doesn't play well with `pytest`.

All that said, removing `sure` seems like it would require a *ton* of manual effort.  I know there are tools that automate moving from `nose`- and `unittest`-based assert methods to `pytest`-style  assertions, but I doubt something similar exists for moving away from `sure`.

@bblommers Maybe, at a minimum, we advise against using `sure` for any new tests?

hans-d:
some fragments directly using`unittest` are also to be found.
Working through this code to help change this will aid me in getting a better grip of moto internals. Expecting it to be done in pieces, like the mypy chnages.

bblommers:
Well, I do prefer the sure-style, as it reads better IMO, but you're both right that there's a long list of disadvantages. 

The approach makes sense as well - for now we can just stop new tests from using `sure`, and whenever possible we'll gradually upgrade the existing tests, file by file, service by service.

bblommers:
I haven't found a way to automate everything, but some regexes that I found helpful when tackling this:

F `(\S+)\.should.equal\((.*)\)`
F `(\S+)\.should.equal\(\s+(\S+)\s+\)`
R `assert $1 == $2`

F `(\S+)\.should.not.equal\((.*)\)`
R `assert $1 != $2`

F `(\S+.+)\.should.contain\((.*)\)`
F `(\S+.+)\.should.contain\(\s+(".+")\s+\)`
R `assert $2 in $1`

F `(\S+.+)\.shouldnt.contain\((.*)\)`
[...]

PR #6073, opened 2023-03-14
Techdebt: Sure - Organizations

hans-d:
- Removing sure: #6035
- pytest-ified, by adding fixtures and parametrization (instead of the calling out to functions) and making it more dry
- focus on the actual testing needed, not test because we can
- grouping tests into classes, to give a bit more structure
- grouping tests from 1 big file into 5 main areas
- uncovered some missing test cases
- removed unused code

PS: noting how the various exceptions returned differ in structure re message and code. Did not investigate if this is how it should be. Using the various parametrize setups makes this clearer.


bblommers:
@hans-d I realize the work that has gone into this, and thank you for that. There is undoubtedly space to optimize our testing strategy here.
However, there is no way for me to meaningfully review a PR this large. I will happily look at a PR that only removes the `sure`-dependency. If we really want to improve the structure, let's look at that separately.

PR #6077, opened 2023-03-15
fix: adds "DeleteResponse" to `delete_object` call

jashugan:
fixes https://github.com/getmoto/moto/issues/6076

### DIFF
--- a/tests/test_s3/test_s3.py
+++ b/tests/test_s3/test_s3.py
@@ -1634,7 +1634,7 @@ def test_delete_versioned_bucket_returns_meta():
 
     # Delete the object
     del_resp = client.delete_object(Bucket="blah", Key="test1")
-    assert "DeleteMarker" not in del_resp
+    assert del_resp["DeleteMarker"] is False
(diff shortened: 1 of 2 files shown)


tests/test_s3/test_s3.py
```diff
@@ -1630,19 +1630,26 @@ def test_delete_versioned_bucket_returns_meta():
         Bucket="blah", VersioningConfiguration={"Status": "Enabled"}
     )
 
-    client.put_object(Bucket="blah", Key="test1", Body=b"test1")
+    put_resp = client.put_object(Bucket="blah", Key="test1", Body=b"test1")
 
     # Delete the object
     del_resp = client.delete_object(Bucket="blah", Key="test1")
-    assert "DeleteMarker" not in del_resp
+    assert del_resp["DeleteMarker"] is False
```
bblommers, line 1637: Is this correct? In the ticket you noted that `Deleting an object by key only will create a delete marker for it. The response will include the DeleteMarker flag set to True`

PR #6083, opened 2023-03-17
added `TagList` to DatabaseSnapshot Model as specified in latest AWS …

cloudsurf-digital:
…Models of RDS

Here are the docs: https://docs.aws.amazon.com/pdfs/AmazonRDS/latest/APIReference/rds-api.pdf


cloudsurf-digital:
added a test

bblommers (approved):
Awesome, thanks @cloudsurf-digital!

PR #6095, opened 2023-03-20
batch: use specified network IP address

d4l3k:
`AWS_BATCH_JOB_MAIN_NODE_PRIVATE_IPV4_ADDRESS` is empty when a network is set using `MOTO_DOCKER_NETWORK_NAME`. This checks if the network is present in the container networks and uses that IP address if present. Custom networks are required to use container DNS names for multi node networking.

Test plan:

```
pytest tests/test_batch/test_batch_jobs
MOTO_DOCKER_NETWORK_NAME=foo pytest tests/test_batch/test_batch_jobs
```

This was discovered when adding distributed tests using moto in https://github.com/pytorch/torchx/pull/710 Those tests now pass.

### DIFF
--- a/moto/batch/models.py
+++ b/moto/batch/models.py
@@ -724,6 +724,9 @@ class Job(threading.Thread, BaseModel, DockerModel, ManagedState):
                 else {}
             )
 
+            network_mode = settings.moto_network_mode()
+            network_name = settings.moto_network_name()
+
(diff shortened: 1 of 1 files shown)


d4l3k:
@bblommers when you get time, reviews very appreciated :) -- this is a follow up to https://github.com/getmoto/moto/pull/5840

PR #6123, opened 2023-03-24
Issue#6039: Fix Cloudformation delete stack implementation

priyaoz:
Fixing `delete_stack_instances`  implementation by clearing the `stack_instances` after iterating and deleting the stack_instances. 

ref: https://github.com/getmoto/moto/issues/6039

### DIFF
--- a/moto/cloudformation/models.py
+++ b/moto/cloudformation/models.py
@@ -349,8 +349,7 @@ class FakeStackInstances(BaseModel):
         for instance in self.stack_instances:
             if instance.region_name in regions and instance.account_id in accounts:
                 instance.delete()
-        # clear the stack_instances in the end to prevent IndexError
-        self.stack_instances.clear()
+                self.stack_instances.remove(instance)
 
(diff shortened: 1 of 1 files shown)


moto/cloudformation/models.py
```diff
@@ -346,10 +346,11 @@ def update(
                 instance.parameters = parameters or []
 
     def delete(self, accounts: List[str], regions: List[str]) -> None:
-        for i, instance in enumerate(self.stack_instances):
+        for instance in self.stack_instances:
             if instance.region_name in regions and instance.account_id in accounts:
                 instance.delete()
-                self.stack_instances.pop(i)
+        # clear the stack_instances in the end to prevent IndexError
+        self.stack_instances.clear()
```
bblommers, line 353: This isn't quite right. See the failing test - we should only remove the stack instances that match the provided regions/accounts

priyaoz: Thanks for the feedback , that makes sense - will look into alernative fix.

bblommers (approved):
LGTM! Thanks @priyaoz

PR #6124, opened 2023-03-24
CloudFormation: Implement Fn::Sub variable mapping

jmsanders:
https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/intrinsic-function-reference-sub.html

This places a lot of faith in the existing parsing. I didn't exhaustively test recursion, but I did check that more "complex" mappings (like ones that provide their own calls to Fn::If) work correctly.

### DIFF
--- a/moto/cloudformation/parsing.py
+++ b/moto/cloudformation/parsing.py
@@ -1,3 +1,4 @@
+import string
 import functools
 import json
 import logging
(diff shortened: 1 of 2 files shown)


bblommers (approved):
LGTM - Thanks @jmsanders!

PR #6125, opened 2023-03-24
CloudFormation, Logs: Resolve LogGroup physical id and attributes

jmsanders:
https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-logs-loggroup.html#aws-resource-logs-loggroup-return-values

This brings CloudFormation's Ref behavior more in line with actual AWS behavior. Previously, it returned an instance of a LogGroup model.

I didn't add a test case for this but I'm happy to do so if requested.


bblommers (approved):
LGTM! Thanks @jmsanders

PR #6129, opened 2023-03-26
CloudFormation: Add coverage checklist

jmsanders:
This change introduces a checklist similar to IMPLEMENTATION_COVERAGE.md
to document moto's adherence to the full CloudFormation specification.

The script (invoked via `make cloudformation_coverage`) finds any moto
models that implement `moto.core.common_models.CloudFormationModel` and
checks to see if all CRUD methods are implemented. For `has_cfn_attr`,
it also checks to see if all of the attributes exposed by `Fn::GetAtt` are
implemented. It does not check to see if `physical_resource_id` is
implemented because as far as I can tell, the published AWS spec doesn't
include that information:

https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/cfn-resource-specification.html

A more aggressive approach would be to remove default implementations
from all the abstract methods and allow Python to throw for any
unimplemented method. But that would obviously break a lot of things.

Instead, I think the checklist represents good documentation for users
about what they can and can't expect moto to handle when parsing their
CloudFormation templates. It also serves as a nice to-do list of small
improvements for contributors to add (and I'll likely add a few myself).
Many of these would be particularly good "first issues" for first time
contributors because in general, these methods just call existing
methods.

### DIFF
--- a/scripts/cloudformation_coverage.py
+++ b/scripts/cloudformation_coverage.py
@@ -1,0 +1,111 @@
+#!/usr/bin/env python
+import importlib
+import json
+import mock
+import requests
+
+import moto
+
+# Populate CloudFormationModel.__subclasses__()
+moto.mock_all()
+
+
+def check(condition):
+    if bool(condition):
+        return "x"
+    else:
+        return " "
+
+
+def is_implemented(model, method_name):
+    # method_name in model.__dict__ will be True if the method
+    # exists on the model and False if it's only inherited from
+    # CloudFormationModel.
+    return hasattr(model, method_name) and method_name in model.__dict__
+
+
+class CloudFormationChecklist:
+    def __init__(self, resource_name, schema):
+        self.resource_name = resource_name
+        self.schema = schema
+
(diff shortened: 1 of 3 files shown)


scripts/cloudformation_coverage.py
```diff
@@ -0,0 +1,111 @@
...
+    @property
+    def expected_attrs(self):
+        return list(self.schema.get("Attributes", {}).keys())
+
+    @property
+    def missing_attrs(self):
+        missing_attrs = []
+        for attr in self.expected_attrs:
+            try:
+                # TODO: Change the actual abstract method to return False
```
jmsanders, line 75: I'm only patching this here because I think changing the actual abstract method to return False could maybe be considered a breaking change. But I do think it's more accurate to default to returning False than it is to default to returning True.

bblommers (approved):
This makes a lot of sense @jmsanders. Easier for users to discover what's what, and easier for developers as well to keep track of everything.

Thank you for adding this to Moto!

PR #6136, opened 2023-03-27
ecr.put_image - Support imageManifestMediaType parameter (issue 6135)

iantempest:
Add support for the optional imageManifestMedia, parameter as detailed in https://github.com/getmoto/moto/issues/6135

### DIFF
--- a/tests/test_ecr/test_ecr_boto3.py
+++ b/tests/test_ecr/test_ecr_boto3.py
@@ -350,6 +350,44 @@ def test_put_image():
     response["image"]["registryId"].should.equal(ACCOUNT_ID)
 
 
+@mock_ecr
+def test_put_image_without_mediatype():
+    client = boto3.client("ecr", region_name="us-east-1")
+    _ = client.create_repository(repositoryName="test_repository")
+
+    image_manifest = _create_image_manifest()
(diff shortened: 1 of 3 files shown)


tests/test_ecr/test_ecr_boto3.py
```diff
@@ -350,6 +350,44 @@ def test_put_image():
...
+@mock_ecr
+def test_put_image_without_mediatype():
+    client = boto3.client("ecr", region_name="us-east-1")
+    _ = client.create_repository(repositoryName="test_repository")
+
+    image_manifest = _create_image_manifest()
+    _ = image_manifest.pop("mediaType")
+
+    error_msg = "image manifest mediatype not provided in manifest or parameter"
+    client.put_image.when.called_with(
```
bblommers, line 362: Hi @iantempest, thanks for the PR!

Looks like this line is breaking the Server mode tests. It probably works better if you use the `with pytest.raises` to verify the error message. This is already used in `test_create_repository_error_already_exists` if you want to see an example.

bblommers (changes requested):

bblommers (approved):
LGTM! Thank you for fixing this @iantempest

PR #6138, opened 2023-03-27
feat: dynamodb AttributeUpdate ADD/PUT/DELETE for NS

Basti110:
This should handle `ADD`, `PUT` and `DELETE` to set for number sets (NS) in dynamodb `update_item`
- Added test coverage
- tested with AWS backend


bblommers (approved):
LGTM! Thanks for adding this feature @Basti110

PR #6145, opened 2023-03-28
Combined Java Dependabot Updates

bpandola:
Not sure if this is the best (or only) solution, but the Java dependabot PRs were gumming up the works, so I reset the base branch on those PRs so we could test the changes in combination (on this branch).  This seemed low risk, but was a bit of a manual process.  We may want to explore some of the dependabot configuration options to see if there's a better way going forward.

### DIFF
--- a/other_langs/tests_java/pom.xml
+++ b/other_langs/tests_java/pom.xml
@@ -95,7 +95,7 @@
         </plugin>
         <plugin>
           <artifactId>maven-deploy-plugin</artifactId>
-          <version>3.1.0</version>
+          <version>3.1.1</version>
         </plugin>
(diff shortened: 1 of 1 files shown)


bblommers:
Maybe the simplest improvement would be to [only check once a month](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#scheduleinterval).
Just so we don't have to deal with it too often.

Another useful change would be not to run the test suite. I.e., if there's a Java dependency update, there's no need to run the Python/TF tests. Don't know how easy that is to achieve, though.

Dependabot has a concept of maximum branch length. It's not available in GitHub yet, see https://github.com/dependabot/dependabot-core/issues/396

I wonder what happens if (when it becomes available) we set the length to 10, so the branch name becomes `dependabot`. 
Would it just reuse the same PR for multiple changes?

PR #6152, opened 2023-03-30
EC2: Implement GetLaunchTemplateData

viren-nadkarni:
<https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_GetLaunchTemplateData.html>


bblommers (approved):
LGTM! Thanks @viren-nadkarni

PR #6167, opened 2023-04-02
CloudFormation: Implement Fn::ToJsonString

jmsanders:
https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/intrinsic-function-reference-ToJsonString.html

### DIFF
--- a/moto/cloudformation/parsing.py
+++ b/moto/cloudformation/parsing.py
@@ -246,6 +246,14 @@ def clean_json(resource_json: Any, resources_map: "ResourceMap") -> Any:
                 result.append(f"{region}{az}")
             return result
 
+        if "Fn::ToJsonString" in resource_json:
+            return json.dumps(
+                clean_json(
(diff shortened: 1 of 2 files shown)


moto/cloudformation/parsing.py
```diff
@@ -246,6 +246,14 @@ def clean_json(resource_json: Any, resources_map: "ResourceMap") -> Any:
                 result.append(f"{region}{az}")
             return result
 
+        if "Fn::ToJsonString" in resource_json:
```
jmsanders, line 249: I just want to say kudos to whoever wrote the initial implementation of this parsing module - I've found it remarkably straightforward to extend. 🥳

bblommers: That would be @spulec! https://github.com/getmoto/moto/pull/111

bblommers (approved):
LGTM - thanks again @jmsanders

PR #6173, opened 2023-04-04
Add missing double quota to Cognito special characters.

bartoszj:
Cognito special characters include a double quota mark (`"`), which is missing in the current version: https://docs.aws.amazon.com/cognito/latest/developerguide/user-pool-settings-policies.html


bblommers (approved):
Good catch @bartoszj. Thanks for the PR, and welcome to Moto!

PR #6174, opened 2023-04-04
Validate password in NEW_PASSWORD_REQUIRED

mwpb:
Hi, thanks for creating this useful library. Following on from https://github.com/getmoto/moto/issues/5259 I think that AWS Cognito validates the `new_password` in the `NEW_PASSWORD_REQUIRED` challenge. Therefore it might be nice to validate in `CognitoIdpBackend.respond_to_auth_challenge` in moto also. That is provided in this PR.


bblommers (approved):
LGTM - thanks for the PR @mwpb!

PR #6180, opened 2023-04-06
EC2: Fix egress rules used in ingress revoke method

jiggak:
This fixes the `revoke_security_group_ingress` effectively removing all ingress rules when attempting to remove one rule while leaving the rest of the rules in the group.


bblommers (approved):
LGTM - thanks for the fix @jiggak, and welcome to Moto!

PR #6181, opened 2023-04-06
Allow docker compose to set port 

deepakmeena635:
Hi there, 

Currently, I'm deploying a moto container with other aws and localstack containers as others don't support athena.
It's being done with a docker-compose file which makes it easily manageable. But the problem is on some of my teams machines port 5000 is blocked by os and the process respawns when killed.

It'd be awesome if we can allow the port to be configured in the dockerfile. if there are other solutions that allow it to be done via docker compose I'll be more than happy to implement them.


bblommers:
Hi @deepakmeena635, the server-script already has a `MOTO_PORT` environment variable that you can use:

https://github.com/getmoto/moto/blob/bbb07b4e7ede832d69b8fe0089e0707c915710fc/moto/server.py#L41

Does that fix your problem?

deepakmeena635:
thanks @bblommers, it solves my problem is there any documentation that i can update for anyone else wondering about this

mikegrima:
Is this ready to merge?

deepakmeena635:
Thanks @bblommers and @mikegrima done from my side

Dockerfile
bblommers, line 1: The changes in this file are not necessary, right?

bblommers (changes requested):

Dockerfile
deepakmeena635: yes one second

deepakmeena635: done, thanks for pointing it out

deepakmeena635:
wait

Dockerfile
deepakmeena635: done now

bblommers (approved):
Awesome, thanks for adding this @deepakmeena635!

PR #6184, opened 2023-04-07
issue 1438 sqs fifo dedupe

rafcio19:
Dedupe according to logic in https://docs.aws.amazon.com/AWSSimpleQueueService/latest/APIReference/API_SendMessage.html
and
https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/using-messagededuplicationid-property.html

### DIFF
--- a/moto/sqs/models.py
+++ b/moto/sqs/models.py
@@ -525,17 +525,19 @@ class Queue(CloudFormationModel):
         ]
 
     def add_message(self, message):
-        if (
-            self.fifo_queue
-            and self.attributes.get("ContentBasedDeduplication") == "true"
-        ):
-            for m in self._messages:
-                if m.deduplication_id == message.deduplication_id:
-                    diff = message.sent_timestamp - m.sent_timestamp
-                    # if a duplicate message is received within the deduplication time then it should
-                    # not be added to the queue
-                    if diff / 1000 < DEDUPLICATION_TIME_IN_SECONDS:
-                        return
+        if self.fifo_queue:
+
+            # the cases in which we dedupe fifo messages
+            # from https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/using-messagededuplicationid-property.html
+            if self.attributes.get("ContentBasedDeduplication") == "true" or \
+                    message.deduplication_id:
+                for m in self._messages:
+                    if m.deduplication_id == message.deduplication_id and m.body_md5 == message.body_md5 :
+                        diff = message.sent_timestamp - m.sent_timestamp
+                        # if a duplicate message is received within the deduplication time then it should
+                        # not be added to the queue
+                        if diff / 1000 < DEDUPLICATION_TIME_IN_SECONDS:
+                            return
 
         with self._messages_lock:
             self._messages.append(message)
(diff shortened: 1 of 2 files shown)


mikegrima:
@rafcio19 Ready to merge?

tests/test_sqs/test_sqs.py
```diff
@@ -2742,7 +2747,8 @@ def test_receive_messages_with_message_group_id_on_visibility_timeout():
     with freeze_time("2015-01-01 12:00:00"):
         sqs = boto3.resource("sqs", region_name="us-east-1")
         queue = sqs.create_queue(
-            QueueName="test-queue.fifo", Attributes={"FifoQueue": "true"}
+            QueueName="test-queue.fifo",
+            Attributes={"FifoQueue": "true", "ContentBasedDeduplication": "true"},
```
mikegrima, line 2751: Are there any tests where this is set to `false`?

rafcio19: The [default ](https://github.com/getmoto/moto/blob/0520326626ab796e0dd7953a0bf6ac85f96a6d8b/moto/sqs/models.py#L273) for `ContentBasedDeduplication` is `false`, so all the cases where `ContentBasedDeduplication` is not explicitly set `true` for a fifo queue AND a message is sent test for that scenario.

Apart from the tests I added, here is a few:

1. https://github.com/getmoto/moto/blob/0520326626ab796e0dd7953a0bf6ac85f96a6d8b/tests/test_sqs/test_sqs.py#L572
2. https://github.com/getmoto/moto/blob/0520326626ab796e0dd7953a0bf6ac85f96a6d8b/tests/test_sqs/test_sqs.py#L1315
3. https://github.com/getmoto/moto/blob/0520326626ab796e0dd7953a0bf6ac85f96a6d8b/tests/test_sqs/test_sqs.py#L1673
4. https://github.com/getmoto/moto/blob/0520326626ab796e0dd7953a0bf6ac85f96a6d8b/tests/test_sqs/test_sqs.py#L2086
 

I think mine is a breaking change: it will break existing tests where `ContentBasedDeduplication` is not explicitly set to `true` OR  `MessageDeduplicationId` is not passed. However, it reflects more accurately what happens in AWS. 

Ready to merge I reckon 🤘

moto/sqs/models.py
```diff
@@ -758,6 +763,24 @@ def send_message(
 
         queue = self.get_queue(queue_name)
 
+        if (
```
bblommers, line 766: Just a nitpick - I think these two checks should be behind an `is fifo` check.
The current approach works, because `ContentBasedDeduplication` is only set for `fifo`-queues. In other words: `queue.attributes.get("ContentBasedDeduplication")` always returns None for a non-fifo queue. 

But it's more readable IMO if we make it explicit that these checks are only applicable for FIFO queues.

bblommers (changes requested):

moto/sqs/models.py
rafcio19: Makes sense

rafcio19: Updated the PR

bblommers (approved):
Awesome - thank you so much for fixing this @rafcio19!

PR #6188, opened 2023-04-07
Added support for AWS Config Retention Configuration

mikegrima:
- Added AWS Config Retention Configuration support
- Also added S3 KMS Key ARN support for the Delivery Channel

### DIFF
--- a/docs/docs/services/config.rst
+++ b/docs/docs/services/config.rst
@@ -62,7 +62,7 @@ config
 - [ ] delete_remediation_configuration
 - [ ] delete_remediation_exceptions
 - [ ] delete_resource_config
-- [ ] delete_retention_configuration
+- [X] delete_retention_configuration
 - [ ] delete_stored_query
 - [ ] deliver_config_snapshot
 - [ ] describe_aggregate_compliance_by_config_rules
@@ -91,7 +91,7 @@ config
 - [ ] describe_remediation_configurations
 - [ ] describe_remediation_exceptions
 - [ ] describe_remediation_execution_status
-- [ ] describe_retention_configurations
+- [X] describe_retention_configurations
 - [ ] get_aggregate_compliance_details_by_config_rule
 - [ ] get_aggregate_config_rule_compliance_summary
 - [ ] get_aggregate_conformance_pack_compliance_summary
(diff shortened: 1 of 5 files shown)


mikegrima:
I didn't test the deletion logic with real AWS yet, but based on the docs it should work. I'm going to do a formal test on Monday. The other APIs work correctly.

bblommers (approved):
LGTM!

mikegrima:
Hmm it looks like the test server tests don't like my parameter validation disablement. I'll fix that on Monday.

mikegrima:
BTW, I wanted to document this as I was having some strange test failures related to datetimes being hours off when I was debugging the server tests failing.

Instead of just following the simple instructions here: http://docs.getmoto.org/en/latest/docs/contributing/development_tips/tests.html?highlight=server#servermode-tests which I found after doing all this work :upside_down_face:, I instead tried digging in the makefile and seeing how the build environment sets up the container. I saw this command being run:

```
docker run --rm -t --name motoserver -e TEST_SERVER_MODE=true -e AWS_SECRET_ACCESS_KEY=server_secret -e AWS_ACCESS_KEY_ID=server_key -v `pwd`:/moto -p 5000:5000 -v /var/run/docker.sock:/var/run/docker.sock python:3.7-buster /moto/scripts/ci_moto_server.sh
```

Running that in a terminal will get the server ready for tests to run. Another way to do it is to run `docker build .`, grabbing the image ID, and then running: 

```
docker run --rm -t --name motoserver -e TEST_SERVER_MODE=true -e AWS_SECRET_ACCESS_KEY=server_secret -e AWS_ACCESS_KEY_ID=server_key -p 5000:5000 sha256:IMAGE-ID-HERE
```

Doing this though was causing strange datetime tests to fail, and it was because the timezones weren't set up properly. I needed to add a `TZ=` variable to the container (in my case for US Eastern Time):

```
docker run --rm -t --name motoserver -e TEST_SERVER_MODE=true -e AWS_SECRET_ACCESS_KEY=server_secret -e AWS_ACCESS_KEY_ID=server_key -e TZ=America/New_York -p 5000:5000 sha256:IMAGE-ID-HERE
```

After that all my tests were working perfectly. So if anyone sees issues with running moto server in a docker container and the datetimes are off, this is why (you need to set your timezone via the `TZ` environment variable to the container). Of course, if you just run `moto_server` everything just works perfectly.

PR #6195, opened 2023-04-10
fix/describe-target-health: return unavailable target

mtulio:
## Issue:

describe-target-health is not returning the expected structure when the target is new

## Steps to reproduce:

- create the target group (without registering the target)
- send describe-target-health with a target as a parameter

## Current results:

```
botocore.errorfactory.InvalidTargetException: An error occurred (InvalidTarget) when calling the DescribeTargetHealth operation: The specified target does not exist or is not in the same VPC as the target group.
```

## Expected results

```
Valid `FakeHealthStatus`  with reason `Target.NotRegistered`
```

## References

- https://docs.aws.amazon.com/elasticloadbalancing/latest/APIReference/API_TargetHealth.html
-

### DIFF
--- a/moto/elbv2/models.py
+++ b/moto/elbv2/models.py
@@ -152,7 +152,17 @@ class FakeTargetGroup(CloudFormationModel):
     def health_for(self, target: Dict[str, Any], ec2_backend: Any) -> FakeHealthStatus:
         t = self.targets.get(target["id"])
         if t is None:
-            raise InvalidTargetError()
+            port = self.port
+            if 'port' in target:
+                port = target["port"]
+            return FakeHealthStatus(
+                target["id"],
+                port,
+                self.healthcheck_port,
+                "unavailable",
+                "Target.NotRegistered",
+                "Target is not registered",
+            )
         if t["id"].startswith("i-"):  # EC2 instance ID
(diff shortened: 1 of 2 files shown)


moto/elbv2/models.py
```diff
@@ -152,7 +152,17 @@ def deregister_terminated_instances(self, instance_ids: List[str]) -> None:
...
         if t is None:
-            raise InvalidTargetError()
+            port = self.port
+            if 'port' in target:
+                port = target["port"]
+            return FakeHealthStatus(
+                target["id"],
+                port,
+                self.healthcheck_port,
+                "unavailable",
```
mtulio, line 162: I am not sure about the status value, I assumed it for the desired `Reason=Target.NotRegistered`

bblommers (changes requested):
Looks like our linter is failing atm.

Would be good to have a test for this as well - I don't think this scenario is covered yet in our tests

mtulio:
> Looks like our linter is failing atm.
> 
> Would be good to have a test for this as well - I don't think this scenario is covered yet in our tests

Hey @bblommers , linter fixed and unit added. LMK WDYT. Thanks

bblommers (approved):
LGTM - thanks again @mtulio!

PR #6205, opened 2023-04-12
BatchSimple: jobs can fail or succeed

rafcio19:
Potential solution for https://github.com/getmoto/moto/issues/5168

### DIFF
--- a/moto/batch_simple/README.md
+++ b/moto/batch_simple/README.md
@@ -1,0 +1,7 @@
+# batch_simple
+
+Batch jobs run under `batch_simple` will succeed by default. To make the jobs fail:
+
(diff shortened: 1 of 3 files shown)


bblommers (changes requested):
Our internal tests just use a random `job_name`, which makes it easy to alter. People calling the Batch-service in production probably have a predetermined naming-strategy that may not be easy to change.

I would suggest using an environment variable instead. It's in-line with other configuration options that are used in Moto, it' easy to configure and easy for us to maintain.

One user asked for the option for Batch to fail after a certain period of time. So maybe something like:
`MOTO_SIMPLE_BATCH_FAIL_AFTER=5` 
to indicate a failure after 5 seconds. Any non-negative integer can be used of course - `0` will probably be the most popular to make it fail immediately.

If the env isn't set, then the status can be `SUCCESS` as usual.

PR #6210, opened 2023-04-13
Store temporary access key in the correct account backend

dfangl:
## Motivation

Currently, when performing sts.AssumeRole, we are storing the AccessKey in the account backend of the requesting account.
This, however, leads to the key not being found e.g. for sts.GetCallerIdentity, and also leads to wrong behavior in this method: https://github.com/getmoto/moto/blob/90d67db6eae20ccde57356b694e013fd5864d14a/moto/iam/models.py#L72 which is necessary for account detection.

## Changes
* Store the access key in the store of the account the role belongs to instead of the one assuming the role. There are points to be made for both, depending on who you think the access key belongs to, but otherwise we would need to adapt the surrounding logic, so I just went forward and assumed the access key always belongs to the account of the role.
* Test the behavior by asserting on the output of sts.GetCallerIdentity

### DIFF
--- a/moto/sts/models.py
+++ b/moto/sts/models.py
@@ -99,7 +99,8 @@ class STSBackend(BaseBackend):
             duration,
             external_id,
         )
-        self.assumed_roles.append(role)
+        account_backend = sts_backends[account_id]["global"]
+        account_backend.assumed_roles.append(role)
         return role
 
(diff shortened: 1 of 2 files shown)


bblommers (approved):
Good catch @dfangl - thanks for the PR!

PR #6220, opened 2023-04-17
retrieve S3_UPLOAD_PART_MIN_SIZE from env variable

agilan2001:
Made the retrieve S3_UPLOAD_PART_MIN_SIZE value configurable by retrieving it from an env variable

### DIFF
--- a/moto/settings.py
+++ b/moto/settings.py
@@ -54,7 +54,7 @@ def get_s3_custom_endpoints():
     return []
 
 
-S3_UPLOAD_PART_MIN_SIZE = os.environ.get("S3_UPLOAD_PART_MIN_SIZE", 5242880)
+S3_UPLOAD_PART_MIN_SIZE = int(os.environ.get("S3_UPLOAD_PART_MIN_SIZE", "5242880"))
 
 
 def get_s3_default_key_buffer_size():
--- a/moto/settings.py
+++ b/moto/settings.py
@@ -54,7 +54,7 @@ def get_s3_custom_endpoints():
     return []
 
 
-S3_UPLOAD_PART_MIN_SIZE = 5242880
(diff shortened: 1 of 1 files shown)


bblommers:
Hi @agilan2001, what's the use-case here? Does this solve a particular problem that you're having?

agilan2001:
> Hi @agilan2001, what's the use-case here? Does this solve a particular problem that you're having?

Hi @bblommers,

As of now, S3_UPLOAD_PART_MIN_SIZE is set to be around 5 MB.
So, writing test cases that check multipart_uploads is not possible without creating test files of size 5 MB.
(Multipart Uploads are rejected if the size of a part [that is not the last one] is less than the S3_UPLOAD_PART_MIN_SIZE)
https://github.com/getmoto/moto/blob/b4346e2eea6d09dfb8723f2a43e13bbe45b34710/moto/s3/models.py#L397 

Making S3_UPLOAD_PART_MIN_SIZE configurable with env variables would allow for writing test cases that checks multipart uploads without actually creating test_files of size 5 MB.

bblommers:
Hi @agilan2001, my question was more: what does the environment variable add? Is it not possible to just override `settings.S3_UPLOAD_PART_MIN_SIZE` instead?

agilan2001:
> Hi @agilan2001, my question was more: what does the environment variable add? Is it not possible to just override `settings.S3_UPLOAD_PART_MIN_SIZE` instead?

Hi @bblommers, 
Making moto's S3_UPLOAD_PART_MIN_SIZE configurable with env variables would allow for setting the configuration with moto-server.
I personally feel this could be highly useful for testing AWS services in Non-Python SDKs through moto-server.

bblommers (approved):
LGTM!

PR #6227, opened 2023-04-18
fix: adjust last modified response dict for s3

satheler:
The AWS `parse_timestamp` only accept these formats:
- iso8601
- rfc822
- epuch (value is an integer)

The format that was returned is `rfc1123` that causes the following error:
```
ValueError: Invalid timestamp "Mon, 05 Apr 2023 14:30:00 EDT": Unknown string format: Mon, 05 Apr 2023 14:30:00 EDT
```

For example, when we have try to `head_object` from s3.

The [`parse_timestamp` botocore implementation](https://github.com/boto/botocore/blob/892c5bc73606ec02a6eff2522da1decff88d1c48/botocore/utils.py#L927).

### DIFF
--- a/moto/s3/models.py
+++ b/moto/s3/models.py
@@ -230,7 +230,7 @@ class FakeKey(BaseModel, ManagedState):
     def response_dict(self):
         res = {
             "ETag": self.etag,
-            "last-modified": self.last_modified_RFC1123,
+            "last-modified": self.last_modified_ISO8601,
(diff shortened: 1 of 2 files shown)


bblommers:
This isn't quite right either @satheler - the format is different depending on the kind of request. See https://github.com/boto/boto/issues/466

I assume that's why the tests fail as well, because TF actually expects a `rfc1123`-date in some cases.

PR #6228, opened 2023-04-18
Adds raw_header to MockRawResponse

akshara08:
## Issue
The primary problem with moto integration with aiobotocore is [this](https://github.com/aio-libs/aiobotocore/blob/72b8dd5d7d4ef2f1a49a0ae0c37b47e5280e2070/aiobotocore/endpoint.py#L23) method. There are 2 main issues with that method
1. It expects `raw` to have `raw_headers` just to convert the headers to lowercase. 
2. Moto passes [MockRawResponse](https://github.com/getmoto/moto/blob/79616e11e64add6b732006952abdef5c5d0914ef/moto/core/botocore_stubber.py#L9) which is not awaitable causing the breaks with lines containing `await`.

## How to reproduce the error
```
# test_response.py

@pytest.fixture
def aws_credentials():


@pytest.fixture
def s3(aws_credentials):
    with mock_s3():
        conn = boto3.resource("s3")
        conn.create_bucket(Bucket='testbucket')
        yield conn


def test_moto(s3):
    for i in range(3):
        s3.Bucket('testbucket').put_object(
            Key=f'glob_{i}.txt', Body=f"test glob file {i}"
        )
    path = 's3://testbucket/glob_*.txt'
    fs = s3fs.S3FileSystem()
    files = fs.glob(path) # Fails 
    assert len(files) == 3
```

## Resolution
It's a two part solution
1. The first part can be a addressed by adding `raw_headers` to MockRawResponse
2. The second part of the problem can be addressed in this repo by only `await`ing for awaitable objects. ([PR](https://github.com/aio-libs/aiobotocore/pull/1007))

I haven't written the tests because [this](https://github.com/aio-libs/aiobotocore/pull/1007) needs to be merged for tests to pass. I was able to test locally and it looks fine

All the moto tests are otherwise passing and I'll add tests once the other PR is merged

### DIFF
--- a/moto/core/botocore_stubber.py
+++ b/moto/core/botocore_stubber.py
@@ -7,7 +7,8 @@ from .responses import TYPE_RESPONSE
 
 
 class MockRawResponse(BytesIO):
-    def __init__(self, response_input: Union[str, bytes]):
+    def __init__(self, response_input: Union[str, bytes], headers: Dict[Any, Any]):
+        self.headers = headers
         if isinstance(response_input, str):
             response_input = response_input.encode("utf-8")
         super().__init__(response_input)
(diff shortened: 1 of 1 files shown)


bblommers:
Hi @akshara08!

I don't want to make changes to Moto _just_ because another open source project needs it for their code to work. That is frankly not our problem.

If you can make an argument that Moto's usage of `AWSResponse` does not follow the AWS spec/behaviour, a change is more likely to be accepted. What kind of class/argument does AWS pass to `AWSResponse`, instead of the `MockRawResponse`? Can we subclass it? Are there any other parameters/methods that Moto should consider?

PR #6230, opened 2023-04-19
fix: Corrects ECRClient.batch_get_image tag handling

zellio:
Currently, `ECRClient.batch_get_image` fails to return an image by any tag other than the one most recently PUT. This appears to be contrary to the functionality of `batch_get_image` against the AWS ECR API.

Code change to correct this issue very small but the PR does include a test case.


bblommers (approved):
LGTM! Thanks for fixing this @zellio

PR #6231, opened 2023-04-19
fix: calculating size of record using decoded data and partition key

lucaxchaves:
Currently, the `put_records` limit validations have the following errors:

- The size of the partition key is not included in the calculation
- The data length is calculated using the base64 encoded data, which is approximately 34% larger
- The total records validation is checking for 5000000 bytes instead of 5*(2**20) bytes
- It is not possible to send a record with exactly 1 MiB
- It is not possible to send 5 MiB worth of records.

ref:
https://docs.aws.amazon.com/streams/latest/dev/service-sizes-and-limits.html

### DIFF
--- a/moto/kinesis/models.py
+++ b/moto/kinesis/models.py
@@ -1,4 +1,4 @@
-from base64 import b64encode
+from base64 import b64encode, b64decode
 from collections import OrderedDict
 from gzip import GzipFile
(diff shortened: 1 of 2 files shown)


bblommers (approved):
Awesome - thanks for fixing this @lucaxchaves!

PR #6236, opened 2023-04-20
Implement ReplaceRoute to NetworkInterfaceId

tjohnes:
This PR adds support for using a NetworkInterfaceId as the target in ReplaceRoute.

Also updated test case `test_route_tables.py::test_routes_replace`.


bblommers (approved):
LGTM - thank you for adding this feature @tjohnes, and welcome to Moto!

PR #6242, opened 2023-04-21
sns: numeric filtering fixups + parameter validation

dezeroku:
During work on https://github.com/getmoto/moto/issues/6087 I've run the `tests/test_sns/test_publishing_boto3.py` tests against the "real" AWS (after some modifications) and I've found out that there are issues with numeric filtering.
It's mostly about the types used.

Changes implemented:
1. [SNS filter policy using "numeric" matching](https://docs.aws.amazon.com/sns/latest/dg/numeric-value-matching.html) checks requires numbers (as in int/float) to be provided in the check, not strings
2. Message attributes of type "Number" should be provided (and returned in responses) as strings. They've been ints/floats till now
3. General parameter validation for numeric matching, although limited. The errors from AWS contain the exact location where the parsing error occurs and doingit the same way here, while doable, is too much effort IMHO. If someone wants to improve on this feel free to do so
4. rounding tests rewritten from using `100.1234561` to `100.1234567`. The latter works as expected with AWS.

Both 1 and 2 are "backwards incompatible", but this shouldn't really matter, as that's how it works in the AWS anyway.

### DIFF
--- a/moto/sns/models.py
+++ b/moto/sns/models.py
@@ -836,7 +836,6 @@ class SNSBackend(BaseBackend):
                         # Example: 'Value of < must be numeric\n at [Source: (String)"{"price":[{"numeric":["<","100"]}]}"; line: 1, column: 28]'
                         # While it probably can be implemented, it doesn't feel as important as the general parameter checking
 
-
                         attributes_copy = attributes[:]
                         if not attributes_copy:
                             raise SNSInvalidParameter(
--- a/moto/sns/models.py
+++ b/moto/sns/models.py
@@ -302,6 +302,7 @@ class Subscription(BaseModel):
                         return False
 
                     for attribute_values in attribute_values:
+                        attribute_values = float(attribute_values)
                         # Even the official documentation states a 5 digits of accuracy after the decimal point for numerics, in reality it is 6
                         # https://docs.aws.amazon.com/sns/latest/dg/sns-subscription-filter-policies.html#subscription-filter-policy-constraints
                         if int(attribute_values * 1000000) == int(rule * 1000000):
@@ -347,7 +348,7 @@ class Subscription(BaseModel):
                                 [value] if isinstance(value, str) else value
                             )
                             if attr["Type"] == "Number":
-                                actual_values = [str(attr["Value"])]
+                                actual_values = [float(attr["Value"])]
                             elif attr["Type"] == "String":
                                 actual_values = [attr["Value"]]
                             else:
@@ -361,7 +362,7 @@ class Subscription(BaseModel):
                             message_attributes.get(field, {}).get("Type", "")
                             == "Number"
                         ):
-                            msg_value = message_attributes[field]["Value"]
+                            msg_value = float(message_attributes[field]["Value"])
                             matches = []
                             for operator, test_value in numeric_ranges:
                                 test_value = int(test_value)
--- a/moto/sns/models.py
+++ b/moto/sns/models.py
@@ -836,7 +836,6 @@ class SNSBackend(BaseBackend):
                         # Example: 'Value of < must be numeric\n at [Source: (String)"{"price":[{"numeric":["<","100"]}]}"; line: 1, column: 28]'
                         # While it probably can be implemented, it doesn't feel as important as the general parameter checking
 
-                        from xml.sax.saxutils import escape
 
                         attributes_copy = attributes[:]
                         if not attributes_copy:
@@ -848,12 +847,12 @@ class SNSBackend(BaseBackend):
 
                         if not isinstance(operator, str):
                             raise SNSInvalidParameter(
-                                f"Invalid parameter: Attributes Reason: FilterPolicy: Invalid member in numeric match: {escape(str(operator))}\n at ..."
+                                f"Invalid parameter: Attributes Reason: FilterPolicy: Invalid member in numeric match: {(str(operator))}\n at ..."
                             )
 
                         if operator not in ("<", "<=", "=", ">", ">="):
                             raise SNSInvalidParameter(
-                                f"Invalid parameter: Attributes Reason: FilterPolicy: Unrecognized numeric range operator: {escape(str(operator))}\n at ..."
+                                f"Invalid parameter: Attributes Reason: FilterPolicy: Unrecognized numeric range operator: {(str(operator))}\n at ..."
                             )
 
                         try:
@@ -863,7 +862,7 @@ class SNSBackend(BaseBackend):
 
                         if value is None or not isinstance(value, (int, float)):
                             raise SNSInvalidParameter(
-                                f"Invalid parameter: Attributes Reason: FilterPolicy: Value of {escape(str(operator))} must be numeric\n at ..."
+                                f"Invalid parameter: Attributes Reason: FilterPolicy: Value of {(str(operator))} must be numeric\n at ..."
                             )
 
                         if not attributes_copy:
@@ -878,7 +877,7 @@ class SNSBackend(BaseBackend):
 
                         if second_operator not in ("<", "<="):
                             raise SNSInvalidParameter(
-                                f"Invalid parameter: Attributes Reason: FilterPolicy: Bad numeric range operator: {escape(str(second_operator))}\n at ..."
+                                f"Invalid parameter: Attributes Reason: FilterPolicy: Bad numeric range operator: {(str(second_operator))}\n at ..."
                             )
 
                         try:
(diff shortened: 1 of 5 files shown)


moto/sns/models.py
```diff
@@ -832,6 +832,72 @@ def _validate_filter_policy(self, value):
                             )
                         continue
                     elif keyword == "numeric":
+                        # TODO: All of the exceptions listed below contain column pointing where the error is (in AWS response)
+                        # Example: 'Value of < must be numeric\n at [Source: (String)"{"price":[{"numeric":["<","100"]}]}"; line: 1, column: 28]'
+                        # While it probably can be implemented, it doesn't feel as important as the general parameter checking
+
+                        from xml.sax.saxutils import escape
```
dezeroku, line 839: This function is used to escape XML (`<`, `>`, etc. characters) in the exception thrown, but I feel like it's not the best place to do it. Is there a better way to put it so it works globally?

Without this in place, boto client errors out on parsing the response, due to the forbidden characters.

bblommers: The `RESTError`-class is the ancestor for all exceptions, so that's probably the best choice: 
https://github.com/getmoto/moto/blob/462ff57900abc0a8133a26487052273f9fc4579f/moto/core/exceptions.py#L58

moto/sns/responses.py
```diff
@@ -78,6 +78,7 @@ def _transform_message_attributes(self, message_attributes):
                                 "when calling the Publish operation: "
                                 f"Could not cast message attribute '{name}' value to number."
                             )
+                    transform_value = str(transform_value)
```
bblommers, line 81: If we don't re-assign `transform_value` in the first place, I'm assuming this isn't necessary.

I.e.: instead of doing `transform_value = int(value["StringValue"])`, we can just try to cast (`int(value["StringValue"])`) without assigning it to `transform_value`. That way, we'll keep the original value as a string.

moto/sns/models.py
```diff
@@ -302,6 +302,7 @@ def _field_match(field, rules, message_attributes):
                         return False
 
                     for attribute_values in attribute_values:
+                        attribute_values = float(attribute_values)
```
bblommers, line 305: This for-loop should probably say `for attribute_value in attribute_values`. It's currently re-using the same variable-name, that's never a good idea.

(I know it's not your change, but I suddenly noticed it, and we may as well fix it now.)

tests/test_sns/test_subscriptions_boto3.py
```diff
@@ -477,9 +481,136 @@ def test_subscribe_invalid_filter_policy():
...
+            Attributes={
+                "FilterPolicy": json.dumps({"price": [{"numeric": ["<", "100"]}]})
+            },
+        )
+        assert False
+    except ClientError as err:
+        err.response["Error"]["Code"].should.equal("InvalidParameter")
+        err.response["Error"]["Message"].should.equal(
+            "Invalid parameter: Attributes Reason: FilterPolicy: Value of < must be numeric\n at ..."
+        )
```
bblommers, line 502:
```suggestion
    with pytest.raises(ClientError) as err:
        conn.subscribe(
            TopicArn=topic_arn,
            Protocol="http",
            Endpoint="http://example.com/",
            Attributes={
                "FilterPolicy": json.dumps({"price": [{"numeric": ["<", "100"]}]})
            },
        )

    err.response["Error"]["Code"].should.equal("InvalidParameter")
    err.response["Error"]["Message"].should.equal(
        "Invalid parameter: Attributes Reason: FilterPolicy: Value of < must be numeric\n at ..."
    )
```
This is better readable, IMO. This way we don't need the `assert False` - pytest will automatically fail if the expected error isn't raised.

(Just to be clear, you don't have to change the existing tests in this PR! As long as the new tests follow the correct approach, we can always fix the others later.)

bblommers (changes requested):

tests/test_sns/test_subscriptions_boto3.py
dezeroku: Well, that's what you get for writing tests at night :D I completely missed that you can actually get the exception instance thrown with `pytest.raises()`. Fixing this up

moto/sns/models.py
dezeroku: I am not sure if we understood each other. The issue here is IMHO not that the exception is raised on moto's end (so we can catch the generic `RestError`), but rather that moto happily returns a message that can't be parsed by the boto client.

So it seems like there's nothing on the moto's end making sure that content is escaped before being sent out and it can't really be catched with an exception. I am not completely sure how the moto/boto interaction works though, so I might be wrong.

If it's OK for you then keeping the sanitizing on the local level should be OK, most of the responses won't ever contain problematic characters.

Here's the trace if the `escape`s are removed:
```

>           conn.subscribe(
                TopicArn=topic_arn,
                Protocol="http",
                Endpoint="http://example.com/",
                Attributes={
                    "FilterPolicy": json.dumps({"price": [{"numeric": ["<", "100"]}]})
                },
            )

tests/test_sns/test_subscriptions_boto3.py:484: 
.venv/lib/python3.10/site-packages/botocore/client.py:530: in _api_call
.venv/lib/python3.10/site-packages/botocore/client.py:943: in _make_api_call
    http, parsed_response = self._make_request(
.venv/lib/python3.10/site-packages/botocore/client.py:966: in _make_request
    return self._endpoint.make_request(operation_model, request_dict)
.venv/lib/python3.10/site-packages/botocore/endpoint.py:119: in make_request
    return self._send_request(request_dict, operation_model)
.venv/lib/python3.10/site-packages/botocore/endpoint.py:199: in _send_request
    success_response, exception = self._get_response(
.venv/lib/python3.10/site-packages/botocore/endpoint.py:241: in _get_response
    success_response, exception = self._do_get_response(
.venv/lib/python3.10/site-packages/botocore/endpoint.py:308: in _do_get_response
    parsed_response = parser.parse(
.venv/lib/python3.10/site-packages/botocore/parsers.py:250: in parse
    parsed = self._do_error_parse(response, shape)
.venv/lib/python3.10/site-packages/botocore/parsers.py:557: in _do_error_parse
    root = self._parse_xml_string_to_dom(xml_contents)

self = <botocore.parsers.QueryParser object at 0x7f0c29d9d9f0>
xml_string = b'<?xml version="1.0" encoding="UTF-8"?>\n<ErrorResponse>\n    <Error>\n        <Code>InvalidParameter</Code>\n       ...Message>\n        \n        <RequestId>7a62c49f-347e-4fc4-9331-6e8eEXAMPLE</RequestId>\n    </Error>\n</ErrorResponse>'

    def _parse_xml_string_to_dom(self, xml_string):
        try:
            parser = ETree.XMLParser(
                target=ETree.TreeBuilder(), encoding=self.DEFAULT_ENCODING
            )
            parser.feed(xml_string)
            root = parser.close()
        except XMLParseError as e:
>           raise ResponseParserError(
                "Unable to parse response (%s), "
                "invalid XML received. Further retries may succeed:\n%s"
                % (e, xml_string)
            )
E           botocore.parsers.ResponseParserError: Unable to parse response (not well-formed (invalid token): line 5, column 79), invalid XML received. Further retries may succeed:
E           b'<?xml version="1.0" encoding="UTF-8"?>\n<ErrorResponse>\n    <Error>\n        <Code>InvalidParameter</Code>\n        <Message>Invalid parameter: Attributes Reason: FilterPolicy: Value of < must be numeric\n at ...</Message>\n        \n        <RequestId>7a62c49f-347e-4fc4-9331-6e8eEXAMPLE</RequestId>\n    </Error>\n</ErrorResponse>'

.venv/lib/python3.10/site-packages/botocore/parsers.py:506: ResponseParserError
```

bblommers: I was talking about a Moto-class called `RESTError`, that contains all the logic for exceptions to be properly formatted before being send to boto

tests/test_sns/test_subscriptions_boto3.py
```diff
@@ -425,61 +425,179 @@ def test_subscribe_invalid_filter_policy():
                 "FilterPolicy": json.dumps({"store": [str(i) for i in range(151)]})
             },
         )
-    except ClientError as err:
+
         err.response["Error"]["Code"].should.equal("InvalidParameter")
```
bblommers, line 429: These lines should be un-indented one level. They are currently inside the `with pytest.raises`-block, so when the previous call fails, these lines are never executed.

moto/sns/models.py
dezeroku: Ok, I got it now, thanks. I'll move the logic there

dezeroku: Ended up adapting the logic from https://github.com/getmoto/moto/commit/579b6e264ff297d1787706848f005b5d707bb794 for all core error templates to avoid double-escaping (as would be the case with SES after original changes).
CDATA seems a lot cleaner anyway

tests/test_sns/test_subscriptions_boto3.py
dezeroku: Thanks, it's fixed now.

bblommers (approved):
Awesome - thank you for contributing to Moto @dezeroku!

PR #6245, opened 2023-04-22
S3: add a checksum to a copied object if requested so

rafcio19:
Resolves https://github.com/getmoto/moto/issues/6234


bblommers (approved):
Nice one - thanks @rafcio19!

PR #6253, opened 2023-04-24
[Glue] Triggers

szemek:
Implemented:

- batch_get_triggers
- create_trigger
- delete_trigger
- get_trigger
- get_triggers
- list_triggers
- start_trigger
- stop_trigger

I'll try to implement `update_trigger` and few other `update_*` methods in the next pull request.


moto/glue/models.py
```diff
@@ -757,6 +771,79 @@ def update_schema(
...
+
+        return list(self.triggers.values())
+
+    @paginate(pagination_model=PAGINATION_MODEL)
+    def list_triggers(self, dependent_job_name: str) -> List["FakeTrigger"]:  # type: ignore
+        if dependent_job_name:
+            triggers = []
+            for trigger in self.triggers.values():
+                for action in trigger.actions:
+                    if ("JobName" in action) and (
```
bblommers, line 835: Would you mind adding a test for this? It looks OK, but it's always good to explicitly verify that passing the argument works as expected

moto/glue/responses.py
```diff
@@ -415,6 +415,17 @@ def filter_jobs_by_tags(
             return [job.get_name() for job in jobs]
         return [job.get_name() for job in jobs if self.is_tags_match(job.arn, tags)]
 
+    def filter_triggers_by_tags(
+        self, triggers: List[FakeJob], tags: Dict[str, str]
+    ) -> List[str]:
+        if not tags:
+            return [trigger.get_name() for trigger in triggers]
+        return [
```
bblommers, line 423: Same here, would be good to have a test here as well

bblommers (changes requested):

moto/glue/models.py
szemek: @bblommers thanks for pointing that out 👌  I added tests in the next commit.

bblommers (approved):
LGTM - thank you for adding this @szemek!

PR #6259, opened 2023-04-26
SES v2

rafcio19:
Relates to https://github.com/getmoto/moto/issues/5076

Hey @bblommers this is my first attempt at a service from scratch, and a hybrid of using existing backend too. Let me know if this general direction of things is OK and I will spruce this PR up and run the implementation coverage script etc


moto/sesv2/responses.py
```diff
@@ -0,0 +1,44 @@
...
+    def send_email(self):
+        """Piggy back on functionality from v1 mostly"""
+
+        # parsing of these params is nasty, hopefully there is a tidier way
+        params = json.loads(list(dict(self.querystring.items()).keys())[0])
+        from_email_address = params.get("FromEmailAddress")
+        destination = params.get("Destination")
+        content = params.get("Content")
+
+        message = self.sesv1_backend.send_email(
```
bblommers, line 35: This should really call the `sesv2_backend.send_email` - in there, we can then call `sesv1.backend`.
That way we have a clean separation of IO (in responses) and logic (in models). It also ensures that the method is picked up in the implementation coverage script.

moto/sesv2/urls.py
```diff
@@ -0,0 +1,14 @@
+"""sesv2 base URL and path."""
+from .responses import SESV2Response
+
+url_bases = [
+    r"https?://email\.(.+)\.amazonaws\.com",
```
bblommers, line 5: It looks like this endpoint is the same as sesv1, which means it will be difficult for Moto to figure out which service the request was made to.
From what I can tell, the only difference between them is the requestUri: 

https://github.com/boto/botocore/blob/e3375193deedb139045a19ccf821b30c2a100098/botocore/data/sesv2/2019-09-27/service-2.json#L1186

So we may have to do something similar to what we did for S3Control, which had the same issue:

https://github.com/getmoto/moto/blob/37f14567472b3b40d44b20cbba1cbd631d954f5a/moto/moto_server/werkzeug_app.py#L152

Let me know if you want me to look into this further

bblommers (changes requested):

bblommers:
I realize I didn't actually answer your question - this is indeed in the right direction @rafcio19! On a high-level, this should be everything necessary to add a new service.

Just a general note: I've just introduced typing annotations just now for all services starting with S*, so this will need the annotations as well. After merging in master you can just run `mypy` to see if this check passes.
(PR for reference: https://github.com/getmoto/moto/pull/6261)

moto/sesv2/urls.py
rafcio19: Hmm.. I will need a hand with making `test_server` pass. I added an endpoint for [ListContacts](https://docs.aws.amazon.com/ses/latest/APIReference-V2/API_ListContacts.html) and I thought this will make moto server happy but not so:

```
pytest tests/test_sesv2/test_server.py                                                                                                    <aws:raf>
================================================================================ test session starts ================================================================================
platform darwin -- Python 3.9.13, pytest-7.1.3, pluggy-1.0.0
rootdir: /Users/raf/projects/moto, configfile: setup.cfg
plugins: anyio-3.6.1, mock-3.10.0, xdist-3.2.1, ordering-0.6, cov-4.0.0
collected 1 item                                                                                                                                                                    

tests/test_sesv2/test_server.py F                                                                                                                                             [100%]

===================================================================================== FAILURES ======================================================================================
__________________________________________________________________________________ test_sesv2_list __________________________________________________________________________________

    def test_sesv2_list():
        backend = server.create_backend_app("sesv2")
        test_client = backend.test_client()
    
        resp = test_client.post("/v2/email/contact-lists/test/contacts")
    
>       assert resp.status_code == 200
E       assert 404 == 200
E        +  where 404 = <WrapperTestResponse streamed [404 NOT FOUND]>.status_code

tests/test_sesv2/test_server.py:12: AssertionError
============================================================================== short test summary info ==============================================================================
FAILED tests/test_sesv2/test_server.py::test_sesv2_list - assert 404 == 200
  
```

bblommers: Ah, I don't think we need `test_server.py` actually. It's great if we want to test the raw HTTP responses, but that's not necessary in this case.

Just need to make sure that the regular boto3-tests pass in ServerMode. This was the last failure: 
https://github.com/getmoto/moto/actions/runs/4819031481/jobs/8582608793

Docs on how to run these tests locally: http://docs.getmoto.org/en/latest/docs/contributing/development_tips/tests.html#servermode-tests

rafcio19: It looks like the two(my test_server test and ServerMode tests) are definitely related, they both fail for the same reason -  a 404 from the flask app 🤔 

```
tests/test_sesv2/test_sesv2.py::test_list_contacts 127.0.0.1 - - [27/Apr/2023 15:28:16] "POST /moto-api/reset HTTP/1.1" 200 -
127.0.0.1 - - [27/Apr/2023 15:28:16] "GET /v2/email/contact-lists/test/contacts HTTP/1.1" 404 -
FAILED
```

moto/sesv2/urls.py
```diff
@@ -0,0 +1,14 @@
...
+]
+
+
+response = SESV2Response()
+
+
+url_paths = {
+    "{0}/v2/.*$": response.dispatch,
```
bblommers, line 13:
```suggestion
    "{0}/v2/email/outbound-emails$": response.dispatch,
    "{0}/v2/email/contact-lists/(?P<name>[^/]+)/contacts$": response.dispatch,
```
The URL's need to be specified explicitly, then it will work

tests/test_sesv2/test_server.py
```diff
@@ -0,0 +1,12 @@
+"""Test different server responses."""
+
+import moto.server as server
+
+
+def test_sesv2_list():
+    backend = server.create_backend_app("sesv2")
+    test_client = backend.test_client()
+
+    resp = test_client.post("/v2/email/contact-lists/test/contacts")
```
bblommers, line 10:
```suggestion
    resp = test_client.get("/v2/email/contact-lists/test/contacts")
```
The POST-method resolves to `create_contact`, which isn't implemented yet. The GET-method will resolve to `list_contacts`.

Note that this particular test still failed for me due to an bug in Moto itself - but it should pass after #6266 is merged.

moto/__init__.py
```diff
@@ -171,6 +172,7 @@ def f(*args, **kwargs):
 mock_xray_client = lazy_load(".xray", "mock_xray_client")
 mock_wafv2 = lazy_load(".wafv2", "mock_wafv2")
 mock_textract = lazy_load(".textract", "mock_textract")
+mock_sesv2 = lazy_load(".sesv2", "mock_sesv2", boto3_name="sesv2")
```
bblommers, line 175: This was added twice

tests/test_sesv2/test_sesv2.py
```diff
@@ -0,0 +1,83 @@
...
+                "Subject": {"Data": "test subject"},
+                "Body": {"Text": {"Data": "test body"}},
+            },
+        },
+    )
+
+    # Execute
+    with pytest.raises(ClientError) as e:
+        conn.send_email(**kwargs)
+        assert e
```
bblommers, line 37:
```suggestion
    with pytest.raises(ClientError) as e:
        conn.send_email(**kwargs)
    assert e.response.value["Error"]["Code"] == "MessageRejected"
```
This asserts that the correct error is raised. Note the indentation - after `conn.send_email` fails, nothing else is executed. So asserting that the error is correct should happen outside this context.

bblommers (changes requested):
LGTM like this @rafcio19 - just got a few minor nitpicks.

Did you want to add any other methods?

rafcio19:
Thanks for your help, much appreciated. It would've taken me waaaaay longer to sort this out myself 🚀 
I still want to add a method or 2 more and need to sort out a bunch of mypy stuff. I think this will be ready for a final review within a few days

rafcio19:
@bblommers this is ready for a review. 

N.B. I run implementation_coverage script and and it [trimmed](https://github.com/getmoto/moto/pull/6259/commits/c6c3a11efb71a644c55d2bd69691b78ad4a4220a) quite a lot of stuff I didn't expect. Should I drop this commit?

bblommers:
@rafcio19 The full list of methods is supplied by botocore/boto3, so a diff like this usually happens if it is run against an older version that doesn't support the latest features.

Would you mind updating them both, and just rerunning the script? I'd expect some additions for other features that the latest boto3 supports, but no removals.

`pip install --upgrade botocore boto3`

rafcio19:
@bblommers looks better after upgrade, thanks. I also just noticed that none of the create/delete/list `_contact` methods are getting pulled into the documentation and it looks like that's because I moved these methods out of `SESV2Backend` model and into `ContactList` object(because `Contact` is a child items of `ContactList`) 🤔 
I will refactor to so the backend calls the appropriate method on `ContactList` object and that should be final 🤞

bblommers (approved):
Awesome - thank you so much for adding this to Moto @rafcio19!

PR #6262, opened 2023-04-27
sns: support FilterPolicyScope attribute, including filtering

dezeroku:
This resolves: https://github.com/getmoto/moto/issues/6087
[AWS blog article about the feature](https://aws.amazon.com/blogs/compute/introducing-payload-based-message-filtering-for-amazon-sns/)

## What's added?
1. Recognition of the [`FilterPolicyScope` attribute](https://docs.aws.amazon.com/sns/latest/dg/sns-message-filtering.html#sns-message-filtering-scope) for an SNS subscription
2. Validation of the [filter policy's complexity](https://docs.aws.amazon.com/sns/latest/dg/subscription-filter-policy-constraints.html)
3. Actual logic for filtering the messages based on `FilterPolicyScope` that's set

Both 1 and 2 were pretty much taken as-is from https://github.com/localstack/moto/pull/62.

## General idea
The biggest change is in the `_matches_filter_policy` function, I've created subfunctions for checking the actual conditions (to be shared between both `FilterPolicyScope` variants). For attributes based filtering everything stayed roughly the same, for payload based filtering:
1. A nested list of simplified checks is generated, so nested traversing is not required in the "actual checker" function
2. The checks are run separately, results are concatenated so either one of the checks in a list has to pass or all of them

## Testing
Most of the tests for `MessageBody` variant were copy-pasted from the `MessageAttributes` variant (the ones that were already implemented) and modified accordingly. This is done as a sanity check for the filtering.

AWS docs don't provide many examples that utilize the most interesting part of `MessageBody` filtering, that is nesting.
Few tests were adapted from [AWS blog article](https://aws.amazon.com/blogs/compute/introducing-payload-based-message-filtering-for-amazon-sns/) and some were crafted during the development to clarify some doubts.

All of the tests have been run (with minor modifications required) against AWS and were confirmed valid this way.

## Concerns
1. I wanted to keep the "layout" of everything mostly the same, e.g. all the logic is contained within the original `matches_filter_policy` function. It looks crammed now, do you think separating it to its own class/file would be OK?
2. I feel like general cleanup for this is needed, but I don't really have an idea how to simplify it without overengineering. Any comments would be appreciated

### DIFF
--- a/moto/sns/models.py
+++ b/moto/sns/models.py
@@ -4,7 +4,7 @@ import requests
 import re
 
 from collections import OrderedDict
-from typing import Any, Dict, List, Iterable, Optional, Tuple
+from typing import Any, Dict, List, Iterable, Optional, Tuple, Union, Callable
 
 from moto.core import BaseBackend, BackendDict, BaseModel, CloudFormationModel
 from moto.core.utils import (
@@ -203,7 +203,7 @@ class Subscription(BaseModel):
         group_id: Optional[str] = None,
         deduplication_id: Optional[str] = None,
     ) -> None:
-        if not self._matches_filter_policy(message_attributes):
+        if not self._matches_filter_policy(message_attributes, message):
             return
 
         if self.protocol == "sqs":
@@ -278,7 +278,7 @@ class Subscription(BaseModel):
             )
 
     def _matches_filter_policy(
-        self, message_attributes: Optional[Dict[str, Any]]
+        self, message_attributes: Optional[Dict[str, Any]], message: str
     ) -> bool:
         if not self._filter_policy:
             return True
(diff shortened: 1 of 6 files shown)


bblommers:
This is very comprehensive @dezeroku! Nicely done. It is quite understandable, thanks to the comments, so I don't know whether a clean-up/simplification is necessary.

I do agree with your first concern - it would be better for overall readability if it could be extracted in it's own file/class.
That would have the additional benefit that we can test that class directly. Boto3 adds quite a bit of overhead, so being able to invoke the logic directly (without creating a topic, queue, setting the attributes, etc.) would also improve the performance and readability of the tests.

bblommers (approved):
LGTM! Thank you for adding this feature to Moto @dezeroku, much appreciated!

bblommers:
Hi @dezeroku,

https://payments.getmoto.org/


Feel free to open a bug or discussion if you run into any problems:

Issue #6263, opened 2023-04-27
S3 CopyObject differences with AWS


bentsku:
Hi!

I've been investigating differences between AWS and moto regarding `CopyObject`, and did extensive testing to try to validate the behaviour as closely as possible.

- It seems AWS does not copy the encryption settings from the source key. Affected fields: `ServerSideEncryption`, `BucketKeyEnabled` and `SSEKMSKeyId`.
- AWS does not keep any informations regarding Object Locking from the source key. Affected fields: `ObjectLockMode`, `ObjectLockRetainUntilDate` and `ObjectLockLegalHoldStatus`. 
- AWS does not keep the storage class of the source key.
- AWS does not keep the ACL from the source key (another issue will be opened for ACL "inheritance" which does not happen in AWS). 
- AWS needs less fields to be specified to allow an "in-place" copy: only `ServerSideEncryption`, `MetadataDirective` (set to `REPLACE`), `StorageClass` and `WebsiteRedirectLocation`. Even when set, those do not need to be different than the source key's one to be accepted. 
- When copying from a versioned bucket, moto does not return  the `x-amz-copy-source-version-id` header.

I'm going to submit a PR to fix those issues. The added tests in the PR should allow reproduction of the above issues.

PR #6264, opened 2023-04-27
update CopyObject logic

bentsku:
Hello!
This PR should fix the issues mentioned in #6263

I tried to be as exhaustive as possible in the tests to allow reproduction of the current issues. I've also extensively tested those against AWS.

This would fix the following:
- Don't use the following from the source key:
  - `ServerSideEncryption`
  - `BucketKeyEnabled`
  - `SSEKMSKeyId`
  - `StorageClass`
  - `ACL`
  - `ObjectLockMode`
  - `ObjectLockRetainUntilDate`
  - `ObjectLockLegalHoldStatus`
- Use and set those parameters to the destination key:
  - `ObjectLockMode`
  - `ObjectLockRetainUntilDate`
  - `ObjectLockLegalHoldStatus`
- Rework the check to allow copying "in-place"
- Return `x-amz-copy-source-version-id` if the source bucket/key was versioned

If anything is missing, feel free and I'll try to fix it asap!

### DIFF
--- a/moto/s3/models.py
+++ b/moto/s3/models.py
@@ -2363,23 +2363,27 @@ class S3Backend(BaseBackend, CloudWatchMetricProvider):
         kms_key_id: Optional[str] = None,
         bucket_key_enabled: Any = None,
         mdirective: Optional[str] = None,
-        metadata:Optional[Any] = None,
+        metadata: Optional[Any] = None,
         website_redirect_location: Optional[str] = None,
         lock_mode: Optional[str] = None,
         lock_legal_status: Optional[str] = None,
         lock_until: Optional[str] = None,
     ) -> None:
-        if (
-                src_key.name == dest_key_name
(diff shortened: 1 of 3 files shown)


bblommers:
Thanks for the PR @bentsku! I'm not quite sure why that particular test fails. It may be because of a recent werkzeug upgrade - you may have to merge in the latest master, I pushed a coule of fixes recently

bentsku:
Oh I see my mistake now, I'm passing a `Bucket` resource to the client `Bucket` parameter... Will fix it quickly.

bentsku:
Finally fixed the test again, pushing commits on Friday evening without testing was not a good idea. It should finally be good to go. Sorry again for having to run the pipeline, really sorry!

bblommers:
No worries @bentsku - the CI is free :)

LGTM - thanks again for the PR!

bblommers (approved):

PR #6269, opened 2023-04-28
Scheduler: Fix update_schedule raising ResourceNotFound errors when using schedule groups

atorrescogollo:
The `scheduler:UpdateSchedule` method is broken when using schedule groups since it cannot find the targeted schedule:

```
botocore.errorfactory.ResourceNotFoundException: An error occurred (ResourceNotFoundException) when calling the UpdateSchedule operation: Schedule not found
```

The issue is that the retrieved parameter is `groupName` whereas the right parameter for the UpdateSchedule seems to be `GroupName`.

Also, added a new test for this use case.

### DIFF
--- a/moto/scheduler/responses.py
+++ b/moto/scheduler/responses.py
@@ -60,7 +60,7 @@ class EventBridgeSchedulerResponse(BaseResponse):
         return "{}"
 
     def update_schedule(self) -> str:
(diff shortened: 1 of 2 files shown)


bblommers (approved):
Awesome - thank you for fixing this @atorrescogollo!

PR #6278, opened 2023-05-02
Fix default account id in docs

svdimchenko:
Problem statement: default account id in docs does not correspond with DEFAULT_ACCOUNT_ID in moto model.


bblommers (approved):
Nice catch @svdimchenko - thanks for fixing this!

Issue #6286, opened 2023-05-04
Creating of SQS queues with latest botorecore (1.29.127) is broken


PierreKiwi:
Hello moto team,

Looks like moto is broken with the latest botocore (1.29.127) as AWS has changed the format for interacting with some services (includig SQS). When creating a SQS queue, it returns {} rather than a Queue object (I use a SQS resource).

I have opened an issue in their tracker (see https://github.com/boto/botocore/issues/2930) but it looks like moto might need to make some changes too.

You can reproduce the issue with:

```python
import moto
import boto3
from uuid import uuid4

@moto.mock_sqs
def test_create_queue():
    sqs = boto3.resource("sqs", region_name="us-east-1")

    q_name = str(uuid4())[0:6]
    new_queue = sqs.create_queue(QueueName=q_name)

    print(new_queue)

test_create_queue()
```

With pip install boto3==1.26.126 botocore==1.29.126 moto, the output is `sqs.Queue(url='https://sqs.us-east-1.amazonaws.com/123456789012/a94c6c')`.

With pip install boto3==1.26.127 botocore==1.29.127 moto, the output is `{}`.

kellertk:
> AWS has changed the format for interacting with some services

Just wanted to clarify on this point. AWS has only changed the wire format for SQS. Changing the wire protocol for a service is a somewhat large undertaking and we don't do it often. The SQS service itself supports both the old and new API formats so that old SDK clients will continue to work, but new versions of boto* will only ever talk to SQS in the new format.

JayThomason:
We were just bit by this as well and I used an almost identical test script haha.

My testing seemed to to indicate that boto3 >= 1.26.0 does not work with any versions of moto, while boto3 1.25.X did work, although maybe the issue was in botocore -- I didn't check the version of that.

This is pretty easy to replicate by spinning up a python:3.11 docker container, installing moto and boto3, running the test script, and then switching between different versions of boto3.

SamStephens:
@JayThomason it's botocore - what you're seeing is that boto3 1.26.x will use any available version of botocore that is > 1.29.x and < 1.30.0, whereas boto3 1.15.x will use any available version of botocore that is > 1.28.x and < 1.29.0.

So any install of boto3 1.26.x will pick up botocore==1.29.127 (or later) which has the breaking change, whereas any install of boto3 1.25.x will pick up botocore==1.18.5 and work.

[...]

PR #6287, opened 2023-05-05
Do not unquote the key two times on multipart upload complete

vdytyniak-exos:
Current implementation unquotes key two times, firstly in `self.backend.put_object` and then in `self.backend.put_object_acl` -> `self.get_object`. To fix this issue I have exposed `key_is_clean` in `put_object_acl` and set it to `True`.

### DIFF
--- a/tests/test_s3/test_s3_multipart.py
+++ b/tests/test_s3/test_s3_multipart.py
@@ -150,6 +150,47 @@ def test_multipart_upload():
 
 @mock_s3
 @reduced_min_part_size
+def test_multipart_upload_with_encoded_char():
+    s3 = boto3.resource("s3", region_name=DEFAULT_REGION_NAME)
+    client = boto3.client("s3", region_name=DEFAULT_REGION_NAME)
+    s3.create_bucket(Bucket="foobar")
+
+    part1 = b"0" * REDUCED_PART_SIZE
+    part2 = b"1"
+    mp = client.create_multipart_upload(Bucket="foobar", Key="the%20key")
+    up1 = client.upload_part(
+        Body=BytesIO(part1),
+        PartNumber=1,
+        Bucket="foobar",
+        Key="the%20key",
+        UploadId=mp["UploadId"],
+    )
+    up2 = client.upload_part(
+        Body=BytesIO(part2),
+        PartNumber=2,
+        Bucket="foobar",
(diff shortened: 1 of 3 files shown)


vdytyniak-exos:
> Current implementation unquotes key two times, firstly in `self.backend.put_object` and then in `self.backend.put_object_acl` -> `self.get_object`. To fix this issue I have exposed `key_is_clean` in `put_object_acl` and set it to `True`.

vdytyniak-exos:
@bblommers I see that were trying to rerun the check. Do you think it is related to my changes?

tests/test_s3/test_s3_multipart.py
```diff
@@ -148,6 +148,47 @@ def test_multipart_upload():
     response["Body"].read().should.equal(part1 + part2)
 
 
+@mock_s3
+@reduced_min_part_size
+def test_multipart_upload_with_encoded_char():
```
bblommers, line 153: This is the same test as `test_multipart_upload`, right? 
I think it's cleaner to just parametrize that test, using `pytest.mark.parametrize`

There are other tests in this file using the same fixture, so you can use those as a guide. Let me know if you run into any issues.

bblommers (changes requested):
Hi @vdytyniak-exos, appreciate you taking the time to submit a fix for this issue!

The tests are failing because of #6286, so don't worry too much about unrelated tests.

tests/test_s3/test_s3_multipart.py
vdytyniak-exos: Makes sense. Done.

vdytyniak-exos:
> Hi @vdytyniak-exos, appreciate you taking the time to submit a fix for this issue!
> 
> The tests are failing because of #6286, so don't worry too much about unrelated tests.

Could you please approve the changes?

bblommers (approved):
LGTM - thanks again for fixing this @vdytyniak-exos!

vdytyniak-exos:
> LGTM - thanks again for fixing this @vdytyniak-exos!

@bblommers Glad to fix! When I can expect docker image to be pushed with my fix?

bblommers:
That happens automatically @vdytyniak-exos ! It's already part of the `latest`-tag. I also released `4.1.9` yesterday if you prefer to use a fixed release.

PR #6289, opened 2023-05-06
ECS: allow ecs service to transition to arbitrary running task count

rafcio19:
Closes https://github.com/getmoto/moto/issues/6283

### DIFF
--- a/moto/ecs/models.py
+++ b/moto/ecs/models.py
@@ -2,6 +2,7 @@ import re
 from copy import copy
 from datetime import datetime, timezone
 from typing import Any, Dict, Iterator, List, Optional, Tuple
+from os import getenv
 
 from moto import settings
 from moto.core import BaseBackend, BackendDict, BaseModel, CloudFormationModel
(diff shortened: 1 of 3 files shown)


rafcio19:
@bblommers ready for review

moto/ecs/models.py
```diff
@@ -507,14 +516,31 @@ def __init__(
...
         self.events: List[Dict[str, Any]] = []
         self.launch_type = launch_type
         self.service_registries = service_registries or []
+        self.load_balancers = load_balancers if load_balancers is not None else []
+        self.scheduling_strategy = (
+            scheduling_strategy if scheduling_strategy is not None else "REPLICA"
+        )
+        self.platform_version = platform_version
+        self.tags = tags if tags is not None else []
+        self.pending_count = 0
```
bblommers, line 532: We can probably improve the logic here as well, as this can now become equal to `desired_count - ecs_running_count` (with a minimum of 0, just in case the user sets the running count larger then the desired count

moto/ecs/models.py
```diff
@@ -507,14 +516,31 @@ def __init__(
...
+        self.region_name = cluster.region_name
+        self._account_id = backend.account_id
+        self._backend = backend
+
+        try:
+            ecs_running_count = int(getenv("MOTO_ECS_SERVICE_RUNNING", 0))
```
bblommers, line 538: Same here - theoretically this can be set to a negative number, so I think we should specify `min(ecs_running_count, 0)`

moto/ecs/models.py
bblommers, line 532: This attribute isn't actually used at the moment - `self.deployments` currently uses `"pendingCount": self.desired_count`, which will be invalid with this change

bblommers (changes requested):

bblommers (approved):
Nice one, thanks @rafcio19!

PR #6293, opened 2023-05-07
EC2: validate pagination parameters

rafcio19:
Closes https://github.com/getmoto/moto/issues/6282

### DIFF
--- a/moto/ec2/responses/instances.py
+++ b/moto/ec2/responses/instances.py
@@ -14,6 +14,12 @@ from ._base_response import EC2BaseResponse
 class InstanceResponse(EC2BaseResponse):
     def describe_instances(self) -> str:
         self.error_on_dryrun()
+        # https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/ec2/client/describe_instances.html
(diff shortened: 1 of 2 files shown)


rafcio19:
@bblommers I can't replicate these fails locally when running tests in servermode. I am on the same versions of werkzeug, botocore and boto3 as the github runner 🤔 
Any tips?

bblommers:
Hey @rafcio19, the EC2 tests in ServerMode run in parallel, to speed things up. That means that every test can see resources created in every other test.

Because the `test_get_paginated_instances`-test can no longer provide InstanceId's, it returns every instance created across every test executed so far.

I think the assertions should be changed to:
```
resp2 = client.describe_instances(NextToken=next_token)
assert len(resp2["Reservations"]) >= 7  # 12 total - 5 from the first call, and at least 7 from the second
```
The `assert "NextToken" not in resp2  # This is it - no more pages` can then be removed - we have no idea whether Moto contains enough instances for another page or not.

bblommers (approved):
Awesome - thanks @rafcio19!

Issue #6294, opened 2023-05-07
S3 PutObjectLegalHold and PutObjectRetention do not take versionId into account


bentsku:
Hi! Following a bug report in LocalStack (https://github.com/localstack/localstack/issues/8183), it seems `PutObjectLegalHold` and `PutObjectRetention` do not take into account the `versionId` provided in the query string, and will instead always be applied to the latest version in the stack (the "current" object).

I'll open a PR to fix this issue.

PR #6295, opened 2023-05-07
fix S3 PutObjectRetention and PutObjectLegalHold not using versionId

bentsku:
As reported in #6294, `PutObjectRetention` and `PutObjectLegalHold` would not take into account the `versionId` as it was not accessing it the same way as `PutObjectTagging` for example. I've pulled the retrieval of the `versionId` outside of the `if` blocks, so that all operations needing it could access it.

I've ran the tests against AWS as well and it confirms the behaviour the user was expecting.


bblommers (approved):
LGTM - thanks @bentsku!

PR #6309, opened 2023-05-10
S3: fix checksum calculation

bentsku:
While working against AWS, I realised moto was returning different checksum after copying an object and recalculating its checksum. It seems the calculation was bit wrong, as it was using `hexdigest` instead of the binary `digest` when passing the hash to the base64 encoding.
I've tried to add a test to cover all checksum calculation by using a checksum recalculation property of `CopyObject`.

Also fixed CopyObject behaviour, when copying a key that has a Checksum, if not asked to be recalculated, it will also copy it to the destination key. This has been verified with AWS (same with the values, so we can check that the recalculation works properly, even when not provided by boto in the first request like in `test_copy_key_boto3_with_both_sha256_checksum`).

### DIFF
--- a/moto/s3/models.py
+++ b/moto/s3/models.py
@@ -2418,6 +2418,10 @@ class S3Backend(BaseBackend, CloudWatchMetricProvider):
             # Object copied from Glacier object should not have expiry
             new_key.set_expiry(None)
 
+        if src_key.checksum_value:
+            new_key.checksum_value = src_key.checksum_value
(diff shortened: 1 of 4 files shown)


bblommers:
Hey @bentsku, the linter has been fixed now - if you merge in the latest master, the build should pass again

bblommers (approved):
LGTM! Thanks @bentsku

PR #6318, opened 2023-05-12
Update getting_started.rst

oesteban-vx:
I have tried the unittest sample, but on a non-object variable
```python
mock_s3 = mock_s3()
```
It results in something similar to _variable not defined_, so I've _renamed_ the module to something different. While your code works, copying and pasting would be easier.


bblommers:
Hi @oesteban-vx, thanks for raising this! That example is indeed a bit odd.

I've raised #6325 as an alternative fix, that change is a better example IMO to show the difference between the module (`mock_s3`) and the instance variable (`self.mock_s3`).

PR #6319, opened 2023-05-12
Comprehend: implement canned responses for detect_* functions

rafcio19:
Closes https://github.com/getmoto/moto/issues/6268

### DIFF
--- a/IMPLEMENTATION_COVERAGE.md
+++ b/IMPLEMENTATION_COVERAGE.md
@@ -1197,7 +1197,7 @@
 
 ## comprehend
 <details>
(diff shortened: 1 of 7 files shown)


bblommers:
We can make the response configurable, if necessary. 
There are already API's in place for Athena and RDS Data that allow the user to configure the exact response - we could follow the same pattern here. (Example PR where I've added this feature for RDS Data: https://github.com/getmoto/moto/pull/6168/files)

Let's see what they say -if the requester is OK with a static response, we can stick with the current approach.

bblommers (approved):

PR #6328, opened 2023-05-15
IAM: missing CreateDate from get_group response

rafcio19:
Closes https://github.com/getmoto/moto/issues/6320

### DIFF
--- a/tests/test_iam/test_iam_groups.py
+++ b/tests/test_iam/test_iam_groups.py
@@ -8,6 +8,7 @@ import pytest
 from botocore.exceptions import ClientError
 from moto import mock_iam, settings
 from moto.core import DEFAULT_ACCOUNT_ID as ACCOUNT_ID
+from moto.backends import get_backend
 from freezegun import freeze_time
 from dateutil.tz import tzlocal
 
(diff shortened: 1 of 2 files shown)


tests/test_iam/test_iam_groups.py
```diff
@@ -107,10 +109,23 @@ def test_add_user_to_unknown_group():
...
+        conn.create_group(GroupName=group)
+        conn.create_user(UserName=user)
+        conn.add_user_to_group(GroupName=group, UserName=user)
+
+    # Execute
+    result = conn.get_group(GroupName=group)
+
+    # Verify
+    assert len(result["Users"]) == 1
+    assert result["Users"][0]["CreateDate"] == frozen_time
```
bpandola, line 128: Freezing time doesn't work for the server mode tests and isn't really necessary here.  The important thing is that we assert on the `CreateDate` field being present in the response and that it is a `datetime` object, which means that the value moto sends back is correctly parsed by the boto3 client.

rafcio19:
@bpandola @bblommers ready for review

bblommers (approved):
LGTM! Thanks @rafcio19

PR #6329, opened 2023-05-17
Core: add date to response metadata

rafcio19:
Closes https://github.com/getmoto/moto/issues/6321


rafcio19:
@bblommers ready for review

bblommers (approved):
LGTM - thanks @rafcio19!

bblommers:
Hi @rafcio19,

https://payments.getmoto.org/

PR #6330, opened 2023-05-17
SNS: Cross-account access for topics

viren-nadkarni:
This PR introduces cross-account support for SNS topics.

This basically means that topics will now be retrieved from the appropriate backend based on the value of account ID and region in the ARN. Any account can now retrieve an SNS topic with its ARN.

Cross-account access now works for `GetTopicAttributes`, `SetTopicAttributes`, `AddPermission`, `RemovePermission`, `DeleteTopic`, `Subscribe`, `ListSubscriptionsByTopic`, `Publish`, similar to AWS. 

IAM enforcement is left out of scope.

### DIFF
--- a/moto/events/models.py
+++ b/moto/events/models.py
@@ -4,7 +4,6 @@ import re
 import json
 import sys
 import warnings
-from collections import namedtuple
 from datetime import datetime
 from enum import Enum, unique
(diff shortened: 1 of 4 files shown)


bblommers (approved):

PR #6332, opened 2023-05-19
Add `Tags` filed for the response of `describe_spot_fleet_requests`

### LINKED ISSUE #6310: Tags on spot fleet requests missing from describe-spot-fleet-requests response
I'm trying to filter the response for describe-spot-fleet-requests by tags, but it looks like the response doesn't contain the tags.

To be clear, there are two different TagSpecification fields that AWS supports. One is inside the LaunchSpecifications field applying to the instance requests being create, and one is outside, at the same level in the configuration as the SpotFleetRequestConfig field, applying to the fleet as a whole. Currently only the tags inside the LaunchSpecifications is being returned.

I'm using moto via localstack, so I don't have a direct repro, but a slight modification to one of the existing tests should show this.
The following line of code in the tests adds the tags for the instances in the request:
https://github.com/getmoto/moto/blob/9d8d166571ce2dc4f13aac954ba798c621b578dc/tests/test_ec2/test_spot_fleet.py#L297
If we were to add immediately after that:
`config["TagSpecifications"] = tag_spec`
Then in AWS, the response to describe-spot-fleet-requests includes for example:
```            ...
            "SpotFleetRequestId": "sfr-03a07062-a189-40c3-9b4e-cb3c27f32441",
            "SpotFleetRequestState": "cancelled",
            "Tags": [
                {
                    "Key": "tag-1",
                    "Value": "foo"
                }, ...
            ]
...
```
But with the mocked AWS, my tests are failing because the request-level tags are missing.

Thanks for looking at this!

tsugumi-sys:
closes #6310

### DIFF
--- a/moto/ec2/models/spot_requests.py
+++ b/moto/ec2/models/spot_requests.py
@@ -185,6 +185,7 @@ class SpotFleetRequest(TaggedEC2Resource, CloudFormationModel):
         launch_specs: List[Dict[str, Any]],
         launch_template_config: Optional[List[Dict[str, Any]]],
         instance_interruption_behaviour: Optional[str],
+        tag_specifications: Optional[List[Dict[str, Any]]],
     ):
 
         self.ec2_backend = ec2_backend
@@ -201,6 +202,11 @@ class SpotFleetRequest(TaggedEC2Resource, CloudFormationModel):
         self.fulfilled_capacity = 0.0
 
         self.launch_specs = []
+        self.tags: Dict[str, Any] = (
+            convert_tag_spec(tag_specifications)
+            if tag_specifications is not None
+            else {}
+        )
 
         launch_specs_from_config = []
         for config in launch_template_config or []:
@@ -456,6 +462,7 @@ class SpotRequestBackend:
         launch_specs: List[Dict[str, Any]],
         launch_template_config: Optional[List[Dict[str, Any]]] = None,
         instance_interruption_behaviour: Optional[str] = None,
+        tag_specifications: Optional[List[Dict[str, Any]]] = None,
     ) -> SpotFleetRequest:
 
         spot_fleet_request_id = random_spot_fleet_request_id()
@@ -470,6 +477,7 @@ class SpotRequestBackend:
             launch_specs=launch_specs,
             launch_template_config=launch_template_config,
             instance_interruption_behaviour=instance_interruption_behaviour,
+            tag_specifications=tag_specifications,
         )
         self.spot_fleet_requests[spot_fleet_request_id] = request
         return request
--- a/moto/ec2/models/spot_requests.py
+++ b/moto/ec2/models/spot_requests.py
@@ -2,6 +2,7 @@ from collections import defaultdict
 from typing import Any, Dict, List, Optional, Tuple, TYPE_CHECKING
 
 from moto.core.common_models import BaseModel, CloudFormationModel
+from moto.ec2.exceptions import InvalidParameterValueErrorTagSpotFleetRequest
 
 if TYPE_CHECKING:
     from moto.ec2.models.instances import Instance
@@ -208,9 +209,7 @@ class SpotFleetRequest(TaggedEC2Resource, CloudFormationModel):
             tags = convert_tag_spec(tag_specifications)
             for resource_type in tags:
                 if resource_type != "spot-fleet-request":
-                    raise ValueError(
-                        f"The value for `ResourceType` must be `spot-fleet-request`, but got {resource_type} instead."
-                    )
+                    raise InvalidParameterValueErrorTagSpotFleetRequest(resource_type)
             self.tags.update(tags)
 
         launch_specs_from_config = []
--- a/moto/ec2/models/spot_requests.py
+++ b/moto/ec2/models/spot_requests.py
@@ -202,11 +202,16 @@ class SpotFleetRequest(TaggedEC2Resource, CloudFormationModel):
         self.fulfilled_capacity = 0.0
 
         self.launch_specs = []
-        self.tags: Dict[str, Any] = (
-            convert_tag_spec(tag_specifications)
-            if tag_specifications is not None
-            else {}
-        )
+
+        self.tags = {}
+        if tag_specifications is not None:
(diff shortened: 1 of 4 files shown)


tests/test_ec2/test_spot_fleet.py
```diff
@@ -64,10 +64,19 @@ def spot_config(subnet_id, allocation_strategy="lowestPrice"):
...
+                        "ResourceType": "instance",
+                        "Tags": [{"Key": "test", "Value": "value"}],
+                    }
+                ],
             },
         ],
         "AllocationStrategy": allocation_strategy,
         "FulfilledCapacity": 6,
+        "TagSpecifications": [
+            {"ResourceType": "instance", "Tags": [{"Key": "test2", "Value": "value2"}]}
```
bpandola, line 78: From the AWS docs:

> The key-value pair for tagging the Spot Fleet request on creation. The value for `ResourceType` must be `spot-fleet-request`, otherwise the Spot Fleet request fails.

moto/ec2/responses/spot_fleets.py
```diff
@@ -89,6 +91,16 @@ def request_spot_fleet(self) -> str:
         <item>
             <spotFleetRequestId>{{ request.id }}</spotFleetRequestId>
             <spotFleetRequestState>{{ request.state }}</spotFleetRequestState>
+            {% for resource_type in request.tags %}
+            <tag>
```
bpandola, line 95: This should be `<tagSet>` according to the AWS model.

moto/ec2/responses/spot_fleets.py
```diff
@@ -89,6 +91,16 @@ def request_spot_fleet(self) -> str:
         <item>
             <spotFleetRequestId>{{ request.id }}</spotFleetRequestId>
             <spotFleetRequestState>{{ request.state }}</spotFleetRequestState>
+            {% for resource_type in request.tags %}
```
bpandola, line 94: I don't think this is correct, according to the model.  I think you need to drop this line and modify your next `for` loop (see my comment there) to only get `spot-fleet-request` tags.

moto/ec2/responses/spot_fleets.py
```diff
@@ -89,6 +91,16 @@ def request_spot_fleet(self) -> str:
...
+            <tag>
+                {% for key, value in request.tags[resource_type].items() %}
```
bpandola, line 96: This should be modified to only pull in tags for the resource type you want.  Something like this:

`{% for key, value in request.tags.get('spot-fleet-request',{}).items() %}`

moto/ec2/responses/spot_fleets.py
```diff
@@ -89,6 +91,16 @@ def request_spot_fleet(self) -> str:
...
+                <item>
+                    <key>{{ key }}</key>
+                    <value>{{ value }}</value>
+                </item>
+                {% endfor %}
+            </tag>
```
bpandola, line 102: This should be `</tagSet>`

moto/ec2/responses/spot_fleets.py
```diff
@@ -89,6 +91,16 @@ def request_spot_fleet(self) -> str:
...
+            {% endfor %}
```
bpandola, line 103: This should be removed.

tests/test_ec2/test_spot_fleet.py
```diff
@@ -87,6 +96,7 @@ def test_create_spot_fleet_with_lowest_price():
     len(spot_fleet_requests).should.equal(1)
     spot_fleet_request = spot_fleet_requests[0]
     spot_fleet_request["SpotFleetRequestState"].should.equal("active")
+    spot_fleet_request["Tags"].should.equal({"Key": "test2", "Value": "value2"})
```
bpandola, line 99: This is incorrect.  The tags will come back as an array of dicts.

Check the docs: https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/ec2/client/describe_spot_fleet_requests.html

bpandola (changes requested):

tests/test_ec2/test_spot_fleet.py
tsugumi-sys: @bpandola 
Thank you for your reviews!
Should we check if `ResourceType` is correct or not here?
``` python
# moto/ec2/models/spot_requests.py
self.tags: Dict[str, Any] = (
    convert_tag_spec(tag_specifications)
    if tag_specifications is not None
    else {}
)
```

moto/ec2/models/spot_requests.py
```diff
@@ -202,6 +203,16 @@ def __init__(
...
+        self.tags = {}
+        if tag_specifications is not None:
+            tags = convert_tag_spec(tag_specifications)
+            for resource_type in tags:
+                if resource_type != "spot-fleet-request":
+                    raise ValueError(
+                        f"The value for `ResourceType` must be `spot-fleet-request`, but got {resource_type} instead."
+                    )
+            self.tags.update(tags)
+
```
bpandola, line 215: Instead of a built-in Python `ValueError` exception, you need to raise an appropriate EC2 custom exception. 

Here is a link to the one you likely want (or should subclass for your specific use-case): https://github.com/getmoto/moto/blob/c9d62e94025fb504881510278150d97d9c4dc002/moto/ec2/exceptions.py#L431

moto/ec2/models/spot_requests.py
```diff
@@ -202,6 +203,16 @@ def __init__(
 
         self.launch_specs = []
 
+        self.tags = {}
+        if tag_specifications is not None:
+            tags = convert_tag_spec(tag_specifications)
+            for resource_type in tags:
+                if resource_type != "spot-fleet-request":
+                    raise ValueError(
```
bpandola, line 211: It's good to add this check, but it also needs to be covered by a test.

Here is an example of how to do that: https://github.com/getmoto/moto/blob/c9d62e94025fb504881510278150d97d9c4dc002/tests/test_ec2/test_spot_instances.py#L34

bpandola (changes requested):

bpandola (approved):

bpandola:
Looks good. Thanks @tsugumi-sys

PR #6333, opened 2023-05-19
S3: Bucket names must be unique across accounts

viren-nadkarni:


moto/s3/models.py
```diff
@@ -937,6 +937,7 @@ def __init__(self, name: str, account_id: str, region_name: str):
         self.default_lock_days: Optional[int] = 0
         self.default_lock_years: Optional[int] = 0
         self.ownership_rule: Optional[Dict[str, Any]] = None
+        s3_backends.bucket_owners[name] = account_id
```
viren-nadkarni, line 940: This could be a module level variable but I feel like it belongs to the backend.

tests/test_s3/test_s3_file_handles.py
```diff
@@ -39,10 +45,20 @@ class TestS3FileHandleClosures(TestCase):
     def setUp(self) -> None:
         if settings.TEST_SERVER_MODE:
             raise SkipTest("No point in testing ServerMode, we're not using boto3")
-        self.s3 = s3model.S3Backend("us-west-1", "1234")
```
viren-nadkarni, line 42: This had to be changed because now we make references to `s3_backends` from with the `S3Backend` class

moto/s3/models.py
```diff
@@ -1645,6 +1647,13 @@ def create_bucket(self, bucket_name: str, region_name: str) -> FakeBucket:
     def list_buckets(self) -> List[FakeBucket]:
         return list(self.buckets.values())
 
+    def get_bucket_global(self, bucket_name: str) -> FakeBucket:
```
bblommers, line 1650: Does it make sense to just rewrite `get_bucket` to include this? I think that would make the cross-account logic available for all methods.

If that's not possible, or it introduces complexity somewhere else, I'd be happy to merge this as is - just trying to understand the pros/cons of this approach 🙂

bblommers (approved):
Awesome - thanks @viren-nadkarni!

moto/s3/models.py
```diff
@@ -1646,10 +1648,14 @@ def list_buckets(self) -> List[FakeBucket]:
         return list(self.buckets.values())
 
     def get_bucket(self, bucket_name: str) -> FakeBucket:
-        try:
+        if bucket_name in self.buckets:
             return self.buckets[bucket_name]
-        except KeyError:
-            raise MissingBucket(bucket=bucket_name)
+
+        if bucket_name in s3_backends.bucket_accounts:
```
bblommers, line 1654: Hi @viren-nadkarni! I realize it's been a while.. but what whas the reasoning behind this part?

I guess it made sense to me at the time, but looking at it again, I wonder why we don't just throw an AccessDenied error.

viren-nadkarni: Hi @bblommers, it's there to access a bucket that belongs to a different account

bblommers: Right. When I looked at it earlier, an AccessDenied error felt more intuitive.

 But thinking about it a little more, all our cross-account is allowed by default, so this implementation makes more sense. Thanks @viren-nadkarni

PR #6341, opened 2023-05-24
add CreateTime property to EC2 key pairs

AlexChesters:
currently the EC2 Key Pair model doesn't expose the `CreateTime` property (this property [is returned](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/ec2/client/describe_key_pairs.html) by `boto3`, however)

I've implemented this by following the same pattern established with the `CreateDate` property on an IAM User in `moto`

### DIFF
--- a/tests/test_ec2/test_key_pairs.py
+++ b/tests/test_ec2/test_key_pairs.py
@@ -1,4 +1,5 @@
 import pytest
+from datetime import datetime
 
 import sure  # noqa # pylint: disable=unused-import
 import boto3
@@ -7,6 +8,7 @@ from botocore.exceptions import ClientError
 from moto import mock_ec2, settings
 from uuid import uuid4
 from unittest import SkipTest
+from freezegun import freeze_time
 
 from .helpers import rsa_check_private_key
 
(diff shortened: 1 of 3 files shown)


tests/test_ec2/test_key_pairs.py
```diff
@@ -77,6 +79,7 @@ def test_key_pairs_create_dryrun_boto3():
 
 
 @mock_ec2
+@freeze_time("2014-01-01 05:00:00")
```
bpandola, line 82: Best to avoid freezing time unless absolutely an essential part of the test, as it doesn't work when we run the tests in server mode.

tests/test_ec2/test_key_pairs.py
```diff
@@ -101,6 +104,9 @@ def test_key_pairs_create_boto3():
     kps[0].should.have.key("KeyPairId")
     kps[0].should.have.key("KeyName").equal(key_name)
     kps[0].should.have.key("KeyFingerprint")
+    kps[0]["CreateTime"].replace(tzinfo=None).should.equal(
+        datetime(2014, 1, 1, 5, 0, 0)
+    )
```
bpandola, line 109: Probably best to just follow the existing convention (`should.have.key`) and not try to assert against the exact date/time (see my previous comment about using `freeze_time`).  The important thing to check is that the date/time moto passes back is properly deserialized by boto into a `datetime` object, e.g. something like `kps[0].should.have.key("CreateTime").should.be.a(datetime)`

bpandola (changes requested):

bpandola (approved):

PR #6342, opened 2023-05-25
ECR: Implement ImageAlreadyExistsException

sclothie:
I hit this error running some boto code and wanted to handle it in a unit test but then noticed that moto didn't implement it so I thought I'd give it a shot.


bblommers (approved):
LGTM - thanks for fixing this @sclothie!

PR #6370, opened 2023-06-05
SSM: Implement physical_resource_id property

snijbell:
The current SSM model does not provide the PhysicalResourceId, resulting in an empty string value when calling the `list_stack_resources` from a CloudFormation client.


bblommers (approved):
LGTM. Thank you for the PR @snijbell, and welcome to Moto!

bblommers:
Hi @snijbell,

https://payments.getmoto.org/

PR #6374, opened 2023-06-06
EC2: allow creation multiple flow logs on same resource id

### LINKED ISSUE #6369: Unexpected FlowLogAlreadyExists Error when all flow logs have no log_group_name
# Issue description
Adding 2 flow logs having different `log_destination` but no `log_group_name`(unset as deprecated) raises an undesired `FlowLogAlreadyExists`. 

# Reproduction (with awslocal & localstack, which make use of moto's EC2 backend under the hood)

Given one VPC with ID 'vpc-123456' and two buckets 'bucket_logs_A' and 'bucket_logs_B'
```
awslocal ec2 create-flow-logs --resource-id "vpc-123456" --resource-type "VPC" --log-destination-type=s3 --log-destination arn:aws:s3:::bucket_logs_A
awslocal ec2 create-flow-logs --resource-id "vpc-123456" --resource-type "VPC" --log-destination-type=s3 --log-destination arn:aws:s3:::bucket_logs_B
```
The first command passes, but the second :
```
An error occurred (FlowLogAlreadyExists) when calling the CreateFlowLogs operation: Error. There is an existing Flow Log with the same configuration and log destination.
```
Although they have a different destination

# Hints on where things happen

Here, you'll find that this condition is always True, when no `log_group_name` is provided for either of the flow_logs, because `None == None` gives `True`. 
https://github.com/getmoto/moto/blob/b180f0a015c625f2fb89a3330176d72b8701c0f9/moto/ec2/models/flow_logs.py#L258
Consequently, despite having different `log_destination` but because they have the same `resource_id`, the overall condition is True and an unexpected error is raised.

The solution is to protect against `None == None` in either cases (either log_group_name are provided XOR log_destination are provided)

lyderichti59:
See title, issue and commit for list of changes and their justifications

### DIFF
--- a/moto/ec2/models/flow_logs.py
+++ b/moto/ec2/models/flow_logs.py
@@ -253,11 +253,8 @@ class FlowLogsBackend:
 
             all_flow_logs = self.describe_flow_logs()
             if any(
-                fl.resource_id == resource_id
-                and (
-                    fl.log_group_name == log_group_name
-                    or fl.log_destination == log_destination
-                )
+                (fl.resource_id, fl.log_group_name, fl.log_destination)
+                == (resource_id, log_group_name, log_destination)
                 for fl in all_flow_logs
             ):
                 raise FlowLogAlreadyExists()
--- a/tests/test_ec2/test_flow_logs.py
+++ b/tests/test_ec2/test_flow_logs.py
@@ -68,6 +68,56 @@ def test_create_flow_logs_s3():
     flow_log["MaxAggregationInterval"].should.equal(600)
 
 
+@mock_s3
+@mock_ec2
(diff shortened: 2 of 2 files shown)


bblommers (approved):
LGTM - thank you for fixing this @lyderichti59!

bblommers:
Hi @lyderichti59,

https://payments.getmoto.org/

lyderichti59:
> Hi @lyderichti59,
> ...
> This website is still in beta, but everything should work. Feel free to open a bug or discussion if you run into any problems: https://github.com/getmoto/payments

Thank you @bblommers ! Much appreciated !

PR #6396, opened 2023-06-12
Fix TableMappings and ReplicationTaskSettings string in DMS.

damjad:
**Problem:** The `TableMappings` string returned in `describe_replication_tasks` for DMS requires `json.loads` twice to get the TableMappings `dict`. However, in `boto3` client, we have to apply `json.loads` only once. 

The same problem occurs for the `ReplicationTaskSettings` property.

For the following DMS task. 
```
client = boto3.client("dms", region_name="us-east-1")

    client.create_replication_task(
        ReplicationTaskIdentifier="test",
        SourceEndpointArn="source-endpoint-arn",
        TargetEndpointArn="target-endpoint-arn",
        ReplicationInstanceArn="replication-instance-arn",
        MigrationType="full-load",
        TableMappings='{"rules":[]}',
    )

    tasks = client.describe_replication_tasks(
        Filters=[{"Name": "replication-task-id", "Values": ["test"]}]
    )
```

**actual boto3** passes the following (expected)

```
task["TableMappings"].should.equal('{"rules":[]}')
json.loads(task["TableMappings"]).should.be.a(dict)
```

**boto3 with `mock_dms`** fails all of the following (found)

```
task["TableMappings"].should.equal('{"rules":[]}')
json.loads(task["TableMappings"]).should.be.a(dict)
```

**Root cause**
There is an extra call to `json.dumps` on `TableMappings` and `ReplicationTaskSettings` string.


bblommers (changes requested):
Looks like the linter is failing @damjad:

https://github.com/getmoto/moto/actions/runs/5248667715/jobs/9481410568

damjad:
> Looks like the linter is failing @damjad:
> 
> https://github.com/getmoto/moto/actions/runs/5248667715/jobs/9481410568

The checks have passed. Please, do let me know if anything else needs to be changed.

PR #6399, opened 2023-06-13
Fix SES V2 Responses

sugyan:
I was getting an error when calling `send_email` with certain raw message data using SES V2 (the current `master` branch test case seems to luckily pass without it happening).
I have tried to find the cause, and found that SES V2 uses `"protocol": "rest-json"`, so the body contains JSON, but the current `SESV2Response` parses it as a query string. This made `self.data` a strange dict, and [it was forced to use `json.loads` to retrieve the data](https://github.com/getmoto/moto/pull/6259/files#diff-4d53d2456ab2314d10a9f59631b895debb9ace766f76890603b6c585a3293c47R100). Here the raw message contains base64 encoding, so when the `=` padding is added, it gets split in parse as a query string!

To avoid this parse failure, I have changed to allow deserialization to the correct schema by adding a `aws_service_spec` value.
However, I noticed that SES V2 does not include headers like `X-Amz-Target`, so it is not possible to determine the method name (event name). In `BotocoreStubber.__call__`, the `event_name` was received, but it was not passed to callback; I wanted to add `event_name` as an argument to the callback function, but that would be too big an interface change. As a desperate measure, I temporarily included it in the headers so that it could be retrieved by `BaseResponse.setup_class`.

Other changes are included to accommodate the use of raw data in `send_email`, since it is not necessary to specify the `FromEmailAddress` or `Destination`.

### DIFF
--- a/moto/core/botocore_stubber.py
+++ b/moto/core/botocore_stubber.py
@@ -58,6 +58,7 @@ class BotocoreStubber:
             try:
                 recorder._record_request(request)
 
+                request.headers["X-Moto-EventName"] = event_name
                 status, headers, body = response_callback(
                     request, request.url, request.headers
                 )
--- a/moto/core/responses.py
+++ b/moto/core/responses.py
@@ -1,3 +1,4 @@
+import base64
 import boto3
 import functools
 import datetime
@@ -236,6 +237,10 @@ class BaseResponse(_TemplateEnvironmentMixin, ActionAuthenticatorMixin):
         """
         use_raw_body: Use incoming bytes if True, encode to string otherwise
         """
+        event_name: Optional[str] = None
+        if "X-Moto-EventName" in headers:
+            event_name = headers["X-Moto-EventName"]
+            del headers["X-Moto-EventName"]
         querystring: Dict[str, Any] = OrderedDict()
         if hasattr(request, "body"):
             # Boto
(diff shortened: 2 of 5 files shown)


sugyan:
The test fails with `TEST_SERVER_MODE` when using BotocoreStubber to inject `event_name` into headers.... It seems that It needed a different way to determine the `input_spec`, such as by looking at the contents of the HTTP request.
I would like to know if there is a better way.

bblommers:
Hi @sugyan, thanks for taking a shot at this! I've raised #6405, which uses an IMO more straightforward solution to fix this.

Can you verify that approach/branch fixes your problem?

sugyan:
Wow! I see…
`self.body` contains the original request data as it is. So by decoding it, the original request data can be correctly restored.
I didn't notice that... With your solution, I think my hack here is not necessary.
This pull-request is ok to close.

Since `self.data` (`self.querystring`) is still a weird result of parsing a json string as a query, I think it would be better to change it to use body json in other methods as well.
I think `self.data` and `get_params_dict` should not be used anymore.

bblommers:
I agree that `get_params_dict` is a bit strange @sugyan - but considering it does work, I think I'll leave it for now. I don't want to potentially introduce new bugs by changing it :slightly_smiling_face: 

Thanks again for raising this!

PR #6403, opened 2023-06-13
RDS: improve global clustering behaviour

rafcio19:


rafcio19:
@bblommers could you rerun the failing tests on https://github.com/getmoto/moto/actions/runs/5268843777/jobs/9527894782?pr=6403#step:9:7096
?
It doesn't look like it's my changes, rather a flakey response from servermode tests

bblommers:
Done @rafcio19!

rafcio19:
This PR caters for an example presented in https://github.com/getmoto/moto/issues/6260#issuecomment-1586361144, not sure if this is exactly what the OP needs.

bblommers:
Looks like this indeed covers the scenario of https://github.com/getmoto/moto/issues/6260#issuecomment-1523778464

The scenario in the first post isn't covered yet, I think, so I'm not going to close the attached issue yet, but I will merge this.

(
```
Create primary cluster with create_db_cluster in us-east-1
Create secondary cluster with create_db_cluster in us-west-2 with the primary cluster's arn as ReplicationSourceIdentifier
*Note that the secondary cluster is now added under the primary cluster's ReadReplicaIdentifiers

Create global database with SourceDBClusterIdentifier as the primary cluster's arn.
The global database will have the primary cluster listed as a cluster member, but the secondary cluster is missing from the list.
```
)

Issue #4944, opened 2022-03-17
Request: Add Python typing annotations


Mainstay-Noah-Huppert:
Hi. Thank you very much for this helpful library.

Would it be possible to add Python type annoations to the source code of this library? Or [MyPy typing stubs](https://mypy.readthedocs.io/en/stable/stubs.html)?

The benefit is that it would allow users to run a static type checker over their testing code which uses moto.

bblommers:
Hi @Mainstay-Noah-Huppert, which parts of Moto do you use that would need type annotations?

Straightforward tests that only use the Moto-decorator wouldn't benefit from any typing, as far as I can see, as the test would only contain boto3-calls.

Note that I'm not against the idea, just trying to gauge the most efficient way of meeting your usecase.

glungley:
Hi both, we have this issue, so maybe I can clarify.

When using moto with the decorators we get errors like the following:
`Untyped decorator makes function "x" untyped [misc]`

Although having types within moto itself would not be of any benefit to these tests. Having the decorators typed would mean we don't have to disable mypy for these tests and lose the value of having it for the boto3-calls.

iainelder:
Python typing annotations would be really useful for developers too.

I just started looking at the Organizations module and I wish I had types :-)

All the backends should return the same types as the boto3 clients do. By following what [boto3-stubs](https://pypi.org/project/boto3-stubs/) declares for the real boto3 library, I think it would make the mocks easier to maintain.

Bert, would you accept taking on mypy as a dependency so that I may add type annotations to the Organizations module?

bblommers:
> 
> All the backends should return the same types as the boto3 clients do. 

This is not necessarily true. Moto only intercepts the HTTP request that boto3 makes to AWS, so boto3 will still run a post-processing step after receiving the (mocked) response.
In a lot of cases, if we return a JSON-response it may look like the response from the boto3-client, but it's not guaranteed.

> Bert, would you accept taking on mypy as a dependency so that I may add type annotations to the Organizations module?

Are you suggesting to add it as a dev-dependency, and running mypy (on the organizations-module only, for now) as part of the linting-process? I'd be OK with that, yes.

iainelder:
> Are you suggesting to add it as a dev-dependency, and running mypy (on the organizations-module only, for now) as part of the linting-process? I'd be OK with that, yes.

[...]

PR #6406, opened 2023-06-14
Techdebt: MyPy all the things

bblommers:
Fixes: #4944

### DIFF
--- a/moto/__init__.py
+++ b/moto/__init__.py
@@ -34,6 +34,7 @@ mock_acmpca = lazy_load(".acmpca", "mock_acmpca", boto3_name="acm-pca")
 mock_amp = lazy_load(".amp", "mock_amp")
 mock_apigateway = lazy_load(".apigateway", "mock_apigateway")
 mock_apigatewayv2 = lazy_load(".apigatewayv2", "mock_apigatewayv2")
+mock_appconfig = lazy_load(".appconfig", "mock_appconfig")
 mock_appsync = lazy_load(".appsync", "mock_appsync")
 mock_athena = lazy_load(".athena", "mock_athena")
 mock_applicationautoscaling = lazy_load(
@@ -210,7 +211,7 @@ mock_all = MockAll
 # logging.getLogger('boto').setLevel(logging.CRITICAL)
 
 __title__ = "moto"
-__version__ = "4.1.9.dev"
+__version__ = "4.1.12.dev"
 
 
 try:
--- a/moto/__init__.py
+++ b/moto/__init__.py
@@ -1,17 +1,31 @@
 import importlib
 import sys
 from contextlib import ContextDecorator
+from moto.core.models import BaseMockAWS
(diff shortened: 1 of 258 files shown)


moto/core/models.py
```diff
@@ -116,9 +120,9 @@ def stop(self, remove_data: bool = True) -> None:
             self.disable_patching()  # type: ignore[attr-defined]
 
     def decorate_callable(
-        self, func: Callable[..., CALLABLE_RETURN], reset: bool, remove_data: bool
-    ) -> Callable[..., CALLABLE_RETURN]:
-        def wrapper(*args: Any, **kwargs: Any) -> CALLABLE_RETURN:
+        self, func: Callable[..., "BaseMockAWS"], reset: bool, remove_data: bool
+    ) -> Callable[..., "BaseMockAWS"]:
+        def wrapper(*args: Any, **kwargs: Any) -> "BaseMockAWS":
```
andersk, line 125: `func` is the user’s function being decorated. There’s no reason to expect it to return `BaseMockAWS`. All we can say is that our wrapper will return the same type that the user’s function returned, whatever it is. The type variable `CALLABLE_RETURN` was more correct.

moto/core/models.py
```diff
@@ -427,7 +430,9 @@ class base_decorator:
     def __init__(self, backends: BackendDict):
         self.backends = backends
 
-    def __call__(self, func: Optional[Callable[..., Any]] = None) -> BaseMockAWS:
+    def __call__(
+        self, func: Optional[Callable[..., Any]] = None
+    ) -> Union[BaseMockAWS, Callable[..., BaseMockAWS]]:
```
andersk, line 435: Same: if `func` is provided, we return a callable returning whatever type `func` returned, not `BaseMockAWS`.

moto/__init__.py
```diff
@@ -1,17 +1,27 @@
...
-    def f(*args, **kwargs):
+TEST_METHOD = TypeVar("TEST_METHOD", bound=Callable[..., Any])
+
+
+def lazy_load(
+    module_name: str,
+    element: str,
+    boto3_name: Optional[str] = None,
+    backend: Optional[str] = None,
+) -> Callable[..., BaseMockAWS]:
```
andersk, line 16: This annotation is wrong when a `func` is provided.

moto/core/models.py
```diff
@@ -65,10 +66,13 @@ def __init__(self, backends: BackendDict):
             self.reset()  # type: ignore[attr-defined]
 
     def __call__(
-        self, func: Callable[..., Any], reset: bool = True, remove_data: bool = True
-    ) -> Any:
+        self,
+        func: Callable[..., "BaseMockAWS"],
+        reset: bool = True,
+        remove_data: bool = True,
+    ) -> Callable[..., "BaseMockAWS"]:
```
andersk, line 73: Same: there’s no reason to expect the user’s `func` to return `BaseMockAWS`.

PR #6412, opened 2023-06-15
Feat: Support All AWS Partitions

joel-aws:
This PR extends support of moto to all AWS Partitions.

One can see all Partitions, Regions, and DNS Suffixes with: `curl https://raw.githubusercontent.com/boto/botocore/develop/botocore/data/endpoints.json | jq '.partitions[] | [.partitionName, .partition, .regions, .dnsSuffix]'`

### DIFF
--- a/moto/core/base_backend.py
+++ b/moto/core/base_backend.py
@@ -195,14 +195,10 @@ class AccountSpecificBackend(Dict[str, SERVICE_BACKEND]):
         self.regions = []
         if use_boto3_regions:
             sess = Session()
-            self.regions.extend(sess.get_available_regions(service_name))
             for partition in sess.get_available_partitions():
-                if partition != "aws":
-                    self.regions.extend(
-                        sess.get_available_regions(
-                            service_name, partition_name=partition
-                        )
-                    )
+                self.regions.extend(
+                    sess.get_available_regions(service_name, partition_name=partition)
+                )
         self.regions.extend(additional_regions or [])
         self._id = str(uuid4())
 
--- a/moto/core/base_backend.py
+++ b/moto/core/base_backend.py
@@ -55,11 +55,13 @@ class BaseBackend:
         urls = {}
         for url_base in url_bases:
             # The default URL_base will look like: http://service.[..].amazonaws.com/...
-            # This extension ensures support for the China, ISO, & ISOB regions
+            # This extension ensures support for the China & ISO regions
             alt_dns_suffixes = {
                 "cn": "amazonaws.com.cn",
                 "iso": "c2s.ic.gov",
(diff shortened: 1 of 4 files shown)


moto/core/base_backend.py
```diff
@@ -189,12 +196,13 @@ def __init__(
...
             sess = Session()
             self.regions.extend(sess.get_available_regions(service_name))
-            self.regions.extend(
-                sess.get_available_regions(service_name, partition_name="aws-us-gov")
-            )
-            self.regions.extend(
-                sess.get_available_regions(service_name, partition_name="aws-cn")
-            )
+            for partition in sess.get_available_partitions():
+                if partition != "aws":
```
bblommers, line 200: Minor nitpick: can we just add all partitions? `get_available_regions("ec2")` is functionally equivalent to `get_availble_regions("ec2", partition_name="aws")`, as far as I know.

So removing the initial `self.regions.extend(sess.get_available_regions(service_name))` and this if-statement should result in the same outcome, and is a bit cleaner.

bblommers (changes requested):
Hi @joel-aws, thanks for the PR! Please see my comment

moto/core/base_backend.py
joel-aws: Great suggestion; fixed.

joel-aws:
I realized I was querying an old `endpoints.json`, so I just added the most recent DNS Suffixes.

bblommers (approved):
LGTM - thanks again @joel-aws!

bblommers:
Hi @joel-aws! 

Another followup question/issue around this feature:
The `get_available_regions()` will be potentially deprecated by the AWS SDK team - see https://github.com/aws/aws-sdk/issues/206#issuecomment-1471354853

The suggested alternative way to get this region-information is to use the SSM parameter store instead, and I am looking into a [feasible implementation](https://github.com/getmoto/moto/pull/6362) for Moto.

The main problem I am running into is that these ISO regions are not exposed in the SSM parameter store. The US-gov and China regions are exposed, so it would make sense to me if the ISO-regions are also part of it.

Is that something you can raise internally, to see if that  information can be included/exposed? I understand that it's probably an entirely different team responsible for this, but I'm assuming that a request from an internal employee would hold more weight than a random support ticket from a regular user :slightly_smiling_face:

joel-aws:
@bblommers -- apologies for the delay in responding!

I am now posing this question internally... On the flip side, we value customer input via service tickets. If you haven't already opened one, could you do that and share the ticket # ?

bblommers:
No problem @joel-aws - thanks for getting back to me.

I haven't been on top of this either, to be honest - I only just opened a ticket now, nr `171584385800169`.

PR #6414, opened 2023-06-16
Fix SESV2Response, and SESBackend.send_raw_email

sugyan:
In relation to #6399 and #6405,  I noticed that unexpected errors occur for certain inputs.

- The value of the email address specified in Destination can also be a value with a display-name, such as `"foo, bar" <to@example.com>`.
  - This could cause the parsing of destinations in `SESBackend.send_raw_email()` to be split into multiple invalid strings because of the simple use of `str.split()`.
  - It seems that `email.utils.getaddresses` can be used to correctly parse them.
- In #6405, only `send_email()` was fixed to decode `self.body` as JSON, but it still needed to be fixed elsewhere as well.
  - For example, `Topics[].DisplayName` in `create_contact_list()` can be any string. Of course, you can include `"="`!
  - This still causes the problem I pointed out in #6399, so `get_params_dict()` should not be used, as I wrote in my comment.

### DIFF
--- a/moto/ses/models.py
+++ b/moto/ses/models.py
@@ -2,7 +2,7 @@ import json
 import email
 import datetime
 from email.mime.base import MIMEBase
-from email.utils import parseaddr
+from email.utils import formataddr, getaddresses, parseaddr
 from email.mime.multipart import MIMEMultipart
 from email.encoders import encode_7or8bit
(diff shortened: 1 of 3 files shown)


bblommers (approved):
Nice one - thanks for fixing this as well @sugyan!

PR #6421, opened 2023-06-18
Techdebt: MyPy improvements

bblommers:


bblommers:
A proper/working version is now part of Moto with #7045

PR #6424, opened 2023-06-19
Add cURL install to Dockerfile

benjamingorman:
See https://github.com/getmoto/moto/issues/6423


mikegrima (approved):
Will merge when tests pass.

PR #6425, opened 2023-06-19
EC2: Return appropriate error when calling EC2.Client.replace_route() with missing route

LewisGaul:
Currently the `EC2.Client.replace_route()` API does not error if a missing route is given, which is inconsistent with AWS behaviour.

The expected error code is `InvalidRoute.NotFound`, in line with what's returned from `delete_route()` and similar cases such as `InvalidRouteTable.NotFound` being returned if an invalid route table ID is given. However, AWS current behaviour is to return `InvalidParameterValue` instead, so that's what this PR gets moto to do (the [docs](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_ReplaceRoute.html) are vague about what error code to expect, but this has been manually checked). This issue has been raised with AWS via a case (unfortunately no public link available), but there is not yet agreement to fix this.

### DIFF
--- a/moto/ec2/exceptions.py
+++ b/moto/ec2/exceptions.py
@@ -444,6 +444,14 @@ class InvalidParameterValueErrorTagSpotFleetRequest(EC2ClientError):
         )
 
 
+class InvalidParameterValueErrorReplaceRoute(EC2ClientError):
+    def __init__(self, cidr: str):
+        super().__init__(
+            "InvalidParameterValue",
+            f"There is no route defined for '{cidr}' in the route table. Use CreateRoute instead.",
+        )
+
+
 class EmptyTagSpecError(EC2ClientError):
     def __init__(self) -> None:
         super().__init__(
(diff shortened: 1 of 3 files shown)


mikegrima:
Looks like a small issue with linting:
```moto/ec2/models/route_tables.py:466:58: error: Argument 1 to "InvalidParameterValueErrorReplaceRoute" has incompatible type "Optional[str]"; expected "str"  [arg-type]```

LewisGaul:
The failing testcase `test_duplicate_network_acl_entry` seems unrelated:
`botocore.exceptions.ClientError: An error occurred (NetworkAclEntryAlreadyExists) when calling the CreateNetworkAclEntry operation: The network acl entry identified by 100 already exists.`

mikegrima:
We have seen tests fail transiently before which could result in other tests also failing. It looks like this one would appear to be transient since all the others have passed.

PR #6427, opened 2023-06-20
pyparsing 3.1.0 support, remove experimental warning on Glue partition filtering

robert-schmidtke:
Pyparsing 3.1.0 was released (https://github.com/pyparsing/pyparsing/releases/tag/3.1.0) and has added `DelimitedList` in favor of `delimited_list`. This PR handles the announced future deprecation.

Also, the experimental warning on Glue partition filtering is removed. I use this excessively in my tests and found no issues, so I think this is safe to do.

### DIFF
--- a/moto/glue/utils.py
+++ b/moto/glue/utils.py
@@ -15,12 +15,18 @@ from pyparsing import (
     Suppress,
     Word,
     alphanums,
-    delimited_list,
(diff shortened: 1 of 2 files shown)


mikegrima (approved):
LGTM - will merge on successful tests.

PR #6428, opened 2023-06-20
AWS Config support for Resource Type Exclusions

mikegrima:
- Added support for AWS Config resource type exclusions
- Updated the data model to account for the newest changes to boto for AWS Config recorders


mikegrima:
@bblommers Should I also update the boto3 and botocore versions in the `setup.cfg`? The config changes are reflected in the latest versions.

bblommers:
> @bblommers Should I also update the boto3 and botocore versions in the `setup.cfg`? The config changes are reflected in the latest versions.

No, I don't think that's necessary. Moto can still be used with an older boto3 version

bblommers (approved):
LGTM - thanks @mikegrima!

PR #6429, opened 2023-06-20
add support for maintenance window targets

pinzon:
This PR adds support for the next operations:

- [x] register_target_with_maintenance_window
- [x] describe_maintenance_window_targets
- [x] deregister_target_from_maintenance_window

### DIFF
--- a/moto/ssm/models.py
+++ b/moto/ssm/models.py
@@ -1,5 +1,4 @@
 import re
-import uuid
 from dataclasses import dataclass
 from typing import Any, Dict, List, Iterator, Optional, Tuple
 from typing import DefaultDict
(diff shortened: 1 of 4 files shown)


moto/ssm/models.py
```diff
@@ -1,4 +1,5 @@
 import re
+import uuid
```
bblommers, line 2: Minor nit: 

users can seed Moto, so that 'random' identifiers are no longer random anymore. That means that for every 'random' identifier, we have to use a seeded generator:
```
from moto.moto_api._internal import mock_random
mock_random.uuid4()
```

Docs if you're interested: http://docs.getmoto.org/en/latest/docs/configuration/recorder/index.html#deterministic-identifiers

bblommers (changes requested):

bblommers (approved):
Thanks @pinzon!

PR #6430, opened 2023-06-21
add support for ssm maintenance window tasks

pinzon:
This PR adds support for some operations related to Maintenance Window Tasks:

- [X] register_task_with_maintenance_window
- [X] describe_maintenance_window_tasks
- [X] deregister_task_from_maintenance_window

I'll rebase it after #6429 is merged

### DIFF
--- a/moto/ssm/models.py
+++ b/moto/ssm/models.py
@@ -1,4 +1,5 @@
 import re
+import uuid
 from dataclasses import dataclass
 from typing import Any, Dict, List, Iterator, Optional, Tuple
 from typing import DefaultDict
--- a/moto/ssm/models.py
+++ b/moto/ssm/models.py
@@ -976,10 +976,77 @@ class FakeMaintenanceWindowTarget:
         return str(random.uuid4())
 
 
+class FakeMaintenanceWindowTask:
+    def __init__(
+        self,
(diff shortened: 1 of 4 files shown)


mikegrima:
Codecov complains that there are 2 spots not covered by tests. Mostly a nitpick, but we would prefer you have coverage for it.

moto/ssm/models.py
```diff
@@ -1,4 +1,5 @@
 import re
+import uuid
```
bblommers, line 2: This should use the `mock_random` as well

tests/test_ssm/test_ssm_maintenance_windows.py
```diff
@@ -274,3 +274,110 @@ def test_deregister_target_from_maintenance_window():
         WindowId=window_id,
     )
     resp.should.have.key("Targets").should.have.length_of(0)
+
+
+def test_describe_maintenance_window_with_no_task_or_targets():
```
bblommers, line 279: Looks like this is missing the `mock_ssm`-decorator

bblommers (changes requested):

bblommers (approved):
Awesome - thanks @pinzon!

PR #6434, opened 2023-06-22
Fixed bug in new Config recorder logic

mikegrima:
Fixed logic bug in handling the default empty `exclusionByResourceTypes` dictionary passed in.


bblommers (approved):
LGTM!

PR #6436, opened 2023-06-23
fix: missing type from param update response

rafcio19:
Resolves https://github.com/getmoto/moto/issues/6416


bblommers (approved):
Awesome - thanks for following up on this @rafcio19!

PR #6437, opened 2023-06-23
add GroupId to SecurityGroupRules

ackdav:
There is a small parity issue within the operation `describe-security-group-rules` ([aws-examples-link](https://docs.aws.amazon.com/cli/latest/reference/ec2/describe-security-group-rules.html#examples)) as it is missing the field `GroupId`. This appeared due to a ticket within [localstack](https://github.com/localstack/localstack/issues/8528), which very nicely describes the issue.

This PR is a quick draft to resolve this - I'm not particularly familiar with ec2 or moto (:smile:) and haven't added a test to this yet. something like `assert "GroupId" in rules[0]` [here](https://github.com/getmoto/moto/blob/dab6318f90dba1fdbb758603879e430e038a3c85/tests/test_ec2/test_security_groups.py#L589)? thx for any hints

### DIFF
--- a/moto/ec2/models/security_groups.py
+++ b/moto/ec2/models/security_groups.py
@@ -558,8 +558,11 @@ class SecurityGroupBackend:
             )
         rules = []
         for group in matches:
-            rules.extend(group.ingress_rules)
-            rules.extend(group.egress_rules)
+            group_rules = {"group_id": group.group_id, "rules": []}
+            group_rules["rules"].extend(group.ingress_rules)
+            group_rules["rules"].extend(group.egress_rules)
+            if len(group_rules["rules"]) > 0:
+                rules.append(group_rules)
 
         return rules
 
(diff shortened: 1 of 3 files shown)


bblommers (changes requested):
> something like assert "GroupId" in rules[0] [here](https://github.com/getmoto/moto/blob/dab6318f90dba1fdbb758603879e430e038a3c85/tests/test_ec2/test_security_groups.py#L589)
>
Yes, a simple assert should be enough to get test coverage. 

Looks like the linter is failing as well: https://github.com/getmoto/moto/actions/runs/5358273739/jobs/9720346981?pr=6437

bblommers (approved):
LGTM - thank you for fixing this @ackdav!

bblommers:

https://payments.getmoto.org/

Feel free to open a bug or discussion if you run into any problems:

PR #6186, opened 2023-04-07
RDS: Various improvements

bblommers:
Fixes #6151 
Fixes #6175 

New Methods:
 - create_db_cluster_parameter_group()
 - create_global_cluster()
 - delete_db_cluster_parameter_group()
 - delete_global_cluster()
 - describe_db_cluster_parameter_groups()
 - describe_db_cluster_parameters()
 - describe_db_subnet_groups()
 - describe_global_clusters()
 - promote_read_replica_db_cluster()
 - remove_from_global_cluster()

Miscellaneous:
 - RDS: Describe_db_clusters() now returns the parameters KmsKeyId, NetworkType, DBSubnetGroupName, ScalingConfiguration

PR #6444, opened 2023-06-25
RDS: Automated snapshots now have the appropriate SnapshotType

### LINKED ISSUE #6419: rds_client.create_db_snapshot, describe_db_snapshots behavior not right due to the snapshot type attribute
## What went wrong?

Based on this aws document [create_db_snapshot](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/rds/client/create_db_snapshot.html), the db snapshot created by this method should have ``SnapshotType = "manual"``.

However, in moto's implementation the [DatabaseSnapshot](https://github.com/getmoto/moto/blob/master/moto/rds/models.py#L1051) model doesn't have snapshot type attribute ( but it is defined in filters, werid)

## How to fix it 

1. add ``snapshot_type`` attribute in ``DatabaseSnapshot.__init__``
2. update ``create_db_snapshot`` method and ``describe_db_snapshots`` method.

## One more thing

I bet the ``create_db_cluster_snapshot`` related methods and datamodel has the same issue.

bblommers:
SnapshotType is set to `manual` for snapshots created by the user, and set to `automated` for snapshots created when a cluster is deleted.
This applies to both DatabaseSnapshots and ClusterSnapshots.

The `snapshot-type`-filter has also been fixed.

This PR also removes some duplicate validation when copying a snapshot. We now just call the `create..`-method, instead of duplicating the entire logic on copy.

Fixes #6419

### DIFF
--- a/moto/rds/models.py
+++ b/moto/rds/models.py
@@ -436,13 +436,20 @@ class ClusterSnapshot(BaseModel):
         "db-cluster-snapshot-id": FilterDef(
             ["snapshot_id"], "DB Cluster Snapshot Identifiers"
         ),
-        "snapshot-type": FilterDef(None, "Snapshot Types"),
+        "snapshot-type": FilterDef(["snapshot_type"], "Snapshot Types"),
         "engine": FilterDef(["cluster.engine"], "Engine Names"),
     }
 
-    def __init__(self, cluster: Cluster, snapshot_id: str, tags: List[Dict[str, str]]):
+    def __init__(
+        self,
+        cluster: Cluster,
+        snapshot_id: str,
+        snapshot_type: str,
+        tags: List[Dict[str, str]],
+    ):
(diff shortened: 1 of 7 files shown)


moto/rds/models.py
```diff
@@ -551,9 +558,7 @@ def __init__(self, **kwargs: Any):
             )
         self.db_cluster_identifier: Optional[str] = kwargs.get("db_cluster_identifier")
         self.db_instance_identifier = kwargs.get("db_instance_identifier")
-        self.source_db_identifier: Optional[str] = kwargs.get(
-            "source_db_ide.db_cluster_identifierntifier"
-        )
+        self.source_db_identifier: Optional[str] = kwargs.get("source_db_identifier")
```
bblommers, line 561: This was introduced in #6186 - looks like an accidental copy/paste

PR #6446, opened 2023-06-26
Fix service name for ACM backend

giograno:
This PR fixes the service name of the `BackendDict` for `acm` (it currently has `ec2` instead).
 
I wondered if there is a rationale behind this or if that's simply a mistake.


viren-nadkarni:
FYI 2a3cb4174b49d1bbd9d285eb52ef1310ef5db032 fixes the service names for remaining BackendDict initialisations. Boto3 now correctly returns the regions for these services.

PR #6449, opened 2023-06-27
DynamoDB: correct no table error

rafcio19:
Closes https://github.com/getmoto/moto/issues/6441


bblommers (approved):
LGTM - thanks for fixing this @rafcio19!

PR #6451, opened 2023-06-27
Core: Reduce slowdown `mock_all`

### LINKED ISSUE #6448: mock_all takes a long time in moto==4.1.12
`mock_all` is a ~6x slower in `moto==4.1.12` than in previous versions. I noticed this because tests in one of my projects that use the decorator started to time out.

I assume one service in particular is taking much longer on startup than it used to (although I haven't dug into the root cause yet).

Here are steps to reproduce:

```
git@github.com:getmoto/moto.git
cd moto

make

git checkout 4.1.12

python -m pytest tests/test_core/test_mock_all.py
=============================================================================================== test session starts ================================================================================================

tests/test_core/test_mock_all.py ..                                                                                                                                                                          [100%]

========================================================================================== 2 passed in 162.07s (0:02:42) ===========================================================================================

git checkout 4.1.11

python -m pytest tests/test_core/test_mock_all.py
=============================================================================================== test session starts ================================================================================================                                                                                                                                                                            

tests/test_core/test_mock_all.py ..                                                                                                                                                                          [100%]

================================================================================================ 2 passed in 27.06s ================================================================================================
```

bblommers:
Fixes #6448 

Some timings
----------------------

How long does it take to execute `pytest tests/test_core/test_mock_all.py`?
```
4.1.11          25.13s
4.1.12          132.66s (0:02:12)
this patch      12.86s
```

Changes:
--------------

Running a profiler points to the fact that `responses.BaseResponse.__eq__` is called a lot. The original response allows for both strings and Patterns to be compared. But because our `CallbackResponse` always only has Patterns, we can simplify the comparison logic.
This is not a big change, but it made for a significant change because of the number of executions.

This PR also changes how we store Responses in our CustomRegistry. The default registry stores everything in a simple list. But as we can have hundreds of Responses, naively going through the list one-by-one is time-consuming.
Changing the storage to type `dict[http_method, [Responses])` reduces the number of responses we have to search through, as we only care about potential Responses for the particular HTTP method of the incoming request (GET, POST, ..)

Lastly:
Supporting the ISO regions is a great change, but it does incur a significant slowdown, because of the additional Responses that we now need to manage.
Considering the number of total users impacted by the slowdown is much greater than the number of people using this, I've elected to now hide this functionality behind a feature flag.
This now needs an environment variable to work:
```
MOTO_ENABLE_ISO_REGIONS=true
```

@joel-aws As the contributor of the ISO regions, do you have any problems with this? 
I really do not like the fact that we're removing a feature, as a breaking change, but the slowdown of always having this feature enabled outweighs the down-sides IMO.


joel-aws:
@bblommers -- I think this is a reasonable change.

Only one request: could you add `MOTO_ENABLE_ISO_REGIONS` to the [docs](https://github.com/getmoto/moto/blob/master/docs/docs/configuration/environment_variables.rst)?

PR #6452, opened 2023-06-28
Elbv2.create rule links target group to lb

florianweyandt1337:
In order to search target_groups by LB-ARN-only when creating a new listener rule after listener creation, we need this link in place.

similar to #1500


bblommers (changes requested):
Looks like the linter is failing @florianweyandt1337: https://github.com/getmoto/moto/actions/runs/5398377211/jobs/9806986485?pr=6452

bblommers:
Hi @florianweyandt1337, hope you don't mind - I pushed a small change to get the linter to pass 

Thank you for the PR!

PR #6455, opened 2023-06-28
Lambda: create list_aliases functionality

freshollie:
- Creates the [list_aliases](https://boto3.amazonaws.com/v1/documentation/api/1.26.85/reference/services/lambda/client/list_aliases.html) functionality
- Redesign internal storage of aliases to correct mistaken implementation (Duplicate alias names could be created under the same lambda pointing to different versions)
- Adds tests for new functionality
- Worth noting I've not included functionality to filter by "version", can be implemented if needed

### DIFF
--- a/moto/awslambda/models.py
+++ b/moto/awslambda/models.py
@@ -1177,7 +1177,7 @@ class LambdaStorage(object):
         for config in self._functions[name]["versions"]:
             if str(config.version) == version:
                 return config
-    
+
         return None
 
     def _get_function_aliases(self, function_name: str) -> Dict[str, LambdaAlias]:
@@ -1209,7 +1209,7 @@ class LambdaStorage(object):
         if name in aliases:
             arn = f"arn:aws:lambda:{self.region_name}:{self.account_id}:function:{function_name}:{name}"
             raise ConflictException(f"Alias already exists: {arn}")
-    
+
         alias = LambdaAlias(
             account_id=self.account_id,
             region=self.region_name,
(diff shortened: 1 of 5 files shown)


freshollie:
Let me know if you want docs updated as part of this PR

moto/awslambda/models.py
```diff
@@ -954,43 +953,6 @@ def _create_zipfile_from_plaintext_code(code: str) -> bytes:
     def delete(self, account_id: str, region: str) -> None:
         lambda_backends[account_id][region].delete_function(self.function_name)
 
-    def delete_alias(self, name: str) -> None:
```
freshollie, line 957: Moved the logic to lambda storage level

moto/awslambda/models.py
```diff
@@ -1205,22 +1167,34 @@ def __init__(self, region_name: str, account_id: str):
         self.region_name = region_name
         self.account_id = account_id
 
+        # function-arn -> alias -> LambdaAlias
+        self._aliases: Dict[str, Dict[str, LambdaAlias]] = defaultdict(lambda: {})
```
freshollie, line 1171: Aliases are associated with the "lambda" not the "lambda version". We can't store these in "lambda version" as it makes it very complex to check if the alias already exists before writing. It also simplifies fetching a list of aliases as we don't need to iterate over each "version" to find it's alias

freshollie:
Will look into the failing test

bblommers (changes requested):
This looks really good. Thank you for this PR @freshollie!

To show our thanks, we'd like to share some of the donations that we've received with you. PR's like this are the big reason that Moto is as successful as it is - so it's only fair that you, as a contributor, get to share the spoils.

https://payments.getmoto.org/

PR #6460, opened 2023-06-29
Route53: Implement GetHealthCheckStatus

viren-nadkarni:
This PR adds support for GetHealthCheckStatus endpoint in Route53

<https://docs.aws.amazon.com/Route53/latest/APIReference/API_GetHealthCheckStatus.html>


bblommers (approved):

PR #6465, opened 2023-06-30
Add type field to error response

viren-nadkarni:
AWS error responses have a field called `Type` which indicates where the blame for the error lies. There is little documentation on this, but HTTP 4xx errors seem to have this field set to `Sender`.

This PR introduces support for this field. To start with, SNS errors are correctly mapped.

The other two templates do not contain the field:
- `SINGLE_ERROR_RESPONSE`: Used for S3 response, see <https://docs.aws.amazon.com/AmazonS3/latest/API/ErrorResponses.html>
- `ERROR_RESPONSE`: Used for EC2 response, see <https://github.com/boto/botocore/blob/82f6e194716e36c838fc016eddc4a2ef8e449ec4/botocore/parsers.py#L614>

Also see: <https://github.com/boto/botocore/blob/82f6e194716e36c838fc016eddc4a2ef8e449ec4/tests/unit/test_parsers.py#L1200-L1271>

### DIFF
--- a/moto/core/exceptions.py
+++ b/moto/core/exceptions.py
@@ -46,7 +46,7 @@ class RESTError(HTTPException):
     request_id_tag_name = "RequestId"
 
     # When this field is set, the `Type` field will be included in the response
-    sender_fault = False
+    sender_fault = True
 
     templates = {
         "single_error": SINGLE_ERROR_RESPONSE,
--- a/moto/core/exceptions.py
+++ b/moto/core/exceptions.py
@@ -3,8 +3,6 @@ from jinja2 import DictLoader, Environment
 from typing import Any, List, Tuple, Optional
 import json
 
-# TODO: add "<Type>Sender</Type>" to error responses below?
-
 
 SINGLE_ERROR_RESPONSE = """<?xml version="1.0" encoding="UTF-8"?>
 <Error>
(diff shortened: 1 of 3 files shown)


moto/core/exceptions.py
```diff
@@ -3,8 +3,6 @@
 from typing import Any, List, Tuple, Optional
 import json
 
-# TODO: add "<Type>Sender</Type>" to error responses below?
-
 
 SINGLE_ERROR_RESPONSE = """<?xml version="1.0" encoding="UTF-8"?>
```
viren-nadkarni, line 7: This doesn't seem to contain the `Type` field

<https://docs.aws.amazon.com/AmazonS3/latest/API/ErrorResponses.html>

moto/core/exceptions.py
```diff
@@ -44,6 +45,9 @@ class RESTError(HTTPException):
     # most APIs use <RequestId>, but some APIs (including EC2, S3) use <RequestID>
     request_id_tag_name = "RequestId"
 
+    # When this field is set, the `Type` field will be included in the response
+    sender_fault = True
```
bblommers, line 49: Minor nitpick - but the attribute-name doesn't make it immediately clear to me what it does. How about `include_type_sender`?

bblommers (changes requested):

moto/core/exceptions.py
viren-nadkarni: Changed in 225023df6 :+1:

bblommers (changes requested):
LGTM, just need to get the tests to pass

bblommers (approved):
Awesome, thanks @viren-nadkarni!

PR #6477, opened 2023-07-03
EC2: fix DescribeLaunchTemplateVersions default response

ackdav:
The default response of ec2's DescribeLaunchTemplateVersions fails due to the variable `template` being referenced without assignment [here](https://github.com/getmoto/moto/blob/db002cb95813255a083ed4af24eb68ae3224f31e/moto/ec2/responses/launch_templates.py#L204). The default should be an empty list. This came to light because of an issue report in [localstack](https://github.com/localstack/localstack/issues/8606).
Granted, this is rather an ugly quick fix, but it seems to work. Alternatively the default could be to create an empty launch template? (Open to adjust if you could give me some guidance what would be a more sensible approach here)

### DIFF
--- a/moto/ec2/responses/launch_templates.py
+++ b/moto/ec2/responses/launch_templates.py
@@ -156,8 +156,10 @@ class LaunchTemplates(EC2BaseResponse):
         template_id = self._get_param("LaunchTemplateId")
         if name:
             template = self.ec2_backend.get_launch_template_by_name(name)
-        if template_id:
+        elif template_id:
             template = self.ec2_backend.get_launch_template(template_id)
+        else:
+            template = None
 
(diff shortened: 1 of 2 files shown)


bblommers (changes requested):
IMO the best (most readable) way would be to always ensure the `template`-variable is set, set to None if necessary.

So all the way at the top we can say:
```
if name:
    template = ...
elif template_id:
    template = ..
else:
    template = None
```

The other changes then become more obvious:
```
if versions and template is not None:
   ...
elif template is not None:
    ret_versions = template.versions
```

ackdav:
:facepalm: of course! that's way more readable... thx for the hint!

bblommers (approved):
Nice one - thank you for fixing this @ackdav!

PR #6479, opened 2023-07-04
ECR: add CreateRepository name validation

ackdav:
This is a small fix to add name validation for the `repository-name` within ECR's `create-repository`.
Currently there is no name validation during the creation - but the [documentation](https://docs.aws.amazon.com/AmazonECR/latest/APIReference/API_CreateRepository.html#API_CreateRepository_RequestSyntax) states that it needs to follow a regex pattern. I found this through a [localstack issue](https://github.com/localstack/localstack/issues/8051).


bblommers (approved):
LGTM - thanks @ackdav!

PR #6480, opened 2023-07-04
Update server_mode.rst

liborjelinek:
Add note on zsh shell error "no matches found moto[server]"


bpandola:
A bit too specific to include in the docs, which are at a much higher/general overview level.

PR #6492, opened 2023-07-06
Route53: create_hosted_zone() now returns the Location

### LINKED ISSUE #6464: The CreateHostedZone response for Route53 does not match the behavior of AWS
## Reporting Bugs

I'm using localstack, which implements its route53 mock service using moto.

The `CreateHostedZone` operation returns a response that misses some info compared to an actual AWS response. In particular, it misses the `Location` header, which is where official AWS SDKs are reading the location.

See https://github.com/localstack/localstack/pull/5639#issuecomment-1064527918 for a detailed analysis submitted to localstack a while ago

bblommers:
Fixes #6464

### DIFF
--- a/moto/route53/responses.py
+++ b/moto/route53/responses.py
@@ -81,6 +81,9 @@ class Route53(BaseResponse):
                 delegation_set_id=delegation_set_id,
             )
             template = Template(CREATE_HOSTED_ZONE_RESPONSE)
+            headers = {
+                "Location": f"https://route53.amazonaws.com/2013-04-01/hostedzone/{new_zone.id}"
+            }
             return 201, headers, template.render(zone=new_zone)
 
         elif request.method == "GET":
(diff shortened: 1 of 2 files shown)


moto/route53/responses.py
```diff
@@ -81,6 +81,9 @@ def list_or_create_hostzone_response(self, request: Any, full_url: str, headers:
                 delegation_set_id=delegation_set_id,
             )
             template = Template(CREATE_HOSTED_ZONE_RESPONSE)
+            headers = {
+                "Location": f"https://route53.amazonaws.com/2013-04-01/hostedzone/{new_zone.id}"
```
stof, line 85: shouldn't this use `full_url + '/' + new_zone.id` instead, to respect the potential custom domain ? I'm wondering whether the current patch will be working for localstack...

bblommers: Hi @stof! If I understand the architecture of LocalStack correctly, the `full_url` will always become `route53.amazonaws.com` for their invocations. So that wouldn't make a difference.

It is a bit tricky to return URL's, because when Moto is run inside Docker, whatever endpoint (and port) we expose inside the container may be very different than what is actually accessible by the user. Maybe LS can rewrite the Location to a URL that makes sense for them, such as `localhost.localstack.cloud`?

PR #6519, opened 2023-07-13
RDS: db instance identifier validation

rafcio19:
Closes https://github.com/getmoto/moto/issues/6511


rafcio19:
@bblommers I think this one is ready to go, only a single TF test set fails. Can you rerun it?

bblommers (approved):
Nice one - thank you @rafcio19!

PR #6520, opened 2023-07-13
EBS: Some fixes to snapshot uploading

jdtsillas:
Two fixes to support uploading a snapshot:

1. responses for putting a block and completing a snapshot should be 201 and 202 respectively rather than 200; HTTP 201 and 202 are expected by the .NET AWS API for the respective operations.

2. If the body cannot be utf-8 decoded, try base64.


bblommers:
Hi @jdtsillas, thanks for the PR! 
What is the reasoning behind the base64-change?

Also, it would be good to have tests for these changes, ideally by extending the existing DotNet tests: 
https://github.com/getmoto/moto/tree/master/other_langs/tests_dotnet/ExampleTestProject

The tests are executed like so:
https://github.com/getmoto/moto/blob/master/.github/workflows/tests_sdk_dotnet.yml

moto/core/responses.py
```diff
@@ -253,8 +253,15 @@ def setup_class(
                 querystring[key] = [value]
 
         raw_body = self.body
+
         if isinstance(self.body, bytes) and not use_raw_body:
-            self.body = self.body.decode("utf-8")
+            try:
+                self.body = self.body.decode("utf-8")
+            except:
+                try
```
bblommers, line 261:
```suggestion
                try:
```

jdtsillas: The base64 change is to support some URL encodings that are used when the snapshot data come in this form. Not sure if it really matters since payload is not used AFAIK. The real fix is to just handle the decode exception. I'll change to using the try try pattern.

bblommers (approved):

PR #6539, opened 2023-07-19
Organizations: Update trusted service principals

viren-nadkarni:
See <https://docs.aws.amazon.com/organizations/latest/userguide/orgs_integrate_services_list.html>

Closes https://github.com/localstack/localstack/issues/8559


bblommers (approved):
LGTM

PR #6540, opened 2023-07-19
VPC: Add CloudFront, Ground Station and Lattice prefix lists

viren-nadkarni:
The address ranges are taken from AWS us-east-1 region.

Closes https://github.com/localstack/localstack/issues/8518


bblommers (approved):
Nice one - thank you @viren-nadkarni!

PR #6541, opened 2023-07-19
Batch: arraySize and child jobs

rafcio19:
Closes https://github.com/getmoto/moto/issues/6512

### DIFF
--- a/moto/batch/models.py
+++ b/moto/batch/models.py
@@ -6,7 +6,6 @@ import threading
 import time
 
 from sys import platform
-from copy import copy
 from itertools import cycle
 from time import sleep
 from typing import Any, Dict, List, Tuple, Optional, Set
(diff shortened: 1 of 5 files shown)


bblommers:
BTW, I would be happy to have a go at the implementation, I just don't know when/if I have time in the coming days.

rafcio19:
Thanks for the tips, leave it with me for another while.

oliver-helix:
Hey @rafcio19  @bblommers 
Thanks so much for your hard work here. Any update since July?

rafcio19:
Hey chaps, I haven't had much time to devote to moto since the summer. If someone else wants to have a go then please do; or you could 🤞 and give me a few weeks.

bblommers:
Hey @rafcio19, I just had a crack at it. Instead of an `ArrayJob`, I realized it was easier to just create additional regular jobs in the `submit_job`-method. The parent-job itself will not be executed, but has a list of child-jobs. 
When describing a job that happens to have any child-jobs, we can check immediately check the status of those child-jobs - so they are no longer hardcoded to 'SUCCEEDED'.

Based on my testing, this is all in-line with how AWS behaves

rafcio19:
Go nuts @bblommers, I've had it long enough. Will try to pick up more moto issues in the near future

PR #6550, opened 2023-07-21
Fix: SQS BatchRequestTooLong exception

sophcass:
When sending messages in batch to AWS SQS a `BatchRequestTooLong` is thrown by AWS if the length of all the messages put together is greater than the AWS limit. See AWS documentation [here](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/APIReference/API_SendMessageBatch.html#:~:text=Status%20Code%3A%20400-,BatchRequestTooLong,-The%20length%20of).

This PR fixes the logic that determines if a `BatchRequestTooLong` exception should be thrown. Prior to this PR, a `BatchRequestTooLong` exception is thrown only if a message in the batch of messages to be sent is greater than the AWS limit. After this PR, an exception will be thrown if the sum of all the messages exceeds the AWS limit.

### DIFF
--- a/moto/sqs/models.py
+++ b/moto/sqs/models.py
@@ -884,15 +884,8 @@ class SQSBackend(BaseBackend):
         ):
             raise InvalidBatchEntryId()
 
-        body_length = next(
-            (
-                len(entry["MessageBody"])
-                for entry in entries.values()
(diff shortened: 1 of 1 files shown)


bblommers (approved):
LGTM - thank you for fixing this @sophcass !

PR #6551, opened 2023-07-22
Fix code scanning alerts

bpandola:
The two range-based regexs were clearly incorrect.  The unescaped periods in the `meteringmarketplace` urls seem like an oversight, unless those urls work in some unique way I'm not understanding.


bblommers (approved):
Thanks @bpandola

PR #6553, opened 2023-07-22
Deny login if user account is disabled

quirozs:
When a user is disabled, it should raise a `NotAuthorizedException` exception when attempting to login


bblommers (approved):
LGTM - thank you for adding this feature to Moto @quirozs!

PR #6560, opened 2023-07-25
[Bug fix] Add required deduplication_id parameter to sns:PublishBatch

ckreiling:
This fixes a bug when publishing to an SNS FifoTopic. Currently any message published to a FifoTopic results in the following error:

> The topic should either have ContentBasedDeduplication enabled or MessageDeduplicationId provided explicitly


bblommers (approved):
LGTM - thank you for the fix @ckreiling, and welcome to Moto!

PR #6586, opened 2023-08-01
Add ECR support to lambda `update_function_code()`

### LINKED ISSUE #6584: lambda publish_version() doesn't generate new version
This problem was discovered with moto 4.1.12 but appears to exist with 4.1.14 also.

When calling lambda `publish_version()` on a new function, version "1" is created as expected.

If we then modify the function payload (by calling `update_function_code()` to change the image URI for an ECR-based function) and call `publish_version()` again, it appears that the existing version "1" is updated instead of creating a new version "2" as expected. (I don't actually care what the new version number is -- only that it actually gets created.) "Live" lambda does create a new version in this circumstance, as you'd expect.

If the root cause of this problem is not immediately obvious, let me know, and I will try to boil down our library code and unit test method into something small enough to be reasonably portable and reproducible.

rbailey-godaddy:
This apparently got missed, resulting in a null operation when this method was called for an ECR-based lambda.

This fix consolidates code management logic into a single internal helper method, and then adjusts `create_function()` and `update_function_code()` implementations to use that helper. There are a few puzzling aspects to the logic which are worthy of further review by someone who has a clue:

* The check for S3 input is slightly different; I went with the old "update" version which fires only when both `S3Bucket` and `S3Key` are present (vs old create behavior that didn't directly check `S3Key`). The Amazon documentation is vague regarding whether or not both must be present. I am uncertain if `and "S3Key" in updated_spec` should remain or be removed (or worse, continue to be treated differently for each use case).
* The "create" and "update" paths used totally different helpers with totally different exceptions for validating the actual S3 object. This code preserves the original logic for each so that the expected exceptions are returned and unit tests remain happy. I don't understand why "update" doesn't use the "create" logic, when the Amazon documentation suggests both raise the same exceptions.
* "Create" filled in dummy data for null `key`, whereas "update" did not. Now they both do, although it is unclear if this code is reachable since both helpers seem to raise exceptions on errors.

Fixes #6584

### DIFF
--- a/moto/awslambda/models.py
+++ b/moto/awslambda/models.py
@@ -479,61 +479,7 @@ class LambdaFunction(CloudFormationModel, DockerModel):
             datetime.datetime.utcnow()
         )
 
-        if "ZipFile" in self.code:
-            (
-                self.code_bytes,
-                self.code_size,
-                self.code_sha_256,
-                self.code_digest,
-            ) = _zipfile_content(self.code["ZipFile"])
-
-            # TODO: we should be putting this in a lambda bucket
-            self.code["UUID"] = str(random.uuid4())
-            self.code["S3Key"] = f"{self.function_name}-{self.code['UUID']}"
-        elif "S3Bucket" in self.code:
-            key = _validate_s3_bucket_and_key(self.account_id, data=self.code)
-            if key:
-                (
-                    self.code_bytes,
-                    self.code_size,
-                    self.code_sha_256,
-                    self.code_digest,
-                ) = _s3_content(key)
-            else:
-                self.code_bytes = b""
-                self.code_size = 0
-                self.code_sha_256 = ""
-        elif "ImageUri" in self.code:
-            if settings.lambda_stub_ecr():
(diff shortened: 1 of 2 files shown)


bblommers (changes requested):
Would you mind adding a test for this @rbailey-godaddy?

There is quite a bit of duplication going on now between `create` and `update`, and I'm happy to tackle that at a later stage - but having tests to verify that we're not going to break this (again) is always useful.

rbailey-godaddy:
> Would you mind adding a test for this @rbailey-godaddy?

Sure thing -- I'm OOO for the remainder of this week, but I'll pick it up again next week.

rbailey-godaddy:
@bblommers -- I have consolidated the logic for both functions into a single helper, although it is weirder than I hope it should be in order to preserve existing behavior examined by existing unit tests. I also added a new unit test for the "update ECR" case, and revised the PR description to better reflect what is going on.

bblommers (approved):
LGTM - thanks for fixing this @rbailey-godaddy!

PR #6593, opened 2023-08-04
Techdebt: Replace sure with regular assertions in SDB

kbalk:
Thought I would help out with the edits in the tests/test_s* directories.


bblommers (approved):
Awesome - thanks for the help @kbalk!

PR #6601, opened 2023-08-04
Techdebt: Replace sure with regular assertions in s3bucket_path

kbalk:


bblommers (approved):
LGTM

PR #6603, opened 2023-08-06
Techdebt: Replace sure with regular assertions in S3

kbalk:


bblommers (approved):
Thank you so much for tackling this @kbalk! This was a massive piece, so I really appreciate all the work you've put in here (and in the other PR's!)

PR #6614, opened 2023-08-07
Techdebt: Replace sure with regular assertions in sagemaker

kbalk:

Also, I noticed a test failure when running on my system that does not show up when the tests are run within the GitHub workflow.  Created a ticket for the issue:  https://github.com/getmoto/moto/issues/6615.

Lastly, my editor runs pylint and pylint is not happy with variable names less than 3 characters.  With past PRs I've made the corrections, but I stopped doing that with this PR as it can be harder to look at the changes relevant just for the PR and it's a little tedious.

### DIFF
--- a/tests/test_sagemaker/test_sagemaker_cloudformation.py
+++ b/tests/test_sagemaker/test_sagemaker_cloudformation.py
@@ -1,8 +1,8 @@
-import boto3
+import re
 
-import pytest
-import sure  # noqa # pylint: disable=unused-import
+import boto3
 from botocore.exceptions import ClientError
+import pytest
 
 from moto import mock_cloudformation, mock_sagemaker
 from moto.core import DEFAULT_ACCOUNT_ID as ACCOUNT_ID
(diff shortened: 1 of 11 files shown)


bblommers (approved):
Thanks @kbalk !

kbalk:
> We do have `pylint` as well actually, but it only runs a few checks. The config is in `setup.cfg`, if you want to setup your editor to adhere to it.

I could set up my editor to use moto's pylint config, but I don't mind fixing some pylint errors that aren't currently caught.  

> I've found (in other projects as well) that a hard rule against short variable names is a bit too strict. Yes, a variable named `x` is useless, but a variable named `s3` is perfectly clear. And I'm sure there are other examples where a 1/2 letter variable is good enough.

I don't think `s3` is always so clear.  I think `s3_client` or `s3_resource` can be more helpful, so I didn't mind making the change when I saw `s3`.

bblommers:
> I don't think s3 is always so clear. I think s3_client or s3_resource can be more helpful, so I didn't mind making the change when I saw s3.
> 

Ah, that's true. I tend to only use `s3_client` in my projects, almost never `s3_resource`, so that's less of an issue then.

PR #6626, opened 2023-08-09
awslambda: `create_function()` should fail on duplicate

rbailey-godaddy:
In real life, `lambda:CreateFunction` fails if the caller attempts to create a function that already exists. Accurately model this behavior.

Yes, the actual AWS response really is written in broken English. ;-)

### DIFF
--- a/moto/awslambda/responses.py
+++ b/moto/awslambda/responses.py
@@ -6,7 +6,7 @@ from urllib.parse import unquote
 from moto.core.utils import path_url
 from moto.utilities.aws_headers import amz_crc32, amzn_request_id
 from moto.core.responses import BaseResponse, TYPE_RESPONSE
-from .exceptions import FunctionAlreadyExists
+from .exceptions import FunctionAlreadyExists, UnknownFunctionException
 from .models import lambda_backends, LambdaBackend
 
 
@@ -320,7 +320,7 @@ class LambdaResponse(BaseResponse):
         function_name = self.json_body["FunctionName"].rsplit(":", 1)[-1]
         try:
             self.backend.get_function(function_name, None)
-        except:
+        except UnknownFunctionException:
             fn = self.backend.create_function(self.json_body)
             config = fn.get_configuration(on_create=True)
             return 201, {}, json.dumps(config)
--- a/moto/awslambda/responses.py
+++ b/moto/awslambda/responses.py
@@ -6,6 +6,7 @@ from urllib.parse import unquote
 from moto.core.utils import path_url
 from moto.utilities.aws_headers import amz_crc32, amzn_request_id
 from moto.core.responses import BaseResponse, TYPE_RESPONSE
+from .exceptions import FunctionAlreadyExists
 from .models import lambda_backends, LambdaBackend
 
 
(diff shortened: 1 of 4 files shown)


moto/awslambda/responses.py
```diff
@@ -316,9 +316,27 @@ def _list_aliases(self, function_name: str) -> TYPE_RESPONSE:
         return 200, {}, json.dumps(result)
 
     def _create_function(self) -> TYPE_RESPONSE:
-        fn = self.backend.create_function(self.json_body)
-        config = fn.get_configuration(on_create=True)
-        return 201, {}, json.dumps(config)
+        function_name = self.json_body["FunctionName"].rsplit(":", 1)[-1]
+        try:
+            self.backend.get_function(function_name, None)
```
bblommers, line 321: You can create a dedicated exception in `exceptions.py`. As long as it extends `JsonRestError`, Moto should automatically take care of the formatting.
That means that this can be simplified to

```suggestion
            self.backend.get_function(function_name, None)
            raise FunctionAlreadyExists  # or whatever name you choose
```

tests/test_awslambda/test_lambda.py
```diff
@@ -1205,19 +1205,20 @@ def test_create_function_with_already_exists():
...
+            Handler="lambda_function.lambda_handler",
+            Code={"S3Bucket": bucket_name, "S3Key": "test.zip"},
+            Description="test lambda function",
+            Timeout=3,
+            MemorySize=128,
+            Publish=True,
+        )
 
-    assert response["FunctionName"] == function_name
+        assert exc.value.response["Error"]["Code"] == "ResourceConflictException"
```
bblommers, line 1221: This should be out-dented - so outside the `pytest.raises`-block. Because the `create_function` throws an error, no other statements inside this block will be executed

bblommers (changes requested):

moto/awslambda/responses.py
rbailey-godaddy: Thanks! This is much cleaner. I have added a commit to define the exception this way.

tests/test_awslambda/test_lambda.py
rbailey-godaddy: Yeesh, I learned something here. I was thinking the `with` was establishing scope -- which it is not -- and not behaving like a `try` block, which it really is. Fixed here (and now I need to go back and look at my own code elsewhere LOL)

moto/awslambda/responses.py
rbailey-godaddy: FWIW I am not raising immediately as you suggested in order to avoid having that being trapped by the try block which is catching the expected "there's nothing here" error. Arguably it would be better to trap that condition explicitly and then raise FunctionAlreadyExists there, but I'm trying to weasel my way out of doing the research to determine the correct exception raised in this case...

bblommers (changes requested):
Looks like the build is failing, but the code LGTM so far!

bblommers (approved):
LGTM! Thank you for adding this @rbailey-godaddy

PR #6627, opened 2023-08-09
Lambda: model concurrency limits for `put_function_concurrency()`

rbailey-godaddy:
In real life, AWS lambda does not allow unbounded concurrency. Refine behavior of `put_function_concurrency()` to provide an approximation of the correct behavior:

* The AWS default limit of 1,000 concurrent lambda instances per account and region can be overridden with the `MOTO_LAMBDA_CONCURRENCY_QUOTA` environment variable, if needed;
* If cumulative reservations result in fewer than 100 unreserved instances (i.e. by default, 900 reserved instances) then an `InvalidParameterValueException` is raised and no change is made;
* The simulation applies the limit across all defined functions (rather than only functions in the same account and region) -- my judgment is that this distinction is unimportant in nearly all test scenarios (and previous behavior can be obtained by redefining the quota to some really huge number sufficient to accommodate all functions under test)

### DIFF
--- a/moto/awslambda/models.py
+++ b/moto/awslambda/models.py
@@ -2034,6 +2034,21 @@ class LambdaBackend(BaseBackend):
     def put_function_concurrency(
         self, function_name: str, reserved_concurrency: str
     ) -> str:
+        # Actual lambda restricts concurrency to 1000 (default) per region/account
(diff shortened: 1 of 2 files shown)


rbailey-godaddy:
I mangled this; closing and will reopen a replacement PR

PR #6628, opened 2023-08-09
awslambda: enforce concurrency quota in `put_function_concurrency()`

rbailey-godaddy:
In real life, AWS lambda does not allow unbounded concurrency. Refine behavior of `put_function_concurrency()` to provide an approximation of the correct behavior:

* By default, no concurrency limits are enforced (old behavior);
* If environment variable `MOTO_LAMBDA_CONCURRENCY_QUOTA` is defined (to a positive integer), then enforcement is enabled (use 1000 to match the AWS default limit);
* If cumulative reservations result in fewer than 100 unreserved instances (i.e. by default, 900 reserved instances) then an `InvalidParameterValueException` is raised and no change is made;
* The simulation applies the limit across all defined functions (rather than only functions in the same account and region) -- my judgment is that this distinction is unimportant in nearly all test scenarios (and previous behavior can be obtained by redefining the quota to some really huge number sufficient to accommodate all functions under test)

### DIFF
--- a/moto/awslambda/models.py
+++ b/moto/awslambda/models.py
@@ -2034,6 +2034,21 @@ class LambdaBackend(BaseBackend):
     def put_function_concurrency(
         self, function_name: str, reserved_concurrency: str
     ) -> str:
+        # Actual lambda restricts concurrency to 1000 (default) per region/account
+        # across all functions; we approximate that behavior by summing across all
+        # functions (hopefully all in the same account and region) and allowing the
+        # caller to simulate an increased quota.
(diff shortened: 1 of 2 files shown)


bblommers (changes requested):
The Lambda tests (and a few other services) run in parallel in ServerMode, in an effort to speed up the build. Because all tests run in parallel, multiple functions are calling `put_function_concurrency` in parallel, and the new error is raised when we don't want it.

It's a bit of a hack, but maybe we can just run these new tests in new/unused regions? As long as tests are using ReservedConcurrency in different regions it should not interfere with eachother

bblommers (approved):
LGTM - thank you for adding this feature @rbailey-godaddy!

PR #6638, opened 2023-08-12
Techdebt: Replace sure with regular assertions in SWF

kbalk:

Some of the files were DOS-formatted.  I converted them to UNIX format.


bblommers (approved):
Thanks @kbalk!

PR #6694, opened 2023-08-17
Techdebt: Replace sure with regular assertions in ResourceGroups

kbalk:


bblommers (approved):
I'll merge all these when the builds have finished

Issue #6697, opened 2023-08-18
Cognito initiate_auth USER_SRP_AUTH flow should not authorize users which the status is FORCE_CHANGE_PASSWORD


schmittjoaopedro:
As per the bug reported in this issue #4739 the same thing happen when authorizing an user via `USER_SRP_AUTH` flow instead of `USER_PASSWORD_AUTH`.

For me this issue showed up because I use `pycognito` to easy the communication with `cognito`, specifically I'm using the `pycognito` to challenge the user with a new password by calling the following method:

```python
cognito_client.new_password_challenge(request.old_password, request.new_password)
```

And this particular method calls another method `set_new_password_challenge` that happens to execute the following flow:

```python
def set_new_password_challenge(self, new_password, client=None):
        boto_client = self.client or client
        auth_params = self.get_auth_params()
        response = boto_client.initiate_auth(
            AuthFlow="USER_SRP_AUTH",
            AuthParameters=auth_params,
            ClientId=self.client_id,
        )
        if response["ChallengeName"] == self.PASSWORD_VERIFIER_CHALLENGE:
            challenge_response = self.process_challenge(
                response["ChallengeParameters"], auth_params
            )
            tokens = boto_client.respond_to_auth_challenge(
                ClientId=self.client_id,
                ChallengeName=self.PASSWORD_VERIFIER_CHALLENGE,
                ChallengeResponses=challenge_response,
            )

            if tokens["ChallengeName"] == self.NEW_PASSWORD_REQUIRED_CHALLENGE:
                challenge_parameters = response["ChallengeParameters"]
                challenge_response.update(
                    {
                        "USERNAME": challenge_parameters["USERNAME"],
                        "NEW_PASSWORD": new_password,
                    }
                )
                new_password_response = boto_client.respond_to_auth_challenge(
                    ClientId=self.client_id,
                    ChallengeName=self.NEW_PASSWORD_REQUIRED_CHALLENGE,
                    Session=tokens["Session"],
                    ChallengeResponses=challenge_response,
                )
                return new_password_response
            return tokens

        raise NotImplementedError(
            f"The {response['ChallengeName']} challenge is not supported"
        )
```

In this case, when challenging cognito moto with `PASSWORD_VERIFIER` challenge, it is ignoring that the user status is `FORCE_CHANGE_PASSWORD` and is authorizing it in, when in reality it should respond the client with a new challenge to request a new password:

[...]

PR #6698, opened 2023-08-18
CognitoIDP: user status FORCE_CHANGE_PASSWORD in USER_SRP_AUTH should require new password

schmittjoaopedro:
This PR is a proposal to fix the issue raised up at #6697

When the user initiates a USER_SRP_AUTH authentication flow, it's challenged with the password verification, in this scenario, if the user status is FORCE_CHANGE_PASSWORD, cognito should responde with a new challenge to require the new password (NEW_PASSWORD_REQUIRED) instead of authorizing the user.
Once the user provides a new password, the access token is emitted.

### DIFF
--- a/moto/cognitoidp/models.py
+++ b/moto/cognitoidp/models.py
@@ -1525,6 +1525,15 @@ class CognitoIdpBackend(BaseBackend):
                     "ChallengeParameters": {},
                 }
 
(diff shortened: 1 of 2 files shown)


bblommers (approved):
LGTM - thank you for fixing this @schmittjoaopedro!

Issue #6695, opened 2023-08-17
Moto doesn't mock out the right response for DescribeDBClusterSnapshots


klauern:
When debugging a CloudCustodian test using moto, I found that the response from `DescribeDBClusterSnapshots` isn't including all of the fields that it can, notably the `TagList` fields: https://docs.aws.amazon.com/AmazonRDS/latest/APIReference/API_DBClusterSnapshot.html

The template specifies the following: https://github.com/getmoto/moto/blob/master/moto/rds/models.py#L465-L492

But TagList is nowhere to be found.

PR #6699, opened 2023-08-18
RDS: Fix ClusterSnapshot template to include tags

dballweg:
ClusterSnapshot template is missing the TagList.
This means that describe_db_cluster_snapshots will make all the snapshots look like there are not tags.

This will solve #6695


bblommers (approved):
LGTM - thanks for the PR @dballweg!

PR #6702, opened 2023-08-20
KMS: Add `RSASSA_PSS_SHA_256` private key

tsugumi-sys:
- Expanding `generate_private_key` method to be able to return multiple private key algorithm by using abstract class.
- Add `RSASSA_PSS_SHA_256` private key.
- We can use the following unittest for this changes.
https://github.com/getmoto/moto/blob/5e1b166f2521eacd737f2ccb241fbc0b2e8bb931/tests/test_kms/test_kms_boto3.py#L1166-L1194


bblommers (approved):
LGTM! A nice way to abstract the signing logic

PR #6708, opened 2023-08-21
KMS: Support 3072 and 4098 key sizes for RSA

tsugumi-sys:
Support key sizes 3072 and 4098 key sizes for RSA private key.


bblommers (approved):
LGTM!

PR #6715, opened 2023-08-22
S3: add granting logging perms using bucket policy

pvditt:
**Context**
Related issue: https://github.com/getmoto/moto/issues/6659

This PR enables for moto s3 buckets to enable logging if they have an appropriate bucket policy set (in addition to the existing logic that enables this if an appropriate ACL is set)

**Testing**
Added to existing unit tests.

### DIFF
--- a/tests/test_s3/test_s3_logging.py
+++ b/tests/test_s3/test_s3_logging.py
@@ -3,10 +3,12 @@ import json
 import boto3
 import pytest
 from botocore.client import ClientError
+from unittest import SkipTest
 from unittest.mock import patch
 
 
 from moto import mock_s3
+from moto import settings
 from moto.core import DEFAULT_ACCOUNT_ID
 from moto.s3 import s3_backends
 from moto.s3.models import FakeBucket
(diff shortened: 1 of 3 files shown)


pvditt:
@bblommers can you please talk a look at this PR when you get the chance?

tests/test_s3/test_s3_logging.py
```diff
@@ -261,3 +268,301 @@ def test_log_file_is_created():
...
+
+
+@mock_s3
+def test_invalid_bucket_logging_when_permissions_are_false():
+    s3_client = boto3.client("s3", region_name=DEFAULT_REGION_NAME)
+    bucket_name = "mybucket"
+    log_bucket = "logbucket"
+    s3_client.create_bucket(Bucket=bucket_name)
+    s3_client.create_bucket(Bucket=log_bucket)
+    with patch(
```
bblommers, line 280: Patching like this will not work in ServerMode. You can skip these tests if you like:
```
from moto import settings
from  unittest import SkipTest
if settings.TEST_SERVER_MODE:
    raise SkipTest("Can't patch permission logic in ServerMode")
```

pvditt:
@bblommers thanks for pointing that out - I just updated those tests and have linting passing locally. Can you re-review when you get the chance?

bblommers (commented):
I should've noticed this earlier - but all of the tests are actually failing in ServerMode. 
Just because we're creating the bucket using the `boto3.client`, which creates the bucket in the MotoServer, but then we try to access the bucket using the in-memory `s3_backend` - and it doesn't exist there.

So we should probably just skip all of them.

bblommers (approved):
Awesome - thank you for fixing this @pvditt!

PR #6722, opened 2023-08-24
SQS: perform validation on all messages before attempting a send

amlodzianowski:
Follow up to https://github.com/getmoto/moto/pull/6721

I extracted the individual SQS message validation logic into a new `_validate_message` method and I am calling that from `send_message_batch` for all messages before attempting to send them via `send_message`. This replicates boto3 behavior. All tests are passing locally without changes required.

Tagging @bblommers for visibility.

### DIFF
--- a/moto/sqs/models.py
+++ b/moto/sqs/models.py
@@ -101,7 +101,6 @@ class Message(BaseModel):
 
     @property
     def attribute_md5(self) -> str:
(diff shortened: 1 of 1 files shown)


bblommers (approved):
LGTM - thank you for fixing this @amlodzianowski!

Issue #5761, opened 2022-12-12
Feature request: Add support for ECDSA_SHA_256 signing


cwkang1998:
Currently, moto ignores and hardcodes signing_algorithm to RSASSA_PSS_SHA_256.

I would love to add support for ECDSA_SHA_256 signing, for `["ECC_NIST_P256", "ECC_SECG_P256K1"]` key specs. Of course, eventually hopefully we can add support for the other `ECDSA` and `RSASSA` signing algorithms as well.

Would be trying to implement this myself.

bblommers:
Thanks for raising this @cwkang1998, and welcome to Moto! A PR would be very welcome - we have a section in our docs to get you started with contributing to Moto: 
http://docs.getmoto.org/en/latest/docs/contributing/index.html 

Please let us know if you need any help.

tsugumi-sys:
I take on this task!

tsugumi-sys:
| key spec | signing algorithm |
| :---: | :---: |
| `SYMMETRIC_DEFAULT` | |
| RSA_2048, 3072, 4096 | |
| RSA_2048, 3072, 4096 | |
| RSA_2048, 3072, 4096 | |
| RSA_2048, 3072, 4096 | |
| RSA_2048, 3072, 4096 | |
| RSA_2048, 3072, 4096 | |
| RSA_2048, 3072, 4096 | |
| RSA_2048, 3072, 4096 | |

PR #6737, opened 2023-08-28
KMS: Add signing by ECDSA Private key

tsugumi-sys:
- Adding ECDSA Private key and implementing signing and verification message.
- Adding unit tests:
  - 1. A message is successfully signed and verified.
  - 1. A message is successfully signed but verification fails.
  - 2. Signing fails becauase of an invalid signing algorithm for key spec.

### DIFF
--- a/moto/kms/models.py
+++ b/moto/kms/models.py
@@ -167,12 +167,12 @@ class Key(CloudFormationModel):
             return None  # type: ignore[return-value]
         elif self.key_spec in KeySpec.ecc_key_specs():
             if self.key_spec == KeySpec.ECC_NIST_P384:
-                return [SigningAlgorithm.ECDSA_SHA_384]
-            elif self.key_spec == KeySpec.ECC_NIST_P512:
-                return [SigningAlgorithm.ECDSA_SHA_512]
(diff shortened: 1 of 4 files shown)


bblommers (changes requested):
Looks like one test is now failing - everything else looks good.

Does this fix #5761? Or is there still some logic missing?

tsugumi-sys:
@bblommers 
Thanks! The failed test is fixed!

https://github.com/getmoto/moto/issues/5761 is almost fixed except for `SM2` algorithm.

`SM2` algorithm is not supported in [pyca/cryptography](https://cryptography.io/en/latest/)..., so I think we have two choices.

1. Finding other library (e.g. [supermemo2](https://github.com/alankan886/SuperMemo2))
2. Not supporting `SM2` in moto and use `RSA` instead of it. It's already done in the following code.

https://github.com/getmoto/moto/blob/7cf432b41203680c001f629c037175da8c13640a/moto/kms/utils.py#L248-L249

What do you think?

bblommers:
Thank you for finishing this @tsugumi-sys! 

For `SM2`, let's just default to `RSA`. If someone explicitly needs the `SM2`-algorithm, we can always change this later.

PR #6743, opened 2023-08-31
IAM policies: allow s3 accesspoint arns

### LINKED ISSUE #6742: Allow using s3 accesspoint arns in iam policies
[S3 Accesspoints](https://docs.aws.amazon.com/AmazonS3/latest/userguide/using-access-points.html) have arns which look like this:

```
        "Resource": "arn:aws:s3:us-west-2:123456789012:accesspoint/my-access-point"
```
and
```
        "Resource" : "arn:aws:s3:us-west-2:123456789012:accesspoint/my-access-point/object/*",
```

The current implementation does not allow specifying the region in s3 arns: https://github.com/getmoto/moto/blob/d3f5e3d68b0517be1172319087ee68c0834a157a/moto/iam/policy_validation.py#L61

Right now there is an MalformedPolicyDocument error thrown it should not throw an error.

thomaschaaf:
fixes #6742

### DIFF
--- a/moto/iam/policy_validation.py
+++ b/moto/iam/policy_validation.py
@@ -56,15 +56,13 @@ VALID_CONDITION_PREFIXES = ["ForAnyValue:", "ForAllValues:"]
 
 VALID_CONDITION_POSTFIXES = ["IfExists"]
 
-SERVICE_TYPE_REGION_INFORMATION_ERROR_ASSOCIATIONS = {
+SERVICE_TYPE_REGION_INFORMATION_ERROR_ASSOCIATIONS: Dict[str, Any] = {
     "iam": {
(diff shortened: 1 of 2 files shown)


bblommers (changes requested):
Hi @thomaschaaf, welcome to Moto!

Looks like the tests are failing, but the overall approach for this fix looks good so far.

thomaschaaf:
@bblommers should pass the tests now

thomaschaaf:
@bblommers  how can I run that test locally? I don't understand what is falling.

bblommers:
@thomaschaaf Looks like an intermittent failure, so don't worry about that

bblommers (approved):
LGTM - thank you @thomaschaaf!

PR #6746, opened 2023-08-31
Fix usage of a single object instances for multiple concurrent requests

### LINKED ISSUE #6745: Moto uses a single instance of the ...Response class for all requests for some services
## Motivation
In LocalStack, we began seeing some strange behavior, especially with s3 buckets, which were not created in the region or accounts the request targets. This leads to test failures, with rather confusing logs.

For example, a bucket in account 123456789012 in region us-east-1 is created twice. Since the create bucket operation is idempotent in us-east-1, it should prove without an error. However, this fails, if some other s3 operation runs at the same time under a different account.

This test can reproduce this issue (it is a test against LocalStack, but the culprit here is indeed moto):

```python
    def test_parallel_bucket_creation(self, aws_client_factory):
        num_threads = 10
        create_barrier = threading.Barrier(num_threads)
        errored = False

        def _create_bucket(runner: int):
            nonlocal errored
            bucket_name = f"bucket-{short_uid()}"
            s3_client = aws_client_factory(
                region_name="us-east-1", aws_access_key_id=f"{runner:012d}"
            ).s3
            create_barrier.wait()
            try:
                s3_client.create_bucket(Bucket=bucket_name)
                s3_client.create_bucket(Bucket=bucket_name)
            except Exception:
                LOG.exception("Create bucket failed")
                errored = True

        thread_list = []
        for i in range(1, num_threads + 1):
            thread = threading.Thread(target=_create_bucket, args=[i])
            thread.start()
            thread_list.append(thread)

        for thread in thread_list:
            thread.join()

        assert not errored
```

## The problem
After some investigation of this issue, which (before above test) was occurring only occasionally, we found the culprit.
For some services (list can be found at the end), moto uses a single `...Response` instance for all requests. So, concurrent requests will affect each others state on the `setupClass` method which is used over all response classes. There are unaffected services, which use the `dispatch` method, but a rather large part, including S3, is affected.

In S3, for example, these are the problematic lines:
https://github.com/getmoto/moto/blob/d3f5e3d68b0517be1172319087ee68c0834a157a/moto/s3/urls.py#L13-L23

They use a single instance, for all requests, which leads to the mentioned behavior.

## Impact

The impact is of course only noticeable in parallel scenarios. In our case, we found it out only in terraform tests, which perform a lot of S3 operations for validation, while our lambda provider wanted to create a bucket to store an archive in a different account.
Also, the impact is mostly region/account specific, since this is the state most often used from the `...Response` classes. However, other state (like headers) are affected too, but the impact is less noticeable.

Please note that we did not test the behavior against moto itself, but from the import path of `urls.py`, it seems like moto would be affected equally as LocalStack, if we did not miss something.

## Proposed fixes

* We have a fix in the pipeline against our moto fork (https://github.com/localstack/moto/pull/70) which I am happy to open here as well. However, we wanted to discuss first if this is the route forward. It basically uses the unbound version of a method as a parameter to a function, which calls this method on a new instance then. It was the least intrusive solution we found quickly, and the type checker should catch wrong usages of the `dispatch_method` class method. We tested this solution with above test, and it works.

* We could choose a version were we do not pass the whole method, but just a string with the method name. Would work the same, but without type checking. However, this would be a tad less "hacky".

* A refactoring of the found services to use `dispatch`. This would be more impactful, and we would basically parse the request url twice, which we might want to avoid.

If you have any more ideas for solution, we would be happy to work on them to get something merged here as well, and remove our downstream fix.

## Affected Services

These services:

```
amp
apigateway
apigatewayv2
appconfig
appsync
awslambda
cloudfront
cognitoidp
databrew
ebs
elastictranscoder
glacier
greengrass
guardduty
instance_metadata
iot
managedblockchain
mq
pinpoint
quicksight
route53
s3
s3control
eventbridgescheduler
```

use a single instance for all the requests. Unaffected services use the dispatch method. There are some services which use the classmethod dispatch on an instance, but since that does not make a difference (even though it can be confusing), we do not necessarily need to change it.

dfangl:
Fixes #6745

## Motivation
Currently, for the following services:
```
amp
apigateway
apigatewayv2
appconfig
appsync
awslambda
cloudfront
cognitoidp
databrew
ebs
elastictranscoder
glacier
greengrass
guardduty
instance_metadata
iot
managedblockchain
mq
pinpoint
quicksight
route53
s3
s3control
eventbridgescheduler
```
we use a single instance of the respective `...Response` class to handle all the requests.

In LocalStack, this leads to issues with parallel requests influencing each others state, most notably of region and account.

There are a couple of different ways to fix this, we chose one that was minimally intrusive to the existing logic, but ensured a separate instance of the Response class for every request.

Unaffected services use the `dispatch` method. There are some services which use the classmethod `dispatch` on an instance, but since that does not make a difference (even though it can be confusing), we did not change it with this PR.

## Changes
* Create `method_dispatch` class method in the BaseResponse class, which will take an unbound function as input, and execute the corresponding method on a new instance on the class. This ensures proper typing and having the type checker catch potential errors, while being able to keep the current pattern of dispatching.
* Refactor affected services to use this new function instead of calling a method on a singleton instance of the Response class.
* Some minor typing changes to avoid typing errors

### DIFF
--- a/moto/glacier/responses.py
+++ b/moto/glacier/responses.py
@@ -62,6 +62,7 @@ class GlacierResponse(BaseResponse):
     def vault_archive_response(
         self, request: Any, full_url: str, headers: Any
     ) -> TYPE_RESPONSE:
+        self.setup_class(request, full_url, headers, use_raw_body=True)
         return self._vault_archive_response(request, full_url, headers)
 
     def _vault_archive_response(
--- a/tests/test_core/test_server.py
+++ b/tests/test_core/test_server.py
@@ -43,4 +43,4 @@ def test_domain_dispatched_with_service():
     dispatcher = DomainDispatcherApplication(create_backend_app, service="s3")
     backend_app = dispatcher.get_application({"HTTP_HOST": "s3.us-east1.amazonaws.com"})
     keys = set(backend_app.view_functions.keys())
-    assert "S3Response.key_response" in keys
+    assert "moto.s3.responses.key_response" in keys
--- a/moto/amp/urls.py
+++ b/moto/amp/urls.py
@@ -6,17 +6,18 @@ url_bases = [
 ]
 
 
-response = PrometheusServiceResponse()
-
-
 url_paths = {
-    "{0}/workspaces$": response.dispatch,
-    "{0}/workspaces/(?P<workspace_id>[^/]+)$": response.dispatch,
-    "{0}/workspaces/(?P<workspace_id>[^/]+)/alias$": response.dispatch,
-    "{0}/workspaces/(?P<workspace_id>[^/]+)/logging$": response.dispatch,
-    "{0}/workspaces/(?P<workspace_id>[^/]+)/rulegroupsnamespaces$": response.dispatch,
-    "{0}/workspaces/(?P<workspace_id>[^/]+)/rulegroupsnamespaces/(?P<name>[^/]+)$": response.dispatch,
-    "{0}/tags/(?P<resource_arn>[^/]+)$": response.dispatch,
-    "{0}/tags/(?P<arn_prefix>[^/]+)/(?P<workspace_id>[^/]+)$": response.tags,
-    "{0}/tags/(?P<arn_prefix>[^/]+)/(?P<workspace_id>[^/]+)/(?P<ns_name>[^/]+)$": response.tags,
+    "{0}/workspaces$": PrometheusServiceResponse.dispatch,
+    "{0}/workspaces/(?P<workspace_id>[^/]+)$": PrometheusServiceResponse.dispatch,
+    "{0}/workspaces/(?P<workspace_id>[^/]+)/alias$": PrometheusServiceResponse.dispatch,
+    "{0}/workspaces/(?P<workspace_id>[^/]+)/logging$": PrometheusServiceResponse.dispatch,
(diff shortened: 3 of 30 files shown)


tests/test_core/test_server.py
```diff
@@ -43,4 +43,4 @@ def test_domain_dispatched_with_service():
     dispatcher = DomainDispatcherApplication(create_backend_app, service="s3")
     backend_app = dispatcher.get_application({"HTTP_HOST": "s3.us-east1.amazonaws.com"})
     keys = set(backend_app.view_functions.keys())
-    assert "S3Response.key_response" in keys
+    assert "moto.s3.responses.key_response" in keys
```
dfangl, line 46: This needed to be change since it is now a function, and not a method anymore. This test also necessitates the `functools.wrap` in the BaseResponse `method_dispatch` method.

moto/glacier/responses.py
```diff
@@ -62,6 +62,7 @@ def _vault_response_delete(self, vault_name: str, headers: Any) -> TYPE_RESPONSE
     def vault_archive_response(
         self, request: Any, full_url: str, headers: Any
     ) -> TYPE_RESPONSE:
+        self.setup_class(request, full_url, headers, use_raw_body=True)
```
dfangl, line 65: This was missing here, I assume it never occurred, since there are other calls using glacier and setting up the class before. Now it is a new one, which leads to this new line.

bblommers (approved):
Makes sense to me - thank you for the PR @dfangl!

PR #6750, opened 2023-09-01
EC2: Add missing transitGatewayArn field for CreateTransitGatewayResponse

mnitchev:
Fixes a small parity issue with AWS. Here's an link to the [examples](https://docs.aws.amazon.com/cli/latest/reference/ec2/create-transit-gateway.html#examples) and I also tried it out against actual AWS and it does indeed populate the ARN on create.

This also fixes this bug https://github.com/localstack/localstack/issues/9042 in `localstack`

### DIFF
--- a/moto/ec2/models/transit_gateway.py
+++ b/moto/ec2/models/transit_gateway.py
@@ -46,6 +46,10 @@ class TransitGateway(TaggedEC2Resource, CloudFormationModel):
     def owner_id(self) -> str:
         return self.ec2_backend.account_id
 
+    @property
+    def arn(self) -> str:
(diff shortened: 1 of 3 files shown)


bblommers (changes requested):
Hi @mnitchev, can we move the ARN as a property to the `TransitGateway` class instead? 
Just like the `owner_id` is created:
https://github.com/getmoto/moto/blob/093052bd9be65078aca9c9b57068b5648f9b5f7b/moto/ec2/models/transit_gateway.py#L46

We should also pass the region to the ARN, instead of hardcoding it, using `self.ec2_backend.region_name`

mnitchev:
@bblommers The linter failed, but it should be fine now. Can you approve the workflows again?

bblommers (approved):
LGTM - thank you for fixing this @mnitchev!

PR #6752, opened 2023-09-01
Introducing a new linter: ruff

tsugumi-sys:
A linter: ruff is introduced in this PR.
Currently, a few rules are ignored and will be fixed.

related: https://github.com/getmoto/moto/issues/6739


requirements-dev.txt
```diff
@@ -13,3 +13,4 @@ typing-extensions<=4.5.0; python_version < '3.8'
 packaging
 build
 prompt_toolkit
+ruff
```
bblommers, line 16: This isn't necessary - the second line in this file installs everything from `requirements-tests.txt`, which already contains `ruff`

bblommers (changes requested):
Let's remove `flake8` (invocation + dependency) at the same time - no point in running them both

bblommers (approved):
Awesome - thank you @tsugumi-sys!

PR #6765, opened 2023-09-04
Fix ruff error of f601

jamesdongdong:
# background
There are repeated dictionary keys in ip_protocol_keywords in `moto/ec2/models/security_groups.py`
and they are not detected because error F601 is in the ignore list.

# what i did
- delete F601 from ignore list
- delete the repeated keys


bblommers (approved):
LGTM - thank  you @jamesdongdong!

PR #6771, opened 2023-09-04
fix(secretsmanager): fix update of no-value initial secret

JoshuaLicense:
Fixes the updating of an initial no-value secret.

If a secret was initially created without a value, then subsequent updates causes a `KeyError: None` as `default_version_id` is `None`.

Noticed downstream on the latest Localstack build.


bblommers (approved):
LGTM - thanks for the fix @JoshuaLicense!

PR #6772, opened 2023-09-04
AWSLambda: Layers are now loaded onto the Docker image

### LINKED ISSUE #3779: Using lambda layers to provide missing dependencies
Hello, I don't think this one has been raised before:

I need to use the "Klayers-python37-requests" layer to be able to use the `requests` library in my lambdas.

I can fetch the layer in moto (actually download and then zip the downloaded version), and create lambdas with this layer, but if I'm not wrong it's not actually providing the actual dependency  when running the lambdas? In practice, I see `No module named 'requests'` everywhere when running the lambdas.

Request is: can we run the code provided by the layers to make the libraries available to the running code in the lambdas

bblommers:
Fixes #3779 

Layers specified in the `create_function` will now be uploaded into the Docker-volume, so that the contents will be available when executing the function.

If the layer has been created with invalid data, and the extraction/loading is unsuccessful, a warning will be logged, but execution continues.

Note that this all requires Layers to be available locally, so users will have to download them manually before executing any tests.


mcahill7:
This does not seem to work as expected:
`botocore.errorfactory.ResourceNotFoundException: An error occurred (ResourceNotFoundException) when calling the CreateFunction operation: One or more LayerVersion does not exist ['arn:aws:lambda:us-east-1:336392948345:layer:AWSSDKPandas-Python310:16']`

PR #6774, opened 2023-09-05
EC2: Support filtering transit gateways by tag

mnitchev:
Noticed transit gateways can't be filtered by tags, which is supported in AWS. Wasn't really sure how to do this, since there are several ways to do filters in the different models. Let me know if this isn't the right way to do it.


bblommers (approved):
LGTM - thank you for adding this feature @mnitchev!

PR #6775, opened 2023-09-05
fix secretsmanager PutSecretValue on empty secret

bentsku:
Hello! It seems a case has been missed with #6720, when a secret would be created empty then updated with `PutSecretValue`. This would raise a `KeyError`:
```python
  File "/opt/code/localstack/.venv/lib/python3.10/site-packages/moto/secretsmanager/models.py", line 131, in reset_default_version
    self.versions[previous_current_version_id]["version_stages"] = ["AWSPREVIOUS"]  # type: ignore
KeyError: None
```

I don't know `secretsmanager` too well, but maybe when creating an empty secret, it does not create a version? 
So I've added a check to see if there was a previous version before trying to override it as `AWSPREVIOUS`. 

Added a test case checking this specific case. 

This issue was reported in the LocalStack Slack with `cdktf`, I was able to reproduce against LocalStack with a stripped down Terraform file:
### `main.tf`
```hcl
resource "aws_secretsmanager_secret" "example" {
  name = "example"
}

resource "aws_secretsmanager_secret_version" "example-version" {
  secret_id     = aws_secretsmanager_secret.example.id
  secret_string = "example-string-to-protect"
}
```

### DIFF
--- a/moto/secretsmanager/models.py
+++ b/moto/secretsmanager/models.py
@@ -126,9 +126,10 @@ class FakeSecret:
             if "AWSPREVIOUS" in old_version["version_stages"]:
                 old_version["version_stages"].remove("AWSPREVIOUS")
 
-        # set old AWSCURRENT secret to AWSPREVIOUS
-        previous_current_version_id = self.default_version_id
-        self.versions[previous_current_version_id]["version_stages"] = ["AWSPREVIOUS"]  # type: ignore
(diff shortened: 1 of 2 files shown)


bblommers:
Hey @bentsku! Someone beat you to it - this has just been fixed in #6771 :slightly_smiling_face:

bentsku:
Ah oops, didn't check the open PR before! Thanks a lot for the fix!

PR #6778, opened 2023-09-06
S3: fix Range request handling

bentsku:
Hello,

It seems AWS has a weird way of handling HTTP `Range` request for S3. It seems it does not follow the RFC, and is instead very forgiving, failing only in certain circumstances. 

Basically, if the range request is badly formatted or not following the RFC, S3 will just treat the request as a non-range request and will return the entire object with a 200 status code.
Only certain conditions will trigger an exception and 416, mainly if the beginning of the request is above the end of the object and the request is a valid Range request (meaning it can be parsed), or if the request is a "suffix" request and it ends with 0.

I've thoroughly tested those changes against AWS, and it's been reported by one of our users.

I've updated the code and added the tests for most of the weird cases I could think of.
If you feel like the logic could be handled better in `_handle_range_header`, I'm all ears, I just tried to return early and return a regular response if AWS would do it too.

### DIFF
--- a/moto/s3/responses.py
+++ b/moto/s3/responses.py
@@ -1169,25 +1169,35 @@ class S3Response(BaseResponse):
     ) -> TYPE_RESPONSE:
         length = len(response_content)
         last = length - 1
+
         _, rspec = request.headers.get("range").split("=")
         if "," in rspec:
-            raise NotImplementedError("Multiple range specifiers not supported")
+            return 200, response_headers, response_content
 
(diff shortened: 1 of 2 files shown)


bentsku:
Seems like the test failure might be unrelated?

bblommers:
> Seems like the test failure might be unrelated?

Yup, unrelated - Docker is just a little fickle

bblommers (approved):
LGTM - thanks for fixing this @bentsku! Good to have this aligned with AWS.

PR #6783, opened 2023-09-07
shelve based persistence

morsecodist:
Our team uses moto server as a local version of AWS for local development environments, typically with docker-compose. This allows us to put persistence files into a mount so we can start and stop our development environment restoring state. This is a feature localstack has but we prefer to use a more open source tool if we can. I know this isn't the primary use case for your package but this adds a basic form of persistence in an unobtrusive, optional way. It may theoretically be helpful to speed up tests in CICD as well.

The persistence is strictly opt-in via an environment variable and nothing should change for people who don't want it. I have added a test to ensure that persistence has no effect by default, as well as other test cases. The overall change is extremely small and should not incur a huge maintenance burden. I am using shelve which uses pickle so theoretically there could be some issues with serializing and deserializing certain items but this should only matter for those using persistence. Also, I am using pickle to merge properties into classes instead of serializing the classes themselves.


morsecodist:
Oh thanks @bblommers couldn't find that when looking into things. Thank you so much for pointing it out. Will close.

PR #6789, opened 2023-09-08
Batch: Using enum for job status

tsugumi-sys:
- Adding Enum for Batch job status and some utility methods.
- Adding unit tests for the methods. 
- Replacing some codes where job status is hard coded.


bblommers (approved):
Nice one - thanks @tsugumi-sys!

PR #6793, opened 2023-09-08
STS: verify that a role exists to allow sts:AssumeRole

ajoga:
Hey,

I started this PR by writing the test case, as I found odd that it was possible to `assume_role()` with a non-existing IAM role.

Then I tried to fix it, but:

- it's becoming to feel like ["by design" that this check does not exist to begin with](https://github.com/ajoga/moto/blob/4e81d4e29bb75b5014691a8074c89cab7c083687/moto/sts/models.py#L101)
- Given that assuming a role that does not exist yields an AccessDenied error, I wonder if the logic that check that the role can be assume should be instead in moto.iam.access_control.IAMRequestBase.check_action_permitted() ?

I'm pushing this code now as a draft to open the discussion, can you help me contributing to this or let me know if this does not make sense?

### DIFF
--- a/moto/core/exceptions.py
+++ b/moto/core/exceptions.py
@@ -148,6 +148,14 @@ class AccessDeniedError(RESTError):
             "AccessDenied", f"User: {user_arn} is not authorized to perform: {action}"
         )
 
+class AccessDeniedOnResourceError(RESTError):
+    code = 403
+
+    def __init__(self, user_arn: str, action: str, resource: str):
+        super().__init__(
+            "AccessDenied", f"User: {user_arn} is not authorized to perform: {action} on resource: {resource}"
+        )
+
 
 class AuthFailureError(RESTError):
     code = 401
--- a/moto/sts/models.py
+++ b/moto/sts/models.py
@@ -5,6 +5,7 @@ import re
 import xmltodict
 
 from moto.core import BaseBackend, BaseModel, BackendDict
+from moto.core.exceptions import AccessDeniedOnResourceError
 from moto.core.utils import iso_8601_datetime_with_milliseconds
 from moto.iam.models import iam_backends, AccessKey
 from moto.sts.utils import (
@@ -94,10 +95,19 @@ class STSBackend(BaseBackend):
         policy: str,
         duration: int,
         external_id: str,
+        access_key_id: str
     ) -> AssumedRole:
         """
         Assume an IAM Role. Note that the role does not need to exist. The ARN can point to another account, providing an opportunity to switch accounts.
         """
+        caller = self.get_caller_identity(access_key_id)
+        backend = iam_backends[self.account_id]["global"]
+        role_exists = False
+        for role in backend.roles.values():
+            if role.arn == role_arn:
+                role_exists = True
+        if not role_exists:
+            raise AccessDeniedOnResourceError(user_arn=caller[1], action="sts:AssumeRole", resource=role_arn)
         account_id, access_key = self._create_access_key(role=role_arn)
         role = AssumedRole(
             account_id,
--- a/moto/sts/responses.py
+++ b/moto/sts/responses.py
@@ -43,6 +43,7 @@ class TokenResponse(BaseResponse):
         policy = self.querystring.get("Policy", [None])[0]
         duration = int(self.querystring.get("DurationSeconds", [3600])[0])
         external_id = self.querystring.get("ExternalId", [None])[0]
+        access_key_id = self.get_access_key()
 
         role = self.backend.assume_role(
             role_session_name=role_session_name,
@@ -50,6 +51,7 @@ class TokenResponse(BaseResponse):
             policy=policy,
             duration=duration,
             external_id=external_id,
+            access_key_id=access_key_id
         )
         template = self.response_template(ASSUME_ROLE_RESPONSE)
         return template.render(role=role)
--- a/tests/test_core/test_auth.py
+++ b/tests/test_core/test_auth.py
@@ -632,6 +632,35 @@ def test_access_denied_with_temporary_credentials():
         == f"User: arn:aws:sts::{ACCOUNT_ID}:assumed-role/{role_name}/{session_name} is not authorized to perform: rds:CreateDBInstance"
     )
 
+@set_initial_no_auth_action_count(3)
+@mock_sts
+def test_access_denied_for_non_existing_role():
+    user_name = "new-test-user"
+    role_name = "test-role"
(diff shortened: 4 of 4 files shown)


bblommers (commented):
Hi @ajoga! This is indeed by design.

When running Moto in ServerMode, the only way to create resources in a different account is by using a IAM Role from that account. So if we introduce a check that the `role` must exist, we have a chicken and egg problem: we need to create a `role` to create (say) an S3 bucket, but to create the `role` in a different account, we need to create a `role` to give us access, .. etc.

Allowing any role-name is the easiest, most straightforward way to get around that conundrum.

ajoga:
Thank you for the feedback @bblommers .

Since an sts:AssumeRole on a non-existing role yields an AccessDenied, I see an alternative way to test a code path for a non-existing role to use a principal lacking permissions to assume such role 

I managed to get something like this to work:

```python
@set_initial_no_auth_action_count(3)
@mock_sts
def test_access_denied_with_not_allowing_policy():
    user_name = "test-user"
    forbidden_role_arn = f"arn:aws:iam::{ACCOUNT_ID}:role/forbidden"
    inline_policy_document = {
        "Version": "2012-10-17",
        "Statement": [
            {"Effect": "Deny", "Action": ["sts:AssumeRole"], "Resource": forbidden_role_arn}
            ],
    }
    access_key = create_user_with_access_key_and_inline_policy(
        user_name, inline_policy_document
    )
    client = boto3.client(
        "sts",
        region_name="us-east-1",
        aws_access_key_id=access_key["AccessKeyId"],
        aws_secret_access_key=access_key["SecretAccessKey"],
    )
        client.assume_role(
            RoleArn=forbidden_role_arn,
            RoleSessionName='azerr'
        )
    assert ex.value.response["Error"]["Code"] == "AccessDenied"
    assert ex.value.response["ResponseMetadata"]["HTTPStatusCode"] == 403
    assert (
        ex.value.response["Error"]["Message"]
        == f"User: arn:aws:iam::{ACCOUNT_ID}:user/{user_name} is not authorized to perform: sts:AssumeRole"
    )

```
(very inspired from https://github.com/getmoto/moto/blob/fa9aa95156d1e11a3c5a2e06ab8b085ae98c2c1f/tests/test_core/test_auth.py#L269)

But when I tried to set the inline_policy_document to something more "realistic" for the test, ie with broader permissions but a deny on that specific role, my assume_role that should work is denied as well:

```python
@set_initial_no_auth_action_count(3)
@mock_sts
def test_access_denied_with_not_allowing_policy():
    user_name = "test-user"
    forbidden_role_arn = f"arn:aws:iam::{ACCOUNT_ID}:role/forbidden"
    allowed_role_arn = f"arn:aws:iam::{ACCOUNT_ID}:role/allowed"
    inline_policy_document = {
        "Version": "2012-10-17",
        "Statement": [
            {"Effect": "Deny", "Action": ["sts:AssumeRole"], "Resource": forbidden_role_arn},
            {"Effect": "Allow", "Action": ["sts:AssumeRole"], "Resource": "*"}
            ],
    }
    access_key = create_user_with_access_key_and_inline_policy(
        user_name, inline_policy_document
    )
    client = boto3.client(
        "sts",
        region_name="us-east-1",
        aws_access_key_id=access_key["AccessKeyId"],
        aws_secret_access_key=access_key["SecretAccessKey"],
    )
        client.assume_role(
            RoleArn=forbidden_role_arn,
            RoleSessionName='azerr'
        )
    assert ex.value.response["Error"]["Code"] == "AccessDenied"
    assert ex.value.response["ResponseMetadata"]["HTTPStatusCode"] == 403
    assert (
        ex.value.response["Error"]["Message"]
        == f"User: arn:aws:iam::{ACCOUNT_ID}:user/{user_name} is not authorized to perform: sts:AssumeRole"
    )
    client.assume_role(
        RoleArn=allowed_role_arn,
        RoleSessionName='azerr'
    )
```
> ./tests/test_core/test_auth.py::test_access_denied_with_not_allowing_policy Failed: [undefined]botocore.exceptions.ClientError: An error occurred (AccessDenied) when calling the AssumeRole operation: User: arn:aws:iam::123456789012:user/test-user is not authorized to perform: sts:AssumeRole

Reading this line shows it's a known limitation: https://github.com/getmoto/moto/blob/bef70a47d93f10d6371bcc5edf7877cce0083cf2/moto/iam/access_control.py#L6

From debugging I noticed that the above `client.assume_role` call goes up to there an the `resource` attribute is not set properly: https://github.com/getmoto/moto/blob/c8b5470e25012477f817a619e88f9ad1bea08bbe/moto/core/responses.py#L131

I tried understanding how [this PR for S3](https://github.com/getmoto/moto/pull/6637/files#diff-4f4b29ffb4a51b743c9757969944ca9b98dc74195b63f97682bfa7a704310657) got the resource policy checks implemented, but the call stack for s3 functions are going through the `s3/reponses.py` whereas the assume_role call stack goes though `core/responses.py`: I am a bit at loss as what/where/how to implement the missing logic for the `resource` parameter to be passed to moto.core.responses.ActionAuthenticatorMixin._authenticate_and_authorize_action()

I tried to see if I could contribute something in this regard but I may try to chew more than I can bite ; what do you think? Could you give me some pointers?

bblommers:
All S3-methods() call `_authenticate_and_authorize_s3_action`, which is also defined in `moto/core/responses.py`. 
Other services call `_authenticate_and_authorize_normal_action` - but both versions end up calling `_authenticate_and_authorize_action()`, just using different parameters.

> what/where/how to implement the missing logic for the resource parameter to be passed to moto.core.responses.ActionAuthenticatorMixin._authenticate_and_authorize_action()
>

IMO, the logical place would be to determine the `resource` when calling `_authenticate_and_authorize_normal_action`:

https://github.com/getmoto/moto/blob/fa9aa95156d1e11a3c5a2e06ab8b085ae98c2c1f/moto/core/responses.py#L496

The problem then becomes of course determining what the value for `resource` is, based on just the request (headers/body) - it's not immediately obvious to me how to do that cleanly...

Maybe we can delegate this to the service-specific backends? So something like `STSResponse._determine_resource(request)`.

The action flow in `call_action` then becomes:
```
resource = "*"
if hasattr(self, '_determine_resource'):
     resource = self._determine_resource(request)
self._authenticate_and_authorize_normal_action(resource)
```

Open to suggestions here, maybe (hopefully!) I'm missing something obvious.

ajoga:
Closing this PR as it didn't make sense to begin with, continuing the other discussion in  #6799

PR #6796, opened 2023-09-10
Elastic Beanstalk: Add the `delete_application` method support

jfmainville:
Add the `delete_application` method support to the Elastic Beanstalk resource. This PR contains the base configuration for this method.


jfmainville:
Need to rebase the branch to only include the required commits.

PR #6797, opened 2023-09-10
Elastic Beanstalk: Add the `delete_application` method support

jfmainville:
Add the `delete_application` method support for the Elastic Beanstalk resource. This PR contains the base configuration required to support this action.


moto/elasticbeanstalk/models.py
```diff
@@ -148,5 +159,19 @@ def _find_environment_by_arn(self, arn: str) -> FakeEnvironment:
                     return env
         raise KeyError()
 
+    def delete_application(
+        self, application_name: str, terminate_env_by_force: bool
+    ) -> FakeApplication:
+        if application_name:
+            if application_name in self.applications:
+                application = self.applications[application_name]
+                application.status = "deleting"
```
bblommers, line 168: What's the reasoning behind setting the status, rather then just removing the application altogether?

The `describe_applications` call doesn't seem to return a status, so a user would not be able to tell the difference between an active application, and an application that was 'deleted' in the background.

How does this work in AWS - is the application still returned in `describe_applications` for another x minutes, before AWS actually deletes it?

moto/elasticbeanstalk/models.py
```diff
@@ -42,12 +47,18 @@ def region(self) -> str:
...
+        backend: "EBBackend",
+        application_name: str,
+    ):
         self.backend = weakref.proxy(backend)  # weakref to break cycles
         self.application_name = application_name
         self.environments: Dict[str, FakeEnvironment] = dict()
         self.account_id = self.backend.account_id
         self.region = self.backend.region_name
+        self.status = ""
+        self.terminate_env_by_force = False
```
bblommers, line 61: Let's remove this variable - it doesn't have any impact on how users interact with Moto, so there's no point in keeping it.

moto/elasticbeanstalk/models.py
```diff
@@ -148,5 +159,19 @@ def _find_environment_by_arn(self, arn: str) -> FakeEnvironment:
...
+        if application_name:
+            if application_name in self.applications:
+                application = self.applications[application_name]
+                application.status = "deleting"
+                if terminate_env_by_force:
+                    application.terminate_env_by_force = terminate_env_by_force
+                else:
+                    application.terminate_env_by_force = False
+                return application
+        raise ApplicationNotFound(application_name)
```
bblommers, line 174: Would you mind adding a test for this scenario? Just to verify the error handling works correctly.

http://docs.getmoto.org/en/latest/docs/contributing/development_tips/tests.html#negative-tests

bblommers (changes requested):

moto/elasticbeanstalk/models.py
jfmainville: Thanks for the feedback! You're absolutely right, the `status` field shouldn't be included, I thought it was always included in all the resources with different naming conventions, but for the Elastic Beanstalk nothing is returned. I removed this field from the model.

moto/elasticbeanstalk/models.py
jfmainville: The variable `status` was removed.

moto/elasticbeanstalk/models.py
jfmainville: That's right, I added the `test_delete_unknown_application` test to validate that we indeed get receive the `ApplicatioNotFound` error.

```python
@mock_elasticbeanstalk
def test_delete_unknown_application():
    conn = boto3.client("elasticbeanstalk", region_name="us-east-1")

    application_name = "myapp"
    unknown_application_name = "myapp1"

    conn.create_application(ApplicationName=application_name)
        conn.delete_application(ApplicationName=unknown_application_name)
    assert err["Code"] == "ApplicationNotFound"
    assert (
        err["Message"]
        == f"Elastic Beanstalk application {unknown_application_name} not found."
    )
```

moto/elasticbeanstalk/models.py
```diff
@@ -148,5 +158,18 @@ def _find_environment_by_arn(self, arn: str) -> FakeEnvironment:
                     return env
         raise KeyError()
 
+    def delete_application(
+        self, application_name: str, terminate_env_by_force: bool
+    ) -> FakeApplication:
+        if application_name:
+            if application_name in self.applications:
+                application = self.applications[application_name]
```
bblommers, line 166: I've just tested this out against AWS - they just completely remove the application on deletion. So let's do the same:
`self.applications.pop(application_name)`

The response does not contain any information, at all, so we also don't have to return anything in this method.

jfmainville: Thanks for the validation! Also, it's the first time that I have to handle a function that doesn't return anything in Moto, I made the following adjustment to the `delete_application` method in the model, is it compliant with the in-place structure for these scenarios: 

model.py
```python
    def delete_application(
        self,
        application_name: str,
    ) -> None:
        if application_name:
            if application_name in self.applications:
                self.applications.pop(application_name)
            else:
                raise ApplicationNotFound(application_name)
```
responses.py
```python
    def delete_application(self) -> str:
        application_name = self._get_param("ApplicationName")
        self.elasticbeanstalk_backend.delete_application(
            application_name=application_name,
        )
        return DELETE_APPLICATION_TEMPLATE
```

Also, I removed the `TerminateEnvByForce` parameter from the model and response as it doesn't have any impact in this scenario as nothing is returned.

bblommers (approved):
LGTM - thank you for adding this feature @jfmainville!

PR #6799, opened 2023-09-10
IAM validation: Resource is now processed during authorization

ajoga:
This PR uses the validation logic to validate the resource attribute of the principal-attached policies during `sts:AssumeRole`

It relates to the discussion in https://github.com/getmoto/moto/pull/6793#issuecomment-1712505836

What do you think of this draft @bblommers ? I implemented following your directions. I don't have enough experience to make an alternative proposition, and what you offered makes sense. It feels like it's going to be rather intense to implement each map of (service, method) to the corresponding resource attribute, but I don't see an alternative. 

If it looks good for you, I'll include in this PR the logic in TokenResponse._determine_resource() for every actions already supported for STS before submitting it as PR.

It's breaking some tests and there's one I've looked into I don't get:

> FAILED tests/test_s3/test_s3_bucket_policy.py::TestBucketPolicy::test_block_or_allow_get_object[kwargs2-403] - Failed: DID NOT RAISE <class 'botocore.exceptions.ClientError'>

On the previous commit that works, with a breakpoint at this line, I don't understand how the exception raises correctly ? https://github.com/ajoga/moto/blob/fa9aa95156d1e11a3c5a2e06ab8b085ae98c2c1f/tests/test_s3/test_s3_bucket_policy.py#L68 

When I compare my commit that breaks the test and the previous one that does not, I see no difference.

Also, do you see other tests cases needed?

I'll close the initial PR

### DIFF
--- a/moto/core/responses.py
+++ b/moto/core/responses.py
@@ -157,10 +157,10 @@ class ActionAuthenticatorMixin(object):
         else:
             ActionAuthenticatorMixin.request_count += 1
 
-    def _authenticate_and_authorize_normal_action(self) -> None:
+    def _authenticate_and_authorize_normal_action(self, resource: str = "*") -> None:
         from moto.iam.access_control import IAMRequest
 
-        self._authenticate_and_authorize_action(IAMRequest)
+        self._authenticate_and_authorize_action(IAMRequest, resource)
 
     def _authenticate_and_authorize_s3_action(
         self, bucket_name: Optional[str] = None, key_name: Optional[str] = None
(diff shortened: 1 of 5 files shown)


bblommers:
>  It feels like it's going to be rather intense to implement each map of (service, method)
>

Intense is the right word, I think! But that's the beauty of this approach - it can be done on a service by service basis, only if/when people need it.

> It's breaking some tests:
>

Took me a while to get my head around.. but I think the `TestBucketPolicy` test is just wrong. The Moto approach to bucket policies is: [allow, unless there is an explicit deny](https://github.com/getmoto/moto/blob/fa9aa95156d1e11a3c5a2e06ab8b085ae98c2c1f/moto/s3/responses.py#L1299).
But the failing test cases all expect a failure when we allow access to different resources.

So these test cases can just be removed IMO.

> Also, do you see other tests cases needed?
>
The new test in this PR look good. But we should extend this to also test the other methods, if you do extend the `_determine_resource`-logic to include other actions.

PR #6808, opened 2023-09-13
ELBv2: added validation for target group

macnev2013:
# Motivation
- This PR aims to make target group in parity with AWS.

# Added
- Parameter validation for `create_target_group`
- Default values for `create_target_group`
- Adjusted test cases accordingly
- Fixed `create_target_group` response

# Notes
- Related PR for AWS Validated Test Case: https://github.com/localstack/localstack-ext/pull/2064

### DIFF
--- a/moto/elbv2/models.py
+++ b/moto/elbv2/models.py
@@ -1093,13 +1093,13 @@ Member must satisfy regular expression pattern: {expression}"
                     param = "VPC ID" if param == "vpc_id" else param.lower()
                     raise ValidationError(f"A {param} must be specified")
 
-        if vpc_id := kwargs.get("vpc_id"):
+        if kwargs.get("vpc_id"):
             from moto.ec2.exceptions import InvalidVPCIdError
 
             try:
-                self.ec2_backend.get_vpc(vpc_id)
+                self.ec2_backend.get_vpc(kwargs.get("vpc_id"))
             except InvalidVPCIdError:
-                raise ValidationError(f"The VPC ID '{vpc_id}' is not found")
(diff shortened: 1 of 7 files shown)


moto/elbv2/models.py
```diff
@@ -1080,6 +1087,90 @@ def create_target_group(self, name: str, **kwargs: Any) -> FakeTargetGroup:
                 "HttpCode must be like 200 | 200-399 | 200,201 ...",
             )
 
+        if target_type in ("instance", "ip", "alb"):
+            for param in ("protocol", "port", "vpc_id"):
+                if not kwargs.get(param):
+                    param = "VPC ID" if param == "vpc_id" else param.lower()
+                    raise ValidationError(f"A {param} must be specified")
+
+        if vpc_id := kwargs.get("vpc_id"):
```
bblommers, line 1096: We're still supporting Python 3.7, so the `:=` operator will not work unfortunately

macnev2013: Updated

moto/elbv2/models.py
```diff
@@ -1080,6 +1091,124 @@ def create_target_group(self, name: str, **kwargs: Any) -> FakeTargetGroup:
...
+            try:
+                self.ec2_backend.get_vpc(kwargs.get("vpc_id"))
+            except InvalidVPCIdError:
+                raise ValidationError(
+                    f"The VPC ID '{kwargs.get('vpc_id')}' is not found"
+                )
+
+        kwargs_patch = {}
+
+        conditions: Dict[str, Any] = {
```
bblommers, line 1112: I like this - makes it very easy to see all defaults in one place

bblommers (approved):
LGTM - thank you @macnev2013!

PR #6812, opened 2023-09-14
ELBv2: fix parity in describe_target_group

macnev2013:
# Motivation
This PR fixes minor parity gaps with aws in `describe_target_groups` api

# Fixed
- sorting the list of target group before returning it
- multiple arguments in describe call raises an exception

# Added
- Added test cases for `describe_target_groups` api


bblommers (approved):
LGTM!

PR #6817, opened 2023-09-15
ELBv2: added validation for create_target_group() and set_ip_address_type()

macnev2013:
# Fixed
- `create_target_group`: parameter validation added for `TargetType == 'lambda'`
- `set_ip_address_type `:  parameter validation added for ip address type

# Test
- Adjusted test case and added new scenarios


bblommers (approved):
LGTM - thank you!

PR #6824, opened 2023-09-18
Opensearch: Add the `list_domain_names` method support

jfmainville:
Add the Opensearch `list_domain_names` method support. The pagination functionality hasn't been implemented yet, just the base configuration for listing all the domain names for Opensearch.

### DIFF
--- a/docs/docs/services/opensearch.rst
+++ b/docs/docs/services/opensearch.rst
@@ -61,7 +61,7 @@ opensearch
 - [ ] get_package_version_history
 - [ ] get_upgrade_history
 - [ ] get_upgrade_status
-- [ ] list_domain_names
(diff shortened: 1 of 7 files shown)


bblommers:
Hi @jfmainville , hope you don't mind - I pushed a small fix to ensure that we can use `list_domain_names()` in ServerMode.

The ES and OpenSearch use the same endpoint, `es.amazonaws.com` - that's why we need to tell the ES-service to redirect to OpenSearch in some cases.
The differentiator is the version, so ES requests are pointed to `/2015-01-01/domain`, Opensearch uses `/2021-01-01/domain`.

bblommers (approved):
Thanks for the PR @jfmainville!

PR #6826, opened 2023-09-18
EC2: Cross-account VPC peering connections

viren-nadkarni:
AWS allows VPCs to be peered as long as their CIDR ranges do not overlap. Peering can happen within the same account/region or cross-account/region.

https://github.com/getmoto/moto/pull/2195 added VPC peering support across-regions.

This PR introduces VPC peering support across-accounts.

Exisiting tests have been parametrised to run all assertions within the same account and across accounts. 

Closes: https://github.com/localstack/localstack/issues/8550

### DIFF
--- a/moto/ec2/exceptions.py
+++ b/moto/ec2/exceptions.py
@@ -632,21 +632,23 @@ class InvalidVPCRangeError(EC2ClientError):
         super().__init__("InvalidVpc.Range", f"The CIDR '{cidr_block}' is invalid.")
 
 
-# accept exception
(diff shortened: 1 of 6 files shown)


bblommers (changes requested):
Looks like the linter is failing: https://github.com/getmoto/moto/actions/runs/6224588248/job/16893138639

All tests that now patch `os.environ` will have to be skipped in ServerMode, otherwise they'll fail.

The logic itself looks good to me!

bblommers (approved):
Awesome - thank you @viren-nadkarni!

PR #6839, opened 2023-09-21
Fix tag specifications crash for EC2 resources

RaJiska:
It seems like moto mishandles tags for a some EC2 resources with the error `AttributeError: 'list' object has no attribute 'get'`.

Commands to reproduce:
```
awslocal ec2 create-vpc --cidr-block 10.0.0.0/16
awslocal ec2 create-subnet --vpc-id "$VPC_ID" --cidr-block 10.0.20.0/20 --tag-specifications '{"Tags":[{"Key":"Name","Value":"ok"},{"Key":"Hello","Value":"World"}]}'
```

I could replicate this error for VPC, Subnets, and Route Tables and just fixed the other EC2 resources which had the same faulty pattern.

### DIFF
--- a/moto/ec2/responses/amis.py
+++ b/moto/ec2/responses/amis.py
@@ -7,7 +7,9 @@ class AmisResponse(EC2BaseResponse):
         name = self.querystring.get("Name")[0]  # type: ignore[index]
         description = self._get_param("Description", if_none="")
         instance_id = self._get_param("InstanceId")
-        tag_specifications = self._get_multi_param("TagSpecification", skip_result_conversion=True)
+        tag_specifications = self._get_multi_param(
+            "TagSpecification", skip_result_conversion=True
+        )
 
         self.error_on_dryrun()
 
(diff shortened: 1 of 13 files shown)


RaJiska:
@bblommers Fixed linting

RaJiska:
This last failing job, is it a random fail? It doesn't look related to the changes.

bblommers:
Hi @RaJiska!

From the [docs](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/ec2/client/create_subnet.html):

> The tags to apply to a resource when the resource is being created. When you specify a tag, you must specify the resource type to tag, otherwise the request will fail.
>

When you _do_ specify `ResourceType=subnet`, it works as expected.

bblommers:
AWS throws this error:

> An error occurred (InvalidParameter) when calling the CreateSubnet operation: Tag specification resource type must have a value
>

bblommers:
Validation is now part of Moto thanks to https://github.com/getmoto/moto/pull/6860, so I'll close this.

PR #6851, opened 2023-09-26
EC2: Cross-account transit gateway peering attachments

viren-nadkarni:
This PR adds cross-region and cross-account peering attachments for transit gateways.

<https://docs.aws.amazon.com/vpc/latest/tgw/tgw-peering.html>

It takes an approach similar to VPC peering. See: https://github.com/getmoto/moto/pull/6826


bblommers (approved):

I'm sure there are plenty of areas where this is still missing, but it's nice to see the multi-account usage increase across the board.

PR #6858, opened 2023-09-27
IoT: Set Certificate ID to SHA256 hexdigest of DER

patstrom:
AWS IoT sets to certificate id using the sha256 of the certificate DER representation, not the PEM representation.

### DIFF
--- a/moto/iot/models.py
+++ b/moto/iot/models.py
@@ -176,9 +176,11 @@ class FakeCertificate(BaseModel):
         region_name: str,
         ca_certificate_id: Optional[str] = None,
     ):
-        m = hashlib.sha256()
-        m.update(certificate_pem.encode("utf-8"))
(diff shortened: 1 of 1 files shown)


patstrom:
It seems the tests are failing since FakeCertificate is instantiated using invalid PEM. This works before this patch since it doesn't actually check for anything, but fails in this patch since it tries to convert the PEM to DER.

I unfortunately do not have time to update the tests. Feel free to close this PR if no one else is keen.

bblommers:
Hi @patstrom, I wouldn't be surprised if some users also pass an invalid certificate, so it may be for the better to keep our tests as is.

If this is something you need, you can always make it an if/else - use DER representation where possible, fall back to the current algorithm otherwise. But if you don't have the time currently, I'll just close it for now.

PR #6860, opened 2023-09-27
EC2: Add validations for create_subnet tags

pinzon:
Wrong inputs are common among users and the crashing of the service instead of returning the correct error message confuse them. This PR adds some extra validations for the situation encountered in localstack/localstack#9195

### DIFF
--- a/moto/ec2/models/subnets.py
+++ b/moto/ec2/models/subnets.py
@@ -13,7 +13,6 @@ from ..exceptions import (
     InvalidAvailabilityZoneError,
     InvalidCIDRBlockParameterError,
     InvalidParameterValueError,
-    InvalidParameter,
     InvalidSubnetConflictError,
     InvalidSubnetIdError,
     InvalidSubnetRangeError,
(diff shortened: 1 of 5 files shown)


moto/ec2/responses/subnets.py
```diff
@@ -11,9 +11,7 @@ def create_subnet(self) -> str:
         ipv6_cidr_block = self._get_param("Ipv6CidrBlock")
         availability_zone = self._get_param("AvailabilityZone")
         availability_zone_id = self._get_param("AvailabilityZoneId")
-        tags = self._get_multi_param("TagSpecification")
-        if tags:
-            tags = tags[0].get("Tag")
+        tags = self._get_multi_param("TagSpecification", skip_result_conversion=True)
```
bblommers, line 14:
```suggestion
        tags = self._parse_tag_specification()
```

Tag-handling is.. how shall I put this delicately.. a bit of a mess. Across all services.

But for EC2, I think that the `parse_tag_specification` is the best approach. 

If you wouldn't mind refactoring that method to add this validation functionality about missing/incorrect `ResourceType`s, it will be really easy to extend this behaviour to other resource-types in the future.

bblommers (changes requested):

bblommers (changes requested):
Looks like some of the tests are still failing @pinzon: https://github.com/getmoto/moto/actions/runs/6341641228/job/17227852825?pr=6860

(Github had a lot of issues yesterday with the CI, so almost every job failed - but these failures look genuine..)

The change itself looks good!

moto/ec2/models/subnets.py
```diff
@@ -317,8 +317,9 @@ def create_subnet(
         ipv6_cidr_block: Optional[str] = None,
         availability_zone: Optional[str] = None,
         availability_zone_id: Optional[str] = None,
-        tags: Optional[List[Dict[str, str]]] = None,
+        tags: Optional[dict[str, dict[str, str]]] = None,
```
bblommers, line 320: I think these should be uppercase:

```suggestion
        tags: Optional[Dict[str, Dict[str, str]]] = None,
```

https://github.com/getmoto/moto/actions/runs/6384917966/job/17330933479?pr=6860

bblommers (approved):
Awesome, thanks @pinzon!

PR #6866, opened 2023-09-29
cognito-idp: Use default values for unspecified settings in update_user_pool

kmdmnak:
Hi!

In this PR, I've fixed the behavior of `update_user_pool`.

I discovered that `update_user_pool` overwrites default values even when the request doesn't contain those parameters.

However, the [AWS docs](https://docs.aws.amazon.com/cognito-user-identity-pools/latest/APIReference/API_UpdateUserPool.html) state:

>If you don't provide a value for an attribute, Amazon Cognito sets it to its default value.

Therefore, I've added a method to extend the default config, ensuring that UserPool uses it during both creation and configuration updates.

Best regards.

### DIFF
--- a/moto/cognitoidp/models.py
+++ b/moto/cognitoidp/models.py
@@ -399,22 +399,7 @@ class CognitoIdpUserPool(BaseModel):
         self.name = name
         self.status = None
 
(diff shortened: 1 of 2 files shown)


bblommers (approved):
LGTM - thank you for fixing this @kmdmnak!

PR #6882, opened 2023-10-04
ELBv2: fixed exception message for target group

macnev2013:
## Motivation
This Pull Request introduces a test case that validates the behavior of health check parameters for target groups.

## Changes
- When the protocol is set to HTTP or HTTPS and HealthCheckIntervalSeconds is equal to HealthCheckTimeoutSeconds, AWS generates an exception (this has been addressed in Moto).
- Test cases have been added to verify this behavior.


bblommers (approved):
LGTM - thanks @macnev2013!

PR #6886, opened 2023-10-05
ELBv2: target-group creation and modification

macnev2013:
# Fixes
-  `modify-target-group-attributes` : fixed validation for `stickiness` and `deregistration_delay.timeout_seconds` and added exception handling
- added new exception `InvalidConfigurationRequest`
- added new tags in target-group `creation` and `describe` response
- added test cases

### DIFF
--- a/tests/terraformtests/bin/run_go_test
+++ b/tests/terraformtests/bin/run_go_test
@@ -20,8 +20,8 @@ PATCH="etc/0001-Patch-Hardcode-endpoints-to-local-server.patch"
 cd terraform-provider-aws || exit
 echo "Running tests $2 for service $1..."
 TF_ACC=1 \
-  AWS_ACCESS_KEY_ID=test AWS_SECRET_ACCESS_KEY=test AWS_DEFAULT_REGION=us-west-1 \
-  AWS_ALTERNATE_ACCESS_KEY_ID=test AWS_ALTERNATE_SECRET_ACCESS_KEY=test AWS_ALTERNATE_REGION=us-east-2 \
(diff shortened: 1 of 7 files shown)


tests/terraformtests/bin/run_go_test
```diff
@@ -19,5 +19,9 @@ PATCH="etc/0001-Patch-Hardcode-endpoints-to-local-server.patch"
 (
 cd terraform-provider-aws || exit
 echo "Running tests $2 for service $1..."
-AWS_ACCESS_KEY_ID=test AWS_SECRET_ACCESS_KEY=test TF_ACC=true go test ./internal/service/$1/ -v -timeout 60m -run $2
+TF_ACC=1 \
+  AWS_ACCESS_KEY_ID=test AWS_SECRET_ACCESS_KEY=test AWS_DEFAULT_REGION=us-west-1 \
```
bblommers, line 23: I think this is why some of the other TF tests are now failing - not all services are available in `us-west-1`. Can we keep `us-east-1` as the default?

macnev2013: image

- Region and Alternate Region should be different.

macnev2013: @bblommers , Only one job is failing. Can't find traces of the variable issue in here. Can you please take a look?

bblommers: Those two tests (`TestAccRoute53ZoneAssociation_crossAccount` and `TestAccRoute53ZoneAssociation_crossAccountAndRegion`) used to be skipped, because `AWS_ALTERNATE_ACCESS_KEY_ID` was not set. 

Now that the variable is set, they are executed - and fail because Moto doesn't yet support all required methods.

I'll just merge this now, and remove these tests from the list afterwards.

Thanks for getting this working @macnev2013!

bblommers (approved):

PR #6894, opened 2023-10-08
Rudimentary support for IVS

pepijn-motosumo:
This PR adds rudimentary support for IVS (Interactive Video Service). It's my first PR here so please let me know of anything I can improve!

For now I just added the endpoints related to channel management. I might come back and add more another time. I also tried to get the tagging endpoints working but it was giving me trouble, I may ask for help with that at some point. :-)

I didn't go very deep with the level of emulation, I'm not sure what's expected there or what the goal is. For now the only "real"ness it delivers is generating an error if a resource does not exist.

Boto documentation: https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/ivs.html
AWS user guide: https://docs.aws.amazon.com/ivs/latest/LowLatencyUserGuide/what-is.html
AWS api reference: https://docs.aws.amazon.com/ivs/latest/LowLatencyAPIReference/Welcome.html

### DIFF
--- a/moto/ivs/models.py
+++ b/moto/ivs/models.py
@@ -1,0 +1,131 @@
+"""IVSBackend class with methods for supported APIs."""
+from moto.core import BaseBackend, BackendDict
+from moto.ivs.exceptions import ResourceNotFoundException
+from moto.utilities.paginator import paginate
+from moto.moto_api._internal import mock_random
+
+
+class IVSBackend(BaseBackend):
+    """Implementation of IVS APIs."""
+
+    PAGINATION_MODEL = {
+        "list_channels": {
+            "input_token": "next_token",
+            "limit_key": "max_results",
+            "limit_default": 100,
+            "unique_attribute": "arn",
+        },
+    }
+
+    def __init__(self, region_name, account_id):
+        super().__init__(region_name, account_id)
+        self.channels = []
+
+    def create_channel(
+        self,
+        authorized,
+        insecure_ingest,
+        latency_mode,
+        name,
+        preset,
+        recording_configuration_arn,
+        tags,
+        type,
+    ):
+        channel_id = mock_random.get_random_string(12)
+        channel_arn = (
+            f"arn:aws:ivs:{self.region_name}:{self.account_id}:channel/{channel_id}"
+        )
+        channel = {
+            "arn": channel_arn,
+            "authorized": authorized,
+            "ingestEndpoint": "ingest.example.com",
+            "insecureIngest": insecure_ingest,
+            "latencyMode": latency_mode,
+            "name": name,
+            "playbackUrl": f"https://playback.example.com/{self.region_name}.{self.account_id}.{channel_id}.m3u8",
+            "preset": preset,
+            "recordingConfigurationArn": recording_configuration_arn,
+            "tags": tags,
+            "type": type,
+        }
+        self.channels.append(channel)
+        stream_key_id = mock_random.get_random_string(12)
+        stream_key_arn = f"arn:aws:ivs:{self.region_name}:{self.account_id}:stream-key/{stream_key_id}"
+        stream_key = {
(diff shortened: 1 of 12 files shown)


bblommers:
Hi @pepijn-motosumo, welcome to Moto!


> I also tried to get the tagging endpoints working but it was giving me trouble, I may ask for help with that at some point. :-)

AWS seems to implement tagging slightly different for every service, so it's always fun to figure out how to support this for a new service. Feel free to open a feature request or partial PR if this is something you need.

> I didn't go very deep with the level of emulation, I'm not sure what's expected there or what the goal is. For now the only "real"ness it delivers is generating an error if a resource does not exist.

If this is the level of emulation that works for you, then that is enough. If the next person wants some additional logic/validation, then they can submit their own PR. 🙂 

Overall, the logic/approach here looks good to me. Two minor niggles:
 - the linter is failing: https://github.com/getmoto/moto/actions/runs/6450578552/job/17523685319?pr=6894
 - there are no type annotations, that will be the next thing that will cause our linter to fail

You can run `make lint` locally to verify your changes. This can be a bit slow because of the size of our codebase, so you can also run these commands just for the new folders/files:
```
black moto/ivs tests/test_ivs
ruff moto/ivs tests/test_ivs
pylint moto/ivs tests/test_ivs
mypy
```

pepijn-motosumo:
> Hi @pepijn-motosumo, welcome to Moto!

Hi, thanks!

> Overall, the logic/approach here looks good to me. Two minor niggles:
> 
> * the linter is failing: https://github.com/getmoto/moto/actions/runs/6450578552/job/17523685319?pr=6894
> * there are no type annotations, that will be the next thing that will cause our linter to fail

I pushed fixes for this. :+1:

moto/ivs/models.py
```diff
@@ -0,0 +1,134 @@
...
+
+    def create_channel(
+        self,
+        authorized: bool,
+        insecure_ingest: bool,
+        latency_mode: str,
+        name: str,
+        preset: str,
+        recording_configuration_arn: str,
+        tags: dict[str, str],
```
bblommers, line 33:
```suggestion
        tags: Dict[str, str],
```
If I understand correctly, Python <= 3.8 doesn't support lowercase dict-types like this - it throws an error that `TypeError: 'type' object is not subscriptable`.
That means importing `typing.Dict` ofcourse, and changing this for all instances.

Failing tests:
https://github.com/getmoto/moto/actions/runs/6458787497/job/17534778481?pr=6894

tests/test_ivs/test_ivs.py
```diff
@@ -0,0 +1,192 @@
...
+import boto3
+from botocore.exceptions import ClientError
+from pytest import raises
+from moto import mock_ivs
+from re import fullmatch
+
+
+@mock_ivs
+def test_create_channel_with_name():
+    client = boto3.client("ivs")
```
bblommers, line 11: I should've spotted this earlier - but we have to specify a region for all tests, as there is not default region set in our CI.

Failing tests: https://github.com/getmoto/moto/actions/runs/6458787497/job/17534779886?pr=6894

bblommers (changes requested):

moto/ivs/models.py
pepijn-motosumo: Thanks, that's an easy fix!

tests/test_ivs/test_ivs.py
pepijn-motosumo: I understand. And I actually wondered why the tests generated by the `scaffold.py` script had a `region_name` in the calls, but the tests worked fine without that parameter. My environment doesn't have `AWS_DEFAULT_REGION` so that couldn't be it I thought. But I forgot about my `~/.aws/config` file. :facepalm:

bblommers (approved):
LGTM!

PR #6895, opened 2023-10-09
Small documentation improvements for new developers

pepijn-motosumo:
So I recently read through the documentation and followed the instructions. And I think a few small things can be improved. Let me know if these suggestions are incorrect or if I missed something!

- When I run `moto_server` it starts the version installed by pip, and not the code that I'm actually working on. So I think in the instructions about running the ServerMode tests, `moto_server` should be replaced by `python moto/server.py`.
- I added a reference to the `make lint` target, because I think it may have helped me to prevent pushing code with linting errors.
- Likewise, I updated the section in the documentation for how to test one specific service. I referred to this a lot while writing code because these commands give quick feedback. And I was actually curious why `ruff` was only run on `moto` and `pylint` only on `tests`. Turns out the commands were just incomplete. :-D I also think it makes sense to include `mypy` in this list.

All these changes except the `moto_server` one are actually based on suggestions from @bblommers in https://github.com/getmoto/moto/pull/6894.

### DIFF
--- a/docs/docs/contributing/development_tips/tests.rst
+++ b/docs/docs/contributing/development_tips/tests.rst
@@ -33,7 +33,7 @@ To verify whether your tests pass in ServerMode, you can run the following comma
 
 .. sourcecode:: bash
 
-  moto_server
+  python moto/server.py
   TEST_SERVER_MODE=true pytest -sv tests/test_service/..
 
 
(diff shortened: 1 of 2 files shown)


bblommers (approved):
LGTM - thank you for taking the time to improve this area @pepijn-motosumo!

PR #6896, opened 2023-10-10
S3: Add paging to list_object_versions()

sveniu:
This adds paging to list_object_versions(). The original motivation was to enable moto testing for an S3 point-in-time recovery tool, where proper handling of paging is important.

CC @bblommers

### DIFF
--- a/moto/s3/models.py
+++ b/moto/s3/models.py
@@ -1798,7 +1798,7 @@ class S3Backend(BaseBackend, CloudWatchMetricProvider):
     ]:
         bucket = self.get_bucket(bucket_name)
 
-        common_prefixes: List[str] = []
+        common_prefixes: Set[str] = set()
         requested_versions: List[FakeKey] = []
         delete_markers: List[FakeDeleteMarker] = []
         next_key_marker: Optional[str] = None
@@ -1819,10 +1819,10 @@ class S3Backend(BaseBackend, CloudWatchMetricProvider):
             if key_marker is None:
                 return True
 
-            if version_id_marker is None:
-                return key_marker == ver.name
+            if version_id_marker is None or version_id_marker == "":
+                return ver.name >= key_marker
 
-            return key_marker == ver.name and version_id_marker == ver.version_id
+            return ver.name == key_marker and ver.version_id == version_id_marker
 
         for version in itertools.dropwhile(
             lambda ver: not key_or_version_match(ver),
@@ -1838,35 +1838,44 @@ class S3Backend(BaseBackend, CloudWatchMetricProvider):
             if not name.startswith(prefix):
                 continue
             # separate keys that contain the same string between the prefix and the first occurrence of the delimiter
+            is_common_prefix = False
             if delimiter and delimiter in name[len(prefix) :]:
                 end_of_delimiter = (
                     len(prefix) + name[len(prefix) :].index(delimiter) + len(delimiter)
                 )
                 prefix_including_delimiter = name[0:end_of_delimiter]
-                common_prefixes.append(prefix_including_delimiter)
-                continue
+
+                # Skip already-processed common prefix.
+                if prefix_including_delimiter == key_marker:
+                    continue
+
+                common_prefixes.add(prefix_including_delimiter)
+                name = prefix_including_delimiter
+                is_common_prefix = True
 
             # Only return max_keys items.
             if (
                 max_keys is not None
-                and len(requested_versions) + len(delete_markers) >= max_keys
+                and len(requested_versions) + len(delete_markers) + len(common_prefixes)
+                >= max_keys
             ):
-                next_key_marker = version.name
-                next_version_id_marker = version.version_id
+                next_key_marker = name
+                next_version_id_marker = (
+                    version.version_id if not is_common_prefix else None
+                )
                 break
 
-            # Differentiate between FakeKey and FakeDeleteMarkers
-            if not isinstance(version, FakeKey):
-                delete_markers.append(version)
-                continue
+            if not is_common_prefix:
+                # Differentiate between FakeKey and FakeDeleteMarkers
+                if not isinstance(version, FakeKey):
+                    delete_markers.append(version)
+                    continue
 
-            requested_versions.append(version)
-
-        common_prefixes = sorted(set(common_prefixes))
+                requested_versions.append(version)
 
         return (
             requested_versions,
-            common_prefixes,
+            sorted(common_prefixes),
             delete_markers,
             next_key_marker,
             next_version_id_marker,
--- a/moto/s3/models.py
+++ b/moto/s3/models.py
@@ -1790,13 +1790,19 @@ class S3Backend(BaseBackend, CloudWatchMetricProvider):
         bucket_name: str,
         delimiter: Optional[str] = None,
         key_marker: Optional[str] = None,
+        max_keys: Optional[int] = 1000,
         prefix: str = "",
-    ) -> Tuple[List[FakeKey], List[str], List[FakeDeleteMarker]]:
+        version_id_marker: Optional[str] = None,
+    ) -> Tuple[
+        List[FakeKey], List[str], List[FakeDeleteMarker], Optional[str], Optional[str]
+    ]:
         bucket = self.get_bucket(bucket_name)
 
         common_prefixes: List[str] = []
         requested_versions: List[FakeKey] = []
         delete_markers: List[FakeDeleteMarker] = []
+        next_key_marker: Optional[str] = None
+        next_version_id_marker: Optional[str] = None
         all_versions = list(
             itertools.chain(
                 *(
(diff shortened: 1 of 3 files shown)


bblommers (changes requested):
Hi @sveniu!

Moto has a `@paginator` decorator to simplify this - it should allow us to keep the existing logic intact, with the decorator doing all the heavy lifting around pagination.
See the docs here: http://docs.getmoto.org/en/latest/docs/contributing/development_tips/utilities.html#paginator

sveniu:
Hi @bblommers. Thanks for the pointer, I wasn't aware of that decorator. Looking at it, it seems it doesn't support multiple input tokens such as key-marker + version-id-marker used by the list_object_versions call. I'll see if I can wrap my head around the inner workings and add support for it.

bblommers:
@sveniu You're right, that is not straightforward with the current implementation. Let's stick with this then for now, we can always refactor it later.

Looks like the linter is failing: https://github.com/getmoto/moto/actions/runs/6465933620/job/17577370491?pr=6896

moto/s3/models.py
```diff
@@ -1807,16 +1813,21 @@ def list_object_versions(
...
-        for version in all_versions:
+
+        def key_or_version_match(ver: FakeKey | FakeDeleteMarker) -> bool:
+            if key_marker is None:
+                return True
+
+            if version_id_marker is None:
+                return key_marker == ver.name
+
+            return key_marker == ver.name and version_id_marker == ver.version_id
```
bblommers, line 1824: Should this be `key_marker <= ver.name and version_id_marker <= ver.version_id`?

As the docs specifically state:
> Specifies the key **to start with** when listing objects in a bucket.

bblommers:
Actually - it looks like the linter is failing on `master` as well, due to a MyPy update. I've raised #6898 to fix that.

moto/s3/models.py
sveniu: That would work if version_id was sequential, but it is basically a random string. I first tried just adding `version_id_marker <= ver.version_id`, but it produced flaky results depending on the random version_id values.

I opted for itertools.dropwhile here, and the helper function simply needs to signal when the first match is found, using exact matching, and dropwhile will then return the remaining items. Quite suitable for this purpose, I think. Was thinking to implement that version_id "search" manually with a version_marker_found bool, but found dropwhile to be a good fit. Let me know if you have thoughts on it.

The main loop still applies the prefix and delimiter logic, and with the max_keys paging added, it looks to me like it behaves like the real S3 server does.

I'd be happy to add more tests if there are suggestions.

bblommers: Fair enough. Would you mind merging in the latest master? That should fix the linter - if the existing tests all pass, then I'd be happy to merge then.

bblommers:
@sveniu The linter is still failing - I think you would have to use `Union[..]` instead of `X | Y`:

https://github.com/getmoto/moto/actions/runs/6493285516/job/17635218175?pr=6896

I'll have a look at the failing Terraform tests later on.

bblommers:
OK, so there is a change in behaviour somewhere. The existing test that now fails, `test_list_object_versions_with_delimiter`, succeeds when running it against AWS, so it should pass with Moto as well.

Conversely, the new test `test_list_object_versions_with_paging` currently fails when running it against AWS - so something isn't quite right with the implementation.

sveniu:
Finally fixed how paging is done when common prefixes are involved – at least moto now behaves the same as the real S3 server there, and tests look green now.

@bblommers Dunno if you have more thoughts on this – let me know if you do. Otherwise, the branch is full of fixup commits, and I guess you do a squash-merge usually?

bblommers:
Hi @sveniu, thanks for continuing to chip away at this!

As far as I can tell, there are still two issues with this:
1. the `next_version_id_marker` should actually point to the last `version_id` that is returned (naming is hard..)
2. objects should be sorted by `type` -> `name` -> `date`, so that DeleteMarkers are returned first.

Point 2 is an issue with the existing implementation, and the existing tests also do not really verify ordering. But from what I can see, this is the order in which AWS returns things:
1. DeleteMarkers from first object (sorted by time)
2. First object
3. DeleteMarkers from second object
4. Second object

If you want to test this out, I've modified the new tests slightly so that they pass against AWS.
This `s3_aws_verified`-decorator will automatically create the bucket for you, and will also automatically delete all objects and the bucket itself after the test ends - so you don't have to worry about having to clean your real AWS account after running these tests.

```
diff --git a/tests/test_s3/test_s3.py b/tests/test_s3/test_s3.py
index 63ead96d3..3d5ce2422 100644
--- a/tests/test_s3/test_s3.py
+++ b/tests/test_s3/test_s3.py
@@ -2613,11 +2613,10 @@ def test_list_object_versions_with_versioning_enabled_late():
     assert response["Body"].read() == items[-1]
 
 
-@mock_s3
-def test_list_object_versions_with_paging():
+@pytest.mark.aws_verified
+@s3_aws_verified
+def test_list_object_versions_with_paging(bucket_name=None):
     s3_client = boto3.client("s3", region_name=DEFAULT_REGION_NAME)
-    bucket_name = "000" + str(uuid4())
-    s3_client.create_bucket(Bucket=bucket_name)
     s3_client.put_bucket_versioning(
         Bucket=bucket_name, VersioningConfiguration={"Status": "Enabled"}
     )
@@ -2647,7 +2646,7 @@ def test_list_object_versions_with_paging():
     assert "NextKeyMarker" in page1
     assert page1["NextKeyMarker"] == "obj1"
     assert "NextVersionIdMarker" in page1
-    assert page1["NextVersionIdMarker"] == obj1ver1["VersionId"]
+    assert page1["NextVersionIdMarker"] == obj1ver2["VersionId"]
 
     # Second page should be the last page and have the oldest version.
     page2 = s3_client.list_object_versions(
@@ -2668,11 +2667,10 @@ def test_list_object_versions_with_paging():
     assert page2["Versions"][0]["VersionId"] == obj1ver1["VersionId"]
 
 
-@mock_s3
-def test_list_object_versions_with_paging_and_delete_markers():
+@pytest.mark.aws_verified
+@s3_aws_verified
+def test_list_object_versions_with_paging_and_delete_markers(bucket_name=None):
     s3_client = boto3.client("s3", region_name=DEFAULT_REGION_NAME)
-    bucket_name = "000" + str(uuid4())
-    s3_client.create_bucket(Bucket=bucket_name)
     s3_client.put_bucket_versioning(
         Bucket=bucket_name, VersioningConfiguration={"Status": "Enabled"}
     )
@@ -2706,7 +2704,7 @@ def test_list_object_versions_with_paging_and_delete_markers():
     assert "NextKeyMarker" in page1
     assert page1["NextKeyMarker"] == "obj1"
     assert "NextVersionIdMarker" in page1
-    assert page1["NextVersionIdMarker"] == obj1dmk1["VersionId"]
+    assert page1["NextVersionIdMarker"] == obj1dmk2["VersionId"]
 
     page2 = s3_client.list_object_versions(
         Bucket=bucket_name,
@@ -2728,9 +2726,9 @@ def test_list_object_versions_with_paging_and_delete_markers():
 
     # The next key/version markers should point to obj2 dmk1.
     assert "NextKeyMarker" in page2
-    assert page2["NextKeyMarker"] == "obj2"
+    assert page2["NextKeyMarker"] == "obj1"
     assert "NextVersionIdMarker" in page2
-    assert page2["NextVersionIdMarker"] == obj2dmk1["VersionId"]
+    assert page2["NextVersionIdMarker"] == obj1ver1["VersionId"]
 
     page3 = s3_client.list_object_versions(
         Bucket=bucket_name,
@@ -2756,11 +2754,10 @@ def test_list_object_versions_with_paging_and_delete_markers():
     assert "NextVersionIdMarker" not in page3
 
 
-@mock_s3
-def test_list_object_versions_with_paging_and_delimiter():
+@pytest.mark.aws_verified
+@s3_aws_verified
+def test_list_object_versions_with_paging_and_delimiter(bucket_name=None):
     s3_client = boto3.client("s3", region_name=DEFAULT_REGION_NAME)
-    bucket_name = "000" + str(uuid4())
-    s3_client.create_bucket(Bucket=bucket_name)
     s3_client.put_bucket_versioning(
         Bucket=bucket_name, VersioningConfiguration={"Status": "Enabled"}
     )
@@ -2807,7 +2804,6 @@ def test_list_object_versions_with_paging_and_delimiter():
         Delimiter="with-",
         Prefix="key1",
         KeyMarker="key12-with-",
-        VersionIdMarker="",
         MaxKeys=5,
     )
 ```

By default, the test will still run against Moto, but run it against AWS like so:
```
MOTO_TEST_ALLOW_AWS_REQUEST=true pytest -sv tests/test_s3/test_s3.py -m aws_verified
```

bblommers:
Oh, and to answer your question - Yes, I always squash-merge, so don't worry about the commit history. :slightly_smiling_face:

bentsku:
Hello, just chiming in here regarding the sorting order: if you put an object then delete it without specifying the object version then running `ListObjectVersions` with `MaxKeys=1`, you will get the `Versions` field and not the `DeleteMarkers`, which shows that the time is sorted before the type. Running the test with more keys shows the order is name > time > type it seems.

bblommers:
I'll just merge this as is, considering it is already a major improvement on what we have now. I'll raise a separate PR with the final touches.

Thank you for all your work on this @sveniu!

bblommers (approved):

PR #6901, opened 2023-10-11
EC2: Add default DHCP options set for VPCs

viren-nadkarni:
AWS provides the `AssociateDhcpOptions` operation to associate a DHCP Options Set with a VPC.

There is no `DisassociateDhcpOptions` operation to remove a DHCP Options Set from a VPC. Instead, `DhcpOptionsId` must be set to `default` when calling `AssociateDhcpOptions` to achieve this.

This PR implements this capability.

See: <https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_AssociateDhcpOptions.html#API_AssociateDhcpOptions_RequestParameters>

Closes: https://github.com/localstack/localstack/issues/9222

### DIFF
--- a/moto/ec2/models/dhcp_options.py
+++ b/moto/ec2/models/dhcp_options.py
@@ -69,6 +69,11 @@ class DHCPOptionsSetBackend:
         dhcp_options.vpc = vpc
         vpc.dhcp_options = dhcp_options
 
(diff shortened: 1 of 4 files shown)


bblommers (approved):

PR #6904, opened 2023-10-12
EC2: Add support for ED25519 SSH key pairs

viren-nadkarni:
This PR adds support for ED25519 SSH key pairs in EC2.

It is now possible to use the `KeyType` field in `CreateKeyPair`, as well as the ability to `ImportKeyPair` ED25519 keys.

See:
- https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_CreateKeyPair.html
- https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_ImportKeyPair.html
- https://github.com/localstack/localstack/issues/9237

### DIFF
--- a/moto/ec2/models/key_pairs.py
+++ b/moto/ec2/models/key_pairs.py
@@ -8,11 +8,12 @@ from ..exceptions import (
     InvalidKeyPairFormatError,
 )
 from ..utils import (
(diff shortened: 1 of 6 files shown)


bblommers (approved):

PR #6910, opened 2023-10-13
[iot] update_thing(): delete empty attributes with merge=true

Wurstnase:
https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/iot/client/update_thing.html
- To remove an attribute, call UpdateThing with an empty attribute value.


bblommers (approved):
LGTM - thank you for fixing this @Wurstnase!

PR #6918, opened 2023-10-15
IOT: update_thing() now behaves correctly with merge=True

Wurstnase:
The last patch missed some cases and the behaviour was wrong.
I added all those cases in the test and fixed the problems.

### DIFF
--- a/moto/iot/models.py
+++ b/moto/iot/models.py
@@ -883,7 +883,11 @@ class IoTBackend(BaseBackend):
             if not do_merge:
                 thing.attributes = attributes
             else:
-                thing.attributes = {k: v for k, v in attributes.items() if v}
+                for k, v in attributes.items():
+                    if v:
+                        thing.attributes[k] = v
+                    else:
+                        thing.attributes.pop(k)
 
     def create_keys_and_certificate(
         self, set_as_active: bool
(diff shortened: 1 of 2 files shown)


moto/iot/models.py
```diff
@@ -883,7 +883,11 @@ def update_thing(
             if not do_merge:
                 thing.attributes = attributes
             else:
-                thing.attributes = {k: v for k, v in attributes.items() if v}
+                for k, v in attributes.items():
+                    if v:
+                        thing.attributes[k] = v
+                    else:
+                        thing.attributes.pop(k)
```
bblommers, line 890: How does AWS behave if `k` does not exist here?
I.e., thing.attributes only has attribute`k1`, and a user calls `update_thing(.. {"k2": ""})`?

Wurstnase: No exception, nothing changed. Simply a `{'ResponseMetadata': {..., 'HTTPStatusCode': 200,...}`.

Wurstnase: I just check another special case.

With `"merge": False` and `update_thing(.. {"k2": ""})`, all attributes are deleted. So an empty value will never be saved.

bblommers (approved):
Thanks for the update @Wurstnase!

Issue #6909, opened 2023-10-13
SESV2 send_email saves wrong body in ses_backend


Alexandre-Bonhomme:
When using sesv2 send_email, the email saved in the ses_backend is not what it should be. The body of the email is replaced by the subject.

The issue comes from moto/sesv2/responses.py line 45-50:
```
message = self.sesv2_backend.send_email(  # type: ignore
    source=from_email_address,
    destinations=destination,
    subject=content["Simple"]["Subject"]["Data"],
    body=content["Simple"]["Subject"]["Data"],
)
```
I think it should be corrected like this:
```
message = self.sesv2_backend.send_email(  # type: ignore
    source=from_email_address,
    destinations=destination,
    subject=content["Simple"]["Subject"]["Data"],
    body=content["Simple"]["Body"]["Text"]["Data"],
)
```
It doesn't looks like there is any test to update with this change.

I tried to create a pull request but I'm really not comfortable with github and I could not do it. Could this bug me resolved ?

bblommers:
Hi @Alexandre-Bonhomme, welcome to Moto! Will raise a PR shortly with a fix.

PR #6927, opened 2023-10-18
Fix: SESV2 send_email with HTML body

forzagreen:
Fix of #6909, part 2.

#6913 handled the case of "Text" body. The goal of this PR is to handle the "Html" body case.

Ref: see SES v2 [send_email](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/sesv2/client/send_email.html).
Cc: @bblommers


bblommers (approved):
LGTM - thanks for adding this @forzagreen!

PR #6930, opened 2023-10-20
IoT: Fix bad account IDs in ARNs

viren-nadkarni:
Certain IoT resource types had bad hardcoded account ID (`1`) in their ARNs, like so:

```
arn:aws:iot:us-east-1:1:job/asdf123
```

This PR ensures that the ARNs have the proper account IDs.

The resource types are:
- ThingType
- ThingGroup
- Job
- Rule
- DomainConfiguration


bblommers (approved):
LGTM, thanks for fixing this @viren-nadkarni!

PR #6933, opened 2023-10-21
S3: fix unquoting of S3 object keys in DeleteObjects

bentsku:
As reported in https://github.com/localstack/localstack/issues/9428, moto is decoding the content of the `DeleteObjects` body, but those keys are not going through the URL encoding as they're in the XML, and should be used as is.

I've removed the `get_safe_path` call from `DeleteObjects` and added a test validating the behaviour, tested against AWS.

This devolved into a bit of a rabbit hole, as it seems fixing this still did not delete the keys, because they were saved with an emoji in the name where it should have been using the raw url encoded format like S3.

```python
object_key_emoji = "a/%F0%9F%98%80"
client.put_object(Bucket=bucket_name, Key=object_key_emoji, Body="emoji percent encoding")
```
Result of `ListObjects`:
```json
{
    "Contents": [
        {
            "ETag": "\"ec53baa61c0c0b736a567bdef59250f3\"",
            "Key": "a/😀",
        }
     ]
}
```

I've changed the `get_safe_path` util to make use of Werkzeug `RAW_URI`, as this contains the URI as received by the server. We can then use those proper values to construct the object key, and this circumvents all issue related to the key encoding. This is a rather big change, but if the tests are green, I believe this would be safe to do.

### DIFF
--- a/moto/core/responses.py
+++ b/moto/core/responses.py
@@ -363,6 +363,11 @@ class BaseResponse(_TemplateEnvironmentMixin, ActionAuthenticatorMixin):
         self.uri = full_url
 
         self.path = self.parsed_url.path
+        if self.is_werkzeug_request and "RAW_URI" in request.environ:
+            self.raw_path = urlparse(request.environ.get("RAW_URI")).path
+        else:
+            self.raw_path = self.path
+
         self.querystring = querystring
         self.data = querystring
         self.method = request.method
--- a/moto/core/utils.py
+++ b/moto/core/utils.py
@@ -1,13 +1,12 @@
 import datetime
 import inspect
 import re
-import unicodedata
 from botocore.exceptions import ClientError
 from gzip import decompress
 from typing import Any, Optional, List, Callable, Dict, Tuple
-from urllib.parse import urlparse, unquote
+from urllib.parse import urlparse
 from .common_types import TYPE_RESPONSE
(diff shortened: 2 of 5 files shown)


bblommers:
> This devolved into a bit of a rabbit hole
>

Yup.. that sounds familiar! It took me ages to get it working in the first place, but I wasn't really happy with the outcome. If you can mange to simplify the solution, that would be very welcome.

For context, this was the original PR with some rationale: https://github.com/getmoto/moto/pull/6602

This is now complicated by the introduction of MotoProxy (https://github.com/getmoto/moto/pull/6848), which does not have the concept of a `RAW_URI`.

bentsku:
Thanks for the pointer towards MotoProxy, I had totally missed that. I will validate the failing `x-copy-source*` tests with URL encoding against AWS to see what the behaviour is and what I've missed, and then will work to make it work for `MotoProxy`. Those Werkzeug upgrades were no joke 😓

bblommers:
Just to make things interesting, I've added the `moto-core` label, which runs the test against older werkzeug versions. 

I'm assuming that's not a problem, because `RAW_URI` will be exposed for older werkzeug versions as well - but let's see.

bentsku:
Seems like the failure might have been a flake as everything else is green? Looks this might work! 😄

bblommers (approved):
Yes, this LGTM. Awesome - thank you for fixing this @bentsku!

PR #6938, opened 2023-10-22
AWSLambda: mark layers test as requiring docker

mweinelt:
Otherwise without docker it fails like this;

```
        msg = success_result["Payload"].read().decode("utf-8")
>       assert msg == '"2.31.0"'
E       assert "error runnin... directory'))" == '"2.31.0"'
E         - "2.31.0"
E         + error running docker: Error while fetching server API version: ('Connection aborted.', FileNotFoundError(2, 'No such file or directory'))
```


bblommers (approved):
LGTM - thanks for the PR @mweinelt, and welcome to Moto!

PR #6948, opened 2023-10-23
Panorama: Add mock_panorama

YHallouard:
This pull request implements for panorama :

 - [X] provision_device
 - [X] describe_device
 - [X] update_device_metadata
 - [X] list_devices
 - [X] delete_device


YHallouard:
Hello there,

Sorry to bother you, i'm working on that new part of moto. I just encounter an error when I try the test suite: 

image

```
if http.status_code >= 300:
E           botocore.exceptions.ClientError: An error occurred (UnrecognizedClientException) when calling the ProvisionDevice operation: {"Message":The security token included in the request is invalid.}
```
Do you have an idea ? I'm a bit blocked

bblommers:
Hi @YHallouard, that usually happens if the correct URL's are not configured.

Looking at the API, AWS will send a request to `/devices` for the `provision_device`-feature, and `devices/{DeviceId}` for `describe_device`:
https://github.com/boto/botocore/blob/7c46948d697e1a6fbbeeb85dfb5bf8d467b08e28/botocore/data/panorama/2019-07-24/service-2.json#L483

So that should be represented in `urls.py` like so:
```
url_paths = {
    "{0}/devices$": PanoramaResponse.dispatch,
   "{0}/devices/(?P<DeviceId>[^/]+)$": PanoramaResponse.dispatch,
}
```

There is some documentation here, describing how you could also use a proxy to discover the correct URL:
http://docs.getmoto.org/en/latest/docs/contributing/development_tips/urls.html

Let me know if that works - if it doesn't, I'll have another closer look.

YHallouard:
> http://docs.getmoto.org/en/latest/docs/contributing/development_tips/urls.html

Hi @bblommers ! 

Thank you very much for your precious advices !! It works perfectly ! 

I will still implement the `list_devices`in the PR, to go a little bit further in device section and it will be good. But I just wanted your opinion about the devices themselves. Since they are something that have a life cycle, they can be "ERROR",  "AWAITING_PROVISIONING", "PENDING", "FAILED", "DELETING", "ONLINE", "OFFLINE", "LEASE_EXPIRED", "UPDATE_NEEDED", "REBOOTING", should we fixe that value ? It will not allow to test some functionalities, or could we add a kind of feature_flag in the provision_device ? Even if it does not reflect the aws panorama api ... 

I don't know if it has already been done in moto

bblommers:
We do have a state transition API for this kind of thing. But, like anything else, it depends on whether you need it. If your usecase doesn't need it, then there's nothing wrong with setting a fixed value for lifecycle - that can always be implemented later.

User facing docs for the state transitions: http://docs.getmoto.org/en/latest/docs/configuration/state_transition/index.html

And dev docs: http://docs.getmoto.org/en/latest/docs/contributing/development_tips/new_state_transitions.html

YHallouard:
> We do have a state transition API for this kind of thing. But, like anything else, it depends on whether you need it. If your usecase doesn't need it, then there's nothing wrong with setting a fixed value for lifecycle - it can always be improved on later.
> 
> If you do want to add it, here are the user facing docs for the state transitions: http://docs.getmoto.org/en/latest/docs/configuration/state_transition/index.html
> 
> And the dev docs: http://docs.getmoto.org/en/latest/docs/contributing/development_tips/new_state_transitions.html

I tried an implementation, tell me what do you think about it :)

moto/panorama/models.py
```diff
@@ -0,0 +1,304 @@
...
+    def __init__(
+        self,
+        account_id: str,
+        region_name: str,
+        description: Optional[str],
+        name: str,
+        network_configuration: Optional[Dict[str, Any]],
+        tags: Optional[Dict[str, str]],
+    ) -> None:
+        self.__device_aggregated_status_manager = ManagedState(
```
bblommers, line 63: Normally the Device-class would extend `ManagedState`, giving us the functionality out of the box. 

It took me a while to understand the reasoning behind your approach, but I now understand that's the only way we can keep track of two statuses. 

Could you leave some comments to explain this? That would make it easier to understand next time a developer is looking at this.

moto/panorama/models.py
```diff
@@ -0,0 +1,304 @@
+import base64
+import json
+from datetime import datetime, timedelta
+from dateutil.tz import tzutc
+from typing import Dict, Any, Optional, Union, List
+from typing_extensions import Literal
```
bblommers, line 6: Moto doesn't have the `typing_extensions` dependency (at runtime), and I don't really want to introduce it either.

Can we replace these all just with `attr: str`? If necessary, just leave a comment with the available/allowed values.

moto/panorama/models.py
```diff
@@ -0,0 +1,304 @@
...
+    @property
+    def response_provision(self) -> Dict[str, Union[str, bytes]]:
+        return {
+            "Arn": self.arn,
+            "Certificates": base64.b64encode("certificate".encode("utf-8")).decode(
+                "utf-8"
+            ),
+            "DeviceId": self.device_id,
+            "IotThingName": self.iot_thing_name,
+            "Status": "AWAITING_PROVISIONING",
```
bblommers, line 207: Can we use `self.__device_provisioning_status_manager.status` here, instead of hardcoding the status?

moto/panorama/models.py
```diff
@@ -0,0 +1,304 @@
...
+        self.serial_number = "GAD81E29013274749"
+        self.type: Literal[
+            "PANORAMA_APPLIANCE_DEVELOPER_KIT", "PANORAMA_APPLIANCE"
+        ] = "PANORAMA_APPLIANCE"
+
+    @property
+    def device_aggregated_status(self) -> DEVICE_AGGREGATED_STATUS_TYPE:
+        _device_aggregated_status: DEVICE_AGGREGATED_STATUS_TYPE = self.__device_aggregated_status_manager.status  # type: ignore[assignment]
+        self.__device_aggregated_status_manager.advance()
+        return _device_aggregated_status
```
bblommers, line 127:
```suggestion
        self.__device_aggregated_status_manager.advance()
        return self.__device_aggregated_status_manager.status
```
Looks like we can simplify this a bit, without the intermediary variable. Same with the provisioning_status below

moto/panorama/models.py
```diff
@@ -0,0 +1,304 @@
...
+                "DeviceAggregatedStatus": self.device_aggregated_status,
+                "ProvisioningStatus": self.provisioning_status,
+            },
+        }
+
+    @property
+    def response_provision(self) -> Dict[str, Union[str, bytes]]:
+        return {
+            "Arn": self.arn,
+            "Certificates": base64.b64encode("certificate".encode("utf-8")).decode(
```
bblommers, line 202: We can generate this value once on initialization, and just return that instead - no need to calculate it every time

tests/test_panorama/test_panorama_device.py
```diff
@@ -0,0 +1,423 @@
...
+    assert resp["SerialNumber"] == "GAD81E29013274749"
+    assert resp["Tags"] == {"Key": "test-key", "Value": "test-value"}
+    assert resp["Type"] == "PANORAMA_APPLIANCE"
+
+
+@mock_panorama
+def test_provision_device_aggregated_status_lifecycle() -> None:
+    client = boto3.client("panorama", region_name="eu-west-1")
+    given_device_name = "test-device-name"
+    state_manager.set_transition(
```
bblommers, line 136: This test would have to be skipped in ServerMode as well, as the `state_manager`-object only works in decorator-mode.

(If you want to configure this in ServerMode, you would have to make a HTTP-request to `http://localhost:5000/moto-api/state-manager/set-transition` instead - but I don't think that adds anything here.)

moto/panorama/responses.py
```diff
@@ -0,0 +1,75 @@
...
+        self,
+    ) -> str:
+        device_aggregated_status_filter = self._get_param(
+            "DeviceAggregatedStatusFilter"
+        )
+        max_results = (
+            int(self._get_param("MaxResults"))
+            if self._get_param("MaxResults")
+            else None
+        )
```
bblommers, line 44:
```suggestion
        max_results = self._get_int_param("MaxResults")
```
Should behave the same

moto/panorama/models.py
```diff
@@ -0,0 +1,304 @@
...
+        self.devices_memory[device_obj.device_id] = device_obj
+        return device_obj
+
+    def describe_device(self, device_id: str) -> Device:
+        device = self.devices_memory.get(device_id)
+        if device is None:
+            raise ValidationError(f"Device {device_id} not found")
+        return device
+
+    @paginate(pagination_model=PAGINATION_MODEL)  # type: ignore[misc]
```
bblommers, line 249: Good to see this working OOTB! It's quite a lot of functionality that you get with just a single decorator, so I'm happy to see another example where this 'just' works.

bblommers (changes requested):
See my comments @YHallouard! Let me know if you want me to have a look at the failing test cases as well.

moto/panorama/models.py
YHallouard: Yes it is a good idea: https://github.com/getmoto/moto/pull/6948/commits/ff562d22b7a6f5bdfba5bc7b5e1dc074dfd50934

Tell me if you think that I should explain how to set them, make them advance etc

moto/panorama/models.py
YHallouard: Done: https://github.com/getmoto/moto/pull/6948/commits/ff562d22b7a6f5bdfba5bc7b5e1dc074dfd50934
Sorry I did this In the same commit

moto/panorama/models.py
YHallouard: Done: https://github.com/getmoto/moto/pull/6948/commits/90828eb8248b14b311cfe55e11c01c4b71c8368f

moto/panorama/models.py
YHallouard: Done: https://github.com/getmoto/moto/pull/6948/commits/fad936cc1e06b8f5c3c868db53ef0fc63a8180e3

moto/panorama/models.py
YHallouard: Done: https://github.com/getmoto/moto/pull/6948/commits/194c04a18c5ecb61416ed4f42e9b255165227ae1

tests/test_panorama/test_panorama_device.py
YHallouard: Done: https://github.com/getmoto/moto/pull/6948/commits/1c0dd6474c8970c06c9d6e24048495029a91a366

moto/panorama/responses.py
YHallouard: Done : https://github.com/getmoto/moto/pull/6948/commits/b3deaf8c494c4c811cc19fb9b7dd83ba1efcda5e

I did not saw it ahah

moto/panorama/models.py
YHallouard: Yes, I'm also very impressed ! 💯 Moto is a very good package ;)

YHallouard:
> See my comments @YHallouard! Let me know if you want me to have a look at the failing test cases as well.

Thank you for all your comments, I've been quite busy last month ... 
I took at you comment, there were all pertinent, thank you for your reactivity ! I don't know why the linter is failling, it is not on local and on my github

bblommers:
Hi @YHallouard, this looks great.

How are you running the linter locally? We've switched from `flake8` to `ruff` recently, so if the linting was done with `flake8` it will not pick up all required changes.

It looks like the linter fails because the import-statements are unordered - you can fix that automatically be running:
 ```ruff moto tests --fix```

And then `make lint` afterwards to verify MyPy (the last linting-step) also passes.

YHallouard:
Hi @bblommers, my bad, I was using `ruff` but I forgot that I was work on a fork, a sync and rebase solved the issue - let's see if the pipeline goes all the way green

bblommers (approved):
LGTM. Thank you so much for adding this to Moto @YHallouard!

PR #6955, opened 2023-10-27
Align organizations PolicyNotFoundException with boto3 response

siluk-aws:
At the moment, the PolicyNotFoundException raised by the mock_organizations is not returning the same `Error` responses as the boto3 client.

The mocked response returns a `Code` of `"400"` and an XML string in the `Message` block.

The expected behaviour based on the boto3 client would be to return a `Code` of `"PolicyNotFoundException"` and a message specific to the call, for example, `"We can't find a policy with the PolicyId that you specified."` for [DescribePolicy](https://docs.aws.amazon.com/organizations/latest/APIReference/API_DescribePolicy.html#API_DescribePolicy_Errors).

Steps to reproduce:

```python
import boto3
import moto

client = boto3.client("organizations")

try:
    client.describe_policy(PolicyId="p-definitelydoesnotexist")
except Exception as e:

    print(type(e).__name__)
    # PolicyNotFoundException

    print(e.response["Error"])
    # {'Message': "You specified a policy that doesn't exist.", 'Code': 'PolicyNotFoundException'}


with moto.mock_organizations():
    try:
        client.describe_policy(PolicyId="p-definitelydoesnotexist")
    except Exception as e:

        print(type(e).__name__)
        # ClientError

        print(e.response["Error"])
        # {'Message': '<?xml version="1.0" encoding="UTF-8"?>\n  <ErrorResponse>\n    <Errors>\n      <Error>\n        <Code>PolicyNotFoundException</Code>\n        <Message><![CDATA[You specified a policy that doesn\'t exist.]]></Message>\n        \n      </Error>\n    </Errors>\n  <RequestId>7a62c49f-347e-4fc4-9331-6e8eEXAMPLE</RequestId>\n</ErrorResponse>', 'Code': '400'}
```

### DIFF
--- a/moto/organizations/exceptions.py
+++ b/moto/organizations/exceptions.py
@@ -109,3 +109,10 @@ class TargetNotFoundException(JsonRESTError):
         super().__init__(
             "TargetNotFoundException", "You specified a target that doesn't exist."
         )
+
+
+class PolicyNotFoundException(JsonRESTError):
+    code = 400
+
+    def __init__(self, message: str) -> None:
+        super().__init__("PolicyNotFoundException", message)
--- a/moto/organizations/models.py
+++ b/moto/organizations/models.py
@@ -16,6 +16,7 @@ from moto.organizations.exceptions import (
     AWSOrganizationsNotInUseException,
     AccountNotRegisteredException,
     RootNotFoundException,
+    PolicyNotFoundException,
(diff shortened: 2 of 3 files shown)


bblommers (approved):
LGTM - thank you for fixing this @siluk-aws!

PR #6957, opened 2023-10-27
Update CONTRIBUTING.md

harshhere905:


bblommers:
Hi @harshhere905!

`can execute` is the right verb tense here. It's essentially means: `[..] download a Docker image before they can execute [that image]`.

PR #6966, opened 2023-10-29
Admin: Support Py 3.12

bblommers:
Includes an upgrade to `py-partiql-parser` to a version that supports Py 3.12

From the 3.12 release notes:

> [datetime](https://docs.python.org/3/library/datetime.html#module-datetime): [datetime.datetime](https://docs.python.org/3/library/datetime.html#datetime.datetime)’s [utcnow()](https://docs.python.org/3/library/datetime.html#datetime.datetime.utcnow) and [utcfromtimestamp()](https://docs.python.org/3/library/datetime.html#datetime.datetime.utcfromtimestamp) are deprecated and will be removed in a future version

Moto already took care of this in #6803, but botocore still uses these methods - so it throws a lot of warnings.

The test changes are necessary because we're suddenly also catching these deprecated datetime warnings, not just our own.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -8,7 +8,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [ 3.7, 3.8, 3.9, "3.10", "3.11" ]
+        python-version: [ 3.7, 3.8, 3.9, "3.10", "3.11", "3.12" ]
 
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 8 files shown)


rafrafek:
https://github.com/getmoto/moto/blob/master/moto/core/utils.py#L185 still gives me:
```
DeprecationWarning:
datetime.datetime.utcfromtimestamp() is deprecated and scheduled for removal in a future version.
Use timezone-aware objects to represent datetimes in UTC:
    datetime.datetime.fromtimestamp(timestamp, datetime.UTC).

epoch = datetime.datetime.utcfromtimestamp(0)
```

rafrafek:
And there are other similar examples, like here: https://github.com/getmoto/moto/blob/master/moto/secretsmanager/models.py#L276

bblommers:
@rafrafek Thanks for letting us know - I've raised #7146 to also replace those methods.

PR #6967, opened 2023-10-29
fix:Create .gitpod.yml

PentesterPriyanshu:
added gitpod configuration file for this repo which is missing


PentesterPriyanshu:
@bblommers   review it kindly and merge it

bblommers:
We already support [Github Codespaces](https://github.com/features/codespaces) - I don't intend to support other tools like this.

PentesterPriyanshu:
thanks for reply but in opensource many use it

PR #6975, opened 2023-10-31
Update server_mode.rst with shell usage example

tomers:
Suggest to create a `aws` alias which contains the `--endpoint-url http://localhost:5000` command-line option


bblommers (changes requested):
It does make sense to have a CLI example here, but I think that we should just show that the parameter exists. E.g.:

```
aws ec2 describe-instances --endpoint-url http://localhost:5000
```

If a user wants to create an alias for that, that is up to them.

PR #6977, opened 2023-10-31
Lambda: allow <fn_name>:<alias_name>

matusfaro:
Most calls that allow a `Qualifier` also support passing in an alias as part of the name.

From [AWS docs for get-policy](https://docs.aws.amazon.com/lambda/latest/dg/API_GetPolicy.html) as an example:
```
Name formats
    Function name – my-function (name-only), my-function:v1 (with alias).
    Function ARN – arn:aws:lambda:us-west-2:123456789012:function:my-function.
    Partial ARN – 123456789012:function:my-function.
```

This PR extends functionality to allow passing `my-function:v1` as described in docs.

Limitations:
- Not all API endpoints allow alias, but this change would allow passing it in. It may have some undefined behavior if you use an alias for an endpoint that disallows it. 
- In case we supply both an alias within a name (e.g. `my-function:v1`) AND at the same time provide a qualifier (e.g. `v1`), currently I prefer the alias within the name over the provided qualifier.


bblommers:
Hi @matusfaro, thanks for the PR, and welcome to Moto!

> It may have some undefined behavior if you use an alias for an endpoint that disallows it.
>

That are quite a few endpoints, from what I can see, so I think we should only enable this feature for operations that do allow it.
Assuming the documentation is up-to-date, and I haven't missed anything, that means only enabling it for the following operations:
 - add_permission
 - delete_function
 - get_function
 - get_policy
 - invoke
 - remove_permission

Maybe we can just add an optional argument (`allow_alias`) to the function? I'm not too hung up on the exact implementation though, that's just a suggestion.

matusfaro:
@bblommers 

I changed it up and fixed up a bunch of things:

- Explicit method that allows a qualifier and one that will throw an exception if qualifier is provided:
    - get_function_by_name_or_arn (Will throw when qualifier is specified)
    - get_function_by_name_or_arn_with_qualifier
- Support passing in function name with a qualifier for endpoints that allow it
- Throwing ResourceNotFoundException for endpoints that requires a function but it doesn't exist. Includes: invoke, getPolicy, UpdateFunctionConfiguration
- UnknownFunctionException may have returned an ARN with double qualifier e.g. `...:1:1`. Occurs when you pass in an ARN with an existing qualifier

moto/awslambda/models.py
```diff
@@ -1349,8 +1349,9 @@ def _get_version(self, name: str, version: str) -> Optional[LambdaFunction]:
         return None
 
     def _get_function_aliases(self, function_name: str) -> Dict[str, LambdaAlias]:
-        fn = self.get_function_by_name_or_arn(function_name)
-        return self._aliases[fn.function_arn]
+        fn = self.get_function_by_name_or_arn_with_qualifier(function_name)
+        [arn, _, _] = self.split_function_arn(fn.function_arn)
```
matusfaro, line 1353: Sanitizing ARN here to remove qualifier

moto/awslambda/models.py
```diff
@@ -1401,34 +1402,73 @@ def update_alias(
...
         aliases = self._get_function_aliases(name)
-
         if qualifier in aliases:
-            alias = aliases[qualifier]
-            return self._get_version(name, alias.function_version)
+            alias_version = aliases[qualifier].function_version
 
-        return None
+            # Find by alias pointing to latest
+            if alias_version.lower() == "$latest":
```
matusfaro, line 1463: I had a case where the alias also pointed to latest, added a case here for that

moto/awslambda/models.py
```diff
@@ -1443,28 +1483,66 @@ def list_aliases(self, function_name: str) -> Iterable[LambdaAlias]:
         return sorted(aliases.values(), key=lambda alias: alias.name)
 
     def get_arn(self, arn: str) -> Optional[LambdaFunction]:
+        [arn_without_qualifier, _, _] = self.split_function_arn(arn)
+        return self._arns.get(arn_without_qualifier, None)
+
+    def split_function_arn(self, arn: str) -> [str, str, Optional[str]]:
```
matusfaro, line 1489: Handy utility to parse an ARN into:
- ARN without qualifier
- function name
- qualifier

moto/awslambda/models.py
```diff
@@ -1539,18 +1621,6 @@ def del_function(self, name_or_arn: str, qualifier: Optional[str] = None) -> Non
             ):
                 del self._functions[name]
 
-        else:
```
matusfaro, line 1542: Simplified logic, merged above

moto/awslambda/models.py
```diff
@@ -1813,9 +1883,7 @@ def create_event_source_mapping(self, spec: Dict[str, Any]) -> EventSourceMappin
                 raise RESTError("InvalidParameterValueException", f"Missing {param}")
 
         # Validate function name
-        func = self._lambdas.get_function_by_name_or_arn(spec.get("FunctionName", ""))
-        if not func:
-            raise RESTError("ResourceNotFoundException", "Invalid FunctionName")
```
matusfaro, line 1818: The exception is now thrown within `get_function_by_name_or_arn_with_qualifier`, `func` is never `None` here

moto/awslambda/models.py
```diff
@@ -2100,8 +2168,6 @@ def get_code_signing_config(self, function_name: str) -> Dict[str, Any]:
 
     def get_policy(self, function_name: str) -> str:
         fn = self.get_function(function_name)
-        if not fn:
```
matusfaro, line 2103: `fn` is never `None`, get function throws exception if not found

moto/awslambda/models.py
```diff
@@ -2120,7 +2186,7 @@ def update_function_configuration(
     ) -> Optional[Dict[str, Any]]:
         fn = self.get_function(function_name, qualifier)
 
-        return fn.update_configuration(body) if fn else None
```
matusfaro, line 2123: `fn` is no longer never `None`

moto/awslambda/models.py
```diff
@@ -2134,12 +2200,9 @@ def invoke(
         Invoking a Function with PackageType=Image is not yet supported.
         """
         fn = self.get_function(function_name, qualifier)
-        if fn:
```
matusfaro, line 2137: fn cannot be none

moto/awslambda/responses.py
```diff
@@ -397,7 +397,15 @@ def _delete_function(self) -> TYPE_RESPONSE:
         return 204, {}, ""
 
     @staticmethod
-    def _set_configuration_qualifier(configuration: Dict[str, Any], qualifier: str) -> Dict[str, Any]:  # type: ignore[misc]
+    def _set_configuration_qualifier(configuration: Dict[str, Any], function_name: str, qualifier: str) -> Dict[str, Any]:  # type: ignore[misc]
```
matusfaro, line 400: I'm not sure I understand the original intent of injecting the qualifier into the configuration here, but I simply extended it to fetch the qualifier from the name itself if present.

bblommers:
@matusfaro I just had a quick look, and I think this all makes sense.

Would you mind resolving the merge conflicts? I'd like to confirm that the existing tests pass - after that I'll have another closer look.

matusfaro:
@bblommers Ready for review:
- Merged from master
- Added extra helpful comments
- Renamed methods to indicate qualifier is not allowed: `get_function_by_name` -> `get_function_by_name_forbid_qualifier`
- Fixed linting issues

bblommers:
@matusfaro The tests are currently failing, because the `delete_function` now behaves differently.

It only fails in ServerMode because the state isn't properly deleted in-between tests, and that causes the `test_publish_function` to trip over functions that were not properly deleted in previous tests. 
(Moto does a hard reset after every test when running the tests normally, so then it's not noticeable.)

To reproduce this, when running the tests normally:
In the `test_lambda.py::test_delete_function`, add the following check at the end:
```
func_list = conn.list_functions(FunctionVersion="ALL")["Functions"]
our_functions = [f for f in func_list if f["FunctionName"] == function_name]
assert len(our_functions) == 0
```

That will throw the same failure as in ServerMode, and shows that the function is not deleted properly.

matusfaro:
@bblommers Good catch, sorry about that bug.

I misunderstood how the functions and qualifiers were stored and misused `function.version`.

I'm still getting used to this code base and Python actually. I understand if you want to be cautious about merging this one in, let me know if you want me to test any other part in more depth.

bblommers (approved):
From what I can tell, this is now all in-line with how AWS behaves.

Thank you for adding this functionality!

PR #6994, opened 2023-11-06
RDS: detect rds instance / cluster engine mismatch

rafcio19:
Resolves https://github.com/getmoto/moto/issues/6993


bblommers (approved):
That was quick!

LGTM - thanks @rafcio19!

PR #6995, opened 2023-11-06
Lambda: Allow Qualifier on get policy call

matusfaro:
As per [AWS docs](https://docs.aws.amazon.com/lambda/latest/dg/API_GetPolicy.html#API_GetPolicy_RequestSyntax), the get policy call also accepts a Qualifier parameter as well as an function ARN.

Notes:
- Although the ARN and qualifier is accepted, a larger change will be required to return different policies based on that qualifier which I haven't done as part of this change
- There is already a test `test_get_policy_with_qualifier` that calls this endpoint with a qualifier, although it doesn't test that the returned policy is different with and without the qualifier.

### DIFF
--- a/moto/awslambda/models.py
+++ b/moto/awslambda/models.py
@@ -2098,8 +2098,8 @@ class LambdaBackend(BaseBackend):
         fn = self.get_function(function_name)
         return fn.get_code_signing_config()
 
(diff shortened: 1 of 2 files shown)


bblommers (approved):
LGTM - thanks for adding this @matusfaro!

PR #6999, opened 2023-11-07
ECR: image digest fix

rafcio19:
Resolves https://github.com/getmoto/moto/issues/6997


bblommers (approved):
Awesome, thank you for fixing this @rafcio19!

PR #7004, opened 2023-11-09
EC2: Fix VPC endpoint default policy

viren-nadkarni:
This PR fixes a bad key in default policy for VPC endpoints.


bblommers (approved):
Nice catch @viren-nadkarni

PR #7007, opened 2023-11-09
ECS: fix container definition validation

rafcio19:
Resolves https://github.com/getmoto/moto/issues/7006


bblommers (changes requested):
The tests failures look legitimate @rafcio19 - I'm guessing the CloudFormation templates that we use in our tests are now invalid with these changes

bblommers (approved):
Awesome - thank you @rafcio19!

PR #7009, opened 2023-11-09
Add new ELBv2 ssl protocols, add small helper script to fetch them

dfangl:
## Motivation
The SSL_POLICIES constant holding the ssl protocols is outdated.

## Changes
* Add small helper script to fetch the protocols from AWS and transform them to the format currently used
* Update the `SSL_POLICIES` constants with the latest protocols available (in my AWS account)


scripts/update_ssl_policies.py
```diff
@@ -0,0 +1,41 @@
...
+        if key in KEY_BLACKLIST:
+            continue
+        new_key = camel_case_to_snake_case(key)
+        result[new_key] = transform_policies(value)
+    return result
+
+
+def main():
+    policies = get_ssl_elb_ssl_policies()
+    transformed_policies = transform_policies(policies)
```
bblommers, line 36: Can we sort the output, before printing? 

If AWS happens to completely change the output order, having it sorted on our side would make it easier to track the changes

bblommers (changes requested):
We have a test for this method, `test_describe_ssl_policies`, and that's currently failing.

Maybe we can simplify that test to check that Moto returns `> 0` policies? I don't think there's a need to assert on the exact number of expected SSL policies.

scripts/update_ssl_policies.py
dfangl: Sure, good point!

dfangl: I sorted only the dictionary keys, not the list items itself, as those seem ordered by design.

dfangl:
Thank you for the review, I incorporated the fixes!

dfangl:
Build is still failing, but since it happens only for python 3.12, possibly a flake?

bblommers:
> Build is still failing, but since it happens only for python 3.12, possibly a flake?

Yes, it is - this looks good to me. Thank you @dfangl!

bblommers (approved):

PR #7010, opened 2023-11-09
Add ACM support for ResourceGroupsTaggingAPI

afuentes-19:
### Background

The ResourcesGroupsTagging API method `_get_resources_generator()` in the `models.py` is used to retrieve the resources and their tags. However, there is currently no support for retrieving ACM resources and their tags with this API. 

## Problem 

The ResourcesGroupsTagging API returns an empty response after creating mocked ACM resources:

Screenshot 2023-11-09 at 12 25 57 PM

## Solution

The solution is to simply add the ACM backend log in the `_get_resources_generator()` to retrieve ACM resources and their tags: 

## Testing 
To test, we use the moto functions to start mocking: 
`moto.mock_resourcegroupstaggingapi().start()`

Then we use boto to make the API call: 

```       
client = boto3.client('resourcegroupstaggingapi')
response = client.get_resources()
logging.info(f"resource tagging response: {response}")
```
We ran the unit tests with no failures associated with our changes.


bblommers (changes requested):
Hi @afuentes-19, welcome to Moto! This PR looks good to me so far.

Can you add a test for this behaviour, to verify it works?

It looks like the linter is currently failing as well:
https://github.com/getmoto/moto/actions/runs/6815599637/job/18535131316?pr=7010

afuentes-19:
Hi @bblommers. With the linting problem, the issue stems from the ACM backend retrieving the certificate tags with a `Dict[str, Optional[str]]` type. However, the `format_tags()` function in the resourcegroupstaggingapi requires tags to be of type `Dict[str, str]`. I am wondering whether to modify the ACM backend certificate tags to be of type `Dict[str, str]` or to modify the `format_tags()` function parameter type. Another potential option is to manually extract the keys and values of the ACM tags and create a new dictionary that satisfies the `format_tags()` parameter type. However, in the case of there being a `None` value for some of the ACM tags, such as the ARN, should this value be changed to an empty string or `"None"`. I appreciate your feedback, thanks!

bblommers:
Good question @afuentes-19. The `get_resources` API actually explicitly allows filtering tags with `Value=None`, so I think we should change the definition of the `format_tags` function completely:

`def format_tags(tags: Dict[str, Optional[str]]) -> List[Dict[str, Optional[str]]]`

That obviously has ramifications for the `tag_filter`-method as well, and maybe even the filters themselves. ACM may be the first service (on our side) that allows for key-only tags, but I'm sure there are others. So it makes sense to me for the `ResourceGroupsTaggingAPI` to be able to handle this case natively/without any ACM-specific logic.

afuentes-19:
Hi [@bblommers](https://github.com/bblommers), I wanted to ask about the linter and PR checks in general for moto.

- I've realized there are many linting errors unrelated to my changes - is this usual and do I need to resolve these? When ran locally, I get `Found 45 errors in 25 files (checked 864 source files)`. Also, I have the same question for unit tests as I see there is a unit test failing unrelated to my recent changes.
- Also, I modified the format_tags and related functions ,`tag_filter()` and `format_tag_keys()`, to accept type `Dict[str, Optional[str]]`, but that gave me more linting errors for the other services in the get_resources API.

For example:
`moto/resourcegroupstaggingapi/models.py:167:24: error: Incompatible types in assignment (expression has type "list[dict[str, str]]", variable has type "list[dict[str, Optional[str]]]") [assignment]`
There are 8 other similar errors and they are for other services than ACM - should the tags for these services also be modified? I thought variables of type `Dict[str, str]` would be okay for functions that accept `Dict[str, Optional[str]`, but I am open to suggestions here to resolve this.

bblommers:
What version of `mypy` are you on, and what kind of errors are you seeing? There shouldn't be any, because that check passes on our master branch.

As far as unit tests go - there are a few flaky tests, and also tests that absolutely need Docker to be installed/working. So that may be the reason it is failing locally.

moto/resourcegroupstaggingapi/models.py
```diff
@@ -131,20 +136,31 @@ def tag_filter(tag_list: List[Dict[str, str]]) -> bool:
...
+        ) -> List[Dict[str, Optional[str]]]:
             result = []
             for tag in tags:
                 result.append({"Key": tag[keys[0]], "Value": tag[keys[1]]})
             return result
 
+        # ACM
+        if not resource_type_filters or "acm" in resource_type_filters:
+            for certificate in self.acm_backend._certificates.values():
+                tags = format_tags(certificate.tags)
```
bblommers, line 158: Can you try to rename this variable `tags` to something service-specific, maybe `acm_tags`? I wonder if that fixes the linting problems.

We're essentially reusing the `tags` variable for all services, so I wonder if the problem is just that `mypy` is unhappy with the fact we're re-assigning the variable to different types.

afuentes-19: I am on mypy `v.1.5.1`. I changed the variable to `acm_tags` but still getting the same errors.

bblommers: Hope you don't mind - I just pushed a 'fix' by just changing it to `Dict[str, Any]`. I had already made the change anyway, so there was no point in undoing all that, and then asking you to make the same changes..

It does feel like a `mypy` bug/shortcoming, that it doesn't recognize that the two should be interchangeable.

Resorting to `Any` is not pretty, but at least it works. I've left a note in there (mostly to myself) to replace it with TypedDict later on, that should fix it completely.

afuentes-19: Great! Thank you - I don't mind at all. The linting check is passing now, just failing on the testing check now. 
My unit test here is passing:
`tests/test_resourcegroupstaggingapi/test_resourcegroupstaggingapi.py::test_get_resources_acm PASSED`

bblommers: I've restarted the build, and it passed now. It's still marked as draft, I don't know if you wanted to add anything else? As far as I'm concerned this is good to go!

afuentes-19: Yep, nothing else to add. Gonna change it to ready for review - thanks for your help!

bblommers (approved):
LGTM - thank you for adding this to Moto @afuentes-19!

PR #7013, opened 2023-11-10
Fix EFS access point ID pattern

rpmcginty:
## Details

https://github.com/getmoto/moto/issues/7014

EFS Access point ID string value does not match valid access point ID pattern as described in [AWS EFS documentation](https://docs.aws.amazon.com/efs/latest/ug/API_AccessPointDescription.htm).

EFS access point ID has the following regex pattern:  `fsap-[0-9a-f]{8,40}`

The previous implementation returns a 8 character hex string *without* the `fsap-` prefix. 

**Proposed change**:
- (old) ~`abcd1234`~
- (new) `fsap-abcd1234`

### DIFF
--- a/moto/efs/models.py
+++ b/moto/efs/models.py
@@ -54,8 +54,8 @@ class AccessPoint(BaseModel):
         root_directory: Dict[str, str],
         context: "EFSBackend",
     ):
-        self.access_point_id = mock_random.get_random_hex(8)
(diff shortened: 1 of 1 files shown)


bblommers (approved):
LGTM - thank you for fixing this @rpmcginty, and welcome to Moto!

PR #7015, opened 2023-11-10
Implemented identitystore describe_group

toshyak:


toshyak:
Also thought about small refactor of some identity store mocks, e.g. use Group type instead of dict here https://github.com/getmoto/moto/blob/c04ff771786d75bd0b872673898a7d38edc6f461/moto/identitystore/models.py#L109 . Should I add it here or separate PR is better?

toshyak:
hey @bblommers! Done, it's ready for review

bblommers (approved):
LGTM - thank you @toshyak!

PR #7028, opened 2023-11-15
Add subtitle support for `transcribe`

rosscooperman:
Adds support for translating the `Subtitle.Formats` configuration key into appropriate outputs given `OutputBucketName` and `OutputKey` configuration. Also fixes the formatting of `Transcript.TranscriptFileUri` in the output when `OutputBucketName` and `OutputKey` are both set. When that is the case the job name is not part of the output key.

### DIFF
--- a/moto/transcribe/models.py
+++ b/moto/transcribe/models.py
@@ -48,6 +48,7 @@ class FakeTranscriptionJob(BaseObject, ManagedState):
         identify_language: Optional[bool],
         identify_multiple_languages: Optional[bool],
         language_options: Optional[List[str]],
+        subtitles: Optional[Dict[str, Any]],
     ):
         ManagedState.__init__(
             self,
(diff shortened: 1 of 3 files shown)


moto/transcribe/models.py
```diff
@@ -217,15 +221,27 @@ def advance(self) -> None:
...
             )
             if self._output_bucket_name:
-                transcript_file_uri = f"https://s3.{self._region_name}.amazonaws.com/{self._output_bucket_name}/"
-                if self.output_key is not None:
-                    transcript_file_uri += f"{self.output_key}/"
-                transcript_file_uri += f"{self.transcription_job_name}.json"
+                transcript_file_prefix = (
+                    f"https://s3.{self._region_name}.amazonaws.com/"
+                    f"{self._output_bucket_name}/"
+                    f"{(self.output_key or self.transcription_job_name).replace('.json', '')}"
```
bblommers, line 227: Bit of a nitpick - but `.json` could also be part of the filename (`this.is.not.a.json.file.xml`). So we should probably specifically check whether `.json` is at the end.

bblommers (changes requested):

moto/transcribe/models.py
rosscooperman: You caught me in my laziness! Will update.

rosscooperman: K, updated the test to include a mid-key `.json` and only strip the suffix.

bblommers (approved):
LGTM - thank you @rosscooperman!

PR #7030, opened 2023-11-15
Techdebt: Fix compatibility with botocore 1.32.1

bblommers:
Botocore 1.32.1 sends some files in a compressed format, `.json.gz`, in order to save disk space. So we will need to read that file instead.

https://github.com/boto/botocore/blob/develop/.changes/1.32.1.json

We're currently treating the compressed format as a fallback, but in the future we should probably try to load this version first as an optimization.


potiuk:
Nice one! Thanks for quick turnaround on that one !

PR #7036, opened 2023-11-17
Backporting SQS: Support JSON format to moto-3.1.18

ziyanli-amazon:
This is the exact same commit that fixes [Issue](https://github.com/getmoto/moto/issues/6286)
This commit is based off tag [3.1.18](https://github.com/getmoto/moto/tree/3.1.18)
Just back-porting this [PR](https://github.com/getmoto/moto/pull/6331) to version 3.1.18

### DIFF
--- a/moto/core/common_types.py
+++ b/moto/core/common_types.py
@@ -1,0 +1,5 @@
+from typing import Any, Dict, Tuple, TypeVar, Union
+
+
+TYPE_RESPONSE = Tuple[int, Dict[str, Any], Union[str, bytes]]
+TYPE_IF_NONE = TypeVar("TYPE_IF_NONE")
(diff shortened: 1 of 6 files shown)


bblommers:
Hi @ziyanli-amazon! We don't support Moto 3.x anymore - we only develop for the current version.

Are you running into any issues upgrading to 4.x?

ziyanli-amazon:
@bblommers Hi, the reason for this PR is that we still have some internal customers who depend on `3.x`. No, there's no issue for `4.x`. 
Is there any other way we can patch the `3.x` version ?

bblommers:
Hi @ziyanli-amazon, I've just released 3.1.19. It contains this change, plus a few others that were necessary to get our internal tests to pass.

https://pypi.org/manage/project/moto/release/3.1.19/

Just to make sure this doesn't generate any expections:
My earlier statement still stands - `We don't support Moto 3.x anymore - we only develop for the current version.`. 

The decision to release this version was more curiosity then anything else, to see how much effort it would take.

Like I said, (almost) all of our internal tests pass against this version, but let me know if you run into any issues.

This is the PR with all the changes: https://github.com/getmoto/moto/pull/7042

PR #7038, opened 2023-11-17
Lambda: no log response bug

rafcio19:
Resolves https://github.com/getmoto/moto/issues/7033


bblommers (approved):
Awesome - thank you @rafcio19!

PR #7039, opened 2023-11-17
Fix CloudFormation AWS::Logs::LogGroup tagging

jmsanders:
CloudFormation AWS::Logs::LogGroup expects tags as an array of Tags:

https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-logs-loggroup.html https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-resource-tags.html

Whereas boto3's Logs `create_log_group()` expects tags to be a dict:

https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/logs/client/create_log_group.html

So we need to transform how we provide tags to avoid getting:

```
E       AttributeError: 'list' object has no attribute 'items'
```

### DIFF
--- a/moto/logs/models.py
+++ b/moto/logs/models.py
@@ -410,7 +410,9 @@ class LogGroup(CloudFormationModel):
         **kwargs: Any,
     ) -> "LogGroup":
         properties = cloudformation_json["Properties"]
-        tags = properties.get("Tags", {})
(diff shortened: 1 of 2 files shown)


bblommers (approved):
LGTM - thank you for fixing this @jmsanders!

PR #7043, opened 2023-11-18
EFS: Add support for describing access points using File system ID filter

rpmcginty:
## What is in this change?

The current moto implementation of `describe_access_points` for EFS service does support the full functionality described in boto. Namely it does not support filtering results based on file system Id (see [`EFSClient.describe_access_points`](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/efs/client/describe_access_points.html#)).

This change adds support for filtering based on file system id. 

## Testing

added unit tests to cover additional filter scenarios for `efs.describe_access_points`

### DIFF
--- a/moto/efs/models.py
+++ b/moto/efs/models.py
@@ -698,7 +698,11 @@ class EFSBackend(BaseBackend):
         self.access_points[access_point.access_point_id] = access_point
         return access_point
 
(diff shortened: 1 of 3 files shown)


bblommers (approved):
LGTM - thank you for adding this feature @rpmcginty!

PR #7044, opened 2023-11-18
Techdebt: add regions for various tests

tungol:
following the example of https://github.com/getmoto/moto/pull/4194


bblommers (approved):
Makes sense - thank you @tungol!

PR #7045, opened 2023-11-19
MyPy improvements

tungol:
This is my attempt to complete the work from https://github.com/getmoto/moto/pull/6421

I rebased it, and then used a Protocol to allow lazy_load to work properly. If you don't like the protocol, it also works if lazy_load has a return type of base_decorator and we just cast the inner function to that. 

I wasn't sure if maybe the name for the protocol class should be more distinct to avoid confusion.

I also added the other uses of the mock decorators listed in https://github.com/getmoto/moto/issues/4944#issuecomment-1596169486 to test_mypy.py. That was probably not really necessary, but it seemed like a reasonable thing to do.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -5,6 +5,8 @@ exclude_lines =
     raise NotImplemented.
     return NotImplemented
     def __repr__
+    if TYPE_CHECKING:
+    ^\s*\.\.\.$
 
 [run]
 include = moto/*
--- a/setup.cfg
+++ b/setup.cfg
@@ -37,6 +37,7 @@ install_requires =
     responses>=0.13.0
     Jinja2>=2.10.1
     importlib_metadata ; python_version < '3.8'
+    typing_extensions
 package_dir =
     moto = moto
 include_package_data = True
@@ -274,7 +275,7 @@ disable = W,C,R,E
 enable = anomalous-backslash-in-string, arguments-renamed, dangerous-default-value, deprecated-module, function-redefined, import-self, redefined-builtin, redefined-outer-name, reimported, pointless-statement, super-with-arguments, unused-argument, unused-import, unused-variable, useless-else-on-loop, wildcard-import
 
 [mypy]
-files= moto, tests/test_core/test_mock_all.py, tests/test_core/test_decorator_calls.py, tests/test_core/test_responses_module.py
+files= moto, tests/test_core/test_mock_all.py, tests/test_core/test_decorator_calls.py, tests/test_core/test_responses_module.py, tests/test_core/test_mypy.py
 show_column_numbers=True
 show_error_codes = True
 disable_error_code=abstract
--- a/setup.cfg
+++ b/setup.cfg
@@ -37,7 +37,6 @@ install_requires =
     responses>=0.13.0
     Jinja2>=2.10.1
     importlib_metadata ; python_version < '3.8'
-    typing_extensions
 package_dir =
     moto = moto
 include_package_data = True
--- a/tests/test_core/test_mypy.py
+++ b/tests/test_core/test_mypy.py
@@ -10,6 +10,12 @@ def test_without_parentheses() -> int:
     return 123
 
 
+@mock_s3()
+def test_with_parentheses() -> int:
+    assert boto3.client("s3").list_buckets()["Buckets"] == []
+    return 456
+
+
 @mock_s3
 def test_no_return() -> None:
     assert boto3.client("s3").list_buckets()["Buckets"] == []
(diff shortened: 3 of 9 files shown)


.coveragerc
```diff
@@ -5,6 +5,8 @@ exclude_lines =
     raise NotImplemented.
     return NotImplemented
     def __repr__
+    if TYPE_CHECKING:
+    ^\s*\.\.\.$
```
tungol, line 9: I hope this is an okay way to handle the MR coverage percent check. This excludes the TYPE_CHECKING imports as well as any line that is only an ellipsis

tests/test_core/test_mypy.py
```diff
@@ -0,0 +1,33 @@
+import boto3
+
+from moto import mock_s3
+from moto.core.models import BaseMockAWS
+
+
+@mock_s3
+def test_without_parentheses() -> int:
```
tungol, line 8: @bblommers The name of this test is from your original draft MR, but I wasn't totally sure of it's significance. Was there a version of this *with* parentheses that you'd intended to add?

tests/test_core/test_responses_module.py
```diff
@@ -17,8 +17,8 @@ def setUp(self) -> None:
         if settings.TEST_SERVER_MODE:
             raise SkipTest("No point in testing responses-decorator in ServerMode")
 
-    @mock_s3
     @responses.activate
+    @mock_s3
```
tungol, line 21: Mypy showed a typing error when mock_s3 came first. I'm not sure what the cause was, and that might be something worth checking out

bblommers: For future reference, the MyPY error:
`Type of decorated function contains type "Any" ("Callable[[VarArg(Any), KwArg(Any)], Any]")`

I _think_ this is due to the fact that the `responses`-decorator actually has `Any` in their typing declarations - see for example here:

https://github.com/getsentry/responses/blob/4fa24b7b2342cd756c9b9533d8036132eb11c91b/responses/__init__.py#L990

This method tests a very specific scenario where Moto's decorator comes first  - hence the name, `test_moto_first`. (The next test has the exact same body, but with the decorators reversed - `test_moto_second`).

So to ensure we're not losing any code/scenario coverage, I think this should be reversed. Maybe we can just add a `type: ignore` to get past the fact that `responses` is to loose in their annotations.

tests/test_core/test_mypy.py
bblommers: Yes - it's not used often, but the following is a valid test case:

```
@mock_s3()
def test_with_parentheses() -> int:
    assert boto3.client("s3").list_buckets()["Buckets"] == []
    return 123
```

Can you add that to this PR? It passes the MyPy check, so it's a good verification that all possible Moto usages are covered.

setup.cfg
```diff
@@ -37,6 +37,7 @@ install_requires =
     responses>=0.13.0
     Jinja2>=2.10.1
     importlib_metadata ; python_version < '3.8'
+    typing_extensions
```
bblommers, line 40: Having to introduce this is unfortunate, but unavoidable. I will probably only merge this PR as part of a minor release (4.3.0), to reduce the impact.  (I.e., reduce the number of users that do not update their dependencies along with Moto, and then complain about missing imports..)

bblommers (changes requested):
This is exciting! I remember looking into `Protocol`, but I obviously couldn't quite figure out how that would solve the problem.

I've left a few minor comments, but overall it looks good.

setup.cfg
tungol: I think we can avoid it as a runtime dependency, by gating it behind `if TYPE_CHECKING:`, if you think that's better.

On the other hand, I see that it's already in use in several other places in moto without it being listed here. It's been imported by `moto/identitystore/models.py` since June, and by `moto/utilities/aws_headers.py` April, but there is behind the TYPE_CHECKING gate

If you want to avoid the runtime dependency, I'll fix `moto/identitystore/models.py` too.

tests/test_core/test_mypy.py
tungol: makes sense, I'll have that change pushed up soon.

tests/test_core/test_responses_module.py
tungol: Ah ha! That's what I get for fixing the type error without really paying attention to what it was. I'll stare at it a little longer and put the ignore pragma on here if I can't figure out another solution.

tungol:
I pushed an updated to my branch to fix the test failures, and it hasn't shown up in the MR here. Not sure why or how to fix that.

bblommers (approved):
LGTM - thank you so much for fixing this @tungol!

PR #7052, opened 2023-11-20
LakeFormation: raise AlreadyExistsException when registering an existing resource

JohannesKoenigTMH:
The register-resource method of lakeformation is missing the functionality to raise AlreadyExistsException when a resource is already registered.

Resource:
- https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/lakeformation/client/register_resource.html,
- Testing with AWS CLI


bblommers (approved):
Nice one - additional validation is always welcome.

PR #7053, opened 2023-11-22
Mypy improvements for lazy_load

tungol:
Following up on my previous MR ( https://github.com/getmoto/moto/pull/7045 ), here's what it looks like to get XRaySegment back on moto.lazy_load. It's a little bit of a funky overload, and I also enabled `warn_unused_ignores` in mypy so that I could construct a negative test for the typing on XRaySegment.

I won't mind if you decide not to take this MR, but I wanted to see if it could be done, at least.

Edit: warn_unused_ignores made mypy's approval more fragile than I realized. I walked that back to be just for test_mypy.py, and put back in some of the ignores that I had originally removed. I kept in the removal of a lot of them though, since I think they're genuinely no longer needed. 

I've made the commit history clean; one commit for removing a bunch of `type: ignore` pragmas and one for the lazy_load stuff.

### DIFF
--- a/moto/__init__.py
+++ b/moto/__init__.py
@@ -1,14 +1,13 @@
 import importlib
 import sys
-from contextlib import ContextDecorator
+from contextlib import ContextDecorator, AbstractContextManager
 
-from moto.core.models import BaseMockAWS, base_decorator, BaseDecorator
+from moto.core.models import BaseMockAWS, BaseDecorator
 from typing import Any, Callable, List, Optional, TypeVar, Union, overload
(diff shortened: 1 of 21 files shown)


tungol:
Actually I think this solution is not very elegant, and I'll just close this unless I hit on a better way to handle it.

bblommers:
Thanks for looking into it regardless @tungol!

FWIW, the current solution doesn't bother me that much. It is a special usecase after all, quite different from the regular `mock_service`-decorators, so it makes sense to me that the typing approach is also different.

PR #7056, opened 2023-11-22
Core: Parameter handling: fix parsing of >= 10 members

steffyP:
Fix for https://github.com/localstack/localstack/issues/9663; related to https://github.com/getmoto/moto/issues/6033: 

When more than 10 members need to be parsed, the string-sort leads to `IndexError`. 

For `Tags` this was already tackled and fixed in https://github.com/getmoto/moto/pull/6038. 

* The proposed solution should handle more generic cases. 
   * However, there may still be issues with nested members.

* Added a test case for autoscaling-tags, to verify the tag-parsing still works.
* Added a test for cloudwatch with a couple of metric-queries to cover the case reported in https://github.com/localstack/localstack/issues/9663

### DIFF
--- a/moto/core/utils.py
+++ b/moto/core/utils.py
@@ -326,16 +326,19 @@ def extract_region_from_aws_authorization(string: str) -> Optional[str]:
     return region
 
 
-def params_sort_function(item: Tuple[str, Any]) -> Tuple[str, Any]:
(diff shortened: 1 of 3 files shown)


bblommers (approved):
LGTM - thank you @steffyP!

PR #7066, opened 2023-11-26
SSM: fix overwrite preserving non-indicated values

hanseltime:
# Summary

This fixes a bug I stumbled across when running localstack (which runs moto server in the background) and using SSM parameters.

The Bug:  overwrite does not just change the fields supplied.  It instead, replaces all values that were not explicitily specified.  This does not happen when the same code is tested against AWS.

Changes:

1. Added a test for verifying that metadata is preserved when an overwrite occurs
2. Coalesce all Parameter values to the previous parameter if they are None type
3. update the response model to not default Tags to any empty array since that's meaningful for erasure (it is coalesced in the backend anyway)
4. Add some warnings about `Tier` and `Policies` configuration options since they seem to be new comparatively

### DIFF
--- a/moto/ssm/models.py
+++ b/moto/ssm/models.py
@@ -2106,6 +2106,24 @@ class SimpleSystemManagerBackend(BaseBackend):
                 self._check_for_parameter_version_limit_exception(name)
                 previous_parameter_versions.pop(0)
 
+            # Ensure all the previous values that we didn't overwrite are preserved
+            value = value if value is not None else previous_parameter.value
+            description = (
+                description
+                if description is not None
+                else previous_parameter.description
+            )
+            allowed_pattern = (
+                allowed_pattern
+                if allowed_pattern is not None
+                else previous_parameter.allowed_pattern
+            )
+            keyid = keyid if keyid is not None else previous_parameter.keyid
+            tags = tags if tags is not None else previous_parameter.tags
+            data_type = (
+                data_type if data_type is not None else previous_parameter.data_type
+            )
+
         last_modified_date = time.time()
         self._parameters[name].append(
             Parameter(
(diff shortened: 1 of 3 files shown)


hanseltime:
I've added a test to make sure that we're calling the warnings I added since the patch coverage requires that I do that on a percentage basis.  Please let me know if we don't want that and I can revert.

hanseltime:
So my tests are failing consistently inconsistently due to this error: 

```
assert 'error runnin...est unknown")' == 'Simple Lambda happy path OK'
  - Simple Lambda happy path OK
  + error running docker: 404 Client Error for http+docker://localhost/v1.43/images/create?tag=3.11&fromImage=mlupin%2Fdocker-lambda: Not Found ("manifest for mlupin/docker-lambda:3.11 not found: manifest unknown: manifest unknown")
tests/test_awslambda_simple/test_lambda_simple.py::test_run_function FAILED
```

https://github.com/getmoto/moto/actions/runs/7007205268/job/19061597718

This one looks like an issue with docker on the runner.

Addtionally, I just got an intermittent failure for 
```
FAILED tests/test_ec2/test_elastic_network_interfaces.py::test_unassign_private_ip_addresses__multiple - AssertionError: assert 3 == 4
 +  where 3 = len([{'Primary': True, 'PrivateIpAddress': '54.0.0.1'}, {'Primary': False, 'PrivateIpAddress': '10.0.21.188'}, {'Primary': False, 'PrivateIpAddress': '10.0.44.25'}])
 ```
 
 I'm at a loss here since my fix passes tests locally and these errors are in tests that are completely unrelated to my code paths.  I also don't have the ability to retry jobs.  How can we get this resolved?  Does someone just need to re-run or does this PR need to wait for these CI bugs to be fixed?

bblommers:
Hi @hanseltime, don't worry about the Docker failures - I think it's an issue with the CI itself where Docker is sometimes unavailable. 

Thanks for the PR - these changes LGTM!

bblommers (approved):

PR #7069, opened 2023-11-27
typing for backends

tungol:
This allows typing to work properly when starting with a BackendDict object, e.g. `s3_backends[account_id]["global"].create_bucket(resource_name, region_name)`

Most functions like this were ending up as Any before, but now mypy can figure out what they are

This turned up a lot of mypy errors, which I mostly didn't try to resolve. A lot of them are in create_from_cloudformation_json() functions; I suspect those functions often aren't covered by a test? They range from obvious errors (wrong number of arguments in the call) to more complicated/confusing situations. At any rate, lots of type: ignore pragmas that can be followed up on later.


tungol:
As a side note, how would you feel about an MR enabling isort? I'm used to import sorting, and the imports in a lot of these modules feel pretty messy to me. But I don't want to step on your toes if there's a reason you're not using that already.

tungol:
I'll move this to draft while you work out what can be fixed.

bblommers:
OK, I've merged #7072 with some improvements @tungol . There will still need to be some `type: ignore`'s, but the important (obvious) bugs are now fixed at least.

moto/awslambda/models.py
```diff
@@ -1276,7 +1276,7 @@ def update_from_cloudformation_json(  # type: ignore[misc]
         cloudformation_json: Dict[str, Any],
         account_id: str,
         region_name: str,
-    ) -> "EventSourceMapping":
+    ) -> "Optional[EventSourceMapping]":
```
tungol, line 1279: We'd get none if the json dict passed in tried to create an event source mapping for a lambda function that doesn't actually exist.

making this Optional fixes the issue, but maybe the None path should raise an error instead? I'm not sure how moto generally handles this type of situation

moto/awslambda/models.py
```diff
@@ -1326,12 +1326,14 @@ def create_from_cloudformation_json(  # type: ignore[misc]
         account_id: str,
         region_name: str,
         **kwargs: Any,
-    ) -> "LambdaVersion":
+    ) -> "Optional[LambdaVersion]":
```
tungol, line 1329: same as before: what should moto do if someone tries to create a version of a lambda that doesn't exist? We can either return None or raise an error.

This same situation comes up a couple more times in this MR.

moto/ec2/models/subnets.py
```diff
@@ -411,7 +411,7 @@ def create_subnet(
 
     def describe_subnets(
         self, subnet_ids: Optional[List[str]] = None, filters: Optional[Any] = None
-    ) -> Iterable[Subnet]:
+    ) -> List[Subnet]:
```
tungol, line 414: Not sure why this was set to Iterable in the first place, really

moto/rds/models.py
```diff
@@ -619,12 +622,11 @@ def __init__(self, **kwargs: Any):
             self.availability_zone = f"{self.region_name}a"
         self.multi_az = kwargs.get("multi_az")
         self.db_subnet_group_name = kwargs.get("db_subnet_group_name")
+        self.db_subnet_group = None
```
tungol, line 625: setting this first instead of in the `else` branch means that mypy figures out it's an Optional type automatically

moto/rds/models.py
```diff
@@ -1899,18 +1901,16 @@ def create_subnet_group(
         self.subnet_groups[subnet_name] = subnet_group
         return subnet_group
 
-    def describe_db_subnet_groups(
-        self, subnet_group_name: str
-    ) -> Iterable[SubnetGroup]:
+    def describe_db_subnet_groups(self, subnet_group_name: str) -> List[SubnetGroup]:
```
tungol, line 1904: If you want to avoid incurring the conversion to list on line 1910, this could probably be `describe_db_subnet_groups(self, subnet_group_name: Optional[str] = None)` and then I can type it with an overload.

moto/awslambda/models.py
bblommers: It should raise an error, ideally. There will be lots of scenarios like this, where we should validate how AWS behaves and then reproduce it in Moto.

Considering the effort involved, I'd vote to just `type: ignore` it for now. If/when someone runs into this error, we can always have another look. Same goes for the other scenarios.

moto/rds/models.py
bblommers: Nah, that's not necessary. I don't expect there to be ever more than 10 results, so the overhead for the list-conversion is negligible

moto/__init__.py
```diff
@@ -175,7 +175,7 @@ def f() -> "xray_segment_type":
 mock_quicksight = lazy_load(".quicksight", "mock_quicksight")
 mock_ram = lazy_load(".ram", "mock_ram")
 mock_rds = lazy_load(".rds", "mock_rds")
-mock_rdsdata = lazy_load(".rdsdata", "mock_rdsdata")
+mock_rdsdata = lazy_load(".rdsdata", "mock_rdsdata", boto3_name="rds-data")
```
bblommers, line 178: The ServerMode tests are currently failing for the `rds-data` service - is that related to this change?

tungol: I'll back it out and we'll see. That was just a small thing that I noticed and (thought) would be an easy cleanup, but I probably didn't fully understand it.

PR #7075, opened 2023-11-29
sorting imports with ruff

tungol:
This MR enables the "I" set of rules. Three commits here: 

1. update configuration
2. apply auto-fix (large diff)
3. add the second commit to .git-blame-ignore-revs


bblommers:
Oof.. this is going to take me a while to get through! I'll try to do it ASAP though, to reduce the number of conflicts that will show up if we start merging other PR's first.

tungol:
If you'd prefer, I think I can set it incrementally and break this up into a bunch of different MRs by sub-module, so it can get a little more spread out over time.

I will admit that I did not review every line of the diff in here before pushing up the MR...

bblommers:
No probs @tungol. I've gone through half of it so far without finding any issues, so with a little luck I can finish the review either tomorrow or the day after.

bblommers (approved):
LGTM - thank you @tungol!

PR #7076, opened 2023-11-29
Expose ThreadedMotoServer publicly

ls-jad-elkik:
When running with pyright we'd need to ignore the typing rule with such a rule:

```
from moto.server import ThreadedMotoServer  # pyright: ignore [reportPrivateImportUsage]
```

This PR makes sure it's exposed correctly as documented in [this document](https://github.com/microsoft/pyright/blob/main/docs/typed-libraries.md#library-interface):

>
> Imported symbols are considered private by default. If they use the “import A as A” (a redundant module alias), “from X import A as A” (a redundant symbol alias), or “from . import A” forms, symbol “A” is not private unless the name begins with an underscore. If a file __init__.py uses the form “from .A import X”, symbol “A” is not private unless the name begins with an underscore (but “X” is still private). If a wildcard import (of the form “from X import *”) is used, all symbols referenced by the wildcard are not private.
>

### DIFF
--- a/moto/server.py
+++ b/moto/server.py
@@ -11,8 +11,10 @@ from moto.moto_server.werkzeug_app import (
     DomainDispatcherApplication,
     create_backend_app,
 )
-from moto.moto_server.threaded_moto_server import (  # noqa # pylint: disable=unused-import
-    ThreadedMotoServer,
+
+# Expose ThreadedMotoServer as a public symbol according to PEP-561
+from moto.moto_server.threaded_moto_server import (  # pylint: disable=unused-import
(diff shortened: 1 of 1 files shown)


bblommers:
Hi @ls-jad-elkik! I don't intend to officially support `pyright`, but I'd be happy to get this in - it seems like a relativily simple change.

Can you pull in the latest master? There are some merge conflicts that need to be resolved first.

ls-jad-elkik:
Hi @bblommers, thanks for checking this out.

It seems to me that it's a PEP standard, so it's likely that other type checkers might adopt this convention as well. So it might be worth it either way.

I'll update my branch now.

PR #7084, opened 2023-12-01
Techdebt: Update boto3_name to 'rds-data'

tungol:
This is consistent with redshift-data, cloudtrail-data, etc

Plus a few stray tweaks to testing that I made in the process of debugging.

I figured out what I was missing before; it was the matching change in backends_index.py. I guess it probably doesn't
matter since apparently nothing was broken by 'rdsdata' not matching the name used by boto3. But it didn't seem right to me.

### DIFF
--- a/scripts/ci_moto_server.sh
+++ b/scripts/ci_moto_server.sh
@@ -2,4 +2,4 @@
 
 set -e
 pip install $(ls /moto/dist/moto*.gz)[server,all]
-moto_server -H 0.0.0.0 > /moto/server_output.log 2>&1
+moto_server -H 0.0.0.0 | tee /moto/server_output.log 2>&1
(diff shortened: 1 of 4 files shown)


scripts/ci_moto_server.sh
```diff
@@ -2,4 +2,4 @@
 
 set -e
 pip install $(ls /moto/dist/moto*.gz)[server,all]
-moto_server -H 0.0.0.0 > /moto/server_output.log 2>&1
```
tungol, line 5: this just made it easier to see the errors from the server side while running the docker container locally not in detached mode

scripts/ci_moto_server.sh
```diff
@@ -2,4 +2,4 @@
...
+moto_server -H 0.0.0.0 | tee /moto/server_output.log 2>&1
```
tungol, line 5: This just made it easier to see the errors from the server side while running the docker container locally not in detached mode

bblommers (approved):
Nice one!

PR #7091, opened 2023-12-03
Improve typing for IAM

tungol:
This MR cleans up as many ignores in moto.iam as I felt like I could do safely. I also enabled additional mypy strictness locally to see what else might turn up in there.

Of particular note:

 - updates moto.core.common_models.ConfigQueryModel to be a generic type, parameterized by the kind of backend in use
 - mypy --install-types wasn't installing up botocore-stubs, so I added that to the requirements-dev.txt file
 - I refactored some logic in moto/iam/config.py so that mypy could follow what was going on. It should be identical in effect, but the flow of conditionals was too complicated before.
 - Separate from the issue of typing per se, I added my solution for the issue of outside users manually setting AccessKey.last_used (of which I am one in one codebase I maintain). This allows you to set it to a datetime, but the value that gets stored is a AccessKeyLastUsed object generated from that datetime object. In theory, this could cause an issue for any user who accesses the value directly after setting it initially? I don't know if that means you'll want to hold it back for a major bump or not. There's also the potential issue that setting service and region to "unknown" isn't actually something that AWS would ever return? But speaking personally I would find this convenient.

### DIFF
--- a/moto/core/__init__.py
+++ b/moto/core/__init__.py
@@ -1,20 +1,10 @@
-from .base_backend import SERVICE_BACKEND, BackendDict, BaseBackend
-from .common_models import BaseModel, CloudFormationModel, CloudWatchMetricProvider
-from .models import DEFAULT_ACCOUNT_ID, patch_client, patch_resource
+from .models import DEFAULT_ACCOUNT_ID  # noqa
+from .base_backend import BaseBackend, BackendDict, SERVICE_BACKEND  # noqa
+from .common_models import BaseModel  # noqa
+from .common_models import CloudFormationModel, CloudWatchMetricProvider  # noqa
+from .models import patch_client, patch_resource  # noqa
 from .responses import ActionAuthenticatorMixin
 
 set_initial_no_auth_action_count = (
     ActionAuthenticatorMixin.set_initial_no_auth_action_count
 )
-
-__all__ = [
-    "DEFAULT_ACCOUNT_ID",
-    "SERVICE_BACKEND",
-    "BackendDict",
-    "BaseBackend",
-    "BaseModel",
-    "CloudFormationModel",
-    "CloudWatchMetricProvider",
-    "patch_client",
-    "patch_resource",
-]
--- a/moto/core/__init__.py
+++ b/moto/core/__init__.py
@@ -1,10 +1,20 @@
-from .models import DEFAULT_ACCOUNT_ID  # noqa
-from .base_backend import BaseBackend, BackendDict, SERVICE_BACKEND  # noqa
-from .common_models import BaseModel  # noqa
-from .common_models import CloudFormationModel, CloudWatchMetricProvider  # noqa
-from .models import patch_client, patch_resource  # noqa
+from .base_backend import SERVICE_BACKEND, BackendDict, BaseBackend
+from .common_models import BaseModel, CloudFormationModel, CloudWatchMetricProvider
+from .models import DEFAULT_ACCOUNT_ID, patch_client, patch_resource
 from .responses import ActionAuthenticatorMixin
 
 set_initial_no_auth_action_count = (
     ActionAuthenticatorMixin.set_initial_no_auth_action_count
 )
+
+__all__ = [
+    "DEFAULT_ACCOUNT_ID",
+    "SERVICE_BACKEND",
+    "BackendDict",
+    "BaseBackend",
+    "BaseModel",
+    "CloudFormationModel",
+    "CloudWatchMetricProvider",
+    "patch_client",
+    "patch_resource",
+]
--- a/moto/core/botocore_stubber.py
+++ b/moto/core/botocore_stubber.py
@@ -33,10 +33,7 @@ class BotocoreStubber:
         self.methods.clear()
 
     def register_response(
-        self,
-        method: str,
-        pattern: Pattern[str],
-        response: Callable[..., TYPE_RESPONSE],
+        self, method: str, pattern: Pattern[str], response: Callable[..., TYPE_RESPONSE]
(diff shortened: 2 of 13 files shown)


moto/core/__init__.py
```diff
@@ -1,10 +1,20 @@
...
+from .base_backend import SERVICE_BACKEND, BackendDict, BaseBackend
+from .common_models import BaseModel, CloudFormationModel, CloudWatchMetricProvider
+from .models import DEFAULT_ACCOUNT_ID, patch_client, patch_resource
 from .responses import ActionAuthenticatorMixin
 
 set_initial_no_auth_action_count = (
     ActionAuthenticatorMixin.set_initial_no_auth_action_count
 )
+
+__all__ = [
```
tungol, line 10: doing this clears about the need for the noqa pragmas. Also, mypy complains if you import these elsewhere if the `--no-implicit-reexport` (or `--strict`) flag is set, which was why I did it. I don't think you're using `__all__` declarations anywhere in moto before this, so let me know if it's something you'd like to avoid.

moto/iam/models.py
```diff
@@ -1093,19 +1097,29 @@ def __init__(
...
+    def last_used(self) -> Optional[AccessKeyLastUsed]:
+        return self._last_used
+
+    @last_used.setter
+    def last_used(self, value: Union[datetime, AccessKeyLastUsed, None]) -> None:
         # Some users will set this field manually
         # And they will be setting this value to a `datetime`
         # https://github.com/getmoto/moto/issues/5927#issuecomment-1738188283
-        #
-        # The `to_csv` method calls `last_used.strptime`, which currently works on both AccessKeyLastUsed and datetime
```
tungol, line 1100: this comment isn't technically made obsolete by my change here, but the importance of it is lessened (imo) and I wasn't sure where it would go. I can put it (or something like it) back in, I was waffling on that.

moto/core/__init__.py
bblommers: I think I prefer the existing code. I don't see a usecase where `__all__` is actually useful for Moto-users, and I prefer a few `noqa`'s over having to maintain a duplicate list.

moto/core/botocore_stubber.py
```diff
@@ -33,12 +33,17 @@ def reset(self) -> None:
         self.methods.clear()
 
     def register_response(
-        self, method: str, pattern: Pattern[str], response: Callable[..., TYPE_RESPONSE]
```
bblommers, line 36: There doesn't seem to be any functional changes here - can we keep the existing formatting? Just to try and keep the diff as small as possible

moto/iam/models.py
```diff
@@ -1093,19 +1097,29 @@ def __init__(
...
-        # In the next major release we should communicate that this only accepts AccessKeyLastUsed
-        # (And rework to_csv accordingly)
-        self.last_used: Optional[AccessKeyLastUsed] = None
```
bblommers, line 1103: I'd like to keep this section as is, as well. This improvement does make sense, but it is indeed a breaking change, and I know there are other people who set this field directly - so I only want to change the behaviour in a major version.

My current plan to improve this: let users create an `AccessKeyLastUsed`-object and set that directly, which gives them explicit control over the date-field _and_ the account/region.
It's a few more lines of code, but it is more flexible. Would that work for you?

moto/moto_proxy/proxy3.py
```diff
@@ -62,15 +62,15 @@ def parse_request(
         host: str,
         path: str,
         headers: Any,
-        body: bytes,
+        body: str,
```
bblommers, line 65: This is not correct - the body of a request could be either bytes (in case of an image upload, for instance) or a string. 

What was the reason behind this change?

moto/core/common_models.py
```diff
@@ -94,8 +94,8 @@ def is_created(self) -> bool:
         return True
 
 
-class ConfigQueryModel:
-    def __init__(self, backends: Any):
+class ConfigQueryModel(Generic[SERVICE_BACKEND]):
+    def __init__(self, backends: BackendDict[SERVICE_BACKEND]):
```
bblommers, line 98: Nice - this is going to make development of future ConfigQueryModels much easier

bblommers (changes requested):

moto/core/botocore_stubber.py
tungol: I think I made a change here and then undid it, and the formatting was altered by accidental. Fixed.

moto/core/__init__.py
tungol: fixed.

moto/iam/models.py
tungol: Makes sense to me. Reverted.

moto/moto_proxy/proxy3.py
tungol: Looks like it was a very incorrect type from botocore-stubs on the arguments to AWSPreparedRequest. Fixed.

Looking at the implementation, I think the true type of the body argument to AWSPreparedRequest is something like str | bytes | bytearray | IOBase | None. They just have `str` right now. I'll take it up with them.

moto/iam/models.py
tungol: For what it's worth I'm already doing that in the relevant code: import both iam_backends and AccessKeyLastUsed, create the AccessKeyLastUsed object and set it. 

I thought it might be clever to allow users to skip that and set a datetime object but get a proper AccessKeyLastUsed anyway, but I can also understand why that might be a little too clever and just be confusing.

bblommers (approved):
Awesome - thank you @tungol!

PR #7093, opened 2023-12-04
ResourceGroupsTaggingAPI: add sqs tags

archinksagar:
Adding SQS Tagging functionality to resourcetaggingapi


moto/resourcegroupstaggingapi/models.py
```diff
@@ -18,6 +18,7 @@
 from moto.rds.models import RDSBackend, rds_backends
 from moto.redshift.models import RedshiftBackend, redshift_backends
 from moto.s3.models import S3Backend, s3_backends
+from moto.sqs.models import sqs_backends, SQSBackend
```
bblommers, line 21:
```suggestion
from moto.sqs.models import SQSBackend, sqs_backends
```
I think this is the reason the linter fails: https://github.com/getmoto/moto/actions/runs/7092610448/job/19305827662?pr=7093

bblommers: I'll actually just merge it - let's see if that's the only thing that's blocking the tests

bblommers (approved):
LGTM - thank you for adding this @archinksagar!

PR #7113, opened 2023-12-11
Fix Bad Request - email '"…"' is invalid when uploading to Azure Artifacts

olivierlefloch:
Azure's python repositories perform metadata validation when uploading wheels. When attempting to upload a `Moto` wheel with `twine`, I get the following error message, which this change fixes:

```
Uploading moto-4.2.10-py2.py3-none-any.whl
INFO     Response from https://pkgs.dev.azure.com/…/_packaging/…/pypi/upload: 400 Bad Request - The email '"…"' is invalid. (DevOps Activity ID: …)
INFO     {"$id":"1","innerException":null,"message":"The email '\"…\"' is invalid.","typeName":"Microsoft.VisualStudio.Services.Packaging.Shared.WebApi.Exceptions.InvalidPackageException, Microsoft.VisualStudio.Services.Packaging.Shared.WebApi","typeKey":"InvalidPackageException","errorCode":0,"eventId":3000}
ERROR    HTTPError: 400 Bad Request from https://pkgs.dev.azure.com/…/_packaging/…/pypi/upload Bad Request - The email '"…"' is invalid. (DevOps Activity ID: …)
```

The format without quotes seems supported by `setuptools`' documentation:

https://setuptools.pypa.io/en/latest/userguide/declarative_config.html

`Django` and `manim`'s `setup.cfg` entry for `author_email` also support this format:

- https://github.com/django/django/blob/b287af5dc954628d4b336aefc5027b2edceee64b/setup.cfg#L6
- https://github.com/3b1b/manim/blob/d8428585f84681055fed8aa3fabfb6ae95e4a0ff/setup.cfg#L5


bblommers (approved):
Works for me - thanks for the PR @olivierlefloch!

PR #7121, opened 2023-12-13
CloudFormation: Implement Fn::GetAtt for AWS::EC2::LaunchTemplate

skinitimski:
## Purpose

* Add support for getting attributes of `AWS::EC2::LaunchTemplate` resources.
* Fix bug. Sorta. The real bug is that moto's cloudformation parsing does not honor dependencies between resources -- instead it retries resource updates that raise errors (up to 5 times). This is fine if the update raises errors, but if no error is raised, the updates may process out of order. This was discovered while writing tests for the previous bullet. In our case, we have an AutoScalingGroup that depends on a LaunchTemplate. If the updates to the AutoScalingGroup are handled first, it picks up the current/old version of the LaunchTemplate instead of the new one (because the new one is not created yet). Honoring dependency order appears hard (probably why it is not yet implemented). As a fallback, this PR makes the order of updates deterministic, so that at least if you put the resources in the right order in your template, they will be updated in that order. This is done by using lists rather than sets. The result should be the same -- since previously we were taking the `set` of a `dict`, the collection by definition has unique elements (that is, I cannot find any discernable difference between `set(some_dict)` and `list(some_dict)` except that the latter preserves order).

### DIFF
--- a/moto/cloudformation/parsing.py
+++ b/moto/cloudformation/parsing.py
@@ -740,11 +740,9 @@ class ResourceMap(collections_abc.Mapping):  # type: ignore[type-arg]
         new = other_template["Resources"]
 
         resource_names_by_action = {
-            "Add": set(new) - set(old),
-            "Modify": set(
-                name for name in new if name in old and new[name] != old[name]
-            ),
-            "Remove": set(old) - set(new),
+            "Add": [name for name in new if name not in old],
+            "Modify": [name for name in new if name in old and new[name] != old[name]],
+            "Remove": [name for name in old if name not in new],
         }
 
         return resource_names_by_action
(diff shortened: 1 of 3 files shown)


skinitimski:
> FAILED tests/test_ec2/test_launch_templates_cloudformation.py::test_asg_with_default_launch_template_version - botocore.exceptions.ClientError: An error occurred (InvalidLaunchTemplateName.AlreadyExistsException) when calling the CreateStack operation: Launch template name already in use.

Heh, that sounds like the bug I am working on in another branch! Stay tuned.

bblommers:
This looks good to me so far @skinitimski. I cannot see the test results for some reason, so I'll just close and re-open this PR to force-start them.

Did you mean to open the other PR against our repo?

bblommers (approved):
I'll merge this as is, because it already looks ready to me.

Thank you for adding this to Moto @skinitimski!

skinitimski:
> I cannot see the test results for some reason, so I'll just close and re-open this PR to force-start them.
> 
> Did you mean to open the other PR against our repo?

😅 GitHub is confusing. I had trouble with both things. I have opened the right PR: https://github.com/getmoto/moto/pull/7135

PR #7128, opened 2023-12-15
DynamoDB: UpdateExpressions can contain a new-line

### LINKED ISSUE #7127: DynamoDB UpdateExpression with newlines raises InvalidTokenException
### Context

When calling an API that uses `UpdateExpression`, the `ExpressionTokenizer` raises an `InvalidTokenException` if the expression contains newlines, even though it should be an acceptable expression. This seems similar to the issue #2275, but from what I see the tokenizer's changed quite a lot since then.

It looks like moto raises the following exception when executing this [line](https://github.com/getmoto/moto/blob/4.2.11/moto/dynamodb/parsing/tokens.py#L210).

```
moto.dynamodb.exceptions.InvalidTokenException: 400 Bad Request: {"__type": "com.amazonaws.dynamodb.v20120810#ValidationException", "message": "Invalid UpdateExpression: Syntax error; token: \"\n\", near: \"\n\""}
```

I think the solution could be as simple as adding a newline token and grouping it in with the whitespaces (it worked for me), though I haven't ran tests against it.

### Example

The boto client is able to write the item without any issues, whereas the moto client fails with the following stacktrace.

```
  File "/Users/abbasyadollahi/random/issue.py", line 48, in <module>
    query(moto_client) # This raises
    ^^^^^^^^^^^^^^^^^^
  File "/Users/abbasyadollahi/random/issue.py", line 10, in query
    client.update_item(
  File "/Users/abbasyadollahi/random/env/lib/python3.11/site-packages/botocore/client.py", line 553, in _api_call
  File "/Users/abbasyadollahi/random/env/lib/python3.11/site-packages/botocore/client.py", line 1009, in _make_api_call
botocore.exceptions.ClientError: An error occurred (ValidationException) when calling the UpdateItem operation: Invalid UpdateExpression: Syntax error; token: "
", near: "ADD
"
```

```python
import boto3
import moto

table_name = "test"
my_key = "my_key"
my_count = "1"


def query(client):
    client.update_item(
        TableName=table_name,
        Key={
            "MyKey": {"S": my_key},
        },
        UpdateExpression=(
            """
            ADD
                MyTotalCount :MyCount
            """
        ),
        ExpressionAttributeValues={
            ":MyTotalCount": {"N": my_count},
        },
    )


boto_client = boto3.client("dynamodb")
query(boto_client)  # This passes

with moto.mock_dynamodb():
    moto_client = boto3.client("dynamodb")
    moto_client.create_table(
        TableName=table_name,
        KeySchema=[
            {
                "AttributeName": "MyKey",
                "KeyType": "HASH",
            }
        ],
        AttributeDefinitions=[
            {
                "AttributeName": "MyKey",
                "AttributeType": "S",
            }
        ],
    )
    query(moto_client)  # This raises
```

### Versions

boto3==1.34.0
botocore==1.34.0
moto==4.2.11

bblommers:
Fixes #7127 

Behaviour is now verified against AWS

### DIFF
--- a/moto/dynamodb/parsing/tokens.py
+++ b/moto/dynamodb/parsing/tokens.py
@@ -11,12 +11,12 @@ class Token:
     _TOKEN_INSTANCE = None
     MINUS_SIGN = "-"
     PLUS_SIGN = "+"
-    SPACE_SIGN = " "
     EQUAL_SIGN = "="
     OPEN_ROUND_BRACKET = "("
     CLOSE_ROUND_BRACKET = ")"
     COMMA = ","
     SPACE = " "
+    NEW_LINE = "\n"
     DOT = "."
     OPEN_SQUARE_BRACKET = "["
     CLOSE_SQUARE_BRACKET = "]"
@@ -24,12 +24,12 @@ class Token:
     SPECIAL_CHARACTERS = [
         MINUS_SIGN,
         PLUS_SIGN,
-        SPACE_SIGN,
         EQUAL_SIGN,
         OPEN_ROUND_BRACKET,
         CLOSE_ROUND_BRACKET,
         COMMA,
         SPACE,
+        NEW_LINE,
         DOT,
         OPEN_SQUARE_BRACKET,
         CLOSE_SQUARE_BRACKET,
@@ -193,7 +193,7 @@ class ExpressionTokenizer(object):
             else:
                 self.process_staged_characters()
 
-                if character == Token.SPACE:
+                if character in [Token.SPACE, Token.NEW_LINE]:
                     if (
                         len(self.token_list) > 0
                         and self.token_list[-1].type == Token.WHITESPACE
--- a/tests/test_dynamodb/test_dynamodb_update_expressions.py
+++ b/tests/test_dynamodb/test_dynamodb_update_expressions.py
@@ -1,33 +1,42 @@
 import boto3
+import pytest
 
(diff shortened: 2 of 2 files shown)


moto/dynamodb/parsing/tokens.py
```diff
@@ -17,6 +17,7 @@ class Token:
     CLOSE_ROUND_BRACKET = ")"
     COMMA = ","
     SPACE = " "
+    NEW_LINE = "\n"
     DOT = "."
```
abbasyadollahi, line 21: Unrelated with the issue, but I noticed `EQUAL_SIGN` on line [15](https://github.com/getmoto/moto/pull/7128/files#diff-6960f4c8844f5f277906ab6c10516d64801261a18046cd4a873255d2d25b8ecdR15) and [29](https://github.com/getmoto/moto/pull/7128/files#diff-6960f4c8844f5f277906ab6c10516d64801261a18046cd4a873255d2d25b8ecdR29) can be deleted since they aren't used anywhere throughout the moto repository (it's a duplicate of `SPACE`).

abbasyadollahi (approved):

moto/dynamodb/parsing/tokens.py
bblommers: Good catch @abbasyadollahi - I've removed the field. Thanks for the review!

PR #7129, opened 2023-12-16
CognitoIDP: admin_initiate_auth() now correctly returns a Challenge when 2FA is enabled

tsw025:
Title: 

Summary:

Issue Encountered: In Moto version 4.2.11 with Botocore version 1.32.7 on Python 3.11, an issue was identified where the admin_initiate_auth method incorrectly returned a valid token AuthenticationResult, even when Multi-Factor Authentication (MFA) was enabled for a user. This behavior was inconsistent with the expected results, as a challenge-response (specifically for Time-based One-Time Password (TOTP)) was anticipated.

Issue link: https://github.com/getmoto/moto/issues/7125

### DIFF
--- a/moto/cognitoidp/models.py
+++ b/moto/cognitoidp/models.py
@@ -1449,6 +1449,19 @@ class CognitoIdpBackend(BaseBackend):
                     "Session": session,
                 }
 
+            if (
+                user.software_token_mfa_enabled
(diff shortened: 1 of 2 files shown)


bblommers (approved):
LGTM - thank you so much for fixing this @tsw025!

PR #7134, opened 2023-12-18
EC2: Fix invalid value for NetworkInfo.EnaSupport

viren-nadkarni:
This PR fixes an incorrect dict key and adds valid fallback value for the response field [`EnaSupport`](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_NetworkInfo.html) for the `DescribeInstanceTypes` API operation.

See https://github.com/localstack/localstack/issues/9854


bblommers (approved):
LGTM - thanks for the fix @viren-nadkarni!

PR #7135, opened 2023-12-18
EC2: launch templates created by CloudFormation have a generated name if not provided

skinitimski:
## Purpose

Fixes a bug in the creation of launch templates via CloudFormation. If your template has more than one ***unnamed*** launch template, creation will fail with

> botocore.exceptions.ClientError: An error occurred (InvalidLaunchTemplateName.AlreadyExistsException) when calling the CreateStack operation: Launch template name already in use.

This is because moto's implementation expects that the resource have a `LaunchTemplateName` property even though [it is optional](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-ec2-launchtemplate.html#cfn-ec2-launchtemplate-launchtemplatename).

### DIFF
--- a/moto/ec2/utils.py
+++ b/moto/ec2/utils.py
@@ -233,6 +233,10 @@ def random_launch_template_id() -> str:
     return random_id(prefix=EC2_RESOURCE_TO_PREFIX["launch-template"], size=17)
 
 
+def random_launch_template_name() -> str:
+    return f"LaunchTemplate_{random_resource_id(size=12)}"
+
+
 def random_iam_instance_profile_association_id() -> str:
     return random_id(prefix=EC2_RESOURCE_TO_PREFIX["iam-instance-profile-association"])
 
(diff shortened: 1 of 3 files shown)


moto/ec2/utils.py
```diff
@@ -233,6 +233,10 @@ def random_launch_template_id() -> str:
     return random_id(prefix=EC2_RESOURCE_TO_PREFIX["launch-template"], size=17)
 
 
+def random_launch_template_name() -> str:
+    return f"LaunchTemplate_{random_resource_id(size=12)}"
```
skinitimski, line 237: This is not _quite_ right, it is closer to what is produced by `cloudformation.utils.random_suffix` -- but importing `random_suffix` is circular.

This is probably close enough.

bblommers (approved):
LGTM - thanks for the fix @skinitimski!

PR #7136, opened 2023-12-18
Add cognitoidp.admin_respond_to_auth_challenge

matusfaro:
Implements `cognitoidp`.`admin_respond_to_auth_challenge`. It simply calls the corresponding public endpoint `respond_to_auth_challenge` internally as there is no relevant differences between them.

Since there are no changes in logic, I didn't add any additional tests, but I did convert existing tests so a call to `admin_initiate_auth` and `initiate_auth` uses the appropriate API `admin_respond_to_auth_challenge` and `respond_to_auth_challenge` respectively.

Let me know if there's anything I missed.

### DIFF
--- a/IMPLEMENTATION_COVERAGE.md
+++ b/IMPLEMENTATION_COVERAGE.md
@@ -1180,7 +1180,7 @@
 
 ## cognito-idp
 <details>
-<summary>59% implemented</summary>
+<summary>60% implemented</summary>
 
 - [X] add_custom_attributes
 - [X] admin_add_user_to_group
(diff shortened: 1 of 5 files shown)


bblommers (approved):
The `region_agnostic_backend` loops over all regions, over all UserPools in that region, untils it finds the region that contains the UserPool that contains the correct Session. 

In the `admin_respond_to_auth_challenge` operation we do have access to the `UserPoolId`, so there is a future improvement to only loop over the regions until we have a matching UserPool (and error otherwise). 

But as I said, that as an enhancement - this looks good as is. 

Thank you @matusfaro!

PR #7141, opened 2023-12-19
Techdebt: Fix Route53Resolver tests

bblommers:
These started failing as of botocore-1.34.3

### DIFF
--- a/tests/test_route53resolver/test_route53resolver_endpoint.py
+++ b/tests/test_route53resolver/test_route53resolver_endpoint.py
@@ -3,6 +3,7 @@ from datetime import datetime, timezone
 
 import boto3
 import pytest
+from botocore.config import Config
 from botocore.exceptions import ClientError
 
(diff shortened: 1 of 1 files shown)


tests/test_route53resolver/test_route53resolver_endpoint.py
```diff
@@ -143,6 +146,7 @@ def test_route53resolver_invalid_create_endpoint_args():
     ) in err["Message"]
 
     subnet_too_long = [{"SubnetId": "a" * 33, "Ip": "1.2.3.4"}]
+    print(subnet_too_long)
```
kbalk, line 149: Should the print() be here?  

Thanks for adding the necessary changes to support the newer botocore!

bblommers: Good catch @kbalk! I'll push a fix.

Issue #7142, opened 2023-12-19
Slow s3 exception raising due to usage pattern of jinja2


jogo:
```python
import time

from moto.s3.exceptions import MissingBucket

start = time.time()
for _ in range(1000):
    try:
        raise MissingBucket(bucket="foo")
    except MissingBucket:
        pass

print(round(time.time() - start, 2))
```

with  moto 4.2.12: `1.58`

cProfile data shows it's related to jinja2 usage. Some refactoring should be able to significantly speed this up.
```pstats% stats 30
Tue Dec 19 00:09:04 2023    pstats

         4532388 function calls (4276064 primitive calls) in 2.932 seconds

   Ordered by: cumulative time
   List reduced from 4319 to 30 due to restriction <30>

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
   2658/1    0.010    0.000    2.975    2.975 {built-in method builtins.exec}
        1    0.003    0.003    2.975    2.975 foo.py:1(<module>)
     1000    0.002    0.000    2.501    0.003 /usr/local/lib/python3.8/dist-packages/moto/s3/exceptions.py:103(__init__)
     1000    0.002    0.000    2.499    0.002 /usr/local/lib/python3.8/dist-packages/moto/s3/exceptions.py:76(__init__)
     1000    0.002    0.000    2.496    0.002 /usr/local/lib/python3.8/dist-packages/moto/s3/exceptions.py:44(__init__)
     1000    0.009    0.000    2.494    0.002 /usr/local/lib/python3.8/dist-packages/moto/core/exceptions.py:58(__init__)
     2000    0.002    0.000    2.418    0.001 /usr/local/lib/python3.8/dist-packages/jinja2/environment.py:964(get_template)
     2000    0.007    0.000    2.415    0.001 /usr/local/lib/python3.8/dist-packages/jinja2/environment.py:939(_load_template)
     2000    0.007    0.000    2.393    0.001 /usr/local/lib/python3.8/dist-packages/jinja2/loaders.py:106(load)
     2000    0.008    0.000    2.359    0.001 /usr/local/lib/python3.8/dist-packages/jinja2/environment.py:717(compile)
     1000    0.003    0.000    1.433    0.001 /usr/local/lib/python3.8/dist-packages/jinja2/environment.py:1256(render)
26226/16158    0.014    0.000    1.411    0.000 {method 'join' of 'str' objects}
```

[...]

PR #7143, opened 2023-12-19
Speed up exception handling by re-using Environments

jogo:
Address https://github.com/getmoto/moto/issues/7142

By taking advantage of the caching logic in [Environment](https://jinja.palletsprojects.com/en/3.1.x/api/#jinja2.Environment)

### DIFF
--- a/moto/cloudfront/exceptions.py
+++ b/moto/cloudfront/exceptions.py
@@ -1,7 +1,5 @@
 from typing import Any
 
-from jinja2 import DictLoader, Environment
-
 from moto.core.exceptions import RESTError
 
 EXCEPTION_RESPONSE = """<?xml version="1.0"?>
@@ -18,9 +16,7 @@ EXCEPTION_RESPONSE = """<?xml version="1.0"?>
 class CloudFrontException(RESTError):
     code = 400
     extended_templates = {"cferror": EXCEPTION_RESPONSE}
-    env = Environment(
-        loader=DictLoader(RESTError.extended_templates(extended_templates))
-    )
+    env = RESTError.extended_environment(extended_templates)
 
     def __init__(self, error_type: str, message: str, **kwargs: Any):
         kwargs.setdefault("template", "cferror")
--- a/moto/cloudfront/exceptions.py
+++ b/moto/cloudfront/exceptions.py
@@ -1,6 +1,7 @@
 from typing import Any
 
 from jinja2 import DictLoader, Environment
+
 from moto.core.exceptions import RESTError
 
 EXCEPTION_RESPONSE = """<?xml version="1.0"?>
@@ -17,7 +18,9 @@ EXCEPTION_RESPONSE = """<?xml version="1.0"?>
 class CloudFrontException(RESTError):
     code = 400
     extended_templates = {"cferror": EXCEPTION_RESPONSE}
-    env = Environment(loader=DictLoader(RESTError.templates | extended_templates))
+    env = Environment(
+        loader=DictLoader(RESTError.extended_templates(extended_templates))
+    )
 
     def __init__(self, error_type: str, message: str, **kwargs: Any):
         kwargs.setdefault("template", "cferror")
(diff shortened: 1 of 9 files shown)


jogo:
test code from #7142 before:

Before: `1.58`
After: `0.02`


cProfile data confirms jinja2 call times drop significantly

moto/cloudfront/exceptions.py
```diff
@@ -15,10 +16,11 @@
 
 class CloudFrontException(RESTError):
     code = 400
+    extended_templates = {"cferror": EXCEPTION_RESPONSE}
+    env = Environment(loader=DictLoader(RESTError.templates | extended_templates))
```
bblommers, line 20: The `|`-operator is not supported in Python 3.7, as far as I know. I'll kick off the linter anyway, but I expect that to fail.

moto/elasticache/exceptions.py
```diff
@@ -16,6 +17,8 @@
 class ElastiCacheException(RESTError):
 
     code = 400
+    extended_templates = {"ecerror": EXCEPTION_RESPONSE}
```
bblommers, line 20: The `self.templates["ecerror"] = ERROR...` can now be removed in the `__init__`-method, right?

I saw the same in the `sdb`, `elasticbeanstalk` and `s3` modules.

moto/core/exceptions.py
```diff
@@ -63,8 +64,7 @@ def __init__(
         self.message = message
 
         if template in self.templates.keys():
-            env = Environment(loader=DictLoader(self.templates))
-            self.description: str = env.get_template(template).render(
+            self.description: str = self.env.get_template(template).render(
```
bblommers, line 67:
```suggestion
            self.description: str = self.__class__.env.get_template(template).render(
```
Functionally equivalent (in this case), but this would make it more obvious to a developer that we're calling a class-attribute.

bblommers (changes requested):
Hi @jogo, thanks for the PR! Please see my comments

moto/cloudfront/exceptions.py
jogo: good catch,  this is a 3.9 feature it turns out

moto/elasticache/exceptions.py
jogo: done

moto/core/exceptions.py
```diff
@@ -95,6 +95,12 @@ def to_json(self) -> "JsonRESTError":
         err.code = self.code
         return err
 
+    @classmethod
+    def extended_templates(cls, extended_templates: Dict) -> Dict:
```
jogo, line 99: Perhaps I should refactor this to include the DictLoader Environment boilerplate as well

moto/core/exceptions.py
```diff
@@ -62,9 +63,8 @@ def __init__(
         self.error_type = error_type
         self.message = message
 
-        if template in self.templates.keys():
-            env = Environment(loader=DictLoader(self.templates))
-            self.description: str = env.get_template(template).render(
+        if template in self.env.list_templates():
```
jogo, line 66: Since templates isn't updated only `self.env` is (changed when adding `extended_environment` to make the code less repetitive

jogo:
Unclear what the issue is with the two failing CI jobs:

`Error: Failed to CreateArtifact: Received non-retryable error: Failed request: (409) Conflict: an artifact with this name already exists on the workflow run`
https://github.com/getmoto/moto/actions/runs/7268541069/job/19804663130?pr=7143
https://github.com/getmoto/moto/actions/runs/7268541069/job/19804663439?pr=7143

bblommers:
> Unclear what the issue is with the two failing CI jobs:

That must be unrelated - I'll look at that later on.

LGTM - thank you again for  the PR @jogo!

bblommers (approved):

jogo:
Thank you @bblommers for the quick turnaround on this, much appreciated.

PR #7146, opened 2023-12-20
Techdebt: Replace deprecated utcfromtimestamp-methods

bblommers:
Follow up from #6966

### DIFF
--- a/moto/core/utils.py
+++ b/moto/core/utils.py
@@ -182,13 +182,31 @@ def str_to_rfc_1123_datetime(value: str) -> datetime.datetime:
 
 def unix_time(dt: Optional[datetime.datetime] = None) -> float:
     dt = dt or utcnow()
-    epoch = datetime.datetime.utcfromtimestamp(0)
+    epoch = utcfromtimestamp(0)
     delta = dt - epoch
     return (delta.days * 86400) + (delta.seconds + (delta.microseconds / 1e6))
 
 
 def unix_time_millis(dt: Optional[datetime.datetime] = None) -> float:
     return unix_time(dt) * 1000.0
+
+
+def utcfromtimestamp(value: int) -> datetime.datetime:
+    """
+    Return the UTC datetime corresponding to the POSIX timestamp, with tzinfo None. The resulting object is naive.
+    """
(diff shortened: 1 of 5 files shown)


moto/core/utils.py
```diff
@@ -191,6 +191,24 @@ def unix_time_millis(dt: Optional[datetime.datetime] = None) -> float:
...
+    Return the UTC datetime corresponding to the POSIX timestamp, with tzinfo None. The resulting object is naive.
+    """
+    # Python 3.12 starts throwing deprecation warnings for utcfromtimestamp()
+    # The docs recommend to use fromtimestamp(UTC) instead
+    #
+    # fromtimestamp(UTC) creates an aware datetime - but utcfromtimestamp() creates a naive datetime
+    # That's why we have to `replace(tzinfo=None)` to make now(UTC) naive.
+    if PYTHON_311:
+        # Only available in 3.11
+        from datetime import UTC  # type: ignore
```
rafrafek, line 205:
```python
if PYTHON_311:
    # Only available in 3.11
```
For me both lines suggest that what's inside is not available in Python 3.12.

I think `type: ignore` is not required when using `sys.version_info` https://github.com/microsoft/pyright/issues/4182#issuecomment-1311894451 so it may be better to write it directly than using the custom `PYTHON_311` value.

bblommers: I've updated the comment to be more specific.

The `type: ignore` is a bit annoying indeed, but using a constant gives us two benefits:
 - it's better readable than `sys.version >= ..`
 - it also makes it easier to keep track of version-specific code.

rafrafek: Thanks for updating the comments!

For me `if PYTHON_311` means `== 3.11`, not `>= 3.11`.

For Python programmers `sys.version_info >= (3, 11)` should look pretty readable. It is also easy to track `sys.version_info` usage.

`type: ignore` makes your static type checker less aware of your code.
If you run static type checker against a matrix of Python versions it will not understand what is actually being imported. It may not affect the rest of the code in this example, but it's better to cooperate with your static type checker rather than silence it.

PR #7149, opened 2023-12-21
Update LaunchTemplate by physical ID rather than by name

skinitimski:
## Purpose

Looking up a launch template resource by its name attribute can be tricky when that name is randomly generated by cloudformation rather than one of the declared resource attributes. Instead, do the lookup using the physical resource id.


bblommers (approved):
Makes sense - thanks @skinitimski!

PR #7156, opened 2023-12-22
LakeFormation: extend permissions catalog functionality 

JohannesKoenigTMH:
Implementation for issue https://github.com/getmoto/moto/issues/7150

### DIFF
--- a/moto/lakeformation/models.py
+++ b/moto/lakeformation/models.py
@@ -70,10 +70,6 @@ class Permission:
         return (self.principal == other.principal) and (self.resource == other.resource)
 
     def merge(self, other: Permission) -> None:
-        if not self.equal_principal_and_resouce(other):
-            raise ValueError(
-                "Cannot merge permissions on different a principal or resource."
-            )
         self.permissions = list(set(self.permissions).union(other.permissions))
         self.permissions_with_grant_options = list(
             set(self.permissions_with_grant_options).union(
@@ -82,10 +78,6 @@ class Permission:
         )
 
     def diff(self, other: Permission) -> None:
-        if not self.equal_principal_and_resouce(other):
-            raise ValueError(
-                "Cannot diff permissions on different a principal or resource."
-            )
         if self.permissions is not None:
             self.permissions = list(set(self.permissions).difference(other.permissions))
         if self.permissions_with_grant_options is not None:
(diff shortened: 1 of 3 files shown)


moto/lakeformation/models.py
```diff
@@ -27,6 +30,113 @@ def to_dict(self) -> Dict[str, Any]:
...
+        return {
+            "Permissions": self.permissions,
+            "PermissionsWithGrantOption": self.permissions_with_grant_options,
+            "Resource": self.resource,
+            "Principal": self.principal,
+        }
+
+
+class PermissionCatalog:
+    permissions: Set[Permission]
```
bblommers, line 113: I think this should be at the instance-level, inside `__init__` - otherwise a second PermissionCatalog-instance will overwrite the permissions of the first one

moto/lakeformation/models.py
```diff
@@ -27,6 +30,113 @@ def to_dict(self) -> Dict[str, Any]:
...
+            )
+        self.permissions = list(set(self.permissions).union(other.permissions))
+        self.permissions_with_grant_options = list(
+            set(self.permissions_with_grant_options).union(
+                other.permissions_with_grant_options
+            )
+        )
+
+    def diff(self, other: Permission) -> None:
+        if not self.equal_principal_and_resouce(other):
```
bblommers, line 85: Do we need this check? We completely control when this method is invoked, so we know this will never happen

tests/test_lakeformation/test_lakeformation.py
```diff
@@ -448,14 +514,25 @@ def test_revoke_permissions():
...
     resp = client.list_permissions()
-    assert resp["PrincipalResourcePermissions"] == [
-        {
-            "Principal": {"DataLakePrincipalIdentifier": "asdf"},
-            "Resource": {"Database": {"Name": "db"}},
-            "Permissions": ["SELECT", "ALTER"],
-            "PermissionsWithGrantOption": ["SELECT", "DROP"],
-        }
-    ]
+    print(resp["PrincipalResourcePermissions"])
```
bblommers, line 517: We don't need the print-statement

bblommers (changes requested):
Hi @JohannesKoenigTMH, thanks for the PR! This looks like the right approach - O just have a few minor comments

bblommers (approved):
Yes, LGTM. Thank you for adding this to Moto @JohannesKoenigTMH!

PR #7165, opened 2023-12-28
Sagemaker model: support versioned models

guimorg:
The current implementation for SageMaker Model Registry has some issues regarding the usage of both `ModelPackageGroup` and `ModelPackage` when using the API [CreateModelPackage](https://docs.aws.amazon.com/sagemaker/latest/APIReference/API_CreateModelPackage.html).

As the documentation states:

> [ModelPackageGroupName](https://docs.aws.amazon.com/sagemaker/latest/APIReference/API_CreateModelPackage.html#API_CreateModelPackage_RequestSyntax)
**This parameter is required for versioned models, and does not apply to unversioned models.**

> [ModelPackageName](https://docs.aws.amazon.com/sagemaker/latest/APIReference/API_CreateModelPackage.html#API_CreateModelPackage_RequestSyntax)
This parameter is required for unversioned models. It is not applicable to versioned models.

The current behavior accounts that the `ModelPackageName` is **required** for this API, even though this is not the case and this raises some issues for my team when creating model packages for a model package group.

This MR addresses this and updates tests too.

ps: I was non-orthodox for the class ModelPackage and made some pretty unconventional changes on it. If this is too much I can refactor the class a bit better.

### DIFF
--- a/moto/sagemaker/models.py
+++ b/moto/sagemaker/models.py
@@ -3,7 +3,7 @@ import os
 import random
 import string
 from datetime import datetime
-from typing import Any, Dict, Iterable, List, Optional, Union
+from typing import Any, Dict, Iterable, List, Optional, Union, cast
 
 from dateutil.tz import tzutc
 
@@ -1105,6 +1105,12 @@ class ModelPackage(BaseObject):
             "AdditionalInferenceSpecifications",
             "SkipModelValidation",
         ]
+        if (
+            "ModelPackageGroupName" in response_object
+            and response_object["ModelPackageGroupName"]
+            == response_object["ModelPackageName"]
+        ):
+            del response_object["ModelPackageName"]
         return {
             k: v
             for k, v in response_object.items()
@@ -3295,6 +3301,8 @@ class SageMakerModelBackend(BaseBackend):
             )
         if model_package_group_name is not None:
             model_package_type = "Versioned"
+            if model_package_group_name.startswith("arn:aws"):
+                model_package_group_name = model_package_group_name.split("/")[-1]
         model_package_summary_list = list(
             filter(
                 lambda x: (
@@ -3377,7 +3385,7 @@ class SageMakerModelBackend(BaseBackend):
 
     def create_model_package(
         self,
-        model_package_name: str,
+        model_package_name: Optional[str],
         model_package_group_name: Optional[str],
         model_package_description: Optional[str],
         inference_specification: Any,
@@ -3397,15 +3405,25 @@ class SageMakerModelBackend(BaseBackend):
         additional_inference_specifications: Any,
     ) -> str:
         model_package_version = None
-        if model_package_group_name is not None:
+        if model_package_group_name and model_package_name:
+            raise AWSValidationException(
+                "An error occurred (ValidationException) when calling the CreateModelPackage operation: Both ModelPackageName and ModelPackageGroupName are provided in the input. Cannot determine which one to use."
+            )
+        elif not model_package_group_name and not model_package_name:
+            raise AWSValidationException(
+                "An error ocurred (ValidationException) when calling the CreateModelPackag operation: Missing ARN."
+            )
+        elif model_package_group_name:
+            model_package_name = model_package_group_name
             model_packages_for_group = [
                 x
                 for x in self.model_packages.values()
                 if x.model_package_group_name == model_package_group_name
             ]
             model_package_version = len(model_packages_for_group) + 1
+
         model_package = ModelPackage(
-            model_package_name=model_package_name,
+            model_package_name=cast(str, model_package_name),
             model_package_group_name=model_package_group_name,
             model_package_description=model_package_description,
             inference_specification=inference_specification,
--- a/moto/sagemaker/models.py
+++ b/moto/sagemaker/models.py
@@ -1004,6 +1004,7 @@ class ModelPackage(BaseObject):
         client_token: str,
         region_name: str,
         account_id: str,
+        model_package_type: str,
         tags: Optional[List[Dict[str, str]]] = None,
     ) -> None:
         fake_user_profile_name = "fake-user-profile-name"
@@ -1032,6 +1033,7 @@ class ModelPackage(BaseObject):
         self.inference_specification = inference_specification
         self.source_algorithm_specification = source_algorithm_specification
         self.validation_specification = validation_specification
+        self.model_package_type = model_package_type
         self.model_package_status_details = {
             "ValidationStatuses": [
                 {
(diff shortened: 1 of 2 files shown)


moto/sagemaker/models.py
```diff
@@ -1105,6 +1105,8 @@ def gen_response_object(self) -> Dict[str, Any]:
             "AdditionalInferenceSpecifications",
             "SkipModelValidation",
         ]
+        if "ModelPackageGroupName" in response_object and response_object["ModelPackageGroupName"] == response_object["ModelPackageName"]:
+            del response_object["ModelPackageName"]
```
guimorg, line 1109: This is what I've referred too as being too unconventional. For versioned models - i.e. Model Package Groups - the ModelPackageName is omitted from the output.

moto/sagemaker/models.py
```diff
@@ -1105,6 +1105,12 @@ def gen_response_object(self) -> Dict[str, Any]:
             "AdditionalInferenceSpecifications",
             "SkipModelValidation",
         ]
+        if (
+            "ModelPackageGroupName" in response_object
+            and response_object["ModelPackageGroupName"]
+            == response_object["ModelPackageName"]
+        ):
+            del response_object["ModelPackageName"]
```
guimorg, line 1113: I am wondering if it makes more sense to use the `model_package_type` for this, but we currently do not hold it. I am dropping a new commit with this change that we can later drop if that is the case.

guimorg: I've just dropped this on commit 6915d300cad6c01ed4d9a7d09ee047f208b42924.

moto/sagemaker/models.py
```diff
@@ -1105,6 +1107,10 @@ def gen_response_object(self) -> Dict[str, Any]:
             "AdditionalInferenceSpecifications",
             "SkipModelValidation",
         ]
+        if self.model_package_type == "Versioned":
```
bblommers, line 1110: In the first iteration it wasn't obvious what was going on, but now the logic is clear - so this is a much better solution IMO.

moto/sagemaker/models.py
```diff
@@ -3396,6 +3404,14 @@ def create_model_package(
         sample_payload_url: Any,
         additional_inference_specifications: Any,
     ) -> str:
+        if not bool(model_package_group_name) ^ bool(model_package_name):
```
bblommers, line 3407: Can we simplify this to `if model_package_group_name and model_package_name`? That would be more readable

moto/sagemaker/models.py
```diff
@@ -3404,6 +3420,8 @@ def create_model_package(
                 if x.model_package_group_name == model_package_group_name
             ]
             model_package_version = len(model_packages_for_group) + 1
+            model_package_name = model_package_group_name
+        assert model_package_name
```
bblommers, line 3424: There shouldn't be any assertions in the logic. If it's a valid scenario where the user hasn't provided the right values we can raise a proper exception, otherwise we should trust the logic above that the value is always set

bblommers (changes requested):
Hi @guimorg, thanks for the PR! Please see my comments

moto/sagemaker/models.py
guimorg: Yeah! Sure, good call!

moto/sagemaker/models.py
guimorg: Hmm, this was just for mypy type narrowing, let me check if I can solve it other way.

moto/sagemaker/models.py
guimorg: Hmm, I just noticed that we **need** the XOR because `model_package_name` is now `Optional`.

I feel that this XOR operation is better than `if model_package_name and model_package_group_name or not model_package_name and not model_package_group_name`. 

What do you think?

moto/sagemaker/models.py
guimorg: I've "solved" this with type casting for passing it to the `ModelPackage` class. This is because the `model_package_name` as the function parameter is `Optional`, and because of that even after we treat the case with the if-statement, it still is considered `str | None`.

So we cast it to an `str` after the if-statement.

What do you think?

moto/sagemaker/models.py
guimorg: Hmm, I think it is better to treat this as another if-statement as the message is different if none of them are passed. I will fix it asap.

moto/sagemaker/models.py
```diff
@@ -3397,15 +3405,32 @@ def create_model_package(
...
     ) -> str:
         model_package_version = None
-        if model_package_group_name is not None:
+        if model_package_group_name and model_package_name:
+            raise AWSValidationException(
+                "An error occurred (ValidationException) when calling the CreateModelPackage operation: Both ModelPackageName and ModelPackageGroupName are provided in the input. Cannot determine which one to use."
+            )
+        elif not model_package_group_name and not model_package_name:
+            raise AWSValidationException(
+                "An error ocurred (ValidationException) when calling the CreateModelPackag operation: Missing ARN."
```
guimorg, line 3414: This  is needed since `ModelPackageName` and `ModelPackageGroupName` are `Optional`. Unfortunately, AWS response in this case is pretty general "Missing ARN". (probably referring to any one of these fields).

moto/sagemaker/models.py
```diff
@@ -3397,15 +3405,32 @@ def create_model_package(
...
+            model_package_name = model_package_group_name
             model_packages_for_group = [
                 x
                 for x in self.model_packages.values()
                 if x.model_package_group_name == model_package_group_name
             ]
+            if model_package_group_name not in self.model_package_groups:
+                raise AWSValidationException(
+                    "An error ocurred (ValidationException) when calling the CreateModelPackage operation: Model Package Group does not exist."
+                )
```
guimorg, line 3427: This is also a thing to check, if the specified group doesn't exist, we can't create a ModelPackage for it.

tests/test_sagemaker/test_sagemaker_model_packages.py
```diff
@@ -133,24 +134,21 @@ def test_list_model_packages_approval_status():
...
     client.create_model_package(
-        ModelPackageName="test-model-package-2",
-        ModelPackageDescription="test-model-package-description-3",
         ModelPackageGroupName=group1,
     )
     client.create_model_package(
         ModelPackageName="test-model-package-without-group",
-        ModelPackageDescription="diff_group",
     )
     resp = client.list_model_packages(ModelPackageGroupName=group1)
```
guimorg, line 153: These changes are referred to the fact that you can't specify both `ModelPackageName` and `ModelPackageGroupName`. Also, we need to first create the `ModelPackageGroup` for creating its packages.

bblommers (approved):
LGTM - thank you for fixing this!

Issue #7164, opened 2023-12-27
Cognito IDP - user not found after being created


nykolaslima:
I have a Flask app that renders HTML templates using jinja2.
I am creating a end to end test with Playwright that will try to make the user login.
The code is using AWS cognito to manage the user's credentials and in the test we are using moto to mock the AWS cognito.
In the test setup we create a user pool and a user inside this pool. After creating the user, executing the `list_users` methods returns the inserted user.

When our test hits the flask api that was started by `pytest flask` and we try to execute the same `list_users`. we got an empty array as response, as the pool has no users.

here is the code:
```
@pytest.yield_fixture(scope="function", autouse=True)
    def cognito_client(self):
        idp = mock_cognitoidp()
        idp.start()

        cognito_client = boto3.client('cognito-idp',
                                      region_name=get_key('AWS_REGION'),
                                      aws_access_key_id=get_key('AWS_ACCESS_KEY_ID'),
                                      aws_secret_access_key=get_key('AWS_SECRET_ACCESS_KEY'),
                                      aws_session_token=get_key('AWS_SESSION_TOKEN'))

        # configure default users pool
        user_pool = cognito_client.create_user_pool(PoolName='users')
        os.environ['COGNITO_USER_POOL_ID'] = user_pool['UserPool']['Id']
        user_pool_client = cognito_client.create_user_pool_client(UserPoolId=user_pool['UserPool']['Id'],
                                                                  ClientName='test')
        os.environ['COGNITO_CLIENT_ID'] = user_pool_client['UserPoolClient']['ClientId']

        yield cognito_client

        del os.environ['COGNITO_USER_POOL_ID']
        del os.environ['COGNITO_CLIENT_ID']
        idp.stop()
        
@fixture(scope="function", autouse=True)
    def logged_user(self, cognito_util, logged_user_password, cognito_client):
        user = UserTableFactory.create().to_domain()
        cognito_util.create_confirmed_user(user.email, logged_user_password)
        # here we got a the user that was just inserted to the pool
        cognito_client.list_users(
            UserPoolId=get_key('cognito_user_pool_id'),
            Limit=123,
        )
        return user

def test_should_sign_in(self, logged_user, logged_user_password, page, cognito_client):
        page.goto(url_for('main.load_sign_in', _external=True))
        page.get_by_label("Email").click()
        page.get_by_label("Email").fill(logged_user.email)
        page.get_by_label("Password").click()
        page.get_by_label("Password").fill(logged_user_password)
[...]

PR #7166, opened 2023-12-28
docs: Improve Moto concurrency safety documentation

nykolaslima:
Enhance Moto's documentation to explicitly explain its lack of support for multithreaded access/multiprocessing. Emphasize the Python threading model where a new thread copies the global context from the main thread, highlighting that this is a one-time snapshot at thread creation. Clarify the implications for read-only access and recommend performing Moto operations within the same thread context as the application or test logic to ensure consistent behavior.

related to #7164

### DIFF
--- a/docs/docs/faq.rst
+++ b/docs/docs/faq.rst
@@ -19,6 +19,14 @@ Is Moto concurrency safe?
 
 No. Moto is not designed for multithreaded access/multiprocessing.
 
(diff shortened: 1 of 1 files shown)


bblommers (approved):
Yes, this makes a lot of sense. Thank you for adding this to Moto @nykolaslima!

PR #7180, opened 2024-01-03
Cloudformation: broken changeset update

rafcio19:
Closes https://github.com/getmoto/moto/issues/7155

### DIFF
--- a/tests/test_cloudformation/test_cloudformation_stack_crud_boto3.py
+++ b/tests/test_cloudformation/test_cloudformation_stack_crud_boto3.py
@@ -33,7 +33,7 @@ dummy_template = {
         "EC2Instance1": {
             "Type": "AWS::EC2::Instance",
             "Properties": {
-                "ImageId": 123,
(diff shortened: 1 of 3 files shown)


tests/test_cloudformation/test_cloudformation_stack_crud_boto3.py
```diff
@@ -2309,3 +2319,21 @@ def get_role_name():
                 AssumeRolePolicyDocument="some policy",
                 Path="/my-path/",
             )["Role"]["Arn"]
+
+
+@mock_cloudformation()
+def test_update_changeset_yaml_and_json():
```
bblommers, line 2325: Can we re-use an existing test to verify this is fixed? For example, the `test_describe_change_set` already verifies the  creating/describing/executing of a change set. It should be trivial to parametrize that test instead, and verify the entire flow with either JSON or YAML

bblommers (changes requested):

PR #7184, opened 2024-01-03
Implements sso-admin AWS Managed Policies 

joelmccoy:
Added AWS Managed policy operations for sso-admin permissionsets.

The following API calls are implemented:

- [detach_managed_policy_from_permission_set](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/sso-admin/client/detach_managed_policy_from_permission_set.html)
- [attach_managed_policy_to_permission_set](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/sso-admin/client/attach_managed_policy_to_permission_set.html)
- [list_managed_policies_in_permission_set](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/sso-admin/client/list_managed_policies_in_permission_set.html)

Added additional exception handling.

Also references the managed policies [here](https://github.com/getmoto/moto/blob/master/moto/iam/aws_managed_policies.py) for validation of a proper AWS Managed Policy existing.

### DIFF
--- a/moto/ssoadmin/models.py
+++ b/moto/ssoadmin/models.py
@@ -1,12 +1,21 @@
+import json
 from typing import Any, Dict, List
 
 from moto.core import BackendDict, BaseBackend, BaseModel
 from moto.core.utils import unix_time
+from moto.iam.aws_managed_policies import aws_managed_policies_data
 from moto.moto_api._internal import mock_random as random
 from moto.utilities.paginator import paginate
 
-from .exceptions import ResourceNotFoundException
+from .exceptions import (
+    ConflictException,
+    ResourceNotFoundException,
+    ServiceQuotaExceededException,
+)
 from .utils import PAGINATION_MODEL
+
+# https://docs.aws.amazon.com/singlesignon/latest/userguide/limits.html
+MAX_MANAGED_POLICIES_PER_PERMISSION_SET = 20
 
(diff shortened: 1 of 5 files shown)


moto/ssoadmin/models.py
```diff
@@ -83,13 +96,27 @@ def generate_id(instance_arn: str) -> str:
...
+
+
 class SSOAdminBackend(BaseBackend):
     """Implementation of SSOAdmin APIs."""
 
     def __init__(self, region_name: str, account_id: str):
         super().__init__(region_name, account_id)
         self.account_assignments: List[AccountAssignment] = list()
         self.permission_sets: List[PermissionSet] = list()
+        self.aws_managed_policies: Dict[str, Any] = json.loads(
```
bblommers, line 117: This is quite a big JSON file, so loading is relatively expensive. Can we lazy-load this, so that it's only called when used in `attach_managed_policy_to_permission_set`?

That way, people who do not use this method (and don't need the data) aren't impacted.

joelmccoy: Good call!  Fixed this.  Let me know if this change resolves your concern @bblommers .

bblommers (approved):
LGTM - thanks @joelmccoy!

PR #7194, opened 2024-01-06
Introduce `mock_aws()` decorator

bblommers:
Removes existing decorators

The functionality behind `mock_lambda_simple` and `mock_batch_simple` can now be reproduced by using the `config`-parameter:
```
@mock_aws(config={"batch": {"use_docker": False}, "lambda": {"use_docker": False}})
```

Service-specific backends are now loaded when a request first comes in for that service

The `BackendDict` now keeps track of which backends/services are loaded, making it easy to reset all data in one go


kbalk:
@bblommers - I think the scaffold templates would need to be updated, e.g. `scripts/template/lib/__init__.py.j2`?

I saw that the docstring at the top of some of these files had been deleted with these changes.  `pylint` (if you don't use the same exceptions as `moto`) will not be happy about that.  I wish they could have been kept, but I don't think they were that prevalent and not worth adding back at this point.

bblommers:
Good call about the templates @kbalk! I'll change those as well.

Regarding the docstrings - do you mean the the ones in `__init__`-files?
```
"""cloudtrail module initialization; sets value for base decorator."""
```
I removed them all because it didn't add any useful info - I wasn't aware that `pylint` had an opinion on them.

kbalk:
> Regarding the docstrings - do you mean the the ones in `__init__`-files?
> 
> ```
> """cloudtrail module initialization; sets value for base decorator."""
> ```
> 
> I removed them all because it didn't add any useful info - I wasn't aware that `pylint` had an opinion on them.

In any file but an empty one:
https://pylint.readthedocs.io/en/latest/user_guide/messages/convention/missing-module-docstring.html

PR #7199, opened 2024-01-07
RDS DB snapshot none handling

SoenkeD:
When trying to execute the action `CreateDBSnapshot` a deserialization error occurs. When looking at the response from Moto it became clear what the problem was:
If the value used in the template was none / nil / null it occurs in the response XML as "None" which is not valid, if a boolean value is expected. 
The proposed solution removes the values from the XML if they are none. It works with the Golang SDK.


bblommers (approved):
Makes sense - thanks for the fix @SoenkeD!

PR #7204, opened 2024-01-10
EC2: Add tag support for security group rules

viren-nadkarni:
This PR adds tagging support for security group rules.

<https://docs.aws.amazon.com/vpc/latest/userguide/security-group-rules.html#working-with-security-group-rules>

See https://github.com/localstack/localstack/issues/8642


bblommers (approved):
LGTM - thanks for the PR!

PR #7222, opened 2024-01-17
Fix more test failures on platforms with 32-bit time_t

mgorny:
Fix two more test failures that are caused by platform `time_t` not being able to represent dates beyond 2038.  This is the same approach as used in #4954.


mgorny:
Thank you!

PR #7227, opened 2024-01-18
Add SageMaker Feature Group

bogdangi:


tests/test_sagemaker/test_sagemaker_feature_groups.py
```diff
@@ -0,0 +1,90 @@
...
+        ],
+        RoleArn="arn:aws:iam::123456789012:role/AWSFeatureStoreAccess",
+        OfflineStoreConfig={
+            "DisableGlueTableCreation": False,
+            "S3StorageConfig": {"S3Uri": "s3://mybucket"},
+        },
+    )
+    resp = client.describe_feature_group(FeatureGroupName="some-feature-group-name")
+
+    assert resp == dict(
```
bblommers, line 59: Minor niggle - but I think I prefer individual asserts, instead of a comparison against the entire dict. I.e.: 
```
assert arn == arn
assert name == name
... etc
```

tests/test_sagemaker/test_sagemaker_feature_groups.py
```diff
@@ -0,0 +1,90 @@
...
+        },
+    )
+
+    assert (
+        resp["FeatureGroupArn"]
+        == "arn:aws:sagemaker:us-east-2:123456789012:feature-group/some-feature-group-name"
+    )
+
+
+@freeze_time("2024-01-17 17:27:37")
```
bblommers, line 38: Freezetime doesn't work in ServerMode, so this would fail atm.

Considering the exact value of the creation date doesn't matter, I would prefer to just check if the value exists:
```
assert isinstance(resp["CreationDate"], datetime)
```
That way we don't have to mock the date, and it will work regardless of the exact date/format.

bblommers (changes requested):
Hi @bogdangi, thanks for the PR! The logic looks good - just got some comments on the test.

Note that the linter is currently failing as well:
https://github.com/getmoto/moto/actions/runs/7582115487

tests/test_sagemaker/test_sagemaker_feature_groups.py
bogdangi: Make sense.

tests/test_sagemaker/test_sagemaker_feature_groups.py
bogdangi: Yes, I see it done ;)

bblommers (approved):
LGTM - thanks again @bogdangi!

PR #7230, opened 2024-01-19
Refactoring: Adding typing for event messagre structure

tsugumi-sys:
How about adding typing for event message structure?

memo: `typing.TypedDict` introduced in python3.8, and `typing.Required` introduced in 3.11.

Document: https://docs.aws.amazon.com/eventbridge/latest/userguide/eb-event-patterns.html


bblommers (changes requested):
Hey @tsugumi-sys, this makes sense overall. I'd like to prevent us from having to add the `typing_extensions` dependency at runtime though.

Do you have any objections if this will only become part of Moto v5?

We will drop support for Python 3.7 in that version, so that would allow us to always use `from typing import TypedDict`.

If I'm not mistaken, we can simply use TYPE_CHECKING for the `required` attribute:
```
from typing import TYPE_CHECKING

if TYPE_CHECKING:
    from typing_extensions import Required
```
And then the Required in between quotes, so that it's ignored at runtime:
```
"detail-type": "Required[Union[str, List[str]]]"
```

bblommers:
Changing the base-branch to `v5` becomes a bit messy, so this may need a new branch/PR

tsugumi-sys:
@bblommers 
> Do you have any objections if this will only become part of Moto v5?
No, I don't.  Sounds good!

I'll create a new PR! Thanks:)

tsugumi-sys:
@bblommers 
I've created a new PR!
https://github.com/getmoto/moto/pull/7234

PR #7231, opened 2024-01-19
Feature: Backup api

archinksagar:
Please provide feedback on this initial draft of Backup API. 
I have added the following API endpoints.
1. create_backup_plan
2. create_backup_vault
3. get_backup_plan - in progress
4. describe_backup_vault - in progress
5. delete_backup_plan
6. list_backup_plans
7. list_backup_vaults - in progress
8. list_tags
9. tag_resource
10. untag_resource

Things yet to be implemented
1. describe_backup_vault
2. Deletion_date and LastExecutionDate in get_backup_plan, list_backup_plans
3. NumberOfRecoveryPoints, MinRetentionDays, MaxRetentionDays, LockDate in list_backup_vaults
4. Pagination
5. Exceptions


tests/test_backup/test_server.py
```diff
@@ -0,0 +1,13 @@
+"""Test different server responses."""
```
bblommers, line 1: This file doesn't seem to add anything, so let's just delete it

moto/__init__.py
```diff
@@ -235,6 +235,11 @@ def f() -> "xray_segment_type":
 mock_xray = lazy_load(".xray", "mock_xray")
 mock_xray_client = lazy_load(".xray", "mock_xray_client")
 mock_wafv2 = lazy_load(".wafv2", "mock_wafv2")
+mock_backup = lazy_load(".backup", "mock_backup", boto3_name="backup")
+mock_backup = lazy_load(".backup", "mock_backup", boto3_name="backup")
+mock_backup = lazy_load(".backup", "mock_backup", boto3_name="backup")
+mock_backup = lazy_load(".backup", "mock_backup", boto3_name="backup")
+mock_backup = lazy_load(".backup", "mock_backup", boto3_name="backup")
```
bblommers, line 242: This needs to only be added once (I know it's a bug in the `scaffold`-script, it adds this line without checking)

moto/backup/models.py
```diff
@@ -0,0 +1,204 @@
...
+    def to_list_dict(self) -> Dict[str, Any]:
+        dct = self.to_get_dict()
+        dct.pop("BackupPlan")
+        dct["BackupPlanName"] = self.backup_plan.get("BackupPlanName")
+        return dct
+    
+    def random_str_utf8(self) -> str:
+        ran_str = mock_random.get_random_string(length=48)
+        utf8_bytes = ran_str.encode('utf-8')
+        return utf8_bytes.decode('utf-8')
```
bblommers, line 62: What is the reasoning behind the decode/encode operation? It looks like we can just `return mock_random.get_random_string(length=48)` to get the same outcome. (Or even get rid of the function altogether..)

bblommers (changes requested):
Hi @archinksagar, thanks for the PR! This looks good overall, I just have a few nitpicks.

Note that the linter is currently failing as well: https://github.com/getmoto/moto/actions/runs/7586740652/job/20665598379?pr=7231

bblommers:
> Things yet to be implemented

Out of interest - do you want to add them to this PR as well? Part of a later PR? Or is it just a general note that they are for someone else to implement?

All options are fine by me - I'm just trying to get a feel for how 'ready' this PR is :slightly_smiling_face:

archinksagar:
Thanks for your feedback. Except for pagination and describe_backup_vault , I will implement Exceptions and others I listed. Apart from that, I realized i have to modify "resourcegroupstaggingapi" to use vault tags. I am having trouble figuring out the values for these fields, I don't know when and how they get populated- "Deletion_date" and "LastExecutionDate", "NumberOfRecoveryPoints". Is there any documentation I can refer?

bblommers:
If I read [the docs here](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/backup/client/get_backup_plan.html) correctly:

 - DeletionDate would only be set after `delete_backup_plan` is called
 - LastExecutionDate can be empty (None), and maybe should be - as there is no option yet to `start_restore_job`

The term `RecoveryPoints` is the same as `Backup` - so NumberOfRecoveryPoints is just the number of backups that single vault contains. I guess that would be always `0` for now, until we implement `start_backup_job`

archinksagar:
hi @bblommers I have added code to resourcegrouptaggingapi and made few other modifications. I did run make lint, and fixed linter errors. Not sure why the linter is failing.

bblommers:
Hey @archinksagar, which version of `black` you do have installed? Different versions will produce different results. Moto uses `black==22.3.0`

Running `black moto tests` should fix that (step of the) build.

https://github.com/getmoto/moto/actions/runs/7659295607/job/20874125053?pr=7231

moto/backup/models.py
```diff
@@ -0,0 +1,236 @@
...
+        ran_str = mock_random.get_random_string(length=48)
+        self.version_id = ran_str
+        self.creator_request_id = creator_request_id
+        self.backup_plan = backup_plan
+        adv_settings = backup_plan.get("AdvancedBackupSettings")
+        self.advanced_backup_settings = adv_settings or []
+        self.deletion_date: Optional[float] = None
+        # Deletion Date is updated when the backup_plan is deleted
+        self.last_execution_date = None  # start_restore_job not yet supported
+        rules = backup_plan.get("Rules")
```
bblommers, line 33:
```suggestion
        rules = backup_plan["Rules"]
```

The `Rules` argument is mandatory anyway. This should fix the mypy error (`moto/backup/models.py:34:21: error: Item "None" of "Any | None" has no attribute "__iter__" (not iterable)  [union-attr]`)

archinksagar:
@bblommers Its failing at test/test with "def delete_backup_plan(self, backup_plan_id: str) -> tuple[str, str, float, str]:
E   TypeError: 'type' object is not subscriptable" error. 
How do i test it in my local, is it 'make test-only'? That would take a long time. Is there a command to test only backup?

bblommers:
> @bblommers Its failing at test/test with "def delete_backup_plan(self, backup_plan_id: str) -> tuple[str, str, float, str]: E TypeError: 'type' object is not subscriptable" error. How do i test it in my local, is it 'make test-only'? That would take a long time. Is there a command to test only backup?

@archinksagar You could just run `pytest -sv tests/test_backup` to only get the tests in that directory.
I think the problem is that `tuple` is only available as of Python 3.9, that's why 3.7 and 3.8 fail to compile but everything else does run. So if you only test it in a recent Python compiler, the problem would not show up.

You can use `from typing import Tuple` instead (note the upper-case T) - that will work in every Python version.

bblommers:
@archinksagar Hope you don't mind - I just pushed a fix that should get the ServerMode tests to pass.

The initial version intercepted all URL's (`/.+`), which is fine for our own request matcher. But when running the tests in ServerMode, the request matching is done by the `werkzeug`-module, and they require the URL's to be spelled out.

archinksagar:
> @archinksagar Hope you don't mind - I just pushed a fix that should get the ServerMode tests to pass.
> 
> The initial version intercepted all URL's (`/.+`), which is fine for our own request matcher. But when running the tests in ServerMode, the request matching is done by the `werkzeug`-module, and they require the URL's to be spelled out.

@bblommers Thanks for fixing it.

bblommers (approved):
LGTM - thank you so much for adding this feature to Moto @archinksagar!

PR #7234, opened 2024-01-20
Refactoring: Adding typing for event message structure

tsugumi-sys:
Adding typing for the event message structure.
(supported only in moto v5. Check https://github.com/getmoto/moto/pull/7230)

Document: https://docs.aws.amazon.com/eventbridge/latest/userguide/eb-event-patterns.html

### DIFF
--- a/moto/events/utils.py
+++ b/moto/events/utils.py
@@ -1,3 +1,25 @@
+from typing import TYPE_CHECKING, Any, Dict, List, TypedDict, Union
+
+if TYPE_CHECKING:
+    from typing_extentions import Required
+
+
+# NOTE: Typing is based on the following document https://docs.aws.amazon.com/eventbridge/latest/userguide/eb-event-patterns.html
+EventMessageType = TypedDict(
+    "EventMessageType",
+    {
+        "version": str,
+        "id": str,
+        "detail-type": Required[Union[str, List[str]]],
+        "source": Required[Union[str, List[str]]],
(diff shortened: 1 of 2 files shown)


moto/events/utils.py
```diff
@@ -1,3 +1,25 @@
+from typing import TYPE_CHECKING, Any, Dict, List, TypedDict, Union
```
bblommers, line 1: The `Any` and `Dict` should probably also be only imported when `TYPE_CHECKING` - the linter is currently failing:
https://github.com/getmoto/moto/actions/runs/7594740349/job/20686572806?pr=7234

moto/events/utils.py
```diff
@@ -1,3 +1,25 @@
...
+
+
+# NOTE: Typing is based on the following document https://docs.aws.amazon.com/eventbridge/latest/userguide/eb-event-patterns.html
+EventMessageType = TypedDict(
+    "EventMessageType",
+    {
+        "version": str,
+        "id": str,
+        "detail-type": Required[Union[str, List[str]]],
+        "source": Required[Union[str, List[str]]],
```
bblommers, line 14: These two should also be quoted, that's why the linter is failing

tsugumi-sys: Oh yes, thanks!

bblommers (approved):
LGTM - thanks @tsugumi-sys!

PR #7238, opened 2024-01-22
Using Enum for s3 notification events

tsugumi-sys:
- S3 notification events are updated.
- Using an enum for the event list.
- Refactoring error message.

Documents for the supported event: https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3/client/put_bucket_notification_configuration.html

image


bblommers (approved):

PR #7243, opened 2024-01-23
RDS instance restore to point in time - fix failing to copy DB subnet…

SoenkeD:
When restoring a RDS instance, where the source instance has a DB subnet group, Moto throws a 500 error. This is due to it not being able to deep copy DB subnet group. 
The attribute can be removed as the DB Subnet group (name) is part of the RestoreRDSInstanceToPointInTime endpoint.


bblommers (approved):
Works for me - thanks for the PR @SoenkeD!

PR #7246, opened 2024-01-23
SageMaker: correct describe_model_package_group does not exist msg

guimorg:
## Description

The message that boto3 raises when a `ValidationException` happens (the model package group does not exist) is actually:

```shell
❯ aws sagemaker describe-model-package-group --model-package-group-name hello

An error occurred (ValidationException) when calling the DescribeModelPackageGroup operation: ModelPackageGroup arn:aws:sagemaker:us-east-1:12345678910:model-package-group/hello does not exist.
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -32,6 +32,7 @@ lint:
 
 format:
 	black moto/ tests/
+	ruff --fix moto/ tests/
 
 test-only:
 	rm -f .coverage
--- a/moto/sagemaker/models.py
+++ b/moto/sagemaker/models.py
@@ -3330,8 +3330,14 @@ class SageMakerModelBackend(BaseBackend):
     ) -> ModelPackageGroup:
         model_package_group = self.model_package_groups.get(model_package_group_name)
         if model_package_group is None:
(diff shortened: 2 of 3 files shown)


Makefile
```diff
@@ -32,6 +32,7 @@ lint:
 
 format:
 	black moto/ tests/
+	ruff --fix moto/ tests/
```
guimorg, line 35: When we perform `make lint`, we use `ruff`, but when formatting we use `black`. This creates a small inconsistency in the development workflow, especially because `ruff` mentions this:

```
❯ make lint
Running ruff...
ruff check moto tests
tests/test_sagemaker/test_sagemaker_model_package_groups.py:2:1: I001 [*] Import block is un-sorted or un-formatted
Found 1 error.
[*] 1 fixable with the `--fix` option.
make: *** [Makefile:23: lint] Error 1
```

guimorg:
I feel that some tests are failing but are not related to any changes in this MR. Can somebody give a hand to what to do in this case?

bblommers (approved):
Good catch @guimorg - thanks for fixing this!

PR #7249, opened 2024-01-25
Support Route53 create_hosted_zone raising ClientError ConflictingDomainExists

codeocelot:
Project does not support Route53 create_hosted_zone raising ConflictingDomainExists if one attempts to create a hosted zone with the same name on the same delegation set.  According to the docs, this should raise a ClientError named ConflictingDomainExists.

Simple Reproduction

```
@mock_route53
def test():
    conn = boto3.client('route53')
    conn.create_hosted_zone(Name='test.com', CallerReference='1', DelegationSetId='1')
    conn.create_hosted_zone(Name='test.com', CallerReference='2', DelegationSetId='1')
```

[Reference](https://docs.aws.amazon.com/sdk-for-ruby/v3/api/Aws/Route53/Types/ConflictingDomainExists.html)

Introduced test covers:
- Two create_hosted_zone requests on the same delegation set == should raise exception
- Two create_hosted_zone reqs with unnamed delegation set ids == should not raise
- Two create_hosted_zone reqs with different, named delegation set ids == should not raise

I read & followed the contribution guide, `make format` and tests pass but lmk if I missed anything!

### DIFF
--- a/moto/route53/exceptions.py
+++ b/moto/route53/exceptions.py
@@ -12,6 +12,19 @@ class Route53ClientError(RESTError):
         super().__init__(*args, **kwargs)
 
 
+class ConflictingDomainExists(Route53ClientError):
+    """Domain already exists."""
+
+    code = 400
+
+    def __init__(self, domain_name: str, delegation_set_id: str) -> None:
+        message = (
+            f"Cannot create hosted zone with DelegationSetId DelegationSetId:{delegation_set_id} as the DNSName"
+            f"{domain_name} conflicts with existing ones sharing the delegation set"
+        )
+        super().__init__("ConflictingDomainExists", message)
+
+
 class InvalidInput(Route53ClientError):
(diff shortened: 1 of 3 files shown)


bblommers:
Hi @codeocelot!

There is `make lint` as well, for additional validation, and that currently fails: https://github.com/getmoto/moto/actions/runs/7648573846/job/20841579582?pr=7249

Note the `str | None` error in the linked GithubAction - this error will not show up if you're only testing a recent Python version, but it would not work for Python < 3.10.

bblommers:
Hope you don't mind @codeocelot - I pushed a fix for the typing, just make sure it's part of the release going out today

bblommers (approved):
LGTM - thanks for the PR!

codeocelot:
Apologies I didn't come back earlier and check in. @bblommers thanks for the help resolving the linting issues.  Glad it's already released!

PR #7251, opened 2024-01-26
SNS: allow SetSubscriptionAttributes to unset FilterPolicy

bentsku:
We've had a report that a user could not "unset" a FilterPolicy by setting it to an empty string like they could in AWS.
This was due to always trying to decode the FilterPolicy as JSON, raising a `JSONDecodeError`. 

I've introduce a conditional check for the value, and manually set the filter policy and the matcher to None to not have a filter in place anymore. 

I've also sneaked a little change to what is returned from `GetSubscriptionAttributes`: SNS does not return the FilterPolicy if it is not set (None or empty string), and will not return the FilterPolicyScope either in that case. Also, if a FilterPolicy is set without manually setting a FilterPolicyScope, it will return the default filter policy scope as well in the fields. This was validated against AWS.

### DIFF
--- a/moto/sns/models.py
+++ b/moto/sns/models.py
@@ -732,8 +732,17 @@ class SNSBackend(BaseBackend):
             raise SNSNotFoundError(
                 "Subscription does not exist", template="wrapped_single_error"
             )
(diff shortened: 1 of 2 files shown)


bblommers (approved):
LGTM!

PR #7252, opened 2024-01-26
S3: Adding notification for eventbridge

tsugumi-sys:
Implement S3 EventBridge notification.
- Expand notification configuration API.
- Send notifications to EventBridge backend (skip tests when moto is in proxy or server mode). 
- Testing on `put_object` notification.

Document of EventBridge event message structure: https://docs.aws.amazon.com/AmazonS3/latest/userguide/ev-events.html

Document of EventBridge notification: https://docs.aws.amazon.com/AmazonS3/latest/userguide/EventBridge.html


moto/s3/models.py
```diff
@@ -2315,11 +2320,7 @@ def put_bucket_notification_configuration(
          - AWSLambda
          - SNS
          - SQS
-
```
bblommers, line 2318: We should keep the documentation that only `s3:ObjectCreated:Copy` and `s3:ObjectCreated:Put` are supported. 

Those are still the only two methods where we actually invoke `s3/notifications.py::send_event`.

moto/s3/notifications.py
```diff
@@ -157,6 +164,97 @@ def _send_sns_message(
...
+def _send_event_bridge_message(
+    account_id: str,
+    bucket: Any,
+    event_name: str,
+    key: Any,
+) -> None:
+    try:
+        from moto.events.models import events_backends
+
+        event = copy.deepcopy(_BASE_EVENT_MESSAGE)
```
bblommers, line 176: The `_BASE_EVENT_MESSAGE`-import should also be local

moto/s3/notifications.py
```diff
@@ -157,6 +164,97 @@ def _send_sns_message(
...
+        event["region"] = bucket.region_name
+        event["resources"] = [f"arn:aws:s3:::{bucket.name}"]
+        event["detail"] = {
+            "version": "0",
+            "bucket": {"name": bucket.name},
+            "object": {
+                "key": key.name,
+                "size": key.size,
+                "eTag": key.etag.replace('"', ""),
+                "version-id": "IYV3p45BT0ac8hjHg1houSdS1a.Mro8e",
```
bblommers, line 190: Can we use `key.version_id` here?

bblommers (changes requested):
This looks really great @tsugumi-sys - just got a few minor comments.

moto/s3/models.py
tsugumi-sys: @bblommers 
That's true, thanks

bblommers (approved):
That looks great - thank you @tsugumi-sys!

PR #7253, opened 2024-01-26
S3: Adding a validation method for s3 notification event name

tsugumi-sys:
- Adding a validation method for `S3NotificationEvent`.
- Adding event name validation where necessary.

### DIFF
--- a/moto/s3/notifications.py
+++ b/moto/s3/notifications.py
@@ -3,6 +3,8 @@ from datetime import datetime
 from enum import Enum
 from typing import Any, Dict, List
 
+from . import exceptions
+
 _EVENT_TIME_FORMAT = "%Y-%m-%dT%H:%M:%S.%f"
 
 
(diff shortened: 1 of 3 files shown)


tsugumi-sys:
@bblommers 
I think that the CI failures are not related to my changes?

moto/s3/notifications.py
```diff
@@ -89,6 +103,9 @@ def send_event(
     if bucket.notification_configuration is None:
         return
 
+    if not S3NotificationEvent.is_event_valid(event_name):
```
bblommers, line 106: This doesn't seem necessary - we always control when `send_event` is invoked, so there's no point in raising an exception

moto/s3/notifications.py
```diff
@@ -49,11 +51,23 @@ class S3NotificationEvent(str, Enum):
...
+        if event_name in [e[:3] for e in self.events()]:
+            return True
+        return False
+
 
 def _get_s3_event(
     event_name: str, bucket: Any, key: Any, notification_id: str
 ) -> Dict[str, List[Dict[str, Any]]]:
     etag = key.etag.replace('"', "")
+    if not S3NotificationEvent.is_event_valid(event_name):
```
bblommers, line 69: Same here

moto/s3/responses.py
```diff
@@ -2140,7 +2140,7 @@ def _notification_config_from_body(self) -> Dict[str, Any]:
                         n["Event"] = [n["Event"]]
 
                     for event in n["Event"]:
-                        if event not in S3NotificationEvent.events():
+                        if not S3NotificationEvent.is_event_valid(event):
```
bblommers, line 2143: This logic is now duplicated into the `Notification`-constructor. That seems to a more logical place to validate this value - so I think we can remove the check here. Or am I missing something?

bblommers (changes requested):

moto/s3/responses.py
tsugumi-sys: That also sounds good to me :) Thanks

bblommers (approved):

PR #7263, opened 2024-01-29
Mark `mock_aws` for explicit reexport

andersk:
This is required for mypy to recognize this as exported with [`--no-implicit-reexport`](https://mypy.readthedocs.io/en/stable/command_line.html#cmdoption-mypy-no-implicit-reexport) (implied by `--strict`).


bblommers (approved):
LGTM - thanks @andersk!

PR #7266, opened 2024-01-29
Update example in README

Apakottur:
Update README to use new decorator `mock_aws` instead of the old `mock_s3`.
This change was introduced in the newest version - 5.0.0


bblommers (approved):
Good catch @Apakottur - thanks for the PR, and welcome to Moto!

PR #7280, opened 2024-01-30
Fix ecs container memory validation 7272

MikiPWata:
fixes #

### DIFF
--- a/moto/ecs/models.py
+++ b/moto/ecs/models.py
@@ -1178,8 +1178,9 @@ class EC2ContainerServiceBackend(BaseBackend):
                 raise EcsClientException(
                     f"Tasks using the Fargate launch type do not support pidMode '{pid_mode}'. The supported value for pidMode is 'task'."
                 )
-
-        self._validate_container_defs(memory, container_definitions)
+        self._validate_container_defs(
+            memory, container_definitions, requires_compatibilities
+        )
 
         if family in self.task_definitions:
             last_id = self._get_last_task_definition_revision_id(family)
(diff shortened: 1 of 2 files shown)


bblommers (commented):
Hi @MikiPWata!

If I run the new test against AWS I still get the memoryReservation error:
```
Invalid setting for container 'hello_world2'. At least one of 'memory' or 'memoryReservation' must be specified.
```

The docs are indeed a bit vague. It does say `If using the Fargate launch type, this parameter is optional.` - but maybe they just mean `it's optional - as longo as you specify the memoryReservation instead`?

MikiPWata:
Hi @bblommers !
Thanks for the comment!
I'm pretty sure we don't need to specify neither memory / memoryResevation for container definitions when running on Fargate(I actually do have containers without specifying memory/memoryReservation on ECS)

MikiPWata:
@bblommers 
Have you specified EC2 for the test you ran? or Fargate?

bblommers:
@MikiPWata I tried the test that you added, `test_register_task_definition_memory_validation_fargate`, so specifically for FarGate.

Could it be an API change, that your containers were started without memory a long time ago, and AWS only added this validation later?

MikiPWata:
@bblommers 
I see. I do update task definitions and containers regularly but I don't get any memory specification issues....

MikiPWata:
@bblommers 
I think I know what I should do.
we need to specify either or both memory or memoryReservation on task-container level if you did not specify memory on task-definition level
image

I need to check if memory was set on task definition and if not, I need to check if memory or memoryReservation is set in container defs

bblommers (approved):
LGTM - thank you for investigating and fixing this @MikiPWata!

PR #7281, opened 2024-01-30
EC2: Add availability zones for eu-central-2

viren-nadkarni:
See https://github.com/localstack/localstack/issues/10122

Information obtained from https://www.cloudregions.io/aws/region/eu-central-2


bblommers (approved):
LGTM!

PR #7284, opened 2024-01-30
Techdebt: DynamoDB: Reduce number of client-creations during tests

bblommers:
Reduces the runtime of just the DynamoDB tests from 38 secs to 17 secs


bblommers:
Initializing a session is the part that takes the most time when creating a new client. And Moto forces boto3 to recreate a session at the start of every mock, so we spend a lot of time creating sessions/loading data/parsing JSON/etc. 

There is probably a better way to improve the client creation that doesn't involve rewriting every test, so I'm going to close this for now.

PR #7285, opened 2024-01-31
Remove python-jose and ecdsa dependencies

### LINKED ISSUE #7244: Unmaintained dependency python-jose
Moto with extra cognitoidp depends on `python-jose`, which in turn depends on `ecdsa`, which in turn was flagged by dependabot: https://github.com/advisories/GHSA-wj6h-64fc-37mp

It doesn't seem like there will be a fix in python-ecdsa and python-jose hasn't had any releases for years, so I am wondering if it is possible to replace python-jose with [joserfc](https://github.com/authlib/joserfc) or something similar? I could try to make a PR if it sounds reasonable. :)

estyrke:
Fixes #7244


tests/test_ec2/test_key_pairs.py
```diff
@@ -29,6 +29,7 @@
 
 RSA_PUBLIC_KEY_RFC4716 = b"""\
 ---- BEGIN SSH2 PUBLIC KEY ----
+cOmmENt: moto@github.com
```
bblommers, line 32: Can we have a few more tests with example RFC4716-compliant keys here, to verify that they all work?
I'm thinking with comment, without any comment, with commend + `x-comment`, etc.

I noticed that the spec has a few examples: 
https://www.rfc-editor.org/rfc/rfc4716#section-3.6

bblommers (changes requested):
Hi @estyrke, thanks for the PR! Please see my comments

PR #7294, opened 2024-02-01
docs: proxy - add excludes via NO_PROXY

morph027:


bblommers (changes requested):
Hi @morph027, `NO_PROXY` is language specific as far as I know - is it Golang? Definitely not something supported by all languages.

If you do want to add it to the docs, just reword it to make this explicit: `Some languages may support a way to ... , such as NO_PROXY in Golang, ... `

morph027:
You're right, it's an instruction for libs and tool based on [`curl`](https://curl.se/docs/tutorial.html). I'll add an update.

morph027:
We tested it w/ the PHP AWS SDK, boto3 and awscli (ok, based on boto3).

bblommers (approved):
That works for me. Thank you for contributing @morph027!

PR #6303, opened 2023-05-09
fix copy in place for S3 bucket with encryption

bentsku:
This PR would fix #6300

With S3 new default encryption for bucket, I've overlooked one case where the bucket would have a manually set encryption configuration. In that case, S3 will allow copying in place. For now, it is rather simple to fix, we just need to check if the bucket has encryption, but this will become a bit more complicated when we implement the default encryption with our bucket has well. 

I've also sneaked a little fix regarding `BucketKeyEnabled` header, as it should not be returned when the encryption is not `aws:kms` (verified with AWS as well).

PR #7303, opened 2024-02-02
S3: Allow in place copies when bucket versioning is enabled

### LINKED ISSUE #7300: self-copying an old version works in S3, doesn't work in moto
## Short description
In S3, you can copy from an old version of an object as the source to the same key as the destination. Moto however gives an error.

## Reproducer

```python
from random import getrandbits
import os
import boto3

BUCKET_NAME="moto-self-copy-reproducer-bucket"

def create_bucket(client):
    client.create_bucket(
        Bucket=BUCKET_NAME,
    )
    client.put_bucket_versioning(
        Bucket=BUCKET_NAME,
        VersioningConfiguration={
            "MFADelete": "Disabled",
            "Status": "Enabled",
        },
    )

def print_key(client, keyname):
    resp=client.get_object(
        Bucket=BUCKET_NAME,
        Key=keyname,
    )
    print("\nthe key contains:", resp["Body"].read())

def main():
    client = boto3.client("s3")
    keyname = f"moto-self-copy-reproducer/key-{getrandbits(16)}"
    create_bucket(client)
    client.put_object(
        Body="Hello, world!",
        Bucket=BUCKET_NAME,
        Key=keyname
    )
    print_key(client, keyname)
    client.put_object(
        Body="New content",
        Bucket=BUCKET_NAME,
        Key=keyname
    )
    print_key(client, keyname)
    resp = client.list_object_versions(
        Bucket=BUCKET_NAME,
        Prefix=keyname,
    )
    print(resp)
    ver_id=next(v for v in resp["Versions"] if v["IsLatest"] == False)["VersionId"]
    print("reverting key to version:", ver_id)
    client.copy_object(
        Bucket=BUCKET_NAME,
        Key=keyname,
        CopySource={
            "Bucket": BUCKET_NAME,
            "Key": keyname,
            "VersionId": ver_id,
        },
    )
    print_key(client, keyname)
if __name__ == "__main__":
    main()
```

pyproject.toml:

<details>

```toml
[tool.poetry]
name = "moto-reproducer"
version = "0.1.0"
description = ""
authors = ["Your Name <you@example.com>"]
readme = "README.md"

[tool.poetry.dependencies]
python = "^3.11"
moto = {extras = ["server","all"], version = "^5.0.0"}
boto3 = "^1.34.0"

[build-system]
requires = ["poetry-core"]
build-backend = "poetry.core.masonry.api"
```

</details>

## Actual output

If you then do `poetry run moto_server` and parallel to it you do `AWS_ENDPOINT_URL=http://127.0.0.1:5000 AWS_ACCESS_KEY_ID=test AWS_SECRET_ACCESS_KEY=test poetry run python reproducer.py`, you get the output:

```
the key contains: b'Hello, world!'

the key contains: b'New content'
{'ResponseMetadata': {'RequestId': 'TgfmqeoLiFA6Ub3yxJjHWlHugX9LJ84AwSmGBNQMSM269aEf5p8R', 'HTTPStatusCode': 200, 'HTTPHeaders': {'server': 'Werkzeug/3.0.1 Python/3.11.6', 'date': 'Thu, 01 Feb 2024 23:41:46 GMT', 'x-amzn-requestid': 'TgfmqeoLiFA6Ub3yxJjHWlHugX9LJ84AwSmGBNQMSM269aEf5p8R', 'content-type': 'text/html; charset=utf-8', 'content-length': '1226', 'access-control-allow-origin': '*', 'connection': 'close'}, 'RetryAttempts': 0}, 'IsTruncated': False, 'KeyMarker': '', 'VersionIdMarker': '', 'Versions': [{'ETag': '"46f58a4ea62a138e33ee79e136333c5c"', 'Size': 11, 'StorageClass': 'STANDARD', 'Key': 'moto-self-copy-reproducer/key-35689', 'VersionId': '5b75650d-1d2a-43f7-b39c-96f39f24dd54', 'IsLatest': True, 'LastModified': datetime.datetime(2024, 2, 1, 23, 41, 46, tzinfo=tzutc()), 'Owner': {'DisplayName': 'webfile', 'ID': '75aa57f09aa0c8caeab4f8c24e99d10f8e7faeebf76c078efc7c6caea54ba06a'}}, {'ETag': '"6cd3556deb0da54bca060b4c39479839"', 'Size': 13, 'StorageClass': 'STANDARD', 'Key': 'moto-self-copy-reproducer/key-35689', 'VersionId': '3613e99e-c3eb-4ebd-bacd-49cc874659b4', 'IsLatest': False, 'LastModified': datetime.datetime(2024, 2, 1, 23, 41, 46, tzinfo=tzutc()), 'Owner': {'DisplayName': 'webfile', 'ID': '75aa57f09aa0c8caeab4f8c24e99d10f8e7faeebf76c078efc7c6caea54ba06a'}}], 'Name': 'moto-self-copy-reproducer-bucket', 'Prefix': 'moto-self-copy-reproducer/key-35689', 'Delimiter': 'None', 'MaxKeys': 1000}
reverting key to version: 3613e99e-c3eb-4ebd-bacd-49cc874659b4
  File "src/moto-reproducer/reproducer.py", line 61, in <module>
    main()
  File "src/moto-reproducer/reproducer.py", line 49, in main
    client.copy_object(
  File ".cache/pypoetry/virtualenvs/moto-reproducer-VFSGIBU6-py3.11/lib/python3.11/site-packages/botocore/client.py", line 553, in _api_call
  File ".cache/pypoetry/virtualenvs/moto-reproducer-VFSGIBU6-py3.11/lib/python3.11/site-packages/botocore/client.py", line 1009, in _make_api_call
botocore.exceptions.ClientError: An error occurred (InvalidRequest) when calling the CopyObject operation: This copy request is illegal because it is trying to copy an object to itself without changing the object's metadata, storage class, website redirect location or encryption attributes.
```

## Expected output

If you use proper AWS instead, you get:

```
AWS_PROFILE=dev poetry run python reproducer.py 

the key contains: b'Hello, world!'

the key contains: b'New content'
{[censored]]}
reverting key to version: gTwd._wy7XYL3K6d.nAjyI2owY868nh3

the key contains: b'Hello, world!'
```

## Cause of the issue

I would say the problem lies in the check linked below: it doesn't check if `src_key.version_id` has some content or not.

https://github.com/getmoto/moto/blob/911b5f19c0564e9f2e5ca7119c272784096630b8/moto/s3/models.py#L2663

## Why would this be useful?

I'm trying to implement a functionality that uses this copy endpoint to roll back a file's version to a specific older date. The copy endpoint seems like the most efficient way to me.

arpad-m:
AWS allows self-copies for keys when the bucket has versioning enabled.

Fixes #7300

Related earlier work: #6303

### DIFF
--- a/moto/s3/models.py
+++ b/moto/s3/models.py
@@ -108,7 +108,7 @@ class FakeKey(BaseModel, ManagedState):
         storage: Optional[str] = "STANDARD",
         etag: Optional[str] = None,
         is_versioned: bool = False,
-        version_id: str = "null",
+        version_id: Optional[str] = None,
         max_buffer_size: Optional[int] = None,
         multipart: Optional["FakeMultipart"] = None,
         bucket_name: Optional[str] = None,
@@ -170,7 +170,7 @@ class FakeKey(BaseModel, ManagedState):
 
     @property
     def version_id(self) -> str:
-        return self._version_id
+        return self._version_id or "null"
 
     @property
     def value(self) -> bytes:
@@ -2674,7 +2674,7 @@ class S3Backend(BaseBackend, CloudWatchMetricProvider):
                     mdirective == "REPLACE",
                     website_redirect_location,
                     bucket.encryption,  # S3 will allow copy in place if the bucket has encryption configured
-                    src_key.version_id != "null" and bucket.is_versioned,
+                    src_key._version_id and bucket.is_versioned,
                 )
             ):
                 raise CopyObjectMustChangeSomething
--- a/moto/s3/models.py
+++ b/moto/s3/models.py
@@ -2674,6 +2674,7 @@ class S3Backend(BaseBackend, CloudWatchMetricProvider):
                     mdirective == "REPLACE",
                     website_redirect_location,
                     bucket.encryption,  # S3 will allow copy in place if the bucket has encryption configured
+                    bucket.is_versioned,
                 )
             ):
                 raise CopyObjectMustChangeSomething
--- a/moto/s3/models.py
+++ b/moto/s3/models.py
@@ -2674,7 +2674,7 @@ class S3Backend(BaseBackend, CloudWatchMetricProvider):
                     mdirective == "REPLACE",
                     website_redirect_location,
                     bucket.encryption,  # S3 will allow copy in place if the bucket has encryption configured
-                    bucket.is_versioned,
+                    src_key.version_id != "null" and bucket.is_versioned,
                 )
             ):
                 raise CopyObjectMustChangeSomething
--- a/tests/test_s3/test_s3_copyobject.py
+++ b/tests/test_s3/test_s3_copyobject.py
@@ -810,7 +810,7 @@ def test_copy_object_in_place_with_versioning():
 
     response = client.copy_object(
         Bucket=bucket_name,
-        CopySource={"Bucket": bucket_name, "Key": key},
+        CopySource={"Bucket": bucket_name, "Key": key, "VersionId": version_id},
         Key=key,
     )
     assert response["ResponseMetadata"]["HTTPStatusCode"] == 200
--- a/tests/test_s3/test_s3_copyobject.py
+++ b/tests/test_s3/test_s3_copyobject.py
@@ -791,6 +791,12 @@ def test_copy_object_in_place_with_versioning():
     client.create_bucket(Bucket=bucket_name)
     key = "source-key"
 
+    response = client.put_object(
+        Body=b"",
+        Bucket=bucket_name,
+        Key=key,
+    )
+
     response = client.put_bucket_versioning(
         Bucket=bucket_name,
         VersioningConfiguration={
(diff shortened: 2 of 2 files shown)


bblommers (changes requested):
Hi @arpad-m! The new test, `test_copy_object_in_place_with_versioning` currently fails against AWS.

It does work if you pass in a VersionId, so I think your first instinct was correct.

arpad-m:
Hmm yeah it seems that the S3 bucket I was testing it against has server side encryption enabled. I have changed it to check if the `VersionId` is specified instead.

By making the internal `_version_id` to be nullable we can now distinguish `VersionId: "null"` from `VersionId` not being specified. [According to docs](https://docs.aws.amazon.com/AmazonS3/latest/userguide/Versioning.html#versioning-states), "null" is the VersionId that AWS assigns to objects when versioning is being enabled, so it's a naturally occuring ID.
 I would presume that AWS treats former as if `VersionId` is actually specified, but I can't test it as I don't have a bucket at hand with versioning and no encryption. If my assumption is not correct, we can remove the last commit.

arpad-m:
I have added a commit that has a test case for the `null` case. If the test works on AWS, then the extra commit was worth it, if not then AWS *does* treat `null` differently from the other possible values for a version ID.

bblommers (approved):
Yes - looks like that passes against AWS, so this all looks good.

Thank you for fixing this @arpad-m!

PR #7315, opened 2024-02-06
Techdebt: Use typing for event message

tsugumi-sys:
- The typing for event messages is applided in `moto/events/models.py`.
- A base event message is added.
- The typing of `time` field in event message structure is expanded for internal use in moto.


bblommers (approved):
LGTM! Thanks @tsugumi-sys

PR #7317, opened 2024-02-06
Support tags in Glue create_database

### LINKED ISSUE #7316: glue.create_database does not create tags
The AWS API allows to specify resource tags when creating a Glue database. However, moto will ignore those tags and not tag the database. As a demonstration, the following test fails:
```
from moto import mock_aws
import boto3

@mock_aws
def test_glue():
    database_name = "test"
    expected_tags = {"key": "value"}

    client = boto3.client("glue", region_name="eu-west-1")
    client.create_database(
        DatabaseInput={"Name": database_name, "LocationUri": "s3://somewhere"},
        Tags=expected_tags,
    )
    response = client.get_tags(ResourceArn=f"arn:aws:glue:eu-west-1:123456789012:database/{database_name}")
    assert response["Tags"] == expected_tags
```

Pytest result:
```
>       assert response["Tags"] == expected_tags
E       AssertionError: assert {} == {'key': 'value'}
E         
E         Right contains 1 more item:
E         {'key': 'value'}

glue_test.py:15: AssertionError
```

I use Moto 5.0.1.

MartinAltmayerTMH:
This PR resolves #7316.

I believe that some other taggable Glue resources are affected by the same problem. So if this PR goes into the right direction, I might add tagging-support for those resources as well.

### DIFF
--- a/moto/glue/models.py
+++ b/moto/glue/models.py
@@ -135,11 +135,8 @@ class GlueBackend(BaseBackend):
             database_name, database_input, catalog_id=self.account_id
         )
         self.databases[database_name] = database
-        if tags:
-            resource_arn = f"arn:aws:glue:{self.region_name}:{self.account_id}:database/{database_name}"
-            self.tagger.tag_resource(
-                resource_arn, self.tagger.convert_dict_to_tags_input(tags)
-            )
+        resource_arn = f"arn:aws:glue:{self.region_name}:{self.account_id}:database/{database_name}"
+        self.tag_resource(resource_arn, tags)
         return database
 
     def get_database(self, database_name: str) -> "FakeDatabase":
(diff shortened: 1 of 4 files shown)


moto/glue/models.py
```diff
@@ -132,6 +135,11 @@ def create_database(
             database_name, database_input, catalog_id=self.account_id
         )
         self.databases[database_name] = database
+        if tags:
```
bblommers, line 138: It would be better to use `self.tag_resource(resource_arn, tags)` here.

No need for the if-statement, less code duplication, and it ensures that a call to `get_tags` always succeeds, regardless of whether the user supplied tags in the first place.

Keep in mind that you may need to change the signature of the `tag_resource` method, because the `tags`-argument then becomes `Optional`.

bblommers (changes requested):
Hi @MartinAltmayerTMH, thanks for the PR! Just got a minor nit - the approach and test look good.

moto/glue/models.py
MartinAltmayerTMH: Thanks for your quick feedback! Using `tag_resource` is a good idea, of course. I've implemented your suggestion.

MartinAltmayerTMH:
I've had a look at various other Glue resources:
- `create_table` and `create_partition` do not support tags in the AWS API.
- Many other resources already support tags in moto (e.g. `create_job`, `create_crawler`, `create_schema`)

So I believe there are no more `create` methods where tagging is missing.

bblommers (approved):
LGTM - thanks again @MartinAltmayerTMH!

PR #7323, opened 2024-02-08
Add ASG Scheduled Scaling Action batch APIs

hearde:
Add support for batch put and batch delete APIs for Scheduled Scaling Actions on AutoScaling Groups. Delete nonexistent action now fails. Also adds support for the `TimeZone` property for actions. The behavior changes I've made in this PR are to align with the behavior I can observe by experimenting with the API.

- [Batch put API reference](https://docs.aws.amazon.com/autoscaling/ec2/APIReference/API_BatchPutScheduledUpdateGroupAction.html)
- [Batch delete API reference](https://docs.aws.amazon.com/autoscaling/ec2/APIReference/API_BatchDeleteScheduledAction.html)

### DIFF
--- a/moto/autoscaling/models.py
+++ b/moto/autoscaling/models.py
@@ -321,11 +321,11 @@ class FakeScheduledAction(CloudFormationModel):
         max_size: Optional[int],
         min_size: Optional[int],
         scheduled_action_name: str,
-        start_time: str,
-        end_time: str,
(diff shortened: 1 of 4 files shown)


bblommers (approved):
LGTM. Welcome to Moto @hearde, and thank you for adding these features!

PR #7326, opened 2024-02-09
feat: use AWS Lambda entry point if ImageConfig.EntryPoint is set

oxidase:
With the proposed change it is possible to use `public.ecr.aws/lambda/python:3.12` image in a hacky way as
```
@mock.patch.dict(
    os.environ,
    {
        "MOTO_DOCKER_LAMBDA_IMAGE": "public.ecr.aws/lambda/python",
    },
)
def test_send_message():
    ...

    func = lambda_.create_function(
        FunctionName=function_name,

        Runtime=f"{sys.version_info.major}.{sys.version_info.minor}",
        ImageConfig={
            "EntryPoint": [
                "/var/lang/bin/python3",
                "-c",
                f"import sys, json; from {handler_package} import {handler_callback} as cb; cb(json.loads(sys.argv[2]), None)",
            ],
        },
   )
...
```

### DIFF
--- a/moto/awslambda/models.py
+++ b/moto/awslambda/models.py
@@ -1001,6 +1001,10 @@ class LambdaFunction(CloudFormationModel, DockerModel):
                             "host.docker.internal": "host-gateway"
                         }
 
+                    # Change entry point if requested
+                    if self.image_config.entry_point:
+                        run_kwargs["entrypoint"] = self.image_config.entry_point
+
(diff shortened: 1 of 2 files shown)


bblommers (approved):
LGTM - thank you for adding this to Moto @oxidase!

oxidase:
Thanks @bblommers !

PR #7329, opened 2024-02-09
Feature: RDS Proxy

JoshLevyMN:
Adds initial support for RDS Proxy, including create, describe, and tagging.


bblommers (changes requested):
Hi @JoshLevyMN! Please see my comments.

It looks like the linter is failing as well: https://github.com/getmoto/moto/actions/runs/7849632093/job/21423405368?pr=7329

JoshLevyMN:
Hi @bblommers, thanks for the feedback, it should all be addressed with the latest commit.

bblommers (approved):
Awesome - thank you for adding these features to Moto @JoshLevyMN!

PR #7351, opened 2024-02-17
Update requirements.txt to reference new location of requirements

adamtheturtle:


bblommers (approved):
Good catch @adamtheturtle, thanks for the PR!

PR #7352, opened 2024-02-17
Add numbers as `Decimal` objects

jtherrmann:
Fixes https://github.com/getmoto/moto/issues/7349

TODO:

- [ ] I've only updated the `DynamoType.__add__` method, but there are other places where floating-point arithmetic is performed that should probably be updated as well, e.g. `DynamoType.__sub__`.
- [ ] Add test cases for these changes?
- [ ] Fix failing tests

On my machine, `make test` fails with:

```
moto/dynamodb/models/dynamo_type.py:109:48: error: Unsupported left operand type for + ("object")  [operator]
Found 1 error in 1 file (checked 900 source files)
make: *** [Makefile:30: lint] Error 1
```

### DIFF
--- a/moto/dynamodb/models/dynamo_type.py
+++ b/moto/dynamodb/models/dynamo_type.py
@@ -100,8 +100,12 @@ class DynamoType(object):
         if self.type != other.type:
             raise TypeError("Different types of operandi is not allowed.")
         if self.is_number():
-            self_value = decimal.Decimal(self.value) if "." in self.value else int(self.value)
-            other_value = decimal.Decimal(other.value) if "." in other.value else int(other.value)
+            self_value = (
+                decimal.Decimal(self.value) if "." in self.value else int(self.value)
+            )
+            other_value = (
+                decimal.Decimal(other.value) if "." in other.value else int(other.value)
+            )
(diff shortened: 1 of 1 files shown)


bblommers:
Hi @jtherrmann, thanks for getting this started. Do you want to tackle the TODO's yourself, or are you OK if someone else picks this up?

jtherrmann:
@bblommers I'd love it if someone else could finish it off!

bblommers:
@jtherrmann No problem! I wasn't able to push to your branch, so I've created #7365 that contains your commit + some tests.

Regarding your TODO's:
 > but there are other places where floating-point arithmetic is performed that should probably be updated as well

Only `ADD` is supported/required for numerical values actually, so I think this is all that is necessary. From the docs:

> If the existing attribute is a number, and if Value is also a number, then Value is mathematically added to the existing attribute. If Value is a negative number, then it is subtracted from the existing attribute.

The failing test was `mypy` complaining, because of a known issue where it can't find a suitable parent-type: https://github.com/python/mypy/issues/3816.
As far as I can tell, the only way around this is to manually decide on a suitable type (`Union[Decimal, int]`)

Issue #7363, opened 2024-02-19
Implement S3 EventBridge Notification


tsugumi-sys:
Since moto currently supports only `Object Created` event, let's add the other events described [here](https://docs.aws.amazon.com/AmazonS3/latest/userguide/EventBridge.html).

https://docs.aws.amazon.com/AmazonS3/latest/userguide/EventBridge.html

Event type | Description | PR
-- | -- | --
Object Created | An object was created.The reason field in the event message structure indicates which S3 API was used to create the object: PutObject, POST Object, CopyObject, or CompleteMultipartUpload. | PutObject: https://github.com/getmoto/moto/pull/7252
Object Deleted (DeleteObject)Object Deleted (Lifecycle expiration) | An object was deleted.When an object is deleted using an S3 API call, the reason field is set to DeleteObject. When an object is deleted by an S3 Lifecycle expiration rule, the reason field is set to Lifecycle Expiration. For more information, see Expiring objects.When an unversioned object is deleted, or a versioned object is permanently deleted, the deletion-type field is set to Permanently Deleted. When a delete marker is created for a versioned object, the deletion-type field is set to Delete Marker Created. For more information, see Deleting object versions from a versioning-enabled bucket.
Object Restore Initiated | An object restore was initiated from S3 Glacier or S3 Glacier Deep Archive storage class or from S3 Intelligent-Tiering Archive Access or Deep Archive Access tier. For more information, see Working with archived objects.
Object Restore Completed | An object restore was completed.
Object Restore Expired | The temporary copy of an object restored from S3 Glacier or S3 Glacier Deep Archive expired and was deleted.
Object Storage Class Changed | An object was transitioned to a different storage class. For more information, see Transitioning objects using Amazon S3 Lifecycle.
Object Access Tier Changed | An object was transitioned to the S3 Intelligent-Tiering Archive Access tier or Deep Archive Access tier. For more information, see Amazon S3 Intelligent-Tiering.
Object ACL Updated | An object's access control list (ACL) was set using PutObjectACL. An event is not generated when a request results in no change to an object’s ACL. For more information, see Access control list (ACL) overview.
Object Tags Added | A set of tags was added to an object using PutObjectTagging. For more information, see Categorizing your storage using tags.
Object Tags Deleted | All tags were removed from an object using DeleteObjectTagging. For more information, see Categorizing your storage using tags.

tsugumi-sys:
I'm working on `Object Storage Class Changed` event.

[...]

PR #7368, opened 2024-02-20
S3: Sending notification to EventBridge `ObjectCreated:Post`

tsugumi-sys:
Sending a notification when an object is uploaded to S3 via POST request.

related: #7363

### DIFF
--- a/moto/s3/responses.py
+++ b/moto/s3/responses.py
@@ -2259,6 +2259,7 @@ class S3Response(BaseResponse):
                 multipart=multipart,
                 encryption=multipart.sse_encryption,
                 kms_key_id=multipart.kms_key_id,
+                request_method=request.method,
             )
             key.set_metadata(multipart.metadata)
 
--- a/moto/s3/responses.py
+++ b/moto/s3/responses.py
@@ -2262,7 +2262,6 @@ class S3Response(BaseResponse):
                 multipart=multipart,
                 encryption=multipart.sse_encryption,
                 kms_key_id=multipart.kms_key_id,
-                request_method=request.method,
             )
             key.set_metadata(multipart.metadata)
 
--- a/moto/s3/responses.py
+++ b/moto/s3/responses.py
@@ -141,7 +141,6 @@ ACTION_MAP = {
             "restore": "RestoreObject",
             "uploadId": "PutObject",
             "select": "SelectObject",
-            "DEFAULT": "PutObject",
         },
     },
     "CONTROL": {
--- a/moto/s3/responses.py
+++ b/moto/s3/responses.py
@@ -141,6 +141,7 @@ ACTION_MAP = {
             "restore": "RestoreObject",
             "uploadId": "PutObject",
             "select": "SelectObject",
+            "DEFAULT": "PutObject",
         },
     },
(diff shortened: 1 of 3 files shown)


bblommers (changes requested):
If I'm reading the documentation right, the `POST object`-event only happens for a regular  HTTP POST request. It looks like your test verifies the CompeteMultiPart-action, which is a separate event.

From the [docs](https://docs.aws.amazon.com/AmazonS3/latest/API/RESTObjectPOST.html):
> The POST operation adds an object to a specified bucket by using HTML forms. [...] Parameters are passed [...] as form fields to POST in the multipart/form-data encoded message body.

So in normal language (i.e., Python code): 
```
requests.post(f"https://{bucket_name}.s3.amazonaws.com/", {"key": "key2", "file": "nothing"})
```

moto/s3/responses.py
```diff
@@ -2259,6 +2259,7 @@ def _key_response_post(
                 multipart=multipart,
                 encryption=multipart.sse_encryption,
                 kms_key_id=multipart.kms_key_id,
+                request_method=request.method,
```
bblommers, line 2262: This would send a `POST` event immediately after the `CompleteMultipartUpload`-action is finished. If I understand it correctly, AWS will only send a single event in this case. So we should avoid sending a `POST` event completely here.

tests/test_s3/test_s3_eventbridge_integration.py
```diff
@@ -1,16 +1,23 @@
...
+from tests.test_s3.test_s3_multipart import reduced_min_part_size
 
 REGION_NAME = "us-east-1"
+REDUCED_PART_SIZE = 256
 
 
 @mock_aws
-def test_pub_object_notification():
+def test_put_object_notification_ObjectCreated_PUT():
+    if not settings.TEST_DECORATOR_MODE:
```
bblommers, line 18: What's the reason behind skipping the tests in ServerMode?

bblommers (changes requested):
This is the part of the code where we handle a multipart/formdata POST request, and where we should send an event:

https://github.com/getmoto/moto/blob/ce074824a45bd36780b0061b896b9b1117478b14/moto/s3/responses.py#L1119

tests/test_s3/test_s3_eventbridge_integration.py
tsugumi-sys: Sorry, this part was added by mistake.

tsugumi-sys:
hmmm, I don't understand why the tests failed?
https://github.com/getmoto/moto/actions/runs/8058630565/job/22011891131?pr=7368
https://github.com/getmoto/moto/actions/runs/8058630565/job/22011891477?pr=7368


I tried to reproduce the error in my local computer, but it didn't fail....

bblommers:
@tsugumi-sys Looks like that's a issue specific to Python 3.11 and up. When Moto reads a file from a form-upload, it doesn't close the stream. I guess it is closed automatically in older versions of Python, somehow.

I've opened #7405 with a fix - once that's merged, I'll merge this one as well.

Thank you for the PR!

bblommers (approved):

PR #7379, opened 2024-02-21
CognitoIDP: Add `SMS_MFA` support to admin_initiate_auth and respond_to_auth_challenge methods

AriFreyr:
- Return `SMS_MFA` challenge if user has sms MFA enabled in `admin_initiate_auth`
- Support `SMS_MFA` type in `respond_to_auth_challange`


bblommers (approved):
LGTM - thank you for adding this to Moto @AriFreyr!

PR #7380, opened 2024-02-21
DynamoDB: fix: support top level lists for Dynamo Item to_json

verysamuel:
The tests in this example (prior to the fix) reproduce an issue where the exception `TypeError: Object of type DynamoType is not JSON serializable` is raised when `Dynamo.Item.to_json` is passed a top level list.

I have chosen to test this via a condition expression with `ReturnValuesOnConditionCheckFailure="ALL_OLD"` as this is a very close reproduction to the issue I faced.
Some additional tests for top level string sets and lists in dictionaries were added to ensure different paths through the new and existing code are handled correctly.

Also fixes issue with `UPDATED_NEW` handling for lists:
- Before this change: when `update` is called to change one value in a list, `_build_updated_new_attributes` attempts to return the list with the value at the position of the changed element as an empty dictionary (which leads to a deserialization error)
- With this change: when `update` is called to change one value in a list, `_build_updated_new_attributes` returns the whole changed list if any element in the original list differs from the changed list - otherwise the whole original list is returned.

### DIFF
--- a/moto/dynamodb/models/dynamo_type.py
+++ b/moto/dynamodb/models/dynamo_type.py
@@ -324,6 +324,12 @@ class Item(BaseModel):
                     for key, value in attribute.value.items()
                 }
                 attributes[attribute_key] = {attribute.type: attr_value}
+            elif isinstance(attribute.value, list):
+                attr_value = [
+                    value.to_regular_json() if isinstance(value, DynamoType) else value
+                    for value in attribute.value
+                ]
+                attributes[attribute_key] = {attribute.type: attr_value}
             else:
                 attributes[attribute_key] = {attribute.type: attribute.value}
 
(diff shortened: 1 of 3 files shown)


verysamuel:
@bblommers Having a difficult time understanding why [the test called `test_dynamodb/test_dynamodb.py::test_update_return_updated_new_attributes_when_same` failed](https://github.com/getmoto/moto/actions/runs/8001300362/job/21855180785#step:9:2065) when it seems unrelated to this PR.

That being said, it seems to point to an existing issue in how lists are handled with `UPDATED_NEW`. As such, I've included a fix for it in this PR. I've updated the description to reflect that.

bblommers (approved):
This looks great @verysamuel - always good to see such a well-tested change.

Thank you for fixing this!

PR #7392, opened 2024-02-25
RDS: Allow copying snapshots of deleted DB instances

oranav:
Fix a regression introduced in #6444, in which copying a snapshot of an already deleted DB instance results in raising a `DBInstanceNotFoundError`.
The expected behavior is that copying a snapshot is supported regardless whether its source database exists or not.

Change `create_db_snapshot` to accept a `Database` object (instead of an identifier), and use this flow in the `copy_db_snapshot` implementation. This works because the actual `Database` object is not deleted (as it is referred to by the snapshot), but rather it is already removed from the `databases` dictionary.

### DIFF
--- a/moto/rds/models.py
+++ b/moto/rds/models.py
@@ -1782,14 +1782,18 @@ class RDSBackend(BaseBackend):
 
     def create_db_snapshot(
         self,
-        db_instance_identifier: str,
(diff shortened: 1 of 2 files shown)


bblommers (approved):
Makes sense @oranav - thank you for fixing this!

PR #7402, opened 2024-02-28
ECR: Add get registry scanning configuration operation

dfangl:
## Motivation
Currently, moto supports the `put_registry_scanning_configuration` operation, however the return value does not match the specification (https://docs.aws.amazon.com/AmazonECR/latest/APIReference/API_PutRegistryScanningConfiguration.html).
The `get_registry_scanning_configuration` operation is currently not supported.

Also, the matching of repository filters against the names is not correct (wildcards not properly handled, check for equality instead of "contains", as described here: https://docs.aws.amazon.com/AmazonECR/latest/userguide/image-scanning.html#image-scanning-filters)

## Changes
* Add `get_registry_scanning_configuration` operation
* Fix return values by put operation
* Add test for the new functionality
* Apply scanning configuration to new repositories as well
* Reset scanning configuration of repositories if a new one is put


bblommers (changes requested):
Hi @dfangl! The code changes themselves LGTM - looks like it's just MyPy that's currently unhappy:
https://github.com/getmoto/moto/actions/runs/8082866988/job/22084668361?pr=7402

dfangl:
Yes, did not come around yet to fix it, will fix it in the morning! I would also like to lock the modification of newly created repositories to be safe, I thought it might not be necessary, but on second thought there are ways to get inconsistent state without!

Thanks for checking it out already!

tests/test_ecr/test_ecr_scanning_config.py
```diff
@@ -133,9 +137,120 @@ def test_put_registry_scanning_configuration():
             "repositoryArn": repo_arn,
             "repositoryName": repo_name,
             "scanOnPush": False,
-            "scanFrequency": "SCAN_ON_PUSH",
-            "appliedScanFilters": [
-                {"filter": f"{repo_name[:4]}*", "filterType": "WILDCARD"}
-            ],
+            "scanFrequency": "MANUAL",
+            "appliedScanFilters": [],
```
dfangl, line 141: Manually tested it against AWS. If a new scanning policy is pushed which does not match the existing repositories, they are overwritten with the default again.

bblommers (approved):
LGTM - thank you for adding this to Moto @dfangl!

PR #7403, opened 2024-02-28
SQS: message send param validation

rafcio19:
Closes https://github.com/getmoto/moto/issues/7398

### DIFF
--- a/moto/sqs/models.py
+++ b/moto/sqs/models.py
@@ -815,7 +815,7 @@ class SQSBackend(BaseBackend):
                 )
                 raise InvalidParameterValue(msg)
 
-            if delay_seconds > 0:  # type: ignore[operator]
+            if delay_seconds > 0:
                 raise InvalidParameterValue(
                     f"Value {delay_seconds} for parameter DelaySeconds is invalid. Reason: "
                     "The request include parameter that is not valid for this queue type."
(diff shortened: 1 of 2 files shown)


rafcio19:
@bblommers the tests are now taking ~15mins, is this version 5 improvements?? I'm pretty sure they used to take 3 times as much 👏 👏 👏 

On the PR - could you rerun the failing tests? 3.8 defo looks like CI issue, perhaps 3.12 is too

bblommers:
Hi @rafcio19! Yes, those are all v5 improvements. Now that we have the new [`config`-attribute](docs.getmoto.org/en/latest/docs/configuration/index.html) the IAM managed policies are no longer loaded by default, and it's now also possible to configure the boto3 session to not reset. Those two made all the difference in performance, so I'm quite happy with that!

About the PR: AWS does allow you to configure a fifo-queue with the DelaySeconds-attribute - but I think the current implementation would error when sending a message against a fifo-queue with that attribute.

Would be good to add a test to Moto with that specific configuration, to ensure it works:
```
sqs.create_queue(QueueName="test3.fifo", Attributes={"FifoQueue": "true", "DelaySeconds": "10"})
sqs.send_message(
    QueueUrl="test3.fifo",
    MessageBody="test",
    DelaySeconds=0,
    MessageGroupId="test",
    MessageDeduplicationId=str(uuid4()),
)
# message can be received as usual...
```

bblommers (approved):
Awesome - thank you so much @rafcio19!

PR #7406, opened 2024-02-28
Feature: Route53Domains

NeevCohen:
Adding support for Route53Domains endpoints

This PR is a work in progress


moto/route53domains/validators.py
```diff
@@ -0,0 +1,1044 @@
...
+    "world",
+    "wtf",
+    "xyz",
+    "zone",
+)
+
+VALID_DOMAIN_REGEX = re.compile(
+    r"(?:[a-z0-9](?:[a-z0-9-]{0,61}[a-z0-9])?\.)+[a-z0-9][a-z0-9-]{0,61}[a-z0-9]"
+)
+PHONE_NUMBER_REGEX = re.compile(r"\+\d*\.\d{10}$")
```
bblommers, line 604:
```suggestion
PHONE_NUMBER_REGEX = re.compile(r"\+\d+\.\d+$")
```
Minor niggle - but phone numbers are notoriously hard to validate, that's why it stood out to  me.

The exact length of a phone number differs by country, so I think we should be very liberal in what we accept

tests/test_route53domains/test_route53domains_domain.py
```diff
@@ -0,0 +1,214 @@
...
+):
+    """Test good register domain API calls."""
+    route53domains_client = boto3.client("route53domains", region_name="global")
+    route53_client = boto3.client("route53", region_name="global")
+    route53domains_client.register_domain(**register_domain_parameters)
+
+    res = route53_client.list_hosted_zones()
+    for zone in res["HostedZones"]:
+        if zone["Name"] == "domain.com":
+            return
```
bblommers, line 95: Pytest doesn't like having `return`-statements inside a test, so this should probably be rewritten to something like:
```
assert "domain.com" in [zone["Name"] for zone in res["HostedZones"]]
```

moto/route53domains/validators.py
```diff
@@ -0,0 +1,1044 @@
+import re
+from datetime import datetime, timedelta, timezone
+from ipaddress import ip_address
+from typing import Dict, List
+
+from moto.core.common_models import BaseModel
+from moto.moto_api._internal import MotoRandom
```
bblommers, line 7:
```suggestion
from moto.moto_api._internal import mock_random as random
```
That allows you to use the regular methods further down (`random.uuid4()`), makes it a bit easier to read.

moto/route53domains/models.py
```diff
@@ -0,0 +1,161 @@
...
+        for operation in self.__operations.values():
+            if (
+                operation.domain_name == requested_operation.domain_name
+                and operation.type == requested_operation.type
+            ):
+                raise DuplicateRequestException()
+
+    def list_operations(
+        self,
+        submitted_since_timestamp: int | None = None,
```
bblommers, line 93: Moto still supports Python 3.8, which doesn't support the |-operand. So they should all be converted:
`Optional[int] = None`

bblommers (commented):
Hi @NeevCohen! The overall structure looks good, that's exactly how all other services look like as well.

I've commented on a few things that stood out to me, but nothing major so far. :+1:

moto/route53domains/models.py
NeevCohen: Fair enough, although I really like the new `|` operator

NeevCohen:
@bblommers Thanks for the quick feedback! I've changed the things you commented about. 
I'll keep working on this and keep you updated if I run into any problems

NeevCohen:
Hey @bblommers, I've implemented some more endpoints - `list_domains`, `delete_domain` and `get_operation_detail`.
Want to go over the changes?

moto/route53domains/models.py
```diff
@@ -0,0 +1,309 @@
...
+                    key=lambda d: d.domain_name,
+                    reverse=(sort.sort_order == DomainSortOrder.descending),
+                )
+            else:
+                domains_to_return.sort(
+                    key=lambda d: d.expiration_date,
+                    reverse=(sort.sort_order == DomainSortOrder.descending),
+                )
+
+        start_idx = 0 if marker is None else int(marker)
```
bblommers, line 187: The pagination can probably be handled by the `@paginate`-decorator. That would simplify the logic, and give you a lot of validation out of the box.
See the documentation here: http://docs.getmoto.org/en/latest/docs/contributing/development_tips/utilities.html#paginator

I think that the docs has all the information you need, but let me know if run into any problems.

moto/route53domains/responses.py
```diff
@@ -0,0 +1,143 @@
...
+        super().__init__(service_name="route53-domains")
+
+    @property
+    def route53domains_backend(self) -> Route53DomainsBackend:
+        """Return backend instance"""
+        return route53domains_backends[self.current_account]["global"]
+
+    def register_domain(self) -> str:
+        """Register a domain"""
+        domain_name: Optional[str] = self._get_param("DomainName")
```
bblommers, line 20:
```suggestion
        domain_name = self._get_param("DomainName")
```
The `_get_param`-operations returns `Any`, if I'm not mistaken. So if the only goal is to make `mypy` happy you could remove all the explicit types.

tests/test_route53domains/test_route53domains_domain.py
```diff
@@ -0,0 +1,581 @@
...
+            "Fax": "+1.1234567890",
+        },
+        "PrivacyProtectAdminContact": True,
+        "PrivacyProtectRegistrantContact": True,
+        "PrivacyProtectTechContact": True,
+    }
+
+
+@pytest.fixture(name="invalid_domain_parameters")
+def generate_invalid_domain_parameters(domain_parameters: Dict) -> Dict:
```
bblommers, line 60: We don't validate typing in the majority of the tests, so they are not strictly necessary. (If you feel that it helps you with development, then feel free to keep it of course.)

bblommers (changes requested):

moto/route53domains/models.py
NeevCohen: Awesome! I changed the functions as per your suggestion and it worked. I also removed the test for testing the pagination logic since it seems out of the scope of the test module.

I really liked the pagination utility, really neat solution!

NeevCohen: Awesome! I changed the functions as per your suggestion and it worked. I also removed the test for testing the pagination logic since it seems out of the scope of the test module.

I really liked the pagination utility, really neat solution!

moto/route53domains/responses.py
NeevCohen: Honestly it was really just for documentation's sake, so that anyone who reads the code in the future would know that type they should be. Anyway, I removed these type hints since they upset mypy as well

tests/test_route53domains/test_route53domains_domain.py
NeevCohen: Yeah I kinda like it when these are type-hinted

NeevCohen:
@bblommers I've updated the code as per your suggestions, I think it's pretty complete in terms of features I wanted to implement, at least for now. Should I mark this as `Ready for review`?

bblommers (approved):
LGTM - thank you for adding this to Moto @NeevCohen!

PR #7407, opened 2024-02-29
S3: Adding tests of EventBridge notification `ObjectCreated:Copy`

tsugumi-sys:
- Adding a test of EventBridge notification for `ObjectCreated:Copy`.
- Some helper methods to set up AWS resources and retrieve sent event logs were added for reuse.
- `disable_notification` argument were added for `put_object` method to handle notification.


bblommers (approved):
LGTM! Thanks again @tsugumi-sys

PR #7418, opened 2024-03-03
StepFunctions: Proper Parser

### LINKED ISSUE #3076: Step functions get_execution_history
Hi, This is not a bug. I just saw on the service coverage page, that step function boto3 client wasn't completely covered. I couldn't find an open issue for it, but I am curious if there is a timeline for boto3.client("stepfunctions").get_execution_history() or if that has been mentioned before? I use this method for reporting, and it would be a lot easier to mock through moto.

### LINKED ISSUE #3544: Full Step Functions Execution Support
I didn't see this in the issues, so my apologies if it's already filed. 

This is a feature request for full Step Function execution support, such that test code could call sfn.StartExecution and get a production-like execution of a state machine, including retries, error handling, concurrency, choice evaluation, etc, and actually runs application code.
Something like StepFunctionsLocal from AWS, which is in java: https://docs.aws.amazon.com/step-functions/latest/dg/sfn-local.html

I appreciate that implementing all of step functions is a daunting task! I will describe a subset that I think would be useful to a large set of developers. 

My ideal implementation would support State Language and lambda task integration. 

I suppose of course it could it integrate with moto.lambda to run lambda functions, but I think some kind of configuration interface/escape hatch enabling test writers to implement their own implementation of any task type would keep it flexible enough for users to do some minimal work to keep up with new step functions features. 

In my case, I would hook the lambda tasks directly to my own worker code and skip the lambda worker model that moto has, as I don't use this

bblommers:
This parser was originally created for LocalStack, but was easily adopted for Moto's use.

TODO:

 - Tests for other possible State-types (catch, retry, map, etc: see https://github.com/localstack/localstack/tree/master/tests/aws/services/stepfunctions/templates/scenarios/statemachines)
 - Tests for other service integrations (see https://github.com/localstack/localstack/tree/master/tests/aws/services/stepfunctions/templates/services/statemachines)
 - Typing?
 - Decide whether we keep the raw (G4) ANTLR files. They are in LocalStack and also in https://github.com/antlr/grammars-v4/pull/3981

Fixes #3076 
Fixes #3544


moto/settings.py
```diff
@@ -114,7 +115,12 @@ def moto_server_host() -> str:
     if is_docker():
         return get_docker_host()
     else:
-        return "http://host.docker.internal"
+        try:
+            localhost = "http://localhost"
+            assert requests.get(f"{localhost}:{moto_server_port()}").status_code == 200
```
bblommers, line 120: This should have a timeout

PR #7420, opened 2024-03-04
S3: Add EventBridge Notification for DeleteObject

MikiPWata:
PR for DeleteObject in #7363 
 - send EventBridge notification when Object is Deleted(for buckets that are versioned and not versioned)
 - added a test case for sending eventbridge notification


bblommers (approved):
LGTM - thank you for adding this @MikiPWata!

Issue #7412, opened 2024-03-01
Cloudfront: update_distribution issue


izipris:
Hi folks,

I'd like to start with saying that Moto is a great library and it helps me a lot, thanks for your work!

Recently I've started working with the Cloudfront service, and I noticed some unexpected behavior of Moto while using the `update_distribution` method - the configuration of the distribution is actually not changed. I wrote some test to illustrate it:
```
@mock_aws
def test_update_distribution():
    cloudfront_service.create_distribution(DistributionConfig=distribution_config)
    distributions = cloudfront_service.list_distributions()['DistributionList']['Items']
    assert len(distributions) == 1
    actual_distribution = distributions[0]
    assert actual_distribution['Enabled']
    get_config_response = cloudfront_service.get_distribution_config(Id=actual_distribution['Id'])
    actual_distribution_config = get_config_response['DistributionConfig']
    etag = get_config_response['ETag']
    actual_distribution_config['Enabled'] = False  # Setup: change the 'Enabled' field to False
    cloudfront_service.update_distribution(DistributionConfig=actual_distribution_config, Id=actual_distribution['Id'],
                                           IfMatch=etag)  # Execution: set the new configuration
    distributions = cloudfront_service.list_distributions()['DistributionList']['Items']
    assert len(distributions) == 1
    actual_distribution = distributions[0]
    assert not actual_distribution['Enabled']  # Verification: make sure 'Enabled' is false FIXME: this assertion fails
```
I was digging a little bit and found in [moto/cloudfront/models.py#L361](https://github.com/getmoto/moto/blob/master/moto/cloudfront/models.py#L361) that when updating a distribution, the new configuration is stored in the `config` variable, however the new configuration **is not** applied, since applying the values of a configuration is happening only in [the constructor of DistributionConfig](https://github.com/getmoto/moto/blob/master/moto/cloudfront/models.py#L148).

Does my thinking makes sense? Would you like to get a PR for this?

bblommers:
Hi @izipris! That does make sense - we definitely don't process the configuration correctly there.

A PR would be very welcome!

izipris:
Hi @bblommers , I just added a PR. I followed [this guide](https://docs.getmoto.org/en/latest/docs/contributing/checklist.html) - verified that `pylint` is happy and all tests passed. Please let me know if I missed anything.
Thanks!

PR #7423, opened 2024-03-04
Cloudfront update_distribution - apply configuration changes

izipris:
Fix [issue #7412](https://github.com/getmoto/moto/issues/7412) - apply distribution configuration changes when using Cloudfront `update_distribution`


bblommers (approved):
Makes sense @izipris - thanks for the fix!

PR #7424, opened 2024-03-05
EC2: SG rule tagging

rafcio19:
Closes https://github.com/getmoto/moto/issues/7408

### DIFF
--- a/moto/ec2/responses/security_groups.py
+++ b/moto/ec2/responses/security_groups.py
@@ -106,17 +106,17 @@ class SecurityGroups(EC2BaseResponse):
                 prefix_list_ids,
             ) = parse_sg_attributes_from_dict(querytree)
 
-            yield (
-                group_name_or_id,
-                ip_protocol,
-                from_port,
-                to_port,
-                ip_ranges,
-                sg_rule_tags,
-                source_groups,
-                prefix_list_ids,
-                security_rule_ids,
-            )
+            yield {
+                "group_name_or_id" :group_name_or_id,
+                "ip_protocol": ip_protocol,
+                "from_port": from_port,
+                "to_port": to_port,
+                "ip_ranges": ip_ranges,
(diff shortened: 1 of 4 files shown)


moto/ec2/responses/security_groups.py
```diff
@@ -86,6 +86,15 @@ def _process_rules_from_querystring(self) -> Any:
                 d = d[subkey]
             d[key_splitted[-1]] = value
 
+        sg_rule_tags = {}
```
bblommers, line 89: I think we should process the tag-retrieval inside the `authorize_..`-method itself. It is a toplevel attribute, and arguably unrelated to the other SecurityGroupRule-properties that we're parsing.

Ideally we can just use 
`tags = self._parse_tag_specification().get("security-group-rule", {})`,
similar to how it's done in `create_security_group`. That way we don't have to have any complicated parsing here

moto/ec2/responses/security_groups.py
```diff
@@ -121,22 +131,23 @@ def _process_rules_from_querystring(self) -> Any:
...
+                "prefix_list_ids": prefix_list_ids,
+                "security_rule_ids": security_rule_ids,
+            }
 
     def authorize_security_group_egress(self) -> str:
         self.error_on_dryrun()
 
         for args in self._process_rules_from_querystring():
-            rule, group = self.ec2_backend.authorize_security_group_egress(*args)
+            rule, group = self.ec2_backend.authorize_security_group_egress(**args)
```
bblommers, line 150:
```suggestion
            rule, group = self.ec2_backend.authorize_security_group_egress(**kwargs)
```
I like the change to make them keyword-arguments! Makes it easier to debug. Just a minor nitpick that the name should reflect this.

bblommers (changes requested):

bblommers (approved):
Nice one - thanks for fixing this @rafcio19!

PR #7429, opened 2024-03-06
fix: add dynamodb type mismatch validator

louislung:
when doing dynamodb `update_item`, it should check for type mismatch 
e.g. for attribute defined as `String`, it should throws an error when `NULL`, or other data type e.g. `Number`, `Binary` is received


bblommers (approved):
LGTM. Thank you for fixing this @louislung, and welcome to Moto!

PR #7430, opened 2024-03-06
added resourcegroupstaggingapi support for ssm

vgiridhar97:


tests/test_resourcegroupstaggingapi/test_resourcegroupstaggingapi.py
```diff
@@ -759,6 +759,32 @@ def test_get_resources_sns():
         "Tags"
     ]
 
+@mock_aws
+def test_get_resources_ssm():
+    import json
+    import pkgutil
+    import yaml
+
+    template_file = pkgutil.get_data(__name__, "test_templates/good.yaml")
```
bblommers, line 768: Can we reuse this from the SSM tests?

I think we can just do a toplevel import:
`from tests.test_ssm/test_ssm_docs import _get_yaml_template`

That way we don't have to duplicate anything

moto/resourcegroupstaggingapi/models.py
```diff
@@ -509,6 +513,23 @@ def format_tag_keys(
...
+        # SSM 
+        if not resource_type_filters or "ssm" in resource_type_filters:
+            for document in self.ssm_backend._documents.values():
+                doc_name = document.describe()["Name"]
+                tags = self.ssm_backend._get_documents_tags(doc_name)
+                if not tags or not tag_filter(
+                    tags
+                ): # Skip if no tags, or invalid filter
+                    continue
+                # pytest.set_trace()
```
bblommers, line 526: This should be removed

moto/ssm/models.py
```diff
@@ -1286,6 +1287,7 @@ def create_document(
         self._documents[ssm_document.name] = documents
 
         if tags:
+            # pytest.set_trace()
```
bblommers, line 1290: Same here (and the `pytest`-import)

bblommers (changes requested):
Hi @vgiridhar97! Just a few minor points - the change itself looks good to me.

moto/resourcegroupstaggingapi/models.py
vgiridhar97: hi, i've removed this comment in my latest commit

moto/ssm/models.py
vgiridhar97: hi, i've removed this comment in my latest commit

tests/test_resourcegroupstaggingapi/test_resourcegroupstaggingapi.py
vgiridhar97: hi, i've changed it to the toplevel import in my latest commit

bblommers (changes requested):
Nitpick - but the additional yaml-file is no longer necessary now.

Looks like the linter is failing as well: https://github.com/getmoto/moto/actions/runs/8254430099/job/22578588665

moto/ssm/models.py
```diff
@@ -8,6 +8,7 @@
 from typing import Any, DefaultDict, Dict, Iterator, List, Optional, Tuple
 
 import yaml
+import pytest
```
bblommers, line 11:
```suggestion
```

bblommers (approved):
LGTM - thank you for adding this @vgiridhar97!

PR #7435, opened 2024-03-07
S3: Sending notification to EventBridge `OBJECT_CREATED_COMPLETE_MULTIPART_UPLOAD_EVENT`

tsugumi-sys:
- Sending a notification to EventBridge when multipart upload is completed.
- Unittest is added.

related #7363

### DIFF
--- a/moto/s3/models.py
+++ b/moto/s3/models.py
@@ -2606,10 +2606,17 @@ class S3Backend(BaseBackend, CloudWatchMetricProvider):
         bucket = self.get_bucket(bucket_name)
 
         response_meta = {}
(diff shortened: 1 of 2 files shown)


bblommers (changes requested):
Hey @tsugumi-sys!

The current implementation tries to send two notifications - one for CompleteMultiupload, and one for PutObject.

I've pushed a change to the `complete_multipart`-method so that all the logic is in one place - if you could pull in the latest changes from master, you'll see that it will make controlling the notifications a bit easier.

bblommers (approved):

PR #7441, opened 2024-03-08
Docs: refer to client, not client2, in sample

douglasnaphas:
Later examples have a `client1` and a `client2`. This one only has `client`.


bblommers (approved):
Good catch @douglasnaphas - thank you for fixing this!

PR #7445, opened 2024-03-08
Fix requests mock for custom S3 endpoints

dhirving:
Prior to the commit that introduced mock_aws (a7f3b367b), it was possible to use the requests mock to access S3 presigned URLs that had a custom endpoint.

This commit restores that behavior by applying the hostname-remapping logic from BotocoreStubber to CustomRegistry.

### DIFF
--- a/moto/core/responses_custom_registry.py
+++ b/moto/core/responses_custom_registry.py
@@ -45,11 +45,15 @@ class CustomRegistry(responses.registries.FirstMatchRegistry):
         match_failed_reasons = []
 
         # Handle non-standard AWS endpoint hostnames from ISO regions or custom S3 endpoints.
-        url_with_standard_aws_domain = urlunparse(
-            get_equivalent_url_in_aws_domain(request.url)
-        )
-        request_with_standard_aws_domain = request.copy()
-        request_with_standard_aws_domain.prepare_url(url_with_standard_aws_domain, {})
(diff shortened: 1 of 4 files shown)


moto/core/responses_custom_registry.py
```diff
@@ -41,8 +43,16 @@ def find(self, request: Any) -> Tuple[Optional[responses.BaseResponse], List[str
         )
         found = []
         match_failed_reasons = []
+
+        # Handle non-standard AWS endpoint hostnames from ISO regions or custom S3 endpoints.
+        url_with_standard_aws_domain = urlunparse(
+            get_equivalent_url_in_aws_domain(request.url)
+        )
+        request_with_standard_aws_domain = request.copy()
```
bblommers, line 51: Copying and preparing the URL is only necessary in some cases, but it currently happens for every incoming HTTP request. The possible performance downsides of that worry me a little.

Can we make the copy+prepare conditional, and only execute that part if the standardized URL is different from the original?

dhirving: Yeah that's a good call.  There's no guarantee that URLs will roundtrip unchanged through urlparse/urlunparse, so I put the modification check a little further up.  That will also save some string copying.

bblommers (approved):
Awesome - thank you for fixing this @dhirving!

PR #7448, opened 2024-03-09
S3: Sending notification to EventBridge `LIFECYCLE_TRANSITION_EVENT`

tsugumi-sys:
- Sending an event to EventBridge when the object storage class is changed.
- (Object storage class is also changed by bucket lifecycle, but I think we cannot detect it in the current moto implementation.)

### DIFF
--- a/moto/s3/models.py
+++ b/moto/s3/models.py
@@ -2783,6 +2783,16 @@ class S3Backend(BaseBackend, CloudWatchMetricProvider):
             new_key,
         )
 
+        if src_key.storage_class != new_key.storage_class:
+            notifications.send_event(
+                self.account_id,
+                notifications.S3NotificationEvent.LIFECYCLE_TRANSITION_EVENT,
+                bucket,
+                new_key,
+            )
+
+        # Send notifications if object storage class is changed.
+
     def put_bucket_acl(self, bucket_name: str, acl: Optional[FakeAcl]) -> None:
         bucket = self.get_bucket(bucket_name)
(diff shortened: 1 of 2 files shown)


bblommers (commented):
This sounds like a good usecase to introduce a getter/setter method for the `storage_class` attribute.

If we send the event inside the setter method, we can be sure that this always happens - regardless of who/where/why the storage class was changed.
We'll just have to be careful to only send an event when there is a change, and not on initialization.

What do you think?

tsugumi-sys:
> This sounds like a good usecase to introduce a getter/setter method for the storage_class attribute.

I totally agree with this :)
Thanks @bblommers ! I'll implement a setter and getter for storage class attribute in this PR.

tsugumi-sys:
we need this implementation: https://github.com/getmoto/moto/pull/7455

bblommers:
The `status`-attribute should already be there @tsugumi-sys - it's defined in the parent class. 
My PR just overrides it and adds additional functionality.

bblommers:
I've done some testing against AWS, and changing the object storage class using the `copy_object` method does **not** lead to a `LIFECYCLE_TRANSITION_EVENT` - it only sends a `COPY_EVENT`. 

This PR contains a CDK stack if you want to test this out for yourself: https://github.com/getmoto/moto/pull/9378

For this to be supported, we would first need to support actual lifecycle transisions, and find a way to mock those. But this proposed implementation would not work,  unfortunately, so I'll close this PR.

PR #7449, opened 2024-03-09
VPCEndpoint: add state filtering

SoenkeD:
AWS EC2 endpoint support filtering by the endpoints state ( https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_DescribeVpcEndpoints.html ). It seems, that only a mapping of the objects value to the filter was missing.


bblommers (changes requested):
Hi @SoenkeD, would you mind adding a test for this?

There is an existing test for the other filter, in `tests/test_ec2/test_vpcs.py::test_describe_vpc_gateway_end_points`, so we can just extend that one.

PR #7455, opened 2024-03-10
S3: Add Notifications for Restore-actions

bblommers:
See @7363

@tsugumi-sys The `restore_object`-function was implemented in S3, it was just hidden in `responses.py`.
This PR:
 - extracts the logic into the models-class
 - Adds a notification when a Restore-action is initiated and completed

Any feedback/improvements that you can see?


tsugumi-sys:
Thanks, @bblommers!
Oh, I missed `restore_object` implementation in moto!
LGTM :)

PR #7464, opened 2024-03-12
Cognito: update UserLastModifiedDate when changing user attributes

schmittjoaopedro:
The current implementation is not updating the Cognito IDP User field UserLastModifiedDate when updating or removing attributes. So this change proposes to fix this.


bblommers (approved):
LGTM - thanks for the fix @schmittjoaopedro!

PR #7468, opened 2024-03-14
S3: `Days` must not be provided for select request when restoring object

tsugumi-sys:
- Raise error when both `Days` and `Type` parameters exist in restore object request.

image
https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3/client/restore_object.html

related: https://github.com/getmoto/moto/issues/7467

### DIFF
--- a/moto/s3/responses.py
+++ b/moto/s3/responses.py
@@ -2272,10 +2272,19 @@ class S3Response(BaseResponse):
             )
 
         elif "restore" in query:
-            es = minidom.parseString(body).getElementsByTagName("Days")
-            days = es[0].childNodes[0].wholeText
+            params = minidom.parseString(body)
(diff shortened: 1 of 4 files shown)


moto/s3/responses.py
```diff
@@ -2272,10 +2272,19 @@ def _key_response_post(
             )
 
         elif "restore" in query:
-            es = minidom.parseString(body).getElementsByTagName("Days")
-            days = es[0].childNodes[0].wholeText
+            params = minidom.parseString(body)
```
bblommers, line 2275: Is there any value in using `xmltodict` here? We use it somewhere else in this file as well.

I don't know what the output for that would look like, but maybe it's easier than parsing the XML manually.

bblommers (approved):
Nice one - thank you!

PR #7477, opened 2024-03-15
adding services to IMPLEMENTED_ENDPOINT_SERVICES, to be returned by describe_vpc_endpoint_services calls

jamesnixon-aws:
Addresses: https://github.com/getmoto/moto/issues/7469

Services retrieved via a real call to describe-vpc-endpoint-services.


jamesnixon-aws:
Not sure what the test failures are caused by, but they seem unrelated to the changes. There are nested import errors for `moto.accessanalyzer`. Any ideas on the cause are welcome.

bblommers:
Hi @jamesnixon-aws! The logic behind the implementation is basically:
```
for every_service in IMPLEMENTED_SERVICES:
    lazyload the service module (implementation for that service)
    request the vpc_endpoint details from that module
```
The errors for `moto.accessanalyzer` appear because Moto doesn't have a service module for most items from the list - so the call to `importlib.import_module('moto.accessanalyzer')` is bound to fail.

I've raised a different PR with a fix: #7509. 
That implementation returns a sensible default config for services that Moto doesn't know about, and also fixes a bug where Moto would only cache service configs for a single region.

Thank you for getting started with this, and for providing the list of supported services!

PR #7478, opened 2024-03-15
Sets autoscalinggroup CreatedTime during creation.

jamaalscarlett:
Use the current time instead of hard coded '2013-05-06T17:47:15.107Z' when creating an autoscaling group.
Updated template to return the correct created time.
Added unit test.


bblommers (approved):
LGTM - thanks for the fix @jamaalscarlett!

jamaalscarlett:
> LGTM - thanks for the fix @jamaalscarlett!
My pleasure :)

PR #7479, opened 2024-03-16
Server: check ready state with `threading.Event`

serozhenka:
This PR aims to replace the busy waiting of the `ThreadedMotoServer` startup with the more clear `threading.Event` object. This change will also reduce the `ThreadedMotoServer.start` method execution time, as the Werkzeug server startup time is usually less than 0.1 seconds.


bblommers (approved):
Makes sense - that is indeed much cleaner.

Thank you for the PR @serozhenka!

PR #7489, opened 2024-03-19
CloudFormation: Return Description field in Outputs

Wurstnase:


moto/cloudformation/responses.py
```diff
@@ -818,6 +818,7 @@ def set_stack_policy(self) -> str:
           <member>
             <OutputKey>{{ output.key }}</OutputKey>
             <OutputValue>{{ output.value }}</OutputValue>
+            <Description>{{ output.description }}</Description>
```
bblommers, line 821:
```suggestion
            {% if output.description %}<Description>{{ output.description }}</Description>{% endif %}
```
Nitpick, but AWS only sends this tag if there is an actual value.

Looks like there's some existing tests failing as well.

bblommers (changes requested):

Wurstnase:
I will fix the other failing tests.

bblommers (approved):
LGTM - thank you @Wurstnase!

PR #7491, opened 2024-03-19
Lambda: throw error on empty resource policy

rafcio19:
Closes https://github.com/getmoto/moto/issues/7475


bblommers (approved):
Awesome - thanks for the fix @rafcio19!

PR #7493, opened 2024-03-19
feat: Workspaceapi 

archinksagar:
This is the initial implementation of AWS Workspace api. Can I get a feedback on this.
1. I have added 15 API endpoints. 
2. I have changed 'us-west-1' to 'us-west-2' for glue and ec2 vpc in test_resourcetaggingapi because that region is not supported by AWS workspace and I was getting errors.
3. There are few unknowns, for e.g. computer_name
4. I have added tagging support in resourcegroupstaggingapi


bblommers (approved):
This is great @archinksagar. Very thorough tests, and early integration with ResourceGroupsTagging-API is always useful to increase our feature coverage there.

Thank you for adding this to Moto!

PR #7494, opened 2024-03-19
add validations and tag filter to describe-security-group-rules operation

pinzon:
This pull request includes additional validations for the filters in the ec2 describe-security-group-rules operation. This resolves the issue identified in localstack/localstack#10315.


pinzon:
My master branch wasn't up to date so I didn't notice that the feature was already implemented in #7424

PR #7496, opened 2024-03-20
updated default value for share status in route53resolver

macnev2013:
# Changes
- Updated default value for `shared_status` in route53resolver
- Removed unused code from `Makefile` which was used for terraform tests
- Updated the test cases


bblommers (approved):
LGTM - thanks @macnev2013 !

PR #7497, opened 2024-03-20
Lambda: simple lambda custom responses

rafcio19:
Customise Lambda response by posting desired results to a new, static moto endpoint.

Resolves https://github.com/getmoto/moto/issues/7481

### DIFF
--- a/moto/awslambda_simple/models.py
+++ b/moto/awslambda_simple/models.py
@@ -24,7 +24,7 @@ class LambdaSimpleBackend(LambdaBackend):
         response_headers: Any,
     ) -> Optional[Union[str, bytes]]:
 
-        default_result = ""
+        default_result = "Simple Lambda happy path OK"
         if self.lambda_simple_results_queue:
             default_result = self.lambda_simple_results_queue.pop(0)
         return str.encode(default_result)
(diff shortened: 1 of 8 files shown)


moto/awslambda_simple/models.py
```diff
@@ -20,9 +24,10 @@ def invoke(
         response_headers: Any,
     ) -> Optional[Union[str, bytes]]:
 
-        if body:
-            return str.encode(body)
-        return b"Simple Lambda happy path OK"
+        default_result = ""
```
bblommers, line 27:
```suggestion
        default_result = str.encode(body) if body else b"Simple Lambda happy path OK"
```

No reason to change that for existing  users - I'm sure someone's tests depend on that specific output :slightly_smiling_face:

bblommers: I agree that an empty string makes a lot more sense, btw - but it is technically a breaking change, so we should only do that as part of a major release.

rafcio19: Good point but the rationale is: this more accurately reflects what happens with real lambda in AWS, by default nothing is returned.

Happy to change this back though 🤷

bblommers (approved):
LGTM - thank you @rafcio19!

PR #7500, opened 2024-03-20
Add the Pagination Utility to the Elasticache describe_cache_clusters model + response

atanner27:
Add the Pagination Utility to the Elasticache describe_cache_clusters model + response.

Heavily based on the route52 implementation + using the existing utility function


bblommers (approved):
LGTM - thank you for adding this @atanner27!

PR #7502, opened 2024-03-21
Admin: Using `ruff format` instead of `black`

### LINKED ISSUE #7473: Adim: How about using `ruff format` instead of `black`
Moto currently uses ruff (check) and black as formatter.

```make
# Makefile
format:
	black moto/ tests/
	ruff --fix moto/ tests/
```

We can use [ruff also as a formatter](https://docs.astral.sh/ruff/formatter/) like this.

```make
# Makefile
format:
	black moto/ tests/
	ruff --fix moto/ tests/
        ruff format moto/ tests/
```

I think black can be replaced as ruff format. @bblommers, what do you think?

tsugumi-sys:
- Add `ruff format` command, instead of `black`.
- Fix ruff version in `requirements-dev.txt`.
- Remove `black` from codes.
- Reformat existing files with ruff.

closes #7473


tsugumi-sys:
@bblommers 
The coverage check fails because some codes are not covered by unittests.
I'll create other pull requests and add some unittests to fix these problems.

bblommers:
Hey @tsugumi-sys. The coverage check fails because the relative coverage is too low compared to the base commit - but that should be fixed if you rebase to the latest master. 

The current base commit is a month old, when the coverage was still at 96+%, but one week ago I merged #7418 which dropped it to 94%. If you rebase to latest master, and update the base commit to one that already has 94% coverage, it should be fine.

tsugumi-sys:
@bblommers 
I see, the commit for comparison was wrong.
Humm but I'm on the latest master branch and I cannot rebase?

image

image

bblommers:
Hm.. interesting. Maybe CodeCov just got confused, in which case we definitely shouldn't pay (too much) attention  to it :slightly_smiling_face: 

I did just (< 2 minutes ago)  merge another large PR which introduced a merge conflict here - sorry about that.

tsugumi-sys:
> Hm.. interesting. Maybe CodeCov just got confused, in which case we definitely shouldn't pay (too much) attention to it 🙂
I agree :)
I'll fix the conflict now

bblommers (approved):
Awesome - thank you @tsugumi-sys!

PR #7504, opened 2024-03-21
S3: Implement restore object with select job type

tsugumi-sys:
- Implement select job for restore object request.
- Adding validation for request parameters.
- Adding unittest for the parameter validation and select job.

boto3 documantation: https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3/client/restore_object.html

I'm not sure we should use `put_object` or deep copy when create an object to restore. What do you think @?


bblommers (changes requested):
@tsugumi-sys I think using `put_object` makes most sense here, as opposed to `deepcopy`.

Do you know what the default filename is that AWS uses? If I read the docs correctly, it sounds like they create a new file with some sort of standardized name, maybe something like `{optional_prefix_}output_of_restore_request_{uuid}`. But it would be nice to validate that.

I'd be happy to test this in AWS as well at some point and try it out, just to validate the theory - but that may take me some time.

tsugumi-sys:
@bblommers 

> I'd be happy to test this in AWS as well at some point and try it out, just to validate the theory - but that may take me some time.

I can check this :)

bblommers:
@tsugumi-sys AWS confirmed this is currently broken, but do not have a timeline to fix this. Reading between the lines, it sounds like they don't intend to fix this at all. So I don't think we should add this to Moto either.

It's unfortunate, because it is a really cool feature!

tsugumi-sys:
@bblommers 
Ah, I see. Thank you for letting me know :)

PR #7506, opened 2024-03-21
Fix: ruff check error in `test_resourcegroupstaggingapi.py

tsugumi-sys:
- Fix the following error by ruff: 
- black format is applied.
```
tests/test_resourcegroupstaggingapi/test_resourcegroupstaggingapi.py:851:9: F841 Local variable `image` is assigned to but never used.
```


bblommers (approved):
LGTM!

PR #7510, opened 2024-03-22
Adim: Add format check by ruff in Github Actions

tsugumi-sys:
- Add format checking by ruff in Github Action workflow.

image
https://docs.astral.sh/ruff/formatter/


bblommers (approved):
:+1:

PR #7511, opened 2024-03-22
fixed secretsmanager list api to support filtering correctly

macnev2013:
# Motivation
This PR addresses the issue in the list operation of Secrets Manager where filters were incorrectly applied to the entire key rather than just the prefix.

# Issue
Previously, filters were matched against the entire key, which is incorrect. The correct behavior is to match filters against the prefix only.

- Reference: [AWS Boto3 Secrets Manager list_secrets documentation](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/secretsmanager/client/list_secrets.html#list-secrets)
- This issue was initially reported in [LocalStack issue #4820](https://github.com/localstack/localstack/issues/4820).
- AWS validated test case implemented [here](https://github.com/localstack/localstack/pull/10520).
- More details on the list api: https://docs.aws.amazon.com/secretsmanager/latest/userguide/manage_search-secret.html

### DIFF
--- a/moto/ec2/models/__init__.py
+++ b/moto/ec2/models/__init__.py
@@ -1,4 +1,4 @@
-from typing import Any, Dict, List
+from typing import List
 
 from moto.core.base_backend import BackendDict, BaseBackend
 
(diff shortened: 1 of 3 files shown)


macnev2013:
Hi @bblommers, can you retrigger the pipeline? It seems the CI is failing due to unrelated changes.

bblommers (approved):
LGTM - thanks for fixing this @macnev2013!

PR #7512, opened 2024-03-23
Admin: Fix remaining ruff and pylint error

tsugumi-sys:
- Fix remaining ruff formatter error
- Fix remaining pylint warnings (redefined-outer-name, redefined-outer-name)
- Fix github action ci command to run `make lint`

(`ruff format` caused a line break and broke the type check. )

including this PR #7513

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -68,7 +68,7 @@ jobs:
       run: |
         pip install -r requirements-dev.txt
     - name: Lint
-      run:
+      run: |
         mkdir .mypy_cache
         make lint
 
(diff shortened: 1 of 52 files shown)


bblommers:
The `make lint` is currently not executed in the CI - that's why we haven't seen these errors before.

Can you add this fix to this PR? https://github.com/getmoto/moto/pull/7513/files
That ensures the linter is actually executed, and we can catch these errors earlier.

tsugumi-sys:
@bblommers 
I see :) Sure, I'll add my fix to your PR :)

bblommers (approved):
Awesome - thank you for fixing this @tsugumi-sys!

PR #7515, opened 2024-03-24
VPCEndpoint: add security group modification

SoenkeD:
Currently it is not possible to add / remove security groups for VPCEndpoints (with ModifyVpcEndpoint() ). The changes of this PR allow this and also load the SecurityGroup names when DescribeVpcEndpoint().

### DIFF
--- a/moto/ec2/responses/vpcs.py
+++ b/moto/ec2/responses/vpcs.py
@@ -224,12 +224,17 @@ class VPCs(EC2BaseResponse):
         add_route_tables = self._get_multi_param("AddRouteTableId")
         remove_route_tables = self._get_multi_param("RemoveRouteTableId")
         policy_doc = self._get_param("PolicyDocument")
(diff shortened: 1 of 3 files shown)


moto/ec2/responses/vpcs.py
```diff
@@ -251,9 +256,19 @@ def describe_vpc_endpoints(self) -> str:
         vpc_end_points = self.ec2_backend.describe_vpc_endpoints(
             vpc_end_point_ids=vpc_end_points_ids, filters=filters
         )
+
+        security_group_ids = []
+        for service in vpc_end_points:
+            security_group_ids.extend(service.security_group_ids)
```
bblommers, line 262: The failing error is because `service.security_group_ids` can be None - so you're effectively running `security_group_ids.extend(None)`, which throws the `TypeError: 'NoneType' object is not iterable` that you can see in the  logs.

bpandola:
Superseded by #9149

PR #7529, opened 2024-03-27
Cloudformation: fix CF delete for Launch template

rafcio19:


bblommers (commented):
Hey @rafcio19, does that mean we can delete `delete_from_cloudformation_json` ? It looks like that implementation isn't working correctly anyway, so we don't need to keep both.

Alternative is to fix `delete_from_..` I guess, instead of introducing the `delete`-method. I don't know what is easier - I have no preference either way though.

rafcio19:
Good point, I removed `delete_from_cloudformation_json`. The way deletes are attempted it makes more sense to use a common delete method: https://github.com/rafcio19/moto/blob/cf_delete_error/moto/cloudformation/parsing.py#L893

The upside is that the delete method is "self-aware", no need to figure out resource names or ids like with `delete_from_cloudformation_json`.

The crux of [the issue](https://github.com/getmoto/moto/issues/7527) was that launch template's name was a [CF sub](https://github.com/getmoto/moto/pull/7529/files#diff-8aad8f517fd5b893b2951d65fd603a0aa2877d2286262e35933703b57d7bd2cfR321) instead of a string, and that was parsed incorrectly.

rafcio19:
@akashkiran-s for now this code lives in this PR and my branch only, not in `v5.0.3`. Once merged, it will become a part of dev release, then a proper release.

akashkiran-s:
@rafcio19 I meant that I added the changes on top of v5.0.3 and tested the bug on local moto server.
I believe that `self.id` should also be passed
Screenshot 2024-03-28 at 1 11 25 AM
 
Can you try it out once with using the local moto server?

rafcio19:
@akashkiran-s it works as expected, I suggest you wait until this is merged and you can install the built package:

```bash
(moto) ➜  moto git:(cf_delete_error) aws --endpoint-url http://localhost:5000 cloudformation create-stack --stack-name test01 --template-body file:///tmp/template --parameters ParameterKey="Subnets",ParameterValue=subnet-f966c513\\,subnet-1a1fe2e6 ParameterKey="StackName",ParameterValue="test01" --region us-west-2
{
    "StackId": "arn:aws:cloudformation:us-west-2:123456789012:stack/test01/586840fe-6090-4a45-8a24-e5b8e6e16534"
}
(moto) ➜  moto git:(cf_delete_error) aws --endpoint-url http://localhost:5000 cloudformation delete-stack --stack-name test01 --region us-west-2  
(moto) ➜  moto git:(cf_delete_error) aws --endpoint-url http://localhost:5000 cloudformation list-stacks --region us-west-2                     

{
    "StackSummaries": [
        {
            "StackId": "arn:aws:cloudformation:us-west-2:123456789012:stack/test01/586840fe-6090-4a45-8a24-e5b8e6e16534",
            "StackName": "test01",
            "TemplateDescription": "Trying to create ec2 with auto scaling group",
            "CreationTime": "2024-03-28T08:37:48+00:00",
            "StackStatus": "DELETE_COMPLETE"
        }
    ]
}

akashkiran-s:
@rafcio19 Sorry for the confusion, it does work as expected.

bblommers (approved):
Great, thanks for investigating and fixing this @rafcio19.

And @akashkiran-s thanks for confirming this works!

PR #7535, opened 2024-03-28
Cloudformation: ASG parameter error

rafcio19:
Closes https://github.com/getmoto/moto/issues/7528

### DIFF
--- a/moto/autoscaling/models.py
+++ b/moto/autoscaling/models.py
@@ -437,7 +437,7 @@ class FakeAutoScalingGroup(CloudFormationModel):
         autoscaling_backend: "AutoScalingBackend",
         ec2_backend: EC2Backend,
         tags: List[Dict[str, str]],
-        mixed_instance_policy: Optional[Dict[str, Any]],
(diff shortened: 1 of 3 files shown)


rafcio19:
@bblommers I don't know why codecov is failing.  
It's been happening on a few previous PRs but a rebase from master or a re-run of CI usually cleared it, not this time though. Mine is a tiny change 🤷

bblommers:
> Report is 143 commits behind head on master.

@rafcio19 That's why it's failing. Compared to a month ago, the codecov dropped a bit indeed. I don't know why they don't just compare to head, as that would be a more realistic diff... I'm essentially yolo-ing PR's for now, ignoring these failures (and hoping that CodeCov catches up at some point).

 ¯\\_(ツ)_/¯

bblommers (approved):
Thanks for the fix!

rafcio19:
@bblommers if CodeCov continues to disappoint then running [coverage](https://coverage.readthedocs.io/en/latest/) locally would be a good substitute.

PR #7547, opened 2024-04-01
fix tag operations in secretsmanager

macnev2013:
# Description
This PR resolves the problem outlined in https://github.com/localstack/localstack/issues/10477.

# Breaking Scenarios

### Scenario 1
  - Secret creation without tags.
  - Addition of tags to the secret.
  - Removal of all tags from the secret.
  - Describing the secret reveals a `Tags` key with an empty list `[]`. *(Previously, this functionality was defective in moto.)*

### Scenario 2
  - Secret creation without tags.
  - Addition of tags to the secret.
  - Complete tag removal using `untag_resource` by specifying all tags.
  - On describing the secret, it should not contain of any tags. *(Previously, this was not correctly implemented in moto.)*

# Fixes
- Changing the default value of tags to `None` to identify weather the tags ever added.
- Addition of `NoneType` handling in filtering functions
- Updated logic for `untag_resource` to remove multiple tags at once.

# Testing
- Updated existing test cases to cover the scenario fixed

### DIFF
--- a/moto/secretsmanager/list_secrets/filters.py
+++ b/moto/secretsmanager/list_secrets/filters.py
@@ -13,19 +13,23 @@ def description_filter(secret: "FakeSecret", descriptions: List[str]) -> bool:
 
 
 def tag_key(secret: "FakeSecret", tag_keys: List[str]) -> bool:
+    if not secret.tags:
+        return False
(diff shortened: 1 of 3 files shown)


bblommers (approved):
LGTM - thank you @macnev2013!

PR #7553, opened 2024-04-02
DynamoDB: duplicate error should be thrown for batch writer

rafcio19:
Closes https://github.com/getmoto/moto/issues/7538


bblommers (approved):
LGTM - thank you @rafcio19!

Issue #7554, opened 2024-04-02
Paginated DynamoDB query does not reproduce `ScanIndexForward=False` correctly


rbailey-godaddy:
This continues analysis of #7531, which still fails a test case in my environment.

Consider the following fixture:

```python

        # Create a dummy tasks table with GSI used by this function
        self.db_client = boto3.client("dynamodb")
        self.db_client.create_table(
            AttributeDefinitions=[
                {"AttributeName": "taskId", "AttributeType": "S"},
                {"AttributeName": "accountId", "AttributeType": "S"},
                {"AttributeName": "sortKey", "AttributeType": "S"},
            ],
            TableName="tasks",
            KeySchema=[
                {"AttributeName": "taskId", "KeyType": "HASH"},
            ],
            GlobalSecondaryIndexes=[
                {
                    "IndexName": "accountId",
                    "KeySchema": [
                        {"AttributeName": "accountId", "KeyType": "HASH"},
                        {"AttributeName": "sortKey", "KeyType": "RANGE"},
                    ],
                    "Projection": {
                        "ProjectionType": "INCLUDE",
                        "NonKeyAttributes": [
                            "findingStats",
                            "scopeKey",
                            "status",
                            "terminated",
                        ],
                    },
                },
            ],
            BillingMode="PAY_PER_REQUEST",
        )

        # Stash some references
        self.tasks_table = boto3.resource("dynamodb").Table("tasks")
        task_core.tasks_table = None

        # Create some fixture data
        self.tasks_table.put_item(
            Item={
                "taskId": "task-1",
                "accountId": "111111111111",
                "scopeKey": "111111111111#us-west-2#test-scanner",
                "sortKey": "20230801010203#task-1",
                "terminated": True,
            }
        )
        self.tasks_table.put_item(
            Item={
                "taskId": "task-2",
                "accountId": "111111111111",
                "scopeKey": "111111111111#us-west-2#test-scanner",
                "sortKey": "20230801020203#task-2",
                "terminated": True,
            }
        )
        self.tasks_table.put_item(
            Item={
                "taskId": "task-3",
                "accountId": "111111111111",
                "scopeKey": "111111111111#us-west-2#test-scanner",
                "sortKey": "20230801030203#task-3",
                "terminated": True,
            }
        )
        self.tasks_table.put_item(
            Item={
                "taskId": "task-4",
                "accountId": "222222222222",
                "scopeKey": "222222222222#us-west-2#test-scanner",
[...]

PR #7555, opened 2024-04-02
DynamoDB: query/scan secondary index fixes

rbailey-godaddy:
Recent logic added in 194088 to accommodate a missing start key brain-damaged pagination because determination of previously-encountered items was not implemented correctly for all use cases.

* Make sure comparisons are performed in the sense of `ScanIndexForward` (previously, this was assumed to be `True` -- the default -- and returned incorrect results when reverse sorting was requested)
* When querying/scanning a secondary index, make comparisons using the index attribute(s) instead of the primary table attribute(s)

This commit now passes my project's internal test suite, and also passes all existing moto DynamoDB unit tests. I expect to add a commit with a unit test that exercises a query with reverse sorting on a GSI with different key attributes, which describes my application's unit test that failed prior to this fix.

### DIFF
--- a/moto/dynamodb/models/table.py
+++ b/moto/dynamodb/models/table.py
@@ -698,19 +698,27 @@ class Table(CloudFormationModel):
                         f"Range Key comparison but no range key found for index: {index_name}"
                     )
 
+            actual_hash_attr = index_hash_key["AttributeName"]
+            actual_range_attr = (
+                index_range_key["AttributeName"] if index_range_key else None
+            )
+
             possible_results = []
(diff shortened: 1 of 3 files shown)


rbailey-godaddy:
Fixes: #7554

bblommers (approved):
Great - thank you for fixing this @rbailey-godaddy! That was indeed an edge case that I hadn't considered yet.

rbailey-godaddy:
Good deal! I have confirmed I get no regressions when testing with this version

Issue #7532, opened 2024-03-27
Missing field validation on put_events 


fernando-alves:
When using event bridge client, it's possible to publish events with empty require fields like `DetailType` and `Source`. If those are empty, I would expect the response to fail with an `InvalidArgument` error, similar to the result in aws:

```
[cloudshell-user@xxxx ~]$ aws events put-events --entries '[ { "Source": "com.testing.moto", "DetailType": "", "Detail": "{}", "EventBusName": "xxxxxx" } ]'
{
    "FailedEntryCount": 1,
    "Entries": [
        {
            "ErrorCode": "InvalidArgument",
            "ErrorMessage": "Parameter DetailType is not valid. Reason: DetailType is a required argument."
        }
    ]
}
```

I believe the current implementation checks for the presence of the fields, but not their value:

https://github.com/getmoto/moto/blob/d767a2799bb1a6a6f2ed05ceb7cbb0c789968b76/moto/events/models.py#L1303-L1324

**How to reproduce the issue:**

```python
import boto3
from moto import mock_aws


@mock_aws
def test_put_events_with_empty_detail_type():
    client = boto3.client("events", region_name="us-east-1")
    event_bus = client.create_event_bus(Name="testing-bus")
    event_bus_arn = event_bus["EventBusArn"]

    expected_response = {
        "FailedEntryCount": 1,
        "Entries": [
            {
                "ErrorCode": "InvalidArgument",
                "ErrorMessage": "Parameter DetailType is not valid. Reason: DetailType is a required argument.",
            }
        ],
    }

    response = client.put_events(
        Entries=[
            {
                "Source": "com.testing.moto",
                "DetailType": "",
                "Detail": "{}",
                "EventBusName": event_bus_arn,
            },
        ]
    )

    assert expected_response == response
```

This test fails with:

```
AssertionError: assert {'Entries': [...ntryCount': 1} == {'Entries': [...Attempts': 0}}
E
E         Differing items:
E         {'Entries': [{'ErrorCode': 'InvalidArgument', 'ErrorMessage': 'Parameter DetailType is not valid. Reason: DetailType is a required argument.'}]} != {'Entries': [{'EventId': '7f0ca1c6-5567-4702-a035-4f9a4be9f2aa'}]}
E         {'FailedEntryCount': 1} != {'FailedEntryCount': 0}
E         Right contains 1 more item:
[...]

PR #7557, opened 2024-04-02
Events: put_events should check if detail type, detail or source are empty

fernando-alves:
As discussed on #7532

### DIFF
--- a/moto/events/models.py
+++ b/moto/events/models.py
@@ -1301,21 +1301,21 @@ class EventsBackend(BaseBackend):
 
         entries = []
         for event in events:
-            if "Source" not in event:
+            if not event.get("Source"):
                 entries.append(
                     {
                         "ErrorCode": "InvalidArgument",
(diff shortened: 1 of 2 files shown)


rafcio19:
LGTM! The tests could be simplified by using pytest's [parametrize](https://docs.pytest.org/en/7.1.x/how-to/parametrize.html) - a single test with the  payload dict passed to `Entries` as parameter. 

See a straight forward example:

https://github.com/getmoto/moto/blob/fc5421a78d51711ed6b0dfb5f0b27f40a845fa60/tests/test_batch/test_batch.py#L7

fernando-alves:
@rafcio19 I've updated the test, is this what you had in mind?

rafcio19:
Looks great, but needs to be approved by the boss @bblommers

bblommers (approved):
LGTM! The tests look much cleaner now as well, which is nice.

Thank you for fixing this @fernando-alves!

PR #7563, opened 2024-04-04
fix(cognito-idp): avoid KeyError on unexisting refresh token

serozhenka:
If an unexisting refresh token was passed to the `initiate_auth` with the `REFRESH_TOKEN_AUTH` flow, firstly the operation fails because of `KeyError` and after a few retrials, Moto returns a 500 status code response.

The fix was to use the `.get()` method to get a dictionary member or `None` if such doesn't exist instead of direct dictionary access with `[]`.


bblommers (approved):
Make sense - thanks for fixing this @serozhenka!

PR #7571, opened 2024-04-05
SQS: sqs q name validation

rafcio19:
Closes https://github.com/getmoto/moto/issues/7559

### DIFF
--- a/tests/test_sqs/test_sqs.py
+++ b/tests/test_sqs/test_sqs.py
@@ -203,6 +203,7 @@ def test_create_fifo_queue_with_high_throughput():
         "name-with-hyphens",
         "Name-with_all_the_THings",
     ],
+    ids=["random", "underscores", "hyphens", "combined"],
 )
 def test_create_queue(q_name):
     sqs = boto3.resource("sqs", region_name=REGION)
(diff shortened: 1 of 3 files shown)


rafcio19:
@bblommers do you know why these tests are failing? Works locally for me, except some SSO tests  🤔

tests/test_sqs/test_sqs.py
```diff
@@ -171,10 +195,18 @@ def test_create_fifo_queue_with_high_throughput():
 
 
 @mock_aws
-def test_create_queue():
+@pytest.mark.parametrize(
+    "q_name",
+    [
+        str(uuid4())[0:6],
```
bblommers, line 201: The SQS tests are run in parallel, and each parallel process collects all test-names. After execution, the results of all processes are merged and checked for consistency. 

I think the error that we're seeing is that the test-names are not consistent across all test processes, due to this randomization. If I understand it correctly, this could be fixed by explicitly specifying the test-id's:
```
ids=["random", "underscores", "hyphens", "combined"]
```
As long as the test-ids are equal across every run, it should not be a problem.

bblommers: If you have `pytest-xdist` installed, you can run `pytest tests/test_sqs -n 4` locally to validate this theory

bblommers (approved):
Great - thanks for adding this @rafcio19!

PR #7597, opened 2024-04-15
Feature: Bedrock and Bedrock Agents

JoshLevyMN:
Support for Bedrock and Agents for Bedrock services. 
Includes support for create, delete, list, get, and tagging operations for custom models, model customization jobs, model invocation logging, agents, and knowledge bases.


moto/bedrock/models.py
```diff
@@ -0,0 +1,546 @@
...
+    def __init__(self, logging_config: Dict[str, Any]) -> None:
+        self.logging_config = logging_config
+
+
+class BedrockBackend(BaseBackend):
+    """Implementation of Bedrock APIs."""
+
+    def __init__(self, region_name: str, account_id: str) -> None:
+        super().__init__(region_name, account_id)
+        self.ModelCustomizationJobs: Dict[str, ModelCustomizationJob] = {}
```
bblommers, line 189: Variables should all by `snake_case` across both Backends, to keep the style consistent

moto/bedrock/models.py
```diff
@@ -0,0 +1,546 @@
...
+    def stop_model_customization_job(self, job_identifier: str) -> None:
+        if job_identifier in self.ModelCustomizationJobs:
+            self.ModelCustomizationJobs[job_identifier].status = "Stopped"
+        else:
+            raise ResourceNotFoundException(
+                f"Model customization job {job_identifier} not found"
+            )
+        return
+
+    def list_model_customization_jobs(
```
bblommers, line 287: Can we use the existing `@paginator`-decorator here? That way we can avoid all the custom pagination logic.

http://docs.getmoto.org/en/latest/docs/contributing/development_tips/utilities.html#paginator

moto/bedrockagent/responses.py
```diff
@@ -0,0 +1,162 @@
...
+class AgentsforBedrockResponse(BaseResponse):
+    """Handler for AgentsforBedrock requests and responses."""
+
+    def __init__(self) -> None:
+        super().__init__(service_name="bedrock-agent")
+
+    @property
+    def bedrockagent_backend(self) -> AgentsforBedrockBackend:
+        """Return backend instance specific for this region."""
+        # TODO
```
bblommers, line 20: Nitpick, but it would be good to have this

moto/bedrockagent/urls.py
```diff
@@ -0,0 +1,13 @@
+"""bedrockagent base URL and path."""
+
+from .responses import AgentsforBedrockResponse
+
+url_bases = [
+    r"https?://bedrock-agent\.(.+)\.amazonaws\.com",
+]
+
+url_paths = {
+    # "0/.*$": AgentsforBedrockResponse.dispatch,
```
bblommers, line 10: These additional URL's don't seem necessary

moto/bedrockagent/responses.py
```diff
@@ -0,0 +1,162 @@
...
+        # bedrockagent_backends is not yet typed
+        # Please modify moto/backends.py to add the appropriate type annotations for this service
+        return bedrockagent_backends[self.current_account][self.region]
+
+    # add methods from here
```
bblommers, line 25: The scaffold script adds them as a pointer/reminder when you start implementing a service, but there's no need to keep them now

moto/bedrockagent/responses.py
```diff
@@ -0,0 +1,162 @@
...
+
+    def list_tags_for_resource(self) -> str:
+        resource_arn = unquote(self.path.split("/")[-1])
+        tags = self.bedrockagent_backend.list_tags_for_resource(
+            resource_arn=resource_arn
+        )
+        return json.dumps(dict(tags=tags))
+
+
+# add templates from here
```
bblommers, line 162: Same here

bblommers (changes requested):
Hi @JoshLevyMN, thanks for the PR! Please see my comments.

I haven't gone through the tests yet, but I'll do that on a second/third pass.

JoshLevyMN:
Thanks for the feedback @bblommers, it should all be addressed now. 
All of the testserver tests are failing, and I'm not sure how to fix it.

bblommers:
Hi @JoshLevyMN, I believe the ServerMode tests are failing because of a bug with how Moto handles request matching. I'll fix that first, and then revisit this. All the changes LGTM btw!

bblommers (approved):

PR #7599, opened 2024-04-16
Rekognition: add DetectCustomLabels support

akx:
Response data copied from https://docs.aws.amazon.com/rekognition/latest/customlabels-dg/detecting-custom-labels.html#text-response.


akx:
I think the `TestNDeploy / rubytest / test (3.2) (pull_request)` failure is a flake? It seems to be failing in an SQS-related thing.

bblommers (approved):
LGTM - thank you for the PR @akx!

PR #7600, opened 2024-04-16
Sagemaker: Add update trial component

YHallouard:


YHallouard:
Hi, 

Can I have more insight on the two test failed ? They don't seems to be related to anything I changed, so I'm little puzzled

bblommers (approved):
LGTM - thanks for adding even more functionality here @YHallouard!

PR #7601, opened 2024-04-16
ec2: improve describe_instances() filters

rbailey-godaddy:
Extend `Filters=` support in `ec2.describe_instances()` so that it now supports `product-code` and `product-code.type` filters, with the following restriction:

* All codes registered in `amis.json` in an image's `product_codes` property are treated as "marketplace" codes, i.e. it is not possible to specify an image with "devpay" codes.

The latent formatting support clearly intended these to be used as marketplace codes, and that's the use case I require, so this scratches my itch as-is.

### DIFF
--- a/moto/ec2/models/amis.py
+++ b/moto/ec2/models/amis.py
@@ -145,6 +145,10 @@ class Ami(TaggedEC2Resource):
             return self.owner_id
         elif filter_name == "owner-alias":
             return self.owner_alias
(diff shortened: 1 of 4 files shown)


bblommers (approved):
LGTM - thanks @rbailey-godaddy!

rbailey-godaddy:
Confirmed working, thanks!

PR #7603, opened 2024-04-16
Athena: Check for None on config before iterating to avoid execute_query error

mcintoac-aws:
Resolves: https://github.com/getmoto/moto/issues/7602

Add a None type check before iterating. Check is to a recent addition to the Athena Execution model. This allows for None configs, as is supported by Athena for configurations overridden by Workspace configs.


bblommers (approved):
Makes sense - thanks for fixing this @mcintoac-aws!

PR #7620, opened 2024-04-22
add validation to describe_security_group_rules filter

pinzon:
This pull request includes an additional validation for the filters in the ec2 describe-security-group-rules operation. This resolves the issue identified in https://github.com/localstack/localstack/issues/10315.

### DIFF
--- a/moto/ec2/exceptions.py
+++ b/moto/ec2/exceptions.py
@@ -499,6 +499,14 @@ class InvalidInternetGatewayIdError(EC2ClientError):
         )
 
 
+class InvalidGroupIdMalformedError(EC2ClientError):
+    def __init__(self, group_id: str):
+        super().__init__(
+            "InvalidGroupId.Malformed",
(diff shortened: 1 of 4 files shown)


bblommers (commented):
Hey @pinzon, thanks for adding this!

Do you want to fix the error message as well as part of this PR?
The error message just underneath this validation is currently:
```
if not matches:
    raise InvalidSecurityGroupNotFoundError(
        "No security groups found matching the filters provided."
    )
```
And I think that should just be
```
if not matches:
    raise InvalidSecurityGroupNotFoundError(group_ids)
```
Haven't tested it, but it looks like that would solve the bug completely.

pinzon:
Hi @bblommers , Thank you for your suggestion. I extended the test to cover also the situation where a SG group doesn't exist and AWS doesn't raises that exception.

bblommers (approved):
Great - thank you @pinzon!

Issue #7591, opened 2024-04-11
Default cache cluster for elasticcache


archinksagar:
When we call create_cache_cluster, an additional cluster ("default") gets created, This is causing problem with our testing.
What is the reason for adding the default cluster? Can that be removed?

https://github.com/getmoto/moto/blob/cc1193076090f9daf90970db286b6b41581dd8bc/moto/elasticache/models.py#L157

bblommers:
@jfmainville I think this was added in your PR originally, #6754 

Do you remember why you added the default cluster? Is that automatically added in specific regions, or automatically created when a user creates clusters/users? 
I've tried this out in AWS, but i can't replicate a scenario where the `default` cluster is created automatically.

jfmainville:
Indeed, it's an oversight on my end, the default `cache_cluster` creation shouldn't be included as part of the Elasticache cache clusters initialization process, I created the MR #7622 to fix this issue.

bblommers:
Fixed as of moto >= 5.0.6.dev30. Thank you for raising this @archinksagar, and thanks again for the fix @jfmainville!

PR #7622, opened 2024-04-22
Refactor: Remove the Elasticache default cluster configuration

jfmainville:
# Purpose

Due to an oversight, a default cache cluster is being created during the initialization phase of the Elasticache model. The default cluster is being created under the `cache_cluster` dictionary.

# Changes

- Removed the default `cache_cluster` creation step
- Update the `test_describe_all_cache_clusters` Elasticache test to exclude the action to delete the default `cache_cluster`

#7591

### DIFF
--- a/moto/elasticache/models.py
+++ b/moto/elasticache/models.py
@@ -152,44 +152,7 @@ class ElastiCacheBackend(BaseBackend):
             no_password_required=True,
         )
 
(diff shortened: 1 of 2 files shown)


bblommers (approved):
Great, thank you so much for fixing this @jfmainville!

PR #7624, opened 2024-04-23
Panorama:  Add more feature about application instances

YHallouard:


bblommers (changes requested):
Hi @YHallouard, would you mind fixing the merge conflicts?

YHallouard:
> Collaborator

Hi @bblommers, tests are running well on my computer, I don't really understand why it is failing on CI

tests/test_panorama/test_application_instance.py
```diff
@@ -0,0 +1,394 @@
...
+    # Then
+    assert isinstance(response["ApplicationInstanceId"], str)
+    assert response["ApplicationInstanceId"].startswith("applicationInstance-")
+    assert len(response["ApplicationInstanceId"]) == 42
+
+
+@mock_aws
+def test_describe_application_instance() -> None:
+    # Given
+    panorama_client = boto3.client("panorama")
```
bblommers, line 75:
```suggestion
    panorama_client = boto3.client("panorama", "eu-west-1")
```

We assume that the ARN starts with `arn:aws:panorama:eu-west-1..` - but that is only the case if the region is set to `eu-west-1`. If the region is not explicitly configured, our CI assumes it is set in `us-east-1`, which results in a different ARN.

This will need to be changed in the other tests as well.

YHallouard:
Hi @bblommers, sorry for the delay, I totally forgot that Pull request...

tests/test_panorama/test_application_instance.py
YHallouard: 6de03bd1c02cf1d20ddfae1bff52e7373ccbe08f

bblommers:
> Hi @bblommers, sorry for the delay, I totally forgot that Pull request...

No problem @YHallouard, thank you as always for contributing!

bblommers (approved):

PR #7625, opened 2024-04-23
Workspaces: Fix directory type

archinksagar:
Changed the directory type of workspace directories from lowercase to Uppercase. This is for describe_workspace_directories api call.
https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/workspaces/client/describe_workspace_directories.html


bblommers (changes requested):
Looks like the linter is failing @archinksagar.

Just a hunch - are you using black? As we've recently switched to `ruff`, so the fix would be:
```
ruff format moto/workspaces
ruff check --fix moto/workspaces
```

bblommers (approved):
Hi @archinksagar, I reverted the formatting changes, that fixed the build.

Thank you for the PR!

PR #7655, opened 2024-05-01
Update getting_started.rst

clarkebishop:
Bucket="b1" throws an error.

Bucket names must be at least 3 characters long.

Bucket="bb1" works.

The code example does not work. This fixes it by making the Bucket names longer.


bblommers (approved):
Great catch @clarkebishop - thank you for both PR's!

PR #7665, opened 2024-05-03
IAM: add instance profile tagging

SoenkeD:
IAM instance profile missed an implementation to add & remove tags (besides creation). This PR adds the functionality.


bblommers (approved):
LGTM - thank you for adding these methods @SoenkeD!

PR #7667, opened 2024-05-04
Update Data Brew Recipe Latest Working Version Create Date

runstache:
Corrects an issue where the Create Date of the Latest Working version remains the same as the published version. This should be the Publish Date as observed through interactions with AWS API Interactions for Publish Recipe.

Steps:

* Added field setting in moto/databrew/models.py
* Updated Publish Recipe Tests to compare publish date and latest working create date

Issues Addressed:

* [Issue 7666](https://github.com/getmoto/moto/issues/7666)


bblommers (approved):
LGTM - thank you for the fix @runstache!

PR #7673, opened 2024-05-06
S3: add EventBridge notification for Object Tagging

MikiPWata:
Added EventBridge notification for S3:PutObectTagging

### DIFF
--- a/moto/s3/models.py
+++ b/moto/s3/models.py
@@ -2291,6 +2291,10 @@ class S3Backend(BaseBackend, CloudWatchMetricProvider):
     ) -> FakeKey:
         if key is None:
             raise MissingKey(key=key_name)
+        
+        # get bucket for eventbridge notification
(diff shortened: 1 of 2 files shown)


MikiPWata:
Hey @bblommers!
Im trying to add eventbridge notification for S3:PutObjectTagging, but linter wouldn't pass.
Do you have any better approach on how to get the bucket in `notification.send_event ` for `put_object_tagging`?

bblommers:
Hmm.. I don't think there is a clean way to do this @MikiPWata. We know that the `bucket_name` will realistically always be there, so I think we can just add a `# type: ignore` to get the linter to pass

MikiPWata:
@bblommers 
thanks for the comment!
Do you mind reviewing this?
thanks in advance!

bblommers (approved):
LGTM - thank you @MikiPWata!

PR #7677, opened 2024-05-07
RouteTables: Filter by route.nat-gateway-id

yoelsusanto:
## Background
In my work, we do DescribeRouteTables and filter by `route.nat-gateway-id`. However, it is currently not implemented yet and returns the following error:
```
moto.ec2.exceptions.FilterNotImplementedError: The filter 'route.nat-gateway-id' for DescribeRouteTables has not been implemented in Moto yet. Feel free to open an issue at https://github.com/getmoto/moto/issues
```

## Changes
- Implement filter by `route.nat-gateway-id`
- Add test


bblommers (approved):
Great - thank you for the PR @yoelsusanto, and welcome to Moto!

PR #7679, opened 2024-05-07
EC2: Add test for DescribeNetworkAcls using the association.subnet-id filter

marcciosilva:
I attempted to add coverage to DescribeNetworkAcls when using the `association.subnet-id` filter (since [this issue](https://github.com/localstack/localstack/issues/10097) was reported a while ago) to ensure it works as expected.

This also fixes a small bug where network ACLs could be deleted even when associated to subnets (according to [AWS's documentation](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_DeleteNetworkAcl.html), this cannot happen).

### DIFF
--- a/moto/ec2/models/network_acls.py
+++ b/moto/ec2/models/network_acls.py
@@ -63,8 +63,8 @@ class NetworkAclBackend:
 
     def delete_network_acl(self, network_acl_id: str) -> "NetworkAcl":
         if any(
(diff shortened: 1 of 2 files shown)


bblommers (approved):
LGTM - thank you for fixing this and for adding the test case @marcciosilva!

PR #7697, opened 2024-05-16
EC2: Eni detach error

rafcio19:
Resolves https://github.com/getmoto/moto/issues/7678

### DIFF
--- a/tests/test_ec2/test_elastic_network_interfaces.py
+++ b/tests/test_ec2/test_elastic_network_interfaces.py
@@ -951,14 +951,14 @@ def test_elastic_network_interfaces_describe_attachment():
 def test_eni_detachment():
     # Setup
     ec2resource, ec2client, vpc, subnet = setup_vpc(boto3)
-    ec2client.run_instances(ImageId="ami-12c6146b", MinCount=1, MaxCount=1)
+    resp = ec2client.run_instances(ImageId="ami-12c6146b", MinCount=1, MaxCount=1)[
+        "Instances"
+    ][0]
 
     # Execute
-    enis = ec2client.describe_network_interfaces()
+    eni_id = resp["NetworkInterfaces"][0]["Attachment"]["AttachmentId"]
     with pytest.raises(ClientError) as ex:
-        ec2client.detach_network_interface(
-            AttachmentId=enis["NetworkInterfaces"][0]["Attachment"]["AttachmentId"]
-        )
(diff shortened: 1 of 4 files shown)


rafcio19:
@bblommers I skipped the new test in server mode as it was failing and I couldn't repro locally.

bblommers (changes requested):
The ServerMode error can be reproduced by running the tests in parallel:
```
MOTO_CALL_RESET_API=false TEST_SERVER_MODE=true pytest -sv tests/test_ec2/test_elastic_network_interfaces.py -n 4
```

The reason it fails is because there will be other ENI's present (created by other tests) - so the `len(enis)` will always be greater than 0.

I think the logic should be changed so that we check the number of ENI's on that specific instance:
```
for eni ..:
    if eni.attachment_id..:
        if len(eni.instance.nics) ==1:
            raise LastEniDetachError
        detach_eni..
```

tests/test_ec2/test_elastic_network_interfaces.py
```diff
@@ -1040,12 +936,46 @@ def test_elastic_network_interfaces_describe_attachment():
...
     ).get("SourceDestCheck")
     assert my_eni_source_dest_check["Value"] is True
+
+
+@mock_aws
+@pytest.mark.skipif(settings.TEST_SERVER_MODE, reason="Fails in servermode")
+def test_eni_detachment():
+    # Setup
+    ec2resource, ec2client, vpc, subnet = setup_vpc(boto3)
+    ec2client.run_instances(ImageId="ami-12c6146b", MinCount=1, MaxCount=1)
```
bblommers, line 955:
```suggestion
    resp = ec2client.run_instances(ImageId="ami-12c6146b", MinCount=1, MaxCount=1)["Instances"][0]
    eni_id = resp["NetworkInterfaces"][0]["Attachment"]["AttachmentId"]

    # Execute
    with pytest.raises(ClientError) as ex:
        ec2client.detach_network_interface(AttachmentId=eni_id)
```

Because the tests are run in parallel, and other ENI's exist, we can't just call `describe_network_interfaces` and get the first one - because that is not guaranteed to be ours.
We can get our ENI from the `run_instances`-call though.

(Note that I'm adding this as a suggestion, but I haven't formatted/linted this... )

rafcio19: 🤘

bblommers (approved):
Great - thank you @rafcio19!

PR #7700, opened 2024-05-16
cache methods_url for BaseResponse._get_action_from_method_and_request_uri

bentsku:
Following #7626, we've noticed a regression in performance while calling API Gateway. Trying to pin down the issue, it seems `BaseResponse._get_action_from_method_and_request_uri` was responsible for it, as the caching is now disabled (following #6746) as we create a new `APIGatewayResponse` for every call to `BaseResponse.dispatch`. API Gateway used `BaseResponse.dispatch_method` before the refactor, which didn't need to route the incoming request to the right operation. 

This PR pulls building the `method_url` out of `_get_action_from_method_and_request_uri` in order to be cacheable (the cache key is the service name and the region, as it seems we're creating the specs based on it, not sure the region is really mandatory?). We then used the cached result in order to regex match the incoming request to find the right operation. 

Not exactly sure how to test this exactly, but by testing locally I could see the slowdown (1s for every APIGW request) had disappeared. 

Thanks!

### DIFF
--- a/moto/core/responses.py
+++ b/moto/core/responses.py
@@ -81,6 +81,21 @@ def _decode_dict(d: Dict[Any, Any]) -> Dict[str, Any]:
     return decoded
 
 
+@functools.cache
+def _get_method_urls(service_name: str, region: str) -> Dict[str, Dict[str, str]]:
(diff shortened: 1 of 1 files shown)


bblommers (approved):
Great - thank you for fixing this @bentsku!

PR #7714, opened 2024-05-21
ECS: update_service reflects correct counts

rafcio19:
Resolves https://github.com/getmoto/moto/issues/7664


bblommers (approved):
Great - thank you for fixing this @rafcio19!

PR #7717, opened 2024-05-22
Fit service name for sso-admin

giograno:
This small PR fixes the service name passed to the `BackendDict` for the SSO Admin service.


bblommers (approved):
LGTM - thank you @giograno!

PR #7723, opened 2024-05-28
Fix BackendDict instantiation service names

viren-nadkarni:
This PR updates certain BackendDict initialisations so that they use proper Botocore service name.

Earlier, these services used `ec2` or `lambda` since `boto3.Session().get_available_regions(...)` used to return an empty list.

There are still a few instantiations remaining that were not updated for the above reason, these are:
- `redshiftdata` should use `redshift-data`
- `apigatewaymanagementapi` should use `apigatewaymanagementapi`
- `bedrockagent`
- `ec2instanceconnect` should use `ec2-instance-connect`

### DIFF
--- a/moto/applicationautoscaling/models.py
+++ b/moto/applicationautoscaling/models.py
@@ -673,4 +673,6 @@ class FakeScheduledAction(BaseModel):
         self.end_time = end_time
 
 
-applicationautoscaling_backends = BackendDict(ApplicationAutoscalingBackend, "application-autoscaling")
+applicationautoscaling_backends = BackendDict(
+    ApplicationAutoscalingBackend, "application-autoscaling"
+)
--- a/moto/applicationautoscaling/models.py
+++ b/moto/applicationautoscaling/models.py
@@ -673,4 +673,4 @@ class FakeScheduledAction(BaseModel):
         self.end_time = end_time
 
 
-applicationautoscaling_backends = BackendDict(ApplicationAutoscalingBackend, "ec2")
+applicationautoscaling_backends = BackendDict(ApplicationAutoscalingBackend, "application-autoscaling")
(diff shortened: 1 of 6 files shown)


bblommers:
Hi @viren-nadkarni!

> boto3.Session().get_available_regions(...) used to return an empty list

Do you know from which version of boto3 this changed? If this only happened in a recent version, then people that are stuck on an older version of `boto3` will suddenly have an empty region-list, so they will no longer be able to use Moto - and that worries me a bit.

What is the effect (on LS) of changing these names?

viren-nadkarni:
@bblommers This has been out in Boto for quite a while. We made this change in LocalStack in July 2023, see https://github.com/localstack/moto/pull/68

Also related https://github.com/getmoto/moto/pull/6446

viren-nadkarni:
For the new `iot-data` change, I tested Boto3 1.28.0 (7 July 2023). It was working with that release:

```
...
Successfully installed boto3-1.28.0 botocore-1.31.85 jmespath-1.0.1 python-dateutil-2.9.0.post0 s3transfer-0.6.2 six-1.16.0 urllib3-2.0.7

$ py
Python 3.10.12 (main, Nov 20 2023, 15:14:05) [GCC 11.4.0] on linux
>>> import boto3
>>> boto3.session.Session().get_available_regions('iot-data')
['ap-east-1', 'ap-northeast-1', 'ap-northeast-2', 'ap-south-1', 'ap-southeast-1', 'ap-southeast-2', 'ca-central-1', 'eu-central-1', 'eu-north-1', 'eu-west-1', 'eu-west-2', 'eu-west-3', 'me-central-1', 'me-south-1', 'sa-east-1', 'us-east-1', 'us-east-2', 'us-west-1', 'us-west-2']
```

bblommers (approved):
Ah, that should be fine then. Thank you @viren-nadkarni!

PR #7732, opened 2024-05-30
EC2: Fix inconsistent private/public key pair attribute

viren-nadkarni:
Currently in the internal representation for EC2 Key Pairs, the `KeyPair` class defines an attribute `material`. However this attribute is not consistently used.

If the key pair object is created using `ImportKeyPair`, `material` contains the base64 encoded public key. If the key pair is created using `CreateKeyPair`, `material` contains the PEM encoded private key.

This PR introduces makes changes such that `material` always contains the private key material (wherever applicable) and adds a new attribute `material_public` always contains the public key in OpenSSH format. There is no impact on the external API.

### DIFF
--- a/moto/ec2/models/key_pairs.py
+++ b/moto/ec2/models/key_pairs.py
@@ -23,14 +23,16 @@ class KeyPair(TaggedEC2Resource):
         self,
         name: str,
         fingerprint: str,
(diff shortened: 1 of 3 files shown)


bblommers (approved):
LGTM!

PR #7750, opened 2024-06-05
EC2: describe_images() - feat: support filtering images by ExecutableUsers=['self']

daeMOn63:
This adds a missing mapping on `self` to the current `account_id` for the `describe_images` `exec_users` endpoint parameter, as well as a check bypass for public images.

Similar to the `Owners` field, `ExecutableUsers` can be filtered by `self` (ref: https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_DescribeImages.html#API_DescribeImages_RequestParameters)

Currently, filtering images with `exec_users=["self"]` returns an empty list, even when having public images.

### DIFF
--- a/moto/ec2/models/amis.py
+++ b/moto/ec2/models/amis.py
@@ -274,18 +274,21 @@ class AmiBackend:
         else:
             # Limit images by launch permissions
             if exec_users:
-                # support filtering by ExecutableUsers=['self']
-                if "self" in exec_users:
-                    exec_users = list(
-                        map(lambda o: self.account_id if o == "self" else o, exec_users)  # type: ignore[attr-defined]
-                    )
-
                 tmp_images = []
                 for ami in images:
-                    for user_id in exec_users:
-                        for lp in ami.launch_permissions:
-                            if lp.get("UserId") == user_id:
-                                tmp_images.append(ami)
+                    if ami.is_public:
+                        tmp_images.append(ami)
+                    else:
+                        # support filtering by ExecutableUsers=['self']
+                        if "self" in exec_users:
+                            exec_users = list(
+                                map(lambda o: self.account_id if o == "self" else o, exec_users)  # type: ignore[attr-defined]
+                            )
+
+                        for user_id in exec_users:
+                            for lp in ami.launch_permissions:
+                                if lp.get("UserId") == user_id:
+                                    tmp_images.append(ami)
                 images = tmp_images
 
             # Limit by owner ids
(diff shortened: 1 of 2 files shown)


daeMOn63:
Not sure what's up with the server tests; they're running fine locally for me

moto/ec2/models/amis.py
```diff
@@ -276,10 +276,22 @@ def describe_images(
...
+                        tmp_images.append(ami)
+                    else:
+                        # support filtering by ExecutableUsers=['self']
+                        if "self" in exec_users:
+                            exec_users = list(
+                                map(
+                                    lambda o: self.account_id if o == "self" else o,  # type: ignore[attr-defined]
+                                    exec_users,
+                                )
+                            )
```
bblommers, line 289:
```suggestion
                            exec_users = list(set(exec_users))
                            exec_users.remove("self")
                            exec_users.append(self.account_id)
```
If I understand the logic correctly, this should be equivalent? And better readable IMO.
The `set` operation is just there for the corner-case that users send a request like `exec_users=["self", "self"]`.

bblommers (changes requested):
Hi @daeMOn63!

The default behaviour is to remove the state after every test, by calling the reset API.

The CI runs the EC2 tests in parallel though, and the state is not discarded after every test. So I think what's happening is that the failing tests are suddenly seeing images that were created by other test cases. Hence errors like the `AssertionError: assert 1 == 0`.

We typically handle this by verifying that an image exists, rather then verifying the length of the result. I.e.:
```
assert my_image in list_of_images
```
instead of
```
assert len(list_of_images) == 1
```
This way, the `list_of_images` could contain any number of images (created by other tests), and we only check the existence of the image created by that particular test.

You should be able to reproduce these errors locally by running the following command:
```
MOTO_CALL_RESET_API=false TEST_SERVER_MODE=true pytest -sv -n 4 tests/test_ec2
```

Speaking of tests - would you mind adding a test case (/extending an existing test case) to verify that this behaviour works? The logic looks correct, but it's always good to verify it works (and keeps working).

daeMOn63:
Thanks a lot @bblommers for all the details! 

Indeed I was missing the `MOTO_CALL_RESET_API` flag to reproduce locally the failures. 
With this, I managed to reproduce, fix, and extend the tests to cover the  changes

Cheers!

bblommers (approved):
LGTM - thank you again for adding this to Moto @daeMOn63!

PR #7752, opened 2024-06-06
RDS: parse more input parameters

rafcio19:
https://github.com/getmoto/moto/issues/7734


bblommers (approved):
LGTM! Good to have the validation here as well. Thank you for the PR @rafcio19!

PR #7754, opened 2024-06-07
fix(eks): set taints in create_nodegroup

johnko:
# The Bug

## Scenario
```
moto.eks.create_nodegroup(
  cluserName="foo",
  nodegroupName="bar",
  taints=[{"key":"soft","effect":"PREFER_NO_SCHEDULE"}]
)

print(
  moto.eks.describe_nodegroup(
    cluserName="foo",
    nodegroupName="bar",
  ).get('nodegroup', {}).get('taints')
)
```

### Expectation

I expect `taints` to be `[{"key":"soft","effect":"PREFER_NO_SCHEDULE"}]`

### Observation

I see `taints` is `[]`

## The fix

This pull request

### DIFF
--- a/moto/eks/responses.py
+++ b/moto/eks/responses.py
@@ -76,6 +76,7 @@ class EKSResponse(BaseResponse):
         remote_access = self._get_param("remoteAccess")
         node_role = self._get_param("nodeRole")
         labels = self._get_param("labels")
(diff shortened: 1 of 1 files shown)


bblommers (approved):
Thanks for the fix @johnko!

PR #7764, opened 2024-06-12
SSM: return ssm param if arn used as a name

rafcio19:
Closes https://github.com/getmoto/moto/issues/7748


bblommers (approved):
LGTM - thanks @rafcio19!

PR #7767, opened 2024-06-12
feat: Shield API

archinksagar:
This is the initial moto implementation of AWS Shield api. I have added 7 API endpoints. Can I get a feedback on this.

### DIFF
--- a/moto/shield/models.py
+++ b/moto/shield/models.py
@@ -1,0 +1,202 @@
+"""ShieldBackend class with methods for supported APIs."""
+
+from typing import Any, Dict, List
+
+from moto.core.base_backend import BackendDict, BaseBackend
+from moto.core.common_models import BaseModel
+from moto.moto_api._internal import mock_random
+from moto.shield.exceptions import (
+    InvalidParameterException,
+    InvalidResourceException,
+    ResourceAlreadyExistsException,
+    ResourceNotFoundException,
+    ValidationException,
+)
+from moto.utilities.tagging_service import TaggingService
+
+
+class Protection(BaseModel):
+    def __init__(
+        self, account_id: str, name: str, resource_arn: str, tags: List[Dict[str, str]]
+    ):
+        self.name = name
+        self.resource_arn = resource_arn
+        self.protection_id = str(mock_random.uuid4())
+        self.tags = tags
+        self.health_check_ids: list[
+            str
+        ] = []  # value is returned in associate_health_check method.
+        # value is returned in enable_application_layer_automatic_response and disable_application_layer_automatic_response methods.
+        self.application_layer_automatic_response_configuration: dict[str, Any] = {}
+        self.protection_arn = (
+            f"arn:aws:shield::{account_id}:protection/{self.protection_id}"
+        )
+
+        resource_types = {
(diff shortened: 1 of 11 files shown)


moto/shield/models.py
```diff
@@ -0,0 +1,202 @@
...
+
+
+class Protection(BaseModel):
+    def __init__(
+        self, account_id: str, name: str, resource_arn: str, tags: List[Dict[str, str]]
+    ):
+        self.name = name
+        self.resource_arn = resource_arn
+        self.protection_id = str(mock_random.uuid4())
+        self.tags = tags
```
bblommers, line 25: Nitpick - it looks like tags is never used, so we don't have to store it here

moto/shield/models.py
```diff
@@ -0,0 +1,202 @@
...
+                    raise ValidationException(
+                        "Error validating the following inputs: inclusionFilters.resourceTypes"
+                    )
+                type_protections = [
+                    protection
+                    for protection in self.protections.values()
+                    if protection.resource_type == resource_types[0]
+                ]
+            try:
+                protections = list(
```
bblommers, line 167: Can we change this to:

```
protections = []
protections.extend(type_protections)
#two more...
protections = list(set(protections))
```
I think that's logically equivalent? Less fancy, but more readable IMO.

moto/shield/models.py
```diff
@@ -0,0 +1,202 @@
...
+        return self.tagger.list_tags_for_resource(resource_arn)["Tags"]
+
+    def tag_resource(self, resource_arn: str, tags: List[Dict[str, str]]) -> None:
+        self.tagger.tag_resource(resource_arn, tags)
+
+    def untag_resource(self, resource_arn: str, tag_keys: List[str]) -> None:
+        self.tagger.untag_resource_using_names(resource_arn, tag_keys)
+
+
+shield_backends = BackendDict(ShieldBackend, "ec2")
```
bblommers, line 202: I can't test this now, otherwise I would try it out. But can we change `"ec2"` to `"shield"`? Or will the tests fail if we do that?

bblommers (changes requested):

moto/shield/models.py
archinksagar: @bblommers It gives me **KeyError: 'aws'**, when I change it to "shield"

moto/shield/models.py
archinksagar: The above logic  fails this test : **test_list_protections_with_resource_arn_and_protection_name():**.  It gives me "AssertionError: assert 2 == 1". According to this: https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/shield/client/list_protections.html
**Shield Advanced returns protections that exactly match all of the filter criteria that you provide.**. That means the resources listed should satisfy all the filter conditions in the **InclusionFilters**.

moto/shield/models.py
archinksagar: @bblommers  Removed the tags

moto/shield/models.py
bblommers: Ah, gotcha, I misunderstood the logic

bblommers (approved):
LGTM - thank you again @archinksagar!

PR #7768, opened 2024-06-13
elbv2: add client_keep_alive.seconds to modify load balancer attributes

sannya-singal:
As per the current implementation if we use `client_keep_alive.seconds` in `elbv2` we get `InvalidConfigurationRequest` error thrown. 

This PR adds `client_keep_alive.seconds ` attribute to the `modify_load_balancer_attributes` handler and writes a test to verify the changes. 

https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-elasticloadbalancingv2-loadbalancer-loadbalancerattribute.html


bblommers (approved):
Great, thank you @sannya-singal!

PR #7769, opened 2024-06-14
Add correct partition for get_federation_token response

dfangl:
## Motivation

The `get_federation_token` operation did not return the arn with the correct partition in the `FederatedUser.Arn` field, it was hardcoded to `aws`.

## Changes
* Return the federated user ARN with the correct partition for the `get_federation_token` response


bblommers (approved):
Nice catch @dfangl - thanks for the fix!

PR #7784, opened 2024-06-20
Fix cfn-lint breaking changes

viren-nadkarni:
[cfn-lint 1.3.0](https://pypi.org/project/cfn-lint/1.3.0) was released on 2024-06-19.

Currently Moto pins it as `cfn-lint>=0.40.0` which causes the latest version to be picked up and integration to break.

This PR fixes the breaking changes and updates its usage.

### DIFF
--- a/moto/cloudformation/utils.py
+++ b/moto/cloudformation/utils.py
@@ -1,4 +1,3 @@
-import os
 import string
 from typing import Any, List
 from urllib.parse import urlparse
@@ -59,24 +58,10 @@ def yaml_tag_constructor(loader: Any, tag: Any, node: Any) -> Any:
 
 def validate_template_cfn_lint(template: str) -> List[Any]:
     # Importing cfnlint adds a significant overhead, so we keep it local
-    from cfnlint import core, decode
-
-    # Save the template to a temporary file -- cfn-lint requires a file
-    filename = "file.tmp"
-    with open(filename, "w") as file:
-        file.write(template)
-    abs_filename = os.path.abspath(filename)
-
-    # decode handles both yaml and json
-    try:
-        template, matches = decode.decode(abs_filename, False)
-    except TypeError:
-        # As of cfn-lint 0.39.0, the second argument (ignore_bad_template) was dropped
-        # https://github.com/aws-cloudformation/cfn-python-lint/pull/1580
-        template, matches = decode.decode(abs_filename)
+    from cfnlint import api, config, core
 
     # Set cfn-lint to info
-    core.configure_logging(None)
+    config.configure_logging(None, False)
 
     # Initialize the ruleset to be applied (no overrules, no excludes)
     rules = core.get_rules([], [], [])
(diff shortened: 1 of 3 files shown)


moto/cloudformation/utils.py
```diff
@@ -59,24 +58,10 @@ def _f(loader: Any, tag: Any, node: Any) -> Any:
 
 def validate_template_cfn_lint(template: str) -> List[Any]:
     # Importing cfnlint adds a significant overhead, so we keep it local
-    from cfnlint import core, decode
-
-    # Save the template to a temporary file -- cfn-lint requires a file
-    filename = "file.tmp"
-    with open(filename, "w") as file:
-        file.write(template)
-    abs_filename = os.path.abspath(filename)
```
viren-nadkarni, line 68: Cfnlint no longer requires a file input

moto/cloudformation/utils.py
```diff
@@ -59,24 +58,10 @@ def _f(loader: Any, tag: Any, node: Any) -> Any:
...
-
-    # decode handles both yaml and json
-    try:
-        template, matches = decode.decode(abs_filename, False)
-    except TypeError:
-        # As of cfn-lint 0.39.0, the second argument (ignore_bad_template) was dropped
-        # https://github.com/aws-cloudformation/cfn-python-lint/pull/1580
-        template, matches = decode.decode(abs_filename)
```
viren-nadkarni, line 76: The new `cfnlint.api.lint(...)` handles this itself

moto/cloudformation/utils.py
```diff
@@ -59,24 +58,10 @@ def _f(loader: Any, tag: Any, node: Any) -> Any:
...
+    from cfnlint import api, config, core
 
     # Set cfn-lint to info
-    core.configure_logging(None)
```
viren-nadkarni, line 79: `cfnlint.core.configure_logging(...)` was removed

moto/cloudformation/utils.py
```diff
@@ -85,9 +70,7 @@ def validate_template_cfn_lint(template: str) -> List[Any]:
     regions = ["us-east-1"]
 
     # Process all the rules and gather the errors
-    matches = core.run_checks(abs_filename, template, rules, regions)
```
viren-nadkarni, line 88: `cfnlint.core.run_checks(...)` was removed

bblommers:
Hi @viren-nadkarni, thanks for the PR!

V1 only came out recently, so I think we want to support both v0 and v1 for now.
I'm sure there are plenty of people who want to continue using v0 for now, and I don't want to force anyone to upgrade.

I'd be happy to make the necessary changes to support both versions, we already do that for a few other dependencies (see moto/core/versions py) - I just don't know when I'll have time for that.

------

Related announcement from AWS, for future reference:
https://aws.amazon.com/blogs/devops/aws-cloudformation-linter-v1/

bblommers (approved):
Awesome, thank you for fixing this @viren-nadkarni!

PR #7789, opened 2024-06-24
EC2: Enabled regex matching in filters

aranjan1002:
There could be filters which uses regex instead of full string for filter value. I added the ability to filter using regex, when only one filter value is provided. 

This was talked about in a closed issue: https://github.com/getmoto/moto/issues/7735#issuecomment-2144092920


bblommers (approved):
LGTM - thank you for following up on this and implementing this @aranjan1002!

PR #7794, opened 2024-06-27
elbv2: check for too many certificates

giograno:
Attaching too many certificates to a load balancer listener (with the `AddListenerCertificates` operation) should raise a `TooManyCertificatesException`. This check is not currently implemented.

25 seems to be the limit as per [AWS docs](https://docs.aws.amazon.com/elasticloadbalancing/latest/application/load-balancer-limits.html).


bblommers (approved):
Great - thanks for adding this check @giograno!

PR #7796, opened 2024-06-28
Fix docs typos

xofbd:
A few typos/fixes I saw when reading the docs.


bblommers (approved):
Good catch @xofbd - thank you for fixing this!

PR #7813, opened 2024-07-02
ResourceGroupsTagging: Add support for `elasticfilesystem` resource type filter

### LINKED ISSUE #7805: EFS not supported by resourcegroupstaggingapi
Apologies, I'm back again with another resource groups tagging api feature request. It looks like EFS file systems aren't supported in moto's implementation, though they are supported in this api by AWS (https://docs.aws.amazon.com/resourcegroupstagging/latest/APIReference/supported-services.html).

```python
@mock_aws
class MyTestClass(unittest.TestCase):
    def setUp(self):
        efs = boto3.client("efs", region_name="us-east-1")
        efs.create_file_system(CreationToken="1234")

    def test_stuff(self):
        boto3.client("resourcegroupstaggingapi", region_name="us-east-1")
            .get_resources()["ResourceTagMappingList"] # No resources are returned, expected the file system created in setUp()
```
Taking a look in the implementation it looks like this is missing altogether so this would be a net-new feature rather than a bug with any existing code https://github.com/getmoto/moto/blob/master/moto/resourcegroupstaggingapi/models.py.

Relevant Versions:
moto - 5.0.9
boto3 - 1.34.135
Python - 3.11.6
MacOS - 14.5

bpandola:
Closes #7805

### DIFF
--- a/moto/resourcegroupstaggingapi/models.py
+++ b/moto/resourcegroupstaggingapi/models.py
@@ -8,6 +8,7 @@ from moto.core.exceptions import RESTError
 from moto.dynamodb.models import DynamoDBBackend, dynamodb_backends
 from moto.ec2 import ec2_backends
 from moto.ecs.models import EC2ContainerServiceBackend, ecs_backends
+from moto.efs.models import EFSBackend, efs_backends
 from moto.elb.models import ELBBackend, elb_backends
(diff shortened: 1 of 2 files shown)


bblommers (approved):
That was quick! Thanks @bpandola!

PR #7833, opened 2024-07-09
RDS: Boolean Value Fixes

bpandola:
Just a few little things I came across while working on something else.  Yes, `None` is technically "falsy", but ensuring these boolean values can only be `True` or `False` makes the code more correct and makes a difference when serializing responses.  

@bblommers These commits be atomic!  😄  Please `Rebase and merge` instead of squashing.  Thanks!


bblommers (approved):
Great, thank you @bpandola!

PR #7840, opened 2024-07-10
S3: Fix bucket policy enforcement for s3:PutObject when creating new objects

### LINKED ISSUE #7837: S3: Not throwing an exception with a Deny bucket policy for s3:PutObject
Hello, I'm using pytest v8.2.2, boto3, botocore v1.34.142 and moto v5.0.11 installed with poetry, and the following code snippet passes just fine without throwing an exception:
```
def test_s3_deny(mock_s3):
    s3 = boto3.client("s3")
    policy = {
        "Version": "2012-10-17",
        "Statement": [
            {
                "Effect": "Deny",
                "Principal": "*",
                "Action": ["s3:PutObject"],
                "Resource": "arn:aws:s3:::mybucket/*",
            }
        ],
    }
    s3.create_bucket(Bucket="mybucket")
    s3.put_bucket_policy(Bucket="mybucket", Policy=json.dumps(policy))
    s3.put_object(Body="test", Bucket="mybucket", Key="obj")
```

I don't know if it's relevant, but here's my conftest.py:
```
import os
import pytest
from moto import mock_aws
import boto3


@pytest.fixture(scope="function", autouse=True)
def aws_credentials():
    os.environ["AWS_DEFAULT_REGION"] = "us-east-1"

@pytest.fixture(scope="function")
def mock_s3():
    with mock_aws():
        yield boto3.client("s3")
```

According to #5883, it should throw an exception and weirdly there is a [test for this case](https://github.com/getmoto/moto/blob/master/tests/test_s3/test_s3_bucket_policy.py). I haven't tried without pytest because I don't think it's the problem here

Related to #5820

OwlInSpace:
Previously, policy enforcement for s3:PutObject only occurred if the key already existed (`if key:`), I changed this to (`if bucket:`) and now policy enforcement works when creating new objects too

I don't know if I broke anything or if I did something terribly wrong, but all the S3 tests pass

Fixes #7837

### DIFF
--- a/moto/s3/responses.py
+++ b/moto/s3/responses.py
@@ -1315,16 +1315,17 @@ class S3Response(BaseResponse):
             if bucket_permissions == PermissionResult.DENIED:
                 return 403, {}, ""
 
-            # If the request is not authorized, and not signed,
-            # that means that the action should be allowed for anonymous users
-            if key and not authorized_request and not signed_url:
-                # We already know that the bucket permissions do not explicitly deny this
-                # So bucket permissions are either not set, or do not explicitly allow
-                # Next check is to see if the ACL of the individual key allows this action
-                if bucket_permissions != PermissionResult.PERMITTED and (
-                    key.acl and not key.acl.public_read
-                ):
-                    return 403, {}, ""
+            if key:
+                # If the request is not authorized, and not signed,
+                # that means that the action should be allowed for anonymous users
+                if not authorized_request and not signed_url:
+                    # We already know that the bucket permissions do not explicitly deny this
+                    # So bucket permissions are either not set, or do not explicitly allow
+                    # Next check is to see if the ACL of the individual key allows this action
+                    if bucket_permissions != PermissionResult.PERMITTED and (
+                        key.acl and not key.acl.public_read
+                    ):
+                        return 403, {}, ""
             elif signed_url and not authorized_request:
                 # coming in from requests.get(s3.generate_presigned_url())
                 if self._invalid_headers(request.url, dict(request.headers)):
--- a/moto/s3/responses.py
+++ b/moto/s3/responses.py
@@ -1299,7 +1299,9 @@ class S3Response(BaseResponse):
             bucket = self.backend.get_bucket(bucket_name)
         except S3ClientError:
             key = bucket = None
-        if key:
+
+        # Enforces policy when creating new objects, `if key` does not https://github.com/getmoto/moto/issues/7837
+        if bucket:
             resource = f"arn:{bucket.partition}:s3:::{bucket_name}/{key_name}"  # type: ignore[union-attr]
 
             # Authorization Workflow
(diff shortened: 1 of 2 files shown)


moto/s3/responses.py
```diff
@@ -1313,21 +1315,21 @@ def _key_response(
...
+                # that means that the action should be allowed for anonymous users
+                if not authorized_request and not signed_url:
+                    # We already know that the bucket permissions do not explicitly deny this
+                    # So bucket permissions are either not set, or do not explicitly allow
+                    # Next check is to see if the ACL of the individual key allows this action
+                    if bucket_permissions != PermissionResult.PERMITTED and (
+                        key.acl and not key.acl.public_read
+                    ):
+                        return 403, {}, ""
+            elif signed_url and not authorized_request:
```
bblommers, line 1329: If we had rewritten the original if-statement, the full logic was:
`if not key and signed_url and not authorized_request`

If we indent the if-statement here, the logic changes to:
`if bucket and not key and signed_url and not authorized_request`

So in order to keep this logic the same, I think we should revert the indentation and change the logic to make it explicit:

```suggestion
        if not key and signed_url and not authorized_request:
```

Whether the bucket exists or not shouldn't matter here (I think...), so it shouldn't be inside the `if bucket`-statement.

(Unless I'm missing something? Feel free to push back if I've got something wrong in the logic!)

tests/test_s3/test_s3_bucket_policy.py
```diff
@@ -70,7 +70,10 @@ def test_block_or_allow_get_object(self, kwargs, boto3_status, unauthorized_stat
 
         assert requests.get(self.key_name).status_code == unauthorized_status
 
-    def test_block_put_object(self):
+    @pytest.mark.parametrize(
```
bblommers, line 73: Good call to parametrize this! Much better than writing a completely new test.

bblommers (changes requested):

moto/s3/responses.py
OwlInSpace: I indented it mainly because `key` is `None` when the bucket doesn't exist according to this:
```
try:
    key = self.backend.get_object(bucket_name, key_name)
    bucket = self.backend.get_bucket(bucket_name)
except S3ClientError:
    key = bucket = None
```
so I thought it would be better to skip the condition when the bucket is missing. I can reverse the indentation to make the logic the same as it was before, the difference is negligible I think, but I may be missing something

OwlInSpace: But, by not indenting it, wouldn't that put the possibility that it would return a 403 if the bucket doesn't exist?

bblommers: I've just tested this against AWS - they always throw a 403 if the headers are wrong/missing, regardless of whether the bucket exists or not. 

To be completely correct, we would have to move this entire if-statement to the top, and simplify it to: `if signed_url and not authorized_request and self._invalid_headers()`.
Only after that do we need to check if the bucket/key even exists.

But let's keep it as is for now - one improvement at the time :slightly_smiling_face:

bblommers (approved):
Thank you for fixing this @OwlInSpace!

PR #7842, opened 2024-07-11
Fix creation of random token in events model.

salvacorts:
wrt https://github.com/getmoto/moto/issues/3780#issuecomment-2222182309

I tested this patch with my localstack instance and now all works as expected. I.e. I no longer get this exception:
```
localstack-main   |   File "/opt/code/localstack/.venv/lib/python3.11/site-packages/moto/events/models.py", line 1036, in _gen_next_token
localstack-main   |     token = os.urandom(128).encode("base64")  # type: ignore
localstack-main   |             ^^^^^^^^^^^^^^^^^^^^^^
localstack-main   | AttributeError: 'bytes' object has no attribute 'encode'
```

### DIFF
--- a/moto/events/models.py
+++ b/moto/events/models.py
@@ -3,6 +3,7 @@ import json
 import os
 import re
 import sys
+import base64
 import warnings
 from collections import OrderedDict
 from enum import Enum, unique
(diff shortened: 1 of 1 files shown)


bblommers:
Hi @salvacorts, thank you for letting us know and taking the time to send in a PR!

I've decided to merge #7847 instead, where we got rid of the custom pagination/token logic altogether. This bug is now fixed as of `moto >= 5.0.12.dev14`.

PR #7856, opened 2024-07-15
DynamoDB uses <> for not equal

jbsilva:
Parse `<>` as `NOT EQUAL` in DynamoDB key condition expression.


bblommers (changes requested):
Hi @jbsilva, would you mind adding a test for this?

There are a few tests for KeyConditionExpression in `tests/test_dynamodb/test_dynamodb_table_with_range_key.py`, so AFAICT that's the best place to add another test for this usecase.

jbsilva:
Hi. I'm closing this PR as I noticed it's actually not correct.
`<>` is unfortunately not allowed in DynamoDB KeyConditionExpression (it is in FilterExpression).

PR #7857, opened 2024-07-15
Network Manager - Add Devices, Links & Sites

### LINKED ISSUE #7844: Network Manager: Add additional endpoints for Sites, Links & Devices
It would be useful to add support for the endpoints to Network Manager:
- Devices
- Links
- Sites

The actions would include **Create**, **Get** & **Delete** for each endpoint.

https://docs.aws.amazon.com/networkmanager/latest/APIReference/API_GetDevices.html
https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/networkmanager/client/get_sites.html

zkarpinski:
Resolves #7844 

Network Manager Endpoints:

- **Devices**: Create, Get, Delete
- **Links**: Create, Get, Delete
- **Sites**: Create, Get, Delete
- ListTagsForResource

https://docs.aws.amazon.com/networkmanager/latest/APIReference/API_GetDevices.html
https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/networkmanager/client/get_sites.html

### DIFF
--- a/moto/networkmanager/models.py
+++ b/moto/networkmanager/models.py
@@ -185,6 +185,58 @@ class Link(BaseModel):
         }
 
 
+class Device(BaseModel):
+    def __init__(
+        self,
+        account_id: str,
+        partition: str,
+        global_network_id: str,
+        aws_location: Optional[Dict[str, str]],
+        description: Optional[str],
+        type: Optional[str],
+        vendor: Optional[str],
+        model: Optional[str],
+        serial_number: Optional[str],
+        location: Optional[Dict[str, str]],
+        site_id: Optional[str],
+        tags: Optional[List[Dict[str, str]]],
+    ):
+        self.global_network_id = global_network_id
+        self.aws_location = aws_location
+        self.description = description
+        self.type = type
+        self.vendor = vendor
+        self.model = model
+        self.serial_number = serial_number
+        self.location = location
+        self.site_id = site_id
+        self.tags = tags or []
+        self.device_id = "device-" + "".join(mock_random.get_random_hex(18))
+        self.device_arn = (
+            f"arn:{partition}:networkmanager:{account_id}:device/{self.device_id}"
+        )
+        self.created_at = datetime.now(timezone.utc).strftime("%Y-%m-%dT%H:%M:%S.%fZ")
+        self.state = "PENDING"
+
+    def to_dict(self) -> Dict[str, Any]:
+        return {
+            "DeviceId": self.device_id,
+            "DeviceArn": self.device_arn,
+            "GlobalNetworkId": self.global_network_id,
+            "AWSLocation": self.aws_location,
+            "Description": self.description,
+            "Type": self.type,
+            "Vendor": self.vendor,
+            "Model": self.model,
+            "SerialNumber": self.serial_number,
+            "Location": self.location,
+            "SiteId": self.site_id,
+            "Tags": self.tags,
+            "State": self.state,
+            "CreatedAt": self.created_at,
+        }
+
+
 class NetworkManagerBackend(BaseBackend):
     """Implementation of NetworkManager APIs."""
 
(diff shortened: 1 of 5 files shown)


zkarpinski:
@bblommers I can't figure out why all my **Get** features aren't working as intended. The request only picks up the first element instead of the entire list. 

Per the documentation, the ids should be a list.
https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/networkmanager/client/get_sites.html

moto/networkmanager/responses.py
```diff
@@ -109,3 +109,150 @@ def describe_global_networks(self) -> str:
...
+        site = self.networkmanager_backend.delete_site(
+            global_network_id=global_network_id,
+            site_id=site_id,
+        )
+        return json.dumps(dict(Site=site.to_dict()))
+
+    def get_sites(self) -> str:
+        params = self._get_params()
+        global_network_id = unquote(self.path.split("/")[-2])
+        site_ids = params.get("siteIds")
```
bblommers, line 139: Looks like the `_get_params()` -method isn't quite smart enough to understand this usecase @zkarpinski.

You can directly get these values from the querystring, that will show all values:
```
self.querystring.get("siteIds")
```

zkarpinski: Thanks @bblommers ! I made the corrections.

moto/networkmanager/models.py
```diff
@@ -195,6 +360,186 @@ def describe_global_networks(
...
+    ) -> List[Link]:
+        # TODO: Implement filtering by site_id, type, provider
+        gn_links = self.links.get(global_network_id) or {}
+        queried = []
+        if not link_ids:
+            queried = list(gn_links.values())
+        elif isinstance(link_ids, List) and len(link_ids) == 1:
+            link_id = link_ids[0]
+            if link_id not in gn_links:
+                raise ResourceNotFound(link_id)
```
bblommers, line 458: Is this error only raised when a single `link_id` is supplied? I'd would expect the logic to be the same regardless of how many `link_ids` are supplied.

There are three possible logic paths here of course, I don't know what AWS does:
 - error only if none of the provided `link_ids` are correct
 - error if a single link is found, and show that ID
 - error when some links are not found, and supply all `link_ids` that are not found

tests/test_networkmanager/test_networkmanager.py
```diff
@@ -191,3 +205,331 @@ def test_describe_global_networks():
...
+    site_ids = []
+    for i in range(NUM_SITES):
+        site_id = client.create_site(
+            GlobalNetworkId=gn_id,
+            Description="Test site #{i}",
+        )["Site"]["SiteId"]
+        site_ids.append(site_id)
+    sites_to_get = site_ids[0:NUM_TO_TEST]
+    resp = client.get_sites(GlobalNetworkId=gn_id, SiteIds=sites_to_get)["Sites"]
+    assert len(resp) == NUM_TO_TEST  # Fix this - Fails if greater than 1
```
bblommers, line 258: Bit of a nitpick - but these comments (and the ones below) can be removed now, I think?

bblommers (changes requested):
Hi @zkarpinski, thanks for the PR! This looks OK overall - just a few minor comments.

moto/networkmanager/models.py
```diff
@@ -24,6 +24,24 @@
         "limit_default": 100,
         "unique_attribute": "core_network_arn",
     },
+    "get_sites": {
```
bblommers, line 27: Good to see pagination supported from the start!

moto/networkmanager/models.py
zkarpinski: Tested it, it'll return an empty list. No error. Fixing logic

zkarpinski:
@bblommers Resolved the feedback & updated exception logic.

bblommers (approved):
LGTM!

PR #7859, opened 2024-07-15
ACM: Enforce ExportCertificate constraints, and issue Private CA certificates via RequestCertificate

### LINKED ISSUE #7858: ACM: Export Certificate permits exporting unexportable certificates
boto3: 1.34.144 
moto: 4.1.0

The current implementation of ExportCertificate permits one to export imported or amazon issued certificates.
This is contrary to Amazon's documentation and the reality of their implementation, which only permits exporting Private CA certificates:
https://docs.aws.amazon.com/acm/latest/userguide/export-private.html

```
import boto3
from moto import mock_aws

mock = mock_aws()
mock.start()

acm = boto3.client("acm", region_name="us-east-1")

request = acm.request_certificate(DomainName="example.com")

cert_arn = request["CertificateArn"]

# This should fail with a validation error as cert_arn is not a private certificate
export = acm.export_certificate(
            CertificateArn=cert_arn,
            Passphrase="test"
         )
````

Tookmund:
Closes #7858 

This PR checks the constraints on ExportCertificate as Amazon does, and only allows exporting private certificates.

Additionally, this enables creating private certificates via the RequestCertificate call.

Ideally those certificates would also enter the ISSUED state before they could be exported, but this is a start to begin supporting ACM/PCA interactions correctly.

### DIFF
--- a/moto/acm/models.py
+++ b/moto/acm/models.py
@@ -213,14 +213,10 @@ class CertBundle(BaseModel):
             encryption_algorithm=serialization.NoEncryption(),
         )
 
-        cert_type = "AMAZON_ISSUED"
-        if cert_auth_arn is not None:
-            cert_type = "PRIVATE"
-
         return cls(
             certificate=cert_armored,
             private_key=private_key,
-            cert_type=cert_type,
+            cert_type="PRIVATE" if cert_authority_arn is not None else "AMAZON_ISSUED",
             cert_status="PENDING_VALIDATION",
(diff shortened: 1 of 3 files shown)


Tookmund:
Apologies missed the contributing guide previously, will reformat this.
Not sure if the project would prefer this as one/two squashed commits, but I can do that if wanted.

bblommers (approved):
This is great @Tookmund - thank you for improving Moto and bringing us closer to parity with AWS!

PR #7865, opened 2024-07-17
IoT: Set Certificate ID to SHA256 hexdigest of DER.

mcli:
Previous algorithm was based on the PEM format and produced an incorrect certificate ID.
Includes configurable certificate validity for backwards compatibility.

### DIFF
--- a/docs/docs/configuration/index.rst
+++ b/docs/docs/configuration/index.rst
@@ -23,6 +23,7 @@ If you are using the decorators, some options are configurable within the decora
         },
         "iam": {"load_aws_managed_policies": False},
         "stepfunctions": {"execute_state_machine": True},
+        "iot": {"use_valid_cert": True},
     })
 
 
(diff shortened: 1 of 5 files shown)


mcli:
This is an spin of https://github.com/getmoto/moto/pull/6858 which was closed.  Make the default to generate the correct cert ID and allow tests to optionally configure to use previous algorithm which supported invalid certificates in the test.

mcli:
Fixes https://github.com/getmoto/moto/issues/7866

bblommers (changes requested):
Hi @mcli, thanks for picking this up. The change makes sense to me - but we would have to keep the current (incorrect) behaviour as the default, and make the DER calculation the opt-in behaviour. We don't want to introduce a breaking change for users.

mcli:
> Hi @mcli, thanks for picking this up. The change makes sense to me - but we would have to keep the current (incorrect) behaviour as the default, and make the DER calculation the opt-in behaviour. We don't want to introduce a breaking change for users.

Hi @bblommers , Ok.  Set the default to `False`.  Thanks for the comments and I agree with the approach.  Propose to change the default in the next major release.

bblommers (approved):

PR #7867, opened 2024-07-18
Route53 list tags for resources

armichaud:
Implements Route53.Client.list_tags_for_resources API

### DIFF
--- a/moto/route53/urls.py
+++ b/moto/route53/urls.py
@@ -9,17 +9,10 @@ from .responses import Route53
 url_bases = [r"https?://route53(\..+)?\.amazonaws.com"]
 
 
-def tag_response1(request: Any, full_url: str, headers: Any) -> TYPE_RESPONSE:
+def tag_response(request: Any, full_url: str, headers: Any) -> TYPE_RESPONSE:
     return Route53().list_or_change_tags_for_resource_request(
         request, full_url, headers
     )
-
-
-def tag_response2(request: Any, full_url: str, headers: Any) -> TYPE_RESPONSE:
-    return Route53().list_tags_for_resources_request(
-        request, full_url, headers
-    )
-
 
 url_paths = {
     r"{0}/(?P<api_version>[\d_-]+)/hostedzone$": Route53.dispatch,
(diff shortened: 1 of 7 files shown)


armichaud:
Hey @bblommers , it looks like my unit test is failing when trying to call `ChangeTagsForResource`. This is definitely a valid boto3 API, and the test is passing when I run it locally. Are there particularities to the remote test runner that I've missed?

moto/route53/urls.py
```diff
@@ -36,8 +30,10 @@ def tag_response2(request: Any, full_url: str, headers: Any) -> TYPE_RESPONSE:
     r"{0}/(?P<api_version>[\d_-]+)/healthcheck$": Route53.dispatch,
     r"{0}/(?P<api_version>[\d_-]+)/healthcheck/(?P<health_check_id>[^/]+)$": Route53.dispatch,
     r"{0}/(?P<api_version>[\d_-]+)/healthcheck/(?P<health_check_id>[^/]+)/status$": Route53.dispatch,
-    r"{0}/(?P<api_version>[\d_-]+)/tags/healthcheck/(?P<zone_id>[^/]+)$": tag_response1,
-    r"{0}/(?P<api_version>[\d_-]+)/tags/hostedzone/(?P<zone_id>[^/]+)$": tag_response2,
+    r"{0}/(?P<api_version>[\d_-]+)/tags/healthcheck/(?P<zone_id>[^/]+)$": tag_response,
+    r"{0}/(?P<api_version>[\d_-]+)/tags/hostedzone/(?P<zone_id>[^/]+)$": tag_response,
```
bblommers, line 34:
```suggestion
    r"{0}/(?P<api_version>[\d_-]+)/tags/hostedzone/(?P<zone_id>.+)$": tag_response,
```
I believe this is the reason the test is currently failing in ServerMode. The `zone_id` contains url-encoded slashes (`%2Fhostedzone%2Fzone_id`), but I guess the matcher does some un-encoding before trying to match the incoming URL - so this regex needs to accept a `zone_id` that can contain slashes.

bblommers (approved):
LGTM - thank you for implementing this @armichaud!

moto/route53/urls.py
armichaud: Gotcha, thanks @bblommers for catching this and making the change! One question though – wouldn't this also then apply to the `/tags/healthcheck` url as well?

PR #7891, opened 2024-07-25
AWS Transfer Family

### LINKED ISSUE #7790: Plans to add support for the Transfer client?
Thank you for the great work on this package! Are there any plans to support the [Transfer client](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/transfer.html), particulalry these methods:

- [create_user](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/transfer/client/create_user.html) and [describe_user](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/transfer/client/describe_user.html) and [delete_user](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/transfer/client/delete_user.html); and
- [import_ssh_public_key](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/transfer/client/import_ssh_public_key.html) and [delete_ssh_public_key](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/transfer/client/delete_ssh_public_key.html).

If not, then what’s your recommendation to handle such non-moto supported services?

armichaud:
Add Transfer client APIs requested in #7790 

Implements create_user, describe_user, delete_user, import_ssh_public_key, delete_ssh_public_key, create_server, describe_server, delete_server


armichaud:
@bblommers this should be good to go

jenstroeger:
> Implements create_user, describe_user, delete_user, import_ssh_public_key, delete_ssh_public_key

I think we also need [create_server()](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/transfer/client/create_server.html) and [delete_server()](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/transfer/client/delete_server.html) in order to use the user related functions. Otherwise, user requests return a
```
botocore.exceptions.ClientError: An error occurred (ServerNotFound) when calling the DescribeUser operation: There are no transfer protocol-enabled servers with ID [...]
```
where the server id was an incorrectly hardwired value in current tests.

moto/transfer/models.py
```diff
@@ -0,0 +1,181 @@
...
+    def describe_server(self, server_id):
+        if server_id in self.servers:
+            ServerNotFound(server_id=server_id)
+        return self.servers[server_id]
+
+    def delete_server(self, server_id):
+        if server_id in self.servers:
+            ServerNotFound(server_id=server_id)
+        del self.servers[server_id]
+        return
```
jenstroeger, line 86: Shouldn’t this be `not in`?

jenstroeger (commented):
Wonderful, thank you! I’ll play around with it today 👍🏼

armichaud:
@jenstroeger ready for rereview

jenstroeger:
> @jenstroeger ready for rereview

Great, thank you! I’ll test this change in more detail tomorrow 👍🏼

armichaud:
> > @jenstroeger ready for rereview
> 
> Great, thank you! I’ll test this change in more detail tomorrow 👍🏼

Linting errors resolved, ready for review

armichaud:
@bblommers Not sure what's up with `test_cloudformation_stack_integration.py::test_lambda_function`

moto/transfer/types.py
```diff
@@ -0,0 +1,185 @@
...
+    LOGICAL = "LOGICAL"
+
+
+class UserHomeDirectoryMappingType(str, Enum):
+    FILE = "FILE"
+    DIRECTORY = "DIRECTORY"
+
+
+class UserHomeDirectoryMapping(TypedDict):
+    Entry: str
```
bblommers, line 20: These properties (from all the TypedDict's _and_ from @dataclasses) should be snake-cased, to be in-line with the rest of the code base. So `entry`, `target`, etc. 

`type` is a keyword, so our linter may complain if you use that as variable name, but that can just become something like `mapping_type`.

bblommers (changes requested):
Hi @armichaud! Having TypedDicts and dataclasses to store the models is a good idea, that makes for some clean code. Just a minor change request regarding the variable names.

Don't worry about the failing test - it's a known flake, so it will probably pass next time.

armichaud:
@bblommers I have refactored everything such that the backend converts everything to snake_case and back to camel case. 

Just a bit of feedback on that pattern: I understand it's the convention for the repo, but I'm not sure what it adds besides boilerplate. 

Moreover, the linter is super aggressive, and at times I wasn't sure what its rules really accomplished. For example, it threw type errors at things like this:

```python
server = {...}
if self.attr is not None:
    server["attr"] = self.attr 
```

but had no problem with this:

```python
attr = {}
if self.attr is not None:
    attr = self.attr
server = {..., "attr": attr}
```

The volume of linting issues made it just too annoying to work with TypedDicts, so I scrapped them and just made most of the type hints generic.

bblommers (approved):
This looks great - thank you for adding this to Moto @armichaud!

The snake case is indeed just convention, but does make development easier when switching between different services - there's less of a mental effort in trying to understand the code. It's also easier to read IMO, but I understand that's a personal preference.

The typing - that's indeed a bit of a pain. In that particularly example, I've learned that `mypy` likes`self.attr = attr or {}` best, but that is just experience. Overall, `mypy` has caught (and still does catch) quite a few number of bugs, so it is worth it IMO.

jenstroeger:
Thank you @armichaud, I’ll test more tomorrow! 

> To show our thanks, we'd like to share some of the donations that we've received with you. PR's like this are the big reason that Moto is as successful as it is - so it's only fair that you, as a contributor, get to share the spoils.

That’s a great idea, well deserved!

armichaud:
> This looks great - thank you for adding this to Moto @armichaud!
> 
> The snake case is indeed just convention, but does make development easier when switching between different services - there's less of a mental effort in trying to understand the code. It's also easier to read IMO, but I understand that's a personal preference.
> 
> The typing - that's indeed a bit of a pain. In that particularly example, I've learned that `mypy` likes`self.attr = attr or {}` best, but that is just experience. Overall, `mypy` has caught (and still does catch) quite a few number of bugs, so it is worth it IMO.

Makes sense! Consistency is definitely valuable, and I agree that the headache with linting is probably worth the bugs that it catches. I suspect that there are reasons for those preferences that have to do with the subtleties of typing in python that I'm just not as familiar with.

jenstroeger:
I started testing and one issue I came across so far:
```diff
diff --git a/moto/transfer/models.py b/moto/transfer/models.py
index 360b82549..ca9f5e69e 100644
--- a/moto/transfer/models.py
+++ b/moto/transfer/models.py
@@ -131,7 +131,7 @@ class TransferBackend(BaseBackend):
         self,
         home_directory: Optional[str],
         home_directory_type: Optional[UserHomeDirectoryType],
-        home_directory_mappings: List[Dict[str, Optional[str]]],
+        home_directory_mappings: Optional[List[Dict[str, Optional[str]]]],
         policy: Optional[str],
         posix_profile: Optional[Dict[str, Any]],
         role: str,
@@ -150,7 +150,7 @@ class TransferBackend(BaseBackend):
             tags=(tags or []),
             user_name=user_name,
         )
-        if len(home_directory_mappings) > 0:
+        if home_directory_mappings:
             for mapping in home_directory_mappings:
                 user.home_directory_mappings.append(
                     {
```
For [create_user()](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/transfer/client/create_user.html) that argument is optional (because it’s not labeled “[REQUIRED]” like e.g. the `Role` argument).

PR #7897, opened 2024-07-26
docs: Add note to Multi-Account support

mbressan-varmour:
I had an issue when thinking that two roles created using this approach would not see eachother's bucket.

I think this note could be important for someone else.


bblommers (approved):
Good catch @mbressan-varmour - thank you for improving our documentation!

PR #7913, opened 2024-08-01
Allow non-default partitions in events and resourcegroups regexes

dfangl:
## Motivation
With #7716 there were already some changes to support arns with `aws-us-gov` and `aws-cn` (for example) partitions in validation regexes.

However, it seems two were missed, in `events` and `resourcegroups`.

## Changes
* Use the `ARN_PARTITION_REGEX` to make two validation regex compatible with non-default partitions.


bblommers (approved):
Great - thank you for following up and fixing this @dfangl!

PR #7923, opened 2024-08-02
CognitoIDP: initiate_auth with USERNAME_PASSWORD_AUTH and SMS/Software Token MFA

wazy:
The initiate_auth call with a AuthFlow of USERNAME_PASSWORD_AUTH was not accounting for SMS MFA and was returning a AuthenticationResult instead of a Challenge as expected.

-- EDIT --
It also was not accounting for software token MFA and not returning a Challenge as expected.

### DIFF
--- a/moto/cognitoidp/models.py
+++ b/moto/cognitoidp/models.py
@@ -1945,6 +1945,13 @@ class CognitoIdpBackend(BaseBackend):
                     "Session": session,
                 }
 
+            if user.sms_mfa_enabled and user.preferred_mfa_setting == "SMS_MFA":
+                return {
+                    "ChallengeName": "SMS_MFA",
+                    "ChallengeParameters": {},
+                    "Session": session,
+                }
+
             access_token, expires_in = user_pool.create_access_token(
                 client_id, username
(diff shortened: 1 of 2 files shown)


bblommers (changes requested):
Hi @wazy, can you extend the test to also `respond_to_auth_challenge`? Just to confirm it is still possible to complete an authentication flow (i.e. get an AccessToken/IdToken) in this scenario.

wazy:
@bblommers added respond_to_auth_challenge in the test as requested :+1:

wazy:
@bblommers I noticed that software token mfa was also not being challenged with USER_PASSWORD_AUTH so went and added that code and a test with respond_to_auth_challenge as well.

wazy:
Fixed test `test_initiate_auth_USER_PASSWORD_AUTH` which had buggy behavior. It should not have any MFA enabled and preferred if it expects AuthenticationResult from initiate_auth and not a challenge.

bblommers (approved):
Great - thank you for fixing this, and welcome to Moto @wazy!

PR #7924, opened 2024-08-02
Feature: KMS MultiRegionConfiguration

Singha22:
Adds support for KMS to add `MultiRegionConfiguration`support when `MultiRegion = True`. 
- Users can view their replica keys and primary keys 
- Support to prevent users from creating replica's of keys in already existing region


moto/kms/models.py
```diff
@@ -263,6 +275,7 @@ def __init__(self, region_name: str, account_id: Optional[str] = None):
         super().__init__(region_name=region_name, account_id=account_id)  # type: ignore
         self.keys: Dict[str, Key] = {}
         self.key_to_aliases: Dict[str, Set[str]] = defaultdict(set)
+        self.multi_region_configuration: dict = {"MultiRegionConfiguration": {}}
```
bblommers, line 278: This doesn't seem to be used anywhere? The value is set in the `replicate_key`-method, but it's not read anywhere afterwards.

moto/kms/models.py
```diff
@@ -348,6 +379,16 @@ def describe_key(self, key_id: str) -> Key:
...
         if r"alias/" in str(key_id).lower():
             key_id = self.get_key_id_from_alias(key_id)  # type: ignore[assignment]
+
+        key = self.keys[self.get_key_id(key_id)]
+
+        if key.multi_region:
+            if key.arn == key.multi_region_configuration["PrimaryKey"]["Arn"]:
+                return self.keys[self.get_key_id(key_id)]
+            else:
+                key.multi_region_configuration["MultiRegionKeyType"] = "REPLICA"
```
bblommers, line 389: Can we move this to the `replicate_key`-method? That's where we decide that a given key is a replica, so it makes sense to me to set the KeyType as well.

bblommers (changes requested):

moto/kms/models.py
bblommers: Ah, yes, I understand - it makes sense as it is then.

PR #7925, opened 2024-08-02
Consolidate `Neptune` and `RDS` backends

bpandola:
Despite being documented as two distinct services (with two distinct boto3 clients), the Neptune and RDS services share a single AWS backend (RDS).  This is reflected in the botocore service definition for Neptune, which lists `rds` as its endpoint.  Calling the `DescribeDBClusters` API method with either the `neptune` or `rds` boto3 client results in the exact same list of clusters--no filtering is done on the AWS backend.  Neptune clusters are commingled with Aurora clusters.

@bblommers I think this makes more sense than trying to separate these services (and it prevents a lot of code duplication), but I want to get your opinion since you originally implemented/split out the Neptune service. (As I continue to work on improving the RDS backend overall, I do plan to handle some of the `if self.engine == "neptune` conditionals in a more elegant way.)


bblommers:
@bpandola It is much cleaner like this, with RDS handling everything. If I remember correctly, I did try this approach first, but ran into some problems differentiating when a call came from Neptune and when from RDS. I don't remember exactly what went wrong, but based on this PR, it does work, so I'm more than happy to go ahead with this approach.

moto/rds/models.py
```diff
@@ -1651,24 +1662,17 @@ def __init__(self, region_name: str, account_id: str):
...
 
-    @property
-    def neptune(self) -> NeptuneBackend:
-        return neptune_backends[self.account_id][self.region_name]
+    def db_cluster_options(self, engine) -> List[Dict[str, Any]]:  # type: ignore
+        from moto.rds.utils import decode_orderable_db_instance
 
-    @property
-    def db_cluster_options(self) -> List[Dict[str, Any]]:  # type: ignore
-        if self._db_cluster_options is None:
```
bblommers, line 1663: Can we revert this part? I saw the commit where you had to revert the `@cache`-property, but that does mean we'd need our manual cache back - this is a very expensive operation, IIRC

bblommers (approved):
Aside from the cache note, this looks good to me.

We would need to find a different approach for the implementation coverage script, to document that RDS coverage == Neptune coverage. 
That doesn't have to be now, or even as part of this PR, just noting it as an FYI.

moto/rds/models.py
bpandola: Done.  The `cache` decorator is not available in 3.8, but `lru_cache` is, and they are functionally equivalent for our purposes here.

bpandola:
@bblommers 
> We would need to find a different approach for the implementation coverage script, to document that RDS coverage == Neptune coverage. That doesn't have to be now, or even as part of this PR, just noting it as an FYI.

Good catch.  I did push an additional commit to fix the implementation coverage script (it was failing before).  I think it should work fine now, with no additional work required.

PR #7944, opened 2024-08-07
[EMR] Put and Get BlockPublicAccessConfiguration

armichaud:


armichaud:
@bblommers do you have a sense of what might be causing this test error?
```
>           raise TypeError(f"Unknown type {vtype}")
E           TypeError: Unknown type structure
moto/core/responses.py:1145: TypeError
```

It seems like the XML template defined for the GetBlockPublicAccessConfiguration response is fundamentally similar to the other EMR responses, and it also seems like there's nothing special about how the API's backend is invoked.

bblommers:
Hi @armichaud, Moto automatically tries to parse the incoming request, and that is failing. More specifically, the AWS spec says that the request should be an empty dictionary (https://github.com/boto/botocore/blob/7f23afdfdbcab57e319a8d86622b4f0216fc7244/botocore/data/emr/2009-03-31/service-2.json#L2192) - but Moto doesn't know how to handle that (yet!).

You can fix this by adding another elif to `moto.core.responses.py::to_str`:

```
--- a/moto/core/responses.py
+++ b/moto/core/responses.py
@@ -1141,6 +1141,8 @@ def to_str(value: Any, spec: Dict[str, Any]) -> str:
         return str(value)
     elif value is None:
         return "null"
+    elif vtype == "structure":
+        return ""
     else:
         raise TypeError(f"Unknown type {vtype}")
```

moto/emr/responses.py
```diff
@@ -1418,3 +1457,36 @@ def remove_auto_scaling_policy(self) -> str:
...
+            <BlockPublicSecurityGroupRuleRange>
+              <MinRange>{{rule_range['min_range']}}</MinRange>
+              <MaxRange>{{rule_range['max_range']}}</MaxRange>      
+            </BlockPublicSecurityGroupRuleRange>
+          {% endfor %}
+        </BlockPublicSecurityGroupRuleRanges>
+      </BlockPublicAccessConfiguration>
+      <BlockPublicAccessConfigurationMetadata>
+        <CreationDateTime>{{creation_date_time}}</CreationDateTime>
+        <CreatedByArn>{{created_by_arn}}</CreationByArn>
```
bblommers, line 1485: This XML is mismatched - that's the next error after fixing the `Unknown type structure` error :slightly_smiling_face:

armichaud:
@bblommers thanks for the quick response! This is ready for review, but I wanted to run one thing by you: ideally, the CreatedByArn would return the Arn for the IAM user that created the configuration, but I'm not sure if that's possible. Is there an equivalent of `STS.client.get_caller_identity` that can be called from within a boto model?

armichaud:
@bblommers ready for final review

bblommers (approved):
LGTM - thank you @armichaud!

PR #7953, opened 2024-08-09
Fix issues for non-default partitions in elastictranscoder, sagemaker, and SSM

dfangl:
## Motivation
During extensive testing, I found some more errors when accepting / generating ARNs for partitions other than `aws` in moto.

## Changes
* Accept role ARNs from non-default partitions in elastictranscoder
* Accept cluster ARNs from non-default partitions in sagemaker
* Properly generate the SSM prefix, to strip it correctly in the `get_parameter` operation.


bblommers (approved):
Love it - thanks @dfangl!

PR #7958, opened 2024-08-09
ACM-PCA: Ensure backend is serialisable

viren-nadkarni:
## Background

Changes introduced in https://github.com/getmoto/moto/pull/7950 make the x509 Issuer a part of the `CertificateAuthority` model. Firstly, it is not necesary to retain this attribute in-memory. Secondly, because the x509 Issuer object is natively implemented in Rust, it cannot be serialised with pickle which breaks certain functionality in LocalStack.

## Changes

This PR changes the issuer into a class property which builds the x509 Issuer object on-demand.


bblommers (approved):
LGTM!

PR #7959, opened 2024-08-09
CloudFormation: create_change_set ChangeSetName more rules

tinngo365:
### Pull Request: Refactor Validation in `create_change_set`

**Summary:**

This PR refactors the `create_change_set` function to improve validation by centralizing it in a new `validate_create_change_set` function. This change ensures all parameters are validated consistently and makes it easier to extend validation in the future.

**Changes:**

- **Added `validate_create_change_set` Function**:
  - Centralizes validation for all parameters, currently including `change_set_name` rules (e.g., must start with a letter, be alphanumeric, and under 128 characters).

- **Updated `create_change_set`**:
  - Calls `validate_create_change_set` to validate inputs before processing.

**Testing:**

- Updated unit tests to cover validation scenarios.

[Documentation](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/cloudformation/client/create_change_set.html#:~:text=A%20change%20set%20name%20can%20contain%20only%20alphanumeric%2C%20case%20sensitive%20characters%2C%20and%20hyphens.%20It%20must%20start%20with%20an%20alphabetical%20character%20and%20can%E2%80%99t%20exceed%20128%20characters.)

image

### DIFF
--- a/moto/cloudformation/utils.py
+++ b/moto/cloudformation/utils.py
@@ -1,6 +1,6 @@
 import os
 import string
-from typing import Any, Dict, List, Optional
+from typing import Any, List
 from urllib.parse import urlparse
 
 import yaml
@@ -138,17 +138,7 @@ def get_stack_from_s3_url(template_url: str, account_id: str, partition: str) ->
     return key.value.decode("utf-8")  # type: ignore[union-attr]
 
 
-def validate_create_change_set(
-    stack_name: str,
-    change_set_name: str,
-    template: str,
-    parameters: Dict[str, str],
-    description: str,
-    change_set_type: str,
-    notification_arns: Optional[List[str]] = None,
-    tags: Optional[Dict[str, str]] = None,
-    role_arn: Optional[str] = None,
-) -> Any:
+def validate_create_change_set(change_set_name: str) -> None:
     if not (change_set_name and change_set_name[0].isalpha()):
         raise ValidationError(f"Invalid change set name: {change_set_name}")
 
--- a/moto/cloudformation/utils.py
+++ b/moto/cloudformation/utils.py
@@ -1,12 +1,14 @@
 import os
 import string
-from typing import Any, List
+from typing import Any, Dict, List, Optional
 from urllib.parse import urlparse
 
 import yaml
 
 from moto.moto_api._internal import mock_random as random
 from moto.utilities.utils import get_partition
+
+from .exceptions import ValidationError
 
 
 def generate_stack_id(stack_name: str, region: str, account: str) -> str:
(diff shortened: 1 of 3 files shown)


tests/test_cloudformation/test_cloudformation_stack_crud_boto3.py
```diff
@@ -1067,15 +1067,15 @@ def test_get_template_summary_for_stack_created_by_changeset_execution():
     conn.create_change_set(
         StackName="stack_from_changeset",
         TemplateBody=json.dumps(dummy_template3),
-        ChangeSetName="test_changeset",
```
tinngo365, line 1070: Had to change this to follow the new rules

tinngo365:
@bblommers could you explain what is wrong?

moto/cloudformation/utils.py
```diff
@@ -134,3 +136,29 @@ def get_stack_from_s3_url(template_url: str, account_id: str, partition: str) ->
 
     key = s3_backends[account_id][partition].get_object(bucket_name, key_name)
     return key.value.decode("utf-8")  # type: ignore[union-attr]
+
+
+def validate_create_change_set(
+    stack_name: str,
```
bblommers, line 142: Can we change this here to only include the arguments that we're actually using?
The less arguments, the easier it is to read - and we can always change the signature later

moto/cloudformation/utils.py
```diff
@@ -134,3 +136,29 @@ def get_stack_from_s3_url(template_url: str, account_id: str, partition: str) ->
...
+    change_set_name: str,
+    template: str,
+    parameters: Dict[str, str],
+    description: str,
+    change_set_type: str,
+    notification_arns: Optional[List[str]] = None,
+    tags: Optional[Dict[str, str]] = None,
+    role_arn: Optional[str] = None,
+) -> Any:
```
bblommers, line 151: Minor - but we can just simplify this to return `None`

tests/test_cloudformation/test_cloudformation_stack_crud_boto3.py
```diff
@@ -2462,6 +2462,65 @@ def test_create_and_update_stack_with_unknown_resource():
             )
 
 
+@mock_aws
+def test_valid_change_set_name():
```
bblommers, line 2466: I don't think this test is necessary? There are other tests that create a changeset, so they verify implicitly that the name is valid

bblommers (changes requested):

tinngo365:
Appreciate the feedback! Thank you

bblommers (approved):
LGTM - thank you for adding this to Moto @tinngo365!

PR #7970, opened 2024-08-12
Admin: GZIP all JSON resources to reduce filesize

### LINKED ISSUE #7909: Large size on disk when installing for just one service
This isn't really a bug, but more just unexpected behavior for me.

`moto` is a relatively large dependency on-disk, installed it's around 58MB for me. A huge portion of that seems to be big json files that define things like AMI's available for every region, all ec2 types, etc.

We only use `moto` for s3, and when I saw there was a way to install `moto[s3]` I assumed that also meant I wouldn't be downloading big json files for all the other AWS services `moto` supports mocking out. 

Is there any way to exclude data for other services? `moto` is now the largest dependency of our pretty heavy monolithic application. `botocore` has a [similar issue](https://github.com/boto/botocore/issues/1543), but currently it's down to 17MB on disk.

bblommers:
Fixes #7909 

- Changes all existing `.json` files to `.json.gz`
- Changes all necessary touch points to load the new `.json.gz` files
- Changes the `load_resource`-method to automatically decompress `.json.gz` files
- Update all existing scripts that pull data from AWS to compress data first
- Adds the missing CI action to also pull ECS optimized AMI's for SSM

After the dev-release is out I can post some stats on pre/post size, both download and on-disk.

@bpandola Have I missed anything?

@viren-nadkarni FYI. Are you aware of any touch points in LS where this may break things, i.e. other dependencies on these JSON files?


bpandola:
With all due respect @bblommers, I think this is the wrong approach.  I think the files should only be zipped for the distribution, but should remain human readable/editable in the source tree.  This is what the `botocore` team did and I think the reasons should be clear.  If need be, I can elaborate further; but I feel pretty strongly that 1) shrinking the size of the distribution is a nice-to-have and not a need-to-have, and 2) doing so should not be destructive to the `moto` source.

Unfortunately--unless I'm just not seeing it--the release/distribution job for `botocore` doesn't appear to be in their public repo, so we can't just lift-and-shift; but the commit to support gzip'd models is fairly minor and I believe what we should be aiming for here: https://github.com/boto/botocore/commit/47fbea3e66b4e0e6d8d4877c2681a1b86eab3eda

bblommers:
Hey @bpandola, that's a very good point -  I hadn't considered that approach.

Setup tools has the option to add custom hooks, maybe we can use those to compress at build time - that way we keep readability at source and reproducible builds.

PR #7971, opened 2024-08-12
Utility to iterate over flattened backends

giograno:
This PR adds a utility method to iterate over a flattened view of a `BackendDict`.

Each record is a tuple of account ID, region, and the corresponding `BaseBackend`.


bblommers (approved):
LGTM!

PR #7972, opened 2024-08-12
Quantum Ledger Database

armichaud:
Implements

- create_ledger
- update_ledger
- delete_ledger
- describe_ledger
- tag_resource
- list_tags_for_resource


armichaud:
@bblommers Any idea why the test thinks that CreateLedger is undefined here?

moto/qldb/urls.py
```diff
@@ -0,0 +1,11 @@
+"""qldb base URL and path."""
+
+from .responses import QLDBResponse
+
+url_bases = [
+    r"https?://qldb\.(.+)\.amazonaws\.com",
+]
+
+url_paths = {
+    "0/.*$": QLDBResponse.dispatch,
```
bblommers, line 10:
```suggestion
    "{0}/.*$": QLDBResponse.dispatch,
```

Typo - the key is converted into a regex to match the full url of the incoming request, with `{0}` acting as a substitute for the host (`qldb.amazonaws.com`), but it will not match anything as a literal `0`.

I figured this out by debugging the `BotocoreStubber` 
https://github.com/getmoto/moto/blob/52d2f06800f5c1db8b5fdb040e6537556ff5b45e/moto/core/botocore_stubber.py#L45

All requests are handled by that class, so debugging (or print-statements) can sometimes very quickly tell you why a request is not routed correctly. (In decorator-mode at least - server mode is a different story..)

armichaud: Looking back, it seems like this was generated by the scaffold command. Worthy of an issue?

armichaud:
@bblommers ready for review

moto/qldb/urls.py
bblommers: I raised https://github.com/getmoto/moto/pull/7981 to fix the scaffolding. Thank you for pointing that out!

bblommers:
@armichaud I pushed a fix for the failing ServerMode tests - the URL's needed to be explicitly specified. 

#7981 Also improves that area, so it now automatically adds the correct URL's for most scenarios.

bblommers (approved):
Thank you for the PR! Great to see so many and meaningful contributions to Moto 🥇

PR #7973, opened 2024-08-12
Batch: add support for parameters in Job commands

pinzon:
This PR adds support for using parameters in job commands. It also implements the ability to overwrite parameters during the submit job operation.


bblommers (changes requested):
Hi @pinzon, left some comments. Note that the tests are also failing atm.

bblommers (approved):
Awesome - thank you @pinzon!

PR #7978, opened 2024-08-14
feat: Deal with APIGateway GetAPI with an empty string

juanitosvq:
Fixes https://github.com/getmoto/moto/issues/7974

AWS returns a 200 empty response when doing a `get_api` with an empty string in `apigatewayv2`. Same behaviour is seen with `get_rest_api` in `apigateway`. `boto` is returning a 404 `not yet implemented` error. This PR tries to mock that use case.

```
import boto3
import json

# apigateway.get_rest_api
client = boto3.session.Session(profile_name="dev", region_name="us-east-1").client("apigateway")
resp = client.get_rest_api(restApiId="")

print(json.dumps(resp, default=str))

# apigatewayv2.get_api
client = boto3.session.Session(profile_name="dev", region_name="us-east-1").client("apigatewayv2")
resp = client.get_api(ApiId="")

print(json.dumps(resp, default=str))
```

Both returns:
```
{
    "ResponseMetadata": {
        "RequestId": "e156a293-3eff-4b81-be31-6712e05c843f",
        "HTTPStatusCode": 200,
        "HTTPHeaders": {
            "date": "Wed, 14 Aug 2024 14:11:18 GMT",
            "content-type": "application/json",
            "content-length": "14959",
            "connection": "keep-alive",
            "x-amzn-requestid": "e156a293-3eff-4b81-be31-6712e05c843f",
            "x-amz-apigw-id": "cgIpBINbIAMEcoA="
        },
        "RetryAttempts": 0
    }
}
```

As discussed in https://github.com/getmoto/moto/issues/7974#issuecomment-2288226432, this seems an outlier in the way AWS normally behaves, so it was decided we could have a custom handler for this use case, instead of trying to make it work with the generic dispatcher architecture used everywhere else.

Feedback is more than welcome!

### DIFF
--- a/moto/apigateway/responses.py
+++ b/moto/apigateway/responses.py
@@ -1,5 +1,5 @@
 import json
-from typing import Dict, List
+from typing import Any, Dict, List
 from urllib.parse import unquote
 
 from moto.core.responses import TYPE_RESPONSE, BaseResponse
@@ -122,7 +122,7 @@ class APIGatewayResponse(BaseResponse):
         return 200, {}, json.dumps(rest_api.to_dict())
 
     @staticmethod
-    def get_rest_api_without_id(*args) -> TYPE_RESPONSE:
+    def get_rest_api_without_id(*args: Any) -> TYPE_RESPONSE:
         return 200, {}, "{}"
 
     def put_rest_api(self) -> TYPE_RESPONSE:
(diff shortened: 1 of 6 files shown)


bblommers (approved):
Great - thank you for fixing this @juanitosvq!

PR #7984, opened 2024-08-15
Admin: Support Py 3.13

bblommers:
The SES changes are due to a logic bug that only manifests in Python3.13.

When no CC/BCC fields are set, the `fieldvalues` would be set to something like: `['to@example.com, foo@example.com', '', '']`
Sending that to the builtin `emails.utils.getaddresses`-method in Python 3.12 results in:
`[('', 'to@example.com'), ('', 'foo@example.com'), ('', '')]`

But in Python 3.13, this results in an empty list.

This PR ensures that there are no empty strings in `fieldvalues`, which in turns means that the `getaddresses`-method keeps working.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -8,7 +8,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [ 3.8, 3.9, "3.10", "3.11", "3.12" ]
+        python-version: [ 3.8, 3.9, "3.10", "3.11", "3.12", "3.13.0-rc.1" ]
 
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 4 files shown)


cas--:
We encountered this bug in our CI pipeline where different 3.10 point releases have flip-flopped between success (`3.10.11`, `3.10.13`, `3.10.14`) and failure (`3.10.12`, `3.10.15`) based on this getaddresses call. 

So for reference this actually resolve a problem with all the latest Python versions since they now all default to a new `strict=True` argument based around this bug fix: https://github.com/python/cpython/issues/102988

* Python 3.8.20 onwards
* Python 3.9.20 onwards
* Python 3.10.15 onwards
* Python 3.11.10 onwards
* Python 3.12.16 onwards
* Python 3.13

PR #7989, opened 2024-08-16
DynamoDB: empty update expression now returns a ValidationException

felixscherz:
Hi, this is intended to fix an issue with dynamodb where an emtpy update expression given to `update_item` raises an unexpected exception: https://github.com/getmoto/moto/issues/7988

### DIFF
--- a/moto/dynamodb/models/__init__.py
+++ b/moto/dynamodb/models/__init__.py
@@ -437,11 +437,6 @@ class DynamoDBBackend(BaseBackend):
     ) -> Item:
         table = self.get_table(table_name)
 
-        if not update_expression:
-            raise MockValidationException(
-                "Invalid UpdateExpression: The expression can not be empty;"
-            )
-
         # Support spaces between operators in an update expression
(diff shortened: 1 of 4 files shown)


felixscherz:
I see some tests are failing, I'll have another look:)

bblommers (changes requested):
Hi @felixscherz, thanks for opening a PR for this!

The `update_expression` can be None when calling `update_item` in `models`, which would be absolutely valid. I would suggest to add this validation in `responses.py` instead, where the request first comes in:
https://github.com/getmoto/moto/blob/856ded18148ad3a2e25458950a8dbee64f95fa40/moto/dynamodb/responses.py#L892

If the user-provided value is an empty string here, we should throw an exception - everything else can stay the same.

felixscherz:
Ok great, thanks for the help! I will have a look at the `responses.py`

bblommers (approved):
Great - thank you for adding this to Moto @felixscherz!

PR #7991, opened 2024-08-16
RDS Fix bug that causes second PIT restore to fail

egrif:
Previous code broke if you tried to create more than one PIT restore due to a KeyError on 'db_subnet_group' when trying to delete the key.  This just checks for existence before deleting the key

Added a test to the existing `test_restore_db_instance_to_point_in_time()`.  Don't know if it's wanted or if I should ass it to a separate function


bblommers (approved):
That looks great - thank you for fixing this @egrif!

PR #7995, opened 2024-08-17
Admin: Compress JSON in binary distributions

### LINKED ISSUE #7909: Large size on disk when installing for just one service

bblommers:
Fixes #7909 

Registers a new custom Command when running `python -m build` that finds all JSON files and compresses them.
This only happens during `bdist` creation, so the source distribution is left intact.

@bpandola How do you feel about this approach?

### DIFF
--- a/moto/cognitoidp/models.py
+++ b/moto/cognitoidp/models.py
@@ -1,7 +1,5 @@
 import datetime
 import enum
-import json
-import os
 import re
 import time
 import typing
@@ -15,7 +13,7 @@ from moto.core.common_models import BaseModel
 from moto.core.utils import utcnow
 from moto.moto_api._internal import mock_random as random
 from moto.utilities.paginator import paginate
-from moto.utilities.utils import get_partition, md5_hash
+from moto.utilities.utils import get_partition, load_resource, md5_hash
 
 from ..settings import get_cognito_idp_user_pool_id_strategy
 from .exceptions import (
@@ -439,10 +437,8 @@ class CognitoIdpUserPool(BaseModel):
         self.access_tokens: Dict[str, Tuple[str, str]] = {}
         self.id_tokens: Dict[str, Tuple[str, str]] = {}
 
-        with open(
-            os.path.join(os.path.dirname(__file__), "resources/jwks-private.json")
-        ) as f:
-            self.json_web_key = jwk.RSAKey.import_key(json.loads(f.read()))
+        jwks_file = load_resource(__name__, "resources/jwks-private.json")
+        self.json_web_key = jwk.RSAKey.import_key(jwks_file)
 
     @property
     def backend(self) -> "CognitoIdpBackend":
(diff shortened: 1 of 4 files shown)


moto/ec2/models/instance_types.py
```diff
@@ -17,12 +17,15 @@
 INSTANCE_TYPE_OFFERINGS: Dict[str, Any] = {}
 for _location_type in listdir(root / offerings_path):
     INSTANCE_TYPE_OFFERINGS[_location_type] = {}
-    for _region in listdir(root / offerings_path / _location_type):
-        full_path = offerings_path + "/" + _location_type + "/" + _region
+    for data_file in listdir(root / offerings_path / _location_type):
```
bblommers, line 20: In the previous iteration, `_region` would actually mean the filename, for example `us-east-1.json`. This rewrite has slightly more sensible variable-names, and also deals with both compressed and regular filenames (`us-east-1.json` and `us-east-1.json.gz`).

moto/cognitoidp/models.py
```diff
@@ -439,10 +437,8 @@ def __init__(
         self.access_tokens: Dict[str, Tuple[str, str]] = {}
         self.id_tokens: Dict[str, Tuple[str, str]] = {}
 
-        with open(
-            os.path.join(os.path.dirname(__file__), "resources/jwks-private.json")
-        ) as f:
-            self.json_web_key = jwk.RSAKey.import_key(json.loads(f.read()))
+        jwks_file = load_resource(__name__, "resources/jwks-private.json")
```
bblommers, line 440: This is a JSON file, so this is now also compressed, and needs to be handled as such. 

Compressing isn't really necessary for this particular file, as it is very small, but it does make things consistent.

bpandola (approved):
Well I certainly like that the files changed count is 4 and not 400! 😄 

Yeah, this looks simple and effective.  Nice work!

bpandola:
@bblommers I don't have any idea how many people would need/want to edit any of these JSON files manually after installing moto, but obviously this makes it difficult (though not impossible) to do so.  Botocore seems to only compress a handful of their largest data files.  I understand you're making a trade-off here for consistency's sake (i.e. compress all JSON files).  I don't have a strong opinion on this, but just wanted to mention it as food for thought.

bblommers:
I've never encountered people manually editing/loading the JSON files, but I'm sure there is some use-case out there. Let's see - if this breaks the [workflow](https://xkcd.com/1172/) for enough people we can always revisit.

Thank you for reviewing and the feedback @bpandola !

PR #8006, opened 2024-08-20
Feature/ssm: Add gov cloud region us-gov-east-1 and us-gov-west-1

maxhoheiser:
This PR adds the SSM AWS Linux resources available in two not publicly available gov cloud regions - us-gov-east-1 and us-gov-west-1.

The data was provided by a customer who preferred not to share the original AMI values; hence, these values are randomly generated and differ from the official AWS values.

### DIFF
--- a/moto/ssm/resources/ami-amazon-linux-latest/us-gov-east-1.json
+++ b/moto/ssm/resources/ami-amazon-linux-latest/us-gov-east-1.json
@@ -1,0 +1,191 @@
+[
+    {
+        "ARN": "arn:aws-us-gov:ssm:us-gov-east-1::parameter/aws/service/ami-amazon-linux-latest/al2023-ami-kernel-6.1-arm64",
+        "DataType": "text",
+        "LastModifiedDate": 1721692759.417,
+        "Name": "/aws/service/ami-amazon-linux-latest/al2023-ami-kernel-6.1-arm64",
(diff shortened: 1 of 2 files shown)


bblommers (commented):
Hi @maxhoheiser! Thank you for sharing these.

The SSM parameters for the other region are updated automatically using [this script](https://github.com/getmoto/moto/blob/master/scripts/ssm_get_default_amis.py), on a weekly or monthly basis.
That gives two major guarantees:
 - The names/details are always up to date, and
 - The mentioned AMI's are available via EC2

Introducing these (essentially hardcoded) parameters would give off the impression that the same guarantees apply to `us-gov`-regions, which Moto can't actually commit to. So I'm very hesitant to merge this - I don't want to be in a situation where there's an expectation that we will _always_ support this functionality, given that the ARN's will be out-of-date in 4 months.

It would be amazing to have support for `us-gov`-regions, so I'm open to other suggestions on how to achieve this.

maxhoheiser:
Hi Bert,

That's a very valid point, I think for now I also don't have a suggestion on how to automate this to be able to guarantee updates. I will merge it into our LocalStack fork of moto for now move forward, and keep you update if I get access to gov-cloud so we can have a more automated way. 
Let's close this for now until we have a more satisfactory approach.

PR #8007, opened 2024-08-20
S3: create_bucket() now has additional LocationConstraint-validation

### LINKED ISSUE #7774: S3: LocationConstraints should be validated
Two scenario's:

 - (How) Does AWS complain if the region is incorrect in the same partition?
 - (How) does AWS complain if the region is incorrect in a different partition?

```
s3_resource = boto3.resource("s3", region_name="us-east-2")
s3_resource.create_bucket(Bucket="foobar", CreateBucketConfiguration={"LocationConstraint": "eu-central-1"})
```

```
s3_resource = boto3.resource("s3", region_name="us-east-2")
s3_resource.create_bucket(Bucket="foobar", CreateBucketConfiguration={"LocationConstraint": "us-gov-east-2"})
```

bblommers:
Fixes #7774

The provided LocationConstraint must be equal to the region that the request was send to, same as with AWS.

Note that this may catch a lot of people out if region names are not explicitly specified, and it is taken from either a previous (cached) boto3-client, a configuration setting or from a environment variable.

For example, if `~/.aws/config` states that `region = us-east-1`, the following now (correctly) fails:
```
s3_conn = boto3.client("s3")
s3_conn.create_bucket(Bucket=bucket_name, "CreateBucketConfiguration": {"LocationConstraint": "us-west-2"})
```

Moto was (until now) quite forgiving and simply assumed that you really meant to send this request to `us-west-2`, but will now throw a `InvalidLocationConstraint`-exception.

### DIFF
--- a/moto/s3/exceptions.py
+++ b/moto/s3/exceptions.py
@@ -203,6 +203,26 @@ class IllegalLocationConstraintException(S3ClientError):
         )
 
 
+class IncompatibleLocationConstraintException(S3ClientError):
+    code = 400
+
+    def __init__(self, location: str) -> None:
+        super().__init__(
+            "IllegalLocationConstraintException",
+            f"The {location} location constraint is incompatible for the region specific endpoint this request was sent to.",
+        )
+
+
+class InvalidLocationConstraintException(S3ClientError):
+    code = 400
+
+    def __init__(self) -> None:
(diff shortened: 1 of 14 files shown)


bblommers:
FYI @dfangl

PR #8012, opened 2024-08-21
[AppMesh] Routing

armichaud:
Implements the following methods for AppMesh

- describe_virtual_router
- create_virtual_router
- update_virtual_router
- delete_virtual_router
- list_virtual_routers
- describe_route
- create_route
- update_route
- delete_route
- list_routes


armichaud:
@bblommers this is ready for review. It doesn't completely close out https://github.com/getmoto/moto/issues/8001, but this was already fairly long so I thought I'd submit it as is and open a separate PR for the virtual_node APIs.

moto/appmesh/responses.py
```diff
@@ -97,3 +101,151 @@ def tag_resource(self) -> str:
             tags=tags,
         )
         return json.dumps(dict())
+
+    def describe_virtual_router(self) -> str:
+        mesh_name = self._get_param("meshName")
+        mesh_owner = self._get_param("meshOwner")
```
bblommers, line 107:
```suggestion
        mesh_owner = self._get_param("meshOwner") or self.current_account
```

The `meshOwner`-parameter is not mandatory, so I assume it defaults to the current account.

There are two ways to handle this, I suppose - either default to the current account, or make sure that the `AppMeshBackend.describe_virtual_router` can deal with an optional `mesh_owner`. It currently always assumes that it is supplied.

I have no preference how it's handled - this is just one suggestion.

I'm pointing it out for this particular method, but as far as I can see, this parameter is optional for a lot of methods - and the current implementation assumes it always exists. So this should be fixed across the board.

tests/test_appmesh/test_appmesh.py
```diff
@@ -148,3 +156,501 @@ def test_tag_and_list_tags_for_resource(client):
...
+
+    connection = client.create_virtual_router(
+        meshName=MESH_NAME,
+        meshOwner=mesh_owner,
+        virtualRouterName=ROUTER_2,
+        spec={"listeners": [{"portMapping": {"port": 443, "protocol": "http2"}}]},
+        tags=[{"key": "router_traffic", "value": "https"}],
+    )
+    router2 = connection.get("virtualRouter")
+    assert router2 is not None
```
bblommers, line 204: This assertion is technically not necessary, just because the next line immediately uses the variable. If the variable is None, the test will fail anyway with a `KeyError`. There are quite a few examples like this.

For me, it would be cleaner to remove it - just because less lines of code make it easier to read.

It's very much personal preference though, and not particularly important, so I'm not expecting any changes. It's just an FYI - if you find it easier to read _with_ the explicit assertions, feel free to keep it. :slightly_smiling_face:

bblommers (changes requested):

moto/appmesh/responses.py
armichaud: @bblommers good call, I've made sure owner is optional across the board. Ready for re-review

bblommers (approved):
LGTM - thank you @armichaud!

PR #8016, opened 2024-08-21
Match wafv2 backend service name to service

dfangl:
## Motivation
Currently, the service name in the wafv2 backend dict is "waf-regional". This mismatch leads to errors when trying to load the service based on the service name in the backend dict (as in LocalStack persistence).

It would be nice to match the service name to the actual service, so this mismatch is prevented.
The `waf-regional` name is for waf classic, not wafv2: https://docs.aws.amazon.com/waf/latest/developerguide/classic-waf-chapter.html

## Changes
* Change wafv2 backend service name from `waf-regional` to `wafv2`

### DIFF
--- a/moto/wafv2/models.py
+++ b/moto/wafv2/models.py
@@ -422,6 +422,4 @@ class WAFV2Backend(BaseBackend):
         ]
 
 
-wafv2_backends = BackendDict(
-    WAFV2Backend, "waf-regional", additional_regions=PARTITION_NAMES
-)
+wafv2_backends = BackendDict(WAFV2Backend, "wafv2", additional_regions=PARTITION_NAMES)


bblommers (approved):
To give some background: Moto uses the service-name here to determine the regions in which this service is available, using `boto3.Session().get_available_regions(..)`.

Whenever there is a mismatch between the service and name, that is typically because `boto3` doesn't have any regions for that service. For WAFv2, the service was [added to Moto](https://github.com/getmoto/moto/commit/1b7e015e19a63135db94f1c5de2388c296598435) in Aug 2021. 

Because `boto3.Session().get_available_regions('wafv2')` didn't return any regions, we used `waf-regional` instead - the service that was most similar that did have a list of regions.

There are other services where the service-name doesn't match, for the same reason. So for the long-term, it may be better if LS has it's own names/mapping to determine what service it's reasoning about.

In this case, `boto3` added [the regions for WAFv2](https://github.com/boto/botocore/commit/fbd4ce4382e47024b11649db391545adee005d21) in June 2022, so I think we can safely swap it now. If people use an older version of `boto3`, they won't be able to use `WAFv2` in the latest version of Moto, but I don't imagine that's a very large group of people.

dfangl:
Thank you for the context - it indeed makes sense!

We will look into proper mechanisms should this issue arise again in the future, and not solved that easily.

PR #8021, opened 2024-08-22
Resolved issue with Configuration being a required parameter for Athena Workgroups

chris-p-baker:
Bug introduced by pull request https://github.com/getmoto/moto/pull/7946 which force Configuration to become a required parameter for Athena Workgroups when it is not. This change allow for the work in the pull request mentioned to function without making Configuration required.

### DIFF
--- a/moto/athena/models.py
+++ b/moto/athena/models.py
@@ -62,6 +62,9 @@ class WorkGroup(TaggableResourceMixin, BaseModel):
         self.description = description
         self.configuration = configuration
 
+        if not configuration:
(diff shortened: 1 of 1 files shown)


bblommers (changes requested):
Hi @chris-p-baker! Thank you for the PR. I think we can make this a one-liner:
```
self.configuration = configuration or {}
```

Normally I would have merge this as-is, because it works fine. But the linter is now failing, so while adressing that, we might as well simplify the fix.

PR #8024, opened 2024-08-22
S3: delete_objects() now respects BucketPolicy and ObjectLocks

### LINKED ISSUE #8017: Object retention (ObjectLock) or no-delete policy is not always considered when deleting objects
Firstly: my environment is built using pip install. I have these versions:
```
moto     5.0.9
boto3    1.35.0
botocore 1.35.0
```
Well, no, firstly, thanks for the great work!

My intention was writing a file which couldn't be deleted. To that purpose, I've looked for everything I need to create a properly configured bucket, and then write an object with any required option.

I'd say I have succeeded because, when trying to remove the file, I get a `Forbidden` error. The problem is that if I use an Bucket(...).filter(...).delete(), the file is removed.

This test works, but shouldn't: Commented are some lines I'd say would be required to have a read-only file. The only required thing (`s3_object.delete()` will work without it) is the policy configuration.
I am now thinking that this is not a problem about OblectLocks, but about respecting the policy. Using OblectLocks makes no difference (or I am missing something to make them work).

I must say I can't tell my code works on the real AWS (I can't create such buckets) but being unable to remove the file in one way and having the file removed in a different one (both using moto), I'd say we have something.

```python
import boto3
import json
from moto import mock_aws as MockAWS
import pytest

def test_write_only_file():
    with MockAWS():
        do_test_write_only_file()


def do_test_write_only_file():
    s3r = boto3.resource('s3')
    s3c = boto3.client("s3")

    bucket_name = "some_bucket"

    s3c.create_bucket(
        Bucket=bucket_name,
#       ObjectLockEnabledForBucket=True
    )

    bucket_policy = {
        "Version": "2012-10-17",
        "Statement": [
            {
                "Sid": "DenyDelete",
                "Effect": "Deny",
                "Principal": "*",
                "Action": "s3:DeleteObject",
                "Resource": f"arn:aws:s3:::{bucket_name}/*",
            }
        ]
    }

    # Set the bucket policy
    s3c.put_bucket_policy(
        Bucket=bucket_name,
        Policy=json.dumps(bucket_policy)
    )

#   s3c.put_bucket_versioning(
#       Bucket=bucket_name,
#       VersioningConfiguration={
#           'Status': 'Enabled'
#       }
#   )

#   s3c.put_object_lock_configuration(
#       Bucket=bucket_name,
#       ObjectLockConfiguration={
#           'ObjectLockEnabled': 'Enabled',
#           'Rule': {
#               'DefaultRetention': {
#                   'Mode': 'COMPLIANCE',
#                   'Days': 30  # Retain the object for 30 days, for example
#               }
#           }
#       }
#   )

    # Finally, create an object
    s3_key = "some/path/filename"
    s3_object = s3r.ObjectSummary(bucket_name, s3_key)

    with pytest.raises(ClientError) as e:
        s3_object.load()  # Obviously, the file is still not there
    assert "Not Found" in str(e.value)

    s3_object.put(
        Body=b'this can not be deleted',
#       ObjectLockLegalHoldStatus="ON",
#       ObjectLockMode='COMPLIANCE',
#       ObjectLockRetainUntilDate="2024-12-12T00:00:00.000Z"
    )
    s3_object.load()

    with pytest.raises(ClientError) as e:
        s3_object.delete()
    assert "Forbidden" in str(e.value)

    # Object is not deleted, so this works
    s3_object.load()

    result = s3r.Bucket(bucket_name).objects.filter(Prefix=s3_key).delete()
    print(result)  # We can see the file listed in result[0]["Deleted"], it should be in ...["Errors"]

    # This shouldn't raise (object not removed) but it does, the object is not there!!
    with pytest.raises(ClientError) as e:
        s3_object.load()
    assert "Not Found" in str(e.value)
```

bblommers:
Fixes #8017

### DIFF
--- a/moto/s3/responses.py
+++ b/moto/s3/responses.py
@@ -28,7 +28,6 @@ from .exceptions import (
     AccessForbidden,
     BucketAccessDeniedError,
     BucketAlreadyExists,
-    BucketMustHaveLockeEnabled,
     DuplicateTagKeys,
     HeadOnDeleteMarker,
     IllegalLocationConstraintException,
@@ -769,9 +768,6 @@ class S3Response(BaseResponse):
         if "object-lock" in querystring:
             config = self._lock_config_from_body()
 
-            if not self.backend.get_bucket(bucket_name).object_lock_enabled:
-                raise BucketMustHaveLockeEnabled
-
             self.backend.put_object_lock_configuration(
                 bucket_name,
                 config.get("enabled"),  # type: ignore
@@ -1185,6 +1181,9 @@ class S3Response(BaseResponse):
     ) -> TYPE_RESPONSE:
         template = self.response_template(S3_DELETE_KEYS_RESPONSE)
         body_dict = xmltodict.parse(self.body, strip_whitespace=False)
+        bypass_retention = (
+            self.headers.get("x-amz-bypass-governance-retention") == "True"
+        )
 
         objects = body_dict["Delete"].get("Object", [])
         if not isinstance(objects, list):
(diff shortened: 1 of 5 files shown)


moto/s3/responses.py
```diff
@@ -769,9 +768,6 @@ def _bucket_response_put(
         if "object-lock" in querystring:
             config = self._lock_config_from_body()
 
-            if not self.backend.get_bucket(bucket_name).object_lock_enabled:
```
bblommers, line 772: This check doesn't seem to exist in AWS (anymore)

Issue #8027, opened 2024-08-23
MFA_SETUP ChallengeName not supported


jpmckinney:
If the user pool has configured MFA, then I would expect an MFA_SETUP challenge after, let's say, a NEW_PASSWORD_REQUIRED challenge, when setting up a new user.

I'll create a PR with some code that gets part of the way there. (#8028)

bblommers:
Hi @jpmckinney, thanks for raising this. Just to clarify, was this the workflow that you  had in mind?

1. client.create_user_pool(...)
2. client.set_user_pool_mfa_config(...)
3. client.create_user_pool_client(...)
4. client.admin_create_user(..., TemporaryPassword=temporary_password)
5. client.initiate_auth(..., AuthParameters={"USERNAME": username, "PASSWORD": temporary_password})
6. result = client.respond_to_auth_challenge(..., ChallengeName="NEW_PASSWORD_REQUIRED")
7. client.associate_software_token(AccessToken=result["Session"])

I've been trying to reproduce it like this, and I think it's correct - but AWS throws an `Invalid Access Token`-error in the last step.
Is there anything that I'm missing in this setup?

jpmckinney:
Yes, that’s the workflow - except the only parameter for associate_software_token is Session.

jpmckinney:
And after associate, there needs to be client.verify_software_token(Session, SecretCode)

bblommers:
> Yes, that’s the workflow - except the only parameter for associate_software_token is Session.

Aah.. of course. I was looking at the flow/ordering of the commands for so long that I completely overlooked the `Session`-parameter.

Thank you as well for getting started on a solution @jpmckinney - I've submitted #8037 to add support for this, which builds on your work in #8028.

PR #8028, opened 2024-08-23
Issue MFA_SETUP challenge

jpmckinney:
#8027

These changes are naive – they are just intended to show which methods need to change.

https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/cognito-idp/client/associate_software_token.html
https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/cognito-idp/client/verify_software_token.html


bblommers:
These changes/commits have been committed to Moto as part of https://github.com/getmoto/moto/pull/8037.

PR #8039, opened 2024-08-27
ACM-PCA: Use serialisable types instead of native objs

viren-nadkarni:
This is a follow up to https://github.com/getmoto/moto/pull/7958

The `ImportCertificateAuthorityCertificate` would create a `cryptography.hazmat.bindings._rust.x509.Certificate` object as assign it as a `CertificateAuthority` class member. This prevents it from being serialised with pickle.

This PR reworks the certificate and private key attributes so that all members are picklable.


bblommers:
Hey @viren-nadkarni! Do you want to add some tests for this as well in Moto? Just to prevent it from breaking again in the future

moto/acmpca/models.py
```diff
@@ -287,7 +297,7 @@ def not_valid_before(self) -> Optional[float]:
     def import_certificate_authority_certificate(
         self, certificate: bytes, certificate_chain: Optional[bytes]
     ) -> None:
-        self.certificate = x509.load_pem_x509_certificate(certificate)
+        self.certificate_bytes = certificate
```
bblommers, line 300:
```suggestion
        x509.load_pem_x509_certificate(certificate)
        self.certificate_bytes = certificate
```
Ideally we keep the validation that `certificate`-argument contains a valid X509 certificate. 

As an alternative, something like `assert self.certificate is not None` would also work, as that loads the certificate behind the scenes, but it's less obvious why we're making that assertion.

bblommers (changes requested):

moto/acmpca/models.py
viren-nadkarni: Great catch, added the check and test in ecfa1b336. Thanks @bblommers !

bblommers (approved):
Great, thanks @viren-nadkarni!

PR #8041, opened 2024-08-27
Catch and verify an expected `UserWarning` in the test suite

kurtmckee:
The test suite validates that an SQS FIFO queue with content-deduplication disabled does not receive messages from EventBridge, but doesn't catch and verify the `UserWarning` that gets raised in the code because of this ([recent example in CI](https://github.com/getmoto/moto/actions/runs/10570098846/job/29284013829#step:7:865)). As a result, that `UserWarning` gets reported by pytest at the end of the test run.

This PR uses `pytest.warns()` to catch and verify the expected `UserWarning`.

### DIFF
--- a/tests/test_events/test_events_integration.py
+++ b/tests/test_events/test_events_integration.py
@@ -1,3 +1,4 @@
+import contextlib
 import json
 import os
 from datetime import datetime
@@ -111,7 +112,13 @@ def test_send_to_sqs_fifo_queue():
 
     # when
     event_time = datetime(2021, 1, 1, 12, 23, 34)
-    with pytest.warns(UserWarning, match="you must enable content-based deduplication"):
+    # Catch and validate a UserWarning when running locally.
+    context = (
+        pytest.warns(UserWarning, match="you must enable content-based deduplication")
+        if not settings.TEST_SERVER_MODE
+        else contextlib.nullcontext()
+    )
+    with context:
         client_events.put_events(
             Entries=[
                 {
--- a/tests/test_events/test_events_integration.py
+++ b/tests/test_events/test_events_integration.py
@@ -4,6 +4,7 @@ from datetime import datetime
 from unittest import SkipTest, mock
 
 import boto3
+import pytest
 
 from moto import mock_aws, settings
 from moto.core import DEFAULT_ACCOUNT_ID as ACCOUNT_ID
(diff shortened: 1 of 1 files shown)


bblommers (changes requested):
Hi @kurtmckee! Thank you for improving our CI logs.

The tests are run twice - once regular, and once in ServerMode.
Moto in ServerMode runs in separate Docker-container, so we can't intercept any user warnings there.

We could handle this with an if-else (if normal: intercept warning & put_events, else: put_events without intercept). I'm not sure if it's worth the effort though, considering it doesn't improve the readability of the test.

If you can think of a different non-invasive way to achieve this, I'm all ears, but the warnings in the CI don't bother me enough to warrant big changes to the test.

kurtmckee:
Hello @bblommers, and thanks for the quick review!

The benefit to reducing these warnings is to move toward escalating warnings to errors. It is possible to filter out warnings caused by project dependencies, but it's good hygiene to address (or filter out) existing warnings and then -- from a position of zero warnings -- escalate all warnings to errors. This allows CI to catch new Python `DeprecationWarning`s, or warnings issued by dependencies, early.

It looks like the CI failures are unrelated to this change; were you mentioning the two modes of execution to suggest that this change caused some of the CI failures seen here? Here are the failing tests in CI ([Python 3.12, Ubuntu](https://github.com/getmoto/moto/actions/runs/10580339607/job/29315352846?pr=8041#step:9:1130)):

* `tests/test_core/test_proxy.py::test_http_get_request_can_be_passed_through`
* `tests/test_core/test_proxy.py::test_http_post_request_can_be_passed_through`
* `tests/test_core/test_proxy.py::test_https_request_can_be_passed_through`

but this PR only touched this test:

* `tests/test_events/test_events_integration.py::test_send_to_sqs_fifo_queue`

Looking at the Python 3.12 / Ubuntu CI logs, I'm not seeing that this is related, so please let me know if I'm misunderstanding!

bblommers:
@kurtmckee Those tests all reach out to website that was temporarily unavailable - that should be fine when restarting the build.  I was talking about these logs:
https://github.com/getmoto/moto/actions/runs/10580339607/job/29315346057

> E           Failed: DID NOT WARN. No warnings of type (<class 'UserWarning'>,) were emitted.
> E            Emitted warnings: [].

kurtmckee:
Aha, I see now. Thanks for pointing that out!

If you're game for it, lemme see if there's something simple that can be done for this scenario, otherwise you can close this PR. 👍

bblommers (approved):
LGTM - thank you for fixing this @kurtmckee!

PR #8043, opened 2024-08-27
Batch: fix job parameters definition

pinzon:
As this [document](https://docs.aws.amazon.com/batch/latest/userguide/job_definition_parameters.html#parameters) shows. The parameters are defined in a command like `Ref::parameter` and not `$parameter`. This was a small error introduced in my last PR.


bblommers (approved):
LGTM - thanks for the follow-up @pinzon!

PR #8046, opened 2024-08-28
[S3]Add Eventbridge notification for object storage class change

MikiPWata:


MikiPWata:
Hey @bblommers!
I'm trying to implement eventbridge notification for object storage class change based on [previous PR](https://github.com/getmoto/moto/pull/7448)
any tips on how I should implement it? I'm thinking adding getter and setter is not enough, as we change object's storage class by creating a copy...

bblommers:
Hi @MikiPWata!

If I'm reading the documentation right, there are three scenario's where the StorageClass is set in AWS:
1. When creating an object
2. When copying an object (_and_ the StorageClass-argument is set)
3. When a LifeCycle rule kicks in

We can ignore 1, because that's not a change. Number 3 is irrelevant, because we don't act on the lifecycle rules.

So we only have to worry about number 2, the copy operation - and because that's not yet implemented, we have full control on how we do that. Am I correct so far, or have a missed a scenario?


If it's just the copy that we have to worry about, than it makes sense to use a `getter`/`setter` pattern IMO. We can use the `setter` during object creation as well, to check that the StorageClass is valid - as that doesn't happen atm.

MikiPWata:
Hey @bblommers!
Thanks for your comment!
I think you are right, we just need to worry about the copy operation.
I tried implementing getter and setter, but the following error comes up from all test cases for S3's Eventbridge notification.

`FAILED tests/test_s3/test_s3_eventbridge_integration.py::test_delete_object_notification - botocore.errorfactory.BucketAlreadyOwnedByYou: An error occurred (BucketAlreadyOwnedByYou) when calling the CreateBucket operation: Your previous request to create the named bucket succeeded and you already own it.`

As I'm not that proficient in getters and setters, would love to get some advice on that. 

thanks in advance

MikiPWata:
@bblommers 
This PR currently contains the changes I made. Can you take a look at it?
There's no issue when I rollback the `storage_class` getter and setter, so I'm pretty sure that my implementation is wrong.

bblommers:
Ah, I understand @MikiPWata. The indentation of both the getter and setter is wrong, I guess that broke things in a weird way. If you indent both methods to make it part of the class, the other tests do pass.

bblommers:
I've done some testing against AWS, and changing the object storage class using the `copy_object` method does **not** lead to a `LIFECYCLE_TRANSITION_EVENT` - it only sends a `COPY_EVENT`. 

This PR contains a CDK stack if you want to test this out for yourself: https://github.com/getmoto/moto/pull/9378

For this to be supported, we would first need to support actual lifecycle transisions, and find a way to mock those. But this proposed implementation would not work,  unfortunately, so I'll close this PR.

PR #8050, opened 2024-08-28
Appmesh virtual node

armichaud:
Implements describe, create, list, update, and delete virtual_node APIs


armichaud:
@bblommers ready for review

IMPLEMENTATION_COVERAGE.md
```diff
@@ -751,12 +751,11 @@
 
 ## bedrock
 <details>
-<summary>39% implemented</summary>
+<summary>43% implemented</summary>
 
 - [ ] create_evaluation_job
 - [ ] create_guardrail
 - [ ] create_guardrail_version
-- [ ] create_model_copy_job
```
bblommers, line 759: Usually when the `implementation`-script suddenly removes methods like this, it's because the installed `botocore` module is outdated, and the feature didn't exist yet in the older `botocore` module.

We'll always update the docs before an actual release, so it's not a big deal if these methods disappear inbetween, just as an FYI; if you update `botocore`/`boto3` before running the `implementation`-script, this shouldn't happen.

moto/appmesh/utils/spec_parsing.py
```diff
@@ -0,0 +1,681 @@
...
+    Match as VirtualNodeMatch,
+)
+from moto.appmesh.dataclasses.virtual_router import PortMapping as RouterPortMapping
+from moto.appmesh.exceptions import (
+    MissingRequiredFieldError,
+)
+
+
+def clean_dict(obj: Dict[str, Any]) -> Dict[str, Any]:  # type: ignore[misc]
+    return {
```
bblommers, line 72: This method isnt' used anywhere - looks like the duplicate method in `utils.py` is used instead.

bblommers (approved):
LGTM - thank you continuing to improve this area in Moto @armichaud!

The two comments that I've left are not blockers, so I'll just go ahead and merge it as is - feel free to address them in a follow up PR.

PR #8053, opened 2024-08-29
S3: fix CreateBucket from us-east-1 in other regions, and set bucket region to LocationConstraint

bentsku:
Follow-up from #8007: it seems the PR was a bit too strict.

AWS S3 allows you to create buckets in other regions if your client region is set in `us-east-1` (and only from that region, as its behavior is already quite specific). 

Due to this, the bucket can also have a region different from the client region.

### DIFF
--- a/moto/s3/responses.py
+++ b/moto/s3/responses.py
@@ -895,7 +895,9 @@ class S3Response(BaseResponse):
             if self.body and not location_constraint:
                 raise MalformedXML()
 
-            bucket_region = location_constraint if location_constraint else DEFAULT_REGION_NAME
+            bucket_region = (
+                location_constraint if location_constraint else DEFAULT_REGION_NAME
+            )
             try:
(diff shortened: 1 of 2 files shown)


bblommers (changes requested):
Hi @bentsku, thanks for fixing this!

Would you mind deleting the bucket at the end of the test? Tests marked with `aws_verified` actually get executed against AWS once a week, so we have to make sure that all resources within those tests are deleted. :slightly_smiling_face:

bentsku:
Hi @bblommers! Ah right, thanks, I was actually wondering about the clean up! I couldn't really use `@s3_aws_verified` for it, hopefully as the test doesn't do much, if the first step fails, we won't leave resources dangling. Thanks for the quick review!

bblommers (approved):
Great, thank you @bentsku!

PR #8054, opened 2024-08-29
ACM InUseBy with elbv2

giograno:
This PR is similar to https://github.com/getmoto/moto/pull/4414, implementing the same use case for ELBV2.
In a nutshell, it updates the `InUseBy` property of an ACM certificate when attaching it to an ELBV2 listener.

I also fixed the `ListCertificate` operation that should return a boolean `InUse` property, which gets computed from the `InUseBy` attached to the certificate description.


bblommers (approved):
That's a nice and clean solution - thank you @giograno!

PR #8056, opened 2024-08-29
SageMaker: add tagging api support for additional services

jamarcelin:
Added ResourceGroupsTaggingAPI support for Sagemaker services:

- Models
- Notebook Instances
- Endpoing Configs, Endpoints
- Experiments
- Pipelines
- Pipeline Executions
- Processing Jobs
- Trials
- Trial Components
- Training Jobs
- Transform Jobs
- Notebook Instance Lifecycle Configurations
- Model Cards
- Model Package Groups
- Model Packages
- Feature Groups


moto/sagemaker/models.py
```diff
@@ -205,6 +205,7 @@ def __init__(
         client_request_token: str,
     ):
         self.pipeline_execution_arn = pipeline_execution_arn
+        self.arn = self.pipeline_execution_arn
```
bblommers, line 208: I realize that it's quite a lot of work to rewrite these - but I think we should all settle on `self.arn`, and get rid of the `self.pipeline_execution_arn`. Having two properties to point to the same value is going to be confusing, and can lead to drift.

(Same for all the other models..)

bblommers (changes requested):
Hi @jamarcelin! It looks like a test is now failing as well: https://github.com/getmoto/moto/actions/runs/10622064984/job/29445807888?pr=8056

jamarcelin:
Hey @bblommers, the services have all been updated to only use `self.arn`. Thanks!

bblommers (approved):
Great - thank you @jamarcelin!

srausser:
@bblommers it looks like you missed the transform job adding the arn back to the payload
`response["TransformJobArn"] = response.pop("Arn")`

or am i missing something

bblommers:
Good catch @srausser - I've raised https://github.com/getmoto/moto/pull/8886 so we now return the TransformJobArn

PR #8058, opened 2024-08-30
Techdebt: Improve typing/remove dead code

bblommers:


moto/ec2/responses/launch_templates.py
```diff
@@ -187,11 +187,7 @@ def describe_launch_template_versions(self) -> str:
         ret_versions = []
         if versions and template is not None:
             for v in versions:
-                if str(v).lower() == "$latest" or "$default":
```
bblommers, line 190: This if-statement always returns True, because `or "$default"` is always True.

[Originally](https://github.com/getmoto/moto/commit/2c47ee10b182e63865bc9b95674fafab38a3ff27) it obviously meant to say `if str(v).lower() in ["$latest", "$default"]`. However, because the `get_version` method already checks whether `v` is $latest/$default/numeric, the whole if-statement is pointless.

(`get_version` was introduced here: https://github.com/getmoto/moto/commit/cf2690ca1e2e23e6ea28defe3e05c198d9581f79)

moto/awslambda/responses.py
```diff
@@ -321,31 +321,22 @@ def get_function_code_signing_config(self) -> str:
 
     def get_function_concurrency(self) -> TYPE_RESPONSE:
         path_function_name = unquote(self.path.rsplit("/", 2)[-2])
-        function_name = self.backend.get_function(path_function_name)
-
-        if function_name is None:
-            return 404, {"status": 404}, "{}"
+        self.backend.get_function(path_function_name)
```
bblommers, line 324: `get_function` already raises an `UnknownFunction` exception if necessary, so these explicit 404's are never reached

PR #8059, opened 2024-08-30
EventBridge: Add Name param to ListConnections response

viren-nadkarni:
This PR adds the missing 'Name' parameter to EventBridge [ListConnections](https://docs.aws.amazon.com/eventbridge/latest/APIReference/API_ListConnections.html#API_ListConnections_ResponseSyntax) operation.

This patch is from the LocalStack fork.


bblommers (approved):
:partying_face:

PR #8080, opened 2024-09-03
[Shield] create and describe subscription

armichaud:


armichaud:
@bblommers here's just a small PR to unblock something for my team

bblommers (approved):
LGTM!

PR #8082, opened 2024-09-04
AppMesh: Support tagging for Virtual Nodes/Routers

armichaud:


bblommers (approved):
LGTM - thanks for following up here @armichaud!

PR #8086, opened 2024-09-05
ec2: fix security groups describe rule response

sannya-singal:
When using the `describe-security-group-rules` call with a security-group-id `filter`, it correctly returns the egress and ingress rules. However, if we try to use the `describe-security-group-rules` call and specify the `--security-group-rule-ids`, it does not return the rule as expected. 

This PR will now consider both egress and ingress rules while using `--security-group-rule-ids` param in the command. 

Fixes: https://github.com/localstack/localstack/issues/11445

### DIFF
--- a/moto/ec2/models/security_groups.py
+++ b/moto/ec2/models/security_groups.py
@@ -634,13 +634,17 @@ class SecurityGroupBackend:
         results = []
 
         if sg_rule_ids:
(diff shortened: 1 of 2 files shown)


bblommers (approved):
LGTM - thanks for the fix @sannya-singal!

PR #8095, opened 2024-09-07
[transfer] Fix date_imported to use a correct ISO 8601 date

mgorny:
Fix `date_imported` key to use a proper ISO 8601 format (with hyphens and colons) rather than concatenated numbers.  Botocore expects the former rather than the latter, and when given a bare number it incorrectly interprets it as UNIX timestamp rather than `%Y%m%d%H%M%S` date.

This can particularly be seen on systems with 32-bit time_t, where the resulting timestamp overflows (it would correspond to year 643378 today), and causes `tests/test_transfer/test_transfer.py` failures such as:

```
E       RuntimeError: Unable to calculate correct timezone offset for "20240907201715"
```

When using `%Y-%m-%d %H:%M:%S` format, the relevant tests pass both on systems with 64-bit and 32-bit time_t.


bblommers (changes requested):
Hi @mgorny! I've just tested this against AWS, and they actually return an epoch timestamp here - so I think we should do the same.

We have an existing utility method for that:
```
from moto.core.utils import unix_time
now = unix_time()
```

I don't know whether that affects 32-bit systems? If so, we may just have to disable the tests instead.

mgorny:
Sure, updated it to do that. It's also going to work on 32-bit arches, at least until 2038 (and then, everything else will fail too, so not an issue).

bblommers (approved):
Great, thank you @mgorny!

mgorny:
Thanks!

PR #8106, opened 2024-09-09
ELBv2: Fix CF bug when passing certificates to ELB listener creation

Ivan-Lukianov:
Context:

Recently our tests started failing with a next error:

```
pattern = '^arn:(aws|aws-cn|aws-iso|aws-iso-b|aws-us-gov):iam:'
string = [{'CertificateArn': 'arn:aws:acm:us-east-1:123456789012:certificate/<cert_id>', 'certificate_arn': 'arn:aws:acm:us-east-1:123456789012:certificate/<cert_id>'}]
flags = 0

    def search(pattern, string, flags=0):
        """Scan through string looking for a match to the pattern, returning
        a Match object, or None if no match was found."""
>       return _compile(pattern, flags).search(string)
E       TypeError: expected string or bytes-like object, got 'list'
```

I tracked down an issue to this change - https://github.com/getmoto/moto/pull/8054

In my understanding in case of listener creation from cloudformation `certificate` variable is not a string, but an array and this newly added code is failing. 

This PR adds a change so certificate passed as a string to `create_listener` method in cloudformation case similar to what is in the [elb client code](https://github.com/getmoto/moto/blob/master/moto/elbv2/responses.py#L462-L465).

### DIFF
--- a/moto/elbv2/models.py
+++ b/moto/elbv2/models.py
@@ -361,8 +361,12 @@ class FakeListener(CloudFormationModel):
 
         default_actions = elbv2_backend.convert_and_validate_properties(properties)
         certificates = elbv2_backend.convert_and_validate_certificates(certificates)
+        if certificates:
+            certificate = certificates[0].get("certificate_arn")
+        else:
+            certificate = None
         listener = elbv2_backend.create_listener(
-            load_balancer_arn, protocol, port, ssl_policy, certificates, default_actions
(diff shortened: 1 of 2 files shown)


bblommers (approved):
Hi @Ivan-Lukianov, thank you for the fix and for adding the test case!

It is an interesting decision from AWS to require a list of certificates, and at the same time state that ` You must provide exactly one certificate`... 
But with that in mind, the `if/else` statement that always picks the first certificate makes sense.

PR #8108, opened 2024-09-09
feat: Implementation of  memorydb api

archinksagar:
This is the initial moto implementation of AWS MemoryDB API.

### DIFF
--- a/IMPLEMENTATION_COVERAGE.md
+++ b/IMPLEMENTATION_COVERAGE.md
@@ -296,7 +296,7 @@
 
 ## appconfig
 <details>
-<summary>33% implemented</summary>
+<summary>34% implemented</summary>
 
 - [X] create_application
 - [X] create_configuration_profile
@@ -312,7 +312,6 @@
 - [ ] delete_extension
 - [ ] delete_extension_association
 - [X] delete_hosted_configuration_version
-- [ ] get_account_settings
 - [X] get_application
 - [ ] get_configuration
 - [X] get_configuration_profile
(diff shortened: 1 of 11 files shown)


bblommers (changes requested):
Hi @archinksagar, it looks like all tests fail in ServerMode - see https://github.com/getmoto/moto/actions/runs/10778987259/job/29891829117?pr=8108

Could you also merge in the latest master? I just pushed a fix to make Moto compatible with the latest botocore, and the build for this PR may still fail if that change is not included.

archinksagar:
@bblommers Thanks for reviewing, I merged the latest master. If the tests fail in server mode, do I need to spell out the URL paths in urls.py? Would that fix it?

bblommers:
Hi @archinksagar, yes, manually adding the URL's usually fixes the ServerMode tests. Let me know if it doesn't, or if you run into other issues, and I can have a closer look.

archinksagar:
@bblommers  how to I get all the correct URL paths for memorydb?

bblommers:
@archinksagar You can run the MotoServer locally, and Moto will tell you. Within the virtual environment that you develop Moto in, run:
```
moto_server
```
And then run the tests against it:
```
TEST_SERVER_MODE=true pytest -sv tests/test_memorydb
```
The server logs will then tell you which URL's are being visited, and which requests are failing.

archinksagar:
@bblommers Server mode tests were failing because it did not find the correct URL pattern in werkzeug_app. I added few changes to that file. Could you review again and let me know if the changes are right in werkzeug_app.py. Thankyou.

bblommers (approved):
This looks great - thank you @archinksagar!

PR #8117, opened 2024-09-11
Workspaces Web Support

jamarcelin:
Added mocking for workspaces web (workspaces secure browser)


jamarcelin:
@bblommers Hello! I am having some issues here. I have implemented all of the methods that are being tested in the workspaces web test file but when running the tests half of the methods give a "Not Yet Implemented" error. I think the issue is in the urls.py but I am not sure. Any help would be appreciated, thanks!

bblommers (approved):

PR #8119, opened 2024-09-11
Apigateway Account APIs

mattheidelbaugh:
- Implement get_account and update_account for apigateway
- Added rootResourceId to the RestAPI class and its to_dict function so that it will be returned as it was missing ([create_rest_api](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/apigateway/client/create_rest_api.html), [get_rest_api](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/apigateway/client/get_rest_api.html))

### DIFF
--- a/moto/apigateway/models.py
+++ b/moto/apigateway/models.py
@@ -1528,6 +1528,46 @@ class GatewayResponse(BaseModel):
             dct["responseTemplates"] = self.response_templates
         return dct
 
+class Account(BaseModel):
+    def __init__(self):
+        self.cloudwatch_role_arn: Optional[str] = None
+        self.throttle_settings: Dict[str, Any] = {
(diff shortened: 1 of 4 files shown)


mattheidelbaugh:
Some notes/questions:
- It seems as though we don't get the boto validation for patchOperations.  Validation for this doesn't seem to be implemented for any other resources, so I didn't do it here.  Can certainly add that if desired.  Could easily be done in a utility function and applied to all functions which take patchOperations
- For some reason I can't discern, MyPy is unhappy with my Account class which takes no arguments, so I've ignored those errors.

zkarpinski (approved):
lgtm

bblommers (approved):
LGTM - thank you @mattheidelbaugh!

PR #8131, opened 2024-09-19
[BUG FIX] route53 list_tags_for_resources no results & hostedzone ids not decoding properly

### LINKED ISSUE #8130: [BUG] route53 tagging for hostedzones & list_tags_for_resources does not work properly
There are two bugs for route53.

To recreate, add `assert len(response["ResourceTagSets"]) == 2` in test_route53.py -> test_list_tags_for_resources() above `for set in response["ResourceTagSets"]:` on line 588.

list_tags_for_resources() doesn't actually return anything and it's stems from 3 bugs.

1. The **list_or_change_tags_for_resource_request** api doesn't decode the url encoding in the ID. Thus, characters like "/" are encoded when tags are added. This impacts hostedzones which have slashes in the ID.  **list_tags_for_resources** the id are not encoded, so the key is never found and returns no results.
2. The list of ids isn't capture correctly, it's passing a dictionary instead of the list of ids.
3. The template for list_tags_for_resources doesn't work as expected and needs refactoring.

The test-case mentioned doesn't test what you expect and thus passes incorrectly.

zkarpinski:
Resolves #8130 

Adds additional test scenarios

### DIFF
--- a/moto/route53/models.py
+++ b/moto/route53/models.py
@@ -648,9 +648,11 @@ class Route53Backend(BaseBackend):
         return {}
 
     def list_tags_for_resources(self, resource_ids: List[str]) -> Dict[str, str]:
-        resources = {}
+        resources = []
         for id in resource_ids:
-            resources.update(self.list_tags_for_resource(id))
+            resource_set = {"resource_id": id}
+            resource_set["tags"] = self.list_tags_for_resource(id)
+            resources.append(resource_set)
         return resources
 
     def list_resource_record_sets(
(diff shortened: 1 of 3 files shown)


zkarpinski:
Hey @bblommers Not sure why this is failing with an empty list of tags in **ServerMode**. Hoping you can take a look when you have a chance.

bblommers:
Hi @zkarpinski! The failing test was because of another encoding issue. In ServerMode, the URL comes already unquoted, so we have to handle both cases:
```
/2013-04-01/tags/hostedzone//hostedzone/PNOBIQEFOI7M961        # ServerMode
/2013-04-01/tags/hostedzone/%2Fhostedzone%2F6ZT6E5GX46CTE59    # Regular decorators
```
I've pushed a change that should fix this.

bblommers (approved):
Thanks for fixing this and adding a more elaborate test @zkarpinski!

PR #8132, opened 2024-09-19
NetworkManager: Update resource states to 'AVAILABLE' from 'PENDING' after create response

zkarpinski:
For NetworkManager: Update resource states to 'AVAILABLE' from 'PENDING' after create response to more accurately represent the mocked resource

### DIFF
--- a/moto/networkmanager/models.py
+++ b/moto/networkmanager/models.py
@@ -270,6 +270,11 @@ class NetworkManagerBackend(BaseBackend):
         except (KeyError, ValueError, AttributeError):
             raise ResourceNotFound(arn)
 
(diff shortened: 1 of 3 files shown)


mattheidelbaugh (commented):
Probably not necessary but I think you could use the Managed State model here if you wanted

bblommers:
> Probably not necessary but I think you could use the Managed State model here if you wanted

That is true, @mattheidelbaugh, that would achieve the same result. For this scenario, with only one immediate transition, the current implementation is probably the simplest, so I'm happy with that.

Thanks for the PR @zkarpinski!

bblommers (approved):

PR #8134, opened 2024-09-20
Workspaces Web User Settings Support

jamarcelin:


jamarcelin:
@bblommers Hello! Any idea how to fix this linting issue? The line with the `# type: ignore[misc]` keeps getting moved around by the formatter.

moto/workspacesweb/responses.py
```diff
@@ -38,6 +38,26 @@ def browser_settings(request: Any, full_url: str, headers: Any) -> TYPE_RESPONSE
...
+        handler.setup_class(request, full_url, headers)
+        if request.method == "GET":
+            return handler.get_user_settings()
+        else:
+            return handler.delete_user_settings()
+
+    @staticmethod
+    def user_access_logging_settings(
+        request: Any, full_url: str, headers: Any
+    ) -> TYPE_RESPONSE:  # type: ignore[misc]
```
bblommers, line 53:
```suggestion
    def user_access_logging_settings(  # type: ignore[misc]
        request: Any, full_url: str, headers: Any
    ) -> TYPE_RESPONSE:
```
Our formatter `ruff` isn't smart enough to understand where the `type: ignore` needs to go, so I've moved it manually to the right line. I haven't tested this locally, but I think that should fix it :crossed_fingers:

bblommers (approved):
LGTM - thank you for adding this to Moto @jamarcelin!

PR #8143, opened 2024-09-23
[BUG Fix] route53 - list_tags_for_resources: Resolve when a single ID is provided

zkarpinski:
Fixes two bugs.

1. When list_tags_for_resource is passed only a single ID, the request is a string and not a list
2. The regex for deriving resource_type was always None


zkarpinski:
@bblommers Another bug fix for the route53 endpoint that slipped through :(

bblommers (approved):
Apologies for that @zkarpinski - I did see there was a difference between a call with a single resource and one with multiple resources, but obviously didn't test both scenarios.

Thanks for the fix!

PR #8149, opened 2024-09-24
[FSx] Update file system id generation to expected regex

rpmcginty:
This change intends to address https://github.com/getmoto/moto/issues/8148

Modifies how fsx fs id is generated so that:
* there are not collisions for requests that happen in parallel
* it complies with the expected regex.

### DIFF
--- a/moto/fsx/models.py
+++ b/moto/fsx/models.py
@@ -1,4 +1,5 @@
 """FSxBackend class with methods for supported APIs."""
+import uuid
 
 from datetime import datetime
 from typing import Any, Dict, List, Optional, Tuple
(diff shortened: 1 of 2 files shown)


bblommers:
Hi @rpmcginty, thanks for raising a PR here.

Do you know what AWS actually uses as the length of this identifier? The documentation is a bit confused:
> FileSystemId

    The system-generated, unique 17-digit ID of the file system.

    Type: String

    Length Constraints: Minimum length of 11. Maximum length of 21.

    Pattern: ^(fs-[0-9a-f]{8,})$

    Required: No


Based on this I can't tell whether it's supposed to 8, 17, or anything between 11 and 21.

rpmcginty:
> Hi @rpmcginty, thanks for raising a PR here.
> 
> Do you know what AWS actually uses as the length of this identifier? The documentation is a bit confused:
> 
> > FileSystemId
> 
> ```
> The system-generated, unique 17-digit ID of the file system.
> 
> Type: String
> 
> Length Constraints: Minimum length of 11. Maximum length of 21.
> 
> Pattern: ^(fs-[0-9a-f]{8,})$
> 
> Required: No
> ```
> 
> Based on this I can't tell whether it's supposed to 8, 17, or anything between 11 and 21.

yeah it is a bit confusing I suppose. The character count does include the "fs-" so I think the minimum is in fact 11 characters (not 8). Having dealt with fsx ids I have only seen 17 characters so I can update it if you would like me to.

bblommers (approved):
I'm happy with having a length of 8/11 @rpmcginty, I was just curious. As long as the identifiers are unique, I don't think the exact format matters.

Thank you for fixing this!

PR #8150, opened 2024-09-24
[BUG FIX] EC2: Exclude existing tags to be overwritten from the TagLimitExceeded validation

elstak:
Although EC2 API does not provide a method to update existing tags, [CreateTags](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_CreateTags.html) will overwrite existing tag keys with the new given value. Therefore, existing tags passed as argument to `create_tags()` method should not be taken into consideration to validate `TagLimitExceeded` exception when the resource underneath exists.

It has been tested in AWS that, updating one or more tag values of a resource which tag count is already on the limit (50), works well using AWS CLI, boto3 and Terraform.

### DIFF
--- a/moto/ec2/models/tags.py
+++ b/moto/ec2/models/tags.py
@@ -26,7 +26,14 @@ class TagBackend:
             if resource_id in self.tags:
                 if (
                     len(self.tags[resource_id])
-                    + len([tag for tag in tags if not tag.startswith("aws:")])
+                    + len(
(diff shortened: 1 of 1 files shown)


ayetealab (approved):

bblommers (approved):
Makes sense - thank you for the PR @elstak, and welcome to Moto!

PR #8154, opened 2024-09-24
ECS - modify-ebs-default-kms-keyid support

zkarpinski:
Adds support for the API endpoint **ModifyEbsDefaultKmsKeyId**

Ref:
https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_ModifyEbsDefaultKmsKeyId.html
https://awscli.amazonaws.com/v2/documentation/api/latest/reference/ec2/modify-ebs-default-kms-key-id.html


bblommers (approved):
LGTM! Especially the test - very useful to assert how be should behave when reverting the key to the default.

PR #8156, opened 2024-09-24
IAM - Set value of AccountSummary.AccountAccessKeysPresent

zkarpinski:
Edge case to be able to modify the value of AccountSummary AccountAccessKeysPresent from 0 to 1 and back.

### DIFF
--- a/moto/iam/models.py
+++ b/moto/iam/models.py
@@ -3424,6 +3424,9 @@ class IAMBackend(BaseBackend):
         for value in tagKeys:
             del profile.tags[value]
 
+    def _set_account_access_keys_present(self, present: bool) -> None:
+        self.account_summary._account_access_keys_present = int(present)
+
 
 iam_backends = BackendDict(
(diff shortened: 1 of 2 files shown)


bblommers (commented):
Hi @zkarpinski, I'm a little hesitant to add and support API's like this. In principle we only support access via the AWS SDK's - everything behind `service_backends` is not guaranteed to be stable, and subject to change.

Is there a particular reason why this can't be done directly?
> backend = iam_backends[ACCOUNT_ID]["global"]
> backend.account_summary._account_access_keys_present = 0

zkarpinski:
That works for me! Never even thought of that. Stuck in the mindset that the data is protected and we need a function to update it.

That works for me. Any value in modify the tests to show this as an example or nah? I can close this if there's not value.

bblommers:
> Any value in modify the tests to show this as an example or nah?

No, I don't think it's necessary. If someone asks for an example, I can always point them to this PR. :slightly_smiling_face:

PR #8160, opened 2024-09-26
Opensearch Ingestion Service Implementation

mattheidelbaugh:
- OSIS implementation and testing
- Fixed type hint for create_vpc_endpoint()

This is my first time implementing a full service so any and all feedback is welcome especially where it relates to the extent of the functionality to mimic.  Currently I'm validating that SGs and Subnets exist but nothing else (e.g. opensearch pipeline sink existence, valid pipeline source for using buffer_options, etc.).  I'm also creating VPC endpoints when the VpcOptions.VpcEndpointManagement field is set to Service to mirror what AWS does.  Let me know if I should forgo that in favor of just sending back a fake vpce id.  I am not creating Cloudwatch logs groups if they don't exist although AWS will.


mattheidelbaugh:
@bblommers For the life of me I can't figure out how to get the arn out of the Tagging APIs based on the botocore service model.  If you could take a look that would be very much appreciated

bblommers:
Hi @mattheidelbaugh, thank you for the PR!

I've pushed a fix for the failing tagging tests. There were two issues:
 - The ARN was part of the querystring, so the full URL was `/tags/..?arn=arn:aws:...`. I've updated `responses.py` to just use `self._get_param`, as that (also) pulls parameters from the querystring
 - The tags were set as part of the model _and_ in the TaggingService, but when calling the `tag_resource`-method, only the TaggingService was updated. That means that the tags that are part of the `Pipeline`-model are very quickly outdated.
I've refactored this so that the `TaggingService` is the source of truth, and the `Pipeline`-model refers to the TaggingService if it wants to return the list of it's tags.

I've also refactored the ec2 imports a bit. It's big service that does a lot of things in the background, so importing it takes a lot of time. Having localized imports ensures that we only suffer that cost when necesssary.

And finally the State Transitions. First of all: Really cool to see this part of the implementation! The tests are an excellent way of showing why this is so useful, giving users the chance to test the proper behaviour in all stages of the Pipeline lifecycle.

These tests did fail in ServerMode, because the `statemanager.set_transition` command only works when using the decorator. I've refactored this to call the `/moto-api/state-manager/set-transition`-endpoint instead in ServerMode.

bblommers:
> especially where it relates to the extent of the functionality to mimic.

In an ideal world we mimic the full functionality of course, but realistically there are limits to what we can do.

Validating that the subnets/security groups are present and correct is a great start IMO, and also nice to automatically create a VPC endpoint.
If/when someone needs the Cloudwatch LogGroup to exist, or wants even more validation, we can always implement that later.

bblommers (approved):

PR #8162, opened 2024-09-26
batch: fix bug in add_parameters_to_command

giograno:
https://github.com/getmoto/moto/pull/7973 introduced support for parameters in job commands.
However, the `_add_parameters_to_command` function had a bug that resulted in duplicated arguments, when the number of arguments was greater than one.

This was due to the nested list comprehension logic adopted.

For instance, the  given the following job parameters `{'dossier_md5': 'abc', 'event_id': 'E12340060311'}` and the following command `['--dossier_md5', 'Ref::dossier_md5', '--event_id', 'Ref::event_id']` we were given the following output:
```
['--dossier_md5', '--dossier_md5', 'abc', 'Ref::dossier_md5', '--event_id', '--event_id', 'Ref::event_id', 'E12340060311']
```

This PR adjusts the logic accordingly and adds a simple unit test to verify the fix.

### DIFF
--- a/moto/batch/models.py
+++ b/moto/batch/models.py
@@ -638,12 +638,17 @@ class Job(threading.Thread, BaseModel, DockerModel, ManagedState):
         if not self.parameters:
             return command
 
-        new_command = [
-            command_part.replace(f"Ref::{param}", value)
+        return [
+            next(
+                (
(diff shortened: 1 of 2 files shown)


bblommers (approved):
LGTM - thanks for the followup @giograno!

PR #8165, opened 2024-09-27
CloudFront: add get_invalidation support

ligangty:


ligangty:
Seems the server mode test give the error that only contains "404" for not found exception but no specified message. Just give a fix for the tests.

bblommers (approved):
Hi @ligangty, one of the URL's was missing, so Moto didn't know how to handle the `get_invalidation()` call in ServerMode. I've added the appropriate URL and fixed the error handling.

LGTM - thank you for contributing this feature to Moto!

PR #8167, opened 2024-09-27
Unify EC2 resource tags and add new ones

silv-io:
https://github.com/getmoto/moto/pull/8142 introduced the output for EC2 VPC subnets in the resourcestaggingapi service. 

Further requests from users have shown that there are quite a lot more missing which can easily be added.

This PR adds the missing resources from users and unifies the EC2 resource handling to be more easily readable (inspired by how, e.g., RDS is handled in the same file)


bblommers (approved):
Looks great - that simplifies things a lot. Thank you @silv-io!

PR #8168, opened 2024-09-27
Task - Polly: update resources from latest aws polly describe-voices

moyamartin:
I'm updating this because I needed to make `moto` work with the `Gregory` voice which is not included


bblommers (approved):
Makes sense - thanks for updating this @moyamartin!

moyamartin:
Happy to help @bblommers!

PR #8169, opened 2024-09-27
QuickSight support users with special characters

MSSputnik:
In our production environment there are federated users which ontain a / in their name, e.g.: _QuickSightAuthor/my.name@mycompany.com_

This change was created to support users with special charaters in their name.

No new features ware added.


bblommers (approved):
LGTM - nice to see such an extensive test suite.

Thank you for the PR @MSSputnik!

PR #8170, opened 2024-09-27
Fix Cloudformation ECS task definition memory validation

jmsanders:
Very similar to: https://github.com/getmoto/moto/pull/7280

But for cases when register task definition is being called via cloudformation.


bblommers (approved):
LGTM - nice to see all possible scenarios tested here.

Thanks for the PR @jmsanders!

PR #8172, opened 2024-09-28
Scheduler: start_date parameter validation (create_schedule)

vinicius-oa:
There is a missing validation on the `start_date` parameter in `create_schedule` action if the `schedule_expression` is recurrence and `start_date` is today's date, so there might be a `ValidationException` with the message: _**The StartDate you specify cannot be earlier than 5 minutes ago.**_


bblommers (approved):
LGTM! Nice to have a test as well to clearly show the problematic behaviour.

Thank you for this PR @vinicius-oa, and welcome to Moto!

PR #8178, opened 2024-09-29
Scheduler: generalized start_date parameter validation (create_schedule)

vinicius-oa:
[#8172](https://github.com/getmoto/moto/pull/8172) covered only a use case I was working with. Still, I got suspicious about a few behavior nuances and delved a bit into it, then discovered that it is not only the current today date that throws that exception but any other date before it. The mentioned PR's code raises an exception for `start_date` with today's date but higher properties such as hours, minutes, etc (even though I was able to test all of these properties in  localhost, in GH actions for Python it fails for others times smaller properties as minutes, microseconds, etc. _probably due to -j flag for parallel jobs_). This new PR is more robust and covers other cases slipping on the previous one.


bblommers (approved):
This makes a lot of sense. I already saw the old test fail once randomly, and this change should remove the flakiness.

Thanks for the followup @vinicius-oa!

Issue #8180, opened 2024-09-30
Shield describe-subscription error response incorrect


zkarpinski:
The response errror for shield.DescribeSubscription for accounts without a subscription is incorrect.

**Actual:**
```
"SubscriptionNotFoundException",
"No Shield subscription has been created for this account",
```

**Expected:**
`An error occurred (ResourceNotFoundException) when calling the DescribeSubscription operation: The subscription does not exist.`

I'll update this

Ref:
https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/shield/client/describe_subscription.html

PR #8181, opened 2024-09-30
Update Shield exception for no subscription on describe_subscriptions

zkarpinski:
Resolves: #8180 & added test to validate

Code should be "ResourceNotFound"
Message should be "The subscription does not exist."

Reference: https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/shield/client/describe_subscription.html


bblommers (approved):
LGTM - thanks for the fix @zkarpinski!

PR #8193, opened 2024-10-03
Fix BaseResponse.uri_to_regexp

MSSputnik:
Found an issue on how the URIs send by the boto3 client converted into function names.

I have 2 URI:
1. `/accounts/{AwsAccountId}/namespaces/{Namespace}/groups`
2. `/accounts/{AwsAccountId}/namespaces/{Namespace}/users/{UserName}/groups`

These URIs were converted into these regular expressions:
1. `/accounts/(?P<AwsAccountId>.*)/namespaces/(?P<Namespace>.*)/groups`
2. `/accounts/(?P<AwsAccountId>.*)/namespaces/(?P<Namespace>.*)/users/(?P<UserName>.*)/groups`

Unfortunately both request URIs match regular expression 1:
1. `/accounts/1235/namespaces/default/groups`
2. `/accounts/123456789012/namespaces/default/users/authoremail%40example.com/groups`

Investigating on how AWS defines the requestUri in the service.json files I found that there are 2 types of parameters.
1. e.g. quicksight: `/accounts/{AwsAccountId}/namespaces/{Namespace}/users/{UserName}/groups`
2. e.g. workspaceweb: `/portals/{portalArn+}/trustStores`

As you can see in the requestUri from workspaceweb there is a + at the end of the parameter. Checking the code this means, that the parameter can contain a / as it is a Arn. The same I found in the S3 API.

To fix this issue I changed the function BaseResponse.uri_to_regexp to replace the parameter without the + at the end with `(?P<{name}>[^/]+)` to allow any text but no /.  For the parameter with a + I used `(?P<{name}>.+)` to allow any character with minimum length of 1.
I found the usage of .* is not recommended as this causes issues with URIs form the mediastoredata API where `^/$` is ListItems and `^/{Path+}$` is GetObject.

On my development machine most of the test ran successfully. In local mode 2 tests fail and in server mode 3 tests fail. Checking this failuers, I think they are unrelated to this change and therefore I created this pull request.

### DIFF
--- a/moto/core/responses.py
+++ b/moto/core/responses.py
@@ -521,16 +521,16 @@ class BaseResponse(_TemplateEnvironmentMixin, ActionAuthenticatorMixin):
                 return elem.replace("$", r"\$")
 
             # When the element ends with +} the parameter can contain a / otherwise not.
-            no_slash_allowed = False if "+" == elem[-2:-1] else True
+            slash_allowed = True if "+}" == elem[-2:] else False
             name = (
                 elem.replace("{", "")
                 .replace("}", "")
                 .replace("+", "")
                 .replace("-", "_")
             )
-            if no_slash_allowed:
-                return f"(?P<{name}>[^/]+)"
-            return f"(?P<{name}>.+)"
+            if slash_allowed:
+                return f"(?P<{name}>.+)"
+            return f"(?P<{name}>[^/]+)"
 
         elems = uri.split("/")
         regexp = "/".join([_convert(elem) for elem in elems])
--- a/moto/core/responses.py
+++ b/moto/core/responses.py
@@ -521,7 +521,7 @@ class BaseResponse(_TemplateEnvironmentMixin, ActionAuthenticatorMixin):
                 return elem.replace("$", r"\$")
 
             # When the element ends with +} the parameter can contain a / otherwise not.
-            slash_allowed = True if "+}" == elem[-2:] else False
+            slash_allowed = elem.endswith("+}")
             name = (
                 elem.replace("{", "")
                 .replace("}", "")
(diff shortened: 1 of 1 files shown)


MSSputnik:
Hello @bblommers ,

I checked the failing tests and they are also failing without my changes. (Using the current moto main branch).

The first 9 tests have in common that they use the old requests library 0.15. The tests using 0.17.0 succeed.
The 10th test fails due to some docker problems.

Can you please check?
I think I can not do anything about the failing tests.
Thanks

moto/core/responses.py
```diff
@@ -505,35 +505,35 @@ def _dispatch(self, request: Any, full_url: str, headers: Any) -> TYPE_RESPONSE:
...
-        def _convert(elem: str, is_last: bool) -> str:
+        def _convert(elem: str) -> str:
             if not re.match("^{.*}$", elem):
                 # URL-parts sometimes contain a $
                 # Like Greengrass: /../deployments/$reset
                 # We don't want to our regex to think this marks an end-of-line, so let's escape it
                 return elem.replace("$", r"\$")
+
+            # When the element ends with +} the parameter can contain a / otherwise not.
+            no_slash_allowed = False if "+" == elem[-2:-1] else True
```
bblommers, line 524: Any objections to making the naming positive, and having a more direct check? Something like this is more readable IMO:
```
slash_allowed = elem.endswith("+}")
if slash_allowed:
    return ...
return ...
```

bblommers:
Don't worry about the tests @MSSputnik. I've restarted them a few times to check the performance, as regex changes can make quite a big impact. But that all looks good as far as I'm concerned - the tests just happen to be particularly flaky today.

It is a good change though! I did look into this recently as well, because some ARN's weren't being matched properly, but I didn't notice the difference in behaviour between `{param}` and `{param+}`. 
I have a feeling that this could simplify a lot of URL matching logic in various services.

The only question I had was around variable naming essentially - having a negative variable isn't always easy to read, and having a double negative (`not_x = False if..`) is even more difficult to wrap your head around IMO.

MSSputnik:
I updated the condition. 
The reason for using the negative clause were historical. In first I just replaced the is_last condition. No technical cause.

bblommers (approved):
Great - thank you so much @MSSputnik!

Issue #8171, opened 2024-09-27
DynamoDB Scan does not support segments


stephen-christie-ss:
I am making a Lambda which splits a scan into multiple segments for "threaded" processing, and hoped to test the logic with Moto before trying it for real. It seems that Moto doesn't understand the concept, however.

AWS documentation:
https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Scan.html#Scan.ParallelScan

Moto implementation:
https://github.com/getmoto/moto/blob/12cd1868b5e725b8201714f30d5ddfd0324fd141/moto/dynamodb/responses.py#L833

bblommers:
Hi @stephen-christie-ss, welcome to Moto! Marking it as an enhancement to implement this feature.

PR #8194, opened 2024-10-03
Changes did in responses.py

ShreyaShedge:
Addressing issue #8171,

Did This Changes: 
1. Update the `scan()` method in the`responses.py` in `src/dynamodb`.
2. Added support for segmented scans, modifed this method to handle the `Segment` and `TotalSegments` parameters.


bblommers:
This is impossible to review as is, because of all the indentation changes.

Issue #7126, opened 2023-12-15
Find a way to add extensions (i.e. callbacks) to moto


retpolanne:
I have an environment where I want to test the whole EC2 creation flow, from the EC2 api request to a locally-simulated cloud VM with cloud-init. 

I created this tool - https://github.com/retpolanne/cloud-vm-sim - that has an API endpoint that receives a userdata and creates a QEMU VM that can reach to this userdata and run cloud-init. I've extended the EC2 code to call this service on the mock and it works nicely :) 

However, I wanted this to be cleaner - I don't want moto to have this code as this is useful just for me. I wanted to be able to specify a plugins directory and load them and call whatever is in my plugin before the EC2 response.

retpolanne:
My code for cloud-sim call is here - https://github.com/getmoto/moto/commit/882e010f3c4b282821c308e5eb9be957c39962eb

bblommers:
Hi @retpolanne! Having an extensions/plugin system is quite the implementation/maintenance effort, and I'm not sure there would be enough usage to justify this.

As an alternative solution:
Would it work to use mock Moto, and extend the functionality that way? That looks like a simpler solution, and achieves the same result:

```
original_add_instance = InstanceBackend.add_instances
def add_instance_and_run_code(self, image_id, min_count, user_data, security_group_names, **kwargs):
    print("call vm as desired")
    return original_add_instance(self, image_id, min_count, user_data, security_group_names, **kwargs)


@mock_ec2
@mock.patch.object(InstanceBackend, "add_instances", add_instance_and_run_code)
def test_instance_launch_and_terminate():
    client = boto3.client("ec2", region_name="us-east-1")
    # whatever regular test you have...
```

retpolanne:
@bblommers thanks! In my case, I'm using moto as a standalone server for apps running in another language, so that wouldn't really work :c

bblommers:
Ah, I understand. That makes a bit more difficult.

I'll mark it as a feature request, let's see if we can come up with a solution for this.

bpandola:
#10203 is a concrete example of something we wouldn't want to merge into Moto directly, but would be a good candidate for some sort of extension mechanism.

PR #8195, opened 2024-10-03
Plugin

ShreyaShedge:
Addressing Issue #7126,

1. Made a Plugin Directory to add extensions.
2. Loading Plugins Implement a function `load_plugins()` to scan the plugin directory, import each plugin module, and instantiate any classes that implement the plugin interface.
3. Calling Plugin Methods Create a sample plugin `cloud_vm_sim.py `that implements the interface and defines the behavior you want, like calling your cloud VM service with userdata.

### DIFF
--- a/moto/dynamodb/responses.py
+++ b/moto/dynamodb/responses.py
@@ -1,1309 +1,1313 @@
-import copy
-import itertools
-import json
-from functools import wraps
-from typing import Any, Callable, Dict, List, Optional, Union
-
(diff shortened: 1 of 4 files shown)


bblommers:
This does not compile, is not integrated into the rest of the code, can't be extended/customized by the user...

It looks like what an LLM would output. Please don't use code like that as-is (or force others to spend time reviewing it) - it's nowhere near good enough.

Issue #8092, opened 2024-09-06
feature/enhancement: Can moto read cloud formation/cdk and provision resources?


bc-venkata-edara:
## Enhancement

It will be great if moto reads cloudformation or cdk code to provision resources. That way we don't have to rewrite the code for testing. 

As an example I want to test my lambda code, so I just deploy cloud formation using moto locally and trigger it.

bblommers:
Hi @bc-venkata-edara, we do support mocking CloudFormation. We don't have support to provision every possible service/resource, but Lambda Functions are definitely supported.

As the CDK is a separate tool, you might have to run Moto in ServerMode and point the CDK towards it. 
See the docs on how to get started with ServerMode here:
https://docs.getmoto.org/en/latest/docs/server_mode.html

Note that I don't know how to point the CDK towards Moto - I've never used it personally. But most SDK's support some sort of `EndpointURL` configuration, so I imagine it's the same for the CDK.

jabalazs:
Thank you for working on this package!

I am trying to use CDK to provision resources and am stumbling on an error in the bootstrap step. The detailed steps I'm following are:

1. Fresh install of `moto==5.0.26`
2. Run `moto_server`
3. Attempt to bootsrap CDK with `AWS_ENDPOINT_URL=http://localhost:5000 cdk bootstrap -v`

I get the following error:
```
CDK toolkit version: 2.172.0 (build 0f666c5)
Command line arguments: {
  _: [ 'bootstrap' ],
  v: 1,
  verbose: 1,
  lookups: true,
  'ignore-errors': false,
  ignoreErrors: false,
  json: false,
  j: false,
  debug: false,
  ec2creds: undefined,
  i: undefined,
  'version-reporting': undefined,
  versionReporting: undefined,
  'path-metadata': undefined,
  pathMetadata: undefined,
  'asset-metadata': undefined,
  assetMetadata: undefined,
  'role-arn': undefined,
  r: undefined,
  roleArn: undefined,
  staging: true,
  'no-color': false,
  noColor: false,
  ci: false,
  unstable: [],
  'bootstrap-bucket-name': undefined,
  b: undefined,
  'toolkit-bucket-name': undefined,
  toolkitBucketName: undefined,
  bootstrapBucketName: undefined,
  'bootstrap-kms-key-id': undefined,
  bootstrapKmsKeyId: undefined,
  'example-permissions-boundary': undefined,
  epb: undefined,
  examplePermissionsBoundary: undefined,
[...]

PR #8196, opened 2024-10-04
added a feature and implemented in provisioner.py

divyayewale29:
Addressing issues #8092,

Implement this feature in small, manageable steps.

1. created a new file `provisioner.py` in `moto/cloudformation`.
2. Implement CloudFormation template parsing: Add a method to read and parse JSON/YAML CloudFormation templates.
3. implemented this functions `parse_cloudformation()` , `provision_resources()` and created instance `AWS::EC2::Instance`.

### DIFF
--- a/moto/cloudformation/provisioner.py
+++ b/moto/cloudformation/provisioner.py
@@ -1,0 +1,34 @@
+import json
+from moto import mock_ec2, mock_s3
+
+
+def parse_cloudformation(template_path):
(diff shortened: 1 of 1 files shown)


bblommers:
This looks like what an LLM would output. Please don't use LLM code without reviewing it and ensuring that it works (or force others to spend time reviewing it). As is, this does not even compile.

PR #8200, opened 2024-10-04
Step Funtions - Updated State Machine Create, Describe, Update to support logging, tracing & encryption config.

zkarpinski:
Added additional configuration support for the SFN State Machines

### DIFF
--- a/moto/stepfunctions/models.py
+++ b/moto/stepfunctions/models.py
@@ -35,6 +35,8 @@ class StateMachine(CloudFormationModel):
         roleArn: str,
         tags: Optional[List[Dict[str, str]]] = None,
         encryptionConfiguration: Optional[Dict[str, Any]] = None,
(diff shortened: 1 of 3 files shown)


zkarpinski:
@bblommers Not sure how to reach that bit of missed coverage. Wondering if you could take a quick look when you have a chance.

bblommers (approved):
LGTM!

The entire stepfunctions parser has been borrowed from Localstack. I _think_ that they use the `resource_providers` to automatically serialize the response. But none of the files in that folder are used in Moto, so adding code to it will drop the code coverage a bit.

I'll merge it as is - the entire `parser` implementation needs a bit more attention anyway; there are also other code snippets that are no longer used, a ton of functionality that we don't have tests for, etc.

But I'll leave that for another day, there's no need to block this PR. Thank you @zkarpinski!

Issue #7562, opened 2024-04-04
Validate user password in `respond_to_auth_challenge` when the challenge name is `PASSWORD_VERIFIER`


serozhenka:
In Cognito's [respond_to_auth_challenge](https://github.com/getmoto/moto/blob/master/moto/cognitoidp/models.py#L1548-L1571) method I can see that the only things that are validated are the `challenge_name` and the `challenge_responses` payload. The content of `challenge_responses` is not used to validate the user's password, hence no `UnauthorizedException` is raised when logging in with an invalid password. Wish to have support for that soon.

Moto version: 5.0.4

bblommers:
Hi @serozhenka!

From the docs: the `challenge_responses` needs four parts:
```
"PASSWORD_CLAIM_SIGNATURE": "[claim_signature]"
"PASSWORD_CLAIM_SECRET_BLOCK": "[secret_block]"
"TIMESTAMP": [timestamp]
"USERNAME": "[username]"
```

As an enhancement, we can verify whether all four values exist, and we can easily verify that `username` and `secret_block` have the correct value.

I haven't found any official AWS documentation on how to calculate the signature though. There are various open source implementations (like `pycognito` for Python), but without any official documentation, I'm hesitant to add any other checks on these values, without having 100% certainty that there is only one possible value/calculation.

serozhenka:
@bblommers gotcha, I will have a look in my spare time and let you know if I find anything.

serozhenka:
@bblommers I haven't found any official docs either, but there is an [amplify-js](https://github.com/aws-amplify/amplify-js/) repository maintained officially by AWS. And there is the code that does signature generation right [here](https://github.com/amazon-archives/amazon-cognito-identity-js/blob/6b87f1a30a998072b4d98facb49dcaf8780d15b0/lib/CognitoUser.js#L291-L306). It should be fairly easy to reverse-engineer what they are doing there and transfer it into Python (which was probably the way all the SRP auth implementations took). Just to note, I have a Python version of that which we have been using in production for over half a year now and it works smoothly.

If you give the green light, I can try to prepare a draft PR for that in my spare time by carefully reverse-engineering what AWS does in the amplify-js repo.

bblommers:
@serozhenka I don't want to break this for users who use a different signature that happens to be supported as well. So if you want to add the basic checks, that would be very welcome - but I don't want to go any further.

[...]

PR #8204, opened 2024-10-05
Fixed: Validate user password in respond_to_auth_challenge

dubey12345:
Addressing issue #7562,

1.updated the file `moto/cognitoidp/models.py`.
2.changes did in the` CognitoIdpBackend` class.
3.Added the `respond_to_auth_challenge` method as shown in the artifact.
4.Added the new helper methods `_validate_password_verifier`,` _generate_secret_block`, and `_generate_signature`.


bblommers:
> # ... rest of the existing code ...

That's quite a important note @dubey12345. If your LLM tells you to use the `rest of the existing code` in this place, you should probably do so.

If you want to contribute sincerily to Moto, I would suggest you read our documentation first:
https://docs.getmoto.org/en/latest/docs/contributing/installation.html#contributing-installation

PR #8211, opened 2024-10-07
QuickSight finalize user and group handling

MSSputnik:
__Enhancements:__
* Added all missing user handling functions
* Added all missing group handling functions
* Added pagination support for all user and group list functions
* Added search support for groups

__Fixes:__
* Use function to get URL path parameters

### DIFF
--- a/moto/quicksight/responses.py
+++ b/moto/quicksight/responses.py
@@ -29,8 +29,8 @@ class QuickSightResponse(BaseResponse):
         params = json.loads(self.body)
         group_name = params.get("GroupName")
         description = params.get("Description")
-        aws_account_id = self.path.split("/")[-4]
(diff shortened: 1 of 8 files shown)


moto/quicksight/responses.py
```diff
@@ -29,8 +29,8 @@ def create_group(self) -> str:
         params = json.loads(self.body)
         group_name = params.get("GroupName")
         description = params.get("Description")
-        aws_account_id = self.path.split("/")[-4]
-        namespace = self.path.split("/")[-2]
+        aws_account_id = self._get_param("AwsAccountId")
```
bblommers, line 32: Nice! That's a much cleaner way to get the parameters

bblommers (approved):
LGTM - thank you for adding these features @MSSputnik!

PR #8215, opened 2024-10-08
Step Function - Add support for State Machine to resourcegrouptaggingapi

### LINKED ISSUE #8212: Enhancement: Add Step Functions to support resourcegroupstagging
Per the [AWS docs](https://docs.aws.amazon.com/resourcegroupstagging/latest/APIReference/supported-services.html) Step Function support the Resource Groups Tagging API.

It looks like there is a custom implementation for tags for step functions. @bblommers is there a special case with Step Functions or is this a simple use-case to add this support?

zkarpinski:
resolves #8212


bblommers (approved):
LGTM!

PR #8216, opened 2024-10-09
Support custom ids

cloutierMat:
# Motivation

This pr aims to introduce a mechanism by which we can define custom ids in order to be able to create a resource with deterministic ids.

A `ResourceIdentifier` base class will allow us to quickly create objects to manage ids. Each resource can implement their own `generate` method giving us much flexibility when it comes to the type of id created.

The `MotoIdManager` implements an interface over the registered custom ids. It also contains a method `add_id_source` that will allow us the possibility to further expand the behaviour of the id manager.

# Changes

- Implementation of the id manager
- Integration of the `ResourceIdentifier` in `apigateway` and `secretsmanager` as to provide examples of how it can be used.
- Added tests for implemented services.

# Not in scope

Many services can benefit from this upgrade, but only 2 are implemented here to provide a sample of how to use. Follow PRs will spread the functionality to more services. This should help the review process over smaller PRs!

### DIFF
--- a/moto/apigateway/models.py
+++ b/moto/apigateway/models.py
@@ -1312,9 +1312,11 @@ class RestAPI(CloudFormationModel):
     ) -> Deployment:
         if stage_variables is None:
             stage_variables = {}
+        # Since there are no unique values to a deployment, we will use the stage name for the deployment.
+        # We are also passing a list of deployment ids to generate to prevent overwriting deployments.
         deployment_id = ApigwDeploymentIdentifier(
-            self.account_id, self.region_name, name
-        ).generate()
+            self.account_id, self.region_name, stage_name=name
+        ).generate(list(self.deployments.keys()))
         deployment = Deployment(deployment_id, name, description)
         self.deployments[deployment_id] = deployment
         if name:
@@ -2177,7 +2179,7 @@ class APIGatewayBackend(BaseBackend):
             self.account_id,
             self.region_name,
             # The value of an api key must be unique on aws
-            payload.get("value"),
+            payload.get("value", ""),
         ).generate()
         key = ApiKey(api_key_id=api_key_id, **payload)
         self.keys[key.id] = key
--- a/moto/apigateway/models.py
+++ b/moto/apigateway/models.py
@@ -1166,7 +1166,10 @@ class RestAPI(CloudFormationModel):
 
     def add_child(self, path: str, parent_id: Optional[str] = None) -> Resource:
         child_id = create_apigw_id(
-            self.account_id, self.region_name, "resource", parent_id + "." + path
+            self.account_id,
+            self.region_name,
+            "resource",
+            (parent_id or "") + "." + path,
         )
         child = Resource(
             resource_id=child_id,
@@ -2158,7 +2161,7 @@ class APIGatewayBackend(BaseBackend):
                 if api_key.value == payload["value"]:
                     raise ApiKeyAlreadyExists()
         api_key_id = create_apigw_id(
-            self.account_id, self.region_name, "api_key", payload.get("name")
+            self.account_id, self.region_name, "api_key", payload["name"]
         )
         key = ApiKey(api_key_id=api_key_id, **payload)
         self.keys[key.id] = key
--- a/moto/apigateway/models.py
+++ b/moto/apigateway/models.py
@@ -68,7 +68,7 @@ from .exceptions import (
     ValidationException,
     VpcLinkNotFound,
 )
-from .utils import create_id, to_path
+from .utils import create_apigw_id, create_id, to_path
 
 STAGE_URL = "https://{api_id}.execute-api.{region_name}.amazonaws.com/{stage_name}"
 PATCH_OPERATIONS = ["add", "remove", "replace", "move", "copy", "test"]
@@ -789,6 +789,7 @@ class Stage(BaseModel):
 class ApiKey(BaseModel):
     def __init__(
         self,
+        api_key_id: str,
         name: Optional[str] = None,
         description: Optional[str] = None,
         enabled: bool = False,
@@ -798,7 +799,7 @@ class ApiKey(BaseModel):
         tags: Optional[List[Dict[str, str]]] = None,
         customerId: Optional[str] = None,
     ):
-        self.id = create_id()
+        self.id = api_key_id
         self.value = value or "".join(
             random.sample(string.ascii_letters + string.digits, 40)
         )
@@ -846,6 +847,7 @@ class ApiKey(BaseModel):
 class UsagePlan(BaseModel):
     def __init__(
         self,
+        usage_plan_id: str,
         name: Optional[str] = None,
         description: Optional[str] = None,
         apiStages: Any = None,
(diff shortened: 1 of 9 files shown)


moto/utilities/id_generator.py
```diff
@@ -0,0 +1,118 @@
...
+                found_id := id_source(resource_identifier)
+            ) and found_id not in existing_ids:
+                return found_id
+        return None
+
+
+moto_id_manager = MotoIdManager()
+
+
+def moto_id(fn: Callable[..., str]) -> Callable[..., str]:
```
bblommers, line 94: Not really important in the bigger picture, but I guess this could become a `Callable[ParamSpec, str]`. 

That way we can remove the `type: ignore` on the `generate_str_id` method.

https://github.com/getmoto/moto/blob/5cbfacf3b1e4b214cf1c8f143a8b25ecd5213004/moto/core/responses.py#L224

moto/utilities/id_generator.py
```diff
@@ -0,0 +1,118 @@
...
+    _id_sources: List[Callable[[ResourceIdentifier], Union[str, None]]]
+
+    _lock: threading.RLock
+
+    def __init__(self) -> None:
+        self._custom_ids = {}
+        self._lock = threading.RLock()
+        self._id_sources = []
+
+        self.add_id_source(self.get_custom_id)
```
bblommers, line 53: What's the rationale behind having multiple ID sources?

moto/apigateway/models.py
```diff
@@ -1293,7 +1312,11 @@ def create_deployment(
     ) -> Deployment:
         if stage_variables is None:
             stage_variables = {}
-        deployment_id = create_id()
+        # Since there are no unique values to a deployment, we will use the stage name for the deployment.
+        # We are also passing a list of deployment ids to generate to prevent overwriting deployments.
+        deployment_id = ApigwDeploymentIdentifier(
```
bblommers, line 1317: Having a list of ExistingId's essentially means that it's up to the user to call `moto_id_manager.set_custom_id(..)` before every `Deployment` is instantiated.

So in pseudo code:

```
moto_id_manager.set_custom_id(DeploymentIdentifier, 'suffix')
create_stage(name="mystage")
create_deployment()  # deployment_id = account.region.suffix.mystage
create_deployment()  # deployment_id = random identifier, because 'account.region.suffix.mystage' already exists

moto_id_manager.set_custom_id(DeploymentIdentifier, 'suffix2')
create_deployment()  # deployment_id = account.region.suffix2.mystage
```

Am I understanding the flow correctly?

I can't recall another resource where this is the case, but I'm sure this happens more often.. And it is slightly confusing if the actual behaviour differs for specific Resources.

Do you think it makes sense to implement this in a different way? Maybe make it an incremental generator somehow, so that the names become `account.region.deployment1`, `account.region.deployment2`, etc?

moto/utilities/id_generator.py
cloutierMat: Actually, I tried that. I am not super experienced with mypy. But I was not able to get a scenario where a mix of explicit argument and `**kwargs` can be passed jointly. `Concatenate` did not help either.

I think the clarity and the simpler code in the decorator for using named parameters instead of being forced to use `*args, **kwargs` and deconstruct outweighs the loss on `type: ignore`

bblommers: Absolutely, that's fair - it is better readable like this

moto/utilities/id_generator.py
bblommers: Having a deterministic fallback would make sense here, considering `moto_id_generator` is the only place where LS can tweak the behaviour.

tests/test_utilities/test_id_generator.py
```diff
@@ -0,0 +1,134 @@
...
+RESOURCE = "test-resource"
+
+
+@moto_id
+def generate_test_id(
+    resource_identifier: ResourceIdentifier,
+    existing_ids: ExistingIds = None,
+    tags: Tags = None,
+):
+    return GENERATED_ID
```
bblommers, line 27: Bit of a nitpick, and not blocking a merge, but calling this `GENERIC_ID` may be more fitting, as that's a better opposite to `CUSTOM_ID`

tests/test_secretsmanager/test_secretsmanager.py
```diff
@@ -1950,3 +1951,23 @@ def test_update_secret_version_stage_dont_specify_current_stage(secret_arn=None)
         err["Message"]
         == f"The parameter RemoveFromVersionId can't be empty. Staging label AWSCURRENT is currently attached to version {current_version}, so you must explicitly reference that version in RemoveFromVersionId."
     )
+
+
+@mock_aws
+@pytest.mark.skipif(
+    not settings.TEST_DECORATOR_MODE, reason="Can't access the id manager in proxy mode"
+)
+def test_create_secret_custom_id(account_id, set_custom_id):
```
bblommers, line 1960: Another non-blocking nitpick - the `account_id` fixture is only useful for tests that will also be run against AWS.

It's faster (and better readable IMO) to use the static `ACCOUNT_ID` directly.

moto/utilities/id_generator.py
```diff
@@ -0,0 +1,180 @@
...
+        self._lock = threading.RLock()
+        self._id_sources = []
+
+        self.add_id_source(self.get_id_from_tags)
+        self.add_id_source(self.get_custom_id_from_context)
+
+    def get_custom_id(
+        self, resource_identifier: ResourceIdentifier
+    ) -> Union[str, None]:
+        # retrieves a custom_id for a resource. Returns None
```
bblommers, line 74:
```suggestion
        # retrieves a custom_id for a resource. Returns None _if_ ...
```

moto/apigateway/utils.py
```diff
@@ -1,10 +1,81 @@
...
+from moto.utilities.id_generator import ResourceIdentifier, Tags, generate_str_id
+
+
+class ApigwIdentifier(ResourceIdentifier):
+    service = "apigateway"
+
+    def __init__(self, account_id: str, region: str, name: str):
+        super().__init__(account_id, region, name)
+
+    def generate(
```
bblommers, line 17: It would be nice if we could move this method into the `ResourceIdentifier`-class with some sensible defaults, and only override it if necessary. 

It's currently difficult to tell what is sensible/common of course, without implementing many more `ResourceIdentifier`-classes, so that can wait until we have a better grasp of what identifiers typically look like.

bblommers (approved):
LGTM! The actual implementation (writing a `ResourceIdentifier` class) is fairly simple, and there is plenty of docstrings to understand the actual implementation if people are curious.

There are a few nitpicks, but none of them blocking, so feel fre to address that in a followup PR.

Thank you @cloutierMat!

moto/apigateway/utils.py
cloutierMat: Good point, I will keep that in mind when expanding support to other resources.

On one side it would make using the `generate` method a lot simpler if it explicitly defines the property name(s) to pass.

But we will see, it may be that most will require `name` and will return the same `generate_str_id`.

ps. I have addressed the remaining comments in #8230

PR #8218, opened 2024-10-10
Make .json.gz files deterministic

bmwiedemann:
Make `.json.gz` files deterministic to fix reproducible builds.

See https://reproducible-builds.org/ for why this is good.


bblommers (approved):
Makes sense - thank you for the PR @bmwiedemann!

PR #8220, opened 2024-10-10
APIGateway: fix strict filtering for API keys

bentsku:
Following #8202, the behavior was not fully compliant with AWS, as even if not documented at all by AWS, the `nameQuery` parameter also accepts prefixes. This was validated on AWS for both `GetApiKeys` and `GetUsagePlanKeys`.


bblommers (approved):
TIL - thank you for the fix @bentsku!

PR #8223, opened 2024-10-11
SecretsManager - Allow force deletion of already marked-for-delete secret

dfangl:
## Motivation
AWS supports force deletion of already deleted secrets, but this is blocked in moto.

This PR enables this, including a test.

## Changes
* Allow force-deletion of secrets already marked for deletion


bblommers (approved):
Good catch @dfangl - thank you for fixing this!

PR #8225, opened 2024-10-12
ELBv2: remove_tags() now throws a ResourceNotFound Exception

Bharat23:
## Motivation
This pr aims to add parity with `remove_tags` API of `elbv2` that throws exception in certain scenarios. Currently, no exceptions were being thrown from `remove_tags`.

##  Changes
Include logic to throw exceptions similar to `add_tags` by using the method `_get_resource_by_arn` that can check for various NotFound scenarios.
Added test cases for loadbalancer, listener, listener rule and target group to reflect the new behavior. 

## Not in scope
It doesn't currently cover the singular nature of the API. Even though the parameter `ResourceArns` expects a list but only one ARN can be passed. This use case is not covered yet.

### DIFF
--- a/moto/elbv2/models.py
+++ b/moto/elbv2/models.py
@@ -1990,6 +1990,7 @@ Member must satisfy regular expression pattern: {expression}"
 
     def remove_tags(self, resource_arns: List[str], tag_keys: List[str]) -> None:
         for arn in resource_arns:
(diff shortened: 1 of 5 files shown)


bblommers (approved):
LGTM - great to see tests to cover the various resource types.

Thank you for the PR @Bharat23, and welcome to Moto!

PR #8229, opened 2024-10-14
Adds support for UsePreviousTemplate to create_change_set

jamaalscarlett:
Adds missing support for using the UsePreviousTemplate param with UsePreviousValue in the parameter list with calling the cloudformation create_change_set method.
Adds unit tests.
Code clean-up.

### DIFF
--- a/moto/cloudformation/responses.py
+++ b/moto/cloudformation/responses.py
@@ -175,10 +175,11 @@ class CloudFormationResponse(BaseResponse):
         change_set_name = self._get_param("ChangeSetName")
         stack_body = self._get_param("TemplateBody")
         template_url = self._get_param("TemplateURL")
-        if self._get_param("ChangeSetType", "CREATE") == "UPDATE":
-            use_previous_template = (
+        update_or_create = self._get_param("ChangeSetType", "CREATE")
+        use_previous_template = (
                 self._get_param("UsePreviousTemplate", "false").lower() == "true"
             )
+        if update_or_create == "UPDATE":
             stack = self.cloudformation_backend.get_stack(stack_name)
             self.validate_template_and_stack_body()
 
--- a/moto/cloudformation/responses.py
+++ b/moto/cloudformation/responses.py
@@ -154,11 +154,36 @@ class CloudFormationResponse(BaseResponse):
                 return True
         return False
 
+    def validate_template_and_stack_body(self) -> None:
+        if (
+            self._get_param("TemplateBody") or self._get_param("TemplateURL")
+        ) and self._get_param("UsePreviousTemplate", "false").lower() == "true":
+            raise ValidationError(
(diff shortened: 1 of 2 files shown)


moto/cloudformation/responses.py
```diff
@@ -154,11 +154,36 @@ def stack_name_exists(self, new_stack_name: str) -> bool:
...
+
     def create_change_set(self) -> str:
         stack_name = self._get_param("StackName")
         change_set_name = self._get_param("ChangeSetName")
         stack_body = self._get_param("TemplateBody")
         template_url = self._get_param("TemplateURL")
+        if self._get_param("ChangeSetType", "CREATE") == "UPDATE":
+            use_previous_template = (
+                self._get_param("UsePreviousTemplate", "false").lower() == "true"
+            )
```
bblommers, line 181:
```suggestion
        update_or_create = self._get_param("ChangeSetType", "CREATE")
        use_previous_template = (
                self._get_param("UsePreviousTemplate", "false").lower() == "true"
            )
        if update_or_create == "UPDATE":
```
The `update_or_create` variable is already created a few lines below. It would be cleaner to declare the variable once in the beginning, and then use it everywhere

Note that I'm only suggesting this to give an idea - it will probably fail our linter if we push this change as-is

moto/cloudformation/responses.py
```diff
@@ -168,9 +193,15 @@ def create_change_set(self) -> str:
             for item in self._get_list_prefix("Tags.member")
         )
         parameters = {
-            param["parameter_key"]: param["parameter_value"]
+            param["parameter_key"]: stack.parameters[param["parameter_key"]]
+            if param.get("use_previous_value", "").lower() == "true"
+            and use_previous_template
```
bblommers, line 198: I've extracted this variable outside of the if-statement above so that it's always available - not just when updating a changeset. Otherewise there is a situation where Python would throw a `variable not declared` error.

tests/test_cloudformation/test_cloudformation_stack_crud_boto3.py
```diff
@@ -1568,18 +1568,18 @@ def test_describe_change_set(stack_template, change_template):
         ChangeSetType="CREATE",
     )
 
-    stack = cf.describe_change_set(ChangeSetName="NewChangeSet")
+    change_set = cf.describe_change_set(ChangeSetName="NewChangeSet")
```
bblommers, line 1571: Nice! Always good to have proper naming

bblommers (changes requested):
Hi @jamaalscarlett, thank you for the PR! Looks good overall, just some minor nitpicks.

jamaalscarlett:
@bblommers Thanks for the review, I think I've addressed everything.

bblommers (approved):
Great - thank you again @jamaalscarlett!

PR #8231, opened 2024-10-15
Core: fixed typing for conditional expression

cloutierMat:
# Motivation

With it's latest release, `mypy` has improved it's capabilities to [Inferring Unions for Conditional Expressions](https://github.com/python/mypy/blob/master/CHANGELOG.md#inferring-unions-for-conditional-expressions). This created a new issues with `mypy` and [tests failing](https://github.com/getmoto/moto/actions/runs/11337740019) at the linting step.

# Changes

- Added type annotation to prevent inference at declaration (`var = None`) conflict with inference in conditional expression (`var = "string" if cond() else None`) resulting in `expression has type "Union[Any, str]", variable has type "None"` error.


bblommers (approved):
Great, thanks for fixing this @cloutierMat!

PR #8233, opened 2024-10-16
Pinpoint: Fix region in application ARN

viren-nadkarni:
This PR fixes the hardcoded region in Pinpoint Application resource.

This region is now correctly picked up from the request context.


bblommers (approved):
LGTM!

PR #8234, opened 2024-10-16
ResourceGroupsTaggingAPI: fixed tags for Workspaces Web Portals

jamarcelin:


bblommers (approved):
LGTM - thanks for the fix @jamarcelin!

PR #8235, opened 2024-10-17
Add SHA1 and SHA256 algorithms for EC2 fingerprint calculation

nifadyev:
Resolves [localstack issue](https://github.com/localstack/localstack/issues/9337)

@viren-nadkarni is recommended to make changes in `moto` in [this](https://github.com/localstack/localstack/pull/11567#pullrequestreview-2324592555) comment

### DIFF
--- a/moto/ec2/utils.py
+++ b/moto/ec2/utils.py
@@ -3,7 +3,18 @@ import fnmatch
 import hashlib
 import ipaddress
 import re
-from typing import Any, Dict, List, Optional, Set, Tuple, TypeVar, Union
+from typing import (
+    TYPE_CHECKING,
+    Any,
+    Callable,
+    Dict,
+    List,
+    Optional,
+    Set,
+    Tuple,
+    TypeVar,
+    Union,
+)
 
 from cryptography.hazmat.backends import default_backend
 from cryptography.hazmat.primitives import serialization
@@ -13,11 +24,15 @@ from cryptography.hazmat.primitives.asymmetric.ed25519 import (
     Ed25519PublicKey,
 )
 from cryptography.hazmat.primitives.asymmetric.rsa import RSAPublicKey
+from typing_extensions import TypeAlias
 
 from moto.core.utils import utcnow
 from moto.iam import iam_backends
 from moto.moto_api._internal import mock_random as random
 from moto.utilities.utils import md5_hash
+
+if TYPE_CHECKING:
+    HashType: TypeAlias = hashlib._Hash
 
 EC2_RESOURCE_TO_PREFIX = {
     "customer-gateway": "cgw",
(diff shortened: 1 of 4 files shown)


moto/ec2/utils.py
```diff
@@ -753,14 +754,19 @@ def public_key_parse(
     return public_key
 
 
-def public_key_fingerprint(public_key: Union[RSAPublicKey, Ed25519PublicKey]) -> str:
-    # TODO: Use different fingerprint calculation methods based on key type and source
-    # see https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/verify-keys.html#how-ec2-key-fingerprints-are-calculated
+def public_key_fingerprint(
+    public_key: Union[RSAPublicKey, Ed25519PublicKey], is_created: bool = False
```
bblommers, line 758: The naming of `is_created` is a bit confusing for people without any context. Any objections to rename this to something like `use_sha1: bool`?

A comment in the caller function (`random_rsa_key_pair`) can then explain why we're using sha1:

```
# This fingerprint should use sha1 because it's created by EC2
# See https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/verify-keys.html#how-ec2-key-fingerprints-are-calculated
```

bblommers (changes requested):
Hi @nifadyev, thank you for the PR! I just have a nitpick on the naming.

@viren-nadkarni Do you want to review this as well?

moto/ec2/utils.py
nifadyev: Thanks @bblommers , I will make changes but wait for @viren-nadkarni review first. No point to trigger CI because of these small fixes

viren-nadkarni: Alternatively you could consider moving the logic for determining the hash function to a separate function, and pass the hash constructor callable as an argument to this function.

viren-nadkarni (approved):
LGTM :heavy_check_mark: Thanks for fixing this @nifadyev

moto/ec2/utils.py
nifadyev: Thanks @viren-nadkarni . I think your approach will make the logic more complex. If I understood you correctly, then `public_key_fingerprint` will have 3 args - `public_key, hash_constructor and use_sha1`. We need to pass `use_sha1`, because there is no other way to know if key is newly generated. So I would stick to current solution with `use_sha1` (ex `is_created`) arg

viren-nadkarni: No, my thinking was something like this:

```py
def select_hash_algorithm(
        key: RSAPublicKey| Ed25519PublicKey, is_imported: bool
    ) -> SHA1Type | MD5Type | SHA256Type:
    # This returns the hash constructor based on SSH key format and whether key is imported or not
    ...

hash_constructor = select_hash_algorithm(...)

fingerprint: str = public_key_fingerprint(key, hash_constructor)
```

This ensures that functions are modular, decoupled and unit-testable.

nifadyev: My bad, got it. Then it is definitely a better solution. Will use it asap

nifadyev:
hey @bblommers , could you restart `test 3.9` CI stage? Looks like if failed because of connection issue

bblommers (approved):
This looks great - thank you so much @nifadyev!

PR #8240, opened 2024-10-18
MemoryDB: create_cluster() ensure TLSEnabled can be False

archinksagar:
The response for 'TLSEnabled' and 'AutoMinorVersionUpgrade' is always True even when they are set to False. This PR fixes this issue.

### DIFF
--- a/moto/memorydb/models.py
+++ b/moto/memorydb/models.py
@@ -84,7 +84,11 @@ class MemoryDBCluster(BaseModel):
         else:
             self.engine_version = "7.1"  # Default is '7.1'.
             self.engine_patch_version = "7.1.1"
-        self.auto_minor_version_upgrade = auto_minor_version_upgrade if auto_minor_version_upgrade is not None else True
+        self.auto_minor_version_upgrade = (
+            auto_minor_version_upgrade
+            if auto_minor_version_upgrade is not None
+            else True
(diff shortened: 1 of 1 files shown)


moto/memorydb/models.py
```diff
@@ -67,14 +67,17 @@ def __init__(
         self.maintenance_window = maintenance_window or "wed:08:00-wed:09:00"
         self.port = port or 6379  # Default is set to 6379
         self.sns_topic_arn = sns_topic_arn
-        self.tls_enabled = tls_enabled or True
+        self.tls_enabled = True
+        self.acl_name = acl_name
```
bblommers, line 71: Hi @archinksagar, can I suggest a change that's closer to what we had before?

```suggestion
        self.tls_enabled = tls_enabled if tls_enabled is not None else True
        self.acl_name = "open-access" if tls_enabled is False else acl_name
```
Bit of a nitpick, but this is much more readable IMO - makes it more clear what's happening

Same for the `auto_minor_version_upgrad` below.

archinksagar: @bblommers , Thanks for the suggestion, i modified the code accordingly, please take a look

bblommers (approved):
LGTM - thank you @archinksagar!

PR #8249, opened 2024-10-22
Update `botocore` version specifiers

bblommers:
Reverts #8248 
Should work again now that botocore 1.35.46 is out (see https://github.com/boto/botocore/issues/3284)

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -28,7 +28,7 @@ project_urls =
 python_requires = >=3.8
 install_requires =
     boto3>=1.9.201
-    botocore>=1.14.0,<1.35.45
+    botocore>=1.14.0
     cryptography>=3.3.1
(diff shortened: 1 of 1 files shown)


bblommers:
FYI @bpandola - it looks like this is still failing

bpandola:
@bblommers Yeah, there are two issues (at least).  The `botocore` change contained a regression, but the new code they added breaks our `BotocoreStubber`.  I have an idea for a quick fix... not sure if it will work, though.  I'll provide a PR and some more detail in a bit.

bpandola:
@bblommers This is still an issue with `botocore` at the moment.  I've updated my original issue in their repo with additional information.

bblommers:
Thanks for following up on this @bpandola!

PR #8256, opened 2024-10-23
LAMBDA: use arbitrary docker image for lambda tests via MOTO_DOCKER_LAM…

rafcio19:
…BDA_IMAGE env var

Resolves https://github.com/getmoto/moto/issues/8236


bblommers (approved):
Awesome, thank you @rafcio19!

PR #8260, opened 2024-10-25
Organizations: Allow cross-account access

viren-nadkarni:
This PR unlocks cross-account support for AWS Organizations.

After creating an organization and adding a member account to it, it is now possible to retrieve the details of the parent organization using `DescribeOrganization` from the member account.

Plus:
- Calling `CloseAccount` on an already closed account raises
- Calling `RemoveAccountFromOrganization` on an account which does not belong to any raises

A similar pattern was employed for S3 in https://github.com/getmoto/moto/pull/6333

### DIFF
--- a/moto/organizations/exceptions.py
+++ b/moto/organizations/exceptions.py
@@ -11,6 +11,16 @@ class AccountAlreadyRegisteredException(JsonRESTError):
         )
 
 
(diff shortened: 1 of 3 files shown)


bblommers (approved):
LGTM - thank you @viren-nadkarni!

PR #8265, opened 2024-10-28
Lambda: Properly pass dynamodb streams event name

### LINKED ISSUE #8253: DDB Streams: Event name is always insert
https://github.com/getmoto/moto/blob/036c4a1d9270a4307805922b0c0195e60f5b7676/moto/awslambda/models.py#L2285

should be `item.to_json()["event_name"]` as set by https://github.com/getmoto/moto/blob/036c4a1d9270a4307805922b0c0195e60f5b7676/moto/dynamodb/models/table.py#L202-L207 shouldn't it?

dacevedo12:
Fixes #8253

### DIFF
--- a/moto/awslambda/models.py
+++ b/moto/awslambda/models.py
@@ -2286,7 +2286,7 @@ class LambdaBackend(BaseBackend):
             "Records": [
                 {
                     "eventID": item.to_json()["eventID"],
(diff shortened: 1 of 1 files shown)


bblommers (approved):
Awesome, thanks for the fix @dacevedo12!

PR #8266, opened 2024-10-28
Update sagemaker-runtime.rst

david-katz:
sync call was using wrong client


bblommers (approved):
Good catch @david-katz - thank you for the PR!

PR #8290, opened 2024-11-04
Fix #8289 by only requiring typing_extensions in TYPE_CHECKING mode

### LINKED ISSUE #8289: typing_extensions introduced as prod dependency where it should not have been
A recent commit made typing_extensions a required module, which now causes (our) tests to fail because this module is not installed by us, and moto also does not pull it in as a dependency:
`E   ModuleNotFoundError: No module named 'typing_extensions'`

Current workaround is to install the typing_extensions module ourselves, but obviously this should be fixed at the source.

The offending line is this one:
https://github.com/getmoto/moto/blob/32e7efe8cdbad158ca198184a4fd97137824a478/moto/ec2/utils.py#L27

jvdmr:


bblommers (approved):
LGTM - thank you for fixing this!

PR #8295, opened 2024-11-06
Organizations: Extend validation and errors

viren-nadkarni:
This PR brings some improvements to organizations by implementing validations for running certain operations.

- `CreateOrganization` can only be called if the calling account does not already own an org
- `DeleteOrganization` can only be called if the calling account owns an org
- `CreateAccount` and `CloseAccount` can only be called if the calling account owns the org

Related to https://github.com/getmoto/moto/pull/8260


bblommers (approved):
Nice - thank you @viren-nadkarni!

PR #8308, opened 2024-11-12
TechDebt: remove leftover Python2 compatibility

a-detiste:


bblommers:
The `cfnresponse` module might still be used by Python2 Lambda's invoked by users @a-detiste, so I don't think we can just remove this.

Thank you for the PR though!

a-detiste:
Hi.

Yes I m just going through the package list one by one but I don't know everything about everything. 

Greetings

PR #8320, opened 2024-11-15
RDS: implement CopyTags and Tags parameters for copy_db_snapshot()

snordhausen:
Testing on AWS showed that tags are never merged. Instead, AWS uses either the provided *Tags* value or the tags of the original snapshot.
Also, the old code always copied the tags of the source snapshots, but *CopyTags* actually defaults to False.

I'll apply the same change to copy_db_**cluster**_snapshot once this PR is merged. It behaves in the same way as copy_db_snapshot with regards to tags.


tests/test_rds/test_rds.py
```diff
@@ -1076,6 +1076,48 @@ def test_copy_db_snapshots(delete_db_instance: bool):
...
+
+    original_tags = [{"Key": "original", "Value": "snapshot tags"}]
+    conn.create_db_snapshot(
+        DBInstanceIdentifier="db-primary-1",
+        DBSnapshotIdentifier="snapshot",
+        Tags=original_tags,
+    )
+
+    new_tags = [{"Key": "new", "Value": "tag"}]
+    kwargs_and_expected_results = [
```
bblommers, line 1096: Nice! This is a very clean and clear way of showing the different behaviours.

Just one nitpick - can you use this list to parametrize the entire test? That is going to include a bit of overhead, but having separate tests for each use case makes debugging later on much easier (IMHO).

So the resulting test signature would look something like this:
```
@pytest.mark.parametrize("kwargs,expected_result", [..])
def test_copy_db_snapshots_copytags_and_tags(kwargs, expected_result):
    ...
```

bblommers (changes requested):

snordhausen:
@bblommers I'll take at this eventually, just didn't have time yet.

bblommers (approved):
This looks great - thank you for fixing this in Moto @snordhausen!

PR #8322, opened 2024-11-16
Improve IoT Job and JobTemplate

jsuchenia:
Hello

In this PR I'd like to enhance IoT Job object and JobTemplate

### DIFF
--- a/moto/iot/models.py
+++ b/moto/iot/models.py
@@ -24,6 +24,7 @@ from moto.utilities.utils import get_partition
 
 from .exceptions import (
     CertificateStateException,
+    ConflictException,
     DeleteConflictException,
     InvalidRequestException,
     InvalidStateTransitionException,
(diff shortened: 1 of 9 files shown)


moto/iot/models.py
```diff
@@ -2475,5 +2626,63 @@ def update_indexing_configuration(
...
+
+        self.jobs_templates[job_template_id] = job_template
+        return job_template
+
+    def list_job_templates(
+        self, max_results: int, current_token: Optional[str]
+    ) -> Tuple[List[Dict[str, Any]], Optional[str]]:
+        all_job_templates = [_.to_dict() for _ in self.jobs_templates.values()]
+
+        if current_token is None:
```
bblommers, line 2666: Can we use the `@paginate`-decorator to handle pagination? That would simplify the implementation a bit.

See the documentation here: https://docs.getmoto.org/en/latest/docs/contributing/development_tips/utilities.html#paginator

The `list_job_executions_for_thing`-function uses it as well, so you can use that to see it used in practice.

bblommers (changes requested):

moto/iot/models.py
jsuchenia: Fixed - I was using list_jobs() as an input for my change, so I fixed both methods

bblommers (approved):
LGTM - thank you so much @jsuchenia!

PR #8323, opened 2024-11-16
Admin: Add DotNet tests for SQS

bblommers:


bblommers:
@maheshtirumerla Then I don't know what's going on. Moto thinks everything is OK (considering it returns the 200-responses), and as you can see in this PR, creating a queue does work.

PR #8329, opened 2024-11-18
Docs: managed policies are not loaded by default

iamleot:
Hello folks!
I have spent some time in trying to figure out that when using Moto with OpenTofu and I was able to find only an entry in the 5.0.0-alpha3 release note. I think we should list that change as a breaking change so possible other folks hitting that can hopefully find it more easily!

Thanks!

---

Since Moto 5.0.0 for performance reasons the AWS IAM managed policies are no longer automatically loaded anymore by default.

This can break usages relying on AWS IAM managed policies.

Document how to preserve the pre-5.0.0 behavior.


bblommers (approved):
Good catch @iamleot - thank you for this PR!

iamleot:
Thank you @bblommers!

PR #8333, opened 2024-11-18
Fix typo in Lambda docstring

xofbd:
A small typo I encountered when going through the source code.


bblommers (approved):
Nice catch @xofbd, thank you for the PR!

PR #8334, opened 2024-11-19
Fix required permission for HeadObject, which requires GetObject

amitani:
HeadObject is not part of S3 actions for use in IAM permission policies.
https://docs.aws.amazon.com/service-authorization/latest/reference/list_amazons3.html

The reference for HeadObject states that the required permission is GetObject

> General purpose bucket permissions - To use HEAD, you must have the s3:GetObject permission. 

https://docs.aws.amazon.com/AmazonS3/latest/API/API_HeadObject.html

Indeed, `s3:HeadObject` is not a valid action for a bucket policy (screenshot from AWS Bucket Policy Editor).
Screenshot 2024-11-18 at 6 38 56 PM

### DIFF
--- a/moto/s3/responses.py
+++ b/moto/s3/responses.py
@@ -124,7 +124,7 @@ ACTION_MAP = {
         },
     },
     "KEY": {
-        "HEAD": {"DEFAULT": "HeadObject"},
+        "HEAD": {"DEFAULT": "GetObject"},
         "GET": {
             "uploadId": "ListMultipartUploadParts",
             "acl": "GetObjectAcl",


amitani:
I tried running the tests locally but there are many (seemingly unrelated) failures. Given that there isn't `test_response`, I assume that test doesn't cover this module, which is reasonable given that this is to reflect the actual AWS behavior and unit tests won't add much value. Marking this as "ready for review", but if there is more work to be done, please let me know. I have been using `moto` for a while, but this is my first PR for the repo and I'm not very familiar with how this is set up.

bblommers:
Hi @amitani, we do have some similar tests in [test_s3_auth.py,](https://github.com/getmoto/moto/blob/master/tests/test_s3/test_s3_auth.py) but considering it's such a simple change, I don't think we need an additional test for this. I have verified that this now works as it should, so LGTM.

Thank you for the fix!

bblommers (approved):

PR #8338, opened 2024-11-20
EC2: Generate longer resource IDs

viren-nadkarni:
Since July 2018, AWS uses longer 17 character resource IDs for EC2.

See:
- https://www.amazonaws.cn/en/ec2/faqs/
- https://aws.amazon.com/about-aws/whats-new/2018/02/longer-format-resource-ids-are-now-available-in-amazon-ec2/
- https://aws.amazon.com/blogs/compute/longer-resource-ids-in-2018-for-amazon-ec2-amazon-ebs-and-amazon-vpc/

### DIFF
--- a/moto/ec2/utils.py
+++ b/moto/ec2/utils.py
@@ -89,7 +89,7 @@ def random_resource_id(size: int = 8) -> str:
     return "".join(random.choice(HEX_CHARS) for _ in range(size))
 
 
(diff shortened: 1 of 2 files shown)


bblommers (approved):
Yeah, I suppose it's about time we change the format, after 7 years. :slightly_smiling_face: 

Thanks for the PR!

PR #8349, opened 2024-11-30
S3: Sending notification to EventBridge ObjectAcl:Put

ltoppyl:
- Sending a notification to EventBridge when object acl updated.

related https://github.com/getmoto/moto/issues/7363

### DIFF
--- a/moto/s3/models.py
+++ b/moto/s3/models.py
@@ -2585,6 +2585,7 @@ class S3Backend(BaseBackend, CloudWatchMetricProvider):
             bucket_name=bucket_name,
             key_name=key.name,
             acl=multipart.acl,
+            disable_notification=True,  # avoid sending ObjectAcl:Put events here
         )
 
         notifications.send_event(
--- a/moto/s3/models.py
+++ b/moto/s3/models.py
@@ -2223,6 +2223,7 @@ class S3Backend(BaseBackend, CloudWatchMetricProvider):
         bucket_name: str,
         key_name: str,
         acl: Optional[FakeAcl],
+        disable_notification: Optional[bool] = False,
     ) -> None:
         key = self.get_object(bucket_name, key_name)
         # TODO: Support the XML-based ACL format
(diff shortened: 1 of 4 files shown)


moto/s3/models.py
```diff
@@ -2228,6 +2228,13 @@ def put_object_acl(
         # TODO: Support the XML-based ACL format
         if key is not None:
             key.set_acl(acl)
+            bucket = self.get_bucket(key.bucket_name)  # type: ignore
+
+            notify_event_name = (
+                notifications.S3NotificationEvent.OBJECT_ACL_UPDATE_EVENT
+            )
+
+            notifications.send_event(self.account_id, notify_event_name, bucket, key)
```
bblommers, line 2237: The `put_object_acl`-method is also called internally when completing a multipart upload, and I don't think we want an ACL update notification in that scenario.

We had the same problem when implementing `put_object()`, and added a new parameter called `disable_notification`. I think we should have the same here, so that calling `complete_multipart_upload()` does not result in an ACL event.

https://github.com/getmoto/moto/blob/3eb9530c00802f7bb4478e1b5ac0d14e3fda21ce/moto/s3/models.py#L2202

bblommers (changes requested):

moto/s3/models.py
ltoppyl: Thank you for the review!
I have fixed the issue so that the ACL event is no longer triggered when complete_multipart_upload() is called.

(https://github.com/getmoto/moto/pull/8349/commits/bba4e5dff38afe01bd19e44e16817077c46a2810 and https://github.com/getmoto/moto/pull/8349/commits/afb1e2f7aa8e116e84cde3b30140a400666d3c21)

tests/test_s3/test_s3_eventbridge_integration.py
```diff
@@ -201,7 +201,9 @@ def test_complete_multipart_upload_notification():
     )
 
     events = _get_send_events()
-    assert len(events) == 3  # [PutObject, ObjectTagging, CompleteMultipartUpload]
+    assert (
```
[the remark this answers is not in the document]
ltoppyl: @bblommers 
Apologies, you're absolutely right...
I have reverted the test changes. (https://github.com/getmoto/moto/pull/8349/commits/d79023ad994bf6f7ea168968eb626a2d2e0f4108)

I also found a typo and fixed it while I was at it. (https://github.com/getmoto/moto/pull/8349/commits/ced04713f52fbb8cf6767f8194f90f05f1571fe7)

bblommers (approved):
Awesome - thank you for implementing this @ltoppyl!

PR #8361, opened 2024-12-03
Athena: list_query_execution() not filtering by workgroup name

bfbenf:
Added filtering ability to list_query_execution() using the `WorkGroup` param to better align the function to the documentation.

Changed the types for workgroup in `start_query_execution` as they did not reflect what was being passed in during the tests.


bblommers (approved):
LGTM - thank you for fixing this @bfbenf!

PR #8365, opened 2024-12-05
fix(kms): get_public_key now works with alias (name and arn)

### LINKED ISSUE #8364: KMS get_public_key fails with alias
Hi, 

The following test, trying to use `get_public_key` with an alias is failing:

```python
import boto3
from moto import mock_aws

@mock_aws
def test():
    kms = boto3.client("kms", region_name="eu-west-1")
    key = kms.create_key(
        KeyUsage="SIGN_VERIFY",
        KeySpec="ECC_NIST_P256",
    )
    key_id = key["KeyMetadata"]["KeyId"]
    kms.create_alias(
        AliasName="alias/foo",
        TargetKeyId=key_id,
    )
    kms.get_public_key(KeyId="alias/foo")
```
raises the following exception:
```
/Users/xxx/Library/Caches/pypoetry/virtualenvs/motoreport-c38uzrXm-py3.11/bin/python /Users/xxx/PycharmProjects/cebidhem/motoReport/report/main.py 
  File "/Users/xxx/PycharmProjects/cebidhem/motoReport/report/main.py", line 18, in main
    kms.get_public_key(KeyId="alias/foo")
  File "/Users/xxx/Library/Caches/pypoetry/virtualenvs/motoreport-c38uzrXm-py3.11/lib/python3.11/site-packages/botocore/client.py", line 569, in _api_call
  File "/Users/xxx/Library/Caches/pypoetry/virtualenvs/motoreport-c38uzrXm-py3.11/lib/python3.11/site-packages/botocore/client.py", line 1023, in _make_api_call
botocore.errorfactory.NotFoundException: An error occurred (NotFoundException) when calling the GetPublicKey operation: Invalid keyId alias/foo
```

Somehow, the alias isn't recognized as an alias or it doesn't get translated as the actual keyId.

The only place where I can find the error `Invalid keyId ` is in the `_validate_cmk_id(self, key_id: str)`, which seems weird considering I try to use an alias.

I'm using: 
- Python 3.11.1
- moto 5.0.22
- boto3 1.35.75

I can also confirm that it works normally outside of moto with real AWS calls.

cebidhem:
Fixes #8364 

The error provided was coming from the function `_validate_cmk_id()` which obviously raises an error when used with an alias, and if I understood correctly especially since `describe_key()` is using ` any_id_to_key_id()`.

Tests are passing and none seem to be broken.

### DIFF
--- a/tests/test_kms/test_kms_boto3.py
+++ b/tests/test_kms/test_kms_boto3.py
@@ -1713,7 +1713,7 @@ def test_get_public_key_with_alias():
         == key["KeyMetadata"]["SigningAlgorithms"]
     )
     assert (
-        public_key_response_with_name["SigningAlgorithms"]
+        public_key_response_with_arn["SigningAlgorithms"]
         == key["KeyMetadata"]["SigningAlgorithms"]
     )
     assert "EncryptionAlgorithms" not in public_key_response_with_name
--- a/tests/test_kms/test_kms_boto3.py
+++ b/tests/test_kms/test_kms_boto3.py
@@ -1692,6 +1692,36 @@ def test_get_public_key():
     assert "EncryptionAlgorithms" not in public_key_response
 
 
+@mock_aws
+def test_get_public_key_with_alias():
+    client = boto3.client("kms", region_name="eu-west-1")
+    key = client.create_key(KeyUsage="SIGN_VERIFY", KeySpec="RSA_2048")
+    key_id = key["KeyMetadata"]["KeyId"]
+    client.create_alias(
+        AliasName="alias/foo",
+        TargetKeyId=key_id,
+    )
+    alias_arn = client.list_aliases()["Aliases"][0]["AliasArn"]
+
+    public_key_response_with_name = client.get_public_key(KeyId="alias/foo")
+    public_key_response_with_arn = client.get_public_key(KeyId=alias_arn)
+
+    assert "PublicKey" in public_key_response_with_name, (
+        "PublicKey" in public_key_response_with_arn
+    )
+    assert (
+        public_key_response_with_name["SigningAlgorithms"]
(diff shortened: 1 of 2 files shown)


tests/test_kms/test_kms_boto3.py
```diff
@@ -1692,6 +1692,36 @@ def test_get_public_key():
...
+        TargetKeyId=key_id,
+    )
+    alias_arn = client.list_aliases()["Aliases"][0]["AliasArn"]
+
+    public_key_response_with_name = client.get_public_key(KeyId="alias/foo")
+    public_key_response_with_arn = client.get_public_key(KeyId=alias_arn)
+
+    assert "PublicKey" in public_key_response_with_name, (
+        "PublicKey" in public_key_response_with_arn
+    )
```
bblommers, line 1711: The second argument to `assert` is a message to inform the user what went wrong - you can't use it to chain assertions unfortunately.

As an example, this passes:
```
assert True, False
```
Because the `False` statement would only be used as a message if the condition (first argument) would be False.

But this example fails, and gives us the second argument as a description as to why it fails:
```
>>> assert False, True
AssertionError: True
```

TLDR: These assertions should probably just be split up:
```suggestion
    assert "PublicKey" in public_key_response_with_name
    assert "PublicKey" in public_key_response_with_arn
```

bblommers (changes requested):
Hi @cebidhem, thank you for the PR! The change looks good - just a minor comment about the assertions

tests/test_kms/test_kms_boto3.py
cebidhem: Thanks @bblommers for the explanation, I'm not a true python dev _yet_ so it's helpful 👍 

The requested change has been pushed !

bblommers (approved):
LGTM - thank you for fixing this bug @cebidhem!

PR #8360, opened 2024-12-03
directconnect: MACsec associate/disassociate + LAG create/describe

ptgeorge:
* Additional features for Direct Connect
  * LAG create (`aws directconnect create-lag`)
  * LAGs describe (`aws directconnect describe-lags`)
  * MACsec associate (`aws directconnect associate-mac-sec-key`)
  * MACsec disassociate (`aws directconnect disassociate-mac-sec-key`)
* Corrected minor variable typo (co**nnn**ection_id vs co**nn**ection_id)
* Tests added to cover new functionality
* Linter checks passed locally
* Tests passed locally

Notes
* Scaffolding tool (scripts/scaffold.py) was used to generate code stubs
* MACsec associate/disassociate works for both "normal" connections and LAGs.

PR #8366, opened 2024-12-05
directconnect: doc update (coverage)

ptgeorge:
* Minor doc update related to directconnect ( forgotten in #8360 )


bblommers (approved):
Thanks for the follow-up @ptgeorge.

FWIW, the docs are always updated before every release, so it would have happened anyway, but it is always good to do it as soon as possible.

ptgeorge:
Ah ok good to know!  Thx.

PR #8367, opened 2024-12-05
Validate RDS snapshot identifiers

snordhausen:
...and improve the validation of RDS instance identifiers.

### DIFF
--- a/moto/rds/exceptions.py
+++ b/moto/rds/exceptions.py
@@ -212,6 +212,30 @@ class InvalidDBInstanceIdentifier(InvalidParameterValue):
         )
 
 
+class InvalidDBSnapshotIdentifier(InvalidParameterValue):
+    def __init__(self, snapshot_identifier: str, operation_name: str) -> None:
+        operation_to_parameter_name = {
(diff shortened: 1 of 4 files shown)


bpandola (changes requested):
The `operation_name` is a higher level concept/concern than the model should be aware of/care about.  I get that you're trying to account for the differing parameter names that are subject to the same validation, but the best way to do that would be to do the validation checks in the respective `responses.py` methods, where you'll have direct access to the parameter name(s).  Alternatively, you can keep the check in the model and just use the generic error message for all cases.  While it'd be nice to mirror the actual AWS responses as closely as possible, raising the `InvalidParameterValue` error is the only part that's strictly necessary.  Option 1 better matches the real AWS behavior, but Option 2 is less work.  As the PR author, I'll leave it up to you to decide which path to take.

snordhausen:
@bpandola I tried your idea of doing the checks directly in `responses.py` and I agree this looks much cleaner.

bpandola (approved):
Thanks @snordhausen

PR #8368, opened 2024-12-05
ds: Fix describe_trusts with no directory_id provided

zkarpinski:
Fixes bug where no directory_id is provided, type error.

Per docs: _"If no input parameters are provided, such as DirectoryId or TrustIds, this request describes all the trust relationships belonging to the account."_

https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/ds/client/describe_trusts.html


bblommers (approved):
Good catch @zkarpinski, thanks for the fix!

PR #8372, opened 2024-12-06
RDS: allow snapshot ARNs for copy and restore

snordhausen:
`copy_db_snapshot()` and `restore_db_instance_from_db_snapshot()` should accept both names and full ARNs as snapshot identifier. For copy_db_snapshot(), only the SourceDBSnapshotIdentifier can be an ARN, the TargetDBSnapshotIdentifier has to be a name.


moto/rds/models.py
```diff
@@ -1945,9 +1949,28 @@ def modify_db_instance(
...
+    def extract_snapshot_name_from_arn(self, snapshot_arn: str) -> str:
+        arn_breakdown = snapshot_arn.split(":")
+        region_name, account_id, resource_type, snapshot_name = arn_breakdown[3:7]
+        if resource_type != "snapshot":
+            raise InvalidParameterValue(
+                "The parameter SourceDBSnapshotIdentifier is not a valid identifier. "
+                "Identifiers must begin with a letter; must contain only ASCII "
+                "letters, digits, and hyphens; and must not end with a hyphen or "
+                "contain two consecutive hyphens."
+            )
```
snordhausen, line 1961: Looking into the test coverage...

moto/rds/models.py
```diff
@@ -1945,9 +1949,28 @@ def modify_db_instance(
...
+        if region_name != self.region_name or account_id != self.account_id:
+            raise InvalidParameterValue(
+                "Cross account/region snapshot handling is not yet implemented in moto."
+            )
```
bpandola, line 1965: Raising `NotImplementedError` here would be more appropriate, since this is not an AWS error but  internal to moto.

tests/test_rds/test_rds.py
```diff
@@ -1076,6 +1082,29 @@ def test_copy_db_snapshots(delete_db_instance: bool):
...
+        )
+    assert "is not a valid identifier" in ex.value.response["Error"]["Message"]
+
+    cross_account_arn = f"arn:aws:rds:{DEFAULT_REGION}:112233445566:snapshot:snapshot-1"
+    with pytest.raises(ClientError) as ex:
+        conn.copy_db_snapshot(
+            SourceDBSnapshotIdentifier=cross_account_arn,
+            TargetDBSnapshotIdentifier="snapshot-2",
+        )
+    assert "not yet implemented in moto" in ex.value.response["Error"]["Message"]
```
bpandola, line 1105: No need to cover something that's not implemented.  And, per my previous comment, if you raise `NotImplementedError` CodeCov will ignore that line in the coverage calculation.

bpandola (changes requested):

snordhausen:
@bpandola I fixed it as recommended by you. Please review again.

bpandola (approved):
Thanks @snordhausen

PR #8373, opened 2024-12-06
RDS: Implement Managed Master Password Handling

### LINKED ISSUE #8294: Feature: Determine if managed master password is enable on RDS if describe_db_clusters is used
I am currently working on a script which upgrades RDS instances with Blue-Green Deployments. Right now this requires to disable the Managed Master Password before creating the Blue-Green Deployment, so the script includes setting the Masterpassword statically and after completing the Upgrade, switching back to Managed Master Password.

Right now there is no way to tell if these switches work as expected, since `describe_db_clusters` does not return any information whether the master secret is managed or not. In the original boto3 response, `describe_db_clusters` contain the dict `MasterUserSecret` with further information about the Secret.

I happily contribute this feature if desired.

lukaalba:
Hi everyone,

this is my first pull request in this repo, so if something is missing or unclear, let me know.

As requested in this issue https://github.com/getmoto/moto/issues/8294, this MR adds the implementation of the Managed Master Password handling for db_instance and db_cluster resources. This contains the following actions:

- Enabling/Disabling Managed Master User Password has an impact on the db_instance or db_cluster resource response and contains mocked data
- Providing a custom kms key with `MasterUserSecretKmsKeyId`will be respected in moto's response
- Setting `RotateMasterUserPassword` in a `modify_db_instance` or `modify_db_cluster`-call will set the status of the Master User Secret to `rotating`.

To keep things simple, I do not actually create a secret in moto's mocked secretsmanager when enabling Manage Master User Password. If this does not comply to the how moto works, I can update the behavior.

Closes #8294

### DIFF
--- a/moto/rds/models.py
+++ b/moto/rds/models.py
@@ -804,6 +804,10 @@ class DBInstance(CloudFormationModel, RDSBaseModel):
             self.storage_type = DBInstance.default_storage_type(iops=self.iops)
         self.master_username = kwargs.get("master_username")
         self.master_user_password = kwargs.get("master_user_password")
+        self.master_user_secret = None
+        self.manage_master_user_password = kwargs.get(
+            "manage_master_user_password", False
+        )
         self.auto_minor_version_upgrade = kwargs.get("auto_minor_version_upgrade")
         if self.auto_minor_version_upgrade is None:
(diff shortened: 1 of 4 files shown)


bpandola (approved):
Thanks @lukaalba

PR #8374, opened 2024-12-06
Allow RDS restore_db_instance_from_db_snapshot() to custom db_subnet_group

snordhausen:
Previously, this failed with
`TypeError: cannot pickle 'generator' object`
when the code tried to deepcopy the `backend`. But the backend is not needed, anyway, so just skip it when deepcopying.


bpandola (approved):
Thanks @snordhausen

PR #8376, opened 2024-12-06
DynamoDB: deterministic order on query and filter on gsi

Polandia94:
We need a deterministic order on scan and query operations on GSI.

With this two tables created the same, and added the same elements will have always the same order on GSI.

If we have a table:

GSI_PK | GSI_SK | PK
"a" | "a" | "a"
"a" | "a" | "b"

And we do a scan on gsi, we get two items.

If we do a new scan with ExclusiveStartKey with the key of the first item, now we will be sure that will return only one item. Previously the order to calculate exclusiveStartKey and to return on scan/query were different, now, are the same.

Also, on scan on GSI with the same PK and different SK, the order is correct, previously was using primarykey of table to order


Polandia94:
@bblommers I think the test failed because a 503 on httpbin, not related to this PR

bblommers (commented):
Hi @Polandia94! Can you add a non-trivial test that shows the problem? 

With a test it would be easier to see what AWS is doing, and then determine what the correct approach/fix is.

In an ideal world we always mimick AWS, but they do not guarantee order on `scan`, so the exact order that Moto returns should not matter (as long as it is deterministic).

We have quite a few (AWS verified) tests already for various scenarios, if you want some inspiration:
https://github.com/getmoto/moto/blob/master/tests/test_dynamodb/test_dynamodb_scan.py

Polandia94:
@bblommers i added two tests, both are passing on my branch and failing on master. Im retty sure both will pass on AWS, as I'm using this assumptions on production: 

1. The order of a scan on GSI in same PK is ordered by SK. 
2. You can use ExclusiveStartKey on GSI without have the same row returned twice (as happen on master with my test)

bblommers (approved):
@Polandia94 - I can confirm both tests pass on AWS. It took awhile, apologies for that - but I now had a better look at the changes, and they LGTM.

I've pushed some cosmetic changes to make the linter happy, so the moment the build passes (in 30 or so minutes, hopefully) I will merge.

dacevedo12:
This is unfortunately still a problem in the latest version (5.1.4)

PR #8381, opened 2024-12-08
Kafka V1 Support

jamarcelin:


jamarcelin:
@bblommers I am having some issue with some of the endpoints. On line 147 of moto/kafka/responses.py, it is correctly printing (and returning) the response but the response isn't correct in the tests. I think it's a URL issues but not sure.

bblommers:
Hi @jamarcelin! It looks like some of the response-keys simply need to be lowercased. For example:

```diff
diff --git a/moto/kafka/models.py b/moto/kafka/models.py
index 5cd36741d..e2973d5f8 100644
--- a/moto/kafka/models.py
+++ b/moto/kafka/models.py
@@ -231,11 +231,11 @@ class KafkaBackend(BaseBackend):
     ):
         cluster_info_list = [
             {
-                "ClusterArn": cluster.arn,
-                "ClusterName": cluster.cluster_name,
-                "ClusterType": cluster.cluster_type,
-                "State": cluster.state,
-                "CreationTime": cluster.creation_time,
+                "clusterArn": cluster.arn,
+                "clusterName": cluster.cluster_name,
+                "clusterType": cluster.cluster_type,
+                "state": cluster.state,
+                "creationTime": cluster.creation_time,
             }
             for cluster in self.clusters.values()
         ]
```

That change ensures that the `list_clusters_v2()`-method call succeeds in the `test_create_cluster_v2()`. I imagine that the other failures/tests have similar problems/solutions, but I haven't tested that.

This is the first thing I tried, and it happened to work - just because I have run into this problem in the past. The botocore spec is always good place to confirm things like this though, as you can see that they lowercase all `locationName`'s:

https://github.com/boto/botocore/blob/cacd0e9fdba71d1f976069a072576ad74b888bea/botocore/data/kafka/2018-11-14/service-2.json#L2429

jamarcelin:
@bblommers Gotcha, thanks for the help! I fixed the tests and made some changes. The linter is having an issue with broker_node_group_info from FakeKafkaCluster. Any idea how to resolve this?

bblommers:
Hi @jamarcelin! You can run `ruff format moto/kafka` to fix that. I've pushed a fix for now + some typing fixes - I'll leave the test failures to you :slightly_smiling_face:

jamarcelin:
@bblommers Test should be fixed now, just waiting on check to complete. Thanks again!

jamarcelin:
@bblommers Hey, sorry to keep pinging, it seems the tests are failing in server mode? Is this caused by an issue with the urls?

bpandola:
@jamarcelin I pushed a commit to fix the urls.  They have ARNs in them, which contain forward slashes.

jamarcelin:
@bblommers Awesome! Thanks for the help! This looks good to merge to me. Also, is there a specific cadence for releases? Or is there anyway to trigger a release?

bpandola (changes requested):
Can you add tests to cover the lines flagged by Codecov?  It's unclear, for example, whether or not `FakeKafkaCluster.to_dict()` is unreachable code or just not covered by tests.

moto/kafka/models.py
```diff
@@ -0,0 +1,373 @@
...
+        self.serverless_config = serverless_config
+
+    def _generate_arn(self) -> str:
+        resource_type = (
+            "cluster" if self.cluster_type == "PROVISIONED" else "serverless-cluster"
+        )
+        partition = get_partition(self.region_name)
+        return f"arn:{partition}:kafka:{self.region_name}:{self.account_id}:{resource_type}/{self.cluster_id}"
+
+    def to_dict(self) -> Dict[str, Any]:
```
bblommers, line 76: Can we not remove this method altogether @jamarcelin? As far as I can tell, it's not used anywhere

bblommers:
> Also, is there a specific cadence for releases? Or is there anyway to trigger a release?

I try to do them every two weeks - the next one should come this weekend. :slightly_smiling_face:

bblommers (approved):
Great - thank you @jamarcelin!

PR #8391, opened 2024-12-11
DynamoDB: Make it possible to disable delete protection on a table

kiwi-at-billo:
This was previously impossible due to the conditional that required the argument `deletion_protection_enabled` to be truthy for any update to be applied.


bblommers (approved):
Nice catch @kiwi-at-billo - thank you for the fix, and welcome to Moto!

PR #8392, opened 2024-12-12
CognitoIDP: Fix Limit and Filter params for list_users()

### LINKED ISSUE #8363: Limit shouldn't be applied before Filter when listing users
Using the current version of moto, if you run cognito-idp's list_users() with both Limit and Filter keywords, the limit is applied before the filter. So, if you are searching for a user with a particular sub that isn't the first item in the user pool, and you set Limit to 1, then it will always fail to find it because in `moto/cognito-idp/responses.py`, it calls the backend for the unfiltered list of users limited by the Limit keyword, and then filters that.

iknowright:
Fixes #8363

* move `filt` to the backend list_users logic so that the paginator decorator can pick up the right filter users
* Also added the test case for this specific issue

### DIFF
--- a/moto/cognitoidp/models.py
+++ b/moto/cognitoidp/models.py
@@ -1354,10 +1354,54 @@ class CognitoIdpBackend(BaseBackend):
         raise NotAuthorizedError("Invalid token")
 
     @paginate(pagination_model=PAGINATION_MODEL)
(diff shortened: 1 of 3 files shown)


bblommers (approved):
LGTM - thank you for fixing this @iknowright!

PR #8393, opened 2024-12-12
EC2: add SecurityGroupArn to security groups

pinzon:
## Motiviation 
As stated in the [docs](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/ec2/client/create_security_group.html) the methods for creating and describing a Security Group resource should include the attribute SecurityGroupArn.

## Changes
- ARN generation in the SecurityGroup model
- Addition of the attribute in the response templates.

## Testing 
- Extension of an existing test to assert the presence of the attribute


bblommers (approved):
LGTM - thank you @pinzon!

Issue #8394, opened 2024-12-13
EC2 UpdateRoute does not support DestinationPrefixListId correctly


Zethex:
Unable to replace a route via `UpdateRoute` when specifying a destination of type `DestinationPrefixListId`.

The following code reproduces the bug:

- Create neccessary VPC resources
- Create managed prefix list
- Create route that uses prefix list
- Attempt to update route with a different target

```python
import os
import boto3
from moto import mock_aws

with mock_aws():
    os.environ['AWS_ACCESS_KEY_ID'] = 'testing'
    os.environ['AWS_SECRET_ACCESS_KEY'] = 'testing'
    os.environ['AWS_SECURITY_TOKEN'] = 'testing'
    os.environ['AWS_SESSION_TOKEN'] = 'testing'
    client = boto3.client("ec2", region_name="us-east-1")
    ec2 = boto3.resource("ec2", region_name="us-east-1")
    vpc = ec2.create_vpc(CidrBlock="10.0.0.0/16")
    subnet = ec2.create_subnet(VpcId=vpc.id, CidrBlock="10.0.0.0/24")

    main_route_table_id = client.describe_route_tables(
        Filters=[
            {"Name": "vpc-id", "Values": [vpc.id]},
            {"Name": "association.main", "Values": ["true"]},
        ]
    )["RouteTables"][0]["RouteTableId"]
    main_route_table = ec2.RouteTable(main_route_table_id)

    prefix_list = client.create_managed_prefix_list(
        PrefixListName="examplelist", MaxEntries=2, AddressFamily="IPv4"
    )["PrefixList"]

    igw = ec2.create_internet_gateway()
    eni = ec2.create_network_interface(SubnetId=subnet.id)

    main_route_table.create_route(
        DestinationPrefixListId=prefix_list["PrefixListId"], GatewayId=igw.id
    )
    client.replace_route(
        RouteTableId=main_route_table.id,
        DestinationPrefixListId=prefix_list["PrefixListId"],
        NetworkInterfaceId=eni.id,
    )
```
Although this is a supported operation, the below exception is thrown instead:

`botocore.exceptions.ClientError: An error occurred (InvalidParameterValue) when calling the ReplaceRoute operation: There is no route defined for 'pl-952c30ca08963f2d6' in the route table. Use CreateRoute instead.`

PR #8395, opened 2024-12-13
EC2: update_route() now correctly handles DestinationPrefixListId

Zethex:
Updated EC2 update_route to correctly use the decided destination type instead of `destination_cidr_block`.

Also added additional test cases to verify this logic.

This references issue #8394 .


bblommers (approved):
LGTM - thank you for the fix @Zethex!

PR #8407, opened 2024-12-16
Fix `ExposeHeader` XML tag

kangasta:
Remove extra `d` from `ExposeHeader` XML tag in [GetBucketCors](https://docs.aws.amazon.com/AmazonS3/latest/API/API_GetBucketCors.html) response.


bblommers (changes requested):
Interesting - looks like this has been broken forever.

Can you also improve the existing boto3-test and assert that the `ExposeHeader` is now correctly returned?

https://github.com/getmoto/moto/blob/3c4c72e1c9282e920022afaa7ad818472e9fbf50/tests/test_s3/test_s3.py#L2309

We currently only assert that any CORSRules are returned, but we don't validate whether it actually returns the correct values.

bblommers (approved):
Great - thank you so much for the fix @kangasta!

PR #8409, opened 2024-12-16
Add `MOTO_DISABLE_GLOBAL_CORS` for disabling wildcard CORS

kangasta:


bblommers (changes requested):
Hi @kangasta! This makes sense - having a 'strict' mode where CORS is only enabled when configured can be useful, and this looks like the right approach.

Would you mind merging in the latest master? That should hopefully fix the build/linter.

kangasta:
Thank you for the feedback 👍

I rebased this on top of master and added a testcase for the `DISABLE_GLOBAL_CORS` setting.

bblommers (approved):
LGTM - thank you for adding this feature @kangasta!

Issue #7896, opened 2024-07-26
EC2 instance with predefined IP and SubnetId: network interface stays around after instance termination


wandhydrant:
Hello, and many thanks for Moto!

I ran Moto against some use cases and found a few discrepancies between AWS and Moto. This is my first issue. I hope to have found a real bug, and not to have made a mistake.

When deploying EC2 instances where an IP address has been manually specified, and DeleteOnTermination is set to True, I found that Moto keeps the Network Interface around even after termination. When deploying a new EC2 instance with the same IP, this results in an exception "IP already in use". Here is example code to demonstrate this behaviour, ran with AlmaLinux 8 / Python 3.9.19 in a venv with only moto[server] 5.0.11:

```
import boto3
from moto.server import ThreadedMotoServer

server = ThreadedMotoServer()
server.start()
server._server.passthrough_errors = True  # otherwise we just see a 500

ec2_client = boto3.client("ec2", endpoint_url="http://localhost:5000")

existing_subnet_id = ec2_client.describe_subnets()["Subnets"][0]["SubnetId"]

def create_instance():
    instance = ec2_client.run_instances(
        MaxCount=1,
        MinCount=1,
        ImageId="ami-bb9a6bc2",
        InstanceType="t3a.small",
        NetworkInterfaces=[{
            "DeleteOnTermination": True,
            "PrivateIpAddress": "172.31.0.5",
            "DeviceIndex": 0,
            "SubnetId": existing_subnet_id
        }]
    )["Instances"][0]    
    ec2_client.get_waiter("instance_running").wait(InstanceIds=[instance["InstanceId"]])
    return instance

def get_interfaces():
    return ec2_client.describe_network_interfaces(
        Filters=[{"Name": "subnet-id", "Values": [existing_subnet_id]}]
    )["NetworkInterfaces"]
    
assert len(get_interfaces()) == 0

first_instance = create_instance()
assert len(get_interfaces()) == 1
    
ec2_client.terminate_instances(InstanceIds=[first_instance["InstanceId"]])
ec2_client.get_waiter("instance_terminated").wait(InstanceIds=[first_instance["InstanceId"]])

print("After termination:")
print(get_interfaces())
# => a network interface still exists, with "DeleteOnTermination": False

create_instance()
# => Exception("IP already in use")

server.stop()
```

[...]

PR #8415, opened 2024-12-17
EC2: Remove internal Network Interface traces from the Subnet after deletion

wandhydrant:
Hello! Issue #7896 was not completely fixed, the test code in the bug report did not run completely through. The fixed IP address of the deleted instance was still kept in an internal field of the Subnet, so that another instance with the same IP address could not get created.
I tried fixing this, and added a test. All EC2 tests pass, except two that seem unrelated and also fail for me on current master (one in test_availability_zones_and_regions.py and one in test_security_groups.py).
Thanks!


bblommers (approved):
LGTM - thank you for following up on this and fixing this completely @wandhydrant!

PR #8416, opened 2024-12-17
Add basic addition support to cloudwatch.get_metric_data expressions

sleepdeprecation:
I am writing some code that uses the Cloudwatch `GetMetricData` API's math expression feature, specifically in trying to add two metrics together.

This PR adds support for adding two metrics together using the expression feature.

There are some additional features and validation I would like to add to this (specifically, checking that there aren't any other special characters, so that `m1 + m2 - m3` would fail, and the ability to add more than two metrics together, ex `m1 + m2 + m3`), but I wanted to open the PR at this point to see if this feature would be considered useful and check if this seems like a reasonable approach to the problem.

### DIFF
--- a/moto/cloudwatch/metric_data_expression_parser.py
+++ b/moto/cloudwatch/metric_data_expression_parser.py
@@ -1,13 +1,54 @@
 from typing import Any, Dict, List, SupportsFloat, Tuple
 
+from .exceptions import ValidationError
+
 
 def parse_expression(
     expression: str, results: List[Dict[str, Any]]
 ) -> Tuple[List[SupportsFloat], List[str]]:
     values: List[SupportsFloat] = []
     timestamps: List[str] = []
+
+    if "+" in expression:
+        expression = expression.replace(" ", "")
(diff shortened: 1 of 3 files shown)


bblommers (changes requested):
Hi @sleepdeprecation, thank you for the PR!

Looking at [the docs](https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/using-metric-math.html), these expressions can get quite advanced - so I don't think that this is a long-term solution.

If you mostly want the simple addition for now, then I'm willing to merge it (after the comments are addressed), but future extensions should use a proper parser, as that will make it much easier to extend/maintain the solution.

There are quite a few parsers out there, but my preference would be to use `pyparsing`, just because we're already using it in the Glue-module ([see here](https://github.com/getmoto/moto/blob/master/moto/glue/utils.py)) - so if we use that we don't have to add yet another dependency.

bpandola:
Closing due to lack of activity. Feel free to re-open if/when you get back to this and incorporate the initial feedback.

Issue #8413, opened 2024-12-17
Support Aurora DSQL


dacevedo12:
Fairly new service https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/dsql.html

bblommers:
Hi @dacevedo12! @cm-iwata was kind enough to already provide a PR for the `create_cluster()` operation. Which other operations/features do you want to see supported?

dacevedo12:
Thanks a lot! I think generate_db_connect_admin_auth_token and generate_db_connect_auth_token will be amongst the most used

cm-iwata:
In my opinion, implementing `generate_db_connect_admin_auth_token` and `generate_db_connect_auth_token` has little effect since these operations don't communicate with DSQL endpoints. 
S3 `generate_presigned_url` performs a similar process, so its not have a moto's implementation.

bblommers:
Thanks @cm-iwata - I'll close this then. Please let us know if you'd like to see any other features @dacevedo12.

(Keep in mind that Moto's implementation will always be a mock - i.e., you will not be able to actually connect to the endpoint generated by the `generate_.._token` methods.)

PR #8425, opened 2024-12-22
DSQL: implement create_cluster()

cm-iwata:
This PR is part of #8413 

add DSQL `create_cluster()` implementation.

### DIFF
--- a/moto/backend_index.py
+++ b/moto/backend_index.py
@@ -45,6 +45,7 @@ backend_url_patterns = [
     ("directconnect", re.compile("https?://directconnect\\.(.+)\\.amazonaws\\.com")),
     ("dms", re.compile("https?://dms\\.(.+)\\.amazonaws\\.com")),
     ("ds", re.compile("https?://ds\\.(.+)\\.amazonaws\\.com")),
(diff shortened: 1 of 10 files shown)


moto/backend_index.py
```diff
@@ -45,6 +45,7 @@
     ("directconnect", re.compile("https?://directconnect\\.(.+)\\.amazonaws\\.com")),
     ("dms", re.compile("https?://dms\\.(.+)\\.amazonaws\\.com")),
     ("ds", re.compile("https?://ds\\.(.+)\\.amazonaws\\.com")),
+    ("dsql", re.compile("https?://dsql\\.(.+)\\.api\\.aws")),
```
cm-iwata, line 48: DSQL Endpoint is sub domain of api.aws like `dsql.us-east-1.api.aws` not amazonaws.com

bblommers: Interesting. I wonder whether that's an outlier, or whether they're going to host other (existing?) services there as well.

bblommers (approved):
This looks great - thank you @cm-iwata!

PR #8428, opened 2024-12-22
chore: Fixed workflow to prevent terraform tests from failing

cm-iwata:
Currently, ubuntu-latest is specified as the Runner in Git Hub Actions, but ubuntu-latest will soon be changed to point to ubuntu-24.04.

The ubuntu-24.04 Runner does not have a built-in Terraform command, so terraform tests will suddenly fail one day.

- https://github.blog/changelog/2024-11-05-notice-of-breaking-changes-for-github-actions/
- https://github.com/actions/runner-images/issues/10636

I used a forked repository to perform the following tests.

- I explicitly specified ubuntu-24.04 at the `runs-on` and confirmed that the test failed.
    - https://github.com/cm-iwata/moto/actions/runs/12453251643/job/34763159915
- I then added a step to setup Terraform and confirmed that the test was successful.
    - https://github.com/cm-iwata/moto/actions/runs/12453304689/job/34763283277

### DIFF
--- a/.github/workflows/tests_terraform_examples.yml
+++ b/.github/workflows/tests_terraform_examples.yml
@@ -23,7 +23,7 @@ jobs:
           echo "folders=$folders" >> $GITHUB_OUTPUT
 
   test:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.04
     needs: find-jobs
     strategy:
       fail-fast: false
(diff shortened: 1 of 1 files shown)


.github/workflows/tests_terraform_examples.yml
```diff
@@ -23,7 +23,7 @@ jobs:
           echo "folders=$folders" >> $GITHUB_OUTPUT
 
   test:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.04
```
bblommers, line 26: Nitpick - but can we keep this as `latest`? Just to prevent future maintenance :slightly_smiling_face:

cm-iwata: If use the latest label, there is a risk that tests will suddenly fail one day due to compatibility issues.
That's why I explicitly specified runner version in this PR.

However, if compatibility is maintained when upgrading Runner, as you say, unnecessary maintenance will occur when upgrading.

Both approaches have advantages and disadvantages, but I think it's better to follow a unified policy across the project, so we I modified it to use latest.

e93f3f2325364dc52fbd05527c7a9715d480973b

bblommers: Thank you!

PR #8430, opened 2024-12-22
Add short project description in package metadata

edgarrmondragon:
Just a minuscule addition to package metadata, which is shown in places like the PyPI project page (https://pypi.org/project/moto/) or in the output of some tools, such as `poetry show`.


bblommers (approved):
Makes sense! Thanks @edgarrmondragon

PR #8434, opened 2024-12-24
[Bug Fix] Directory Service - change settings init to deep copy

zkarpinski:
The Settings list of dictionaries is a shadow copy, resulting in unexpected results when there are 2+ directories with changed settings.

A deep copy resolves this issue.


bblommers:
Can you merge in the latest master @zkarpinski? That should fix the build - happy to merge afterwards

bblommers (approved):
Great, thank you!

PR #8462, opened 2025-01-03
Feat: Add KMS methods for Key Rotations and MACs

MauriceBrg:
Implemented New Methods for KMS:

- `generate_mac`
- `rotate_key_on_demand`
- `list_key_rotations`
- `verify_mac`

Added these already implemented methods as dummies to the Backend so they show up in the coverage:

- `list_key_policies`
- `generate_random`

(When I ran `scripts/implementation_coverage.py` it also updated a bunch of other rst-files (>50) with new methods, I'm not sure if I should commit those as well, as most are unrelated to KMS)

### DIFF
--- a/moto/kms/models.py
+++ b/moto/kms/models.py
@@ -4,7 +4,7 @@ import typing
 from collections import defaultdict
 from copy import copy
 from datetime import datetime, timedelta
-from typing import Any, Dict, Iterable, List, Optional, Set, Tuple
+from typing import Any, Dict, Iterable, List, Optional, Set, Tuple, Union
 
 from moto.core.base_backend import BackendDict, BaseBackend
 from moto.core.common_models import BaseModel, CloudFormationModel
@@ -99,6 +99,8 @@ class Key(CloudFormationModel):
             f"arn:{get_partition(region)}:kms:{region}:{account_id}:key/{self.id}"
         )
         self.grants: Dict[str, Grant] = dict()
+
+        self.rotations: List[Dict[str, Any]] = []
 
     def add_grant(
         self,
@@ -460,6 +462,11 @@ class KmsBackend(BaseBackend):
     def get_key_policy(self, key_id: str) -> str:
         return self.keys[self.get_key_id(key_id)].policy
 
+    def list_key_policies(self) -> None:
+        # Marker to indicate this is implemented
+        # Responses uses 'describe_key'
+        pass
+
     def disable_key(self, key_id: str) -> None:
         self.keys[key_id].enabled = False
         self.keys[key_id].key_state = "Disabled"
@@ -525,6 +532,11 @@ class KmsBackend(BaseBackend):
             encryption_context=destination_encryption_context,
         )
         return new_ciphertext_blob, decrypting_arn, encrypting_arn
+
+    def generate_random(self) -> None:
+        # Marker to indicate this is implemented
+        # Responses uses 'os.urandom'
+        pass
 
     def generate_data_key(
         self,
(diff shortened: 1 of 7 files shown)


MauriceBrg:
The two lines missing coverage are for the dummy methods to include `list_key_policies` and `generate_random` in the implementation coverage. Should I write some dummy tests or call them from the responses?

moto/kms/models.py
```diff
@@ -714,5 +731,84 @@ def get_public_key(self, key_id: str) -> Tuple[Key, bytes]:
...
+
+        return key_id
+
+    def list_key_rotations(
+        self, key_id: str, limit: int, marker: Union[str, None]
+    ) -> Tuple[List[Dict[str, Union[str, float]]], bool, Union[str, None]]:
+        key: Key = self.keys[self.get_key_id(key_id)]
+
+        # we treat the marker as an index-offset
+        marker = marker or "0"
```
bblommers, line 754: Does AWS use integers as well for markers, or do they use a random string? If it's random, we can use our '@paginate`-marker - that should simplify the implementation a lot.

See the docs here: https://docs.getmoto.org/en/latest/docs/contributing/development_tips/utilities.html#paginator

moto/kms/models.py
```diff
@@ -714,5 +731,84 @@ def get_public_key(self, key_id: str) -> Tuple[Key, bytes]:
...
+            key_id=key_id,
+            mac_algorithm=mac_algorithm,
+            grant_tokens=grant_tokens,
+            dry_run=dry_run,
+        )
+
+        if mac != regenerated_mac:
+            raise KMSInvalidMacException()
+
+        return key_id, True, mac_algorithm
```
bblommers, line 811: Do we need to return anything here? We already know the `key_id` and `mac_algorithm`, and the `is_valid` will always be `True` - so we do not return any new information. Not returning anything would simplify this method

bblommers:
Hi @MauriceBrg, thanks for the PR! Just left two comments.

> (When I ran scripts/implementation_coverage.py it also updated a bunch of other rst-files (>50) with new methods, I'm not sure if I should commit those as well, as most are unrelated to KMS)

That will be either because Moto hasn't been updated in a while, or you're running the script with an outdated version of `botocore`. Either way, it's not necessary - the implementation docs will always be updated when we release a new version.

> The two lines missing coverage are for the dummy methods to include list_key_policies and generate_random in the implementation coverage. Should I write some dummy tests or call them from the responses?

I'm OK with the current approach. Having 100% coverage is a nice to have, but in the end, it doesn't functionally change anything. We have an implementation, and we have tests for it - that's all I care about.

moto/kms/models.py
MauriceBrg: Makes sense, I shouldn't have blindly followed the scaffold-layout :)

I'll update it.

moto/kms/models.py
MauriceBrg: Somehow missed that this exists, but I should have known that I'm not the first to build a pagination scheme.
I have changed the code to use that utility.

MauriceBrg:
Hey @bblommers 

thanks for the suggestions, I have updated the code to address them and removed the update `IMPLEMENTATION_COVERAGE.md` for this Pull Request :)

I've updated boto3 and botocore, now the `scripts/implementation_coverage.py` would only update a handful of files, but I'll leave that to you for the next release.

```text
boto3==1.35.92
botocore==1.35.92
```

bblommers (approved):
Great, thank you @MauriceBrg!

Issue #8420, opened 2024-12-19
Elasticsearch not showing created OpenSearch domains


jflim:
When calling `list_domains` with [boto3: ElasticsearchService ](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/es.html) client, boto3 documentation mentions that the response includes domains of both Elasticsearch or Opensearch.

In my testing, I was able to use the boto3 elasticsearch client to see any OpenSearch domains that were created. But the moto elasticsearch client seems to only track elasticsearch domains.

Here's a test that shows the current results.

```
@mock_aws
def test_describe_opensearch_domain():
    opensearch_client = boto3.client("opensearch", region_name="us-east-2")
    status = opensearch_client.create_domain(DomainName="moto-opensearch")["DomainStatus"]
    assert status["Created"]
    assert "DomainId" in status
    assert "DomainName" in status
    assert status["DomainName"] == "moto-opensearch"

    # ensure that elasticsearch client can describe opensearch domains as well
    es_client = boto3.client("es", region_name="us-east-2")
    assert es_client.list_domain_names()["DomainNames"] != []
```

### Expected Results 
Test passes.

### Actual Results
```
FAILED test_es.py::test_describe_opensearch_domain - assert [] != []
```

jflim:
@bblommers I can work on this! I'm new to working with moto, so wanted to get your thoughts on two solutions that I considered and which would be more appropriate here. Thanks! :

1) Make it so that OpenSearch and ElasticSearch inherit from a more specific class than BaseBackend so they share `domains` as a class attribute. 
2) Make it so that when `list_domain_names` is called from ElasticSearch, the results will be combined with the results of  `list_domain_names` from the OpenSearch client

PR #8464, opened 2025-01-03
Support for both Elasticsearch and OpenSearch domains for list_domain_names

jflim:
This is for #8420 

Went with Option 1 for now but open to changing it to Option 2 to reduce the amount of code being shared.

Reference:
https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/es/client/list_domain_names.html

### DIFF
--- a/moto/es/exceptions.py
+++ b/moto/es/exceptions.py
@@ -24,3 +24,10 @@ class InvalidDomainName(ElasticSearchError):
 class DomainNotFound(ResourceNotFound):
     def __init__(self, domain_name: str):
         super().__init__("Domain", domain_name)
+
+
+class EngineTypeNotFoundException(JsonRESTError):
+    def __init__(self, engine_type: str):
+        super().__init__(
+            "EngineTypeNotFoundException", f"Engine Type not found: {engine_type}"
+        )
--- a/moto/es/exceptions.py
+++ b/moto/es/exceptions.py
@@ -7,7 +7,7 @@ class ElasticSearchError(JsonRESTError):
     code = 400
 
 
-class ResourceNotFound(ElasticSearchError):
+class ResourceNotFoundException(ElasticSearchError):
     code = 409
 
     def __init__(self, resource_type: str, resource_name: str):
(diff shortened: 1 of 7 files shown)


jflim:
This is for #8420.  

I went with option 1 for now so that both OpenSearch and ElasticSearch use a class attribute to track domains, but now I'm witnessing that during unit tests, the backends are not always resetting in time, so there are few times where a call to `list_domain_names` occurs before the `reset` leading to existing created domains leaking into the next unit tests. I'm going to be looking into the reset logic, but wondering if you might have any ideas as to where I could look to troubleshoot. @bblommers 

Thanks!

jflim:
@bblommers Hi! Can I get feedback on this PR? Also, I'm missing coding coverage bot responses. I had swapped my PR from draft so not sure if that caused the issue or if there's another way to remediate.

bblommers:
Hi @jflim, thank you for the PR!

There is a third option here: let the OpenSearch implementation handle everything. Aside from a few ES-specific attributes, the responses should be roughly the same between ElasticSearch and OpenSearch.

We took the same approach when introducing Neptune - that is now completely handled by the RDS service, minus a few edge cases.

I've opened #8491 with an initial implementation. It's not quite ready yet, but it currently does solve the problem you raised, and domain names are shared between ES and OpenSearch.

jflim:
> Hi @jflim, thank you for the PR!
> 
> There is a third option here: let the OpenSearch implementation handle everything. Aside from a few ES-specific attributes, the responses should be roughly the same between ElasticSearch and OpenSearch.
> 
> We took the same approach when introducing Neptune - that is now completely handled by the RDS service, minus a few edge cases.
> 
> I've opened #8491 with an initial implementation. It's not quite ready yet, but it currently does solve the problem you raised, and domain names are shared between ES and OpenSearch.

That makes sense! So as an example of an edge-case, would we move es-specific functions such as  `es.create_elasticsearch_domain` into the opensearch service now?https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/es/client/create_elasticsearch_domain.html

Also, should I close my PR and branch since you have a branch now working on the fix? New to contributing so just trying to understand how I can help this further.

bblommers:
@jflim The `create_elasticsearch_domain`-method can actually be covered by the OpenSearch `create_domain`-method. The only edge case that I found were two ES-specific attributes, `ElasticsearchVersion` and `ElasticsearchClusterConfig`.

I'll close this PR. Thank you again for getting started on fixing this!

Issue #8422, opened 2024-12-20
Feat: Support for S3 Tables API


felixscherz:
AWS recently introduced [Amazon S3 Tables](https://aws.amazon.com/s3/features/tables/) for managing Apache Iceberg tables on AWS.

`boto3` supports S3 Tables starting from version 1.35.74 https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3tables.html

I would be happy to try and create a PR for this feature:)

bblommers:
Hi @felixscherz, that would be very welcome!

Our docs have a section on [how to contribute](https://docs.getmoto.org/en/latest/docs/contributing/index.html) - see especially the [section on how to get started with a new feature](https://docs.getmoto.org/en/latest/docs/contributing/new_feature.html). Let us know if you run into any problems!

PR #8470, opened 2025-01-05
feat: implement mock for S3Tables service

felixscherz:
Hi, this is in regards to #8422.

I used `scripts/scaffold.py` to setup the `s3tables` service. I had to modify some URL patterns for the S3 service, as they overlap with s3tables.

Supported methods:
- [x] `create_table_bucket`
- [x] `get_table_bucket`
- [x] `list_table_buckets`
- [x] `delete_table_bucket`
- [x] `create_namespace`
- [x] `delete_namespace`
- [x] `list_namespaces` 
- [x] `create_table`
- [x] `get_table`
- [x] `list_tables`
- [x] `delete_table`
- [x] `update_metadata_location`
- [x] `get_metadata_location`
- [x] `rename_table`
- [ ] `delete_table_bucket_policy`
- [ ] `delete_table_policy`
- [ ] `get_table_bucket_maintenance_configuration`
- [ ] `get_table_bucket_policy`
- [ ] `get_table_maintenance_configuration`
- [ ] `get_table_maintenance_job_status`
- [ ] `get_table_policy`
- [ ] `put_table_bucket_maintenance_configuration`
- [ ] `put_table_bucket_policy`
- [ ] `put_table_maintenance_configuration`
- [ ] `put_table_policy`

Currently working on supporting the new type of S3 buckets within the existing S3 backend so that data can be written to/read from S3 tables.


tests/test_s3tables/test_s3tables.py
```diff
@@ -0,0 +1,424 @@
...
+        tableBucketARN=arn,
+        namespace="bar",
+        name="baz",
+        metadataLocation="s3://abc",
+        versionToken=resp["versionToken"],
+    )
+    resp = client.get_table_metadata_location(
+        tableBucketARN=arn, namespace="bar", name="baz"
+    )
+    assert "metadataLocation" in resp
```
bblommers, line 288: Can we use `assert metadataLocation == "s3://abc"` here? Or is the exact value different from what we provided?

tests/test_s3tables/test_s3tables.py
```diff
@@ -0,0 +1,424 @@
...
+def test_delete_table_fails_on_version_token_mismatch():
+    client = boto3.client("s3tables", region_name="us-east-2")
+    arn = client.create_table_bucket(name="foo")["arn"]
+    client.create_namespace(tableBucketARN=arn, namespace=["bar"])
+    resp = client.create_table(
+        tableBucketARN=arn, namespace="bar", name="baz", format="ICEBERG"
+    )
+    token = resp["versionToken"]
+
+    with pytest.raises(client.exceptions.ConflictException):
```
bblommers, line 249: Would be useful to also match on the error, to verify we're throwing the correct error here

tests/test_s3tables/test_s3tables.py
```diff
@@ -0,0 +1,424 @@
...
+    client.create_namespace(tableBucketARN=arn, namespace=["bar"])
+    resp = client.create_table(
+        tableBucketARN=arn, namespace="bar", name="baz", format="ICEBERG"
+    )
+
+    resp = client.get_table(tableBucketARN=arn, namespace="bar", name="baz")
+    s3 = boto3.client("s3", region_name="us-east-2")
+
+    bucket_name = resp["warehouseLocation"].replace("s3://", "")
+    with pytest.raises(s3.exceptions.ClientError):
```
bblommers, line 336: Same here for error matching

tests/test_s3tables/test_s3tables.py
```diff
@@ -0,0 +1,424 @@
...
+    resp = client.create_table(
+        tableBucketARN=arn, namespace="bar", name="baz", format="ICEBERG"
+    )
+
+    resp = client.get_table(tableBucketARN=arn, namespace="bar", name="baz")
+    s3 = boto3.client("s3", region_name="us-east-2")
+
+    bucket_name = resp["warehouseLocation"].replace("s3://", "")
+    s3.put_object(Bucket=bucket_name, Key="test", Body=b"{}")
+    with pytest.raises(s3.exceptions.ClientError):
```
bblommers, line 354: Same here for error matching

moto/s3/models.py
```diff
@@ -1876,13 +1884,25 @@ def create_bucket(self, bucket_name: str, region_name: str) -> FakeBucket:
 
         return new_bucket
 
+    def create_table_storage_bucket(self, region_name: str) -> FakeTableStorageBucket:
+        # every s3 table is assigned a unique s3 bucket with a random name
+        bucket_name = f"{str(uuid4())}--table-s3"
```
bblommers, line 1889: We should use `random.uuid4` here, to ensure the names are deterministic ([see the docs](https://docs.getmoto.org/en/latest/docs/configuration/recorder/index.html#deterministic-identifiers))

moto/s3tables/models.py
```diff
@@ -0,0 +1,485 @@
+"""S3TablesBackend class with methods for supported APIs."""
+
+import datetime
+import re
+from base64 import b64decode, b64encode
+from hashlib import md5
+from typing import Dict, List, Literal, Optional, Tuple, Union
+from uuid import uuid4
```
bblommers, line 8: Same here - we should be using `from moto.moto_api._internal import mock_random as random; random.uuid4()`

moto/s3tables/models.py
```diff
@@ -0,0 +1,485 @@
...
+        )
+        if new_table_bucket.arn in self.table_buckets:
+            raise ConflictException(
+                "The bucket that you tried to create already exists, and you own it."
+            )
+        self.table_buckets[new_table_bucket.arn] = new_table_bucket
+
+        return new_table_bucket
+
+    def list_table_buckets(
```
bblommers, line 171: Can we use the `@paginate`-decorator here? That should simplify the implementation.

See the docs: https://docs.getmoto.org/en/latest/docs/contributing/development_tips/utilities.html#paginator

moto/s3tables/models.py
```diff
@@ -0,0 +1,485 @@
...
+                "A namespace with an identical name already exists in the bucket."
+            )
+
+        ns = Namespace(
+            namespace, account_id=self.account_id, created_by=self.account_id
+        )
+        bucket.namespaces[ns.name] = ns
+        return ns
+
+    def list_namespaces(
```
bblommers, line 240: Same here - ideally we paginate using the decorator, instead of rolling our own logic

moto/s3/models.py
```diff
@@ -2644,6 +2677,8 @@ def list_objects(
 
         MOTO_S3_DEFAULT_MAX_KEYS=5
         """
+        if isinstance(bucket, FakeTableStorageBucket):
```
bblommers, line 2680: Does AWS throw the same error here? 

It seems like a useful feature to let users look at the actual data that was stored, but this is the first time that I'm looking at this service, so maybe I'm missing something obvious :slightly_smiling_face:

moto/s3/models.py
```diff
@@ -1903,6 +1923,19 @@ def delete_bucket(self, bucket_name: str) -> Optional[FakeBucket]:
...
+        bucket = self.get_bucket(bucket_name)
+        assert isinstance(bucket, FakeTableStorageBucket)
+        # table storage buckets can be deleted while not empty
+        if bucket.keys:
+            for key in bucket.keys.values():  # type: ignore
+                if isinstance(key, FakeKey):
+                    key.dispose()
+            for part in bucket.multiparts.values():
+                part.dispose()
+        s3_backends.bucket_accounts.pop(bucket_name, None)
```
bblommers, line 1936: This doesn't seem necessary - we're never adding the bucket to `bucket_accounts`

bblommers (changes requested):
Hi @felixscherz, thank you for the PR! Please see my comments. There's quite a few, but I think they're all minor - let me know if you have any questions.

felixscherz:
Thank you for the review! I'll start addressing your comments:)

moto/s3/models.py
felixscherz: AWS does throw the same error, I think it has to do with the way they want users to use the service i.e. not rely on list operations. They also disable delete_object which raises the same exception. I think its important to replicate this behaviour since AWS not supporting it can throw a lot of people off:)

moto/s3tables/models.py
felixscherz: Yes that seems a lot cleaner! I changed all the list endpoints accordingly

moto/s3tables/models.py
felixscherz: I changed both occurrences to use the moto API

tests/test_s3tables/test_s3tables.py
felixscherz: It should be exactly that. Your comment had me check the actual AWS implementation again however. They validate the metadata location to make sure it is part of the storage bucket assigned to the table, so I am going to add this check and then update the test as well:)

moto/s3/models.py
felixscherz: The bucket is added to `bucket_accounts` as part of the `FakeBucket.__init__` method that's why we're removing it here. It does need to be part of `bucket_accounts` to be able to lookup the actual account the bucket is in I think.

bblommers (approved):
This looks great - thank you so much for adding this to Moto @felixscherz!

PR #8475, opened 2025-01-08
RDS: Check if new instance ID already exists

snordhausen:
When creating a new RDS instance, check if an instance with the same instance identifier already exists and raise an error in that case, just like real AWS does.


snordhausen:
The test failure looks unrelated to the code change.

bblommers (approved):
Makes sense - thank you @snordhausen!

PR #8478, opened 2025-01-09
Scheduler: Fix Exception code/format and default timezone

maxhoheiser:
Service: Event Bridge Scheduler

During extending our test coverage (in this PR: https://github.com/localstack/localstack/pull/12114) we noticed, that there is a slight discrepancy between AWS responses and MOTO responses.
This PR introduces the fixes for the response status code, as well as the response message for scheduler groups.
In addition, it also adds tests in moto for these new changes.


bblommers (approved):
LGTM - thank you for fixing these @maxhoheiser!

PR #8461, opened 2025-01-02
Refactor test_rds.py

snordhausen:
Getting rid of almost 700 lines of code, mostly by refactoring the copy&paste parts.

PR #8483, opened 2025-01-12
RDS: refactoring .get() calls in test cases

snordhausen:
Slightly simplifying the test case for RDS, continuation of #8461.

### DIFF
--- a/tests/test_rds/test_filters.py
+++ b/tests/test_rds/test_filters.py
@@ -55,7 +55,7 @@ class TestDBInstanceFilters:
 
         db_instances = self.client.describe_db_instances(
             Filters=[{"Name": "db-cluster-id", "Values": [db_cluster_identifier]}]
-        ).get("DBInstances")
+        )["DBInstances"]
         assert len(db_instances) == 1
         assert db_instances[0]["DBClusterIdentifier"] == db_cluster_identifier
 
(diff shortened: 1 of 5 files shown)


snordhausen:
Test failures seem unrelated to my code changes.

bpandola:
@snordhausen I don't have any problem with this, but just curious: what was the motivation?  Save a few characters and make things a bit more tidy, or is there some larger issue here?

snordhausen:
@bpandola The goal is to reduce code complexity.
```
foo = bar.get("batz")
```
is equivalent to
```
foo = bar["batz"] if "batz" in bar else None
```
I'm replacing that with
```
foo = bar["batz"]
```
which eliminates the complexity of one if-else-statement.




Additionally, most test cases look like
```
foo = bar.get("batz")
assert foo["abc"] == "xyz"
```
If the `batz` key actually does not exist, the error becomes apparent only in line 2 with an error message `NoneType object is not subscriptable`. One needs to read both line 1 and line 2 to understand that error. But with
```
foo = bar["batz"]
assert foo["abc"] == "xyz"
```
line 1 will raise a `KeyError`. To understand the error, one only needs to read a single line of code.

So in summary, the `.get()` increases code complexity and obscures certain errors.

bpandola (approved):

PR #8485, opened 2025-01-13
RDS: Allow copy_db_cluster_snapshot() for deleted cluster

snordhausen:
Make it possible to call `copy_db_cluster_snapshot()` even when the original cluster was already deleted. A similar bug was fixed in #7392, but for normal RDS instances. This also fixes it for RDS clusters.

While implementing the new code, I wondered what happens when a snapshot with `SnapshotType=="automated"` gets copied. Does the copy have the same property (because it was copied) or does it have `SnapshotType=="manual"` because the snapshot was manually created? Turns out the latter is true, and this is correctly implemented by moto. But there weren't any test cases for this, so the second commit adds them.

### DIFF
--- a/moto/rds/models.py
+++ b/moto/rds/models.py
@@ -2537,6 +2537,25 @@ class RDSBackend(BaseBackend):
             db_cluster_identifier, db_snapshot_identifier, snapshot_type="automated"
         )
 
+    def _create_db_cluster_snapshot(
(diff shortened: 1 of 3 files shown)


snordhausen:
Test failures seem unrelated to my code changes.

bpandola (approved):

PR #8486, opened 2025-01-13
Exclude bad `responses` version

bpandola:
Ref: https://github.com/getsentry/responses/issues/751

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -35,7 +35,7 @@ install_requires =
     xmltodict
     werkzeug>=0.5,!=2.2.0,!=2.2.1
     python-dateutil<3.0.0,>=2.1
-    responses>=0.15.0
+    responses>=0.15.0,!= 0.25.5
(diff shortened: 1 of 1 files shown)


bblommers:
Thanks for fixing the build @bpandola! Looks like they just released a new version that fixes it :slightly_smiling_face: 

https://github.com/getsentry/responses/releases/tag/0.25.6

bpandola:
@bblommers Ha, I guess I could've just waited a few minutes and saved myself some time... 😄 

Looks like there might still be a build issue, though, with the Terraform ACM tests.  I don't really know how any of that works, so hopefully you've got an idea on how to address that one.

bblommers:
> Looks like there might still be a build issue, though, with the Terraform ACM tests. I don't really know how any of that works, so hopefully you've got an idea on how to address that one.

Thanks for the heads up! I'll have a look later.

PR #8495, opened 2025-01-16
Ensure compatibility with latest `botocore` S3 client customizations

bpandola:
The latest version of `botocore` adds some customizations to the S3 client that break some of our S3 tests:

* S3 client behavior is updated to always calculate a CRC32 checksum by default for operations that support it (such as PutObject or UploadPart), or require it (such as DeleteObjects). 
* Botocore will no longer automatically compute and populate the Content-MD5 header.

This PR removes assertions around the default checksum behavior (as it now varies across `botocore` versions).  It also manually supplies the `ContentMD5` parameter for API calls that require it, as `botocore` no longer automatically computes it.

This PR also contains minor `request.body` processing tweaks for S3 as well as during `proxy` or `recording` mode, due to the new `aws-chunked` encoding and the `AwsChunkedWrapper` object that now wraps some S3 request bodies sent from boto3/botocore. 

> [!NOTE]  
> This PR fixes the `moto` build breakages that occurred as a result of the new "default integrity protections" introduced in the AWS SDK for Python v1.36.0.  It also maintains existing compatibility with earlier SDK versions.  There may be additional issues encountered by users of `moto` as a result of these AWS changes, which will have to be addressed in subsequent PRs as they arise.

Ref: https://github.com/boto/boto3/issues/4392
Ref: https://github.com/boto/botocore/commit/590103bde8c18174afb9d17970c7e3d2520ca25a

### DIFF
--- a/moto/moto_proxy/proxy3.py
+++ b/moto/moto_proxy/proxy3.py
@@ -155,12 +155,12 @@ class ProxyRequestHandler(BaseHTTPRequestHandler):
             return
 
         req_body = b""
-        if "Content-Length" in req.headers:
-            content_length = int(req.headers["Content-Length"])
-            req_body = self.rfile.read(content_length)
-        elif "chunked" in self.headers.get("Transfer-Encoding", ""):
+        if "chunked" in self.headers.get("Transfer-Encoding", ""):
             # https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Transfer-Encoding
             req_body = self.read_chunked_body(self.rfile)
+        elif "Content-Length" in req.headers:
+            content_length = int(req.headers["Content-Length"])
+            req_body = self.rfile.read(content_length)
         if self.headers.get("Content-Type", "").startswith("multipart/form-data"):
             boundary = self.headers["Content-Type"].split("boundary=")[-1]
             req_body, form_data = get_body_from_form_data(req_body, boundary)  # type: ignore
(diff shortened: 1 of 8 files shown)


bblommers (approved):
Awesome, thank you for tackling this @bpandola!

jeking3:
Caused by: https://github.com/boto/boto3/issues/4392

moto/moto_proxy/proxy3.py
bpandola, line 1: The diff here looks more complicated than it truly is.  I just reordered the `if` statements so that the `chunked` encoding takes precedence over the `Content-Length` header.  The code within these conditionals was not changed at all. 

This change was necessary because some of the intercepted S3 requests from boto3/botocore now default to a chunked encoding (that also include a `Content-Length` header).

bpandola:
> Where Content-MD5 was required before, it is not any more, but only on S3 proper. Third party S3-like services will still require it. I'm not sure the third party S3-like services support trailing checksums either, and certainly won't support the new CRC64 mode.
> 
> Do we need a way for whatever is running moto to specify the "flavor" of S3 it is emulating now?

It's not clear to me if anything changed server-side at AWS.  The changes in question all seem to affect the client requests generated by boto3/botocore.  As such, this PR just updates `moto` to deal with the new client request format (for some S3 API endpoints).  This PR does not make any modifications to any `moto` responses.

If you run into a specific issue with `moto`, related to this change or not, please open a new issue with the details. Thanks!

bblommers:
I've released `moto==5.0.27.dev16` if people want to try this out

PR #8499, opened 2025-01-18
Core: Allow replace/remove of mocked responses

### LINKED ISSUE #8449: responses_mock.remove() causes infinite loop
`moto.core.responses_custom_registry.CustomRegistry` does not implement `remove()` method,  code will fallback to inherited `responses.registries.FirstMatchRegistry`

test env is current moto master (release 5.0.25) using `make init`

here is a test case which fails due to infinite loop
```
import moto
import requests
import responses
from unittest import TestCase


class TestBad(TestCase):

    @responses.activate
    @moto.mock_aws
    def test_bad(self):
        from moto.core.models import responses_mock

        responses_mock.get(url="https://example.com/", body="test")
        resp = requests.get("https://example.com/")
        assert resp.text == "test"

        # causes infinite loop
        responses_mock.remove("GET", "https://example.com/")


if __name__ == "__main__":
    import unittest
    unittest.main()
```

suggested fix
```
    def remove(self, response: responses.BaseResponse) -> list[responses.BaseResponse]:
        removed_responses = []
        registered = self._registered[response.method]
        while response in registered:
            registered.remove(response)
            removed_responses.append(response)
        return removed_responses
```

bblommers:
Fixes #8449

### DIFF
--- a/moto/core/responses_custom_registry.py
+++ b/moto/core/responses_custom_registry.py
@@ -32,6 +32,23 @@ class CustomRegistry(responses.registries.FirstMatchRegistry):
         self._registered[response.method].append(response)
         return response
 
+    def replace(self, response: responses.BaseResponse) -> responses.BaseResponse:
+        registered = self._registered[response.method]
+        try:
+            index = registered.index(response)
+        except ValueError:
+            raise ValueError(f"Response is not registered for URL {response.url}")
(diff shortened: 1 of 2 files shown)


ralphie0112358 (commented):
👍🏻

PR #8506, opened 2025-01-19
Remove `ParamValidationError` test assertions

bpandola:
The `boto3.client` has a limited set of validation checks that it performs before serializing requests.  If one of these validation checks fails, a `ParamValidationError` exception is raised.  This exception is caught and raised entirely within the AWS client code--requests that result in this exception are never passed along to an AWS (or `moto`) backend.

There are a handful of `moto` test cases that are asserting against this exception being raised.  These tests are covering `botocore` functionality and do not actually exercise or cover any `moto` code. This PR removes these test cases.

The `secretsmanager` backend does include some server-side validation checks.  There were a couple of placeholder test cases that looked like this:

```
# Test is intentionally empty. Boto3 catches and raises ParamValidationError before Moto can see it.
# test_server actually handles this error.
assert True
```

This PR fills in these test cases with proper setup and assertions by using a `boto3.client` configured with `parameter_validation=False`, allowing the requests to be passed as-is to the `moto` backend for validation.


jflim:
Thanks for the check, @bpandola ! I'll remember to not add these tests in the future.

jflim (approved):

bpandola:
@jflim Hey, no big deal.  I was gonna drop a comment on your PR, but it got merged before I was able.  I appreciate that you write good PRs with thorough test coverage.  In the future, if you want to write (and test) server-side validation code in `moto`, you can disable the parameter validation in the `boto3.client` like so:

```

my_config = Config(
    parameter_validation=False,
)
client = session.client(
    "elasticache", region_name="us-west-2", config=my_config
)
```

tests/test_secretsmanager/test_secretsmanager.py
```diff
@@ -1096,11 +1108,32 @@ def test_rotate_secret_rotation_lambda_arn_too_long():
...
-    # test_server actually handles this error.
-    assert True
+@pytest.mark.parametrize(
+    "days",
+    [
+        pytest.param(0, id="below min"),
+        pytest.param(1001, id="above max"),
+    ],
+)
+def test_rotate_secret_rotation_period_validation(days):
```
bblommers, line 1118: Nice! That is much better readable than the previous test

bblommers (approved):
Looks all good to me @bpandola - appreciate the cleanup :pray:

PR #8507, opened 2025-01-20
Resolve `pytest` warning

bpandola:
Pytest will attempt to collect test cases from any class whose name starts with "test", so we have to explicitly mark this class as not a test.


bpandola:
I deliberately cancelled CI for this trivial change, so as not to consume resources unnecessarily (docker rate limits are thing, etc.).

PR #8515, opened 2025-01-22
Elasticsearch - Expose the tagging functionality

### LINKED ISSUE #8514: [Elasticsearch] Add responses for tagging using OpenSearch dispatch
Extending work done in #8491, to expose `list_tags` and `add_tags` for Elasticsearch.

jflim:
Fixes #8514


bblommers (changes requested):
Hi @jflim, would you mind adding some tests for this? It looks good to me, but it would still be nice to have some validation that this is correct

jflim:
> Hi @jflim, would you mind adding some tests for this? It looks good to me, but it would still be nice to have some validation that this is correct

The test actually found an issue with the urls, so I'm glad I tested now 🌞 Retested with OpenSearch tests and tagging tests for opensearch and es.

jflim:
I'm a bit confused as to why this fix works though. I believe I observed that `remove_tags` for OpenSearch was also matching the `/2021-01-01/tags/?` regex but from my understanding the `?` was to handle if a URL ends with `/` or not.

bblommers (approved):

bblommers:
> I'm a bit confused as to why this fix works though. I believe I observed that `remove_tags` for OpenSearch was also matching the `/2021-01-01/tags/?` regex but from my understanding the `?` was to handle if a URL ends with `/` or not.

Looks like this is because we don't do a `fullmatch`. So `../tags` is actually interpreted as `../tags.*` - which also matches `../tags-removal`.

Thank you for the PR!

PR #8518, opened 2025-01-23
Security Hub - get_findings and batch_import_findings

### LINKED ISSUE #8493: Security Hub Service Enhancement: GetFindings
## Enhancement

Looks like Security Hub has not been mocked yet. Starting with `GetFindings`

Singha22:
Resolves #8493 

`get_findings` contains code for only these **2** parameters **out of the 4 parameters**
```python
next_token: Optional[str] = None,
max_results: Optional[int] = None,
```


moto/securityhub/models.py
```diff
@@ -0,0 +1,118 @@
...
+                    raise InvalidInputException(
+                        op="GetFindings",
+                        msg="MaxResults must be a number between 1 and 100",
+                    )
+            except ValueError:
+                raise InvalidInputException(
+                    op="GetFindings", msg="MaxResults must be a number greater than 0"
+                )
+
+        paginator = Paginator(
```
bblommers, line 50: It looks like we can just use the `@paginate`-decorator on the method here, that should simplify the implementation a bit.

Please see the docs here: https://docs.getmoto.org/en/latest/docs/contributing/development_tips/utilities.html#paginator

bblommers (changes requested):
Hi @Singha22, thank you for getting started on this new service!

The implementation looks mostly OK to me - I just have one request to simplify the pagination-implementation a bit.

bblommers (approved):
Awesome - thank you @Singha22!

PR #8530, opened 2025-01-28
DynamoDB: return ConsumedCapacity on delete_item

rick-soptim:
Extends the changes from https://github.com/getmoto/moto/pull/4241 to the DDB DeleteItem endpoint.

DeleteItem-Responses should include an ConsumedCapacity element in the same way as Scan-, Query-, GetItem- and PutItem-Responses already do: https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_DeleteItem.html#DDB-DeleteItem-response-ConsumedCapacity


bblommers (approved):
Makes sense! Thank you for fixing this @rick-soptim

PR #8531, opened 2025-01-28
Organizations: return organization roots for requests made by other accounts within the organization

k-a-il:
Default account is linked to the organization, whereas other accounts within the organization do not have this reference. This PR fixes this issue.

Once an organization is created and a member account is added, the member account can now retrieve the details of the root account by using the `ListRoots` API.

Plus, if a member account is not part of any organization, the `AWSOrganizationsNotInUseException` exception will be raised, as described in the [AWS documentation](https://docs.aws.amazon.com/organizations/latest/APIReference/API_ListRoots.html).

This PR re-uses approach introduced in the #8260

### DIFF
--- a/moto/organizations/models.py
+++ b/moto/organizations/models.py
@@ -482,7 +482,16 @@ class OrganizationsBackend(BaseBackend):
         self._reset()
 
     def list_roots(self) -> Dict[str, Any]:
-        return dict(Roots=[ou.describe() for ou in self.ou if isinstance(ou, FakeRoot)])
+        if self.org:
+            return dict(Roots=[ou.describe() for ou in self.ou if isinstance(ou, FakeRoot)])
+
+        if self.account_id in organizations_backends.master_accounts:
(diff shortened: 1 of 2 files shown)


bblommers (changes requested):
The change LGTM @k-a-il - looks like the linter is currently failing though:
https://github.com/getmoto/moto/actions/runs/13013182740/job/36300800416?pr=8531

Looks like this is a simple matter of running `ruff format moto/organizations/models` to fix it

k-a-il:
> The change LGTM @k-a-il - looks like the linter is currently failing though: https://github.com/getmoto/moto/actions/runs/13013182740/job/36300800416?pr=8531
> 
> Looks like this is a simple matter of running `ruff format moto/organizations/models` to fix it

I forgot to run the linter! I’ve resolved it by running `make format` and `make lint`. It should be fixed now!

bblommers (approved):
Awesome, thank you @k-a-il. Welcome to Moto!

PR #8532, opened 2025-01-28
CloudHSMv2 Integration

Singha22:
Resolves https://github.com/getmoto/moto/issues/8571


moto/cloudhsmv2/models.py
```diff
@@ -0,0 +1,312 @@
...
+    ) -> Tuple[List[Dict[str, str]], Optional[str]]:
+        if resource_id not in self.tags:
+            return [], None
+
+        tags = sorted(self.tags.get(resource_id, []), key=lambda x: x["Key"])
+
+        if not max_results:
+            return tags, None
+
+        if next_token:
```
bblommers, line 145: What is the reasoning behind manipulating the `next_token`?

moto/cloudhsmv2/models.py
```diff
@@ -0,0 +1,312 @@
...
+                if key == "clusterIds":
+                    clusters = [c for c in clusters if c.cluster_id in values]
+                elif key == "states":
+                    clusters = [c for c in clusters if c.state in values]
+                elif key == "vpcIds":
+                    clusters = [c for c in clusters if c.vpc_id in values]
+
+        clusters = sorted(clusters, key=lambda x: x.create_timestamp)
+
+        if not max_results:
```
bblommers, line 251: This doesn't look right - even if we don't have `max_results` specified, we may have a `next_token` that we should honor.

Can we use the `@paginate`-decorator here instead?

bblommers (changes requested):
Thanks for the PR @Singha22.

It looks like the ServerMode tests are currently failing - let me know if you need any help looking into them

moto/cloudhsmv2/models.py
Singha22: The token has to be converted into base64 so without it, we get something like `eyJwYXJhbWV0ZXJDaGVja3N1bSI6IDQyMzg4OTQxMTIsICJ1bmlxdWVBdHRyaWJ1dGVzIjogIlByb2plY3QifQ` instead of `eyJwYXJhbWV0ZXJDaGVja3N1bSI6IDQyMzg4OTQxMTIsICJ1bmlxdWVBdHRyaWJ1dGVzIjogIlByb2plY3QifQ==`. Is this not the right approach?

bblommers: That explains **what** it does - I wanted to know **why** :slightly_smiling_face: 

The `nextToken` that AWS returns is usually a random string, and the user should not care about the exact format of it. So my question was more: why do we need to manipulate the token, and can we not use the token that the `Paginator` returns?

If we don't have to manipulate the token, then we can rip this all out and replace it with the `@paginate` decorator instead, which would be much nicer (IMHO).

If there is a good reason we must manipulate the token, then that's fine as well - I would just want to know why.

Singha22: Hmm yeah I see what you're saying. I was trying to keep it as similar to AWS but I'll just leave it as a TODO for the future if that's fine. I left a comment like in the other services.

Singha22:
Hey @bblommers do you know why I keep on getting these errors in the server tests?

```
E           botocore.exceptions.ClientError: An error occurred (500) when calling the TagResource operation (reached max retries: 4): <!doctype html>
E           <html lang=en>
E           <title>500 Internal Server Error</title>
E           <h1>Internal Server Error</h1>
E           <p>The server encountered an internal error and was unable to complete your request. Either the server is overloaded or there is an error in the application.</p>
```

bblommers:
Hi @Singha22! I've taken the liberty to push the fix already.

The tests failed because the URL's were incomplete. When running Moto in ServerMode, we don't have access to the hostname (because everything is localhost), so instead we use the signing-name specified in the Authorization-header. This usually corresponds exactly to the service-name. Problem is that botocore uses `cloudhsm` as the signing name, and Moto takes that and tries to find a service with `cloudhsm.amazonaws.com` as the URL.

https://github.com/boto/botocore/blob/1941b2b67bd033957c5ddd4855868fe7e010a174/botocore/data/cloudhsmv2/2017-04-28/service-2.json#L13

So as a workaround, I just added `cloudhsm.amazonaws.com` as part one of the 'acceptable' URLs. The original CloudHSM service uses the same signing-name, so if we also build support for that one, we will need to find a way to differentiate. That's a problem for the future though - for now, this will do.

The other problem was reading the parameters. The approach you took (`list(self._get_params().keys())[0]`) doesn't always work, and sometimes results in invalid JSON. All information is present in `self.body` though, in JSON format, so we can use that.

Oh, and the third change: I also took the liberty of deleting `test_server.py`. Manually creating HTTP requests and verifying that Moto accepts them can be very useful if you want to test something that can't be tested with regular `boto3`-methods - but from the looks of it, all scenarios are already covered by the regular Python/boto3 tests.

bblommers (approved):

PR #8533, opened 2025-01-28
Lex Support

jamarcelin:


jamarcelin:
@bblommers Hello! I just started on this Lex Module and I am getting this error in `base_backend.py` when I run the test `test_create_bot` (It's not supposed to pass). I am getting this error:

`E       ModuleNotFoundError: No module named 'moto.lexv2urls'`

Any idea why its looking for a lexv2urls?

bblommers:
Hi @jamarcelin! That was essentially a bug in our core logic. In order to find find the `urls.py`-file for any given service, we take the name of the models-file (`moto.service.models`), replaced `models` with `urls`, and assumed that the result (`moto.service.urls`) was always correct.
That obviously doesn't work when the service name contains the word `models`, as it then tries to find `moto.lexv2urls.urls`, which doesn't exist. I've pushed a fix to the logic itself, so we can keep using the same service name.

(I've just realized that I haven't tested whether this breaks any other services... I expect it to be OK, but I'll keep an eye out. Please tag me again if you see anything else breaking because of this change. :slightly_smiling_face: )

jamarcelin:
@bblommers Sounds good! This PR should be complete now!

jamarcelin:
@bblommers Hey are these test errors an issue?

moto/resourcegroupstaggingapi/models.py
```diff
@@ -163,6 +164,10 @@ def kafka_backend(self) -> KafkaBackend:
     def sagemaker_backend(self) -> SageMakerModelBackend:
         return sagemaker_backends[self.account_id][self.region_name]
 
+    @property
+    def lexv2_backend(self) -> LexModelsV2Backend:
+        return lexv2models_backends[self.account_id][self.region_name]
```
bblommers, line 169: This service is not available in every region @jamarcelin - that's why we're seeing the errors in the CI.

You can just guard this with a simple if-statement, similar to what the `workspaces_backends` above does.

moto/resourcegroupstaggingapi/models.py
```diff
@@ -455,6 +460,26 @@ def format_tag_keys(
 
                 yield {"ResourceARN": f"{kms_key.arn}", "Tags": tags}
 
+        # LexV2
+        lex_v2_resource_map: Dict[str, Dict[str, Any]] = {
+            "lexv2:bot": self.lexv2_backend.bots,
```
bblommers, line 465: Note that `self.lexv2_backend` may become None here, if it does not exist in this region, so we should guard against that here as well

moto/lexv2models/models.py
```diff
@@ -0,0 +1,397 @@
...
+                "description": ba.description,
+                "botVersion": ba.bot_version,
+                "botAliasStatus": ba.status,
+                "creationDateTime": ba.creation_date_time,
+                "lastUpdatedDateTime": ba.last_updated_date_time,
+            }
+            for ba in self.bot_aliases.values()
+        ]
+
+        return bot_alias_summaries, next_token, bot_id
```
bblommers, line 350: We should not return the `next_token` that the user provides. Usually if a user wants to paginate this method, it will be a simple loop like:
```
while not next_token:
    # keep calling the same method
```
If we always return a value here, the loop will never end.

We should just return None here (or completely remove it from the code), to make it clear that pagination is not yet supported

moto/lexv2models/models.py
```diff
@@ -0,0 +1,397 @@
...
+                "botName": bot.bot_name,
+                "description": bot.description,
+                "botStatus": bot.status,
+                "latestBotVersion": 1,
+                "lastUpdatedDateTime": bot.last_updated_date_time,
+                "botType": bot.bot_type,
+            }
+            for bot in self.bots.values()
+        ]
+        return bot_summaries, next_token
```
bblommers, line 229: Same here with the `next_token`

bblommers (changes requested):

bblommers:
The tests failed because the URL's were incomplete.

When running Moto in ServerMode, we don't have access to the hostname (because everything is `localhost`), so instead we use the signing-name specified in the Authorization-header. This usually corresponds exactly to the service-name. Problem is that botocore uses `lex` as the signing name, and Moto takes that and tries to find a service with `lex.amazonaws.com` as the URL.

https://github.com/boto/botocore/blob/c132b4d58cd76e4f786854e7bceaf8a6b67057ff/botocore/data/lex-models/2017-04-19/service-2.json#L11

So as a workaround, I just added `lex.amazonaws.com` as part one of the 'acceptable' URLs. The other Lex services (lex-models, lex-runtime) use the same signing-name, so when we build support for those, we will need to find a different way to differentiate the three services. For now, this will do though.

bblommers (approved):
Thanks for the PR @jamarcelin!

PR #8535, opened 2025-01-29
S3: Return ETag in 304 responses from get_object

apinkney97:
Fixes https://github.com/getmoto/moto/issues/8534


bblommers (approved):
LGTM - thank you for the fix @apinkney97!

apinkney97:
Thanks for merging @bblommers!

PR #8539, opened 2025-01-31
[Elasticache] CreateUser:Engine parameter is case-insensitive

### LINKED ISSUE #8520: [ElastiCache] createUser should map engine to be lowercase
In boto3, I created two Elasticache Redis users and noticed that the engine value for users get stored in lowercase. May need to adapt moto to store the passed values in lowercase.

Create with lowercase
```
import boto3
client = boto3.client('elasticache')
client.create_user(UserId="user1", UserName="user1", Engine = "redis", AccessString = " on ~* +@all", AuthenticationMode = { 'Type': 'iam' },)

```

result:
```
{'UserId': 'user1', 'UserName': 'user1', 'Status': 'active', 'Engine': 'redis', ...}
```

Create with upper case
```
client.create_user(UserId="user2", UserName="user2", Engine = "Redis", AccessString = " on ~* +@all", AuthenticationMode = { 'Type': 'iam' },)
```

result:
```
{'UserId': 'user2', 'UserName': 'user2', 'Status': 'active', 'Engine': 'redis', ...}
```

Looking at the [Elasticache User API docs](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-elasticache-user.html#cfn-elasticache-user-engine), the values accepted are `'valkey'` or `'redis'`

armichaud:
Resolves #8520


armichaud:
@bblommers Small contribution, ready for review

bpandola (changes requested):
I think the heart of the issue is that regardless if the engine type parameter is `Redis` or `redis`, we want to store the value as lowercase.  This changeset, as currently written, would raise an error if `Redis` was passed in, which doesn't match AWS behavior.  Just converting the `engine` parameter to lowercase in `responses.py:create_user` would be the simplest way to address this.

armichaud:
> I think the heart of the issue is that regardless if the engine type parameter is `Redis` or `redis`, we want to store the value as lowercase. This changeset, as currently written, would raise an error if `Redis` was passed in, which doesn't match AWS behavior. Just converting the `engine` parameter to lowercase in `responses.py:create_user` would be the simplest way to address this.

Good feedback @bpandola , thank you for the review. To clarify, does AWS handle this parameter without case sensitivity – i.e. are cases like `Valkey` or `rEDiS` also valid? – or is `Redis` just a special case?  I'm going to assume the later for the time being, but we can easily refactor.

armichaud:
@bpandola Confusingly, the [API docs](https://docs.aws.amazon.com/AmazonElastiCache/latest/APIReference/API_CreateUser.html) state that the currently supported values are `redis` and `valkey`, but the [boto3 docs](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/elasticache/client/create_user.html) state that the only supported value is `Redis`. If I have time later, I'll test both the API and boto3 to observe their behavior, but for now this makes me further inclined to believe that the accepted options are `Redis`,`redis`, and `valkey`.

bpandola (approved):
Thanks @armichaud. I pushed a few commits to get this over the finish line.  And to answer your question: Yes, the engine parameter is completely case insensitive.  `redis`, `Redis`, and `reDis` are all acceptable and I added a unit test to cover this.

PR #8540, opened 2025-01-31
RDS: add OriginalSnapshotCreateTime to DB snapshots

snordhausen:
In contrast to `SnapshotCreateTime`, the `OriginalSnapshotCreateTime` does not change when the snapshot gets copied.

Testing uses only 3 decimals for timestamp because the `iso_8601_datetime_with_milliseconds()` only provides 3. Skip some tests when in server mode, because `freeze_time` does not work there.


moto/rds/models.py
```diff
@@ -1912,6 +1921,7 @@ def copy_db_snapshot(
             db_instance=source_snapshot.database,
             db_snapshot_identifier=target_snapshot_identifier,
             tags=tags,
+            original_created_at=source_snapshot.original_created_at,
```
bpandola, line 1924: Shouldn't this be `orginal_created_at=source_snapshot.created_at`?

tests/test_rds/test_rds.py
```diff
@@ -4,8 +4,10 @@
 import boto3
 import pytest
 from botocore.exceptions import ClientError
+from dateutil.tz import tzutc
+from freezegun import freeze_time
```
bpandola, line 8: Freezing time is a pain (doesn't work in server mode, etc.) and we should really only use it when absolutely necessary.  Isn't it enough to assert that `OrginalSnapshotCreateTIme == SnapshotCreateTime` or `SnapshotCreateTime > OriginalSnapshotCreateTime`?

bpandola (changes requested):

moto/rds/models.py
snordhausen: The `original_created_at` is correct: This field shows when the content was originally created. The `created_at` reflects the time when a particular copy was made.

I just tested this on AWS, taking a snapshot `A`, copying it to `B`, and then copying `B` to `C`. The `OriginalSnapshotCreateTime` is identical for all three of them, and it is equeal to `A`'s `SnapshotCreateTime` value.

tests/test_rds/test_rds.py
snordhausen: > `SnapshotCreateTime > OriginalSnapshotCreateTime`

The internal timekeeping of the DBSnapshot class uses only millisecond precision (just like AWS, so we shouldn't change that). But if the test code runs fast enough (<1ms), the assertion would fail or the test would become flaky. Right now the create+copy takes ~5ms on my system, but we never know what kind of optimizations people will come up with.

So I'd like to keep the freezing to ensure the test stays reliable.

bpandola (approved):
@snordhausen Thanks for verifying this behavior against the real AWS backend, and for dealing with the (ongoing) major refactor of the moto RDS backend.

PR #8541, opened 2025-01-31
Make code ruffer

snordhausen:
I accidentally formatted the code with ruff version 0.9.4 and it produced some code simplifications that ruff 0.3.3 does not object to.


bpandola (approved):
Thanks @snordhausen

PR #8555, opened 2025-02-03
[RDS] Replace XML Jinja Templates with proper XML Response Serializer

bpandola:
This PR adds a response serializer for the AWS Query protocol to the RDS backend, eliminating the need for manually editing XML Jinja templates.  The serializer uses the model information already present in the `botocore` data files to output formatted, type-aware XML responses based on the `moto` RDS model classes.

This changeset attempts to be as clean as possible, with most file changes containing only additions or deletions.  The exception to this is `responses.py`, where each controller method was migrated from a template-based string response to a serialized response, resulting in a large but still understandable diff.

Eliminating 1500+ lines of Jinja template strings and `to_xml()` model methods is a big maintenance win.  Future updates to RDS need only add new attributes to the model--one place instead of the previous multiple (model attribute, model xml template, response template).

> [!NOTE]\
> Future changesets planned:
> * The serialization code, once battle-tested on the RDS backend, will be moved into the `core` directory for import by other endpoints using the AWS Query protocol.
> * Most of the changes in `utils.py` and `viewmodels.py` are just compensating for current discrepancies between the `moto` RDS models and the `botocore` RDS models.  This code was separated out for the time being to make the `models.py` diff pure (all deletions); but will be integrated into the `moto` RDS models over time.

### DIFF
--- a/moto/rds/exceptions.py
+++ b/moto/rds/exceptions.py
@@ -1,23 +1,7 @@
-from jinja2 import Template
-
-from moto.core.exceptions import RESTError
-
-
-class RDSClientError(RESTError):
+class RDSClientError(Exception):
     def __init__(self, code: str, message: str):
-        super().__init__(error_type=code, message=message)
-        template = Template(
-            """
-        <ErrorResponse>
-            <Error>
-              <Code>{{ code }}</Code>
-              <Message>{{ message }}</Message>
-              <Type>Sender</Type>
-            </Error>
-            <RequestId>6876f774-7273-11e4-85dc-39e55ca848d1</RequestId>
-        </ErrorResponse>"""
-        )
-        self.description = template.render(code=code, message=message)
+        super().__init__(message)
+        self.code = code
 
 
 class DBInstanceNotFoundError(RDSClientError):
@@ -48,8 +32,6 @@ class DBSecurityGroupNotFoundError(RDSClientError):
 
 
 class DBSubnetGroupNotFoundError(RDSClientError):
-    code = 404
-
     def __init__(self, subnet_group_name: str):
         super().__init__(
             "DBSubnetGroupNotFoundFault", f"Subnet Group {subnet_group_name} not found."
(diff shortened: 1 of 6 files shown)


bpandola:
@bblommers Let me know if you have any questions or concerns about this one.  It took me way longer to get to this stage than I had hoped, mostly because I've been trying to avoid a huge, indecipherable diff.  But... I'm not sure it's worth it.  We have good test coverage and I'm committed to seeing this through, so if you're ok with this changeset and where this is headed, I may forgo the clean diffs that take forever to implement in favor of messy but quick progress.  Looking forward to hearing your thoughts.

bblommers:
> @bblommers Let me know if you have any questions or concerns about this one. It took me way longer to get to this stage than I had hoped, mostly because I've been trying to avoid a huge, indecipherable diff. But... I'm not sure it's worth it. We have good test coverage and I'm committed to seeing this through, so if you're ok with this changeset and where this is headed, I may forgo the clean diffs that take forever to implement in favor of messy but quick progress. Looking forward to hearing your thoughts.

This looks great @bpandola! Very satisfying to see all the XML gone. :) 

In general, I'm also in favor of messy but quick progress - as you said, our test suite is large enough, so that gives me enough confidence that this will not break anything. The difference between the generic serialization and RDS-specific fixes is also clear, so it's quite easy to see how we can apply this to other services in the future.

The only nitpick I have is really about the location of `serializer.py`: Does it make sense to already move this to `moto/core/..`? It looks like it is completely generic, so moving it away from the `rds` module would make sure we don't accidentally break the separation between generic and RDS-specific serialization logic.

bpandola:
@bblommers wrote:
> The only nitpick I have is really about the location of `serializer.py`: Does it make sense to already move this to `moto/core/..`? It looks like it is completely generic, so moving it away from the `rds` module would make sure we don't accidentally break the separation between generic and RDS-specific serialization logic.

I don't want to move this just yet.  Yes, it is designed to be completely generic, but I may tweak/optimize it a bit as I continue with the rest of my planned RDS changes.  I do have some separate work going to add serializers for all of the AWS protocols (with a comprehensive test suite) but my priority at the moment is RDS so I can get off my fork and get back to upstream `moto`.  Rest assured, I will make this serializer available to other services that can use it, but I want to keep it out of `core` until I feel it is completely baked (and I'm ready to support it more broadly).

PR #8556, opened 2025-02-03
EC2: Require size/snapshot when creating EBS volume

nik-localstack:
Related to https://github.com/localstack/localstack/issues/11522


bblommers (approved):
LGTM - thank you @nik-localstack!

PR #8561, opened 2025-02-04
RDS Backend Improvements 

bpandola:
The RDS models have been updated with all of the attributes from (the now deleted) `viewmodels.py` data transfer classes, encapsulating all entity information in one place.  The models now also inherit a Mixin class that allows for seamless translation of the `botocore` model attributes (e.g. `DBInstanceIdentifier`) to the corresponding `moto` model attributes (e.g. `db_instance_identifier`) during serialization.

Additionally, a dedicated request parser has been added that deserializes the request parameters based on the structure/type information contained in the `botocore` models.  This eliminates the need for manually pulling individual parameters out of the request using a variety of helper methods (`_get_param()`, `_get_bool_param()`, etc.).

> [!NOTE]\
> The diff on this is pretty much indecipherable--lot of changes here--but, importantly, no changes were made to the test suite.  All tests are passing without modification.


moto/rds/models.py
```diff
@@ -573,93 +743,115 @@ class DBInstance(CloudFormationModel, RDSBaseModel):
     resource_type = "db"
 
     def __init__(
-        self, backend: "RDSBackend", db_instance_identifier: str, **kwargs: Any
```
bblommers, line 576: I don't know whether this was intentional, but I do love the fact that we're getting rid of `kwargs` here. Typing them all out is a bit verbose, but I prefer the readability of it.

bpandola: Definitely intentional.  I plan to do it for all the RDS models.

PR #8563, opened 2025-02-04
ELB: Port and protocol validations for create listener

nik-localstack:
When creating a listener, port and protocol should satisfy some conditions base on the load balancer type.

https://docs.aws.amazon.com/cli/latest/reference/elbv2/create-listener.html#options


bblommers (approved):
Great tests @nik-localstack - always good to check every single scenario. 

Thank you!

PR #8564, opened 2025-02-04
RDS: lowercase the DBInstanceIdentifier

snordhausen:
Various RDS API calls treat the DBInstanceIdentifier as lowercase / case insensitive.

[Documentation](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/rds/client/create_db_instance.html) for `create_db_instance` says:
`DBInstanceIdentifier [...] This parameter is stored as a lowercase string.` 

[Documentation](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/rds/client/describe_db_instances.html) for `describe_db_instances` says:
`DBInstanceIdentifier [...] This parameter isn’t case-sensitive.` 

[Documentation](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/rds/client/modify_db_instance.html) for `modify_db_instance` says:
`DBInstanceIdentifier [...] This value is stored as a lowercase string.`
`NewDBInstanceIdentifier [...] This value is stored as a lowercase string.` 

[Documentation](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/rds/client/delete_db_instance.html) for `delete_db_instance` says:
`DBInstanceIdentifier [...] This parameter isn’t case-sensitive.`


snordhausen:
I'll take a look at the test failures a bit later.

bpandola (changes requested):
@snordhausen There is a larger issue here, in that all of the RDS models (DBInstance, DBCluster, DBSnapshot, etc.) have case-insensitive identifiers, so I think we need a higher-level solution.  Peppering `.lower()` throughout the code is messy, and it's too easy to miss a spot.  We definitely want to handle this at the boundaries and not indiscriminately throughout the code.  For example, we could store the identifiers in a case-insensitive dict, we can add a parameter to the filter configuration(s) to indicate if a comparison should be case-insensitive, etc. 

If you'd like to consider this a bit more and propose another solution, feel free.  If the increased scope is not something you want to tackle, I can add it to my TODO list (as I am currently in the process of refactoring the RDS backend and adding lots of new features).

snordhausen:
@bpandola What do you think about doing it [like this](https://github.com/snordhausen/moto/commit/83ba09456c2f00b0e5673b8e5b078b19db5bba21)? That lower-cases the DBInstanceIdentifier before it gets in contact with any backend logic and [replaces the old code](https://github.com/snordhausen/moto/commit/c4c7ff9eb80f42e93c21b48715738e0b8da666c9).

I think it can be extended for the other cases you mentioned, e.g. with a for-loop in `QueryParser.parse()` that lower-cases various keys.

bpandola:
> @bpandola What do you think about doing it [like this](https://github.com/snordhausen/moto/commit/83ba09456c2f00b0e5673b8e5b078b19db5bba21)? That lower-cases the DBInstanceIdentifier before it gets in contact with any backend logic and [replaces the old code](https://github.com/snordhausen/moto/commit/c4c7ff9eb80f42e93c21b48715738e0b8da666c9).
> 
> I think it can be extended for the other cases you mentioned, e.g. with a for-loop in `QueryParser.parse()` that lower-cases various keys.

Although the parser is an obvious boundary, it's intended to be totally generic and not contain any service-specific code.  I wound up implementing this in a different way, but I appreciate you bringing it to our attention and providing a thorough test case.

bpandola:
Superseded by #8777

PR #8567, opened 2025-02-05
Admin: Compatibility with MyPy 1.15

bblommers:
MyPy 1.15 got a lot stricter in the rule: `Type of decorated function contains type "Any"`

 - Changed `list_..` methods to return an Object where possible, instead of Dict
 - Change `Dict[str, Any]` to `Dict[str, specific]` where possible
 - Type: ignored everything else
 - Fixed pagination for AppMesh methods

### DIFF
--- a/moto/appmesh/models.py
+++ b/moto/appmesh/models.py
@@ -1,6 +1,6 @@
 """AppMeshBackend class with methods for supported APIs."""
 
-from typing import Any, Dict, List, Literal, Optional, Union
+from typing import Dict, List, Literal, Optional, Union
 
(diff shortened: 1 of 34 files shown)


bpandola:
@bblommers Do you have any automated tools/processes for adding typing, or are you doing this all manually?

bblommers:
All manual @bpandola! Sometimes it's a bit of a slog, but it can also be useful IMO - a type error is 9 out of 10 times an indication that you're doing it wrong, so that's always good opportunity to revisit/refactor things.

PR #8574, opened 2025-02-07
Allow non kms origin in create_key

waddamski:
This is a change to allow the 'Origin' parameter in create_key to affect the origin of the mocked key.


bblommers (approved):
LGTM - thank you for adding this to Moto @waddamski!

PR #8588, opened 2025-02-12
ECS: tasks include containers objects

rafcio19:
Resolves https://github.com/getmoto/moto/issues/8319

### DIFF
--- a/moto/ecs/models.py
+++ b/moto/ecs/models.py
@@ -386,7 +386,6 @@ class Task(BaseObject, ManagedState):
         self.desired_status = "RUNNING"
         self.task_definition_arn = task_definition.arn
         self.overrides = overrides or {}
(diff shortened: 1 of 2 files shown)


moto/ecs/models.py
```diff
@@ -451,6 +451,7 @@ def response_object(self, include_tags: bool = True) -> Dict[str, Any]:  # type:
             response_object.pop("tags", None)
         response_object["taskArn"] = self.task_arn
         response_object["lastStatus"] = self.last_status
+        response_object["containers"] = [self.containers[0].response_object]
```
bblommers, line 454: Nice that we can just use the `response_object` OOTB!

bblommers (approved):
LGTM - thank you for tackling this @rafcio19!

PR #8603, opened 2025-02-16
Lambda: Pick Lambda image deterministically

### LINKED ISSUE #8602: Non-deterministic selection of Lambda image
At [aiobotocore](https://github.com/aio-libs/aiobotocore), we use `moto` as part of our test suite. We recently started to run CI on arm64 runners and noticed frequent failures due to docker related issues, triggered by moto-backed tests performing Lambda invokations. It turned out that the outdated `lambci/lambda` images are not available for arm64 and thus causing the failures.

Unfortunately, it appears to be impossible to force moto to reliably pick `mlupin/docker-lambda` due to the way moto determines the Lambda image to use:

https://github.com/getmoto/moto/blob/fc7bcea62bab384df68d996e24a52e90637303e9/moto/awslambda/models.py#L1025-L1033

The order in which images are picked in non-deterministic, due to the Python set being used. I would suggest to make this deterministic and will provide a PR shortly.

jakob-keller:
closes #8602

### DIFF
--- a/moto/awslambda/models.py
+++ b/moto/awslambda/models.py
@@ -1023,13 +1023,11 @@ class LambdaFunction(CloudFormationModel, DockerModel):
                     # - lambci/lambda (the repo with older/outdated AWSLambda images
                     #
                     # We'll cycle through all of them - when we find the repo that contains our image, we use it
-                    image_repos = set(
-                        [
-                            settings.moto_lambda_image(),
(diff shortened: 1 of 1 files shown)


bblommers (approved):
Makes sense @jakob-keller. It's a shame that Python does not have an OrderedSet, but this is a great alternative.

Thank you for the PR!

PR #8607, opened 2025-02-18
adding get_administrator_account method for guardduty

ghorondo:


moto/guardduty/models.py
```diff
@@ -86,6 +89,26 @@ def get_detector(self, detector_id: str) -> "Detector":
             raise DetectorNotFoundException
         return self.detectors[detector_id]
 
+    # def get_administrator_account(self, detector_id: str) -> "Detector":
```
bblommers, line 92: This commented-out part doesn't seem necessary

tests/test_guardduty/test_guardduty.py
```diff
@@ -175,3 +175,32 @@ def test_delete_detector():
...
+    assert list_resp["AdminAccounts"][0]["AdminStatus"] == "ENABLED"
+
+    # Get administrator account details
+    resp = guardduty_client.get_administrator_account(DetectorId=detector_id)
+
+    # Assertions for get_administrator_account response
+    assert resp["ResponseMetadata"]["HTTPStatusCode"] == 200
+
+    # Remove the previous assertion about Administrator
+    # As the response you showed doesn't contain this structure
```
bblommers, line 206: This looks like a TODO - and it would be nice to actually see the output (and verify it!)

I _think_ the problem is that you're returning the response fields in upper-case:
```
{
    "Administrator": {
      "AccountId", 
      ..
  }
}
```
But boto3 expects these fields to start with a lowercase letter:
```
{
  "administrator: {
    "accountId": 
    ...
  }
}
```

Let me know if that helps.

bblommers (changes requested):

bblommers (approved):
LGTM - thank you for adding this feature to Moto @ghorondo!

PR #8612, opened 2025-02-21
DynamoDB: Add attribute name use checks to validation layer

BlizzardOf:
This addresses issue https://github.com/getmoto/moto/issues/8474

In the responses file, I added validation checks that every attribute name is used in the various expressions that each request takes. Without major changes, these validations have to be done at the responses level instead of the backend level, because some response functions call the same backend functions, and which expressions need to be included in the check varies between request types.

I had to adapt a couple of the parsers to track which attribute names they used during parsing.

I skipped batch_write_items because [boto doesn't support any expressions in it](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/dynamodb/client/batch_write_item.html), for some reason. I also skipped transact_get_items because it looks like *moto* doesn't support expressions for that call, even though moto does.

I used the aws.verified tag to get the tests working, but I actually haven't verified it against AWS. I was wondering if you had some tips for setting up a proper environment? I have no idea where to get started even.

### DIFF
--- a/moto/dynamodb/responses.py
+++ b/moto/dynamodb/responses.py
@@ -1252,11 +1252,11 @@ class DynamoHandler(BaseResponse):
         # Validate first - we should error before we start the transaction
         for item in transact_items:
             # This logic is common among all types of write items
-            item_values = list(item.values())[0] # Each item only has one of Put, Update, Delete, ConditionCheck
+            item_values = list(item.values())[
+                0
+            ]  # Each item only has one of Put, Update, Delete, ConditionCheck
             condition_expression = item_values.get("ConditionExpression")
-            expression_attribute_names = item_values.get(
-                "ExpressionAttributeNames", {}
-            )
+            expression_attribute_names = item_values.get("ExpressionAttributeNames", {})
             expression_attribute_values = item_values.get(
                 "ExpressionAttributeValues", {}
             )
@@ -1302,7 +1302,9 @@ class DynamoHandler(BaseResponse):
                     attr.get_attribute_name_placeholder() for attr in attr_name_clauses
                 ]
 
-            validate_attribute_names_used(expression_attribute_names, expression_attribute_names_used)
+            validate_attribute_names_used(
+                expression_attribute_names, expression_attribute_names_used
+            )
 
         self.dynamodb_backend.transact_write_items(transact_items)
         response: Dict[str, Any] = {"ConsumedCapacity": [], "ItemCollectionMetrics": {}}
--- a/moto/dynamodb/responses.py
+++ b/moto/dynamodb/responses.py
@@ -1251,6 +1251,24 @@ class DynamoHandler(BaseResponse):
         transact_items = self.body["TransactItems"]
         # Validate first - we should error before we start the transaction
         for item in transact_items:
+            # This logic is common among all types of write items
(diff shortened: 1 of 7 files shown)


moto/dynamodb/responses.py
```diff
@@ -1142,6 +1258,25 @@ def transact_write_items(self) -> str:
                 item_attrs = item["Put"]["Item"]
                 table = self.dynamodb_backend.get_table(item["Put"]["TableName"])
                 validate_put_has_empty_keys(item_attrs, table)
+
+                condition_expression = item["Put"].get("ConditionExpression")
```
bblommers, line 1262: This part seems to be the same for every action. Can we extract that somehow, to keep it DRY?

I don't want to prescribe exactly how you should do it, but I was thinking something along the lines of this:
```
for action in ["Put", ..]:
    if details := item.get(action):
        condition_expression = details.get("ConditionExpression")
        ...
```

bblommers:
> I used the aws.verified tag to get the tests working, but I actually haven't verified it against AWS. I was wondering if you had some tips for setting up a proper environment? I have no idea where to get started even.

Hi @BlizzardOf! You can use the environment variable `MOTO_TEST_ALLOW_AWS_REQUEST=true` to run tests against AWS.

All tests marked with `aws_verified` are setup in such a way that all resources are automatically deleted afterwards, even if the test fails - so you don't have to worry about dangling resources.

To run all tests in a single file/folder against AWS:
```
MOTO_TEST_ALLOW_AWS_REQUEST=true pytest -sv tests/test_dynamodb/exceptions/test_dynamodb_exceptions.py -m aws_verified -n 4
```
Note the `-n 4` - not required, but it will run all tests in parallel to speed things up.


Or if you want to validate a single test:
```
MOTO_TEST_ALLOW_AWS_REQUEST=true pytest -sv tests/test_dynamodb/exceptions/test_dynamodb_exceptions.py::test_query_unused_attribute_name -vv
```

Note that this particular test (`test_query_unused_attribute_name`) actually fails against AWS, so that needs to be fixed first:
```
>       assert (
            err["Message"]
            == "Value provided in ExpressionAttributeNames unused in expressions: keys: {#count}"
        )
E       assert 'Invalid KeyConditionExpression: Syntax error; token: "1", near: "= 1)"' == 'Value provided in ExpressionAttributeNames unused in expressions: keys: {#count}'
E         - Value provided in ExpressionAttributeNames unused in expressions: keys: {#count}
E         + Invalid KeyConditionExpression: Syntax error; token: "1", near: "= 1)"
```

BlizzardOf:
Hm, it's bizarre, I can test against AWS when I run each test individually, but when I run a bunch of them at once, they all fail with 
` An error occurred (UnrecognizedClientException) when calling the CreateTable operation: The security token included in the request is invalid.`
Seems to happen whether I run sequentially or not. You ever see something like that?

Whatever the issue is I can replicate the issue you mention, I should have updates up in a day or so.

BlizzardOf:
Alright, I think it's ready

BlizzardOf:
@bblommers no rush, but this is ready for review

bblommers (approved):
LGTM - thank you for expanding this validation in Moto @BlizzardOf!

Issue #8592, opened 2025-02-13
Submitting a S3 CopyObject to replace user metadata with Content-Length 0 causes Error 500 after the checksum workaround


jeking3:
Example request:
```
[DEBUG] 2025-02-13 20:48:40.015 CURL [140340847924992] (HeaderOut) PUT /s3testbucket/dst/testfile HTTP/1.1
Host: 127.0.0.1:12509
Accept: */*
amz-sdk-invocation-id: 008050B0-3919-4A52-AA41-5E93710D4049
amz-sdk-request: ttl=20250213T205140Z; attempt=4
authorization: ...
content-encoding: aws-chunked
content-length: 0
content-type: binary/octet-stream
user-agent: aws-sdk-cpp/1.11.500 md/internal ua/2.1 api/S3 os/Linux#5.10.233-204.894.x86_64 lang/c++#C++11 md/aws-crt#0.30.1 md/arch#x86_64 md/GCC#7.3.1 m/Z,b
x-amz-api-version: 2006-03-01
x-amz-content-sha256: e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855
x-amz-copy-source: s3testbucket/dst/testfile
x-amz-copy-source-if-match: "620f0b67a91f7f74151bc5be745b7110"
x-amz-date: 20250213T204840Z
x-amz-meta-last-modified: 1739479675000
x-amz-meta-usermetakey: usermetaval
x-amz-metadata-directive: REPLACE
x-amz-security-token: ...
```

Example 500 stack:
```
Error on request:
Traceback (most recent call last):
  File "/test-runtime/lib/python3.9/site-packages/werkzeug/serving.py", line 370, in run_wsgi
    execute(self.server.app)
  File "/test-runtime/lib/python3.9/site-packages/werkzeug/serving.py", line 331, in execute
    application_iter = app(environ, start_response)
  File "/test-runtime/lib/python3.9/site-packages/moto/moto_server/werkzeug_app.py", line 264, in __call__
    return backend_app(environ, start_response)
  File "/test-runtime/lib/python3.9/site-packages/flask/app.py", line 1478, in __call__
    return self.wsgi_app(environ, start_response)
  File "/test-runtime/lib/python3.9/site-packages/flask/app.py", line 1458, in wsgi_app
    response = self.handle_exception(e)
  File "/test-runtime/lib/python3.9/site-packages/flask_cors/extension.py", line 176, in wrapped_function
[...]

PR #8613, opened 2025-02-22
Admin: Add CPP tests

bblommers:
Some very simple tests using C++, just to verify that the MotoServer works against this SDK. I've included a cache for the actual build output as well. Downloading all build tools + building the project takes 5 minutes, so including a cache should avoid some unnecessary compute time.

The need to have this was raised in #8592, a bug report that only occurred when using C++ SDK.

This PR only includes smoke tests - it does not create the exact conditions described in #8592, and therefore does not test this scenario.

Documentation used:
 - https://docs.aws.amazon.com/sdk-for-cpp/v1/developer-guide/sdk-from-pm.html
 - https://docs.aws.amazon.com/sdk-for-cpp/v1/developer-guide/cpp_s3_code_examples.html

Outstanding tasks:
 - Rework this to use proper unit tests (for example by using [GTest](https://google.github.io/googletest/))
 - Include a test to verify the CopyObject scenario works, so that we can prevent it in the future

(FWIW, this is the first time I've touched C++, ever, so excuse the code quality. I'm just happy it works for now :shrug: )

### DIFF
--- a/.github/workflows/tests_sdk_cpp.yml
+++ b/.github/workflows/tests_sdk_cpp.yml
@@ -1,0 +1,39 @@
+name: C++ SDK test
+on: [workflow_call]
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v4
+      with:
+        fetch-depth: 0
+    - name: Set up Python 3.8
+      uses: actions/setup-python@v5
+      with:
+        python-version: "3.8"
+    - name: Start MotoServer
+      run: |
+        pip install build
+        python -m build
+        docker run --rm -t --name motoserver -e TEST_SERVER_MODE=true -e AWS_SECRET_ACCESS_KEY=server_secret -e AWS_ACCESS_KEY_ID=server_key -v `pwd`:/moto -p 5000:5000 -v /var/run/docker.sock:/var/run/docker.sock python:3.10-slim /moto/scripts/ci_moto_server.sh &
+        python scripts/ci_wait_for_server.py
+
+    - name: Check build cache
+      id: build-cache
+      uses: actions/cache@v4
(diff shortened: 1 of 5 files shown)


.github/workflows/tests_sdk_cpp.yml
```diff
@@ -0,0 +1,39 @@
...
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v4
+      with:
+        fetch-depth: 0
+    - name: Set up Python 3.8
+      uses: actions/setup-python@v5
+      with:
+        python-version: "3.8"
```
jeking3, line 15: 3.8 is EOL, 3.10 would allow more runway

other_langs/tests_cpp/hello_s3.cpp
```diff
@@ -0,0 +1,142 @@
...
+        createBucket(bucket1, clientConfig);
+        createBucket(bucket2, clientConfig);
+
+        putObject(bucket1, "test.txt", clientConfig);
+
+        copyObject("test.txt", bucket1, bucket2, clientConfig);
+    }
+
+    Aws::ShutdownAPI(options); // Should only be called once.
+    return result;
```
jeking3, line 141: This returns 0 (success) even if errors occur.

.github/workflows/tests_sdk_cpp.yml
bblommers: Good catch - changed

other_langs/tests_cpp/hello_s3.cpp
bblommers: Great point - I validated manually that the test passed, but that's obviously not enough.

Appreciate the review!

PR #8617, opened 2025-02-23
Admin: Drop support for Python 3.8

bblommers:
Now that we no longer support for 3.8 we can also simplify a bunch of types - i.e. `List[str]` can become `list[str]`, same for Set->set, etc.

That is a bigger change though, so I'll tackle that later.


bpandola:
@bblommers Nice.  Good riddance!  😄

PR #8628, opened 2025-02-26
[EventBridge] Add event-bus support to resourcetaggingapi

zkarpinski:
Adds event-bus support to resourcetaggingapi


bblommers (approved):
Thanks @zkarpinski!

PR #8629, opened 2025-02-26
Only import `typing_extensions` when `TYPE_CHECKING`

### LINKED ISSUE #8627: typing_extensions introduced as prod dependency where it should not have been
https://github.com/getmoto/moto/issues/8289 has regressed in version 5.1.0 while using python versions 3.11 to 3.13 (3.10 and 3.9 do work without requiring typing_extensions).

ModuleNotFoundError: No module named 'typing_extensions'

Sample requirements-dev.txt are below python 3.13:

Working

```
moto==5.0.28
```

Working:

```
moto==5.1.0
typing_extensions==4.12.2
```

Not working:

```
moto==5.1.0
```

bpandola:
Closes #8627

### DIFF
--- a/moto/rds/models.py
+++ b/moto/rds/models.py
@@ -69,8 +69,9 @@ from .utils import (
 )
 
 if TYPE_CHECKING:
-    from moto.ec2.models.subnets import Subnet
     from typing_extensions import Literal, Protocol, Type
+    
+    from moto.ec2.models.subnets import Subnet
 
 
 def find_cluster(cluster_arn: str) -> DBCluster:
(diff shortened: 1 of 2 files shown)


bpandola:
@bblommers I've fallen victim to the pitfalls of Python typing yet again... 😳  I wonder if there's a way we can check for this issue automatically during the CI run.  Maybe install `requirements-test.txt` instead of `requirements-dev.txt`?

bblommers:
Agreed @bpandola - open to suggestions on the best way to do this.

I've opened #8630 at the same time - GMTA! I'll leave it to you to decide which works best

bpandola:
Superseded by #8630

PR #8630, opened 2025-02-26
Admin: Remove typing_extensions module

### LINKED ISSUE #8627: typing_extensions introduced as prod dependency where it should not have been

bblommers:
Fixes #8627 

FYI @bpandola - a little refactoring to get rid of this dependency. I replaced `TypeAlias` completely because it is a) only introduced in 3.10, and b) [was deprecated again in 3.12](https://docs.python.org/3/library/typing.html#typing.TypeAlias). It seems to run fine without it, so I think it's OK? Happy to be corrected though, typing is not my strong suit. :)

### DIFF
--- a/moto/rds/models.py
+++ b/moto/rds/models.py
@@ -9,9 +9,18 @@ import string
 from collections import OrderedDict, defaultdict
 from functools import lru_cache, partialmethod
 from re import compile as re_compile
-from typing import TYPE_CHECKING, Any, Dict, Iterable, List, Optional, Tuple, Union
-
-from typing_extensions import Literal, Protocol, Type
+from typing import (
+    TYPE_CHECKING,
+    Any,
+    Dict,
+    Iterable,
+    List,
+    Literal,
+    Optional,
+    Protocol,
+    Tuple,
+    Union,
+)
 
 from moto.core.base_backend import BackendDict, BaseBackend
 from moto.core.common_models import BaseModel, CloudFormationModel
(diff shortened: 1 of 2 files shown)


bpandola (approved):

bblommers:
> Still wondering if maybe we should only install requirements-tests.txt for CI to try to catch this automatically going forward.

True, that does make sense. Only install `-dev` when linting, installing `-tests` only when running tests.

bpandola:
@bblommers Thanks for doing this.  I felt bad that I had caused a regression, so I was trying (and failing) earlier to address this via the GitHub editor on my phone.  😆

bpandola:
> > Still wondering if maybe we should only install requirements-tests.txt for CI to try to catch this automatically going forward.
> 
> True, that does make sense. Only install `-dev` when linting, installing `-tests` only when running tests.

@bblommers Just to follow up on this: the `cf-lint` package lists `typing_extensions` in its requirements, so it's always installed during our CI runs (`pip install -e .[all,server]`).  I'm sure there's still some way to automate a check that all moto's uses of `typing_extensions` is only after `if TYPE_CHECKING:`, but it won't be as straightforward as I had hoped.  I'm dropping this for now, and crossing my fingers that I don't inadvertently make the same mistake again...

PR #8631, opened 2025-02-26
Lambda: Layer Version Permission methods

archinksagar:
Implemented following methods to lambda
1. add_layer_version_permission
2. get_layer_version_policy
3. remove_layer_version_permission

Implemented following methods in DS
1. create_log_subscription
2. delete_log_subscription
3. list_log_subscriptions


moto/awslambda/policy.py
```diff
@@ -8,15 +8,11 @@
...
-if TYPE_CHECKING:
-    from .models import LambdaFunction
-
-
 TYPE_IDENTITY = TypeVar("TYPE_IDENTITY")
 
 
 class Policy:
-    def __init__(self, parent: "LambdaFunction"):
+    def __init__(self, parent):  # type: ignore[no-untyped-def]
```
bblommers, line 15: Can we improve the typing here and make it a `Union["LambaFunction", "LayerVersion"]`? If we make it explicit what type we expect, that should remove quite a bit of `type: ignore` statements

moto/awslambda/policy.py
```diff
@@ -78,12 +74,17 @@ def del_statement(self, sid: str, revision: str = "") -> None:
     # converts AddPermission request to PolicyStatement
     # https://docs.aws.amazon.com/lambda/latest/dg/API_AddPermission.html
     def decode_policy(self, obj: Dict[str, Any]) -> "Policy":
-        policy = Policy(self.parent)
+        policy = Policy(self.parent)  # type: ignore[no-untyped-call]
         policy.revision = obj.get("RevisionId", "")
+        # get function_arn or arn from parent
+        if hasattr(self.parent, "arn"):
```
bblommers, line 80: If we make the typing explicit, we can then use that info to clarify this if/else statement:
`if isinstance(self.parent, LayerVersion):`

That makes it a lot more readable for future developers.

bblommers (changes requested):
Hi @archinksagar! Thank you for the PR - just a few minor comments on the typing.

Can I also ask you to split the DS changes into a separate PR? It makes reviewing a lot easier, and it also ensures a cleaner Git history.

bblommers (approved):
Great - thank you @archinksagar!

PR #8648, opened 2025-03-02
DynamoDB: Path overlap check for all types of updates

BlizzardOf:
This addresses [issue 8582](https://github.com/getmoto/moto/issues/8582).

I discovered a lot about the AWS api's handling of overlapping paths while making test cases for this one:

- AWS does substitute attribute names in the error messages
- AWS only catches overlapping paths if neither path is an attribute name
- AWS does detect duplicate paths if one or more of them is an attribute name
- AWS splits the path into a list in the error message

Based on this, I found the following issues with the current validation:

- it only checked for conflicts between adds and deletes, or between set actions
- it only checked for duplicate paths
- it didn't handle attribute name substitution

I realized that by limiting a validation to an UpdateExpression, I could add top-level checks to the already existing set of validations for update statements. I felt this was more natural, and the expression names were already available, so I moved the validations that previously belongs to the AST object into the update expression validator.

However, this made it more difficult to do the update validation at the response level for transact_write_items. Rather than duplicate the whole thing, I had the transact_write_items backend pipe validation errors through the transaction failure error handling, since in AWS the validation occurs before any transaction execution.

As part of this I discovered that both updates and transactions can only have one "set" statement, so I added it to the validations as well.

### DIFF
--- a/moto/dynamodb/exceptions.py
+++ b/moto/dynamodb/exceptions.py
@@ -304,16 +304,16 @@ class InvalidAttributeTypeError(MockValidationException):
 
 
 class DuplicateUpdateExpression(InvalidUpdateExpression):
-    def __init__(self, name_1: List[str], name_2: List[str]):
+    def __init__(self, name_1: str, name_2: str):
         super().__init__(
-            f"Two document paths overlap with each other; must remove or rewrite one of these paths; path one: [{", ".join(name_1)}], path two: [{", ".join(name_2)}]"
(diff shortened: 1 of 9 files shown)


bblommers (changes requested):
Hi @BlizzardOf! First of all apologies for the late review - I just haven't had time to look into this.

The changes look great! Adding new validators is a very good approach, and it's nice to have all validation in one place now.
All the new tests pass against AWS, so that's a great step towards parity in this area.

Would you mind resolving the merge conflicts? I'd be happy to merge after that.

BlizzardOf:
Oh yeah meant to comment it's ready, thanks!

PR #8649, opened 2025-03-03
Add AWS Protocol Serializers

bpandola:
These serializers translate any backend result into a properly formatted HTTP response according to the [AWS protocol specifications](https://smithy.io/2.0/aws/protocols/index.html) and the `botocore` model data for a particular service/operation.

This PR contains a battery of low-level protocol tests to ensure adherence to the specifications.  Some future tweaks may be required for integration with individual AWS services.

Benefits:
* No need for Jinja XML templates in response classes.
* No need for `.to_dict()` methods on backend objects.
* No need for manually string formatting date values or coercing boolean values.

Next Steps:
These serializers will be slowly integrated into each individual service backend, starting with the services that will benefit the most (e.g. Query protocol-based services).  Additional features will be added as needs arise; defects will be addressed as they are encountered.

### DIFF
--- a/moto/core/serialize.py
+++ b/moto/core/serialize.py
@@ -1,0 +1,840 @@
+# mypy: disable-error-code="misc, override, var-annotated"
+"""Response serializers for the various AWS protocol specifications.
+
+There are some similarities among the different protocols with respect
+to response serialization, so the code is structured in a way to avoid
+code duplication where possible.  The diagram below illustrates the
+inheritance hierarchy of the response serializer classes.
+
+                           +--------------------+
+                           | ResponseSerializer |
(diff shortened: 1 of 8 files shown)


bpandola:
@bblommers Here are the serializers.  This PR is purely additive, with full test coverage.  I will work on integrating these with one of the simpler backends for each protocol (query, json, rest-xml, etc.).  I fully expect to encounter some issues that may involve some rethinking/refactoring, so I want to make sure I have a reference implementation for each protocol before we start doing this en masse.  As always, I value your feedback, so let me know any thoughts you may have.  I will request your review on the first few reference implementations.  Thanks!

PR #8656, opened 2025-03-06
[Network Manager] Core-Network add OwnerAccountID to list API

zkarpinski:
For future RAM mocking, adding OwnerAccountID CoreNetwork model.
https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/networkmanager/client/list_core_networks.html


bblommers (approved):
LGTM!

PR #8657, opened 2025-03-06
[Service Catalog] - Portfolio Access, Portfolio Share, and Portfolio Work

Singha22:
Resolves https://github.com/getmoto/moto/issues/8692


bblommers (approved):
LGTM @Singha22 - this looks very thorough. Thank you for the PR and for introducing  a new service to Moto!

PR #8660, opened 2025-03-07
WAFv2: fix incorrect ARNs in Cloudfront scope

tiurin:
### Issue
Reported in https://github.com/localstack/localstack/issues/12272.

ARN for WAF resources in Cloudfront scope should specify `us-east-1` region. Currently region is being set to `global` in moto.

I checked in AWS console that it stands true for all WAF resources: Web ACLs, IP sets, Regex pattern sets, Rule groups.

The regression was likely introduced in #7710.

### Solution
Region for global scope backend [defaults](https://github.com/getmoto/moto/blob/d00aa025b6c3d37977508b5d5e81ecad4ca15159/moto/core/base_backend.py#L272) to `"aws"`. It needs to be overridden with `us-east-1` for ARN purposes as Cloudfront-related resources are [managed from](https://docs.aws.amazon.com/waf/latest/APIReference/API_CreateIPSet.html#WAF-CreateIPSet-request-Scope) `us-east-1` and have it specified in ARN. However backend itself should stay separate for global and regional, to distinguish between WAF for us-east-1 and WAF for Cloudfront.

ARNs for global and regional resources for `us-east-1` differ only in resource name prefix - `global/` and `regional/` respectively.

### DIFF
--- a/moto/wafv2/responses.py
+++ b/moto/wafv2/responses.py
@@ -127,7 +127,11 @@ class WAFV2Response(BaseResponse):
 
     def list_tags_for_resource(self) -> TYPE_RESPONSE:
         arn = self._get_param("ResourceARN")
-        self.region = arn.split(":")[3]
+
+        # select correct backend - ARN region is not indicative by itself in case of WAF for Cloudfront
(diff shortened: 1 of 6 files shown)


bblommers (approved):
LGTM @tiurin - welcome to Moto!

PR #8669, opened 2025-03-10
test_cognitoidp: Load jwks-public.json using load_resource

cjwatson:
This works better if the tests are being run on an installed wheel, where `jwks-public.json` has been compressed; and it matches how this test already handles `jwks-private.json`.


bpandola (approved):
Thanks @cjwatson

PR #8672, opened 2025-03-11
s3 put object with object lock header check

Frank-Schmutz:
The current version of moto requires the `Content-Md5` header to be provided when uploading an object with a retention configured using object lock.
However the AWS documentation specifies that the `x-amz-sdk-checksum-algorithm` header may be provided instead [(AWS doc reference)](https://docs.aws.amazon.com/AmazonS3/latest/userguide/object-lock-managing.html#object-lock-put-object). I can confirm that the documentation is accurate after having performed a couple tests including the latter header instead of the content md5 one.
This issue is relevant since botocore does not compute and include the content md5 automatically since version 1.36.0, see [changelog](https://github.com/boto/botocore/blob/1.36.2/CHANGELOG.rst#L37).


Frank-Schmutz:
Resolves https://github.com/getmoto/moto/issues/8662

bpandola (changes requested):
@Frank-Schmutz Thanks for this.  Can you provide a test case that uses the `x-amz-sdk-checksum-algorithm` header?

Frank-Schmutz:
> @Frank-Schmutz Thanks for this. Can you provide a test case that uses the `x-amz-sdk-checksum-algorithm` header?

Test case is added

tests/test_s3/test_s3_lock.py
```diff
@@ -167,6 +167,58 @@ def test_locked_object_governance_mode(bypass_governance_retention, bucket_name=
...
+        ChecksumSHA256=checksum,
+    )
+
+    versions_response = s3_client.list_object_versions(Bucket=bucket_name)
+    response = s3_client.head_object(
+        Bucket=bucket_name,
+        Key=key_name,
+        VersionId=versions_response["Versions"][0]["VersionId"],
+    )
+    assert (
```
bblommers, line 212: I ran this test against AWS, and these two assertions actually fail - AWS does not return any Checksum-headers. 
This is the response I got from AWS, for reference:
```
{'AcceptRanges': 'bytes',
 'ContentLength': 4,
 'ContentType': 'binary/octet-stream',
 'ETag': '"098f6bcd4621d373cade4e832627b4f6"',
 'LastModified': datetime.datetime(2025, 3, 16, 12, 56, 11, tzinfo=tzutc()),
 'Metadata': {},
 'ObjectLockMode': 'GOVERNANCE',
 'ObjectLockRetainUntilDate': datetime.datetime(2025, 3, 16, 12, 56, 20, tzinfo=tzutc()),
 'ResponseMetadata': {'HTTPHeaders': {'accept-ranges': 'bytes',
                                      'content-length': '4',
                                      'content-type': 'binary/octet-stream',
                                      'date': 'Sun, 16 Mar 2025 12:56:11 GMT',
                                      'etag': '"098f6bcd4621d373cade4e832627b4f6"',
                                      'last-modified': 'Sun, 16 Mar 2025 '
                                                       '12:56:11 GMT',
                                      'server': 'AmazonS3',
                                      'x-amz-id-2': '...',
                                      'x-amz-object-lock-mode': 'GOVERNANCE',
                                      'x-amz-object-lock-retain-until-date': '2025-03-16T12:56:20Z',
                                      'x-amz-request-id': 'YK6TEPZDEADV07QH',
                                      'x-amz-server-side-encryption': 'AES256',
                                      'x-amz-version-id': 'PPafCyDPg7EvnM6plxyqX3Ciaqj6hQDT'},
                      'HTTPStatusCode': 200,
                      'HostId': '...',
                      'RequestId': '...',
                      'RetryAttempts': 0},
 'ServerSideEncryption': 'AES256',
 'VersionId': 'PPafCyDPg7EvnM6plxyqX3Ciaqj6hQDT'}
```

There is obviously something wrong in Moto's implementation, if we return these headers when we shouldn't. If you want to hunt down why this is happening, that would be very welcome, but not required - I'd be happy to accept this PR once it passes against AWS.

FYI, if you want to try it out:
> MOTO_TEST_ALLOW_AWS_REQUEST=true pytest -sv tests/test_s3/test_s3_lock.py::test_put_locked_object_with_checksum_algorithm

Adding this env variable will run the test against AWS, and it also automatically deletes the S3 bucket at the end, so you don't have to worry about having to manually delete resources.

bblommers (changes requested):
Thanks for adding the test @Frank-Schmutz - just added a minor remark about the assertions

tests/test_s3/test_s3_lock.py
Frank-Schmutz: I updated the test to pass against AWS. However it won't pass against moto now, because the response for a head operation do not contain the same headers, notably the `x-amz-checksum-sha256` for this particular case.

bpandola: @Frank-Schmutz Is #8675 related?

Frank-Schmutz: @bpandola Just had a look and yes that PR looks like it adds the checksum header returned by AWS.

bpandola: I fixed up and merged #8675, and now your test case is passing. 👍

bpandola (approved):
Thanks @Frank-Schmutz - welcome to Moto!

bpandola:
Just a quick note on the failed codecov report: those lines were not covered when they were originally added, and it'd be virtually impossible to create the scenario needed to trigger those conditions, so we'll have to let it slide.  Even without coverage, I do think it's valuable to have those checks in place on the off chance someone manages to trigger them.

Issue #8622, opened 2025-02-26
DynamoDBStreams: DescribeStream response deserialization error


pelegkaz:
Hi,

I'm using moto as a testcontainer with a golang sdk client in order to mock dynamodbstreams.
When calling DescribeStream, I get error deserializing CreationRequestDateTime.

operation error DynamoDB Streams: DescribeStream, https response error StatusCode: 200, RequestID: eNgJwGwokhpIul3c2cEJcocZj8maXMu83f4kF1d1OnwNs4m4jTBc, deserialization failed, failed to decode response body, expected Date to be a JSON Number, got string instead.

when debugging I see it happens here:

```
		case "CreationRequestDateTime":
			if value != nil {
				switch jtv := value.(type) {
				case json.Number:
					f64, err := jtv.Float64()
					if err != nil {
						return err
					}
					sv.CreationRequestDateTime = ptr.Time(smithytime.ParseEpochSeconds(f64))
				default:
					return fmt.Errorf("expected Date to be a JSON Number, got %T instead", value)
				}
			}
```

I think the fix is simply to change the formatting:
```
@@ -89,7 +90,7 @@ class DynamoDBStreamsBackend(BaseBackend):
                     "ENABLED" if table.latest_stream_label else "DISABLED"
                 ),
                 "StreamViewType": table.stream_specification["StreamViewType"],  # type: ignore[index]
-                "CreationRequestDateTime": table.stream_shard.created_on.isoformat(),  # type: ignore[union-attr]
+                "CreationRequestDateTime": unix_time(table.stream_shard.created_on),  # type: ignore[union-attr]


```
Thanks!

pareshchalke:
Hi, Team
We have a similar issue with DesribeBroker for MQ when using with Terraform

```
 Error: reading MQ Broker (37680b): operation error mq: DescribeBroker, https response error StatusCode: 200, RequestID: ogTQSVZkzyVV5VHRjzydyA1xuyOy7ie3nN6Y2a1AYC7aOTtJCAr9, deserialization failed, failed to decode response body with invalid JSON, expected __timestampIso8601 to be of type string, got json.Number instead
``` 

Please let me know if this should be a separate issue.

Thanks

bpandola:
Fixed in version `5.1.2.dev43`

PR #8673, opened 2025-03-12
Admin: Add Go tests

bpandola:
Add workflow for moto test cases using the AWS SDK for Go.

The Go setup and testing is working, but the job currently fails as a result of issue #8622.  I will submit a PR to fix that soon, but thought it was a good opportunity to add Go to moto's `other_langs` tests.


bpandola:
Pending merge of #8679

bblommers (approved):
Nice. I was planning to tackle this using Terraform, as that uses Go in the background, - but this is probably a better approach.

PR #8674, opened 2025-03-12
fix(autoscaling): Ensure launch template is passed when creating instances in autoscaling groups.

Kimamisa:
Ensure launch template is passed when creating instances in autoscaling groups.
This allows listing all disks which are created when an instance is created.


bblommers (approved):
LGTM - thank you for adding this to Moto @Kimamisa!

PR #8675, opened 2025-03-13
s3: fix {get,head}-object not returning the key's checksum value

terencehonles:
This change updates the get-object and head-object responses to include the ``x-amz-checksum-{algorithm}: {value}`` header. These are parsed by boto3, and validated if returned.


bpandola (approved):
Thanks @terencehonles - welcome to Moto!

Issue #8632, opened 2025-02-27
MQ: DescribeBroker response deserialization error


pareshchalke:
When using Terraform with Moto Server to create an RabbitMQ broker, the JSON response for DescribeBroker contains CreationTimestamp in a format that Terraform does not accept. This causes Terraform to fail deserialization with the following error:

```
│ Error: reading MQ Broker (37680b): operation error mq: DescribeBroker, https response error StatusCode: 200, RequestID: Kcit2L20LGYDSFFaKQDqP2HkrSw4t5bNSitq3VkPLsERZZZwspwZ, deserialization failed, failed to decode response body with invalid JSON, expected __timestampIso8601 to be of type string, got json.Number instead
```

Moto returns CreationTimestamp in a format that Terraform does not accept, causing a deserialization failure.

I believe it requires similar changes - https://github.com/getmoto/moto/pull/8569/commits/216e5861d5b66b51941cce56a54cf67496f04fb2#diff-6b97e164285716cccad08ed0a1c43a0ba6becf99589b1881d91b9a4a2c47b907

bpandola:
@pareshchalke would you be able to provide a minimal (failing) Terraform test case, similar to what we have [here](https://github.com/getmoto/moto/tree/afe69b56318093133182405ed22e0bbceb87959f/other_langs/terraform) for other services?

bpandola:
Fixed in version `5.1.2.dev43`

PR #8679, opened 2025-03-14
Core: add response serialization reference implementations for each AWS protocol

bpandola:
| Protocol | Service | Comments |
| :-----------------------: | --- | --- |
| `ec2` | Amazon Elastic Compute Cloud | Only migrated `TransitGateway` responses, which illustrates how this can be done piecemeal for larger service backends.  Still a good chunk of XML template code deleted, even for this one small part of EC2! |
| `json` - `1.0` | Amazon DynamoDB Streams | Currently has an open issue related to serialization (#8622). Very simple backend (no `exceptions.py` file). |
| `json` - `1.1` | Amazon Elastic Container Registry | This service had a lot of specialized serialization code polluting the backend models.  All gone now! |
| `rest-json` | AmazonMQ | Currently has an open issue related to serialization (#8632).  Lots of `.to_json()` methods deleted.  A good example of modelled exceptions (where HTTP status codes and extra error attributes are serialized according to the AWS service definition).|
| `rest-xml` | N/A | There are only a few backends for this protocol (`cloudfront`, `route53`, and `s3`) but they are complex and will need the AWS protocol-based request parsing I plan to add before we can proceed on converting them.
| `query` | Amazon Relational Database Service | Not part of this PR, but already integrated with the AWS response serializers. |

The diff here is actually pretty readable (especially if each commit is viewed on its own), but I have added some individual comments directly to areas of the changeset whose purpose may not be immediately obvious.  With one notable exception (called out in a commit message/code comment), any test updates were simply to increase coverage and provide additional assertions.

@bblommers - Let me know if you have any questions or concerns.  I was thinking I might also add a pinned issue detailing this ongoing effort.

### DIFF
--- a/moto/mq/models.py
+++ b/moto/mq/models.py
@@ -5,7 +5,7 @@ import xmltodict
 
 from moto.core.base_backend import BackendDict, BaseBackend
 from moto.core.common_models import BaseModel
-from moto.core.utils import unix_time
+from moto.core.utils import utcnow
 from moto.moto_api._internal import mock_random
 from moto.utilities.tagging_service import TaggingService
 from moto.utilities.utils import get_partition
@@ -28,15 +28,13 @@ class ConfigurationRevision(BaseModel):
         data: Optional[str] = None,
     ):
         self.configuration_id = configuration_id
-        self.created = unix_time()
+        self.created = utcnow()
         self.description = description
         self.is_invalid = False
         self.revision_id = revision_id
 
         if data is None:
-            self.data = base64.b64encode(
-                DEFAULT_CONFIGURATION_DATA.encode("UTF-8")
-            ).decode("utf-8")
+            self.data = DEFAULT_CONFIGURATION_DATA
         else:
             self.data = data
 
@@ -54,16 +52,9 @@ class ConfigurationRevision(BaseModel):
             # If anything fails, lets assume it's not LDAP
             return False
 
-    def to_json(self, full: bool = True) -> Dict[str, Any]:
-        resp = {
-            "created": self.created,
-            "description": self.description,
-            "revision": int(self.revision_id),
-        }
-        if full:
-            resp["configurationId"] = self.configuration_id
-            resp["data"] = self.data
-        return resp
+    @property
+    def revision(self) -> str:
+        return self.revision_id
 
 
 class Configuration(BaseModel):
@@ -77,7 +68,7 @@ class Configuration(BaseModel):
     ):
         self.id = f"c-{mock_random.get_random_hex(6)}"
         self.arn = f"arn:{get_partition(region)}:mq:{region}:{account_id}:configuration:{self.id}"
-        self.created = unix_time()
+        self.created = utcnow()
 
         self.name = name
         self.engine_type = engine_type
@@ -114,18 +105,10 @@ class Configuration(BaseModel):
     def get_revision(self, revision_id: str) -> ConfigurationRevision:
         return self.revisions[revision_id]
 
-    def to_json(self) -> Dict[str, Any]:
+    @property
+    def latest_revision(self) -> ConfigurationRevision:
         _, latest_revision = sorted(self.revisions.items())[-1]
-        return {
-            "arn": self.arn,
-            "authenticationStrategy": self.authentication_strategy,
-            "created": self.created,
-            "engineType": self.engine_type,
-            "engineVersion": self.engine_version,
-            "id": self.id,
-            "name": self.name,
-            "latestRevision": latest_revision.to_json(full=False),
-        }
+        return latest_revision
 
 
 class User(BaseModel):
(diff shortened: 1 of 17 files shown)


moto/mq/models.py
```diff
@@ -28,15 +28,13 @@ def __init__(
         data: Optional[str] = None,
     ):
         self.configuration_id = configuration_id
-        self.created = unix_time()
+        self.created = utcnow()
```
bpandola, line 31: Backend models should just use native Python `datetime` objects.  Proper formatting will be handled by the serializer.  (The serializer is forgiving, though, so a timestamp should still work if it's a string, which gives some wiggle room if some are missed when migrating services to the new serialization.)

moto/mq/responses.py
```diff
@@ -89,32 +89,32 @@ def update_broker(self) -> str:
...
-    def describe_broker(self) -> str:
+    def describe_broker(self) -> ActionResult:
         broker_id = self.path.split("/")[-1]
         broker = self.mq_backend.describe_broker(broker_id=broker_id)
-        resp = broker.to_json()
-        resp["tags"] = self.mq_backend.list_tags(broker.arn)
-        return json.dumps(resp)
+        resp = copy.copy(broker)
+        setattr(resp, "tags", self.mq_backend.list_tags(broker.arn))
+        return ActionResult(resp)
```
bpandola, line 102: Generally speaking, backend models should try to align with response models as closely as possible, but the response methods will still have the ability to pull information from disparate sources (multiple backend methods in this case) and combine them into one result.  This particular example is a little hacky, but works with the existing code.  Alternatively, a `tags` attribute could be added to the `broker` backend model.

tests/test_ecr/test_ecr_policy_validation.py
```diff
@@ -70,8 +70,7 @@ def test_validate_error_parse(policy):
 
     # then
     ex = e.value
-    assert ex.code == 400
-    assert ex.error_type == "InvalidParameterException"
+    assert ex.code == "InvalidParameterException"
```
bpandola, line 73: HTTP status codes are a serialization concern (and are contained in the AWS service models).  Moto service exceptions should no longer need to define them directly.

tests/test_mq/test_mq.py
```diff
@@ -305,12 +305,16 @@ def test_describe_broker_unknown():
 
     with pytest.raises(ClientError) as exc:
         client.describe_broker(BrokerId="unknown")
-    err = exc.value.response["Error"]
-    assert err["Code"] == "NotFoundException"
+    resp = exc.value.response
+    error = resp["Error"]
+    metadata = resp["ResponseMetadata"]
+    assert metadata["HTTPStatusCode"] == 404
```
bpandola, line 311: For exceptions that do map to a particular status code, adding an assert to ensure that the serializer is outputting the right one is recommended.

bblommers (approved):
LGTM overall @bpandola! Nice to see this taking shape. 

Just a question on the `ActionResult` - do you see this as  a permanent fixture, or only as a stopgap until we've moved all services to the new serialization method?

And an unrelated musing: there are quite a lot of people who patch parts of Moto to improve the behaviour somehow, and make it work for them. I'm guessing that all these changes will break their custom patches. Note that this is not a criticism, or a reason to stop the process! Only to say that the transition may ruffle some feathers. :slightly_smiling_face:

bpandola:
> LGTM overall @bpandola! Nice to see this taking shape.
> 
> Just a question on the `ActionResult` - do you see this as a permanent fixture, or only as a stopgap until we've moved all services to the new serialization method?
> 
> And an unrelated musing: there are quite a lot of people who patch parts of Moto to improve the behaviour somehow, and make it work for them. I'm guessing that all these changes will break their custom patches. Note that this is not a criticism, or a reason to stop the process! Only to say that the transition may ruffle some feathers. 🙂

1) My current thinking is that `ActionResult` will stay.  1) It avoids having to individually type the return value of every single `responses.py` method, which would be a huge pain for very little value. 2) It's extensible.  We could have a `FilteredResult` or `PaginatedResult` class that could encapsulate that kind of functionality without having to manually put that code into every action method that needs it.  3) It clearly demarcates a specific point in the request/response pipeline, such that in the future we could support pre- and post-action events--similar to the botocore event system that we already hook into--if we needed some specialized processing before/after serialization.  _Is there a downside I'm not considering?_

2) Do you happen to have an example of such a patch that would break?  As a for instance: if someone adds a method to a service's `responses.py` class and is still returning a string template, that will work just as it always has.  Anyone using moto in server mode should be fine...  If they're poking directly at the backend (and I do that in some of my tests that use moto), something could definitely break there, but would hopefully be an easy fix.  Does moto make any guarantees or promises about its internals?  I will definitely pin an issue that describes the motivation/context behind these changes, and we'll see if anyone has anything to say.  I do think the value here is just too hard to ignore.  Ultimately, it should be much *easier* to contribute to moto (and to customize it) once the backend has been decoupled from serialization concerns.  (All that said, I'm happy to slow this down if you think it prudent.  I got my RDS changes upstreamed already, which was the real impetus behind all of this work. 🙂)

bblommers:
There is no downside that I can see for ActionResult @bpandola - I was just curious. It does make sense to standardize the expected result.

And no, I don't have an example patch. There are no guarantees about it's internals, I'm always quick to point that out to people for exactly this reason - so we're free to change what we want. People just always surprise me in how they use Moto, so I'm sure we'll see someone asking for guidance on how to patch things in the new world.

And definitely not a reason to slow down! I'm loving these changes. :slightly_smiling_face:

PR #8680, opened 2025-03-14
IAM Access Control: Add logic with trust relationship verification and role-based access

ArteIIo:
Main purpose of this PR is to introduce to additional feature for iam-like workflows:
- Validate trust relationship's principal and external id condition during assume role operation.
- Add role-based control access by resolving role resource on request.

In case of non-existing Role for both workflows system'll not block the workflow.

Introducing this addition to the codebase would increase library usability for verification access-based scenarios.


moto/iam/access_control.py
```diff
@@ -258,9 +263,41 @@ def check_action_permitted(self, resource: str) -> None:
...
 
+    def _is_assuming_role_operation(self, resource_arn: str) -> bool:
+        if ":role" not in resource_arn.lower():
+            return False
+
+        if not self.backend.has_role_by_arn(resource_arn):
+            return False
+
+        self.backend.get_role_by_arn(resource_arn)
+        return self._action in REQUIRE_RESOURCE_ACCESS_POLICIES_CHECK
```
bblommers, line 280: This loops through all the IAM roles twice - first to see if it exists, and then again to return it. It would be faster to only run through the loop once, by only using the existing `get_role_by_arn`-method:

```suggestion
        try:
            self.backend.get_role_by_arn(resource_arn)
            return self._action in REQUIRE_RESOURCE_ACCESS_POLICIES_CHECK
        except IAMNotFoundException:
            return False
```

moto/iam/access_control.py
```diff
@@ -258,9 +263,41 @@ def check_action_permitted(self, resource: str) -> None:
             elif permission_result == PermissionResult.PERMITTED:
                 permitted = True
 
+        if self._is_assuming_role_operation(resource):
+            permitted = self._check_role_trust_relationship(resource)
```
bblommers, line 267: This should still take the result of the policies into account, right? So something like:

```suggestion
            permitted = permitted and self._check_role_trust_relationship(resource)
```

moto/iam/access_control.py
```diff
@@ -459,6 +518,25 @@ def _check_element_matches(self, statement_element: Any, value: str) -> bool:
...
+            or principal == aws_expected_principals
+            or principal in aws_expected_principals
+        )
+
+    def _check_conditions(self, conditions: Dict[str, Dict[str, str]]) -> bool:
+        expected_conditions = self._statement.get("Condition")
+        if not expected_conditions:
+            return True
+
+        return conditions == expected_conditions
```
bblommers, line 538: There could be multiple 'expected' conditions, right? The policy could say something like:
```
                "Condition": {
                    "StringEquals": {
                        "sts:ExternalId": external_id,
                        "awsAccountId": ...
                    },
                   "StringLike": {
                       ....
                   }
                },
```
So this logic should be a bit more advanced - it can't be a simple comparison.

moto/iam/access_control.py
```diff
@@ -425,6 +475,15 @@ def is_action_permitted(
         if is_action_concerned:
             if self.is_unknown_principal(self._statement.get("Principal")):
                 return PermissionResult.NEUTRAL
+            elif principal and not self._check_principal(principal):
+                return PermissionResult.DENIED
+
+            if conditions and not self._check_conditions(conditions):
+                return PermissionResult.DENIED
+
+            if not self._statement.get("Resource"):
```
bblommers, line 484: This is only true for AssumeRolePolicies, right? For other policies, the Resource is always mandatory. Would be good to have some documentation (i.e, a simple comment) to explain why this if-statement holds true.

bblommers (changes requested):
Hi @ArteIIo, thanks for the PR! Please see my comments.

Note that I haven't had time to look at the tests yet - but I know this PR has been open for some time, so I didn't want to keep you waiting any longer. Once the comments are addressed, I'll have another pass at this.

moto/iam/access_control.py
ArteIIo: Added comment before check

moto/iam/access_control.py
ArteIIo: Redesign conditions evaluation process with extra entities

moto/iam/access_control.py
ArteIIo: Fixed

moto/iam/access_control.py
ArteIIo: Fixed

ArteIIo:
@bblommers I've addressed all the comments. I apologize for a long procession from my side. During that I've redesigned the policies conditions evaluation for it to be more generic and agile. Still there is space for enchantments. As of now it covers `StringEquals` operation and only `sts:ExternalId`as the condition value source. For resolving internal values for condition evaluation there'll be a need to introduce additional resolver by Context key. I've added appropriate comment for that as well for it to the introduces.

bblommers (approved):
Ah, there is is always space for enhancements @ArteIIo - but this already looks great. Thank you for adding this to Moto!

PR #8682, opened 2025-03-14
chore: update SSM default parameters

bpandola:
Reopened from original PR (#8667) that broke the build after it was merged and had to be reverted.


bblommers:
Thanks for handling this @bpandola!

PR #8683, opened 2025-03-15
ECR: describe_image_scan_findings: Allow custom results

bblommers:
See #8586


bpandola:
@bblommers Would you consider trying to implement this after #8679 is merged?  Might be a useful test to see if the new serialization stuff helps or hinders you--or if there are any points of confusion/friction that I can address.

bblommers:
> @bblommers Would you consider trying to implement this after #8679 is merged? Might be a useful test to see if the new serialization stuff helps or hinders you--or if there are any points of confusion/friction that I can address.

Will do!

PR #8688, opened 2025-03-17
RDS: implement CopyDBParameterGoup

dtx14189:
Added support for rds API: copy_db_parameter_group.

In addition, I fixed an incorrect key in the response for describe_db_parameter.

### DIFF
--- a/moto/rds/models.py
+++ b/moto/rds/models.py
@@ -38,6 +38,7 @@ from .exceptions import (
     DBClusterToBeDeletedHasActiveMembers,
     DBInstanceAlreadyExists,
     DBInstanceNotFoundError,
+    DBParameterGroupAlreadyExistsError,
     DBParameterGroupNotFoundError,
     DBProxyAlreadyExistsFault,
     DBProxyNotFoundFault,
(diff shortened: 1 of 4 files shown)


moto/rds/responses.py
```diff
@@ -370,7 +377,7 @@ def modify_db_parameter_group(self) -> ActionResult:
     def describe_db_parameters(self) -> ActionResult:
         db_parameter_group_name = self.parameters.get("DBParameterGroupName")
         db_parameter_groups = self.backend.describe_db_parameter_groups(
-            {"name": db_parameter_group_name}
+            {"db_parameter_group_name": db_parameter_group_name}
```
bpandola, line 380: Good catch.  It's interesting that this wasn't picked up by our existing tests... This defect meant that calling `DescribeDBParameterGroups(DBParameterGroupName="group_name")` was the same as calling `DescribeDBParameterGroups()` with no parameters, which only worked because we were only ever dealing with one DBParameterGroup in our tests.  Anyway, thanks for running into this and fixing it!

moto/rds/models.py
```diff
@@ -2886,6 +2887,48 @@ def create_db_parameter_group(
...
+        if source_db_parameter_group_identifier.startswith("arn:aws:rds:"):
+            source_db_parameter_group_identifier = (
+                source_db_parameter_group_identifier.split(":")[-1]
+            )
+        if source_db_parameter_group_identifier not in self.db_parameter_groups:
+            raise DBParameterGroupNotFoundError(source_db_parameter_group_identifier)
+        if target_db_parameter_group_identifier in self.db_parameter_groups:
+            raise DBParameterGroupAlreadyExistsError(
+                target_db_parameter_group_identifier
+            )
```
bpandola, line 2906: Can you add test coverage for these two exceptions?  Individual tests preferred, but it's also fine if you just cover these two conditions within your existing test.

moto/rds/models.py
```diff
@@ -2886,6 +2887,48 @@ def create_db_parameter_group(
...
+        source_db_parameter_group = self.db_parameter_groups[
+            source_db_parameter_group_identifier
+        ]
+
+        target_db_parameter_group = DBParameterGroup(
+            backend=source_db_parameter_group.backend,
```
bpandola, line 2912: This should be `backend=self`, right?  It seems AWS supports an ARN for the source parameter group, which means it could be in a different region, which would be the wrong backend for our target parameter group.

moto/rds/models.py
```diff
@@ -2886,6 +2887,48 @@ def create_db_parameter_group(
...
+            db_parameter_group_name=target_db_parameter_group_identifier,
+            db_parameter_group_family=source_db_parameter_group.family,
+            description=target_db_parameter_group_description,
+            tags=tags,
```
bpandola, line 2916: What's the AWS behavior with tags here?  If the source parameter group has tags will those be copied over to the target, or does the target only get tags if passed in?  Whatever the behavior, can you add test coverage to make it explicit?

bpandola (changes requested):
@dtx14189 Welcome to moto!  This looks good, just a bit of feedback on additional test coverage.  Thanks.

moto/rds/models.py
dtx14189: Tests added. Thank you.

moto/rds/models.py
dtx14189: Yeah, you are correct. I updated the code, thank you

bpandola (approved):
Thanks @dtx14189 - Nice work.

PR #8691, opened 2025-03-17
[Redshift] Logging - Enable, Disable, Describe

zkarpinski:
Adds the following redshift endpoints:

- [Enable_Logging](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/redshift/client/enable_logging.html)
- Disable_logging
- Describe_logging

### DIFF
--- a/moto/redshift/responses.py
+++ b/moto/redshift/responses.py
@@ -777,16 +777,15 @@ class RedshiftResponse(BaseResponse):
             log_destination_type=log_destination_type,
             log_exports=log_exports,
         )
-        template = self.response_template(ENABLE_LOGGING_TEMPLATE)
-        return template.render(
-            logging_enabled=config["LoggingEnabled"],
-            bucket_name=config["BucketName"],
-            s3_key_prefix=config["S3KeyPrefix"],
-            last_successful_delivery_time=config["LastSuccessfulDeliveryTime"],
-            last_failure_time=config["LastFailureTime"],
-            last_failure_message=config["LastFailureMessage"],
-            log_destination_type=config["LogDestinationType"],
-            log_exports=config["LogExports"],
+        return self.get_response(
+            {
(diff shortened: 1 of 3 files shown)


moto/redshift/responses.py
```diff
@@ -763,3 +763,123 @@ def get_cluster_credentials(self) -> str:
...
+    <LastFailureTime>{{ last_failure_time }}</LastFailureTime>
+    <LastFailureMessage>{{ last_failure_message }}</LastFailureMessage>
+    <LogDestinationType>{{ log_destination_type }}</LogDestinationType>
+    <LogExports>
+{% for logexport in logexports %}
+      <member/>
+{% endfor %}
+    </LogExports>
+  </DescribeLoggingStatusResult>
+</DescribeLoggingStatusResponse>"""
```
bpandola, line 885: Why not follow existing convention here?  None of the other endpoints use XML templates...  I recall the json/xml stuff with Redshift being a boto/boto3 thing, so it may be time to look into whether that complexity is still necessary, but this looks like a case where following the existing convention would actually be easier and less code, no?

zkarpinski: I can make the change. The scaffolding script put this convention, so I just rolled with it.

bpandola: Ah, gotcha.  Not a big deal.  Like I said, the Redshift backend in `moto` is a bit of an anomaly in that it supports both JSON and XML output, which is an edge-case the scaffold script obviously doesn't detect.  Anyway, if you can make things consistent, that'd be great.  (I'll follow-up on whether or not we can drop this dual serialization stuff in a future PR.)  Thanks.

zkarpinski:
@bblommers Anything need to be fixed here or just flaky test? Looks like some test proxies worked.

bpandola (approved):
Thanks @zkarpinski

PR #8696, opened 2025-03-19
EC2: ModifyLaunchTemplate

Peeanio:
Hi, this pull request is to add the ModifyLaunchTemplate response to EC2. The action is used to set the default version of the launch template (https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/ec2/client/modify_launch_template.html). I have linted the commits according to the contribution guide and tests pass.

This is my first pull request, and if there is additional work required or other feedback, I'm happy to work through that.


bblommers (approved):
LGTM - thank you for adding this feature to Moto @Peeanio!

PR #8702, opened 2025-03-21
Optimization for autoscaling groups

pcosquer:
Some optimization for autoscaling groups and volumes, to speed up lookup by instance id


pcosquer:
Sorry, this was mean for a fork, there are some commits that are too bad to be mergeable

PR #8704, opened 2025-03-21
[CloudDirectory] New Resource support

### LINKED ISSUE #8703: [CloudDirectory] Add support for Cloud Directory service mocking
- [ ] CreateDirectory
- [ ] GetDirectory
- [ ] DeleteDirectory
- [ ] ListDirectories
- [ ] TagResource
- [ ] UnTagResource
- [ ] Add to ResourceGroupTaggingAPI

zkarpinski:
Resolves #8703 

- CreateDirectory
- GetDirectory
- DeleteDirectory
- ListDirectories
- TagResource
- UnTagResource
- Added to ResourceGroupTaggingAPI


moto/clouddirectory/responses.py
```diff
@@ -0,0 +1,93 @@
...
+
+    def list_directories(self) -> str:
+        next_token = self._get_param("NextToken")
+        # max_results = self._get_param("MaxResults")
+        state = self._get_param("State")
+        directories = self.clouddirectory_backend.list_directories(
+            state=state,
+        )
+        directory_list = [directory.to_dict() for directory in directories]
+        return json.dumps(dict(Directories=directory_list, NextToken=next_token))
```
bblommers, line 46: This should only return a `NextToken` if pagination is actually implemented - otherwise boto3 will end up in an endless loop when using `boto3_client.get_paginator('list_directories')`

moto/resourcegroupstaggingapi/models.py
```diff
@@ -265,6 +266,22 @@ def format_tag_keys(
...
+                ]
+                for directory in clouddirectory_backend.directories.values():
+                    tags = clouddirectory_backend.tagger.list_tags_for_resource(
+                        directory.directory_arn
+                    )["Tags"]
+                    if not tags or not tag_filter(tags):
+                        continue
+                    yield {"ResourceARN": f"{directory.directory_arn}", "Tags": tags}
+            except Exception:
+                pass
```
bblommers, line 283: What kind of exception can happen here, and why do we ignore it? Note that I don't mind - but it would be nice to have a comment explaining what's happening for future reference :slightly_smiling_face:

bblommers (changes requested):

moto/resourcegroupstaggingapi/models.py
bblommers: Ah, that makes sense, if the `CloudDirectory` service is not available in that region.

It's probably cleaner to use the example from `LexV2` then, with a custom method that returns  `None` if no region exists. Then the whole section can be if-guarded instead, to make sure it only runs when `CloudDirectory` is available.

zkarpinski:
Corrections made: 
Updated resourcegroupstaggingapi logic to handle the regions for cloud-directory
Updated pagination logic & states filter for list_

bblommers (approved):
Great - thanks for the fixes @zkarpinski!

PR #8706, opened 2025-03-22
Admin: Up minimum botocore version to 1.20.88

bblommers:
Ups the minimum botocore version that supports both `is_document_type` and `error_code` - both properties of `botocore.model` that are now used by the serialization code, but were only introduced in [1.1.6.0](https://github.com/boto/botocore/commit/1d91dbbf23ab883f3c92fe7f6c22e0fc726d5c7d) and [1.20.88](https://github.com/boto/botocore/commit/3b87b23002d99df480b147252897b45fd10241a1) respectively.

Also updates the outdated dependency tests to verify that the tests pass with this version.

FYI @bpandola

### DIFF
--- a/.github/workflows/test_outdated_versions.yml
+++ b/.github/workflows/test_outdated_versions.yml
@@ -13,12 +13,12 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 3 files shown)


bpandola:
@bblommers Good catch. Thanks!

PR #8718, opened 2025-03-25
Security Hub Additional Org Functions

Singha22:


tests/test_securityhub/test_securityhub.py
```diff
@@ -154,3 +157,157 @@ def test_get_findings_max_results():
...
+        assert "Administrator" in admin_response_1
+        assert (
+            admin_response_1["Administrator"]["AccountId"] == alternate_admin_account_id
+        )
+        assert admin_response_1["Administrator"]["MemberStatus"] == "ENABLED"
+
+
+# Test passes here but fails on Github Actions CI
+# @mock_aws
+# def test_organization_auto_enable_disabled():
```
bblommers, line 259: Hi @Singha22, can you explain why this is failing in GithubActions? Is it related to the fact that the environment variables are not updated in ServerMode?

If that's the case, we tend to skip the test instead:

https://github.com/getmoto/moto/blob/4eb407460e04f9392fefa9cd4a685648ab693a8c/tests/test_organizations/test_organizations_boto3.py#L46C1-L48C1

bblommers (changes requested):

tests/test_securityhub/test_securityhub.py
bblommers: No, I've never seen that before @Singha22. It runs locally on my machine as well, which is interesting.

Let's enable it anyway, and just disable it for ServerMode tests:
```python
from unittest import SkipTest

from moto import mock_aws, settings

def test_...():
        raise SkipTest("Can only change setting in DecoratorMode")
```
If the CI still fails, I'll have another look.

Singha22: Hey Bert, looks like by adding this it got resolved. The current pipeline error is just `test_stepfunctions` but that's unrelated to the code I have. Thanks for looking into it.

moto/securityhub/models.py
```diff
@@ -30,6 +31,17 @@ class SecurityHubBackend(BaseBackend):
         }
     }
 
+    org_admin_account_details = {
```
bblommers, line 34: Having this data structure globally doesn't work well with the fact that Moto can support [multiple accounts](https://docs.getmoto.org/en/latest/docs/multi_account.html), so I don't think this is correct.

In theory, a user could have a setup with four accounts, `A`, `B`, `C` and `D`, where `A` is the admin account of `B` and `C` is the admin account of `D`. This data structure does not support that scenario. IMO, we should move this inside the `SecurityHub`-backend, so that it is specific to an account.

The [docs for this feature](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/securityhub/client/enable_organization_admin_account.html) say:
> Designates the Security Hub administrator account for an organization. Can only be called by the organization management account.

If we want to support this, we can - but that would mean integrating it with the `Organizations`-service. That is **not** a requirement for this PR to pass, to be clear! Just pointing out that that would be the 'proper' way to validate that we have only have a single admin, and that it's part of the organization.

bblommers (changes requested):

tests/test_securityhub/test_securityhub.py
```diff
@@ -154,3 +157,423 @@ def test_get_findings_max_results():
...
+    admin_account = next(acc for acc in accounts if acc["Name"] == "SecurityHubAdmin")
+    admin_account_id = admin_account["Id"]
+
+    client = boto3.client("securityhub", region_name="us-east-1")
+    client.enable_organization_admin_account(AdminAccountId=admin_account_id)
+
+    get_resp = client.get_administrator_account()
+    assert "Administrator" not in get_resp
+
+    with mock.patch.dict(os.environ, {"MOTO_ACCOUNT_ID": admin_account_id}):
```
bblommers, line 183:
```suggestion
    if not settings.TEST_DECORATOR_MODE:
        # Can't set env variables in ServerMode
        return
    
    with mock.patch.dict(os.environ, {"MOTO_ACCOUNT_ID": admin_account_id}):
```
We will have to skip these parts in ServerMode. Moto runs in a separate process there, so updating environment variables does not work 

Exitting early allows us to keep the rest of the test intact, so we still test the 'regular' (non-multi account) parts.

Note that settings should be a direct import from moto:
```
from moto import settings
```

This applies to all the other tests where we set environment variables as well, of course.

bblommers (changes requested):
Hi @Singha22, thanks for adding the multi-account functionality! I've just have a small comment to fix the tests, happy to merge once the CI pipeline passes.

PR #8720, opened 2025-03-25
Route53: Hosted zone custom id support

cloutierMat:
# Motivation

This pr adds support for creating hosted zone with deterministic ids.

# Changes

Added the `moto_id` wrapper to `create_route53_zone_id` to enable the id manager to return a set custom id


bblommers (approved):
LGTM!

PR #8737, opened 2025-03-31
[MediaConnect] Accept Maintenance value during MediaConnect flow creation

flaker:
The parameter to set the desired maintainance window schedule when creating a Flow is missing.


bblommers:
> I am not sure if someone triggers the test run? I see some random dynamodb test failure (`test_state_machine_calling_dynamodb_put_wait_for_task_token`), along some other random failure? now I see an s3 one.

Yes, a maintainer needs to explicitly trigger the tests for new contributors. These failing tests seem to be flakes though, so don't worry about it.

bblommers (approved):
LGTM - thank you for contributing this feature @flaker, and welcome to Moto!

PR #8752, opened 2025-04-01
S3: delete_objects(): BypassGovernanceMode should handle both True and true

bblommers:
Looks like the casing changed (`True` -> `true` in a recent `botocore` update (potentially [this one](https://pypi.org/project/botocore/1.37.25/)), and Moto didn't understand the provided parameter anymore. This should fix the tests on master.


bpandola:
Just out of curiosity, I went digging for the change.  It's [here](https://github.com/boto/botocore/blame/690efa8ef21bdeadcbd3adc6bb9687555b0f71fa/botocore/serialize.py#L961-L962).

PR #8755, opened 2025-04-02
Feature/cluster tasks setting

ColtonHerrod:
Adding a basic setting to allow generating an ECS cluster with pre-generated running tasks or pending tasks.  This is required for some cluster operations I'm looking to do, and doesn't look to be possible when generating tasks or services through those dedicated functions.

To that end, this PR allows setting arbitrary numbers of running or pending tasks on the cluster using the same pattern as is currently used for ECS services.  If this isn't the preferred pattern, let me know and I'll circle around to take another approach.

- [x] Feature is added
- [x] The linter is happy
- [x] Tests are added
- [x] All tests pass, existing and new

### DIFF
--- a/moto/ecs/models.py
+++ b/moto/ecs/models.py
@@ -77,9 +77,7 @@ class Cluster(BaseObject, CloudFormationModel):
         self.active_services_count = 0
         self.arn = f"arn:{get_partition(region_name)}:ecs:{region_name}:{account_id}:cluster/{cluster_name}"
         self.name = cluster_name
-        self.pending_tasks_count = 0
         self.registered_container_instances_count = 0
-        self.running_tasks_count = 0
         self.status = "ACTIVE"
         self.region_name = region_name
         self.settings = cluster_settings or [
@@ -90,6 +88,23 @@ class Cluster(BaseObject, CloudFormationModel):
         self.default_capacity_provider_strategy = default_capacity_provider_strategy
         self.tags = tags
         self.service_connect_defaults = service_connect_defaults
+
+        try:
+            # negative running count not allowed, set to 0 if so
+            running_tasks_count = max(int(getenv("MOTO_ECS_CLUSTER_RUNNING", 0)), 0)
+        except ValueError:
+            # Unable to parse value of MOTO_ECS_CLUSTER_RUNNING as an integer, set to default 0
+            running_tasks_count = 0
+
+        try:
+            # negative running count not allowed, set to 0 if so
+            pending_tasks_count = max(int(getenv("MOTO_ECS_CLUSTER_PENDING", 0)), 0)
+        except ValueError:
+            # Unable to parse value of MOTO_ECS_CLUSTER_PENDING as an integer, set to default 0
+            pending_tasks_count = 0
+
+        self.running_tasks_count = running_tasks_count
+        self.pending_tasks_count = pending_tasks_count
 
     @property
     def physical_resource_id(self) -> str:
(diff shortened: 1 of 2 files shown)


moto/ecs/models.py
```diff
@@ -91,6 +89,23 @@ def __init__(
         self.tags = tags
         self.service_connect_defaults = service_connect_defaults
 
+        try:
+            # negative running count not allowed, set to 0 if so
+            running_tasks_count = max(int(getenv("MOTO_ECS_CLUSTER_RUNNING", 0)), 0)
```
bblommers, line 94: Considering this is specifically about tasks - do you think it makes sense to rename it to `MOTO_ECS_CLUSTER_TASKS_RUNNING`?

Note that my domain knowledge of ECS is limited, so feel free to push back on this! Just a suggestion, but it looks like that a `Cluster` could also be in a `Running` state, which is different from what we do here.

bblommers, line 94: It would be nice to add both configuration options to the documentation as well. If you add this to the existing class-documentation of `ECSBackend`, it will be semi-automatically picked up and added to [the ECS docs](https://docs.getmoto.org/en/latest/docs/services/ecs.html).

tests/test_ecs/test_ecs_boto3.py
```diff
@@ -47,6 +47,40 @@ def test_create_cluster_with_setting():
     assert cluster["serviceConnectDefaults"] == {"namespace": "ns"}
 
 
+@mock_aws
+def test_create_cluster_with_running_tasks():
+    running_tasks_count = 3
```
bblommers, line 52: I haven't run the tests yet - but I know both will fail in ServerMode, because that runs Moto in a different process and the environment variables will not be applied.

We usually just skip these tests if that's the case:
```python
    raise SkipTest("Can't set environment variables in server mode")
```

moto/ecs/models.py
bblommers, line 94: Third and hopefully last point:

When a user calls the `RunTask`-method, we do update `self.running_tasks_count`.

So it might make more sense to leave `self.running_tasks_count` alone, and return the user-supplied env-var in the `response_object` instead (if supplied)? It might be a bit confusing otherwise, if a user configures a specific value and Moto overrides it without warning.

bblommers (changes requested):

moto/ecs/models.py
ColtonHerrod: Gotcha, that makes sense.  I think I missed that behavior.  The context of this is that I'd like to mimic a running service on a given cluster, and adding a service with desired tasks set the **service** tasks correctly but not the **cluster** tasks.

This seemed like a straightforward way to mimic that behavior, but I may look into refactoring this to make those updates as part of the `create_service` workflow instead.

ColtonHerrod: Actually, after some investigation, it doesn't look like the `run_task` command updates the cluster task count.  It does update the container instance count here on [line 1419](https://github.com/getmoto/moto/blob/master/moto/ecs/models.py#L1419), but there doesn't appear to be an equivalent call to the cluster itself to update those resources.

I think the proper approach there would be to add an `update_cluster_resources` function an call it on service/task creation.  How does that sound?

ColtonHerrod: I've removed this in favor of dynamically updating cluster tasks on `run_tasks` call.

ColtonHerrod: Updated the class documentation: https://github.com/getmoto/moto/pull/8755/files#diff-b1b4c8346e32bf53243d2f29dc4c11bad4fae156fc2101fdb24c0c272dbcc173R1037

PR #8759, opened 2025-04-02
ServiceCatalog-AppRegistry Backend

Peeanio:
Adding service catalog app registry backend. Supports creating/listing applications and associating resources. Associated resources can be tag value or Cloudformation stacks. AWS API raises "ResourceNotFoundException" when the CF stack is not present, but not sure how to search for the stack internally on the backend, or if it is required. 

Happy to work through anything else missing.


moto/servicecatalogappregistry/responses.py
```diff
@@ -0,0 +1,87 @@
...
+    def associate_resource(self) -> str:
+        application = unquote(self._get_param("application"))
+        resource_type = self._get_param("resourceType")
+        resource = unquote(self._get_param("resource"))
+        options = self._get_param("options")
+        app = None
+        app = self._find_app_by_any_value(application)
+        if options is None:
+            options = []
+        new_resource = AssociatedResource(
```
bblommers, line 57: The Backend also has an `associate_resource`-method which does the same thing, we would be the 'proper' way to do it - did you mean to use that instead?

moto/servicecatalogappregistry/models.py
```diff
@@ -0,0 +1,163 @@
...
+                r"^arn:aws:cloudformation:(us(-gov)?|ap|ca|cn|eu|sa)-(central|(north|south)?(east|west)?)-\d:\d{12}:stack/\w[a-zA-Z0-9\-]{0,127}/[a-f0-9]{8}(-[a-f0-9]{4}){3}-[a-f0-9]{12}$",
+                resource,
+            )
+            if match is not None:
+                self.name = resource.split("/")[1]
+            else:
+                self.name = resource
+            cf_backend = cloudformation_backends[account_id][region_name]
+            try:
+                cf_backend.get_stack(self.name)
```
bblommers, line 77: Nice to see an integration with other services immediately builtin :+1:

bblommers (changes requested):
Hi @Peeanio! It looks like the tests in ServerMode are currently failing, see for example: https://github.com/getmoto/moto/actions/runs/14364128261/job/40273952245?pr=8759

See for the documentation on how to run the tests locally in ServerMode, hopefully that gives some insight into why it's not working:
https://docs.getmoto.org/en/latest/docs/contributing/development_tips/tests.html#servermode-tests

Let me know if you get stuck - happy to have a look as well!

moto/servicecatalogappregistry/responses.py
Peeanio: Good point - I think did this out of order and came back to this. Addressed in latest commits

Peeanio:
Thanks for your comments. I took a look at the docs, but could not see how to address the 404's that appear in server mode. My intuition is that there's probably some mismatch in the backend index matching "servicecatalog-appregistry" vs "servicecatalogappregistry" not hitting the defined URLs, but I can't find it even with some changes. Any advice there would be appreciated.

bblommers:
@Peeanio These are the changes necessary to make the build pass:

https://github.com/Peeanio/moto/compare/master...bblommers:servicecatalog-appregistry-fix-urls

Explanation, in order:

1. CodeDeploy was presumably removed by accident from your branch, so I'm just adding that back in
2. MotoServer has no way of differentiating AppRegistry requests from regular ServiceCatalog-requests. Incoming requests only say 'servicecatalog` in the headers, so MotoServer has no choice but to send them to the ServiceCatalog-service. Adding the AppRegistry-URL's to the ServiceCatalog-service is a trick to make sure we still invoke the correct method
3. The regex change is necessary to support ARN's, as they may contain a slash

Peeanio:
Changes looked good, thanks for the explanation.

bblommers (approved):
LGTM - thank you for adding this service to Moto @Peeanio!

PR #8763, opened 2025-04-04
[S3] Add Account to Replication

zkarpinski:
Add account ID to bucket replication when provided, to support mocking cross accounts.
https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3/client/put_bucket_replication.html


bblommers (approved):
LGTM!

PR #8764, opened 2025-04-04
[Route53] Resolver Queries

Singha22:
Resolves https://github.com/getmoto/moto/issues/8765


bblommers (approved):
LGTM - thanks for adding these features @Singha22!

PR #8774, opened 2025-04-07
Admin: enable verified RDS tests to run against AWS

bpandola:
@bblommers Not sure if there's a process that needs to be followed before enabling, so I've requested your review.

There are only three `aws_verified` RDS tests.  Two in `tests/test_rds/test_integrations.py` (which I recently committed) and one that you wrote some time ago (that required a minor change in this PR because a subsequent commit from someone else made `client` a fixture [with mocking enabled]).

I did a one-off run for the tests I wrote, and they do pass:
image
Link to run: https://github.com/getmoto/moto/actions/runs/14187655393/job/39745796366

### DIFF
--- a/.github/workflows/tests_real_aws.yml
+++ b/.github/workflows/tests_real_aws.yml
@@ -45,4 +45,4 @@ jobs:
       env:
         MOTO_TEST_ALLOW_AWS_REQUEST: ${{ true }}
       run: |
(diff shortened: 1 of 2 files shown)


bblommers (approved):
Good catch @bpandola!

There is no process, as far as I'm concerned - feel free to merge PR's like this whenever. These tests are run against AWS on a weekly basis anyway, so if they ever change/break, we'll notice it soon enough. :slightly_smiling_face:

PR #8786, opened 2025-04-09
[Lex] Lexv2 tagging fix

ApogeeAC:
Adding list_tags_for_resource call to get the tags. 

Original `tags = format_tags(resource.tags)` error "resource" object does not have a "tags" attribute.

Added unit test function.


bblommers (approved):
Great - thanks for the fix @ApogeeAC!

PR #8794, opened 2025-04-11
Glue: get_job_runs() endpoint

BlizzardOf:
Not too much to it. Just adding a pretty basic backend object and the piping to allow for getJobRuns functionality.


bblommers (approved):
LGTM!

PR #8795, opened 2025-04-11
feat(acm): Update status to EXPIRED when Not After has past

HarshCasper:
Fixes https://github.com/localstack/localstack/issues/11152

### DIFF
--- a/tests/test_acm/test_acm.py
+++ b/tests/test_acm/test_acm.py
@@ -1,3 +1,4 @@
+import datetime
 import os
 import uuid
 from ipaddress import IPv4Address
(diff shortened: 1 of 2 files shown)


tests/test_acm/test_acm.py
```diff
@@ -860,3 +861,24 @@ def test_elbv2_acm_in_use_by():
...
+    client = boto3.client("acm", region_name="eu-central-1")
+    arn = _import_cert(client)
+
+    resp = client.describe_certificate(CertificateArn=arn)
+    assert resp["Certificate"]["Status"] == "ISSUED"
+
+    expiry_time = resp["Certificate"]["NotAfter"]
+
+    # Test before expiry (1 second before)
+    with freeze_time(expiry_time - datetime.timedelta(seconds=1)):
```
bblommers, line 877: Freezing time is not possible in ServerMode, unfortunately, that is (presumably) why the tests are failing. 

We tend to just skip the tests if that's the case, see other ACM tests for an example:

https://github.com/getmoto/moto/blob/701c6e5f77b93cc3772f5a892500c62084d6705d/tests/test_acm/test_acm.py#L478:L479

bblommers (changes requested):

tests/test_acm/test_acm.py
HarshCasper: Thanks @bblommers, the issue is now fixed!

bblommers (approved):
Thanks @HarshCasper!

PR #8804, opened 2025-04-14
fix(sagemaker-runtime): remove cache & return 202 for invoke endpoint async

hasan-yaman:
- According to [InvokeEndpointAsync docs](https://docs.aws.amazon.com/sagemaker/latest/APIReference/API_runtime_InvokeEndpointAsync.html), it returns 202 not 200.  Fix it and add tests for it.
- Neither `InvokeEndpointAsync` nor `InvokeEndpoint` docs mention about the cache. I also tested `InvokeEndpointAsync` and see for the same InputLocation it returns different responses.

### DIFF
--- a/moto/sagemakerruntime/models.py
+++ b/moto/sagemakerruntime/models.py
@@ -21,7 +21,7 @@ class SageMakerRuntimeBackend(BaseBackend):
 
         You can use a dedicated API to override this, by configuring a queue of expected results.
 
-        A request to `get_query_results` will take the first result from that queue. Subsequent requests using the same details will return the same result. Other requests using a different QueryExecutionId will take the next result from the queue, or return static data if the queue is empty.
+        A request to `get_query_results` will take the first result from that queue.
 
         Configuring this queue by making an HTTP request to `/moto-api/static/sagemaker/endpoint-results`. An example invocation looks like this:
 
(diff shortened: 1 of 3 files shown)


bblommers (changes requested):
Hi @hasan-yaman! The AWS docs don't mention a cache, no - that is Moto-specific to make sure that users can invoke `invoke_endpoint_async` as often as they want, without having to re-add the same response to the queue. 

This behaviour is mentioned in our docs: https://docs.getmoto.org/en/latest/docs/services/sagemaker-runtime.html

Are you running into specific problems here?

hasan-yaman:
hi @bblommers 

I'm sending the same request to Sagemaker Endpoint multiple times and saving output locations to database to be able to later check those locations. When I try to write tests for that code I realized that moto is returning the same response every time which I think can mislead people. That's why I checked the docs again and tested with production infrastructure to see whether Sagemaker has any kind of cache mechanism.

bblommers:
There are up- and downsides to either approach - and you're right, it is different from how AWS behaves. 

Other Moto users already use the cache though, so I don't want to merge this PR and change this behaviour - that would break everyone else's workflow.

Thank you for opening the PR though @hasan-yaman ! Always nice to see people willing to contribute to open source :slightly_smiling_face:

PR #8813, opened 2025-04-15
EC2: RunInstances must raise if default and request subnet is missing

viren-nadkarni:
On AWS, EC2 [RunInstances](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_RunInstances.html) does not require a subnet ID. By default, it uses the [default subnet](https://docs.aws.amazon.com/vpc/latest/userguide/work-with-default-vpc.html) from the default VPC. However, if the default subnet and VPCs are not available (i.e. if they were manually deleted), the subnet ID is mandatory.

This PR implements this behaviour in Moto.

### DIFF
--- a/moto/ec2/exceptions.py
+++ b/moto/ec2/exceptions.py
@@ -59,6 +59,16 @@ class DependencyViolationError(EC2ClientError):
         super().__init__("DependencyViolation", message)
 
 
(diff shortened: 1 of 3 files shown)


bblommers (approved):
Great, thank you @viren-nadkarni!

PR #8814, opened 2025-04-15
[Timestream] Parameter Groups

Singha22:
Resolves https://github.com/getmoto/moto/issues/8791


moto/timestreaminfluxdb/models.py
```diff
@@ -270,6 +286,255 @@ def untag_resource(self, resource_arn: str, tag_keys: List[str]) -> None:
...
+    ) -> Tuple[str, str, str, str, Dict[str, Any]]:
+        validate_name(name)
+
+        for param_group in getattr(self, "db_parameter_groups", {}).values():
+            if param_group["name"] == name:
+                raise ConflictException(
+                    f"A DB parameter group with the name {name} already exists"
+                )
+
+        if not hasattr(self, "db_parameter_groups"):
```
bblommers, line 304: What's the reason for using `hasattr` here? If we make it a instance-variable from the beginning, by adding `self.db_parameter_groups = {}` to the `__init__`-method instead, we could remove a lot of these checks

moto/timestreaminfluxdb/models.py
```diff
@@ -270,6 +286,255 @@ def untag_resource(self, resource_arn: str, tag_keys: List[str]) -> None:
...
+            self.db_parameter_groups = {}
+
+        param_group_id = random_id()
+
+        arn = f"arn:aws:timestream-influxdb:{self.region_name}:{self.account_id}:db-parameter-group/{param_group_id}"
+
+        param_group = {
```
bblommers, line 311: Can we make this a separate class, `ParameterGroup(BaseModel)`, instead of dealing with a dictionary? That makes it a bit easier to reason over.

moto/timestreaminfluxdb/models.py
```diff
@@ -270,6 +286,255 @@ def untag_resource(self, resource_arn: str, tag_keys: List[str]) -> None:
...
+            "description": description or "",
+            "parameters": parameters or {},
+        }
+
+        self.db_parameter_groups[param_group_id] = param_group
+
+        if tags:
+            self.tag_resource(arn, tags)
+
+        return (
```
bblommers, line 324: Once we have a class, we can also return an instance of the class here. That makes it easier for the caller (in this case the `Response`-class) to understand what it's receiving

moto/timestreaminfluxdb/models.py
```diff
@@ -270,6 +286,255 @@ def untag_resource(self, resource_arn: str, tag_keys: List[str]) -> None:
...
+                    "arn": param_group["arn"],
+                    "description": param_group["description"],
+                }
+            )
+
+        return items
+
+    @paginate(pagination_model=PAGINATION_MODEL)
+    def list_db_clusters(self) -> List[Dict[str, object]]:
+        if not hasattr(self, "db_clusters") or not self.db_clusters:
```
bblommers, line 374: As above

moto/timestreaminfluxdb/models.py
```diff
@@ -270,6 +286,255 @@ def untag_resource(self, resource_arn: str, tag_keys: List[str]) -> None:
...
+            return []
+
+        items = []
+        for cluster_id, cluster in self.db_clusters.items():
+            if cluster.get("status") == "DELETED":
+                continue
+
+            items.append(
```
bblommers, line 382: As above - we could just return a `List[Cluster]` instead, making it a lot easier on the caller to understand what it's receiving

moto/timestreaminfluxdb/models.py
```diff
@@ -270,6 +286,255 @@ def untag_resource(self, resource_arn: str, tag_keys: List[str]) -> None:
...
+                f"DB cluster with ID {db_cluster_id} not found"
+            )
+
+        cluster = self.db_clusters.get(db_cluster_id)
+        if not cluster:
+            raise ResourceNotFoundException(
+                f"DB cluster with ID {db_cluster_id} not found"
+            )
+
+        return (
```
bblommers, line 435: As above

bblommers (changes requested):
Hi @Singha22! Thanks for the PR - please see my comments

bblommers (approved):
LGTM - thanks for addressing the comments @Singha22!

PR #8815, opened 2025-04-15
Batch: Allow using IAM instance profile name when creating compute env

nik-localstack:
According to the [docs](https://docs.aws.amazon.com/cli/latest/reference/batch/create-compute-environment.html)

> instanceRole -> (string)
The Amazon ECS instance profile applied to Amazon EC2 instances in a compute environment. This parameter is required for Amazon EC2 instances types. You can specify the short name or full Amazon Resource Name (ARN) of an instance profile. For example, `` ecsInstanceRole `` or  `arn:aws:iam::<aws_account_id> :instance-profile/ecsInstanceRole ` . For more information, see [Amazon ECS instance role](https://docs.aws.amazon.com/batch/latest/userguide/instance_IAM_role.html) in the Batch User Guide .

### DIFF
--- a/moto/batch/models.py
+++ b/moto/batch/models.py
@@ -1344,7 +1344,10 @@ class BatchBackend(BaseBackend):
                     "Error executing request, Exception : Instance role is required."
                 )
             for profile in self.iam_backend.get_instance_profiles():
-                if profile.arn == cr["instanceRole"]:
+                if (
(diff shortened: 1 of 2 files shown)


bblommers (approved):
Makes sense - thanks for fixing this @nik-localstack!

PR #8818, opened 2025-04-16
EC2: Implement CreateDefaultSubnet operation

viren-nadkarni:
This PR implements the EC2 [CreateDefaultSubnet](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_CreateDefaultSubnet.html) operation.

> `CreateDefaultSubnet` operation creates a default subnet with a size `/20` IPv4 CIDR block in the specified Availability Zone in your default VPC. You can have only one default subnet per Availability Zone.

Moto already supports the EC2 [CreateDefaultVpc](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_CreateDefaultVpc.html) and this is updated such that it also [creates the default subnets](https://docs.aws.amazon.com/vpc/latest/userguide/default-vpc-components.html).

The implementation assigns the `/20` CIDR blocks to AZs in sequence as follows, but that's not always the case on AWS:
- ...a: `172.31.0.0/20`
- ...b: `172.31.16.0/20`
- ...c: `172.31.32.0/20`
- etc.

### DIFF
--- a/IMPLEMENTATION_COVERAGE.md
+++ b/IMPLEMENTATION_COVERAGE.md
@@ -2497,7 +2497,7 @@
 - [ ] create_coip_cidr
 - [ ] create_coip_pool
 - [X] create_customer_gateway
-- [ ] create_default_subnet
+- [X] create_default_subnet
 - [X] create_default_vpc
 - [X] create_dhcp_options
 - [X] create_egress_only_internet_gateway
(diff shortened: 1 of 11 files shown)


bblommers (approved):
LGTM - thanks for improving the parity here @viren-nadkarni!

PR #8822, opened 2025-04-17
add support for Textract SNS integration

HarshCasper:
Relates to https://github.com/localstack/localstack/issues/12372

### DIFF
--- a/moto/textract/models.py
+++ b/moto/textract/models.py
@@ -7,6 +7,7 @@ from moto.core.base_backend import BackendDict, BaseBackend
 from moto.core.common_models import BaseModel
 from moto.moto_api._internal import mock_random
 from moto.sns import sns_backends
+from moto.sns.exceptions import TopicNotFound
 
 from .exceptions import InvalidJobIdException, InvalidParameterException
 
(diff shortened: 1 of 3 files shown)


moto/textract/models.py
```diff
@@ -16,12 +19,50 @@ class TextractJobStatus:
...
+            "JobId": self.job_id,
+            "Status": self.job["JobStatus"],
+            "API": "StartDocumentTextDetection",
+            "JobTag": "",  # Not implemented yet
+            "Timestamp": int(time.time() * 1000),  # Convert to milliseconds
+            "DocumentLocation": doc_location,
+        }
+
+        sns_backend = sns_backends[account_id][region_name]
+        sns_backend.publish(
```
bblommers, line 60: Can we catch any `moto.sns.exceptions.TopicNotFound`-exceptions here? At this point we don't know whether the topic exists, and if it doesn't, we should IMO ignore it for now.

In the future we can add a check whether the Topic exists before starting the Job, but as long as we don't have that, I think we should just ignore invalid input.

bblommers (changes requested):

bblommers (approved):
Thanks @HarshCasper!

PR #8824, opened 2025-04-17
RDS: do not throw an error if multiple modify requests are sent with ManageMasterUserPassword=False

lukaalba:
I've stumbled across this while updating moto in my project. Tests which sent two modify requests with `ManageMasterUserPassword=False` are started failing afterwards.
Even though I don't remember why I wrote those tests this way and make them work again, I think it is more convenient if moto does not fail when multiple modify requests are sent with the same parameter. What do you think?

### DIFF
--- a/moto/rds/models.py
+++ b/moto/rds/models.py
@@ -1454,7 +1454,9 @@ class DBInstance(EventMixin, CloudFormationModel, RDSBaseModel):
             self.master_user_secret = MasterUserSecret(
                 self, master_user_secret_kms_key_id
             )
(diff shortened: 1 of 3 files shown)


bpandola (approved):
@lukaalba Thanks for the fix!

bpandola:
@lukaalba The real AWS backend requires that `MasterUserPassword` be passed in whenever `ManageMasterUserPassword` is set to `False`, but `moto` doesn't validate that currently, so... I'm happy to merge this fix in for now just because it obviously shouldn't fail for the case you've identified.

PR #8826, opened 2025-04-18
[Cloudfront] Add TrustedSigners Class

ApogeeAC:
TrustedSigners class added. This allows users to add an AWSAccountNumber to a CacheBehavior/DefaultCacheBehavior and it will auto update the enabled/qty field. Similar to TrustedKeyGroups functionally.

Added a test to support. Format/Checked, tested local and on server.


bblommers (approved):
Great - thank you @ApogeeAC!

PR #8828, opened 2025-04-19
[S3] GLACIER_IR to storage class

### LINKED ISSUE #8748: S3 GLACIER_IR class should not be in ARCHIVE_STORAGE_CLASSES
Currently GLACIER_IR class is in ARCHIVE_STORAGE_CLASSES list, which doesn't allow to retrieve the file. This should not be the case, it's in the name - 'instant retrieval'.

[https://github.com/getmoto/moto/blob/c194259b6096a865d2c990984740008c26f8b1d9/moto/s3/utils.py#L37](url)

Current error:
`E           botocore.errorfactory.InvalidObjectState: An error occurred (InvalidObjectState) when calling the GetObject operation: The operation is not valid for the object's storage class
`

AWS documentation:
[https://docs.aws.amazon.com/AmazonS3/latest/userguide/storage-class-intro.html](url)

> 

S3 Glacier Instant Retrieval (GLACIER_IR) – Use for long-term data that's rarely accessed and requires milliseconds retrieval. Data in this storage class is available for real-time access.

ApogeeAC:
Fixes #8748 

GLACIER_IR moved to storage class as it supports MS Retrieval

Added EXPRESS_ONEZONE.

Test added. Format/Linted

### DIFF
--- a/moto/s3/utils.py
+++ b/moto/s3/utils.py
@@ -34,7 +34,6 @@ user_settable_fields = {
 ARCHIVE_STORAGE_CLASSES = [
     "GLACIER",
     "DEEP_ARCHIVE",
-    "GLACIER_IR",
 ]
 STORAGE_CLASS = [
     "STANDARD",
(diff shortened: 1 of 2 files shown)


bblommers (approved):
Thanks for the fix @ApogeeAC!

PR #8838, opened 2025-04-21
feat(acm-pca): implement `RevokeCertificate` API

HarshCasper:
See https://github.com/localstack/localstack/issues/10299

### DIFF
--- a/tests/test_acmpca/test_acmpca.py
+++ b/tests/test_acmpca/test_acmpca.py
@@ -12,6 +12,7 @@ from cryptography.hazmat.primitives import hashes, serialization
 from cryptography.x509 import DNSName, NameOID
 
 from moto import mock_aws
+from moto.acmpca.models import acmpca_backends
 from moto.core import DEFAULT_ACCOUNT_ID
 from moto.core.utils import utcnow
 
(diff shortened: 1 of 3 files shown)


tests/test_acmpca/test_acmpca.py
```diff
@@ -697,3 +696,84 @@ def test_policy_operations():
...
+    )
+
+    # Revoke certificate
+    client.revoke_certificate(
+        CertificateAuthorityArn=ca_arn,
+        CertificateSerial=serial_number,
+        RevocationReason="KEY_COMPROMISE",
+    )
+
+    # Verify the certificate is revoked by checking the backend directly
```
bblommers, line 764: This check fails in ServerMode unfortunately, because we can't reach the backend when Moto is running in a separate process. 

The next check (that `get_certificate` fails) should work, so ideally we only guard this section (`if TEST_SERVER_MODE: ...`)

bblommers (changes requested):

tests/test_acmpca/test_acmpca.py
HarshCasper: Seems like that isn't working out on the proxy mode. I have added a skip for this whole test in proxy mode. Let me know if that's not ideal!

bblommers: Makes sense @HarshCasper! Looks like there are just some merge conflicts, once that's fixed I'd be happy to merge

Issue #8839, opened 2025-04-21
S3 policy with StringNotEquals condition results in NotImplementedError


ealb0:
Is this a bug? I am using moto 5.1.4. I am trying to unit test `s3.put_object` with a policy that contains the condition `StringNotEquals`. Using this condition works in real S3, but seems to be undefined in `policy_conditions.py`, making it so that I cannot unit test the policy functionality. 

Here is example code:

```
import json
import boto3
from moto import mock_aws

@mock_aws
def test_s3():
    bucket = "mybucket"
    s3 = boto3.client("s3", region_name="us-east-1")
    s3.create_bucket(Bucket=bucket, CreateBucketConfiguration={'LocationConstraint': 'us-east-1'})

    policy = {
        "Version": "2012-10-17",
        "Id": "PutObjectKMSKeyPolicy",
        "Statement": [
            {
                "Effect": "Deny",
                "Principal": "*",
                "Action": "s3:PutObject",
                "Resource": f"arn:aws:s3:::{bucket}/*",
                "Condition": {
                    "StringNotEquals": {
                        "s3:x-amz-server-side-encryption-aws-kms-key-id": "key_id"
                    }
                }
            }
        ]
    }

    s3.put_bucket_policy(Bucket=bucket, Policy=json.dumps(policy))
    s3.put_object(Bucket=bucket, Key="key", Body="text", ServerSideEncryption="aws:kms", SSEKMSKeyId="kms_id")
```

Running the file with pytest results in the error `NotImplementedError: Unsupported condition: StringNotEquals`. However if I revert to 5.1.3, it works.

@ArteIIo I'm tagging you since you're the only one who worked on `policy_conditions`.

Thanks for the help!

ArteIIo:
Hi @ealb0! Are you using the IAM-like mode for your unit tests? The reason that it is failing is that for the policy evaluation moto is only supporting the `StringEquals` operation. The corresponding source code could be found [here](https://github.com/getmoto/moto/blob/master/moto/iam/policy_conditions.py).

@bblommers for such cases I think that I should add just the logic to return `True` for currently not supported values. If you agree, I can handle the PR creation.

ealb0:
[...]

PR #8842, opened 2025-04-22
[Policies conditions] Fixed compatability issue for currently non-supported policy conditions operations

ArteIIo:
With introduction of policies conditions evaluation in #8680, for currently non-supported ones the explicit exception is throwing. However, according to the investigation for #8839, S3 bucket policies are also evaluating through the same iam `access_control` mechanism and introducing blockers for previously working scenarios.

Main goal of this PR is to provide the backward compatibility check - the condition is not supported, then it is not blocking the further flow like it was before the new logic introduction.


bblommers (approved):
Makes sense @ArteIIo - thanks for following up on this!

PR #8845, opened 2025-04-23
[AppSync] Add tags to list and get

zkarpinski:
Adds the missing tags to app sync responses.

### DIFF
--- a/moto/appsync/models.py
+++ b/moto/appsync/models.py
@@ -194,6 +194,7 @@ class GraphqlAPI(BaseModel):
         open_id_connect_config: str,
         lambda_authorizer_config: str,
         visibility: str,
+        backend,
     ):
         self.region = region
         self.name = name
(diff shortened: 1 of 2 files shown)


moto/appsync/models.py
```diff
@@ -321,13 +323,14 @@ def to_json(self) -> Dict[str, Any]:
...
             "visibility": self.visibility,
+            "tags": self.backend.list_tags_for_resource(self.arn),
         }
 
 
 class AppSyncBackend(BaseBackend):
     """Implementation of AppSync APIs."""
 
-    def __init__(self, region_name: str, account_id: str):
+    def __init__(self, region_name: str, account_id: str) -> None:
```
bblommers, line 333: Python typing tries to be smart here, which makes it confusing IMO. Because the return type of an `__init__` is always `None`, it doesn't need to be specified, **unless** there are no arguments.

So these examples are all valid:
`def __init__(self) -> None` --> No arguments, needs a return
`def __init__(self, arg: str)` --> All arguments typed, no return necessary
`def __init__(self, arg: str) -> None` --> All arguments typed, no return necessary, but also not wrong

Anyway, not a big deal either way - just an FYI :slightly_smiling_face:

bblommers (approved):

PR #8846, opened 2025-04-23
[Config] Resource Config Methods

Singha22:
Resolves https://github.com/getmoto/moto/issues/8807

For the `select_resource_config`, I tried to replicate how Moto Athena's SQL works so as of now it will give a response but it is in no shape or form accurate to how actual SQL would work. 

If there is another example I should follow, let me know.


bblommers (approved):
@Singha22 LGTM! 

The SQL that the user can provide is fairly straightforward, [according to the documentation](https://docs.aws.amazon.com/config/latest/developerguide/query-components.html), so we could write a custom parser for this in Moto to support full query support. We do something similar for S3 Select using [a custom parser](https://pypi.org/project/py-partiql-parser/).

However, if manually configuring the results is good enough for your usecase, then that is a good starting point, and we can always extend it later if we want. Thanks for the PR!

PR #8847, opened 2025-04-24
[Neptune] Enable create_db_instance where engine is neptune

zkarpinski:
Enables create_db_instance for Neptune

### DIFF
--- a/moto/rds/utils.py
+++ b/moto/rds/utils.py
@@ -37,6 +37,7 @@ class DbInstanceEngine(str, Enum):
     CUSTOM_SQLSERVER_WEB = "custom-sqlserver-web"
     MARIADB = "mariadb"
     MYSQL = "mysql"
+    NEPTUNE = "neptune"  # TODO: Only allow from Neptune client?
     ORACLE_EE = "oracle-ee"
     ORACLE_EE_CDB = "oracle-ee-cdb"
     ORACLE_SE2 = "oracle-se2"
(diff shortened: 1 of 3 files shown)


moto/rds/utils.py
```diff
@@ -37,6 +37,7 @@ class DbInstanceEngine(str, Enum):
     CUSTOM_SQLSERVER_WEB = "custom-sqlserver-web"
     MARIADB = "mariadb"
     MYSQL = "mysql"
+    NEPTUNE = "neptune"  # TODO: Only allow from Neptune client?
```
bpandola, line 40: No need.  The backend for RDS and Neptune is one and the same (as it is with real AWS).  Only boto3 client-side validation prevents sending `neptune` as the engine type (when using the `rds` client).

tests/test_neptune/test_clusters.py
```diff
@@ -55,6 +55,27 @@ def test_create_db_cluster__with_additional_params():
...
+        DBInstanceClass="db.r5.large",
+    )
+    resp = client.describe_db_instances(DBInstanceIdentifier="instance-id")[
+        "DBInstances"
+    ]
+    assert resp[0]["DBInstanceIdentifier"] == "instance-id"
+    assert resp[0]["DBInstanceClass"] == "db.r5.large"
+    assert resp[0]["DBClusterIdentifier"] == "cluster-id"
+    assert resp[0]["Engine"] == "neptune"
+
```
bpandola, line 77: Why not assert on the neptune defaults you added?

bpandola (changes requested):

zkarpinski:
@bpandola Made the requested changes

bpandola (approved):

PR #8849, opened 2025-04-25
RDS: make failover target parameter optional

JanKaczmarski:
**Make** `target_db_instance_identifier` **parameter optional in** `failover_db_cluster`

In the previous implementation, target_db_instance_identifier was a required parameter. This change aligns the behavior with the AWS API, where the failover target can be selected automatically if the parameter is omitted.

A unit test has been added to cover this updated behavior.

### DIFF
--- a/moto/rds/models.py
+++ b/moto/rds/models.py
@@ -2669,18 +2669,26 @@ class RDSBackend(BaseBackend):
         return self.create_db_cluster(new_cluster_props)
 
     def failover_db_cluster(
-        self, db_cluster_identifier: str, target_db_instance_identifier: str = ""
+        self,
+        db_cluster_identifier: str,
+        target_db_instance_identifier: str = "",
     ) -> DBCluster:
         cluster = self.clusters[db_cluster_identifier]
 
-        if not target_db_instance_identifier:
-            for instance_id, instance in self.databases.items():
-                if not instance.is_cluster_writer:
-                    target_db_instance_identifier = instance_id
+        if len(cluster.members) == 1:
+            raise InvalidDBClusterStateFault(
+                f"Database cluster:{cluster.db_cluster_identifier} should have at least two database instances for failover"
+            )
 
-        instance = self.databases[target_db_instance_identifier]
-        assert isinstance(instance, DBInstanceClustered)
-        cluster.failover(instance)
+        if not target_db_instance_identifier:
+            for instance in cluster.members:
+                if not instance.is_cluster_writer:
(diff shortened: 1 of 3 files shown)


moto/rds/models.py
```diff
@@ -2669,9 +2669,15 @@ def restore_db_cluster_to_point_in_time(
         return self.create_db_cluster(new_cluster_props)
 
     def failover_db_cluster(
-        self, db_cluster_identifier: str, target_db_instance_identifier: str
+        self, db_cluster_identifier: str, target_db_instance_identifier: str = ""
```
bpandola, line 2672: I think `Optional[str] = None` would make it more explicit that the target is not specified.

moto/rds/models.py
```diff
@@ -2669,9 +2669,15 @@ def restore_db_cluster_to_point_in_time(
...
     ) -> DBCluster:
         cluster = self.clusters[db_cluster_identifier]
+
+        if not target_db_instance_identifier:
+            for instance_id, instance in self.databases.items():
+                if not instance.is_cluster_writer:
+                    target_db_instance_identifier = instance_id
+
```
bpandola, line 2680: Won't this blow up if called on a cluster that only has one member?  Can you find out what the AWS response is in that case and raise an appropriate exception here?

moto/rds/models.py
```diff
@@ -2669,9 +2669,15 @@ def restore_db_cluster_to_point_in_time(
         return self.create_db_cluster(new_cluster_props)
 
     def failover_db_cluster(
-        self, db_cluster_identifier: str, target_db_instance_identifier: str
+        self, db_cluster_identifier: str, target_db_instance_identifier: str = ""
     ) -> DBCluster:
         cluster = self.clusters[db_cluster_identifier]
+
+        if not target_db_instance_identifier:
+            for instance_id, instance in self.databases.items():
```
bpandola, line 2677: You don't want to search through all databases, do you?  Don't you just want `cluster.members`?

bpandola (changes requested):

moto/rds/models.py
JanKaczmarski: I tried that, but `self.databases` expects a str key, and `Optional[str]` causes issues with moto—it doesn't allow non-str keys. So while `Optional[str] = None` is more explicit, it clashes with how the value is later used as a dict key.
image

bpandola (approved):
@JanKaczmarski I pushed some additional commits to handle various exceptional cases and increase test coverage.  Thank you for your contribution and welcome to moto.

PR #8851, opened 2025-04-26
RDS: store managed secret as json with appropriate fields

alex-ruehe:
As far as I can tell, the managed secret for an RDS cluster is a json object dumped as string. Sadly I cannot find any official documentation, only confirmation from AWS Support.

At least https://docs.aws.amazon.com/secretsmanager/latest/userguide/reference_secret_json_structure.html mentions that secrets for RDS/Aurora are supposed to have JSON structure

### DIFF
--- a/moto/rds/models.py
+++ b/moto/rds/models.py
@@ -429,7 +429,7 @@ class MasterUserSecret:
         secret = self.secretsmanager.create_managed_secret(
             service_name="rds",
             secret_id=self._generate_secret_name(),
-            secret_string=json.dumps({"username": "admin", "password": "P@55w0rd!"}),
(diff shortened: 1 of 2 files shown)


bpandola (changes requested):
@alex-ruehe Can you add a basic test case to cover this behavior?  You can put it in the [integrations tests suite](https://github.com/getmoto/moto/blob/8f08d11ed313895a115cc07e8e059dc8432d95ad/tests/test_rds/test_integrations.py) since presumably you'll be using the Secrets Manager API to get the secret (and not the RDS API).

alex-ruehe:
@bpandola updated tests to check the structure of the secret

bpandola (approved):

PR #8857, opened 2025-04-28
ServiceCatalog-AppRegistry Implementation Coverage

Peeanio:
Missed running scripts/implementation_coverage.py for previous commits.


bblommers (approved):
Thanks for the followup @Peeanio! 

FWIW, it's not a big deal if this doesn't happen - I always update the implementation coverage before every release, so it's never out of sync for a long time.

PR #8878, opened 2025-05-06
Add tagging for CodeDeploy

ghorondo:
This PR implements tagging functionality for AWS CodeDeploy resources. It adds support for:

Creating applications and deployment groups with tags
Automatic tag inheritance from deployment groups to deployments
The list_tags_for_resource, tag_resource, and untag_resource API operations

### DIFF
--- a/moto/codedeploy/models.py
+++ b/moto/codedeploy/models.py
@@ -199,7 +199,6 @@ class DeploymentInfo(BaseModel):
         self.external_id = ""
         self.related_deployments: Dict[str, Any] = {}
         self.override_alarm_configuration = override_alarm_configuration
-        # Add tags attribute to deployments
         self.tags: List[Dict[str, str]] = []
 
     def to_dict(self) -> Dict[str, Any]:
@@ -444,7 +443,6 @@ class CodeDeployBackend(BaseBackend):
             self.deployment_groups[application_name] = {}
         self.deployment_groups[application_name][dg.deployment_group_name] = dg
 
-        # Tag the deployment group
         if tags:
             dg_arn = f"arn:aws:codedeploy:{self.region_name}:{self.account_id}:deploymentgroup:{application_name}/{deployment_group_name}"
             self.tagger.tag_resource(dg_arn, tags)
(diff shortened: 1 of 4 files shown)


moto/codedeploy/models.py
```diff
@@ -198,6 +199,7 @@ def __init__(
         self.external_id = ""
         self.related_deployments: Dict[str, Any] = {}
         self.override_alarm_configuration = override_alarm_configuration
+        self.tags: List[Dict[str, str]] = []
```
bblommers, line 202: This doesn't seem to do anything

moto/codedeploy/models.py
```diff
@@ -492,5 +512,31 @@ def list_deployment_groups(
             for deployment_group in self.deployment_groups[application_name].values()
         ]
 
+    def list_tags_for_resource(
+        self, resource_arn: str
+    ) -> Dict[str, List[Dict[str, str]]]:
+        try:
+            return self.tagger.list_tags_for_resource(resource_arn)
+        except Exception:
+            return {"Tags": []}
```
bblommers, line 521:
```suggestion
        return self.tagger.list_tags_for_resource(resource_arn)
```

IMO it would be better to fail loudly. If an error occurs, and an Exception is thrown for whatever reason, at least we know about it, and we can fix it. If we hide errors, it becomes much easier to figure out why Moto is not behaving like you'd expect.

moto/codedeploy/models.py
```diff
@@ -492,5 +512,31 @@ def list_deployment_groups(
...
+
+    def tag_resource(
```
bblommers, line 523: Same

moto/codedeploy/models.py
```diff
@@ -492,5 +512,31 @@ def list_deployment_groups(
...
+        try:
+            self.tagger.tag_resource(resource_arn, tags)
+        except Exception:
+            pass
+
+        return {}
+
+    def untag_resource(self, resource_arn: str, tag_keys: List[str]) -> Dict[str, Any]:
+        try:
+            self.tagger.untag_resource_using_names(resource_arn, tag_keys)
```
bblommers, line 535: Same

bblommers (changes requested):
Hi @ghorondo, thanks for the PR! Please see my comments

moto/codedeploy/models.py
ghorondo: Updated!

bblommers (approved):
LGTM - thank you @ghorondo!

PR #8891, opened 2025-05-11
Update Ruff to 0.11.9 - Drop Pylint

bblommers:
The `ruff` formatter is not yet stable, but it makes sense (IMO) to update to the latest version as much as possible, to reduce the amount of changes we'll have to make when `ruff` 1.0 releases.

The latest version of `ruff` supports a ton more of `pylint`'s rules. The ones that are not covered yet, but that we had enabled:
 - arguments-renamed
 - deprecated-module
 - redefined-outer-name
 - signature-differs

I'm sure that thse rules will be part of `ruff` at some point, but in the mean time, I propose we drop `pylint` altogether. Validation of these 4 rules is nice, but it's not worth the time it takes `pylint` to run (again, IMHO).

@bpandola What do you think - any objections to this?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,8 +21,6 @@ lint:
 	@echo "Running ruff..."
 	ruff check moto tests
 	ruff format --check moto tests
-	@echo "Running pylint..."
-	pylint -j 0 moto tests
 	@echo "Running MyPy..."
 	mypy --install-types --non-interactive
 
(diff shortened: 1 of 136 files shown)


bpandola (approved):
@bblommers I don't have a strong opinion on this one way or the other, happy to defer to your judgment.

PR #8893, opened 2025-05-13
EventBridge: Update describe_event_bus schema

ghorondo:
1. Updated the event bus schema to match boto3 API for describe_event_bus and create_event_bus: https://docs.aws.amazon.com/eventbridge/latest/APIReference/API_DescribeEventBus.html
- Added timestamp fields (CreationTime and LastModifiedTime) to the EventBus class
- Added support for optional fields (DeadLetterConfig, Description, and KmsKeyIdentifier) to match the full AWS API schema

### DIFF
--- a/moto/events/models.py
+++ b/moto/events/models.py
@@ -1573,12 +1573,14 @@ class EventsBackend(BaseBackend):
         return self.event_buses[name]
 
     def list_event_buses(self, name_prefix: Optional[str]) -> List[EventBus]:
-        return [
-            event_bus
-            for event_bus in self.event_buses.values()
-            if event_bus.name != "default"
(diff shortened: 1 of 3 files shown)


bblommers (changes requested):
Hi @ghorondo! What's the reasoning behind the decision to remove the default event bus from the response?

I haven't tested this against AWS, but [the docs](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/events/client/list_event_buses.html) explicitly say that it is included:
> Lists all the event buses in your account, including the default event bus

ghorondo:
@bblommers I added the default event bus back to the response, it was a request from a teammate that I should have looked further into. My apologies!

bblommers (approved):
LGTM - thanks for improving this area @ghorondo!

PR #8894, opened 2025-05-13
ACM-PCA: Add SAN extension passthrough from CSR

viren-nadkarni:
This PR adds Subject Alternative Name (SAN) passthrough from Certificate Signing Request (CSR) to the generated certificate. This allows the user to request [any SAN extension](https://docs.venafi.com/Docs/current/TopNav/Content/Certificates/r-UEP-support-SANs.php): RFC822, Directory, URI etc.

This also fixes a misconfiguration in certificate generation where the DNS SAN extension was added if CN was set. In fact, the correct approach is to [pass-through](https://docs.aws.amazon.com/privateca/latest/userguide/template-definitions.html#EndEntityCertificate-V1) SAN extensions as set in the CSR. This is true for all templates (we don't support the API pass-through i.e. the [`ApiPassthrough`](https://docs.aws.amazon.com/privateca/latest/APIReference/API_IssueCertificate.html#privateca-IssueCertificate-request-ApiPassthrough) parameter)

### DIFF
--- a/moto/acmpca/models.py
+++ b/moto/acmpca/models.py
@@ -1,6 +1,7 @@
 """ACMPCABackend class with methods for supported APIs."""
 
 import base64
+import contextlib
 import datetime
 from typing import Any, Dict, List, Optional, Tuple, cast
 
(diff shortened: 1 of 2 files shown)


bblommers (approved):

PR #8902, opened 2025-05-15
SQS: Return MessageGroupId (and similar attributes) on receive_messages

MatthijsPiek:
In our tests we were missing the MessageGroupId which AWS SQS returns when specified in the `MessageSystemAttributeNames` parameters of `receive_message`. See [boto3 documentation](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/sqs/client/receive_message.html).

We added support for returning the MessageGroupId and several other system attributes and added tests for those as well.


bblommers (approved):
LGTM - thank you for contributing @MatthijsPiek, and welcome to Moto!

PR #8905, opened 2025-05-15
[Connect Campaign] - Config, Job, Campaign

Singha22:
Resolves https://github.com/getmoto/moto/issues/8882


Singha22:
@bblommers Hey Bert, having an issue with linting. I ran the `make format` and it re-formatted other files but now those files are causing this linting issue. Should I create another PR for my changes or is there a way to ignore this linting issue?

bblommers:
Hey @Singha22, we recently updated `ruff` to `0.11.9`, so if you're still on the old version of ruff, that would explain the reformatting. It should go away if you explicitly do a `pip install ruff==0.11.9` and then re-format.

Sorry about that! Could have done a better job of communicating this.

bblommers (approved):
LGTM! Thank you @Singha22

Issue #8906, opened 2025-05-15
[RDS] Add support for DB shard group endpoints


LeekSwan:
Adding support for:
[create_db_shard_group](https://docs.getmoto.org/en/latest/docs/services/rds.html#:~:text=create_db_shard_group)
[describe_db_shard_groups](https://docs.getmoto.org/en/latest/docs/services/rds.html#:~:text=describe_db_shard_groups)

PR #8907, opened 2025-05-15
[RDS] Add shard group endpoint support

LeekSwan:
This PR addresses #8906  

Adding support for:
- [create_db_shard_group](https://docs.getmoto.org/en/latest/docs/services/rds.html#:~:text=create_db_shard_group)
- [describe_db_shard_groups](https://docs.getmoto.org/en/latest/docs/services/rds.html#:~:text=describe_db_shard_groups)

Thank you @leslieetubo for helping refactor!


bpandola:
@LeekSwan You still working on this?

LeekSwan:
Yes I am!

tests/test_rds/test_db_shard_group.py
bpandola, line 1: When using `pytest.raises`, adding at least one assertion to make sure you got the error code you were expecting is standard practice.  There are many instances throughout the RDS test suite, if you need an example.

moto/rds/models.py
```diff
@@ -2208,6 +2210,62 @@ def status(self) -> str:
...
+        self.db_shard_group_resource_id = (
+            f"shard-group-{self.db_shard_group_identifier}"
+        )
+        self.endpoint = f"{self.db_cluster_identifier}.{self.region}.rds.amazonaws.com"
+        self.db_shard_group_arn = f"arn:aws:rds:{self.region}:{self.account_id}:shard-group:{self.db_shard_group_identifier}"
+        self.publicly_accessible = publicly_accessible
+        self.tags = tags or []
+
+    def to_dict(self) -> Dict[str, Any]:
+        return {
```
bpandola: I think this can be deleted.  It's never called and shouldn't be necessary given that you appropriately named your class attributes so they can be picked up by the serializer.

bpandola (changes requested):
@LeekSwan Thanks for adding this new functionality.  I have a couple of minor comments, but happy to merge once addressed.

bpandola (approved):

PR #8922, opened 2025-05-23
[EC2] Add modify_instance_metadata_options support

ApogeeAC:
This PR supports the following added capabilities:

- modify_instance_metadata_options

In the models/instances.py a MetadataOptions class was added. Figured it was cleaner to add a class rather a dict in the Instance's init function

Added "DryRun" to the modify_instance_metadata_options in case a user wants to pass it in. Mock support of it doesn't make a lot of sense so I didn't add handling for it. If wanted in the future it can be implemented.


bblommers (approved):
LGTM!

PR #8947, opened 2025-06-03
[NetworkFirewall] Adding of UpdateToken attribute

ApogeeAC:
Added the `update_token` NetworkFirewall's to the `describe_firewall` Response. UpdateToken also used in other client API calls for future usage.


bblommers (approved):
Makes sense!

PR #8949, opened 2025-06-05
Stable cognito user pool client

Chagui-:
Same as #5194 , but for user pool client.

My use case is: 
I have a frontend which uses cognito for login and i want to connect to moto server for testing and developing. Every time I run the compose file, it creates a new user pool client with a different id, and the frontend breaks.


bblommers (approved):
Makes sense @Chagui- - thanks for the PR!

PR #8951, opened 2025-06-07
EC2: Serialize Transit Gateway Route Tables

bblommers:
@bpandola FYI - I finally got around to playing around with your serializer! 

It took some finagling, because the existing models didn't really align with the actual data model, but I think this is an improvement overall.

### DIFF
--- a/moto/ec2/models/core.py
+++ b/moto/ec2/models/core.py
@@ -12,6 +12,10 @@ class TaggedEC2Resource(BaseModel):
             tags = self.ec2_backend.describe_tags(filters={"resource-id": [self.id]})  # type: ignore[attr-defined]
         return tags
 
+    @property
+    def tagSet(self) -> list[dict[str, str]]:
+        return self.get_tags()
+
     def add_tag(self, key: str, value: str) -> None:
         self.ec2_backend.create_tags([self.id], {key: value})  # type: ignore[attr-defined]
 
--- a/moto/ec2/models/transit_gateway_route_tables.py
+++ b/moto/ec2/models/transit_gateway_route_tables.py
@@ -1,10 +1,30 @@
+import itertools
+from datetime import datetime
 from typing import Any, Dict, List, Optional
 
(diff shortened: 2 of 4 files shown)


moto/ec2/models/transit_gateway_route_tables.py
```diff
@@ -38,6 +57,10 @@ def physical_resource_id(self) -> str:
     def create_time(self) -> str:
         return iso_8601_datetime_with_milliseconds(self._created_at)
```
bpandola: String formatting of dates is a serialization concern; models can (and should) now just deal with basic Python `datetime` objects.

moto/ec2/models/transit_gateway_route_tables.py
```diff
@@ -38,6 +57,10 @@ def physical_resource_id(self) -> str:
...
 
+    @property
+    def creationTime(self) -> str:
```
bpandola: The serializer will find ["xformed"](https://github.com/boto/botocore/blob/45f539f2d5a0cfe5d6a7222f403141d334ad4eaa/botocore/__init__.py#L131) attribute names, so you can keep things Pythonic (i.e. `creation_time` in this case).

moto/ec2/models/core.py
```diff
@@ -12,6 +12,10 @@ def get_tags(self) -> List[Dict[str, str]]:
             tags = self.ec2_backend.describe_tags(filters={"resource-id": [self.id]})  # type: ignore[attr-defined]
         return tags
 
+    @property
+    def tagSet(self) -> list[dict[str, str]]:
```
bpandola, line 16: Here, too, you should be able to use the more Pythonic `tag_set`.

bblommers: This works for the other property (creation_time), but not here for some reason - tags aren't serialized when changing it to `tag_set`. :confused: I'll merge it as-is for now.

bpandola: Hmm... ok.  I will put this on my list to fix as part of my next round of serializer updates (coming soon).

PR #8953, opened 2025-06-08
Fix #4600: Implement KMS key policy enforcement

### LINKED ISSUE #4600: mock_kms doesn't respect the kms policy (AccessDeniedException)
Hello, 

I am trying to write a unit test, to validate insufficient permissions of the KMS key policy. To do this I wrote a python unit test
where I mock the creation of a KMS key and the attachment of a KMS policy with Deny all actions to the key.
However, when I run `describe_key()` instead of expected `AccessDeniedException` ClientError exception, I retrieve the key information. It looks like the KMS policy is being ignored.

How to reproduce the issue:
Run `pytest test.py` against the following code:

```python test.py
from moto import mock_kms
import pytest
import boto3
import json

@mock_kms
def test_kms():
    client = boto3.client("kms")
    new_key = client.create_key(
        Description="test-key",
        Origin="AWS_KMS"
    )
    key_id = new_key["KeyMetadata"]["KeyId"]
    policy = {
        "Version": "2012-10-17",
        "Id": "",
        "Statement": [
            {
                "Sid": "test",
                "Effect": "Deny",
                "Principal": "*",
                "Action": [
                    "kms:*"
                ],
                "Resource": "*"
            }
        ]
    }
    client.put_key_policy(
        Policy=json.dumps(policy),
        PolicyName="default",
        KeyId=key_id
    )
        client.describe_key(KeyId=key_id)

    assert "AccessDeniedException" in str(err.value)

```

- Tell us **what you expected to happen**
   The unit test to PASS as it should ` pytest.raises` as ClientError with `AccessDeniedException` message

- Tell us **what actually happens**.
   Received: `Failed: DID NOT RAISE <class 'botocore.exceptions.ClientError'>`

   The API actually allows the describe_key() operation and it output the key details, instead of an exception

 - Tell us **what version of Moto you're using**
```
    moto==2.2.13
    boto==2.49.0
    botocore==1.20.73
    boto3==1.17.73

    Moto is installed in my local MacBook using pip install moto
    Python 3.9.7
    pip 21.3
    pytest==6.0.0
```

mdawoud27:
- Add Key.evaluate_key_policy method to check permissions
- Update KMS operations to respect key policies
- Add wildcard pattern support for policy actions
- Fix edge cases in re_encrypt method
- Add comprehensive tests for policy enforcement


bblommers:
Please don't waste maintainers time with LLM bullshit.

PR #8956, opened 2025-06-08
feat: Add support for AWSTraceHeader message system attribute in SQS

1Utkarsh1:
This change enhances the SQS mock to support the `AWSTraceHeader` message system attribute.

Modifications include:
- Ensuring `Message` objects in `moto/sqs/models.py` correctly store system attributes.
- Updating `SQSBackend.send_message` to pass through system attributes.
- Modifying `SQSResponse.receive_message` in `moto/sqs/responses.py` to include `AWSTraceHeader` in the response when it's present on the message and requested by you.
- Adding a new test case in `tests/test_sqs/test_sqs.py` to verify the functionality of sending and receiving messages with the `AWSTraceHeader` system attribute, covering scenarios where the header is present, absent, requested, or not requested.

### DIFF
--- a/moto/sqs/models.py
+++ b/moto/sqs/models.py
@@ -76,11 +76,14 @@ class Message(BaseModel):
         self,
         message_id: str,
         body: str,
+        # system_attributes is already here, no change needed to the signature.
+        # The previous attempt was trying to add it again.
         system_attributes: Optional[Dict[str, Any]] = None,
     ):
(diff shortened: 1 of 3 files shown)


bpandola:
The issue with the `AWSTraceHeader` should be resolved with #9386.  If you want to update your PR to just include the additional tests and re-submit, feel free.

PR #8958, opened 2025-06-09
feat: Add tests for SQS AWSTraceHeader attribute

1Utkarsh1:
This commit introduces pytest tests for the AWSTraceHeader message system attribute in SQS.

The tests verify that:
1. When a message is sent with the AWSTraceHeader system attribute, and `receive_message` is called with `AWSTraceHeader` in `MessageSystemAttributes`, the attribute is correctly returned with its original value.
2. When `AWSTraceHeader` is not included in the `MessageSystemAttributes` parameter of `receive_message` (i.e., the parameter is omitted, is an empty list, or contains other attribute names), the `AWSTraceHeader` is not present in the response.

The tests use boto3's SQS client and moto's `mock_sqs` decorator, and ensure proper SQS queue cleanup after each test.

### DIFF
--- a/tests/test_sqs/test_sqs_trace_header.py
+++ b/tests/test_sqs/test_sqs_trace_header.py
@@ -1,0 +1,115 @@
+import boto3
+import pytest
+from moto import mock_sqs
+from uuid import uuid4
+
+
+@mock_sqs
+def test_receive_trace_header_if_requested():
+    sqs = boto3.client("sqs", region_name="us-east-1")
+    queue_name = f"test-queue-{uuid4()}"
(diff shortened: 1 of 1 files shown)


bblommers (changes requested):
Hi @1Utkarsh1! The `mock_sqs` decorator has been removed since Moto 4.0 - we now use the `mock_aws`-decorator instead.

However - even with the right decorator, these tests fail. Were they supposed to be part of your other PR (#8956)?

bpandola:
The issue with the `AWSTraceHeader` should be resolved with #9386.  If you want to update your PR based on the original feedback and re-submit, feel free.

PR #8959, opened 2025-06-09
Added athena get_query_runtime_statistics

footballencarta:
* Added `get_query_runtime_statistics` to Athena
* Ran `make implementation_coverage` to update implementation coverage


bblommers (approved):
LGTM - thanks for adding this feature to Moto @footballencarta!

PR #8970, opened 2025-06-12
IAM: integrate with core AWS response serializer

bpandola:
This PR excises the original Jinja/XML template-based responses in favor of direct integration with our dedicated Query Protocol Serializer:

* all serialization concerns have been removed from the model, e.g. base64 encodings, string formatting of datetime objects, etc.
* very minor test updates were required, mostly just for a couple of methods where we're now (automatically) returning additional response fields

Out of scope for this PR:
* the backend handles tagging inconsistently (i.e. some models use the `moto` TaggingService`, but not all)
* pagination response attributes were maintained if they were part of the original XML templates, but they are (still) applied somewhat haphazardly
* hard-coded values contained in the original XML templates have been moved unchanged to either the model or a response method

@bblommers This was a big one, but the diff is pretty readable.  Given that this is a core AWS service likely used by the majority of `moto` users, do let me know if you have any concerns.  There's always a chance that some behavior not covered by our existing tests could have changed, but I was quite careful and the existing models lined up pretty well with their botocore counterparts.

### DIFF
--- a/moto/iam/models.py
+++ b/moto/iam/models.py
@@ -1,4 +1,5 @@
-import base64
+from __future__ import annotations
+
 import json
 import os
 import re
@@ -9,18 +10,12 @@ from urllib import parse
 
 from cryptography import x509
 from cryptography.hazmat.backends import default_backend
-from jinja2 import Template
 
 from moto.core import DEFAULT_ACCOUNT_ID
 from moto.core.base_backend import BackendDict, BaseBackend
 from moto.core.common_models import BaseModel, CloudFormationModel
 from moto.core.exceptions import RESTError
-from moto.core.utils import (
-    iso_8601_datetime_with_milliseconds,
-    iso_8601_datetime_without_milliseconds,
-    unix_time,
-    utcnow,
-)
+from moto.core.utils import unix_time, utcnow
 from moto.iam.policy_validation import (
     IAMPolicyDocumentValidator,
     IAMTrustPolicyDocumentValidator,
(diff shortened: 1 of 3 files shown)


bblommers:
> the backend handles tagging inconsistently (i.e. some models use the moto TaggingService`, but not all)

Yeah.. I originally introduced the TaggingService to get rid of the inconsistencies, but I just added another way of doing it. :slightly_frowning_face:  

The end goal for me would be to:
 - Have a single TaggingService that every service uses
 - Handle tagging semi-automatically (maybe by making it part of the BaseModel?)
 - Let the ResourceGroupsTaggingAPI just call the TaggingService, instead of explicitly checking every single service.

Now with the serializer in play, standardizing the tagging would simplify that as well. 

Hopefully I'll get to it one day :slightly_smiling_face:

bpandola:
> Hopefully I'll get to it one day 🙂

See, this is the kind of thing I want AI to be able to do.  Like, you write this cool Tagging service and now you want to put it in place across the whole codebase, which isn't exactly difficult--just tedious.  Same with the serializer stuff.  I feel like I should be able to say to an AI, "integrate this service with the serializer and make sure all the tests still pass" - but in my experience thus far something like that does not seem to be possible.  I'm definitely using it as an autocomplete on steroids, which is helpful, but not quite the game changer I'd really like to see.

P.S. I might keep at it with the serializer integrations, as I find it quite addicting to delete so much code.  I have a little fantasy of seeing my name on the `moto` contributors page with more code deleted than added. 😄

PR #8973, opened 2025-06-12
kms - use UTC rather than local time

howbazaar:
When scheduling deletion of a KMS key the a local time value is created but then treated as UTC when converted to a float. So use the `utcnow` utility function instead.


bblommers (approved):
Makes sense - thanks for the fix @howbazaar!

PR #8975, opened 2025-06-13
opensearch: fix engine version response format

### LINKED ISSUE #8974: opensearch: wrong response format in create_domain and describe_domain
Related to https://github.com/getmoto/moto/issues/8271

CreateDomain expects the response key `EngineVersion` to be a string, but since https://github.com/getmoto/moto/pull/8348 was introduced, the value is an object, which causes aws sdk go/terraform to fail with:
`deserialization failed, failed to decode response body with invalid JSON, expected VersionString to be of type string, got map[string]interface {} instead`

dacevedo12:
fixes #8974


dacevedo12:
Pipeline is now green :)

bblommers (approved):
Awesome - thanks for fixing all of these @dacevedo12!

PR #8987, opened 2025-06-15
RAM - Implementing ListResourceTypes

VictorTavares-Dev:
Greetings.

Recently, I'm working with the RAM methods and I noticed that we don't have some API methods for this service implemented here. So, I commited some changes in the lib:
- I'm adding the AWS Glue `database`, `table` and `catalog` shared resource types to enable these resources for the currently available API implementations;
- I'm implementing the `list_resource_types()` API, just to pavimentate the next implementations I'm willing to code.

I've been thinking about implementing the permissions methods in another PR, such as:
- `create_permission()`;
- `get_permission()`;
- `list_permissions()`;
- `list_permission_associations()`;
- `list_resource_share_permissions()`

I'm new here, so, feel free to recomend any design pattern for the implementations I added and to guide me with any recomendations for the code and the contributions :)

### DIFF
--- a/moto/ram/responses.py
+++ b/moto/ram/responses.py
@@ -41,6 +41,6 @@ class ResourceAccessManagerResponse(BaseResponse):
         return json.dumps(
             self.ram_backend.get_resource_share_associations(**self.request_params)
         )
-    
+
     def list_resource_types(self) -> str:
         return json.dumps(self.ram_backend.list_resource_types(**self.request_params))
--- a/moto/ram/responses.py
+++ b/moto/ram/responses.py
@@ -41,3 +41,6 @@ class ResourceAccessManagerResponse(BaseResponse):
         return json.dumps(
             self.ram_backend.get_resource_share_associations(**self.request_params)
         )
+    
+    def list_resource_types(self) -> str:
+        return json.dumps(self.ram_backend.list_resource_types(**self.request_params))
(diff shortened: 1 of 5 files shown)


tests/test_ram/test_ram.py
```diff
@@ -530,3 +531,52 @@ def test_get_resource_share_associations_errors():
         "You cannot specify a principal when the association type is RESOURCE"
         in ex.response["Error"]["Message"]
     )
+
+
+@mock_aws
+@pytest.mark.parametrize(
```
bblommers, line 537: ** praise**: Nice! Parametrizing tests is always a good way to avoid duplication

moto/ram/responses.py
```diff
@@ -41,3 +41,6 @@ def get_resource_share_associations(self) -> str:
         return json.dumps(
             self.ram_backend.get_resource_share_associations(**self.request_params)
         )
+
+    def list_resource_types(self) -> str:
+        return json.dumps(self.ram_backend.list_resource_types(**self.request_params))
```
bblommers, line 46: **nitpick** As a general rule, I like to list the input parameters explicitly, instead of using `**kwargs`.

```suggestion
        resource_region_scope = self._get_param("resourceRegionScope", "ALL")
        return json.dumps(self.ram_backend.list_resource_types(resource_region_scope))
```
And then the same in `models.py`:
```python
def list_resource_types(self, resource_region_scope: str) -> Dict[str, Any]:
```
Behaviour is the same, but this is just a bit easier to read.

I know that the other methods in this particular service always use `**self.request_params`, so this is not a blocker IMO. But if you're intending to add more methods to this service, that would be a good guideline to follow. :slightly_smiling_face:

bblommers (approved):
This looks great @VictorTavares-Dev - thank you for contributing, and welcome to Moto!

PR #8989, opened 2025-06-16
Implement ECS `delete_task_definition` and `status` filter for `list_task_definitions`

lasuillard:
While I'm working on my personal project, I found there are some features missing in ECS that I need. So I implemented:

- I found that current `list_task_definitions` does not support `status` filter. I implement it added tests for it.
-The `delete_task_definitions` feature is not implemented, so I implemented it.

I have some questions:

- Because I thought `test_ecs_boto3.py` file quite big, I added tests for `delete_task_definitions` in new file, `test_ecs_task_definitions.py`. If I should merge the file, please let me know.

Thank you for maintaining this library. Please feel free to suggest any changes!

### DIFF
--- a/moto/ecs/models.py
+++ b/moto/ecs/models.py
@@ -1321,14 +1321,19 @@ class EC2ContainerServiceBackend(BaseBackend):
             ):
                 raise TaskDefinitionMemoryError(cd["name"])
 
-    def list_task_definitions(self, family_prefix: str) -> List[str]:
+    def list_task_definitions(
+        self, family_prefix: str, status: str = "ACTIVE"
+    ) -> List[str]:
         task_arns = []
         for task_definition_list in self.task_definitions.values():
             task_arns.extend(
                 [
                     task_definition.arn
                     for task_definition in task_definition_list.values()
-                    if family_prefix is None or task_definition.family == family_prefix
+                    if (
+                        family_prefix is None or task_definition.family == family_prefix
+                    )
+                    and task_definition.status == status
                 ]
             )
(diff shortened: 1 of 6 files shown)


moto/ecs/models.py
```diff
@@ -2445,5 +2463,66 @@ def enable_long_arn_for_name(self, name: str) -> bool:
...
+
+        deleted_task_definitions = []
+        failures = []
+        for task_def in task_definitions:
+            task_definition_name = task_def.split("/")[-1]
+            if ":" in task_definition_name:
+                family, rev = task_definition_name.split(":")
+                revision = int(rev)
+            else:
+                failures.append(
```
bblommers, line 2487: @lasuillard CoPilot is right about the duplicate tests. 

Did you mean to test this edge case, so with a `task_definition_name` without a `:`?

bblommers (changes requested):

moto/ecs/models.py
lasuillard: @bblommers You are right. That's what I exactly meant to do. Previously I manually checked the response of `aws ecs delete-task-definitions` and found that it gives me different response (`reason`) for each cases, so I wrote two very similar tests, `test_delete_task_definitions_invalid_identifier` and `test_delete_task_definitions_nonexistent`.

But now I see the code and the `reason` is same.. I must have made mistake. Thanks for pointing it out. Let me check it again and fix the code.

lasuillard: I pushed new commit to update the test response.

PR #8991, opened 2025-06-16
elasticache: create and describe subnetgroup

algobarb:
Addresses part of https://github.com/getmoto/moto/issues/8932.

This PR introduces two new functions to the elasticache resource:
1. create_cache_subnet_group
2. describe_cache_subnet_groups
3. list_tags_for_resource

Added tests for:
- when there are actual mock subnets referenced in `test_create_cache_subnet_group_and_mock_vpc()`
- and when the subnetIds referenced need default values

## Note
Regarding the missing codecov tests, I left one of the subnet tests commented out because the current `create_subnet` call hard-codes `ipv6Native` parameter to `False` and does not support ipv6 by itself without ipv4 cidr. Since subnet is used by many other moto resources, I decided I should not try updating the ipv6native support in ec2 subnet in the same PR without thorough testing. Let me know if you would like me to remove the commented out `test_cache_subnet_group_with_ipv6_native_subnets` test altogether.

## References
- https://github.com/getmoto/moto/blob/eba33cf1550a7a40a053ca060d95fdd3acd44e68/moto/rds/models.py#L2000C9-L2003C39 - For how `ec2_backend` was used to `describe_subnets`


moto/elasticache/exceptions.py
```diff
@@ -99,6 +99,28 @@ def __init__(self, cache_cluster_id: str):
...
+class CacheSubnetGroupAlreadyExists(ElastiCacheException):
+    code = 404
+
+    def __init__(self, cache_subnet_group_name: str):
+        (
+            super().__init__(
+                "CacheSubnetGroupAlreadyExists",
+                message=f"CacheSubnetGroup {cache_subnet_group_name} already exists.",
+            ),
+        )
```
bblommers:
```suggestion
        super().__init__(
            "CacheSubnetGroupAlreadyExists",
            message=f"CacheSubnetGroup {cache_subnet_group_name} already exists.",
        )
```
A nitpick, but this broke my brain a bit trying to understand what's going on :slightly_smiling_face:

moto/elasticache/models.py
```diff
@@ -137,13 +141,91 @@ def get_tags(self) -> List[Dict[str, str]]:
...
+        if tags is None:
+            tags = []
+        self.cache_subnet_group_name = cache_subnet_group_name
+        self.cache_subnet_group_description = cache_subnet_group_description
+        self.subnet_ids = subnet_ids
+        self.tags = tags
+
+        # Get VPC details from provided subnet IDs
+        # References rds models.py
+        ec2_backend = ec2_backends[account_id][region_name]
```
bblommers, line 163: Can we move the EC2 import down here? It's quite a computationally expensive import, so we should only do it when absolutely necessary.

(`from moto.ec2.models import ec2_backends`)

moto/elasticache/models.py
```diff
@@ -334,6 +417,45 @@ def delete_cache_cluster(self, cache_cluster_id: str) -> CacheCluster:
...
+    ) -> List[CacheSubnetGroup]:
+        if max_records is None:
+            max_records = 100
+        if cache_subnet_group_name:
+            if cache_subnet_group_name in self.cache_subnet_groups:
+                cache_subnet_group = self.cache_subnet_groups[cache_subnet_group_name]
+                return list([cache_subnet_group])
+            else:
+                raise CacheSubnetGroupNotFound(cache_subnet_group_name)
+        cache_subnet_groups = list(self.cache_subnet_groups.values())[:max_records]
```
bblommers:
```suggestion
    def describe_cache_subnet_groups(
        self,
        cache_subnet_group_name: str,
    ) -> List[CacheSubnetGroup]:
        if cache_subnet_group_name:
            if cache_subnet_group_name in self.cache_subnet_groups:
                cache_subnet_group = self.cache_subnet_groups[cache_subnet_group_name]
                return list([cache_subnet_group])
            else:
                raise CacheSubnetGroupNotFound(cache_subnet_group_name)
        cache_subnet_groups = list(self.cache_subnet_groups.values())
```
Second nitpick. Your approach worked - but all the pagination-details are already handled by the decorator, so it's not necessary to duplicate that logic

moto/elasticache/models.py
```diff
@@ -334,6 +417,45 @@ def delete_cache_cluster(self, cache_cluster_id: str) -> CacheCluster:
...
+            region_name=self.region_name,
+            cache_subnet_group_name=cache_subnet_group_name,
+            cache_subnet_group_description=cache_subnet_group_description,
+            subnet_ids=subnet_ids,
+            tags=tags,
+        )
+        self.cache_subnet_groups[cache_subnet_group_name] = cache_subnet_group
+        return cache_subnet_group
+
+    @paginate(PAGINATION_MODEL)
```
bblommers, line 442: **praise** Nice to see pagination builtin from the beginning!

tests/test_elasticache/test_elasticache.py
```diff
@@ -689,3 +689,283 @@ def test_list_tags_cache_cluster():
...
+        ]
+    assert len(resp["CacheSubnetGroup"]["SupportedNetworkTypes"]) == 2
+    assert "ipv4" in resp["CacheSubnetGroup"]["SupportedNetworkTypes"]
+    assert "dual_stack" in resp["CacheSubnetGroup"]["SupportedNetworkTypes"]
+
+
+# The following test is commented out until the create_subnet moto feature
+# adds support for Ipv6Native parameter.
+
+# @mock_aws
```
bblommers: It's good to already have a test in place! That will make it easier to implement this feature when we're ready.

Instead of commenting it out, I would suggest skipping it instead. That makes it slightly easier to read:
```python
from unittest import SkipTest
def test_cache_subnet_group_with_ipv6_native_subnets():
    raise SkipTest("create_subnet() does not support Ipv6Native-parameter yet")
    .... rest of the test
```

bblommers (changes requested):
Hi @algobarb, thanks for the PR! Looks good overall, just a few minor comments

moto/elasticache/exceptions.py
algobarb: Ah, I copied the existing CacheClusterAlreadyExists exception and it did it that way. Will remove it from there too.

algobarb: Updated and removed those extra parentheses.

moto/elasticache/models.py
algobarb: sure! Good to know for other subnet-related features. That makes sense.

moto/elasticache/models.py
algobarb: I see! Thanks.

tests/test_elasticache/test_elasticache.py
algobarb: That's a good idea! Skipped test

algobarb:
Thank you for taking the time to review the PR. I have made the requested changes. @bblommers

bblommers (approved):
LGTM - thanks again for contributing to Moto @algobarb!

PR #8993, opened 2025-06-17
ServiceDiscovery: create hosted zone when created public or private dns namespaces

DBS-ST-VIT:


moto/servicediscovery/models.py
```diff
@@ -275,7 +276,17 @@ def create_private_dns_namespace(
             if namespace.vpc == vpc:
                 raise ConflictingDomainExists(vpc)
         dns_properties = (properties or {}).get("DnsProperties", {})
-        dns_properties["HostedZoneId"] = "hzi"
+
+        # create the hosted zone
+        hosted_zone = route53_backends[self.account_id][
```
bblommers, line 281: This is the right way to access other services! The only improvement I can see is that this can be extracted, and re-used:


```python
@property
def route53_backend():
    return route53_backends[self.account_id][self.partition]

def create_private_dns_namespace(..):
    ...
    hosted_zone = self.route53_backend.create_hosted_zone(..)

def create_public_dns_namespace(..):
    ...
    hosted_zone = self.route53_backend.create_hosted_zone(..)
```

But that's a nitpick, and not blocking a merge.

PR #8994, opened 2025-06-18
RDS: Add rds bluegreen deployment api implementations

lukaalba:
Hi there,

I was not quite sure how you guys handle new implementations of AWS Features, so it started to implement the apis for rds blue green deployments by myself. This PR contains the following changes:

create_blue_green_deployment:
- adds the BlueGreenDeployment resource to the backend
- adds a green rds instance or cluster to the backend

switchover_blue_green_deployment:
- modifies the source rds instance or cluster and renames it with the "old"-suffix
- modifies the target rds instance or cluster and renames it to the source name

delete_blue_green_deployment:
- deletes the BlueGreenDeployment
- if specified, deletes the green rds instance or cluster

describe_blue_green_deployments:
- get a filterable list of existing BlueGreenDeployment resources

minor changes:
- adds storage_throughput property to DBInstance to be able to update this property with BlueGreenDeployments

Feedback would be highly appreciated!

### DIFF
--- a/moto/rds/models.py
+++ b/moto/rds/models.py
@@ -8,6 +8,7 @@ import re
 import string
 import uuid
 from collections import OrderedDict, defaultdict
+from dataclasses import dataclass
 from datetime import datetime
 from enum import Enum
 from functools import lru_cache, partialmethod
@@ -57,6 +58,7 @@ from .exceptions import (
     DBSubnetGroupNotFoundError,
     ExportTaskAlreadyExistsError,
     ExportTaskNotFoundError,
+    InvalidBlueGreenDeploymentStateFault,
     InvalidDBClusterSnapshotStateFault,
     InvalidDBClusterStateFault,
     InvalidDBClusterStateFaultError,
@@ -3258,6 +3260,16 @@ class RDSBackend(BaseBackend):
                 cluster.master_user_secret.delete_secret()
                 del cluster.master_user_secret
 
+        if "new_db_cluster_identifier" in kwargs:
+            kwargs["endpoint"] = cluster.endpoint.replace(
+                cluster.db_cluster_identifier,
+                kwargs.get("new_db_cluster_identifier", ""),
+            )
+            kwargs["reader_endpoint"] = cluster.reader_endpoint.replace(
+                cluster.db_cluster_identifier,
+                kwargs.get("new_db_cluster_identifier", ""),
+            )
+
         kwargs["db_cluster_identifier"] = kwargs.get("new_db_cluster_identifier", None)
         for k, v in kwargs.items():
             if k == "db_cluster_parameter_group_name":
@@ -3981,12 +3993,15 @@ class RDSBackend(BaseBackend):
         bg_kwargs: Dict[str, Any],
     ) -> BlueGreenDeployment:
         bg_name = bg_kwargs.get("blue_green_deployment_name", "")
-        if bg_name in self.bluegreen_deployments:
-            raise BlueGreenDeploymentAlreadyExistsFault(
-                bg_name
-            )  # ToDo: Check the actual thrown exception
+        for bg_deployment in self.bluegreen_deployments.values():
+            if bg_deployment.blue_green_deployment_name == bg_name:
+                raise BlueGreenDeploymentAlreadyExistsFault(
+                    bg_name
+                )  # ToDo: Check the actual thrown exception
         bg_deployment = BlueGreenDeployment(self, **bg_kwargs)
-        self.bluegreen_deployments[bg_name] = bg_deployment  # type: ignore
+        self.bluegreen_deployments[bg_deployment.blue_green_deployment_identifier] = (
+            bg_deployment  # type: ignore
+        )
         return bg_deployment  # type: ignore
 
     def describe_blue_green_deployments(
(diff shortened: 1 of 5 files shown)


moto/rds/models.py
```diff
@@ -3945,6 +3979,108 @@ def describe_db_log_files(self, db_instance_identifier: str) -> List[DBLogFile]:
...
+        self,
+        bg_kwargs: Dict[str, Any],
+    ) -> BlueGreenDeployment:
+        bg_name = bg_kwargs.get("blue_green_deployment_name", "")
+
+        for bg_deployment in self.bluegreen_deployments.values():
+            if bg_deployment.blue_green_deployment_name == bg_name:
+                raise BlueGreenDeploymentAlreadyExistsFault(bg_name)
+        bg_deployment = BlueGreenDeployment(self, **bg_kwargs)
+        self.bluegreen_deployments[bg_deployment.blue_green_deployment_identifier] = (  # type: ignore
```
bblommers: **nitpick** The `type: ignore` statements on here (and the next few lines) don't seem to be necessary

moto/rds/models.py
```diff
@@ -4155,4 +4291,277 @@ def __init__(self, event_type: str, resource: ResourceWithEvents) -> None:
...
 
+class BlueGreenDeployment(RDSBaseModel):
+    class SupportedStates(Enum):
+        PROVISIONING = (1,)
+        AVAILABLE = (2,)
+        SWITCHOVER_IN_PROGRESS = (3,)
+        SWITCHOVER_COMPLETED = (4,)
+        INVALID_CONFIGURATION = (5,)
+        SWITCHOVER_FAILED = (6,)
+        DELETING = 7
```
bblommers: **question** What's the reason that some enums are  tuple, and some of them are an integer? Is it worth leaving a comment to explain why this is necessary?

lukaalba: Actually I don't know, seems like that was added by some code completion. I set them all to integers

moto/rds/models.py
lukaalba: I removed the `type: ignore` comments, but now the type when calling BlueGreenDeployment() seems incorrect, because I get the following error:

```
Type "BaseModel" is not assignable to declared type "BlueGreenDeployment"
  "BaseModel" is not assignable to "BlueGreenDeployment"
 ```

I have seen this error with other constructor calls, too.

File: moto/rds/models.py [hunk 3]
```diff
@@ -3229,6 +3253,16 @@ def modify_db_cluster(self, kwargs: Dict[str, Any]) -> DBCluster:
...
+        if "new_db_cluster_identifier" in kwargs:
+            kwargs["endpoint"] = cluster.endpoint.replace(
+                cluster.db_cluster_identifier,
+                kwargs.get("new_db_cluster_identifier", ""),
+            )
+            kwargs["reader_endpoint"] = cluster.reader_endpoint.replace(
+                cluster.db_cluster_identifier,
+                kwargs.get("new_db_cluster_identifier", ""),
+            )
+
```
bpandola: Using `string.replace()` here is brittle.  Converting the `endpoint` and `reader_endpoint` attributes to computed properties on `DBCluster` would make this code unnecessary.

moto/rds/models.py
```diff
@@ -3947,6 +3981,108 @@ def describe_db_log_files(self, db_instance_identifier: str) -> List[DBLogFile]:
...
+        if (
+            delete_target
+            and bg_deployment.status
+            != bg_deployment.SupportedStates.SWITCHOVER_COMPLETED.name
+        ):
+            if self._is_cluster(bg_deployment.target):
+                cluster = self.describe_db_clusters(
+                    db_cluster_identifier=bg_deployment.target
+                )[0]
```
bpandola: Using the `find_cluster` helper here would obviate the need for manual array indexing.

moto/rds/models.py
```diff
@@ -3947,6 +3981,108 @@ def describe_db_log_files(self, db_instance_identifier: str) -> List[DBLogFile]:
...
+                for member in cluster.members:
+                    self.delete_db_instance(
+                        db_instance_identifier=member.db_instance_identifier
+                    )
+                self.delete_db_cluster(cluster_identifier=cluster.db_cluster_identifier)
+            else:
+                instance = self.describe_db_instances(
+                    db_instance_identifier=bg_deployment.target
+                )[0]
```
bpandola: Using `self.find_db_from_id` here would obviate the need for manual array indexing.

(Yes, we should probably standardize our `find_*` helpers, but that is out of scope for this PR.)

moto/rds/models.py
```diff
@@ -2225,6 +2245,9 @@ def __init__(self, region_name: str, account_id: str):
             GlobalCluster: self.global_clusters,
             OptionGroup: self.option_groups,
         }
+        self.bluegreen_deployments: MutableMapping[str, BlueGreenDeployment] = (
```
bpandola: minor nit: can we call this `blue_green_deployments` to keep in line with AWS naming?

moto/rds/models.py
```diff
@@ -2309,8 +2332,9 @@ def get_snapshot(
         self,
         identifier: str,
         resource_type: type[DBSnapshot] | type[DBClusterSnapshot],
-        not_found_exception: type[DBSnapshotNotFoundFault]
-        | type[DBClusterSnapshotNotFoundError],
+        not_found_exception: (
+            type[DBSnapshotNotFoundFault] | type[DBClusterSnapshotNotFoundError]
+        ),
```
bpandola: This seems unrelated to your changeset.  If reverted, it would make this PR purely additive (all green) which would be ideal!

moto/rds/models.py
```diff
@@ -4157,4 +4293,277 @@ def __init__(self, event_type: str, resource: ResourceWithEvents) -> None:
...
+            return green_instance.db_cluster_arn
+
+    def _get_instance(self, arn: str) -> DBInstance | DBCluster:
+        result = re.findall(r"^arn:[A-Za-z][0-9A-Za-z-:._]*", arn)
+        if len(result) == 0:
+            raise InvalidParameterValue("Provided string is not a valid arn")
+        if self.backend._is_cluster(arn):
+            return self.backend.describe_db_clusters(db_cluster_identifier=arn)[0]
+        else:
+            return self.backend.describe_db_instances(db_instance_identifier=arn)[0]
```
bpandola: Again, you can use the `find_*` helpers to avoid manual array indexing.

moto/rds/models.py
bblommers: Hm, interesting. I assumed it wasn't necessary because `mypy` doesn't complain about it. Are you using a different type checker, or is your IDE complaining?

moto/rds/models.py
lukaalba: Don't know where that change comes from. I've reverted it

moto/rds/models.py
lukaalba: Its just my IDE complaining, so maybe there is a misconfiguration on my side

moto/rds/models.py
lukaalba: Would you mind to review my changes if the properties are implemented as expected?

bpandola (approved):

PR #8997, opened 2025-06-19
IAM: Omit role tags field from response when not set

viren-nadkarni:
This PR fixes a small cosmetic bug which popped up in LocalStack snapshot tests.

Moto IAM CreateRole always returns the `Tags` field.

```
$ aws --endpoint-url=http://localhost:5000 iam create-role \
    --role-name hello0 \
    --assume-role-policy-document '{"Version": "2012-10-17", "Statement": [{"Action": "sts:AssumeRole", "Principal": {"AWS": "123456789012"}, "Effect": "Allow"}]}'
{
    "Role": {
        "Path": "/",
        "RoleName": "hello0",
        "RoleId": "AROARZPUZDIKGYSLBPCKW",
        "Arn": "arn:aws:iam::123456789012:role/hello0",
        "CreateDate": "2025-06-19T12:42:46.393417Z",
        "AssumeRolePolicyDocument": {
            "Version": "2012-10-17",
            "Statement": [
                {
                    "Action": "sts:AssumeRole",
                    "Principal": {
                        "AWS": "123456789012"
                    },
                    "Effect": "Allow"
                }
            ]
        },
        "MaxSessionDuration": 3600,
        "Tags": [],
        "RoleLastUsed": {}
    }
}
```

AWS IAM CreateRole only returns the `Tags` field when specified in the request.

```
# Without tags
$ aws iam create-role \
    --role-name hello1 \
    --assume-role-policy-document '{"Version": "2012-10-17", "Statement": [{"Action": "sts:AssumeRole", "Principal": {"AWS": "123456789012"}, "Effect": "Allow"}]}'                                                                                                    
{                               
    "Role": {                 
        "Path": "/",
        "RoleName": "hello1",
        "RoleId": "AROAZCRR4CRXXXXXXXXXX",
        "Arn": "arn:aws:iam::123456789012:role/hello1",
        "CreateDate": "2025-06-19T12:28:18Z",
        "AssumeRolePolicyDocument": {                                                                    
            "Version": "2012-10-17",                                                                                                                                                                              
            "Statement": [
                {
                    "Action": "sts:AssumeRole",
                    "Principal": {
                        "AWS": "123456789012"
                    },
                    "Effect": "Allow"
                }
            ]
        }
    }
}
```
```
# With tags
$ aws iam create-role \
    --role-name hello2 \
    --assume-role-policy-document '{"Version": "2012-10-17", "Statement": [{"Action": "sts:AssumeRole", "Principal": {"AWS": "123456789012"}, "Effect": "Allow"}]}' --tags Key=colour,Value=red
{
    "Role": {
        "Path": "/",
        "RoleName": "hello2",
        "RoleId": "AROAZCRR4CRXXXXXXXXXX",
        "Arn": "arn:aws:iam::123456789012:role/hello2",
        "CreateDate": "2025-06-19T12:28:42Z",
        "AssumeRolePolicyDocument": {
            "Version": "2012-10-17",
            "Statement": [
                {
                    "Action": "sts:AssumeRole",
                    "Principal": {
                        "AWS": "123456789012"
                    },
                    "Effect": "Allow"
                }
            ]
        },
        "Tags": [
            {
                "Key": "colour",
                "Value": "red"
            }
        ]
    }
}
```

See https://github.com/getmoto/moto/pull/8970 and https://github.com/getmoto/moto/issues/5092

### DIFF
--- a/moto/iam/models.py
+++ b/moto/iam/models.py
@@ -2186,7 +2186,7 @@ class IAMBackend(BaseBackend):
             path=path,
             permissions_boundary=permissions_boundary,
             description=description,
-            tags=[clean_tags[tag] for tag in clean_tags] or None,
+            tags=[clean_tags[tag] for tag in clean_tags],
             max_session_duration=max_session_duration,
             linked_service=linked_service,
         )
--- a/moto/iam/models.py
+++ b/moto/iam/models.py
@@ -2177,6 +2177,10 @@ class IAMBackend(BaseBackend):
             raise EntityAlreadyExists(f"Role with name {role_name} already exists.")
 
         clean_tags = self._tag_verification(tags)
+        tags = None
+        if clean_tags:
+            tags = [clean_tags[tag] for tag in clean_tags]
+
         role = Role(
             account_id=self.account_id,
             partition=self.partition,
@@ -2186,7 +2190,7 @@ class IAMBackend(BaseBackend):
             path=path,
             permissions_boundary=permissions_boundary,
             description=description,
-            tags=[clean_tags[tag] for tag in clean_tags],
+            tags=tags,
             max_session_duration=max_session_duration,
             linked_service=linked_service,
         )
--- a/moto/iam/models.py
+++ b/moto/iam/models.py
@@ -2177,10 +2177,6 @@ class IAMBackend(BaseBackend):
             raise EntityAlreadyExists(f"Role with name {role_name} already exists.")
 
         clean_tags = self._tag_verification(tags)
-        tags = None
-        if clean_tags:
-            tags = [clean_tags[tag] for tag in clean_tags]
-
         role = Role(
             account_id=self.account_id,
             partition=self.partition,
@@ -2190,7 +2186,7 @@ class IAMBackend(BaseBackend):
             path=path,
             permissions_boundary=permissions_boundary,
             description=description,
-            tags=tags,
+            tags=[clean_tags[tag] for tag in clean_tags] or None,
             max_session_duration=max_session_duration,
             linked_service=linked_service,
         )
--- a/moto/iam/responses.py
+++ b/moto/iam/responses.py
@@ -252,6 +252,8 @@ class IamResponse(BaseResponse):
         role_name = self._get_param("RoleName")
         role = self.backend.get_role(role_name)
         role = copy(role)
+        if not role.tags:
+            role.tags = None  # type: ignore[assignment]  # empty tag-list should not be serialized
         setattr(
             role, "AssumeRolePolicyDocument", quote(role.assume_role_policy_document)
         )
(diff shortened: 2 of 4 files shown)


bblommers:
The `tags`-attribute is always a list, either empty or not. When serializing, it uses the standard `xmltodict.unparse`-method to [serialize it](https://github.com/getmoto/moto/blob/888d7b2e638f441c578dd9129c95ede481a27b98/moto/core/serialize.py#L638), which means that for an empty list it will always return `<Tags />`.

My 'solution' is to explicitly set the `tags` to None to prevent it from being serialized, but ideally we come up with a better (more generic) solution.
 - `botocore` does not seem to have any info on whether a field should be serialized when empty, so we cannot use that
 - In an ideal world this behaviour would be centralized, so empty lists (in all services/responses) should never be serialized. I'd be very surprised if AWS has a standard here though, so we can't do that either.

@bpandola Any ideas on how we could handle this?

bpandola:
@bblommers As far as I can tell, the Smithy docs do not explicitly state whether or not empty lists should be serialized (in responses).  There does not appear to be a standard either, as IAM and RDS (both using the "query" protocol) differ in their behavior with respect to returning (or not returning) empty lists.  

I am not in favor of altering the model, i.e. setting `tags` to `None` when the list is empty.  This is a serialization concern and should be handled in the `responses.py` layer.  The other suggestion of copying the model object and manually setting tags to `None` for the response would probably be fine as a one-off, but I think there's a larger issue here.  I'm already working on a PR for IAM to encapsulate the logic for url-encoding policies in responses -- something similar looks like it will be needed for tags.  I see a number of botocore/boto3 issues related to tags not being present in IAM reponses (e.g. https://github.com/boto/boto3/issues/1855), so having the logic in one place will be helpful as we work toward parity with AWS.

I will get something for you to look at soon (since I'm the one who made the breaking change and am most familiar with the new serialization code).  I will incorporate the tests from this PR.  How does that sound?

bblommers:
> I will get something for you to look at soon (since I'm the one who made the breaking change and am most familiar with the new serialization code). I will incorporate the tests from this PR. How does that sound?

Sounds good! Appreciate you looking into this @bpandola :pray:

viren-nadkarni:
Other attributes that seem to be affected:
- `IsTruncated` field of [ListGropusForUser](https://docs.aws.amazon.com/IAM/latest/APIReference/API_ListGroupsForUser.html)
- `Description` field of [CreatePolicy](https://docs.aws.amazon.com/IAM/latest/APIReference/API_CreatePolicy.html)

bpandola:
@viren-nadkarni @bblommers Have a look at #9018 and let me know what you think.  Thanks.

bpandola:
Superseded by #9018

bpandola:
Superseded by #9018

PR #8999, opened 2025-06-21
RAM - Implementing ListPermissions and some refactoring

VictorTavares-Dev:
Hello, there 😊

I'm continuing the saga to implement tests for the RAM API calls. This time, I'm implementing the `list_permissions()` method. Also, I liked the recommendation from @bblommers to list the input parameters explicitly for the responses in `responses.py` and the `models.py` files (as stated [here](https://github.com/getmoto/moto/pull/8987#discussion_r2159588118)), so this generated some refactoring. Here's a summary of the PR:
- **[NEW]**: I'm implementing the `list_permissions()` API, mocking the response with the expected AWS managed permissions in the AWS RAM service with the services. The responses are based on the services configured in the `SHAREABLE_RESOURCES` list in `models.py` file;
- **[REFACTOR]**: since the responses for this new API implementation has a considerable amount of repeated dicts, I decided to refactor the `models.py` file in order to make things cleaner in this file. I don't know if it's the general pattern we use in the repo, but I created a `utils.py` file to centralize utilities and constants for the RAM implementations to be created in the `models.py` file. In order to generate the expected AWS managed permissions, I decided to code the `format_ram_permission()` function, which can be used to create `CUSTOMER_MANAGED` permissions in the future also;
- **[REFACTOR]**: since I created the `utils.py` file to centralize constants, I decided to move the `RESOURCE_TYPES` values to a new constant called `RAM_RESOURCE_TYPES`, which I decided to format dinamically with an auxiliary constant called `_RESOURCE_TYPE_DEFINITIONS`;
- **[REFACTOR]**: I refactored the `responses.py` and `models.py` files to respectively send and receive explicit parameters. I wasn't able to implement the same for the `create_resource_share()` functions. For some reason, when I tested, when this function was called for the second time, some internal error occurred and the test got stuck in an infinite loop internally. So, I decided to leave the `create_resource_share()` function as is;
- **[TESTS]**: new unit tests were added for both the new implementation and the new `utils.py` file.

Feel free to recomend any design pattern for the implementations I added and to guide me with any recomendations for the code and the contributions :)


moto/ram/models.py
```diff
@@ -413,5 +346,49 @@ def list_resource_types(self, **kwargs: Any) -> Dict[str, Any]:
...
+        }
+
+        # Here, resourceType first partition (service) is case sensitive and
+        # last partition (type) is case insensitive
+        if resource_type and not any(
+            str(permission["resourceType"]).split(":")[0] == resource_type.split(":")[0]
+            and str(permission["resourceType"]).split(":")[-1].lower()
+            == resource_type.split(":")[-1].lower()
+            for permission in self.MANAGED_PERMISSIONS
+        ):
```
bblommers: It took me a while to see what's going on in this logic, even with the comment. 

What do you think about this simplification?
 - The `str(..)` should not be necessary, as we know it's already a string
 - We can compare the entire string for make it more readable. If the entire string is equal, then the last partition is also equal

```suggestion
        if resource_type and not any(
            permission["resourceType"].split(":")[0] == resource_type.split(":")[0]
            and permission["resourceType"].lower() == resource_type.lower()
            for permission in self.MANAGED_PERMISSIONS
        ):
```
Note that this is just a suggestion, nothing else! Feel free to push back if you prefer your version.

moto/ram/models.py
```diff
@@ -413,5 +346,49 @@ def list_resource_types(self, **kwargs: Any) -> Dict[str, Any]:
...
+            raise InvalidParameterException(f"Invalid resource type: {resource_type}")
+
+        if resource_type:
+            permissions = [
+                permission
+                for permission in self.MANAGED_PERMISSIONS
+                if str(permission["resourceType"]).split(":")[-1].lower()
+                == resource_type.split(":")[-1].lower()
```
bblommers: Is this split correct here? I would expect that we want to compare the entire `resource_type`, not just the last partition.

It happens to be fine now, but if (for instance) AWS ever introduced `dynamodb:Database`, then we wouldn't be able to tell it apart from `glue:Database`. So comparing the entire string here seems more correct.

bblommers (changes requested):
Thanks for continuing to improve this service @VictorTavares-Dev! Just some nitpicks - everything else looks good

moto/ram/models.py
VictorTavares-Dev: Hey, @bblommers . Thanks for the input! I really liked the change you recommended. It really makes it easier to understand. I committed the changes here [a2a7a36ce](https://github.com/getmoto/moto/pull/8999/commits/a2a7a36ce85dbffc1b39eb55444a11925da8fad3). About the use of `str()`, I just added it here because the linter was having issues with the parts of the code where I'm retrieving the value of `permission["resourceType"]`, returning errors like `Item "bool" of "str | bool" has no attribute "split"`. I believe it's due to the type hint at the return of the `utils.format_ram_permission()` function. Even though it's correct in my opinion, the linter apparently is interpreting the possibility of the `permission["resourceType"]` to be a `bool` instead of a `str`. I tried messing with the type hints, but, I wasn't able to correct this issue unless to explicitly "transform" the `permission["resourceType"]` in a str by using `str()`. If you see any recommendations for this behavior, I'd really appreciate it!

moto/ram/models.py
VictorTavares-Dev: You are absolutely correct! Guess I got carried away with the splits 😂. I fixed this with the expected behavior in the [a2a7a36ce](https://github.com/getmoto/moto/pull/8999/commits/a2a7a36ce85dbffc1b39eb55444a11925da8fad3) commit, and fixed the tests in the [03727c2a8](https://github.com/getmoto/moto/pull/8999/commits/03727c2a8a4a9cfdce07a16b39af206904787141) commit.

moto/ram/models.py
bblommers: Ah, I understand. The 'correct' way would be to create a `TypedDict` and specify that as the return type. The `TypedDict` can specify exactly which fields are present, and which type each field is. With that in place, `mypy` would understand that `permission["resourceType"]` is a string.

As an example what that would look like:
```python
from typing import TypedDict

class PermissionType(TypedDict):
    resourceType: str
    defaultVersion: bool
    ...

def format_ram_permission(..) -> PermissionType:
    return {
        "resourceType": "..",
        "defaultVersion": False,
        ...
    }
```

For now I'm happy to keep it as a string though, to keep it simple.

bblommers (approved):
LGTM - thanks again @VictorTavares-Dev!

PR #9004, opened 2025-06-22
RAM - Some refactoring for the next steps

VictorTavares-Dev:
Hello, again!

In this PR, I'm implementing some refatoring in the `models.py` file. Since I'm willing to implement new logic to interact with the managed permissions specifications, I decided to refactor the models we have and introduce two new model classes, in order to configure the backend for the next RAM implementations. Here's a summary of the changes:
- I created the `ResourceType` class to represent a model of any resource type implemented in the backend of the RAM service;
- I created the `ManagedPermission` class to represent a model of any managed permission implemented in the backend of the RAM service;
- Since I introduced two new classes to represent parts of the service, I added new class attributes for the `ResourceAccessManagerBackend` class. The difference now is that each attribute we had previously now represents a `List` of objects from the classes I mentioned above;
- Since I created the new attributes above, I had to refactor the previously implemented methods in order to work with this new pattern;
- I refactored the `test_ram.py` file, to make use of the `AWS_MANAGED_PERMISSIONS` and `RAM_RESOURCE_TYPES` lists for validation;
- I created the `ManagedPermissionDict` in `utils.py` for better type annotation and removed the unnecessary `str()` calls from `models.py` (thank you @bblommers for the recommendation);
- Lastly, I did some minor changes I noticed by re-reading the `models.py` file (e.g. typo in the `resouce` variable name in `organizations_backend()` method)

I'd like to discuss if this new pattern implementation for the service makes sense. Thanks in advance for the help :)

### DIFF
--- a/moto/ram/models.py
+++ b/moto/ram/models.py
@@ -115,6 +115,7 @@ class ResourceShare(BaseModel):
                     f"Principal ID {principal} is malformed. Verify the ID and try again."
                 )
 
+        for principal in principals:
             self.principals.append(principal)
 
     def add_resources(self, resource_arns: List[str]) -> None:
@@ -132,6 +133,7 @@ class ResourceShare(BaseModel):
                     "You cannot share the selected resource type."
                 )
 
+        for resource in resource_arns:
             self.resource_arns.append(resource)
 
     def delete(self) -> None:
--- a/moto/ram/models.py
+++ b/moto/ram/models.py
@@ -1,12 +1,16 @@
 import re
 import string
+from datetime import datetime, timezone
 from typing import Any, Dict, List, Optional
 
 from moto.core.base_backend import BackendDict, BaseBackend
 from moto.core.common_models import BaseModel
 from moto.core.utils import unix_time, utcnow
 from moto.moto_api._internal import mock_random as random
-from moto.organizations.models import OrganizationsBackend, organizations_backends
+from moto.organizations.models import (
+    OrganizationsBackend,
+    organizations_backends,
+)
 from moto.ram.exceptions import (
     InvalidParameterException,
     MalformedArnException,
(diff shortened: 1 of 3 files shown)


moto/ram/models.py
```diff
@@ -111,7 +115,6 @@ def add_principals(self, principals: List[str]) -> None:
                     f"Principal ID {principal} is malformed. Verify the ID and try again."
                 )
 
-        for principal in principals:
```
bblommers: I think this should stay, actually, despite the performance improvement that it would give, for the simple reason that we only want to add the principals if all of them are valid. 

If a user provides two principals, one valid and the second invalid, with this change the first principal will be added, but the request will fail on the second one. For the user, it would be a bit confusing if some of the principals were successfully added, despite the fact that the request failed.

moto/ram/models.py
```diff
@@ -129,7 +132,6 @@ def add_resources(self, resource_arns: List[str]) -> None:
                     "You cannot share the selected resource type."
                 )
 
-        for resource in resource_arns:
```
bblommers: Same

moto/ram/utils.py
```diff
@@ -1,5 +1,18 @@
 from datetime import datetime, timezone
-from typing import Dict, Optional, Union
+from typing import Optional, TypedDict
+
+
+class ManagedPermissionDict(TypedDict):
```
bblommers, line 5: Nice: This makes it much easier to understand what kind of data is present

bblommers (changes requested):

moto/ram/models.py
VictorTavares-Dev: Oh, I see. I didn't think about this behavior when I refactored this part. It really makes sense. Thanks for the heads up! I fixed it in the [b958b2d6e](https://github.com/getmoto/moto/pull/9004/commits/b958b2d6e6b6c8b7259e0db7a2dab7a287dd7fba) commit

moto/ram/models.py
VictorTavares-Dev: Fixed in the [b958b2d6e](https://github.com/getmoto/moto/pull/9004/commits/b958b2d6e6b6c8b7259e0db7a2dab7a287dd7fba) commit

PR #9009, opened 2025-06-24
fix: SecretsManager - add missing 'tags' property to ReplicaSecret

spennymac:
Fixes an AttributeError raised when accessing the 'tags' attribute on a ReplicaSecret instance. The property was missing, causing code that relies on it to fail with:

  `AttributeError: 'ReplicaSecret' object has no attribute 'tags'`


tests/test_secretsmanager/test_secrets_duplication.py
```diff
@@ -97,6 +97,42 @@ def test_replicate_secret():
     assert replica_secret["ARN"] == replica_arn
 
 
+@mock_aws
+def test_list_replica_secret_with_filters():
```
bblommers, line 101: Very comprehensive test @spennymac - that's great to see.

PR #9010, opened 2025-06-24
set correct value type for Logs.GetQueryResults

pinzon:
## Motivation
According to the [Boto3 operation docs](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/logs/client/get_query_results.html) and the [Go SDK documentation](https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/service/cloudwatchlogs@v1.50.3/types#ResultField), the value field must always be a string.

## Changes
- Ensure the value attribute is always a string

## Testing
- Extended an existing test to assert the type
- Manually verified the behavior against AWS

### DIFF
--- a/moto/logs/models.py
+++ b/moto/logs/models.py
@@ -88,7 +88,7 @@ class LogQuery(BaseModel):
     def to_result_json(self) -> Dict[str, Any]:
         return {
             "results": [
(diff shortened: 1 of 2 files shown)


bblommers (approved):
LGTM! Thanks for the additional parity here @pinzon

PR #9017, opened 2025-06-27
EMR: integrate with core AWS response serializer

bpandola:
This PR excises the original Jinja/XML template-based responses for EMR in favor of direct integration with our dedicated AWS JSON Protocol Serializer.

> [!WARNING]  
> Although it's hard to imagine this impacting anyone (for reasons noted below), this is technically a breaking change. Merging of this Pull Request will drop `moto`'s current support for EMR API requests that use the AWS Query protocol. 

## Some Relevant History
`moto` was originally a mocking package created for use with the AWS [`boto`](https://github.com/boto/boto) library.  The `boto` library communicated with EMR using the AWS Query protocol, which resulted in `moto` using XML string templates for mocking EMR responses.  When [`botocore`](https://github.com/boto/botocore) and [`boto3`](https://github.com/boto/boto3) arrived on the scene, they communicated with EMR using the newer AWS JSON protocol. As `boto3` began to gain acceptance, `moto` was [modified](https://github.com/getmoto/moto/pull/733) to translate the original XML responses (for `boto`) into equivalent JSON responses (for `boto3`), and has supported both protocols for EMR ever since.  In the ensuing years, `boto` was deprecated, is no longer maintained, does not support Python versions beyond 3.4, and support for it has been [removed from `moto`](https://github.com/getmoto/moto/pull/4378).  All this to say, I did not feel compelled to find some way to continue supporting both communication protocols for EMR when integrating it with our core response serializer.  If the original Query protocol-based service definition for EMR exists somewhere--it never existed in `botocore` and I was unable to find it anywhere else--we could continue to support this, but ultimately I don't think it's worth maintaining any longer.

### DIFF
--- a/moto/core/responses.py
+++ b/moto/core/responses.py
@@ -26,7 +26,6 @@ from xml.dom.minidom import parseString as parseXML
 
 import boto3
 import requests
-import xmltodict
 from botocore.model import OperationModel, ServiceModel
 from jinja2 import DictLoader, Environment, Template
 from werkzeug.exceptions import HTTPException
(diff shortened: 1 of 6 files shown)


bblommers (approved):
LGTM! If the `Query`-protocol was only supported in `boto`, it makes sense to drop support for it :+1:

PR #9023, opened 2025-06-29
SES: add SNS event destinations to describe_configuration_set

dmacvicar:
The template only returns the configuration set name.

There are other [ConfigurationSetAttributeNames](https://docs.aws.amazon.com/ses/latest/APIReference/API_DescribeConfigurationSet.html) that should be added in the future, like `trackingOptions | deliveryOptions | reputationOptions`. Same with other types of destinations like `CloudWatchDestination` and `KinesisFirehoseDestination` .

Related to https://github.com/localstack/localstack/issues/12722

### DIFF
--- a/moto/ses/responses.py
+++ b/moto/ses/responses.py
@@ -1,5 +1,5 @@
 import base64
-from typing import Any, Dict, List
+from typing import Any, Dict, List, Optional
 
 from moto.core.responses import BaseResponse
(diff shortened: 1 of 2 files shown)


bblommers (approved):
LGTM - thanks for the fix @dmacvicar, and welcome to Moto! :partying_face:

PR #9041, opened 2025-07-03
remove SNS:Receive from default topic policy

thrau:
We know from snapshot tests we've done in localstack, that `SNS:Receive` is not in the list of actions of the default policy: see the [snapshot file for test_sns.py](https://github.com/localstack/localstack/blob/e558996a92c8635d859d5652ea638f27630ba8a7/tests/aws/services/cloudformation/resources/test_sns.snapshot.json#L28-L56).

This PR simply removes that action from the default policy factory.


thrau:
looks like there's a flaky test in `test_stepfunctions_dynamodb_integration.py`

bpandola (approved):
Thanks @thrau

PR #9053, opened 2025-07-10
ssm: Delete ssm parameter based on resource name

thecodingsysadmin:
Use the `resource_name` to delete the SSM Parameter when using Cloudformation.

Will fallback to using the Cloudformation JSON if `resource_name` is not specified.

The JSON `cloudformation_json` would not match the resource name if Parameter values are being used in the Cloudformation template to represent the SSM ParameterStore name.

In the `cloudformation_json` case below, the SSM parameter that is passed to `ssm_backend.delete_parameter` is the literal value, not the parsed value.

```
{
  "Properties": {
    "Name": {
      "Fn::Sub": "${Name}"
    },
    "Type": "String",
    "Value": {
      "Ref": "Value"
    }
  },
  "Type": "AWS::SSM::Parameter"
}

```

### DIFF
--- a/moto/ssm/models.py
+++ b/moto/ssm/models.py
@@ -359,10 +359,13 @@ class Parameter(CloudFormationModel):
         account_id: str,
         region_name: str,
     ) -> None:
-        ssm_backend = ssm_backends[account_id][region_name]
-        properties = cloudformation_json["Properties"]
+        resource = resource_name
 
-        ssm_backend.delete_parameter(properties.get("Name"))
+        ssm_backend = ssm_backends[account_id][region_name]
+        if not resource:
+            properties = cloudformation_json["Properties"]
+            resource = properties.get("Name")
+        ssm_backend.delete_parameter(resource)
 
     @property
     def physical_resource_id(self) -> str:
--- a/tests/test_ssm/test_ssm_parameterstore_cloudformation.py
+++ b/tests/test_ssm/test_ssm_parameterstore_cloudformation.py
@@ -1,0 +1,148 @@
+import json
+
+import boto3
+import botocore.exceptions
+import pytest
+
+from moto import mock_aws
+
+
+@mock_aws
+def test_cloudformation_lifecycle():
+    ssm_param_name = "test"
+    ssm_param_value = "initial value"
+    stack_template = {
+        "AWSTemplateFormatVersion": "2010-09-09",
+        "Description": "Test Stack",
+        "Resources": {
+            "SSMParameter": {
+                "Type": "AWS::SSM::Parameter",
+                "Properties": {
+                    "Name": ssm_param_name,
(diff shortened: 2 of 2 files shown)


moto/ssm/models.py
```diff
@@ -415,9 +418,9 @@ def __init__(self, ssm_document: "Document"):
         self.versions = {version: ssm_document}
         self.default_version = version
         self.latest_version = version
-        self.permissions: Dict[
-            str, AccountPermission
-        ] = {}  # {AccountID: AccountPermission }
+        self.permissions: Dict[str, AccountPermission] = (
+            {}
+        )  # {AccountID: AccountPermission }
```
thecodingsysadmin: ruff reformat change - no actual changes here.

thecodingsysadmin:
```
pytest -sv tests/test_ssm/test_ssm_parameterstore_cloudformation.py

tests/test_ssm/test_ssm_parameterstore_cloudformation.py::test_cloudformation_lifecycle PASSED
tests/test_ssm/test_ssm_parameterstore_cloudformation.py::test_cloudformation_lifecycle_with_parsing PASSED
```

moto/ssm/models.py
bblommers: Hi @thecodingsysadmin! Looks like the linter is still failing because of `ruff`. Please make sure that you use `ruff==0.11.9` when reformatting, as other versions might produce different results.

thecodingsysadmin: Thanks, I've reformatted using `ruff` and now this change no longer appears.

Apologies for that.

moto/ssm/models.py
```diff
@@ -359,10 +359,13 @@ def delete_from_cloudformation_json(  # type: ignore[misc]
...
     ) -> None:
-        ssm_backend = ssm_backends[account_id][region_name]
-        properties = cloudformation_json["Properties"]
+        resource = resource_name
 
-        ssm_backend.delete_parameter(properties.get("Name"))
+        ssm_backend = ssm_backends[account_id][region_name]
+        if not resource:
+            properties = cloudformation_json["Properties"]
+            resource = properties.get("Name")
```
thecodingsysadmin, line 367: @bblommers looking for some advice on testing - looks like coverage is saying I'm not testing this branch path.

l left this in just in case things change in the future, but from my limited testing `resource_name` is always passed into this method.

Would you rather I delete this if statement and rely solely on `resource_name` or add a `#pragma: no cover ` statement?

I can also add a specific test to this method, but looking at some of the existing tests this project doesn't test methods directly.

bblommers: I'm OK with the current approach @thecodingsysadmin. The coverage is more of a suggestion than a rule, and it is good to keep the if-statement just in case we're missing something - we don't want to break a users' workflow.

PR #9060, opened 2025-07-14
fix(s3): object key in notifications should be url-encoded

### LINKED ISSUE #9054: S3 Bucket notifications - Object keys are not URL-encoded
## Description of the issue

I am trying to test a setup in which a S3 Bucket is configured to send notifications to an SQS queue when objects are created (`s3:ObjectCreated:*`).

As per AWS documentation, the object keys in S3 notifications are URL-encoded. To handle this, my production code uses `urllib.quote.unquote_plus` to retrieve the actual object key.

> The s3 key provides information about the bucket and object involved in the event. The object key name value is URL encoded. For example, "red flower.jpg" becomes "red+flower.jpg" (Amazon S3 returns "application/x-www-form-urlencoded" as the content type in the response).
> -- https://docs.aws.amazon.com/AmazonS3/latest/userguide/notification-content-structure.html

However, moto does not apply the URL-encoding. Therefore, after the URL-decoding, certain object keys (e.g. with plus signs) will end up being the wrong ones.

## Workaround

I use the following pytest fixture to "deactivate" the URL-decoding during the tests. One could use the same `mock.patch` as a unittest class decorator. It goes without saying that I'm not a fan of tampering with production code in order for the tests to pass.

```python
from unittest import mock
import pytest

@pytest.fixture(autouse=True)
def deactivate_url_decoding_s3_object_keys():
    with mock.patch("urllib.quote.unquote_plus", lambda x: x):
        yield
```

<details>
<summary><h2>Minimal reproducible example</h2></summary>

```python
import json

import boto3
import moto


AWS_REGION = "us-east-1"
QUEUE = "my-queue"
BUCKET = "my-bucket"


@moto.mock_aws
def test_s3_sqs_notification():
    # Create SQS queue
    sqs_client = boto3.client("sqs", region_name=AWS_REGION)
    create_queue_response = sqs_client.create_queue(QueueName=QUEUE)

    # Create S3 bucket and notification configuration
    s3_client = boto3.client("s3", region_name=AWS_REGION)
    s3_client.create_bucket(Bucket=BUCKET)
    s3_client.put_bucket_notification_configuration(
        Bucket=BUCKET,
        NotificationConfiguration={
            "QueueConfigurations": [
                {
                    "QueueArn": f"arn:aws:sqs:{AWS_REGION}:123456789012:{QUEUE}",
                    "Events": ["s3:ObjectCreated:*"],
                }
            ]
        },
        SkipDestinationValidation=True,
    )

    # Upload S3 object
    s3_client.put_object(Bucket=BUCKET, Key="filename+with+plus", Body=b"abcdef")
    s3_client.put_object(Bucket=BUCKET, Key="filename with spaces", Body=b"gehijk")

    # Get rid of the test message
    receive_message_response = sqs_client.receive_message(
        QueueUrl=create_queue_response["QueueUrl"]
    )

    # First notification
    receive_message_response = sqs_client.receive_message(
        QueueUrl=create_queue_response["QueueUrl"]
    )
    assert (
        json.loads(receive_message_response["Messages"][0]["Body"])["Records"][0]["s3"][
            "object"
        ]["key"]
        == "filename+with+plus"  # Should be "filename%2Bwith%2Bplus"
    )

    # Second notification
    receive_message_response = sqs_client.receive_message(
        QueueUrl=create_queue_response["QueueUrl"]
    )
    assert (
        json.loads(receive_message_response["Messages"][0]["Body"])["Records"][0]["s3"][
            "object"
        ]["key"]
        == "filename with spaces"  # Should be "filename with spaces"
    )
```
</details>

## Environment

* Python 3.12.11
* Boto3 1.38.27
* Moto 5.1.8

## Solution?

I have identified the following as the piece of code responsible for generating the body of the notification. Intuitively, I should be able to wrap the object key with `urllib.quote.quote_plus` and be done with it.

I am yet to find the tests associated with S3 bucket notifications. I'll try and draft a PR asap and keep this section updated.

VictorColomb:
This PR applies URL-encoding to the object key in the payload of S3 Object Notifications to Lambda/SQS/SNS, as described in [^1].

Fixes #9054.

### Notes

Tests concerning notifications to SNS and SQS are located in a module called `test_s3_lambda_integration`, while the module `test_s3_notification` only contains notifications to EventBridge. It therefore took me a while to find them. Is there a historical reason for this? What would you think about merging all of these tests in the latter module?

I have hijacked existing S3 Object Notification tests to demonstrate the amended behavior. Let me know if you would prefer that I write a new, specific test.

[^1]: https://docs.aws.amazon.com/AmazonS3/latest/userguide/notification-content-structure.html

### DIFF
--- a/moto/s3/notifications.py
+++ b/moto/s3/notifications.py
@@ -3,6 +3,7 @@ import json
 from datetime import datetime
 from enum import Enum
 from typing import TYPE_CHECKING, Any, Dict, List
+from urllib.parse import quote_plus
 
 from moto.core.utils import unix_time
 
@@ -73,6 +74,8 @@ def _get_s3_event(
     # s3:ObjectCreated:Put --> ObjectCreated:Put
     event_name = event_name[3:]
     event_time = datetime.now().strftime(_EVENT_TIME_FORMAT)
+    # https://docs.aws.amazon.com/AmazonS3/latest/userguide/notification-content-structure.html
+    key_name = quote_plus(key.name)
     return {
         "Records": [
             {
@@ -88,7 +91,7 @@ def _get_s3_event(
                         "name": bucket.name,
                         "arn": bucket.arn,
                     },
-                    "object": {"key": key.name, "size": key.size, "eTag": etag},
+                    "object": {"key": key_name, "size": key.size, "eTag": etag},
                 },
             }
         ]


VictorColomb:
Thanks for triggering the tests!

The `tests/test_sqs/test_sqs_cloudformation.py::test_describe_stack_subresources` test has failed in Python 3.13. I cannot reproduce locally with the exact same Python version and pytest options. Furthermore, a test in the same module has failed in the same way [here][1].
Is this a known flaky test? If so, would it be possible to re-run the failed job?


[1]: https://github.com/getmoto/moto/actions/runs/15968877357/job/45035486279

bblommers:
> Tests concerning notifications to SNS and SQS are located in a module called test_s3_lambda_integration, while the module test_s3_notification only contains notifications to EventBridge. It therefore took me a while to find them. Is there a historical reason for this? What would you think about merging all of these tests in the latter module?

Probably historical yes - just a matter of someone not paying attention. Having all the tests in one place (in `test_s3_notification`) would make sense - but it's also not a big deal IMO

> I have hijacked existing S3 Object Notification tests to demonstrate the amended behavior. Let me know if you would prefer that I write a new, specific test.

No, that makes a lot of sense, to use the same test for this behaviour :+1:

bblommers (approved):
LGTM - thanks for fixing this @VictorColomb, and welcome to Moto!

j-mdw:
The escaping is wrongly encoding "/" characters. These are not encoded in S3 event-notifications. Very simple fix here: `quote_plus(key.name, safe='/')`

PR #9066, opened 2025-07-16
Reduce size of the Docker image

Tenzer:
This takes a few steps to reduce the size of the generated Docker image, making it faster to load for developers and CI machines.

The arm64 image was 792 MB before I started this work and the result is a 523 MB image, so a reduction of 269 MB.

I have broken the steps taken into individual commits to make it easier to review. Each commit message mentions the size reduction the change gives.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,7 @@
+# Ignore everything by default
+*
+
+!README.md
+!moto/
+!setup.cfg
+!setup.py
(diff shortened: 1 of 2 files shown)


bblommers:
I've verified these improvements [using Dive](https://github.com/wagoodman/dive) (on AMD64), and the improvements are indeed impressive:

Before: 
Screenshot from 2025-07-19 11-19-05

After:


Screenshot from 2025-07-19 11-18-17

Tenzer:
You are welcome! I used Dive myself to try to identify areas for improvement.

Issue #9070, opened 2025-07-17
[Athena] Add athena capacity reservation endpoint support


LeekSwan:
Adding support for:
- [create_capacity_reservation](https://docs.getmoto.org/en/latest/docs/services/athena.html#:~:text=create_capacity_reservation)
- [get_capacity_reservation](https://docs.getmoto.org/en/latest/docs/services/athena.html#:~:text=get_capacity_reservation)

PR #9071, opened 2025-07-17
[Athena] Add athena capacity reservation endpoint support

LeekSwan:
This PR addresses #9070 

Adding support for:
- [create_capacity_reservation](https://docs.getmoto.org/en/latest/docs/services/athena.html#:~:text=create_capacity_reservation)
- [get_capacity_reservation](https://docs.getmoto.org/en/latest/docs/services/athena.html#:~:text=get_capacity_reservation)


moto/athena/models.py
```diff
@@ -367,6 +379,25 @@ def stop_query_execution(self, exec_id: str) -> None:
...
+            target_dpus=target_dpus,
+            tags=tags
+        )
+        #TODO: Need some validations
+        self.capacity_reservations[cr.name] = cr
+        return None
+        
+
+    def get_capacity_reservation(self, name: str) -> Optional[CapacityReservation]:
+        return self.capacity_reservations[name] if name in self.capacity_reservations else None
```
bblommers:
```suggestion
        return self.capacity_reservations.get(name)
```
Minor nitpick - but this is a bit more readable IMHO

bblommers (changes requested):
Hi @LeekSwan!

Looks like the linter is currently failing:
https://github.com/getmoto/moto/actions/runs/16378019912/job/46282632610?pr=9071

LeekSwan:
Hi @bblommers , thanks for the comments so far. Ive made some changes to reflect them. Ive been running into something curious when doing tagging though. It seems like the "Tags" key and value is being dropped when its passed from response -> test. For example, when running `test_get_capacity_reservation`, I can see that the object created in model.py and response.py has a "Tags" key and value but once it gets to the test file, its suddenly dropped. Any idea what's happening? Ive noticed this happening in the other Athena functions as well.

bpandola:
@LeekSwan Tags aren't returned when you call [GetCapacityReservation](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/athena/client/get_capacity_reservation.html), use [ListTagsForResource](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/athena/client/list_tags_for_resource.html)

LeekSwan:
Oh man. Cant believe I missed that T_T. Thank you!

PR #9075, opened 2025-07-17
Elastic Beanstalk: integrate with core response serializer

bpandola:
Relatively minor changeset required for integration:

* generated new list of solution stacks (previous values were from 2019).
* dropped `Fake` prefix from model names to align with Botocore.
* changed `DeleteApplication` logic when passed a non-existent application name (behavior verified against real AWS backend).

1300+ lines of XML templates deleted!


bblommers:
> 1300+ lines of XML templates deleted!

:partying_face:

PR #9076, opened 2025-07-18
fix: Cloudformation allow changeset updates

thecodingsysadmin:
This commit fixes two issues with Cloudformation change set updates.

Firstly, `moto` returns an exception: `KeyError: 'parameter_value'` for existing parameters if the change set is created without the `UsePreviousTemplate` argument.

This was fixed by removing the condition for `UsePreviousTemplate` in the parameter list generation.

The second issue was default values defined in the Parameter section in the Cloudformation template are not included in the parameter arguments to `CloudFormationBackend.create_change_set` when the change set is being parsed/validated.

This results in a change set creation failing due to missing Parameter values.

The second issue was fixed by replacing references to `FakeStack.parameters` with `FakeStack.stack_parameters` which returns the resource parameters vs just the parameters provided to Cloudformation create/update stack operations.

### DIFF
--- a/moto/cloudformation/responses.py
+++ b/moto/cloudformation/responses.py
@@ -193,10 +193,11 @@ class CloudFormationResponse(BaseResponse):
             for item in self._get_list_prefix("Tags.member")
         )
         parameters = {
-            param["parameter_key"]: stack.parameters[param["parameter_key"]]
-            if param.get("use_previous_value", "").lower() == "true"
-            and use_previous_template
-            else param["parameter_value"]
+            param["parameter_key"]: (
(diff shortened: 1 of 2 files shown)


thecodingsysadmin:
Just checking against an actual account what the change set will look like with default parameters.

Right now we blindly add to them to as ParameterKey:ParameterValue `parameter` argument of `create_change_set`

thecodingsysadmin:
The changes in this PR should be enough to fix the issues with changeset updates.

bblommers (approved):
LGTM - thanks for investigating and fixing this @thecodingsysadmin!

PR #9096, opened 2025-07-23
IAM: integrate exceptions with core serializer

bpandola:
Also a fair bit of clean up and a number of defects fixed.  Some of those existing error code assertions were 🤨


bpandola:
Re patch coverage: the missing lines are just renamed Exception classes, which have never been covered by tests.

PR #9110, opened 2025-07-28
RDS: add support for DBCluster/DBInstance DomainMemberships

leslieetubo:
To reviewers and maintainers @bblommers and @bpandola . This PR adds the domain membership attribute to rds cluster and instance response according to [boto3 api docs](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/rds/client/create_db_cluster.html). 
>NOTE: This is a rudimentary implementation, since boto3 doesn't have an endpoint to add more domain memberships for rds client. For now,  there can be only one domain membership and FQDN (Fully qualified domain name) for db clusters is broken to fill OU, Domain and FQDN attributes.

All tests are passing.


moto/rds/models.py
```diff
@@ -1132,6 +1154,25 @@ def __init__(
         self.multi_az = multi_az
         self.db_subnet_group_name = db_subnet_group_name
         self.db_security_groups = db_security_groups or []
+        self.vpc_security_group_ids = vpc_security_group_ids or []
```
bpandola: Looks like a failed master merge after #9128 - shouldn't be any modification to this attribute in your PR.

moto/rds/models.py
```diff
@@ -1132,6 +1154,25 @@ def __init__(
...
+                "FQDN": domain_fqdn,
+                "IAMRoleName": "rds-directory-service-access-role",
+                "OU": domain_ou,
+                "AuthSecretArn": domain_auth_secret_arn,
+                "DnsIps": domain_dns_ips,
+            }
+            if domain
+            else None
+        )
+        if not self.vpc_security_group_ids:
```
bpandola: Same here.  There shouldn't be any changes to `vpc_security_group_ids`.

moto/rds/models.py
```diff
@@ -845,8 +862,8 @@ def vpc_security_group_ids(
         self._vpc_security_group_ids = vpc_security_group_ids
 
     @property
-    def domain_memberships(self) -> List[str]:
-        return []
```
bpandola, line 849: This was originally just a dummy method, so it got the type hint wrong.  No need to keep `list[str]` around as an option.  Also, don't return `None`.  AWS always returns an empty list if there are no memberships.

moto/rds/models.py
```diff
@@ -586,6 +588,22 @@ def __init__(
...
+                or "rds-directory-service-access-role",
+                "OU": f"OU={domain.split('.')[0]}OU,DC={domain.split('.')[0]},DC=com"
+                if domain
+                else None,
+                "AuthSecretArn": None,  # Not implemented yet
+                "DnsIps": [],
+            }
+            if domain
+            else None
+        )
```
bpandola: How about encapsulating this as a DomainMembership object that takes in the relevant params (`domain`, etc.)?  I'd really like to get away from dicts when we know the shape of the object/attributes ahead of time.

bpandola (changes requested):

moto/rds/models.py
leslieetubo: Ohh yeah you are right. I will make these modifications. Thanks @bpandola

PR #9114, opened 2025-07-30
RDS:  fix blue/green deployment logic around managed master user passwords

lukaalba:
AWS does not support rds instances and clusters which have managed master user credentials. Therefore, the moto implementation of `create_blue_green_deployment` throws an exception.

However, even after modifying the rds instance and disable the manage master user password, `create_blue_green_deployment` still throws an error.

The current implementation checks if the field manage_master_user_password is set to True, but when disabling it, this field was never set to False. Furthermore, since this field does not exist in the boto3 response, I've changed the conditional, so that it checks if master_user_secret exists in the DBInstance or DBCluster object.


moto/rds/models.py
```diff
@@ -4508,8 +4513,11 @@ def _create_green_instance(
     ) -> str:
         source_instance: DBInstance | DBCluster = self._get_instance(self.source)
         green_instance: DBInstance | DBCluster
-        if source_instance.manage_master_user_password:
-            raise SourceDatabaseNotSupportedFault(source_instance.arn)
+        if hasattr(source_instance, "master_user_secret"):
+            if isinstance(source_instance, DBInstance):
+                raise SourceDatabaseNotSupportedFault(source_instance.arn)
+            else:
+                raise SourceClusterNotSupportedFault(source_instance.arn)
```
bpandola: Would you mind extracting this into a helper method, something like `valid_bgd_source_or_raise`?

bpandola (changes requested):
One minor request, then happy to merge.  Thanks.

moto/rds/models.py
lukaalba: I moved it into the  `_get_instance` function and renamed it since there is already a conditional which checks if the database is a cluster or an rds instance

PR #9119, opened 2025-07-30
[EKS] Adding support for update-cluster-config

LeekSwan:
This PR addresses https://github.com/getmoto/moto/issues/9118 by adding support for  [update_cluster_config](https://docs.getmoto.org/en/latest/docs/services/eks.html#:~:text=update_cluster_config)


bblommers (approved):
LGTM - thanks again for contributing and improving Moto @LeekSwan!

Issue #9116, opened 2025-07-30
describe-stack-resources returns logical resources that don't match


iainelder:
Am I missing something? I think the real CloudFormation service would return an empty list here.

Instead Moto returns a list of logical resources that don't match the request.

Versions:

Python 3.13.5
moto 5.1.9
boto3 1.39.16

Expected result: test passes.

Actual result:

```
E       AssertionError: assert [{'LogicalRes...Bucket', ...}] == []
E         
E         Left contains one more item: {'LogicalResourceId': 'Dummy', 'PhysicalResourceId': 'teststack1-dummy-qvdc64atsfme', 'ResourceStatus': 'CREATE_COMPLETE', 'ResourceType': 'AWS::S3::Bucket', ...}
E         Use -v to get more diff

test_moto_stack_resource.py:55: AssertionError
```

Self-contained test code follows. Copy it to an executable file and then run it.

```python
#!/usr/bin/env -S uv run --script
#
# /// script
# requires-python = ">=3.12"
# dependencies = [
#   "pytest",
#   "moto",
#   "boto3",
#   "openapi-spec-validator", # Fixes ModuleNotFoundError.
# ]
# ///

import pytest
from _pytest.monkeypatch import MonkeyPatch
from typing import Iterable
from moto import mock_aws
from boto3 import Session
from botocore.exceptions import ClientError

@pytest.fixture(autouse=True)
def fake_aws_credentials(monkeypatch: MonkeyPatch) -> None:
    for var in "AWS_PROFILE", "AWS_ACCESS_KEY_ID", "AWS_SECRET_ACCESS_KEY":
        monkeypatch.setenv(var, "broken_not_real_profile")
    monkeypatch.delenv("AWS_PROFILE")


@pytest.fixture(autouse=True)
def aws_region(monkeypatch: MonkeyPatch) -> None:
    monkeypatch.setenv("AWS_DEFAULT_REGION", "eu-central-1")


@pytest.fixture
def session() -> Iterable[Session]:
    with mock_aws():
        yield Session()


def test_missing_stack_resource_returns_empty_list(session: Session):

    session.client("cloudformation").create_stack(
        StackName="teststack1",
        TemplateBody="""
            Resources:
                Dummy:
                    Type: AWS::S3::Bucket
        """
    )

    response = session.client("cloudformation").describe_stack_resources(
        StackName="teststack1",
        LogicalResourceId="AnyOtherResource",
    )

    assert "StackResources" in response
[...]

Issue #9073, opened 2025-07-17
Refactor: Request Parsing/Response Serialization


bpandola:
# Summary 

Over the coming weeks and months, Moto's request parsing/response serialization pipeline will be unified and normalized across all services. 

When complete, Moto will have a robust set of core methods for parsing requests and serializing responses applied consistently throughout the project. Backends will be largely free of serialization/parsing concerns, allowing their code to be more focused on an individual service's implementation details.  Whole categories of defects will have been eliminated and various patterns within the codebase will have been greatly simplified, ultimately making the project more stable and easier to maintain/contribute. 

> [!IMPORTANT]
> For those who only interact with Moto via the AWS API boundary (using a client like Boto3, or other SDK's), this work should go virtually unnoticed.

> [!WARNING]
> If you are poking directly at Moto's internals, be aware that this work will include a fair bit of restructuring/renaming of existing modules/methods/classes/attributes etc. 

Individual service backends will be slowly integrated into the new request/response pipeline, starting with the services that will benefit the most (e.g. Query protocol-based services). Additional features will be added as needs arise; defects will be addressed as they are encountered.

All related Issues and Pull Requests will be labeled to provide an at-a-glance view of the work in progress: [here](https://github.com/getmoto/moto/issues?q=label%3Arequest-parsing%2Cresponse-serialization)

# Details

## Intent 

Leverage the AWS service definitions in Botocore to create generalized request parsers and response serializers for Moto.

## Motivation 

From its inception, the core Moto logic delegated some request parsing and most response serialization duties to each individual backend service implementation.   With so many mocked services implemented incrementally over the years by numerous contributors, a lot of cruft has built up around request parsing/response serialization.

The base request parsing code grew haphazardly over time, as it attempted to handle the nuances of each AWS wire protocol. Service-specific parameter parsing code was more ad hoc, resulting in numerous specialized methods peppered throughout the various backends, often duplicating similar methods in other backends.  

Serialization is currently the responsibility of each individual service--with some dumping JSON, some returning XML data, and some doing both.  Backend models are often polluted with serialization concerns based on the expected response format (e.g. timestamps stored as strings, because JSON doesn't have a built-in date type; or boolean values stored as strings to match the XML representation), which unnecessarily complicates the service logic and is prone to error (e.g. what does the conditional `if obj.boolean_value:` evaluate to when `boolean_value` is the string `"false"`?).

[...]

PR #9120, opened 2025-08-01
CloudFormation: Add LogicalResourceId param to DescribeStackResources

iainelder:
Support the LogicalResourceId parameter in the DescribeStackResources API.

When set, it returns the resource wrapped in a list if it exists, or returns an empty list if the resource doesn't exist.

Make the tests clearer by removing an unneeded call to DescribeStacks. We don't need to compare the StackId here.

A solution for #9116 .

### DIFF
--- a/moto/cloudformation/models.py
+++ b/moto/cloudformation/models.py
@@ -1264,9 +1264,16 @@ class CloudFormationBackend(BaseBackend):
         raise ValidationError(stack_name, message)
 
     def describe_stack_resources(
-        self, stack_name: str
+        self, stack_name: str, logical_resource_id: str
     ) -> Tuple[Stack, Iterable[Type[CloudFormationModel]]]:
         stack = self.get_stack(stack_name)
+
+        if logical_resource_id is not None:
+            for res in stack.stack_resources:
+                if res.logical_resource_id == logical_resource_id:
+                    return stack, [res]
+            return stack, []
+
         return stack, stack.stack_resources
 
     def list_stack_resources(
--- a/moto/cloudformation/models.py
+++ b/moto/cloudformation/models.py
@@ -1270,7 +1270,7 @@ class CloudFormationBackend(BaseBackend):
 
         if logical_resource_id is not None:
             for res in stack.stack_resources:
-                if res.logical_resource_id == logical_resource_id:
+                if res.logical_resource_id == logical_resource_id:  # type: ignore[attr-defined]
                     return stack, [res]
             return stack, []
 
--- a/moto/cloudformation/responses.py
+++ b/moto/cloudformation/responses.py
@@ -325,8 +325,9 @@ class CloudFormationResponse(BaseResponse):
 
     def describe_stack_resources(self) -> ActionResult:
         stack_name = self._get_param("StackName")
+        logical_resource_id = self._get_param("LogicalResourceId")
         stack, resources = self.cloudformation_backend.describe_stack_resources(
-            stack_name
+            stack_name, logical_resource_id
         )
 
         result = {
--- a/tests/test_cloudformation/test_cloudformation_stack_crud_boto3.py
+++ b/tests/test_cloudformation/test_cloudformation_stack_crud_boto3.py
@@ -1723,14 +1723,43 @@ def test_describe_stack_resources():
     cf = boto3.client("cloudformation", region_name=REGION_NAME)
     cf.create_stack(StackName=TEST_STACK_NAME, TemplateBody=dummy_template_json)
 
-    stack = cf.describe_stacks(StackName=TEST_STACK_NAME)["Stacks"][0]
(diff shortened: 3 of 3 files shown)


iainelder:
The CloudFromation unit tests pass (`pytest tests/test_cloudformation`).

The linter fails because it doesn't recognize the `logical_resource_id` property on the internal representation of the StackResource data type.

`moto/cloudformation/models.py:1273:20: error: "type[CloudFormationModel]" has no attribute "logical_resource_id"  [attr-defined]`

How do we satisfy the type checker?

bpandola:
> How do we satisfy the type checker?

You'll probably have to use `getattr`, like what had to be done here: https://github.com/getmoto/moto/blob/c618ffdabf2d2130ba206d6f45a849de701e98dd/moto/cloudformation/responses.py#L29

iainelder:
`getattr` is ugly, but it's good enough for now.

I don't know a lot about moto's internals yet, but the types for CloudFormation look odd. It looks like there is confusion between the internal types that exist to implement CloudFormation support for resources in other services, and external types that exist to model the responses that come from the CloudFormation service APIs.

A quick search of the full code suggests that the CloudFormationModel class is a mixin that provides an interface for all the types that can be created via CloudFormation: SQS Queues, EventBridge Rules, KMS Keys, and more.

That's fine, but I don't understand why that type is also returned from describe_stack_resources and describe_stack_resource. Instead I was expecting to see something like the CloudFormation API's [StackResource data type](https://docs.aws.amazon.com/AWSCloudFormation/latest/APIReference/API_StackResource.html).

The StackResourceDTO has the shape I expect, but it's used later than I would expect, in the "responses" layer that wraps the "model" layer.

Am I missing something  basic about how Moto works?

iainelder:
I just noticed that the existing code for `describe_stack_resource` faces the same problem and solves it by telling the type checker to ignore it.

I think it reads better than `getattr`, so I'll do that too.

https://github.com/getmoto/moto/blob/77f7eb5b507b08ebf964b7170e554dddb479899f/moto/cloudformation/models.py#L1258

bpandola:
> Am I missing something basic about how Moto works?

@iainelder Maybe just the broader context: Moto has been developed--mostly piecemeal--by 100+ contributors over more than 12 years.  Type hints were only introduced (to Moto) ~2 years ago, and it took quite a bit of work--and yes--hacks and "ignores" to wrangle the codebase into submission.  CloudFormation is a complex service that interacts with many other service backends.  Some of that code is messy; some of the type-hinting, as you point out, could probably be improved (or might even be incorrect).  I ended up adding the `StackResourceDTO` class to get around the "confusion" in the model that you have also noticed.  I'm currently working on normalizing response serialization in Moto, which is a big enough undertaking on its own that I simply can't stop to fix or improve every bit of code I encounter that could use a fix or improvement (even though I'd like to).

So, no, I don't think you're missing something.  I think what you're pointing out is valid and could definitely be improved.  If that's something you feel like attempting to untangle, PRs are always welcome. 😉

iainelder:
@bpandola, thanks for the context. I'm honestly in awe of how well moto works considering it gets no support from AWS at all.

I've just found the issue about refactoring the serialization: #9073. I'm really glad you've identified this and are working to improve it. I don't know enough about the serialization part to help with it just now, but I am keen to see how it develops.

I do know a lot of the AWS services quite well, and I would like to help you improve support for them to make them as close to the real thing as is practical. I can keep making those improvements as I identify them, keeping in mind the history, your current priorities, and my limited time to contribute.

This is my vision for the service models: After the serialization is normalized, It would be great for all the service models to return types that are basically DTOs that map to AWS's on published model. (For example, StackResourceDTO maps to the [StackResource data type](https://docs.aws.amazon.com/AWSCloudFormation/latest/APIReference/API_StackResource.html)). Does that fit with your vision?

(I know CloudFormation is a bit of a special case because of how it integrates with all the other services, but I'm sure we could find a way to represent it with sensible types that is also a robust implementation.)

iainelder:
Just noticed you've had the same discussion on a different CloudFormation issue and it looks like we're thinking along the same lines for the future development :-D

> All the data necessary to create AWS service definitions is in botocore already, so if/when we go that route, we'll probably write our own parser.

https://github.com/getmoto/moto/issues/9077#issuecomment-3124339782

bpandola:
> I do know a lot of the AWS services quite well, and I would like to help you improve support for them to make them as close to the real thing as is practical. I can keep making those improvements as I identify them, keeping in mind the history, your current priorities, and my limited time to contribute.
> 
Please do!
> This is my vision for the service models: After the serialization is normalized, It would be great for all the service models to return types that are basically DTOs that map to AWS's on published model. (For example, StackResourceDTO maps to the [StackResource data type](https://docs.aws.amazon.com/AWSCloudFormation/latest/APIReference/API_StackResource.html)). Does that fit with your vision?
> 
Yes.

PR #9127, opened 2025-08-01
feat(batch): Enable asterisk match when providing filters to Batch list_jobs

CesarCMa:
Hi!

**Features of the PR:**

a) Add support for asterisk match when providing filters in the Batch `list_jobs` method.
b) Ignore `jobStatus` when the `filters` parameter is provided, to match the behavior described in the Boto3 documentation.

---

### Detail about a)

According to the Boto3 documentation, it is possible to provide filters to the `list_jobs` method using an asterisk so that it filters based on the beginning of the job name:

> "The value of the filter is a case-insensitive match for the job name. If the value ends with an asterisk (*), the filter matches any job name that begins with the string before the ‘*’. This corresponds to the jobName value. For example, test1 matches both Test1 and test1, and test1\* matches both test1 and Test10. When the JOB\_NAME filter is used, the results are grouped by the job name and version."
> *Source: [Boto3 Batch list\_jobs doc](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/batch/client/list_jobs.html)*

The new code enables this kind of filter for `JOB_NAME`.

---

### Detail about b)

The documentation also states that if `filters` are provided, `jobStatus` will be ignored:

> "The filter to apply to the query. Only one filter can be used at a time. When the filter is used, jobStatus is ignored."

The current implementation checks both of them when filtering jobs, which can be a bit misleading when testing interactions with the Batch service.

---

### Notes

In this implementation, I defined a function within the `list_jobs` method to avoid deep indentation. Normally, I would encapsulate this logic in a private function at the module level, but since `models.py` is already quite large, I thought it might be confusing to define it outside the method.

However, feel free to suggest a better approach or a cleaner solution 😄


moto/batch/models.py
```diff
@@ -1865,24 +1865,41 @@ def list_jobs(
             if array_job := self.get_job_by_id(array_job_id):
                 jobs_to_check.extend(array_job._child_jobs or [])
 
-        if job_status is not None and job_status not in JobStatus.job_statuses():
+        if not jobs_to_check:
```
bblommers, line 1870: Nice catch - exiting as early as possible is always a good strategy :clap:

moto/batch/models.py
```diff
@@ -1865,24 +1865,41 @@ def list_jobs(
...
+            if filter["name"] == "JOB_NAME":
+                for value in filter["values"]:
+                    if value.endswith("*"):
+                        pattern = value[:-1].lower()
+                        if job.job_name.lower().startswith(pattern):
+                            return True
+                    else:
+                        if job.job_name.lower() == value.lower():
+                            return True
+            return False
```
bblommers:
```suggestion
                return False
            return True
```
Our current implementation assumes True by default. Only when a `JOB_NAME` is provided and does not match, is a job not included.

The suggested change would keep that behaviour intact, so the logic essentially looks like this:
```
if filter["name"] == "JOB_NAME":
    # try to match..
    ...
    # no match
    return False
# Filter is unsupported - passthrough
return True
```

report.md
bblommers, line 1: Great to have this written out! That can be very useful when we further improve the behaviour of this method in the future.

One nitpick - can we move it to the `moto/batch` folder?

bblommers (changes requested):

report.md
CesarCMa: Ups, this file was unintentionally committed; it's meant to be internal notes to help with documenting the PR.

I'll remove it and instead add a TODO in the function's docstring to clarify that the filter parameter should not be allowed when listing batch array jobs.

moto/batch/models.py
CesarCMa: Right, missed the scenario of non-implemented filters. I've also added a comment to clarify that we return `True` for non supported filters.

Thanks for the feedback!

PR #9144, opened 2025-08-05
Glue: List crawls

BlizzardOf:
This implements the list_crawls function for glue

- Added a crawl object, which has a state associated with it
- Changes the crawler backend object to maintain its history of crawls, and to derive its state from the latest crawl
- Dynamically builds filter functions which are used to filter the crawl history
- Used a pydantic model with an error translation wrapper function to build and validate the filter object in the input. It's not ideal, but I found myself essentially re implementing stuff pydantic already does so I figured it's better to use that
- I discovered that the "_InvalidOperationException" was actually duplicating the error message, the same thing is handled by moto at the top level, so I removed it entirely


moto/glue/models.py
```diff
@@ -207,6 +209,12 @@ class GlueBackend(BaseBackend):
             "limit_default": 100,
             "unique_attribute": "job_run_id",
         },
+        "list_crawls": {
```
bblommers, line 212: Nice to see pagination supported!

moto/glue/utils.py
```diff
@@ -2,9 +2,11 @@
 import operator
 import re
 from datetime import date, datetime
+from enum import Enum
 from itertools import repeat
 from typing import Any, Dict, List, Optional, Union
 
+from pydantic import BaseModel, ValidationError, model_validator
```
bblommers: The `pydantic` is a transitive dependency in Moto - it gets pulled in via `cfn-lint`, but it is not a direct dependency. 

When users install some service dependencies (`pip install moto[glue]`), it will actually not install the `pydantic` dependency, so this will fail.

In general, we try to have the least number of dependencies possible - so in this case, I would prefer to have our own validation logic here, instead of using `pydantic`.

moto/glue/utils.py
```diff
@@ -20,6 +22,7 @@
     one_of,
     pyparsing_common,
 )
+from typing_extensions import Self
```
bblommers: Same issue here - `typing_extensions` is a transitive dependency, so this may not always exist. It would be better to use the explicit type, instead of `Self`.

tests/test_glue/test_datacatalog.py
```diff
@@ -1324,6 +1324,12 @@ def test_get_crawlers_several_items():
 def test_start_crawler():
     client = boto3.client("glue", region_name="us-east-1")
     name = "my_crawler_name"
+
+    helpers.set_transition(
```
bblommers: This is a breaking change. Can we set this as the default transition instead? That should ensure that this behaviour stays the same for existing users.

Default transitions are registered here:

https://github.com/getmoto/moto/blob/master/moto/moto_api/__init__.py

tests/test_glue/test_datacatalog.py
bblommers, line 1: Nice and extensive test suite! That's always great to see :partying_face:

moto/glue/utils.py
BlizzardOf: Yeah, makes sense. It's a shame, it's so convenient for this kind of input validation

tests/test_glue/test_datacatalog.py
BlizzardOf: Okay yeah, I was wondering what that was when I was searching for transition examples

tests/test_glue/test_datacatalog.py
BlizzardOf: Thanks!

PR #9148, opened 2025-08-07
SQS: Support MessageGroupId on standard queues

carterv:
Now that SQS supports setting a MessageGroupId on standard queues as part of the "fair queues" feature, this validation logic isn't necessary. See https://github.com/getmoto/moto/issues/9146 for more context

Linting, type checking, and unit tests were all run locally and passed.


bblommers (approved):
LGTM - thanks for enabling this @carterv, and welcome to Moto!

PR #9149, opened 2025-08-07
EC2: Add sg support to modify_vpc_endpoint mock

dtaskai:
Not too happy with the retrieval of the security group attributes as I could have used `describe_security_groups` instead of looping and doing it individually for each security group ID, but after checking it the describe mock doesn't support pagination as of now and I didn't want to cause trouble downstream. This way we still get the desired functionality as describing the VPC endpoint returns the security groups in the right format and we can add and remove them with `modify_vpc_endpoint`.

### DIFF
--- a/tests/test_ec2/test_vpcs.py
+++ b/tests/test_ec2/test_vpcs.py
@@ -1316,6 +1316,81 @@ def test_modify_vpc_endpoint_remove_route_table():
 
 
 @mock_aws
+def test_modify_vpc_endpoint_add_security_group():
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -1,7 +1,7 @@
 FROM python:3.13-alpine
 
 ADD . /moto/
-ENV PYTHONUNBUFFERED 1
+ENV PYTHONUNBUFFERED=1
```
bblommers, line 4: Nice catch - I'd seen the warnings, but hadn't gotten around to fixing it :+1:

PR #9152, opened 2025-08-08
SES: Add CreatedTimestamp to ReceiptRuleSet metadata

viren-nadkarni:
This PR adds the `Metadata.CreatedTimestamp` attribute to the SES [DescribeReceiptRuleSet](https://docs.aws.amazon.com/ses/latest/APIReference/API_DescribeReceiptRuleSet.html) operation.

Earlier this attribute was hardcoded, but it got lost in #9100


bblommers (approved):
Love it - thank you for fixing this @viren-nadkarni!

PR #9153, opened 2025-08-08
EC2: Prefix lists do not return StateMessage

viren-nadkarni:
This PR adjusts the behaviour of [DescribeManagedPrefixLists](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_DescribeManagedPrefixLists.html) operation to resemble AWS.

AWS normally does not return a value for [StateMessage](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_ManagedPrefixList.html) field.

The behaviour is validated against AWS using a verified test.

### DIFF
--- a/moto/ec2/models/managed_prefixes.py
+++ b/moto/ec2/models/managed_prefixes.py
@@ -24,7 +24,7 @@ class ManagedPrefixList(TaggedEC2Resource):
         self.id = random_managed_prefix_list_id()
         self.prefix_list_name = prefix_list_name
         self.state = "create-complete"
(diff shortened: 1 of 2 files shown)


bblommers (changes requested):
That bug is ancient! [Introduced 4 years ago, almost on the day](https://github.com/getmoto/moto/pull/4169).

Happy to merge once we get the tests to pass!

bblommers (approved):
LGTM! Thanks for validating this against AWS @viren-nadkarni :pray:

PR #9170, opened 2025-08-12
[S3 Control] Adding list_access_points

ghorondo:
This PR adds support for the following S3 Control API:

[list_access_points method](https://docs.getmoto.org/en/latest/docs/services/s3control.html#:~:text= list_access_points)

### DIFF
--- a/tests/test_s3control/test_s3control_access_points.py
+++ b/tests/test_s3control/test_s3control_access_points.py
@@ -87,6 +87,33 @@ def test_get_access_point_full():
     }
 
 
+@mock_aws
+def test_list_access_points():
+    client = boto3.client("s3control", region_name="ap-southeast-1")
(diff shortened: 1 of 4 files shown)


tests/test_s3control/test_s3control_access_points.py
```diff
@@ -58,33 +58,64 @@ def test_get_access_point_minimal():
 
 
 @mock_aws
-def test_get_access_point_full():
```
bblommers: What was the reason for deleting this method? The new tests do not use `get_access_point`, so it doesn't look like they are a replacement

bblommers (changes requested):

tests/test_s3control/test_s3control_access_points.py
rondodev: My mistake, I restored get_access_point. Thanks!

PR #9172, opened 2025-08-13
Elasticache: Allowing tags as part of CreateUser action

arjun-sn24:
Fixes https://github.com/getmoto/moto/issues/8521


bpandola (approved):
Thanks @arjun-sn24!

PR #9173, opened 2025-08-13
[Comprehend] Implement async jobs and resource policies

ghorondo:
This PR adds support for AWS Comprehend by adding support for resource policies and the full lifecycle of all nine asynchronous detection jobs.

1. Implements the following resource policy operations, allowing users to attach, view, and delete policies from Comprehend resources (PutResourcePolicy, DescribeResourcePolicy, DeleteResourcePolicy)

2. Asynchronous Detection Jobs: Added support for the Start*, Describe*, List*, and Stop* operations for all nine asynchronous job types

To handle all the similar Comprehend jobs efficiently, I made one ComprehendJob class to store shared information like the Job ID and status.


bblommers (changes requested):
Hi @ghorondo, thanks for the PR!

Would you mind moving the tests to a dedicated file? The main file is getting quite big already, so splitting them up would help readability.

PR #9175, opened 2025-08-13
ELB: Resource ARNs must be unique

viren-nadkarni:
This PR updates the ARN construction logic for following resources such that they are unique:
- Listeners
- Load balancers
- Target groups

### DIFF
--- a/moto/elbv2/utils.py
+++ b/moto/elbv2/utils.py
@@ -1,9 +1,12 @@
+from moto.moto_api._internal import mock_random
 from moto.utilities.utils import get_partition
 
 
 def make_arn_for_load_balancer(account_id: str, name: str, region_name: str) -> str:
-    return f"arn:{get_partition(region_name)}:elasticloadbalancing:{region_name}:{account_id}:loadbalancer/app/{name}/50dc6c495c0c9188"
+    lb_id = mock_random.get_random_hex(length=16)
+    return f"arn:{get_partition(region_name)}:elasticloadbalancing:{region_name}:{account_id}:loadbalancer/app/{name}/{lb_id}"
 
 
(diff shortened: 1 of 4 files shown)


moto/elbv2/utils.py
```diff
@@ -1,9 +1,12 @@
+from moto.moto_api._internal import mock_random
 from moto.utilities.utils import get_partition
 
 
 def make_arn_for_load_balancer(account_id: str, name: str, region_name: str) -> str:
-    return f"arn:{get_partition(region_name)}:elasticloadbalancing:{region_name}:{account_id}:loadbalancer/app/{name}/50dc6c495c0c9188"
+    lb_id = mock_random.get_random_hex(length=16)
```
bpandola: Let's follow the convention here and pass `lb_id` in as a parameter.  That value is also used to generate other attribute names (e.g. `DNSName`).  You don't have to fix those other attributes in this PR, but having `lb_id` available at the call site will make it easier to add in the future.  Thanks.

moto/elbv2/utils.py
```diff
@@ -1,9 +1,12 @@
...
+    return f"arn:{get_partition(region_name)}:elasticloadbalancing:{region_name}:{account_id}:loadbalancer/app/{name}/{lb_id}"
 
 
 def make_arn_for_target_group(account_id: str, name: str, region_name: str) -> str:
-    return f"arn:{get_partition(region_name)}:elasticloadbalancing:{region_name}:{account_id}:targetgroup/{name}/50dc6c495c0c9188"
+    tg_id = mock_random.get_random_hex(length=16)
```
bpandola: Same here.

bpandola (changes requested):

bblommers (approved):
Nice - thanks @viren-nadkarni!

PR #9177, opened 2025-08-13
[Athena] Adding Athena resource group tagging api

LeekSwan:
This PR looks to resolve https://github.com/getmoto/moto/pull/9039 and implements resource group tagging for the following Athena resources:
- Data catalog
- Capacity Reservations
- Workgroups

[list_tags_for_resource](https://docs.getmoto.org/en/latest/docs/services/athena.html#:~:text=list_tags_for_resource) was also implemented

### DIFF
--- a/moto/resourcegroupstaggingapi/models.py
+++ b/moto/resourcegroupstaggingapi/models.py
@@ -2,6 +2,7 @@ from typing import Any, Dict, Iterator, List, Optional, Tuple
 
 from moto.acm.models import AWSCertificateManagerBackend, acm_backends
 from moto.appsync.models import AppSyncBackend, appsync_backends
+from moto.athena.models import AthenaBackend, athena_backends
 from moto.awslambda.models import LambdaBackend, lambda_backends
 from moto.backup.models import BackupBackend, backup_backends
 from moto.clouddirectory import CloudDirectoryBackend, clouddirectory_backends
@@ -69,6 +70,10 @@ class ResourceGroupsTaggingAPIBackend(BaseBackend):
     @property
     def appsync_backend(self) -> AppSyncBackend:
         return appsync_backends[self.account_id][self.region_name]
+
+    @property
+    def athena_backend(self) -> AthenaBackend:
+        return athena_backends[self.account_id][self.region_name]
 
     @property
     def s3_backend(self) -> S3Backend:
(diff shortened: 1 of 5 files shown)


moto/resourcegroupstaggingapi/models.py
```diff
@@ -310,6 +315,39 @@ def format_tag_keys(
 
                 yield {"ResourceARN": f"{api.arn}", "Tags": tags}
 
+        # Athena
+        if self.athena_backend:
```
bpandola: Why is this necessary?

moto/resourcegroupstaggingapi/models.py
```diff
@@ -310,6 +315,39 @@ def format_tag_keys(
...
+            if not resource_type_filters or "athena" in resource_type_filters:
+                athena_backend = athena_backends[self.account_id][self.region_name]
+
+                # Capacity Reservations
+                for (
+                    capacity_reservation
+                ) in self.athena_backend.capacity_reservations.values():
+                    tags = athena_backend.tagger.list_tags_for_resource(
+                        capacity_reservation.name
```
bpandola: I don't understand the difference between your use of `athena_backend` vs `self.athena_backend`... what am I missing?

bpandola (changes requested):

moto/resourcegroupstaggingapi/models.py
LeekSwan: removed!

moto/resourcegroupstaggingapi/models.py
LeekSwan: It should have been `athena_backend`. Updated!

moto/resourcegroupstaggingapi/models.py
```diff
@@ -70,6 +71,10 @@ def __init__(self, region_name: str, account_id: str):
     def appsync_backend(self) -> AppSyncBackend:
         return appsync_backends[self.account_id][self.region_name]
 
+    @property
+    def athena_backend(self) -> AthenaBackend:
+        return athena_backends[self.account_id][self.region_name]
+
```
bpandola: You have to use this property or get rid of it, otherwise it's uncovered code.  (I would suggest using it on line 320.)

bpandola (changes requested):

PR #9178, opened 2025-08-13
Making start job run a little more functional

BlizzardOf:
This addresses [this issue](https://github.com/getmoto/moto/issues/9033). Previously we were just generating hardcoded mock values for all the values of the job. This allows us to pass in most of the supported values from the start_job_run, otherwise it gets them from the create_job values.

### DIFF
--- a/tests/test_glue/test_glue_job_runs.py
+++ b/tests/test_glue/test_glue_job_runs.py
@@ -80,26 +80,7 @@ def test_get_job_run():
 
     response = client.get_job_run(JobName=job_name, RunId=job_run_id)
     assert response["JobRun"]["Id"] == job_run_id
-    assert response["JobRun"]["Attempt"]
-    assert response["JobRun"]["PreviousRunId"]
-    assert response["JobRun"]["TriggerName"]
-    assert response["JobRun"]["StartedOn"]
-    assert response["JobRun"]["LastModifiedOn"]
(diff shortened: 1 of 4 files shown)


tests/test_glue/test_glue_job_runs.py
```diff
@@ -184,6 +165,142 @@ def test_job_run_transition():
...
+    ("MaxCapacity", 5.0),
+    ("NumberOfWorkers", 3),
+    ("SecurityConfiguration", "my-security-config"),
+    ("Timeout", 60),
+    ("WorkerType", "G.1X"),
+]
+
+
+@mock_aws
+@pytest.mark.parametrize("parameter,value", parameters_and_values)
```
bblommers, line 292: Nice - parametrizing is always a good way to keep the LoC down

tests/test_glue/test_glue_job_runs.py
```diff
@@ -184,6 +165,142 @@ def test_job_run_transition():
     state_manager.unset_transition("glue::job_run")
 
 
+@mock_aws
+def test_job_run_arguments() -> None:
```
bblommers, line 169: Nice clean tests as well, all with a single responsibility!

bblommers (approved):
LGTM - thank you for improving this area in Moto @BlizzardOf!

PR #9180, opened 2025-08-14
Redshift: Clusters must have a unique ID

viren-nadkarni:


bblommers (approved):
Nice catch @viren-nadkarni - thanks for the PR! :pray:

PR #9182, opened 2025-08-14
IAM: Avoid KeyError for older policies

viren-nadkarni:


viren-nadkarni:
Closing as it appears AWS has fully dropped this legacy feature

PR #9192, opened 2025-08-17
SecretsManager: Using ClientRequestToken parameter to validate against existing version(s) of a secret, during CreateSecret/PutSecretValue/UpdateSecret actions

arjun-sn24:
Resolves https://github.com/getmoto/moto/issues/9012


arjun-sn24:
Hey @bblommers, quick question if you don't mind. 

I got stuck trying to reproduce the error in the build. It appears in [this test](https://github.com/getmoto/moto/blob/cc0fa6c6532bd06eb7ac8559be6aa4b6160fc7be/tests/test_secretsmanager/test_secretsmanager.py#L1249). 

Whereas, when I start a local server and execute `TEST_SERVER_MODE=true make test-only`, what I see is the test getting stuck in the while loop inside it 😅 Would really appreciate a pointer or two on what (obvious thing) I seem to be missing.

bblommers:
Hi @arjun-sn24! That is not obvious at all - but that test can only run if Moto runs inside a Docker container.

If you follow the steps from our CI, you should be able to create the image and run it:

https://github.com/getmoto/moto/blob/cc0fa6c6532bd06eb7ac8559be6aa4b6160fc7be/.github/workflows/tests_servermode.yml#L22C1-L25C45

Note that the failing test is marked as `aws_verified`, so we know it succeeds when running this against AWS. If this test fails against Moto, that is a sign that the implementation is still lacking something. It might be worth running the new test against AWS as well, just to verify that AWS behaves as one would expect.

Let us know if you need some help tracking down the problem!

arjun-sn24:
> Hi @arjun-sn24! That is not obvious at all - but that test can only run if Moto runs inside a Docker container.
> 
> If you follow the steps from our CI, you should be able to create the image and run it:
> 
> https://github.com/getmoto/moto/blob/cc0fa6c6532bd06eb7ac8559be6aa4b6160fc7be/.github/workflows/tests_servermode.yml#L22C1-L25C45
> 
> Note that the failing test is marked as `aws_verified`, so we know it succeeds when running this against AWS. If this test fails against Moto, that is a sign that the implementation is still lacking something. It might be worth running the new test against AWS as well, just to verify that AWS behaves as one would expect.
> 
> Let us know if you need some help tracking down the problem!

Thanks @bblommers, that was really helpful. Was able to identify the issue, and have now added the correct patch to make the tests pass.

PR #9201, opened 2025-08-20
ELB: Improve parity for TargetGroup and LoadBalancer responses

viren-nadkarni:


moto/elbv2/responses.py
```diff
@@ -371,12 +374,14 @@ def transform_dict(data: dict[str, str]) -> list[dict[str, str]]:
...
-def transform_certificates(data: list[str]) -> list[dict[str, str]]:
-    return [{"CertificateArn": cert} for cert in data]
+def transform_certificates(data: list[str]) -> Optional[list[dict[str, str]]]:
+    return [{"CertificateArn": cert} for cert in data] or None
 
 
 class ELBV2Response(BaseResponse):
     RESPONSE_KEY_PATH_TO_TRANSFORMER = {
+        "CreateTargetGroupOutput.TargetGroups.TargetGroup.LoadBalancerArns": return_if_not_empty,
+        "CreateTargetGroupOutput.TargetGroups.TargetGroup.HealthCheckPort": return_if_not_empty,
```
bpandola: I want to avoid using transformers unless there is no other way to achieve the desired result.  In this case, `HealthCheckPort` is just a string, so if the model ensures that it's a valid string or `None`, then it should behave as needed, right?

moto/elbv2/models.py
```diff
@@ -180,6 +183,10 @@ def __init__(
     def physical_resource_id(self) -> str:
         return self.arn
 
+    @property
+    def health_check_enabled(self) -> bool:
+        return self.healthcheck_enabled
```
bpandola: Let's just rename the attribute to `health_check_enabled` -- no property needed.  I did that for all the other `healthcheck` names, but I guess I missed this one.  Additionally, this is guaranteed to be a `bool` now, so any string comparison stuff in the model is no longer necessary.

moto/elbv2/responses.py
```diff
@@ -371,12 +374,14 @@ def transform_dict(data: dict[str, str]) -> list[dict[str, str]]:
...
 
-def transform_certificates(data: list[str]) -> list[dict[str, str]]:
-    return [{"CertificateArn": cert} for cert in data]
+def transform_certificates(data: list[str]) -> Optional[list[dict[str, str]]]:
+    return [{"CertificateArn": cert} for cert in data] or None
 
 
 class ELBV2Response(BaseResponse):
     RESPONSE_KEY_PATH_TO_TRANSFORMER = {
+        "CreateTargetGroupOutput.TargetGroups.TargetGroup.LoadBalancerArns": return_if_not_empty,
```
bpandola: Follow the convention and make sure this list is sorted alphabetically.

tests/test_elbv2/test_elbv2_target_groups.py
```diff
@@ -133,6 +133,8 @@ def test_create_target_group_and_listeners():
     target_group = response["TargetGroups"][0]
     target_group_arn = target_group["TargetGroupArn"]
     assert target_group["HealthCheckProtocol"] == "HTTP"
+    assert target_group["HealthCheckEnabled"] is True
+    assert target_group.get("LoadBalancerArns") is None
```
bpandola: I think it's more clear to assert that the key is not in the response, like you did for `HealthCheckPort` below.  We don't serialize `None` values, but the way this assert is written makes it seems like the value is there and it's equal to `None`, which is not the case.

tests/test_elbv2/test_elbv2_target_groups.py
```diff
@@ -148,6 +150,7 @@ def test_create_target_group_and_listeners():
     listener = response["Listeners"][0]
     assert listener["Port"] == 80
     assert listener["Protocol"] == "HTTP"
+    assert listener.get("Certificates") is None
```
bpandola: Same here.  Let's assert that the key is not in the response to be more clear.

moto/elbv2/responses.py
viren-nadkarni: Fixed in 9e69201cb

moto/elbv2/models.py
viren-nadkarni: Fixed in fd43903fa

tests/test_elbv2/test_elbv2_target_groups.py
viren-nadkarni: Fixed in d775de956

tests/test_elbv2/test_elbv2_target_groups.py
viren-nadkarni: Fixed in d775de956

moto/elbv2/responses.py
viren-nadkarni: Indeed, I forgot to remove this one. Fixed in 9c0efffb1

PR #9202, opened 2025-08-20
EC2: Add missing attributes to Elastic Network Interfaces

viren-nadkarni:
This PR adds back following attributes to [NetworkInterfaces](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_NetworkInterface.html):
- RequesterID
- RequesterManaged

They used to be hardcoded in the XML template but were lost in #9052

### DIFF
--- a/moto/ec2/responses/_base_response.py
+++ b/moto/ec2/responses/_base_response.py
@@ -1,12 +1,22 @@
 from typing import Any, Dict, Optional
 
 from moto.core.responses import BaseResponse
+from moto.core.serialize import return_if_not_empty
 
 from ..exceptions import EC2ClientError, EmptyTagSpecError, InvalidParameter
 from ..utils import convert_tag_spec
 
 
 class EC2BaseResponse(BaseResponse):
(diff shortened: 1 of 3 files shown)


moto/ec2/models/elastic_network_interfaces.py
```diff
@@ -75,6 +75,8 @@ def __init__(
         self.public_ip_auto_assign = public_ip_auto_assign
         self.start()
         self.add_tags(tags or {})
+        self.requester_managed = False
+        self.requester_id = "AIDAIT2UOQQY3AUEMOTO"
```
bpandola, line 79: I don't see `RequesterId` showing up in real AWS responses unless `RequesterManaged` is `True`, do you?

bpandola (changes requested):

moto/ec2/models/elastic_network_interfaces.py
viren-nadkarni: Yep, I see both with `RequesterManaged=False`.

PR #9219, opened 2025-08-25
STS: validate `AssumeRole.RoleSessionName` length

### LINKED ISSUE #9218: STS assume_role should raise on too long role session name
To reproduce:

```
client = boto3.client("sts", region_name="us-east-1")
client.assume_role(
            RoleArn=".....",
            RoleSessionName="s" * 65,
            DurationSeconds=900,
        )
```

Should throw

```
(ValidationError) when calling the AssumeRole operation: 1 validation error detected: Value 'sssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssss' at 'roleSessionName' failed to satisfy constraint: Member must have length less than or equal to 64
```

sindrig:
This closes #9218

### DIFF
--- a/tests/test_sts/test_sts.py
+++ b/tests/test_sts/test_sts.py
@@ -133,7 +133,6 @@ def test_assume_role(region, partition):
     )
 
 
-@freeze_time("2012-01-01 12:00:00")
 @mock_aws
 def test_assume_role_with_too_long_role_session_name():
     client = boto3.client("sts", region_name="us-east-1")


tests/test_sts/test_sts.py
```diff
@@ -133,6 +133,36 @@ def test_assume_role(region, partition):
     )
 
 
+@freeze_time("2012-01-01 12:00:00")
```
bpandola: Is this needed? I don't see any datetime assertions...

sindrig: It probably isn't - I copied the scaffolding from another tests. Removing.

bpandola (approved):

PR #9224, opened 2025-08-26
Core: normalize request parsing for services using the AWS Query protocol

bpandola:
This PR opts all of the Query-protocol-based services in to the new request parsing pipeline.  The dedicated QueryParser obviates the need for many of the existing helper methods (e.g. `_get_multi_param()`, `_get_object_map()`, etc.), which have been replaced with the generic `_get_param()`.  Because the new parsing flow takes parameter type information from the Botocore models into account, some minor Moto model changes were also required (mostly around boolean values that were previously stored as strings).

The changeset for this PR is provided as a series of atomic commits, one for each affected service.

### DIFF
--- a/moto/autoscaling/models.py
+++ b/moto/autoscaling/models.py
@@ -1852,8 +1852,10 @@ class AutoScalingBackend(BaseBackend):
             if f["Name"] == "auto-scaling-group":
                 tags = [t for t in tags if t["ResourceId"] in f["Values"]]
             if f["Name"] == "propagate-at-launch":
-                values = [v.lower() for v in f["Values"]]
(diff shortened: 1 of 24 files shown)


bblommers (approved):
I love this! Being able to just use `_get_param()` makes this so much simpler to use :heart:

bpandola:
> I love this! Being able to just use `_get_param()` makes this so much simpler to use ❤️

Yeah, I think this is right approach for the moment.  I know there's been some talk of bringing in the Botocore models as typed Python classes--which, in principle, I'm not against--but I do feel that trying to make that work with the existing codebase would be a *TON* of work.  Hooking into `.get_param()` in this way provides the type guarantees of the AWS models with the least amount of friction.

PR #9230, opened 2025-08-27
[Glue] Basic workflows

BlizzardOf:
Fills out the functionality for workflows and workflow runs. There's a lot of work left to be done in order to make workflows actually consist of other glue resources linked together, but this is already pretty large and stands on its own, so I'd like to get it in first.

### DIFF
--- a/moto/glue/models.py
+++ b/moto/glue/models.py
@@ -1579,7 +1579,7 @@ class GlueBackend(BaseBackend):
     def get_workflow_runs(
         self,
         workflow_name: str,
-    ) -> List[Dict[str, Any]]:
+    ) -> List[Dict[str, Union[str, Dict[str, str]]]]:
         workflow = self.workflows.get(workflow_name)
         if not workflow:
             raise EntityNotFoundException("Entity not found")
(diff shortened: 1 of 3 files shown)


moto/glue/models.py
```diff
@@ -1497,6 +1510,120 @@ def delete_resource_policy(
...
+
+    def get_workflow(
+        self,
+        name: str,
+    ) -> Dict[str, Any]:
+        workflow = self.workflows.get(name)
+        if workflow:
+            return workflow.as_dict()
+        else:
+            raise EntityNotFoundException("Entity not found")
```
bblommers, line 1534: Just a nitpick - but I guess this can be shortened with the walrus operator:

```suggestion
        if workflow := self.workflows.get(name):
            return workflow.as_dict()
        raise EntityNotFoundException("Entity not found")
```

moto/glue/models.py
```diff
@@ -1497,6 +1510,120 @@ def delete_resource_policy(
...
+        return name
+
+    def get_workflow_run(
+        self,
+        workflow_name: str,
+        run_id: str,
+    ) -> Dict[str, Any]:
+        workflow = self.workflows.get(workflow_name)
+        if not workflow:
+            raise EntityNotFoundException("Entity not found")
```
bblommers, line 1575: Another nitpick - considering that this exception is always the same, would it make sense to extract it into a utility method, something like `_find_workflow(workflow_name)`?

moto/glue/models.py
```diff
@@ -1497,6 +1510,120 @@ def delete_resource_policy(
...
+        return workflow.get_run(run_id).as_dict()
+
+    @paginate(pagination_model=PAGINATION_MODEL)
```
bblommers, line 1578: Nice to see the pagination immediately available!

Issue #9195, opened 2025-08-19
Important Update - Amazon CloudWatch protocol changes in Public Preview will affect Moto


yasmewad:
Moto team,

We’re from the AWS SDKs team and are reaching out about an important upcoming change to Amazon CloudWatch that may impact this library and its users.

### What's changing

[Amazon CloudWatch is migrating from the AWS Query protocol](https://aws.amazon.com/about-aws/whats-new/2025/07/amazon-cloudwatch-sdk-optimized-json-cbor-protocols-in-preview/) to the more efficient [AWS JSON 1.0](https://smithy.io/2.0/aws/protocols/aws-json-1_0-protocol.html)  and [Smithy RPC v2 CBOR](https://smithy.io/2.0/additional-specs/protocols/smithy-rpc-v2.html) protocols. This change is currently available in public preview and will become the default behavior via a release in the coming months.

Based on our experience with similar protocol migrations, Moto appears to be affected by these protocol changes. 

### Testing Results

We tested Moto for CloudWatch with the boto3 preview SDK and encountered issues in **73 out of 76 tests**. The following is a brief breakdown of issues:

|Error Category|  Count	| Description	|
|---|---|---|
| Protocol Recognition | 69 |`KeyError: 'smithy-rpc-v2-cbor' ` |
| Parameter Parsing | 4 | `KeyError: 'StartTime'` ( likely due to change from AWS Query to AWS JSON data format) |
| Data Type Parsing | 3 | `TypeError: Parser must be a string or character stream` ( likely due to change from AWS Query to AWS JSON data format) |

We temporarily attempted a fix, adding an entry to the `SERIALIZERS` dictionary in `moto/core/serialize.py` that maps the `protocol` value of `smithy-rpc-v2-cbor` in the model for CloudWatch to JSONSerializer. 

```
"smithy-rpc-v2-cbor": JSONSerializer,
```

This isn’t the clear fix, as the new `protocols` field contains a list of `protocols` that the service supports, where a client will choose its most performant option. For boto3, this will resolve `json` with a `jsonVersion` of `1.0`. This [client priority order can be seen here](https://github.com/boto/botocore/blob/aa532424bd4ab40aefa38fdb715536ee545a31d3/botocore/utils.py#L194-L201) in botocore.

This temporary change fixed around **18 tests**, reducing the failure to **55 tests** with all of them appearing related to parsing issues with the new request / response format in AWS JSON 1.0. The following is a brief breakdown of issues:

[...]

PR #9253, opened 2025-09-01
CloudWatch: support Query/JSON protocols simultaneously

bpandola:
In order to ensure that Moto is ready for Botocore's upcoming switch from the Query protocol to the JSON protocol when communicating with CloudWatch (#9195), this PR adds a dedicated workflow to run the CloudWatch test suite using both protocols.  With these tests passing, Moto will be forward- and backward-compatible with the planned Botocore changes.

> [!WARNING]\
> Botocore is switching from the Query protocol to the JSON protocol for CloudWatch, but other SDKs (notably Golang 2.x) will be switching from the Query protocol to the RPC v2 CBOR protocol, for which there is no current (or planned) support in Moto.

It's worth pointing out that Moto is able to simultaneously support both of these protocols _without making a single modification to the CloudWatch backend_, which is remarkable when compared to the [SQS backend](https://github.com/getmoto/moto/blob/d5d203ddd556d88e394131e21d923dc15da16f16/moto/sqs/responses.py), which went through a similar protocol migration and required considerable effort from Moto's maintainer(s) to add support (#6331).  The CloudWatch backend being protocol-agnostic didn't come for free, as a significant amount of recent work has been dedicated to improving Moto's request parsing and response serialization pipeline (#9073), but this is exactly the kind of scenario to highlight the value of these higher-level abstractions: in this case, allowing Moto to quickly adapt to a major change from AWS with very little friction.

> [!TIP]\
> In order to test the new JSON requests/responses, the workflow is currently manually downloading and installing the [public preview version of Botocore](https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/cloudwatch-protocols-faq.html#getting-started) that implements the CloudWatch protocol changes.  Once these changes are included in a released version of Botocore, we should update the workflow (or simply merge with the [Outdated Dependency Tests](https://github.com/getmoto/moto/blob/d5d203ddd556d88e394131e21d923dc15da16f16/.github/workflows/test_outdated_versions.yml)).

### DIFF
--- a/.github/workflows/test_cloudwatch.yml
+++ b/.github/workflows/test_cloudwatch.yml
@@ -1,0 +1,33 @@
+name: "CloudWatch Protocol Tests"
+on:
+  pull_request:
+    paths:
+      - ".github/workflows/test_cloudwatch.yml"
+      - "moto/core/**"
+      - "moto/cloudwatch/**"
+      - "tests/test_cloudwatch/**"
+jobs:
+  test-cloudwatch-protocols:
+    name: ${{ matrix.botocore-version == 'botocore==1.40.16' && 'Query Protocol' || 'JSON Protocol' }}
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: [ "3.12" ]
+        botocore-version: [ "botocore-1.39.7-py3-none-any.whl", "botocore==1.40.16" ]
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v5
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v5
+        with:
+          python-version: ${{ matrix.python-version }}
+      - name: Install project dependencies
+        run: |
+          curl -L -o Boto3CliV1Artifacts.zip "https://d1l1rpjfz23h36.cloudfront.net/pr2qwos6sy8mfx2/Python_Boto_v3/Boto3CliV1Artifacts.zip"
+          unzip Boto3CliV1Artifacts.zip
+          python -m pip install --upgrade pip
+          pip install -r requirements-dev.txt
+          pip install ${{matrix.botocore-version}}
+      - name: Test CloudWatch
+        run: |
+          pytest -sv tests/test_cloudwatch


bblommers:
This is awesome - thank you so much for jumping on this and fixing it @bpandola !

yasmewad:
I had a question, this PR sets up automated test runs. Is there a separate PR for adding support for the JSON protocol? Are the tests passing? In our local testing the tests failed.

bblommers:
Hi @yasmewad, there have been a few PR's to add support for the JSON protocol, the most important ones were probably #9040 and #9139 

Can you explain a bit how you run the tests?

I don't know if it helps at all, but I've just created a dev-release that contains all these changes: https://pypi.org/project/moto/5.1.12.dev52/

yasmewad:
Thanks for those links. I was confused that this PR had only one file change. It makes sense now. Also, when I mentioned local testing, I meant the previous local testing that was used to raise issue #9195.

Thank you for getting this in quickly! Appreciate it!

yasmewad:
Hello @bblommers , I ran these tests locally on macos in a venv using the `5.1.12.dev52` version and found two failures in the `test_cloudwatch` tests. These seemed to be the comparison of `tzutc()` with what cloudwatch sends. However, these tests did not seem to be related to the JSON protocol being used now in the preview SDK. I just wanted to notify you about it.

Failing tests were:
1. [test_put_metric_alarm_with_anomaly_detection](https://github.com/getmoto/moto/blob/master/tests/test_cloudwatch/test_cloudwatch_boto3.py#L1784)
2. [test_put_metric_alarm_with_percentile](https://github.com/getmoto/moto/blob/master/tests/test_cloudwatch/test_cloudwatch_boto3.py#L1713)

The error was:
```
E       assert tzlocal() == tzutc()
E        +  where tzlocal() = datetime.datetime(2025, 9, 8, 14, 17, 9, tzinfo=tzlocal()).tzinfo
E        +  and   tzutc() = tzutc()
```

It would be great if those could be updated to run in all environments regardless of the timezone.

Thanks.

bpandola:
@yasmewad Hmm... that's odd.  The full test suite is [passing for us](https://github.com/getmoto/moto/actions/runs/17564989999).  I wonder what's different about your setup.  I did already fix a [similar issue](https://github.com/getmoto/moto/commit/11540e8a80382af1117960f6fc7b8c7e5a15c390#diff-b99c9ec92875d99c27ec8339366f81731f902ddb9a25d9f9a1b1e6cfa1ab56a6) as part of the work I did to support the JSON protocol for CloudWatch.  It's not clear to me why these tests are failing for you and not for us, but I'll "fix" them the same way I did the others.  Thanks for letting us know!

yasmewad:
Thanks for the prompt response. I just wanted to report what I found.

bpandola:
@yasmewad Fixed in #9281 - We didn't catch this with our dedicated CloudWatch protocol tests because the local time on GitHub Actions runners is Coordinated Universal Time (UTC).  Thanks again for letting us know!

yasmewad:
Yeah I was wondering the same, that most test environments use UTC. I just happen to run this locally. Thanks for such a quick fix and turn around. Appreciate the help!

PR #9259, opened 2025-09-02
Feature/ssm register/deregister/get patch group

Peeanio:
For SSM service, adds:
register_patch_baseline_for_patch_group
deregister_patch_baseline_for_patch_group
get_patch_baseline_for_patch_group


moto/ssm/models.py
```diff
@@ -1164,6 +1166,29 @@ def generate_id() -> str:
...
+            "SUSE": "arn:aws:ssm:us-west-2:280605243866:patchbaseline/pb-0aee740f9a480ec2e",
+            "CENTOS": "arn:aws:ssm:us-west-2:280605243866:patchbaseline/pb-0b641de5f3a9f3b2f",
+            "ORACLE_LINUX": "arn:aws:ssm:us-west-2:280605243866:patchbaseline/pb-06ac4861e0c6d047f",
+            "DEBIAN": "arn:aws:ssm:us-west-2:280605243866:patchbaseline/pb-04d1ad3cad30d44ff",
+            "MACOS": "arn:aws:ssm:us-west-2:280605243866:patchbaseline/pb-03dbdd88b851b829c",
+            "RASPBIAN": "arn:aws:ssm:us-west-2:280605243866:patchbaseline/pb-0bcede5146d0adbd4",
+            "ROCKY_LINUX": "arn:aws:ssm:us-west-2:280605243866:patchbaseline/pb-0b12e36f68dafa2ba",
+            "ALMA_LINUX": "arn:aws:ssm:us-west-2:280605243866:patchbaseline/pb-07dbd9f0b517b769e",
+            "AMAZON_LINUX_2023": "arn:aws:ssm:us-west-2:280605243866:patchbaseline/pb-0a624803d647da0ab",
+        }
```
bpandola: Is this your account number?  I think we'd need to bring these in with proper partition, region, and Moto account_id.  See my comment below for an example of how to do this.

moto/ssm/models.py
```diff
@@ -1189,6 +1214,7 @@ def __init__(self, region_name: str, account_id: str):
 
         self.windows: Dict[str, FakeMaintenanceWindow] = dict()
         self.baselines: Dict[str, FakePatchBaseline] = dict()
+        self.patch_groups: Dict[str, FakePatchGroup] = dict()
         self.ssm_prefix = (
             f"arn:{self.partition}:ssm:{self.region_name}:{self.account_id}:parameter"
         )
```
bpandola, line 1214: There's a separate backend for each partition, region, and account.  This is an example of how to make the ARNs work across any backend.

moto/ssm/models.py
max-russ: That's not my account number, that's Amazon's account where they store the default baselines. It appears that the account id itself is maybe region specific, but here's the only public resource I could find mentioning that account id specifically. Will look at other regions and their account Ids tomorrow.
https://medium.com/@george.fridman/navigating-the-challenges-of-aws-cloudformation-and-ssm-a-journey-19b0699ceab9

bpandola: Check out the `moto/scripts` directory.  There are a number of scripts (`ssm*.py`) that pull in data from various regions for SSM. Maybe something like is needed here?

max-russ: Thanks for the pointer on the scripts. I made a new one to grab every SSM region and dump into a json file with the patch baselines per region, which should also then set the partition and baseline id correctly. My environment limits most regions, so I can't generate a full list of baselines for every region, but that should be all that's needed to get the tests to pass.

max-russ: Added the scrapped default baselines.

moto/ssm/models.py
max-russ: I'm unsure how to access the other partitions in the script to scrape out those arns, hoping that the script + default is good enough there.

bpandola (approved):
@Peeanio Thanks!

@bblommers This has my approval as-is, but wanted to loop you in (because of your work on #5924) to determine if it's worth making a workflow out of the new script.  (Maybe not if the data doesn't change very often.)

PR #9264, opened 2025-09-04
[Macie]: Adding support for Macie v2

ghorondo:
This PR introduces mock support for AWS Macie v2

Implemented API Calls:

- CreateInvitations
- ListInvitations
- AcceptInvitation
- DeclineInvitations
- ListMembers
- GetAdministratorAccount
- DeleteMember

### DIFF
--- a/moto/macie2/models.py
+++ b/moto/macie2/models.py
@@ -1,0 +1,170 @@
+from datetime import datetime
+from typing import Any, Dict, List, Optional
+
+from moto.core.base_backend import BackendDict, BaseBackend
+from moto.core.common_models import BaseModel
+from moto.moto_api._internal import mock_random
+
+from .exceptions import ResourceNotFoundException
+
+
+class Invitation(BaseModel):
+    def __init__(self, account_id: str, region_name: str, admin_account_id: str):
+        self.account_id = account_id
+        self.invitation_id = mock_random.get_random_hex()
+        self.invited_at = datetime.utcnow()
+        self.relationship_status = "Invited"
+        self.arn = f"arn:aws:macie2:{region_name}:{admin_account_id}:invitation/{self.invitation_id}"
+
+    def to_json(self) -> Dict[str, Any]:
+        return {
+            "accountId": self.account_id,
+            "invitationId": self.invitation_id,
+            "invitedAt": self.invited_at.isoformat(),
+            "relationshipStatus": self.relationship_status,
+        }
+
+
+class Member(BaseModel):
+    def __init__(
+        self,
+        account_id: str,
+        region_name: str,
+        admin_account_id: str,
+        invitation: Invitation,
+    ):
+        self.account_id = account_id
+        self.relationship_status = "Enabled"
+        self.updated_at = datetime.utcnow()
+        self.arn = (
+            f"arn:aws:macie2:{region_name}:{admin_account_id}:member/{self.account_id}"
+        )
+        self.administrator_account_id = admin_account_id
+        self.invited_at = invitation.invited_at
+
+    def to_json(self) -> Dict[str, Any]:
+        return {
+            "accountId": self.account_id,
+            "administratorAccountId": self.administrator_account_id,
+            "arn": self.arn,
+            "email": f"{self.account_id}@example.com",  # Moto does not have access to real emails
+            "invitedAt": self.invited_at.isoformat(),
+            "masterAccountId": self.administrator_account_id,  # masterAccountId is deprecated but still present
+            "relationshipStatus": self.relationship_status,
+            "updatedAt": self.updated_at.isoformat(),
+        }
+
+
+class MacieBackend(BaseBackend):
+    def __init__(self, region_name: str, account_id: str):
+        super().__init__(region_name, account_id)
+        self.invitations: Dict[str, Invitation] = {}
+        self.members: Dict[str, Member] = {}
+        self.administrator_account: Optional[Member] = None
+        self.macie_session: Optional[Dict[str, Any]] = {
+            "createdAt": datetime.utcnow(),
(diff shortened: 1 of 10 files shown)


moto/macie2/models.py
```diff
@@ -0,0 +1,170 @@
...
+
+    def create_invitations(self, account_ids: List[str]) -> List[Dict[str, Any]]:
+        for account_id in account_ids:
+            invitation = Invitation(
+                account_id=account_id,
+                region_name=self.region_name,
+                admin_account_id=self.account_id,
+            )
+            self.invitations[account_id] = invitation
+        return []
```
bblommers: **nitpick**: If there is nothing to return, we should just return nothing - makes it a little easier to read
```suggestion
```

moto/macie2/models.py
```diff
@@ -0,0 +1,170 @@
...
+
+    def list_invitations(self) -> List[Invitation]:
+        return list(self.invitations.values())
+
+    def decline_invitations(self, account_ids: List[str]) -> List[Dict[str, Any]]:
+        for backend_dict in macie2_backends.values():
+            backend = backend_dict.get(self.region_name)
+            if backend and self.account_id in backend.invitations:
+                backend.invitations.pop(self.account_id)
+        return []
```
bblommers: Same here

moto/macie2/models.py
```diff
@@ -0,0 +1,170 @@
...
+                region_name=self.region_name,
+                admin_account_id=self.account_id,
+            )
+            self.invitations[account_id] = invitation
+        return []
+
+    def list_invitations(self) -> List[Invitation]:
+        return list(self.invitations.values())
+
+    def decline_invitations(self, account_ids: List[str]) -> List[Dict[str, Any]]:
```
bblommers: The user-provided `account_ids` aren't used here.

I'm not familiar with Macie at all - but we should we be declining the `account_ids` that the user provided? We currently always decline `self.account_id`

moto/macie2/models.py
```diff
@@ -0,0 +1,170 @@
...
+                backend.invitations.pop(self.account_id)
+        return []
+
+    def accept_invitation(
+        self, administrator_account_id: str, invitation_id: str
+    ) -> None:
+        for backend_dict in macie2_backends.values():
+            backend = backend_dict.get(self.region_name)
+            if not backend:
+                continue
```
bblommers, line 100:
```suggestion
        # Loop through every account available
        for backend_dict in macie2_backends.values():
            # Invitations can only be send to the same region
            # If the region does not exist, then the user hasn't used Macie in this account yet, and we can safely skip it
            backend = backend_dict.get(self.region_name)
            if not backend:
                continue
```
**nitpick**: It would be nice to have a small comment here that explains what we're doing here, and why we're selecting the region.

The comment that I added is just a suggestion, what made sense based on how I'm understanding the implementation - but feel free to adjust/change as you see fit :slightly_smiling_face:

tests/test_macie/test_macie.py
```diff
@@ -0,0 +1,224 @@
...
+
+    admin_client.create_invitations(accountIds=[MEMBER_ACCOUNT_ID])
+
+    response = admin_client.list_invitations()
+    assert len(response["invitations"]) == 1
+    assert response["invitations"][0]["accountId"] == MEMBER_ACCOUNT_ID
+
+
+@mock_aws(config={"macie2": {"initial_account_id": ADMIN_ACCOUNT_ID}})
+def test_accept_invitation():
```
bblommers: **suggestion**: It looks like the functionality of this test is already covered in `test_list_members`? As far as I'm concerned, we can just remove this one. Having more tests is always good, but there is no point in duplicating them

tests/test_macie/test_macie.py
```diff
@@ -0,0 +1,224 @@
...
+from botocore.exceptions import ClientError
+
+from moto import mock_aws
+
+ADMIN_ACCOUNT_ID = "111111111111"
+MEMBER_ACCOUNT_ID = "222222222222"
+REGION = "us-east-1"
+
+
+@mock_aws(config={"macie2": {"initial_account_id": ADMIN_ACCOUNT_ID}})
```
bblommers: **question**: This configuration doesn't seem to be used - is that still necessary?

tests/test_macie/test_macie.py
```diff
@@ -0,0 +1,224 @@
...
+    assert response["ResponseMetadata"]["HTTPStatusCode"] == 200
+
+
+@mock_aws(config={"macie2": {"initial_account_id": ADMIN_ACCOUNT_ID}})
+def test_decline_invitations():
+    sts = boto3.client("sts", region_name=REGION)
+    member_identity = sts.assume_role(
+        RoleArn=f"arn:aws:iam::{MEMBER_ACCOUNT_ID}:role/my-role",
+        RoleSessionName="test-session",
+    )["Credentials"]
```
bblommers, line 78: **praise**: Nice to see cross-account tests here!

bblommers (changes requested):
Hi @ghorondo, thanks for the PR! Please see my comments

moto/macie2/models.py
rondodev: Fixed!

tests/test_macie/test_macie.py
rondodev: Modified the tests to use @mock_aws() decorator

tests/test_macie/test_macie.py
rondodev: Removed test

moto/macie2/models.py
rondodev: Updated to properly iterate through the account_ids parameter

bblommers (approved):
Great - thanks again for the PR @ghorondo!

PR #9265, opened 2025-09-04
[DirectConnect] Updating ARNs to match AWS

LeekSwan:
This PR looks to address the issue with DirectConnect resources not having proper arn format. Some small changes were made using this [doc](https://docs.aws.amazon.com/service-authorization/latest/reference/list_awsdirectconnect.html#awsdirectconnect-dxcon:~:text=arn%3A%24%7BPartition%7D%3Adirectconnect%3A%24%7BRegion%7D%3A%24%7BAccount%7D%3Adxcon/%24%7BConnectionId%7D) as reference for how ARN should be structured.

### DIFF
--- a/moto/directconnect/models.py
+++ b/moto/directconnect/models.py
@@ -68,7 +68,7 @@ class Connection(BaseModel):
 
     def __post_init__(self) -> None:
         if self.connection_id == "":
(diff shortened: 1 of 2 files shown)


bblommers (approved):
Makes sense! Thanks for updating this @LeekSwan - having closer parity with AWS is always good

PR #9272, opened 2025-09-08
feat: Handle empty S3 lifecycle configuration

ego93:
This PR fixes the issue when an empty lifecycle configuration (which is used to delete all lifecycle rules). Moto had a bug in the XML parsing logic that expected a Rule key even when the lifecycle configuration was empty.

A test has been added `Test that empty lifecycle configuration (no rules) works correctly` 

IMPLEMENTATION_COVERAGE.md has also been updated to mention support for `put_bucket_lifecycle_configuration` and `get_bucket_lifecycle_configuration`

https://github.com/getmoto/moto/issues/9271


bblommers (approved):
Looks good to me - thanks for the PR @ego93!

PR #9273, opened 2025-09-08
fix(cognitoidp): aws.cognito.signin.user.admin scope in access token

bdellegrazie:
Include the default `aws.cognito.signin.user.admin` scope in access token. This scope is included by default in all access tokens created by `AdminInitiateAuth` and `InitiateAuth`, however it can be stripped by a custom pre-token-generation lambda trigger.

Ref: https://docs.aws.amazon.com/cognito/latest/developerguide/amazon-cognito-user-pools-using-the-access-token.html


bdellegrazie:
rebased against master

PR #9274, opened 2025-09-08
ELB: Allow secondary_ips.auto_assigned.per_subnet for modify_load_balancer_attributes

nik-localstack:
This is a supported attribute for NLBs according to https://docs.aws.amazon.com/cli/latest/reference/elbv2/modify-load-balancer-attributes.html


bblommers (approved):
LGTM! Thanks for fixing this upstream @nik-localstack

PR #9289, opened 2025-09-14
[WorkSpaces] Add terminate_workspaces and remove EnableWorkDocs deprecated field

ApogeeAC:
Added `terminate_workspaces` and supporting unit test.
Removed the `EnableWorkDocs` parameter as it is no longer supported in Boto3. Since AWS has also ended support for WorkDocs, this field is no longer required


tests/test_workspaces/test_workspaces.py
```diff
@@ -268,26 +266,42 @@ def test_describe_workspaces_only_user_name_used():
 
 
 @mock_aws
-@pytest.mark.parametrize("enable_work_docs", [True, False])
-def test_register_workspace_directory_with_workdocs(enable_work_docs):
-    # The WorkDocs-service has been deprecated:
-    # https://www.reddit.com/r/aws/comments/1cd1iqb/workdocsamazon_has_decided_to_end_support_for_the/
-    # Botocore 1.38.13 has removed support for the EnableWorkdocs-parameter,
```
bblommers, line 275: This version of botocore was released on May 9th, 2025, so almost 5 months ago - makes sense to remove it now IMHO

bblommers (approved):
LGTM!

PR #9291, opened 2025-09-16
[Athena] list capacity reservation endpoint support

LeekSwan:
This PR adds suppport for the following endpoints:

- [list_capacity_reservation](https://docs.getmoto.org/en/latest/docs/services/athena.html#:~:text=list_capacity_reservations)
- [update_capacity_reservation](https://docs.getmoto.org/en/latest/docs/services/athena.html#:~:text=update_capacity_reservation)


bblommers (approved):
LGTM! Thanks @LeekSwan

PR #9293, opened 2025-09-18
Remove QLDB

viren-nadkarni:
AWS retired Quantum Ledger Database (QLDB) on 2025-07-31. It is no longer accessible on AWS and the docs have also been erased.

This PR removes QLDB from Moto.


bblommers:
Looks like the day has come! Thanks again for opening the PR @viren-nadkarni :+1: 

https://github.com/boto/botocore/pull/3583

PR #9296, opened 2025-09-19
SNS: publish_batch: Use MessageAttributes provided by user

### LINKED ISSUE #9292: SNS message attribute keys converted to snake_case in v5.1.12 when consumed from SQS
After I upgraded from moto `v5.1.11` to `v5.1.12`, batch published SNS message attribute keys published to an SQS queue appear to be converted to snake-case instead of the original camel-case when consumed from the queue. Any ideas what might have caused this? Only `sns.publish_batch` shows this behaviour, not `sns.publish`.

```
python: 3.12
boto3: 1.40.31
moto: 5.1.12
```

Example simplified code below for reproducing:

```python
from moto import mock_aws
import boto3
import json
import uuid

@mock_aws
def test_snake_case_conversion_issue():
    sns = boto3.client("sns")
    sqs = boto3.resource("sqs")

    topic_arn = sns.create_topic(Name="topic")["TopicArn"]
    queue = sqs.create_queue(QueueName="queue")
    queue_arn = sqs.meta.client.get_queue_attributes(
        QueueUrl=queue.url, AttributeNames=["QueueArn"]
    )["Attributes"]["QueueArn"]

    sns.subscribe(
        TopicArn=topic_arn,
        Protocol="sqs",
        Endpoint=queue_arn,
    )

    sns.publish_batch(
        TopicArn=topic_arn,
        PublishBatchRequestEntries=[
            {
                "Id": str(uuid.uuid4()),
                "Message": json.dumps(item),
                "MessageAttributes": {
                    "SomeCamelCaseAttr": {"DataType": "String", "StringValue": "value"}
                },
            }
            for item in [{"type": "test"}]
        ],
    )

    msg = queue.receive_messages()[0]
    body = json.loads(msg.body)

    # the assertion fails in moto v5.1.12 but passes in moto v5.1.11
    assert body["MessageAttributes"] == {
        "SomeCamelCaseAttr": {"Type": "String", "Value": "value"}
    }
```

bblommers:
Fixes #9292 

We automatically convert incoming request parameters into snake-case - but this is not desired when we're requesting user-provided dictionaries.

This PR adds an escape hatch, so that we can determine per indidividual parameter whether we want the normalized value, or the original value that the user provided.


bblommers:
FYI @bpandola. I'm not sure whether this is the best fix - but at least it resolves the issue for now.

bpandola:
@bblommers Thanks for taking care of this.  I think, ultimately, I'm going to get rid of the `XFormedDict`.  It originated with the initial work I did with the RDS backend, which has a lot of long parameter lists that get passed to backend methods that use snake_case parameter names along with a `**kwargs` catch-all, but it's more of a liability outside of RDS (as this issue attests).  Being more explicit when retrieving request parameters in `responses.py` and explicitly specifying all needed parameters in the `models.py` method definitions for RDS would not only be more clear but would likely eliminate the need for this specialized dictionary.  Anyway... again, I appreciate you stepping in to fix this issue.  I'll re-evaluate `XFormedDict` to determine if the pitfalls outweigh the benefits moving forward.

PR #9307, opened 2025-09-25
[Athena] Updating ARNs to match AWS 

LeekSwan:
This PR looks to address the issue with Athena resources not having proper arn format. Some small changes were made using this [doc](https://docs.aws.amazon.com/service-authorization/latest/reference/list_amazonathena.html#:~:text=arn%3A%24%7BPartition%7D%3Aathena%3A%24%7BRegion%7D%3A%24%7BAccount%7D%3Adatacatalog/%24%7BDataCatalogName%7D) as reference for how ARN should be structured.

### DIFF
--- a/moto/athena/models.py
+++ b/moto/athena/models.py
@@ -59,7 +59,6 @@ class WorkGroup(TaggableResourceMixin, BaseModel):
             f"workgroup/{name}",
             tags,
         )
-        self.arn = f"arn:{get_partition(self.region_name)}:athena:{self.region_name}:{athena_backend.account_id}:workgroup/{name}"
         self.athena_backend = athena_backend
         self.name = name
         self.description = description
@@ -97,7 +96,6 @@ class DataCatalog(TaggableResourceMixin, BaseModel):
             f"datacatalog/{name}",
             tags,
         )
-        self.arn = f"arn:{get_partition(self.region_name)}:athena:{self.region_name}:{athena_backend.account_id}:datacatalog/{name}"
         self.athena_backend = athena_backend
         self.name = name
         self.type = catalog_type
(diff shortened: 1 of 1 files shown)


moto/athena/models.py
```diff
@@ -59,6 +59,7 @@ def __init__(
             f"workgroup/{name}",
             tags,
         )
+        self.arn = f"arn:{get_partition(self.region_name)}:athena:{self.region_name}:{athena_backend.account_id}:workgroup/{name}"
```
bblommers: As far as I can see, we already create the ARN in the `TaggableResourceMixin`, and in exactly the same format? So this doesn't seem necessary

moto/athena/models.py
```diff
@@ -96,6 +97,7 @@ def __init__(
             f"datacatalog/{name}",
             tags,
         )
+        self.arn = f"arn:{get_partition(self.region_name)}:athena:{self.region_name}:{athena_backend.account_id}:datacatalog/{name}"
```
bblommers: Same with this one

moto/athena/models.py
```diff
@@ -157,6 +159,7 @@ def __init__(
             f"capacityreservation/{name}",
             tags,
         )
+        self.arn = f"arn:{get_partition(self.region_name)}:athena:{self.region_name}:{athena_backend.account_id}:capacity-reservation/{name}"
```
bblommers:
```suggestion
            f"capacity-reservation/{name}",
            tags,
        )
```

This does the same thing, as far as I can see - adjusting the `resource-name` automatically creates the ARN in the correct format 🙂

LeekSwan: Your absolutely right! Ive made the changes. Tyvm!

bblommers (approved):
Great - thank you @LeekSwan!

PR #9341, opened 2025-10-06
ACM: Handle Exportable Public Certificates

SeanHood:
AWS recently added support for exportable public certificates: https://aws.amazon.com/blogs/aws/aws-certificate-manager-introduces-exportable-public-ssl-tls-certificates-to-use-anywhere/

Docs: https://docs.aws.amazon.com/acm/latest/userguide/acm-exportable-certificates.html

### DIFF
--- a/moto/acm/models.py
+++ b/moto/acm/models.py
@@ -129,6 +129,7 @@ class CertBundle(BaseModel):
         cert_type: str = "IMPORTED",
         cert_status: str = "ISSUED",
         cert_authority_arn: Optional[str] = None,
+        cert_options: Optional[Dict[str, Any]] = None,
     ):
         self.created_at = utcnow()
         self.cert = certificate
@@ -140,6 +141,10 @@ class CertBundle(BaseModel):
         self.status = cert_status  # Should really be an enum
         self.cert_authority_arn = cert_authority_arn
         self.in_use_by: List[str] = []
+        self.cert_options = cert_options or {
+            "CertificateTransparencyLoggingPreference": "ENABLED",
+            "Export": "DISABLED",
+        }
 
         # Takes care of PEM checking
         self._key = self.validate_pk()
@@ -377,7 +382,7 @@ class CertBundle(BaseModel):
                 "Type": self.type,  # One of IMPORTED, AMAZON_ISSUED,
                 "ExtendedKeyUsages": [],
                 "RenewalEligibility": "INELIGIBLE",
-                "Options": {"CertificateTransparencyLoggingPreference": "ENABLED"},
+                "Options": self.cert_options,
             }
         }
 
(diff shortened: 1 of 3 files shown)


moto/acm/models.py
```diff
@@ -129,6 +129,10 @@ def __init__(
         cert_type: str = "IMPORTED",
         cert_status: str = "ISSUED",
         cert_authority_arn: Optional[str] = None,
+        cert_options: dict = {
```
johnnymercado: You should avoid mutable values as defaults. While not an issue in the current state, this can be modified down the road and it can start to change value if it's modified at any point in this method. For example:
```
>>> def example(a={"test": "test1"}):
...     print(a)
...     a['test2'] = 'test3'
... 
>>> example()
{'test': 'test1'}
>>> example()
{'test': 'test1', 'test2': 'test3'}
```

moto/acm/models.py
```diff
@@ -556,6 +561,7 @@ def request_certificate(
         subject_alt_names: List[str],
         tags: List[Dict[str, str]],
         cert_authority_arn: Optional[str] = None,
+        cert_options: dict = {},
```
johnnymercado: Same here

moto/acm/models.py
```diff
@@ -129,6 +129,10 @@ def __init__(
         cert_type: str = "IMPORTED",
         cert_status: str = "ISSUED",
         cert_authority_arn: Optional[str] = None,
+        cert_options: Dict[str, Any] = {
```
bblommers: We should default to None here, as mutable defaults are dangerous

moto/acm/models.py
```diff
@@ -140,6 +144,7 @@ def __init__(
         self.status = cert_status  # Should really be an enum
         self.cert_authority_arn = cert_authority_arn
         self.in_use_by: List[str] = []
+        self.cert_options = cert_options
```
bblommers: The default should be used here instead:


```suggestion
        self.cert_options = cert_options or {
            "CertificateTransparencyLoggingPreference": "ENABLED",
            "Export": "DISABLED",
        }
```

moto/acm/models.py
```diff
@@ -526,11 +531,17 @@ def import_certificate(
...
+    def list_certificates(
+        self, statuses: List[str], includes: Dict[str, Any]
+    ) -> Iterable[CertBundle]:
         for arn in self._certificates.keys():
             cert = self.get_certificate(arn)
             if not statuses or cert.status in statuses:
-                yield cert
+                if (
+                    not includes
+                    or includes.get("exportOption", None) in cert.cert_options["Export"]
```
bblommers: We never guarantee that `Export` is part of `cert.cert_options`, as far as I can tell, so we should guard against that as well.

Something like:
```python
if not includes or (
    "Export" in cert.cert_options and includes.get("exportOption", None) in cert.cert_options["Export"]
   )
```
should do, I think? Open to suggestions here if you have a better idea.

moto/acm/responses.py
```diff
@@ -129,10 +129,14 @@ def import_certificate(self) -> str:
...
         statuses = self._get_param("CertificateStatuses")
-        for cert_bundle in self.acm_backend.list_certificates(statuses):
+        includes = self._get_param("Includes")
+        for cert_bundle in self.acm_backend.list_certificates(statuses, includes):
             _cert = cert_bundle.describe()["Certificate"]
             _in_use_by = _cert.pop("InUseBy", [])
             _cert["InUse"] = bool(_in_use_by)
+            _cert["Exported"] = bool(
+                True if cert_bundle.cert_options["Export"] == "ENABLED" else False
+            )
```
bblommers: The `==` operator already ensures that the outcome is a boolean, so if we simplify this, it essentially says:

`bool(True if True else False)`


There is a simpler way to achieve the same thing:


```python
 _cert["Exported"] = cert_bundle.cert_options["Export"] == "ENABLED"
```

SeanHood:
@bblommers Thank you for these suggestions and the explanations, they make a lot more sense now and the diff cleaner too. I've added them in and re-tested.

PR #9351, opened 2025-10-08
[SNS] Adding config support for SNS topics

ghorondo:
This PR adds Config support for SNS topics, allowing users to list, track, and query topic configurations. This enhancement implements the necessary backend methods for config integration, including list_config_service_resources, get_config_resource, and select_resource_config.

### DIFF
--- a/tests/test_core/test_mypy.py
+++ b/tests/test_core/test_mypy.py
@@ -35,8 +35,9 @@ def test_manual() -> None:
     m.stop()
 
 
-x: int = method_with_parentheses()
-assert x == 456
+def test_mock_aws_decorator_return_types() -> None:
+    x: int = method_with_parentheses()
+    assert x == 456
 
-y: int = method_without_parentheses()
-assert y == 123
+    y: int = method_without_parentheses()
+    assert y == 123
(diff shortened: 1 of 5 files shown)


moto/sns/models.py
```diff
@@ -1238,6 +1239,178 @@ def opt_in_phone_number(self, number: str) -> None:
...
+            topics = [t for t in topics if t.arn in resource_ids]
+
+        if resource_name:
+            topics = [t for t in topics if t.name == resource_name]
+
+        start = int(next_token) if next_token else 0
+        end = start + limit
+        topics_page = topics[start:end]
+
+        new_next_token = str(end) if end < len(topics) else None
```
bblommers: Can we use the builtin pagination utility here, `@paginate`? That way we can avoid having to paginate manually here

Example implementation: https://github.com/getmoto/moto/blob/3565ee532ce17d64d4250d258ae0a13e46be8431/moto/amp/models.py#L122

And the docs: https://docs.getmoto.org/en/latest/docs/contributing/development_tips/utilities.html#paginator

rondodev: Done!

PR #9358, opened 2025-10-11
Admin: fix error shape not found warnings

bpandola:
> tests/test_sesv2/test_sesv2.py::test_get_configuration_set
tests/test_sesv2/test_sesv2.py::test_delete_configuration_set
  /home/runner/work/moto/moto/moto/core/serialize.py:401: UserWarning: SESv2 service model does not contain an error shape that matches code ConfigurationSetDoesNotExist from Exception(ConfigurationSetDoesNotExist)
    warnings.warn(warning)

These warnings are not critical, but had gotten a bit noisy lately now that they are visible as part of the "check annotations" on Pull Requests.  In addition to updating a couple of incorrect error codes, this PR also adds the ability to supplement AWS service models with Moto "extras" at load time.  This mechanism is used here to add an entry to the SESv2 service model that maps a SESv1 error message to its newer counterpart, and there's a lot of potential for other useful applications.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -18,6 +18,7 @@ include moto/rds/resources/cluster_options/*.json
 include moto/ssm/resources/*.json
 include moto/ssm/resources/ami-amazon-linux-latest/*.json
 include moto/ssm/resources/ecs/optimized_amis/*.json
+include moto/sesv2/*/*.json
 include moto/support/resources/*.json
 recursive-include moto/moto_server *
 recursive-include tests *
--- a/moto/ses/models.py
+++ b/moto/ses/models.py
@@ -711,7 +711,7 @@ class SESBackend(BaseBackend):
     ) -> ConfigurationSet:
         if configuration_set_name not in self.config_sets:
             raise ConfigurationSetDoesNotExist(
-                f"Configuration set <{configuration_set_name}> does not exist"
+                f"Configuration set <{configuration_set_name}> does not exist."
             )
         return self.config_sets[configuration_set_name]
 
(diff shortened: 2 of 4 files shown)


moto/core/loaders.py
```diff
@@ -0,0 +1,29 @@
...
+
+from moto import MOTO_ROOT
+
+
+class Loader(BotocoreLoader):
+    BUILTIN_EXTRAS_TYPES = ["moto"]
+
+
+def create_loader() -> Loader:
+    return Loader(extra_search_paths=[MOTO_ROOT])
```
bblommers, line 29: @bpandola This is really cool! Nice to be able to use the builtin mechanisms to achieve this.

Just a nitpick: do you think we can achieve this without adding the `MOTO_ROOT`? Just because it adds to the public API, and it's not necessarily something that is useful for users.

If I'm not mistaken, this would achieve the same thing:
```python
import moto

..

def create_loader() -> Loader:
    return Loader(extra_search_paths=[os.path.dirname(os.path.abspath(moto.__file__))])
```

bpandola: Definitely not opposed if you feel strongly about it.  I just did it the way [Botocore does it](https://github.com/boto/botocore/blob/8d33050a8d49901b780602488aca77a392e32dc6/botocore/__init__.py#L140).  Apologies, I didn't consider the ramifications of modifying the main `__init__` file.  Let me know if you want want me to revert it.

bblommers: All good! It may not become a problem, but I want to avoid situations where others start using this variable (and why shouldn't they - it's part of the public API!), and we can't change it anymore.

Relevant XKCD:

bblommers: I've raised #9363 👍

PR #9370, opened 2025-10-14
CI: Test Moto against Python 3.14

bblommers:


bblommers:
This might be beginners luck - but the 3.14 tests seem to be significantly faster:

`3.11`: 18 mins
`3.12`: 18 mins
`3.13`: 17 mins
`3.14`: 13 mins

This is consistent across two runs as well, so this is very promising 🥳

PR #9373, opened 2025-10-15
SQS: Protocol Test Improvements

bpandola:
* Add proper request headers for server mode tests
* Add dedicated Query protocol test for creating queue with tags
* Assert correct error code serialization for QueueNotFound, regardless of wire protocol (JSON or Query)


bpandola:
The missing line reported by CodeCov is covered by the dedicated [SQS Protocol Tests](https://github.com/getmoto/moto/blob/bd787730867483c2b593bfc65f4c34c2dadd4af6/.github/workflows/test_sqs.yml).

PR #9375, opened 2025-10-16
IdentityStore: Add implementation for get-user-id

pcjun97:
This PR adds support for the `get_user_id` operation in the IdentityStore service. This allows retrieving a user's UserId by alternate identifiers such as username or email address.

From AWS document for `list-users`: https://docs.aws.amazon.com/cli/latest/reference/identitystore/list-users.html#description
>  Filtering for a User by the UserName attribute is deprecated. Instead, use the GetUserId API action.

ExternalId alternate identifier is not yet implemented (consistent with `get_group_id`).

### DIFF
--- a/moto/identitystore/models.py
+++ b/moto/identitystore/models.py
@@ -220,6 +220,38 @@ class IdentityStoreBackend(BaseBackend):
 
         return user_id, identity_store_id
 
+    def get_user_id(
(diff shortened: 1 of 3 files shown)


bblommers (approved):
LGTM! Nice to see such an extensive test suite 👍

PR #9376, opened 2025-10-16
Core: deprecate and limit use of `XFormedDict`

bpandola:
The [XFormedDict](https://github.com/getmoto/moto/blob/85f9ee0e352d0d9e10852da3c3e96ac724af709c/moto/core/parsers.py#L258) was created to aid in translating AWS request input keys serialized in either `PascalCase` or `camelCase` into the more Pythonic `snake_case`.  This was done implicitly, meaning that either the original key or the "xformed" key could be used interchangeably to access dictionary values (e.g. `xformed_dict["DBInstanceIdentifier"] == xformed_dict["db_instance_identifier"]`).  

XFormedDict was originally created for and used with the RDS backend--which has lots of long, mostly flat lists of input parameters being passed to backend methods and constructors that accept `**kwargs`--but was later made the default data structure for all backends that were opted in to the new request parsing pipeline (#9224).

Although initially quite helpful in obviating the need for a lot of boilerplate code on the RDS backend, over time it's become more of a liability outside of its original use-case.  Input that contains arbitrary user-provided dictionaries, for example, is one scenario where having the dictionary keys silently modified under the hood is not desirable (see #9292).  It's also debatable whether or not there's any value in being able to write `input_map["param_key"]` instead of `input_map["ParamKey"]`.  Being able to use the `**` operator to unpack an `XFormedDict` into (properly `snake_cased`) keyword arguments is undeniably useful, but explicit is better than implicit (as the Zen of Python states) and we really should be explicitly passing arguments to well-defined method signatures.

All this to say: it's time to get rid of `XFormedDict`.  This PR removes it from all backends except for EMR and RDS, where it is so deeply embedded that it will take a lot more effort (in another PR) to be rid of it entirely.

> [!WARNING]
> There is some chance of regression here, although it is impossible to quantify exactly because of the implicit nature of the `XFormedDict`.  This is actually another argument in favor of its removal from the codebase.  Any potential impact should be easily addressed and (hopefully) minimal.

### DIFF
--- a/moto/applicationautoscaling/models.py
+++ b/moto/applicationautoscaling/models.py
@@ -503,7 +503,7 @@ class FakeApplicationAutoscalingPolicy(BaseModel):
             threshold=42.0,
             statistic="Sum",
             description=f"DO NOT EDIT OR DELETE. For TargetTrackingScaling policy {alarm_action}",
-            dimensions=[{"name": "TableName", "value": table_name}],
+            dimensions=[{"Name": "TableName", "Value": table_name}],
             alarm_actions=[alarm_action],
         )
         alarms.append(alarm1)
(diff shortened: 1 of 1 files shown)


bblommers (approved):
LGTM! Makes sense to reduce the scope of the XFormedDict 👍

Issue #7054, opened 2023-11-22
SQS Response Protocol Migrated from XML to JSON


mkeemon:
Moto (installed via Docker, pip)
- Python version: 3.12.0
- Moto Version: 4.2.9

AWS Ruby SDK
- aws-sdk-core: 3.187.1
- aws-sdk-sqs: 1.67.0

According to AWS documentation, the most recent version of their SDKs utilize JSON instead of XML for SQS responses. It appears this is a backward incompatible change, so migrating to the most recent version of the various AWS SDKs will cause cause issues with Moto.

https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-json-faqs.html

Tests are failing in the [ruby-shoryuken/shoryuken](https://github.com/ruby-shoryuken/shoryuken/pull/753) repository due to the [XML responses](https://github.com/getmoto/moto/blob/b3c3883a78f7dc096984a06e06112c31e96c2e06/moto/sqs/responses.py#L746) in Moto

[Action run in](https://github.com/ruby-shoryuken/shoryuken/actions/runs/6946610660/job/18904539378?pr=753) ruby-shoryuken/shoryuken: 
```
JSON::ParserError:
          #   859: unexpected token at '<CreateQueueResponse><CreateQueueResult><QueueUrl>http://localhost:5000/123456789012/shoryuken-travis-StandardWorker-455c8b98-cade-4220-a7cd-8b9e5bdfed27</QueueUrl></CreateQueueResult><ResponseMetadata><RequestId>F4FSDWJ3GAHTAWXHEQ11QOW56ZOVVDXRLHTUKZLB10T1XSI17DSA</RequestId></ResponseMetadata></CreateQueueResponse>'
          #   ./vendor/bundle/ruby/3.1.0/gems/aws-sdk-core-3.187.1/lib/aws-sdk-core/json/json_engine.rb:8:in `load'

     2.2) Failure/Error:
            queue_url = Shoryuken::Client.sqs.get_queue_url(
              queue_name: StandardWorker.get_shoryuken_options['queue']
            ).queue_url

          Aws::SQS::Errors::NotFound:
            Aws::SQS::Errors::NotFound
          # ./vendor/bundle/ruby/3.1.0/gems/aws-sdk-core-3.187.1/lib/seahorse/client/plugins/raise_response_errors.rb:17:in `call'
[...]

PR #9386, opened 2025-10-19
SQS: integrate with core request parser/response serializer

### LINKED ISSUE #9331: [sqs] weird encoding in message body
I was playing with sqs + event source mapping and found an error on a lambda that works OK with real SQS

turns out, moto formats the message with https://github.com/getmoto/moto/blob/3b56583db6b727febb924ecae6918cead5786b7f/moto/sqs/models.py#L163

which results in message bodies that are not JSON parseable

seems like a pretty old change, do you recall the rationale behind `.replace('"', "&quot;")`?

bpandola:
This integration was a challenge because the SQS service underwent a protocol switch from Query (XML) to JSON back in 2023 that broke Moto (boto/botocore#2930, #6286, #7054) and took quite a bit of effort to resolve (#6331, #7057).  

Some additional commentary from boto/botocore#3515:
> Historical Context
> * SQS migrated protocols before services were able to model support for multiple protocols
> * During this migration, SQS models lost critical information from their "error" sections
> * This information loss was never properly addressed, leading to some persistent breaking changes

This PR makes the SQS backend protocol-agnostic, removing all of the messy conditionals that had been used to determine if the response methods should return XML or JSON.  Because the current Botocore model for SQS is missing critical information needed to support both protocols (as noted above), this PR also includes a `moto-extras` file for SQS to supplement the existing Botocore model with the missing operation/shape attributes.  A dedicated `SqsQuerySerializer` is also included because of an SQS-specific XML format that escapes HTML entities within XML tags.

The `responses.py` file for SQS was essentially rewritten, but relatively minor updates were needed in `models.py`.  No changes to the tests suite were required outside of correcting a couple of error code assertions (which were verified against a real AWS backend).

Fixes #9331

### DIFF
--- a/moto/core/serialize.py
+++ b/moto/core/serialize.py
@@ -978,7 +978,9 @@ class DoublePassEncoder:
 
 class SqsQuerySerializer(QuerySerializer):
     """
-    Special handling of SQS query responses to escape special characters
+    Special handling of SQS Query protocol responses:
+        * support aws.protocols#awsQueryCompatible trait after switch to JSON protocol
+        * escape HTML entities within XML tag text
     """
 
     def __init__(self, *args: Any, **kwargs: Any) -> None:


bpandola:
None of the missing lines in the CodeCov report are a concern.  The missing lines in models.py and responses.py are not covered on master, and are only flagged because they were touched by this PR (I changed the name of the exception class).

moto/sqs/responses.py
```diff
@@ -1,19 +1,10 @@
...
 from urllib.parse import urlparse
 
 from moto.core.common_types import TYPE_RESPONSE
-from moto.core.exceptions import JsonRESTError
-from moto.core.responses import BaseResponse
-from moto.core.utils import (
-    camelcase_to_pascal,
-    camelcase_to_underscores,
-    underscores_to_camelcase,
-)
```
bpandola, line 14: It sure is nice to get rid of all this casing nonsense!  😄

PR #9387, opened 2025-10-19
Add instance variable annotations for SES models

giograno:
This PR reworks the annotation for SES, as suggested in https://github.com/getmoto/moto/issues/9389.

In a few instances, a proper type annotation would require the introduction of new types mapping specific AWS data types. A case would be having a type for [`dkim_attributes`](https://github.com/giograno/moto/blob/7edd15802815da40165faac4df6c6f16c70c3711/moto/ses/models.py#L348) that implements the proper AWS type ([`DkimAttributes`](https://docs.aws.amazon.com/ses/latest/APIReference-V2/API_DkimAttributes.html)). Instead of introducing a new type, I left a generic `typing.Any`.

### DIFF
--- a/moto/core/base_backend.py
+++ b/moto/core/base_backend.py
@@ -43,6 +43,10 @@ class InstanceTrackerMeta(type):
 
 
 class BaseBackend:
+    region_name: str
+    account_id: str
+    partition: str
+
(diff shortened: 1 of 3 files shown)


bpandola:
Closing as a result of https://github.com/localstack/localstack/issues/13588

PR #9392, opened 2025-10-21
DynamoDB: add WarmThroughput parameter to Tables/GSIs

Nick-Source:
DynamoDB in moto is missing the parameter for WarmThroughput on the table itself and in GSIs. I noticed this when testing moto proxy with terraform and saw it fail. This PR adds the parameter to the table and in GSIs. 

### Validation
The values provided for it are only validated on creation as that was the only validation function I found so far. The requirement is that they can't drop lower than either the default OnDemand values or provisioned throughput values. The constants are added as global variables to the tables.py model. The status is always kept as "ACTIVE" too, I found this to be in line with the table status and simplest to do. The validation done isn't extensive and assumes the caller provides the required parameters as well as proper values within their specified ranges. For example, if ProvisionedThroughput is defined it must be valid.

### Default Values
When WarmThroughput isn't specified, this parameter defaults to the values of ProvisionedThroughput or the OnDemand default values.

### Tests
Extra tests were added as well for this new parameter. The update table test checks for an "UPDATING" state too in case this is tested against real AWS infra as I've noticed updating it takes longer. In the other tests, I kept it to check for the "ACTIVE" status since I noticed different tests checking statuses, such as IndexStatus, to match "ACTIVE". I was unsure of how to handle this exactly, so if something else needs to done, it can be changed.

API:
https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_CreateTable.html
https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_DescribeTable.html

Request:
https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_WarmThroughput.html

Response:
https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_TableWarmThroughputDescription.html
https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_GlobalSecondaryIndexWarmThroughputDescription.html

### DIFF
--- a/tests/test_dynamodb/test_dynamodb_create_table.py
+++ b/tests/test_dynamodb/test_dynamodb_create_table.py
@@ -149,6 +149,11 @@ def test_create_table_with_gsi():
                 "ReadCapacityUnits": 0,
                 "WriteCapacityUnits": 0,
             },
+            "WarmThroughput": {
+                "ReadUnitsPerSecond": 12000,
+                "Status": "ACTIVE",
+                "WriteUnitsPerSecond": 4000,
+            },
         }
     ]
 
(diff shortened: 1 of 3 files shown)


lorengordon:
Came here to note this also! Was wondering why my terraform tests were failing, traced it back to AWS provider v6.13.0, then saw they added a mandatory check on the WarmThroughput attribute to mark the resource status as ready. Thank you!

PR #9393, opened 2025-10-22
[FSx] Adding FSx Backup endpoints

LeekSwan:
This PR implements backup for FSx resources and some minor changes

Added endpoints:
- create_backup
- delete_backup
- list_tags_for_resource

Resource group tagging API:
- File Systems
- Backups

Refactored the tag resource methods to use TaggingService

### DIFF
--- a/tests/test_fsx/test_fsx.py
+++ b/tests/test_fsx/test_fsx.py
@@ -190,8 +190,8 @@ def test_tag_resource():
 
     client.tag_resource(ResourceARN=resource_arn, Tags=tags)
 
-    resp = client.list_tags_for_resource(ResourceARN=resource_arn)
-    assert resp["Tags"] == tags
+    resp = client.describe_file_systems(FileSystemIds=[file_system_id])
+    assert resp["FileSystems"][0]["Tags"] == tags
 
 
 @mock_aws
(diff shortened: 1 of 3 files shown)


tests/test_fsx/test_fsx.py
```diff
@@ -119,8 +190,8 @@ def test_tag_resource():
 
     client.tag_resource(ResourceARN=resource_arn, Tags=tags)
 
-    resp = client.describe_file_systems(FileSystemIds=[file_system_id])
```
bblommers, line 122: Nitpick - but can we keep these assertions? 

We still want to validate that the `describe_..` method works as expected (returns the updated tags), we should just extend the test to **also** verify that `list_tags_..` returns the correct tags.

tests/test_fsx/test_fsx.py
```diff
@@ -146,6 +217,6 @@ def test_untag_resource():
 
     client.untag_resource(ResourceARN=resource_arn, TagKeys=tag_keys)
 
-    resp = client.describe_file_systems(FileSystemIds=[file_system_id])
```
bblommers, line 149: Same here

bblommers (changes requested):

PR #9397, opened 2025-10-23
Fix SFN.client.start_execution() idempotency

chriselion:
Resolves https://github.com/getmoto/moto/issues/9394

When an execution with the same name is found, it now compares the inputs:
* If the inputs are different, an `ExecutionAlreadyExists` exception is raised (same as before)
* If the inputs are the same, the original execution is returned, instead of creating a new one.

I updated the existing test case to pass an input, and added another test to check that duplicating the name and input is idempotent.

---

I'm not sure how we should define "equal" inputs here. StateMachine.start_execution() saves the json-decoded input on the Execution
https://github.com/getmoto/moto/blob/86fe36816cd252de3be0454f05f03b1aee7d9d69/moto/stepfunctions/models.py#L133-L139
although the field is supposed to be a `str`
https://github.com/getmoto/moto/blob/86fe36816cd252de3be0454f05f03b1aee7d9d69/moto/stepfunctions/models.py#L310-L318
Since the original input is lost, I'm comparing the json-decoded values.

I can do a little more digging to see what AWS actually does (although if you have some intuition, that would save me some time). If we need to check the exact string equality instead, that'll take some more changes.

### DIFF
--- a/other_langs/tests_sagemaker_client/custom_script.py
+++ b/other_langs/tests_sagemaker_client/custom_script.py
@@ -1,22 +1,21 @@
 import argparse
-import os
 import json
+import os
 
-if __name__ =='__main__':
-
+if __name__ == "__main__":
     parser = argparse.ArgumentParser()
 
     # hyperparameters sent by the client are passed as command-line arguments to the script.
-    parser.add_argument('--epochs', type=int, default=10)
-    parser.add_argument('--batch-size', type=int, default=100)
-    parser.add_argument('--learning-rate', type=float, default=0.1)
+    parser.add_argument("--epochs", type=int, default=10)
+    parser.add_argument("--batch-size", type=int, default=100)
+    parser.add_argument("--learning-rate", type=float, default=0.1)
 
     # an alternative way to load hyperparameters via SM_HPS environment variable.
-    parser.add_argument('--sm-hps', type=json.loads, default=os.environ['SM_HPS'])
+    parser.add_argument("--sm-hps", type=json.loads, default=os.environ["SM_HPS"])
 
     # input data and model directories
-    parser.add_argument('--model-dir', type=str, default=os.environ['SM_MODEL_DIR'])
-    parser.add_argument('--train', type=str, default=os.environ['SM_CHANNEL_TRAIN'])
-    parser.add_argument('--test', type=str, default=os.environ['SM_CHANNEL_TEST'])
+    parser.add_argument("--model-dir", type=str, default=os.environ["SM_MODEL_DIR"])
+    parser.add_argument("--train", type=str, default=os.environ["SM_CHANNEL_TRAIN"])
+    parser.add_argument("--test", type=str, default=os.environ["SM_CHANNEL_TEST"])
 
     args, _ = parser.parse_known_args()
--- a/other_langs/tests_sagemaker_client/test_model_training.py
+++ b/other_langs/tests_sagemaker_client/test_model_training.py
@@ -1,15 +1,17 @@
-from moto import mock_aws
+from sagemaker.modules.configs import InputData, SourceCode
 from sagemaker.modules.train import ModelTrainer
-from sagemaker.modules.configs import SourceCode, InputData
+
+from moto import mock_aws
 
 
 @mock_aws
 def test_model_trainer():
-
     # https://sagemaker.readthedocs.io/en/stable/overview.html#using-modeltrainer
 
     # Image URI for the training job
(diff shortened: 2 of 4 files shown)


tests/test_stepfunctions/test_stepfunctions.py
```diff
@@ -540,25 +540,54 @@ def test_state_machine_start_execution_with_custom_name():
...
+    execution_one = client.start_execution(
+        stateMachineArn=sm["stateMachineArn"],
+        name="execution_name",
+        input='{"a": "b", "c": "d"}',
+    )
+    #
+    execution_two = client.start_execution(
+        stateMachineArn=sm["stateMachineArn"],
+        name="execution_name",
+        input='{"c": "d", "a": "b"}',
```
chriselion: Here's a more concrete example of what I meant in the description - are these inputs equal or not?

bblommers:
Hi @chriselion , thanks for raising the PR!

> although if you have some intuition, that would save me some time

No, I'm afraid not - I would (also) have to do a bunch of manual testing to determine the exact behavior.

This does look like a tricky scenario, so it does require some more research IMO.

I'd be happy to do so, I just can't guarantee a timeline when that would happen. Let me know what you prefer.

(Edit: also marking this PR as a draft, to make it clear it's not quite ready yet.)

chriselion:
I ran this script

```python
import boto3
import json
import os

definition = {
  "StartAt": "Wait",
  "States": {
    "Wait": {
      "Type": "Wait",
      "Seconds": 60,
      "End": True
    }
  },
  "QueryLanguage": "JSONata"
}

input1 = '{"a": "b", "c": "d"}'
input2 = '{"c": "d", "a": "b"}'
execution_name = "testExecution"
assert json.loads(input1) == json.loads(input2)


client = boto3.client("stepfunctions")

print("creating state machine")
resp = client.create_state_machine(
    name="motoRepro1",
    definition=json.dumps(definition),
    roleArn=os.getenv("ROLE_ARN")
)
state_machine_arn = resp["stateMachineArn"]

try:
    execution_response = client.start_execution(
        stateMachineArn=state_machine_arn,
        name=execution_name,
        input=input1,
    )
    execution_arn = execution_response["executionArn"]
    print("Initial execution successful")
except:
    print("Initial execution failed")

try:
    identical_response = client.start_execution(
        stateMachineArn=state_machine_arn,
        name=execution_name,
        input=input1,
    )
    print("Identical execution successful")
    print(f"{(execution_arn==identical_response["executionArn"])=}")
except:
    print("Identical execution failed")

try:
    equivalent_response = client.start_execution(
        stateMachineArn=state_machine_arn,
        name=execution_name,
        input=input2,
    )
    print("Equivalent execution successful")
    print(f"{(execution_arn==equivalent_response["executionArn"])}")
except:
    print("Equivalent execution failed")

print("deleting state machine")
client.delete_state_machine(stateMachineArn=state_machine_arn)
```
which produces this output:
```
creating state machine
Initial execution successful
Identical execution successful
(execution_arn==identical_response["executionArn"])=True
Equivalent execution failed
deleting state machine
```

So I'll update `Execution` to store the string form of the execution_input (which is what the type hint indicates anyway). Hopefully I can get to that today.

chriselion:
I tried this tonight, but it's more complicated than I expected. I removed the `json.load()` call in start_execution() and figured I'd need to remove the corresponding `json.dumps()` here
https://github.com/getmoto/moto/blob/e9a44a25280e1f8ba13ca340e874d06522a21471/moto/stepfunctions/responses.py#L210
but that ended up breaking a lot of tests.

moto/stepfunctions/parser/backend/execution.py
```diff
@@ -145,7 +145,7 @@ def __init__(
         self.region_name = region_name
         self.state_machine = state_machine
         self._cloud_watch_logging_session = cloud_watch_logging_session
-        self.input_data = input_data
+        self.input_data = json.loads(input_data)
```
chriselion, line 148: Now `input_data` is saved as decoded json, and `execution_input` is a `str` - previously the decoding happened in StepFunctionsParserBackend.start_execution, so both would be decoded.

chriselion:
@bblommers - I finally had a chance to come back to this; I think I got it everything working. Part of my initial confusion was the two `Execution` classes, and I only fixed one of them.

~Several tests are failing because of the issue you noted in https://github.com/getmoto/moto/pull/9459 - I'll merge the fix into my branch when it lands.~ Merged master after https://github.com/getmoto/moto/pull/9460

bblommers (approved):
This looks great - thank you for investigating this and for contributing the fix @chriselion!

PR #9402, opened 2025-10-25
[ElastiCache] Make CacheNodeType a required field for Cluster

ApogeeAC:
_Summary_
This PR updates the ElastiCache Cluster model to make the CacheNodeType field required. This ensures that attempts to create a cluster without specifying CacheNodeType now correctly raise an error, matching AWS behavior.

_Changes Introduced_
CacheNodeType is now a required field in the Cluster model.
Added validation to raise an error if CacheNodeType is missing when calling create_cache_cluster.
Added a supporting test to confirm that omitting CacheNodeType raises:
`An error occurred (InvalidParameterValue) when calling the CreateCacheCluster operation: The parameter CacheNodeType must be provided and must not be null.`


bblommers (approved):
LGTM, thanks @ApogeeAC!

PR #9411, opened 2025-10-28
EC2: Add support for modifying all volume params

jkobejs:
**Summary**

This PR adds missing parameters to the modify_volume implementation in the EC2 service to align it with the official [Boto3 modify_volume API](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/ec2/client/modify_volume.html#)
.
**Changes**

Added support to modify all volume parameters.
Updated internal volume model and response serialization to handle these new attributes.
Extended tests to verify that modified parameters persist correctly and match expected API behavior.

**Questions**

Do we need to have validation in modify_volume method like in create_volume?
There are some conditions that we can verify, but it seems that they are documented across various pages in AWS docs.

### DIFF
--- a/moto/ec2/models/elastic_block_store.py
+++ b/moto/ec2/models/elastic_block_store.py
@@ -53,18 +53,28 @@ class VolumeModification:
             )
 
         self.volume = volume
-        self.original_size = volume.size
-        self.original_volume_type = volume.volume_type
-        self.original_iops = volume.iops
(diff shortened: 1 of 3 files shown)


bblommers (approved):
LGTM! Thanks for adding support for these parameters @jkobejs, and welcome to Moto!

jkobejs:
> bblommers

Tnx 😄

PR #9415, opened 2025-10-29
Improve dashboard

mysiki:
Improve dashboard using 
- latest bootstrap 
- latest bootstrap-table
- export JS in external file
- Add sidebar for service
- Add collapse option 
- Add table with sort and search 

image

### DIFF
--- a/moto/moto_server/templates/dashboard.html
+++ b/moto/moto_server/templates/dashboard.html
@@ -157,7 +157,8 @@
   <!-- JavaScript Dependencies -->
 
   <!-- jQuery library - required for Bootstrap Table functionality -->
-  <script src="https://cdn.jsdelivr.net/npm/jquery/dist/jquery.min.js"></script>
+  <script src="https://code.jquery.com/jquery-3.7.1.min.js"
+    integrity="sha256-/JqT3SQfawRcv/BIHPThkBvs0OEvtFFmqPF/lYI/Cxo=" crossorigin="anonymous"></script>
 
   <!-- Bootstrap 5 JavaScript bundle including Popper.js for modals, dropdowns, and other interactive components -->
   <script src="https://cdn.jsdelivr.net/npm/bootstrap@5.3.8/dist/js/bootstrap.bundle.min.js"


bblommers (changes requested):
Hi @mysiki, thanks for the PR!

Having a proper searchable table does make sense, and should make things more readable. This does mean we need two additional dependencies, `jquery` and `bootstrap-table`, but that is an acceptable tradeoff as far as I'm concerned.

## Positive aspects of this change:
 - Search, as I just mentioned
 - Having a separate table per resource type makes it much easier to scroll horizontally, as you can scroll only the table for that particular resource
 - Improved styling, which leads to better readability
 - Having the links to the Github and official docs makes a lot of sense

## Request for changes:

The previous version had a slightly different layout:
1. the list of services was the top, and
2. only resources for a particular service was shown on the page. 

Now, I recognize that this is a personal preference - but I actually liked this better:
1. The page is less cluttered if we keep the left space empty, and
2. It is easier to see which resources belong to a specific service

If we can keep this layout (services at the top, only resources for one particular service showing at the time) but add a proper table per resource type, that would make an ideal combination IMHO.

For reference, this is what the current page looks like:
image

Let me know if you have any questions/concerns!

mysiki:
Hey, thanks for your feedback, I understand what you means. Was my first idea to have a "per service" page and "all". 

I will provide some other design asap.

mysiki:
@bblommers Following you comment, I update it with 2 versions : 
- moto-api
- moto-api/v1 

It's not cleanup but just to know what you prefers / wants. For my point of view, de left side bar is better but in fact it's not really important (:D) so just say me which version you want and I will cleanup the code. 

Nota : on V1 I use bootstrap tab, so all data are load and item selector play show/hide on table. On "side bar" I filter data, means, when you move from one page to another you reload the data.json. 
show/hide make the data load only one time, but have little lag sometime if you have many data. Side bar do not have this lag but as the data.json is parse, you also have few second wait. 

Nota 2 : I remove some stuff like fold / unfold, I think is useless if we already have only the service content. The end goal is to create a code simple to keep up to date, so with minimal side options. 

Also the "all services" is easy with the side bar, let me know if you want it in the selected design. 

My main pbl was : it really hard to check just VPC because of all the AMI stuff etc. So basically, in both version it fix and we can filter/browse data easily. Base on that, I can apply the version that you want. 

I also ask myself, as multiple endpoint seems to be present, do you think that can be a good feature to have more option like "start record", "stop record" etc directly from the dash ? 

image

image

bblommers:
Hi @mysiki, thanks for the update! Let's go with V1 then, that is more readable IMO.

> Also the "all services" is easy with the side bar, let me know if you want it in the selected design.

I don't immediately see a use case for this, so I'd vote to leave it out.

> I also ask myself, as multiple endpoint seems to be present, do you think that can be a good feature to have more option like "start record", "stop record" etc directly from the dash ?

That is an interesting idea! There are quite a few API's that people can use to configure Moto, so being able to invoke that from a UI would make sense. If you want to tackle this, that would be very welcome! My only request would be to do this in a separate PR, just to simplify my job of reviewing everything :)

mysiki:
Change summarize : 

Dashboard update : Main goal, make it more readable

- Update Bootstrap from 3.3.7 to 5.3.8
- Add bootstrap-table 1.25.0 
- Update JQuery from 3.6.4 to 3.7.1 (dependency require for bootstrap table) 
- Remove handlebars dependency 
- Generate smart table with bootstrap-table (filter, size control, sort...)
- Add external link to Moto doc
- Add external link to Moto GitHub
- Sticky top the nav bar (main menu and services list)
- Move JS to separate file
- Add dark mode support

Screenshot result : 
image
image

PR #9418, opened 2025-10-31
Support tagging existing CloudFront distributions

jtschladen:
Support tagging existing CloudFront distributions.

### DIFF
--- a/moto/cloudfront/responses.py
+++ b/moto/cloudfront/responses.py
@@ -28,9 +28,10 @@ class CloudFrontResponse(BaseResponse):
         operation = response._get_param("Operation")
         if operation == "Tag":
             return 204, {}, response.tag_resource()[2]
+        if operation == "Untag":
+            return 204, {}, response.untag_resource()[2]
         if request.method == "GET":
             return 200, {}, response.list_tags_for_resource()[2]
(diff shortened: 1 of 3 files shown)


moto/cloudfront/responses.py
```diff
@@ -21,6 +21,17 @@ def _get_xml_body(self) -> dict[str, Any]:
...
         return cloudfront_backends[self.current_account][self.partition]
 
+    def _get_action_from_method_and_request_uri(
+        self, method: str, request_uri: str
+    ) -> str:
+        # CloudFront uses a query param "Operation" on its tag endpoint
+        if self.querystring.get("Operation"):
+            operation = self.querystring["Operation"]
+            if isinstance(operation, list):
+                operation = operation[0]
```
bblommers: Just a note: you could use `operation := self._get_param("Operation")` here, the `_get_param`-method will handle any list-conversions automatically if necessary

jtschladen:
Thank you @bblommers , this solution is much better! Hopefully it looks correct to you now, please let me know if I missed anything.

I added untag support as well while fixing it up since it only seems logical to support both.

bblommers (approved):
Looks great - thanks for fixing this @jtschladen!

PR #9442, opened 2025-11-05
EC2: integrate with core request parser

bpandola:
This PR opts the EC2 backend into the new request parsing pipeline.  Due to the surface area of the EC2 API, the changeset is deceptively large.  Importantly, only one test needed to be modified (to account for now correctly parsed floating point numbers).

On the responses side, the vast majority of the changes involve replacing direct references to `self.querystring` and any specialized parameter parsing methods (e.g. `self._get_list_prefix()`) with`self._get_param()`.  Changes on the models side are mostly around converting existing datetime, boolean, and integer values that had been stored as strings (to simplify XML serialization) into proper, native Python data types.

> [!WARNING]
> Lots of EC2 input consists of large, deeply nested objects.  Although there is good test coverage, there is a chance that an end-user could provide some specific combination of input that worked previously but is broken when handled by the new parser (in combination with the existing XML templates).  Any potential issues of that sort will be fixed promptly if/when they arise.  Additionally, now that the request parsing has been standardized, work to convert the remaining EC2 backends to the new serialization pipeline can proceed in earnest (which will eliminate any type-related bugs in the existing XML templates).

### DIFF
--- a/moto/acmpca/models.py
+++ b/moto/acmpca/models.py
@@ -65,6 +65,7 @@ class CertificateAuthority(BaseModel):
         self.certificate_bytes: bytes = b""
         self.certificate_chain: Optional[bytes] = None
         self.issued_certificates: dict[str, bytes] = {}
+        self.issued_certificates_certificate_chains: dict[str, bytes] = {}
 
         self.subject = self.certificate_authority_configuration.get("Subject", {})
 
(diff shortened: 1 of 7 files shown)


bblommers (approved):
LGTM! Nice cleanup

If this does lead to any bugs, well.. that just means we get to have more tests/better coverage. That's only a positive thing IMO :slightly_smiling_face:

bpandola:
> If this does lead to any bugs, well.. that just means we get to have more tests/better coverage. That's only a positive thing IMO 🙂
@bblommers Ha! I like this point of view!

PR #9443, opened 2025-11-05
acm-pca: Support CertificateChain in GetCertificate operation

nik-localstack:
Add support for returning CertificateChain in [GetCerficate](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/acm-pca/client/get_certificate.html) operation


bblommers (approved):
LGTM - thanks for adding this to Moto @nik-localstack!

PR #9444, opened 2025-11-06
AutoScaling: Mixed instance autoscaling groups

viren-nadkarni:
Please see the [comment below](https://github.com/getmoto/moto/pull/9444#issuecomment-3539428314) for PR description

### DIFF
--- a/moto/autoscaling/models.py
+++ b/moto/autoscaling/models.py
@@ -985,9 +985,9 @@ class FakeAutoScalingGroup(CloudFormationModel):
                 # to the passed percentage.
                 above_base_capacity = target_capacity - on_demand_base
 
-                on_demand_above_base_capacity = int(
-                    above_base_capacity * percent_above_base
-                ) / 100.0
+                on_demand_above_base_capacity = (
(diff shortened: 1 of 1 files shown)


baermat:
This PR adds more extensive calculations for MixedInstanceTypes for auto scaling groups, which are closer to what AWS does internally.
When a mixed instance types template is passed, the following should happen:
- The base should be filled with on-demand instances
- The remaining capacity up to desired should be filled with either on-demand or spot instances
  - This is determined by the passed percentage
- All of this is measured in abstract units, not in instances. Instances are assigned a certain unit value, which can lead to "extra ceil rounding instances" . If e.g. an instances has weight 4 units, and desired capacity is 8, but the calculation calls for 2u + 3u + 3u, it results in 3 instances, not 2 4u+4u (see test for details).

Limitations: The function currently expects a "prioritized" strategy for the OnDemandAllocationStrategy. Adding more strategies would have been out of scope, so this was left as is.

baermat:
@viren-nadkarni please start the workflows, review the changes and if ready mark as ready for review please :)

PR #9447, opened 2025-11-06
[DMS] Implement Test Connection, Describe Connection and add state transition to Replication Instance creation

gezb:
Implements missing Database Migration Service methods:

- [x] test_endpoint()
- [x] describe_connection()

Also adds optional State Transition on Replication Instance creation


bblommers (approved):
LGTM - thanks @gezb! Nice to see the state transitions in more and more services

PR #9453, opened 2025-11-08
ECS: integrate with core request parser/response serializer

bpandola:
I picked this backend because the ECS API has the concept of conditional response fields, which I thought would be an interesting use-case for the new serialization pipeline.  I initially tried to solve this by just `deepcopy()`-ing the backend objects and deleting attributes that were not to be included, but in addition to being inelegant it actually didn't work with any object that contained [ManagedState](https://github.com/getmoto/moto/blob/d5c8827ba2bb976ed9a1616d0b09eaa401ef4d93/moto/moto_api/_internal/managed_state_model.py#L7) (because `deepcopy()` would call the `.status` attribute and advance the transition counter prematurely).  Anyway, I ended up using the response transformer mechanism that's already in place to do some per-request response transformation, which ended up working perfectly and being quite readable.

The ECS backend was also creating dedicated "response objects" for each model, all of which is now unnecessary (and has been deleted).  Some very minor test updates were needed to correct error codes and field types.

### DIFF
--- a/moto/ecs/responses.py
+++ b/moto/ecs/responses.py
@@ -120,16 +120,18 @@ class EC2ContainerServiceResponse(BaseResponse):
         )
 
     def describe_clusters(self) -> ActionResult:
-        names = self._get_param("clusters")
+        cluster_names = self._get_param("clusters", [])
+        clusters, failures = self.ecs_backend.describe_clusters(cluster_names)
         include = self._get_param("include", [])
-        self.conditional_response_field(DESCRIBE_CLUSTERS_TAGS, "TAGS" in include)
-        clusters, failures = self.ecs_backend.describe_clusters(names)
-        return ActionResult(
-            {
-                "clusters": clusters,
-                "failures": failures,
-            }
-        )
+        if "TAGS" in include:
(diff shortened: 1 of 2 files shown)


moto/ecs/responses.py
```diff
@@ -71,46 +96,49 @@ def put_cluster_capacity_providers(self) -> str:
...
-            }
-        )
+    def describe_clusters(self) -> ActionResult:
+        cluster_names = self._get_param("clusters", [])
+        clusters, failures = self.ecs_backend.describe_clusters(cluster_names)
+        include = self._get_param("include", [])
+        if "TAGS" in include:
+            self._include_in_response(DESCRIBE_CLUSTERS_TAGS)
+        else:
+            self._exclude_from_response(DESCRIBE_CLUSTERS_TAGS)
```
bpandola, line 129: Moto only supports `TAGS` at the moment, but there are lots of other possible values in the `include` parameter.  Supporting these in the future would be quite simple, e.g. having a map of `include` param names with their associated response path, which could simply be looped through to include/exclude as needed per request.

bpandola:
The missing coverage is just exceptions.  I changed the class names, but the lines aren't covered on master either.

PR #9458, opened 2025-11-09
[RDS] Implement add_role_to_db_instance() and add_role_to_db_cluster()

gezb:
[RDS] Implement `add_role_to_db_instance()` and `add_role_to_db_cluster()`

### DIFF
--- a/moto/rds/models.py
+++ b/moto/rds/models.py
@@ -38,11 +38,13 @@ from .exceptions import (
     BlueGreenDeploymentNotFoundFault,
     DBClusterNotFoundError,
     DBClusterParameterGroupNotFoundError,
+    DBClusterRoleAlreadyExists,
     DBClusterSnapshotAlreadyExistsError,
     DBClusterSnapshotNotFoundError,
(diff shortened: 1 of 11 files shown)


moto/rds/models.py
```diff
@@ -1400,6 +1421,20 @@ def physical_resource_id(self) -> Optional[str]:
     def latest_restorable_time(self) -> datetime:
         return utcnow()
 
+    @property
+    def associated_roles(self) -> list[dict[str, Any]]:  # type: ignore[misc]
```
bblommers: Is it possible to just return `self._associated_roles` here? Our response serializer should automatically parse the `AssociatedRole`-class and convert it into a dict, we shouldn't have to manually handle this.

If this does not work, then there is something wrong with our serializer, and we can stick with the current implementation - but returning the `list[AssociatedRole]` directly _should_ work :slightly_smiling_face:

gezb: Thanks I have made the change and added extra tests to cover the missing lines

PR #9459, opened 2025-11-09
Tests: Use pytest.skip instead of unittest.SkipTest, now that's broken in pytest 9.0

bblommers:
Relevant bug report: https://github.com/pytest-dev/pytest/issues/13895

It looks like it may be reverted, but that's still under discussion, and 200+ of our tests break because of it. So I'll probably just merge anyway.

### DIFF
--- a/tests/test_acm/test_acm.py
+++ b/tests/test_acm/test_acm.py
@@ -2,7 +2,7 @@ import datetime
 import os
 import uuid
 from ipaddress import IPv4Address
-from unittest import SkipTest, mock
+from unittest import mock
 
 import boto3
 import pytest
@@ -477,7 +477,7 @@ def test_request_certificate():
 @mock.patch("moto.settings.ACM_VALIDATION_WAIT", 1)
 def test_request_exportable_certificate():
     if not settings.TEST_DECORATOR_MODE:
-        raise SkipTest("Can only change setting in DecoratorMode")
+        raise pytest.skip("Can only change setting in DecoratorMode")
     client = boto3.client("acm", region_name="eu-central-1")
 
     resp = client.request_certificate(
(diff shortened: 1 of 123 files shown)


tests/test_acm/test_acm.py
```diff
@@ -477,7 +477,7 @@ def test_request_certificate():
 @mock.patch("moto.settings.ACM_VALIDATION_WAIT", 1)
 def test_request_exportable_certificate():
     if not settings.TEST_DECORATOR_MODE:
-        raise SkipTest("Can only change setting in DecoratorMode")
+        raise pytest.skip("Can only change setting in DecoratorMode")
```
chriselion, line 480:
```suggestion
       pytest.skip("Can only change setting in DecoratorMode")
```
This raises its own exception:
https://github.com/pytest-dev/pytest/blob/6eb7609c7d738da904de5c8b0b7c9c2a80ddf3e0/src/_pytest/outcomes.py#L137-L142
but you should use it like a normal function call.

chriselion:
It might be simpler to pin to `pytest<9.0.0` in `requirements-tests.txt` for now, at least until the upstream issue is resolved?

tests/test_acm/test_acm.py
bblommers: TIL! Thanks for the hint @chriselion

bblommers:
> It might be simpler to pin to `pytest<9.0.0` in `requirements-tests.txt` for now, at least until the upstream issue is resolved?

Yeah, I agree. I wasn't quite sure that the `pytest` maintainers would actually fix it - but it is already fixed now. I've raised #9460 to do just that :+1:

chriselion:
Oh nice, I didn't see they had reverted. Yeah, I'd expect them to publish a patch version and yank 9.0.0 (so you shouldn't need to exclude it) soon.

PR #9461, opened 2025-11-11
EC2: Fix type for key pair public key material

viren-nadkarni:
This PR fixes a bug in the `KeyPair` resource, where the `KeyPair.material_public` attribute is expected to be `str` but held `byte` instead.

### DIFF
--- a/moto/ec2/responses/key_pairs.py
+++ b/moto/ec2/responses/key_pairs.py
@@ -26,7 +26,7 @@ class KeyPairs(EC2BaseResponse):
 
     def import_key_pair(self) -> str:
         name = self._get_param("KeyName")
-        material = self._get_param("PublicKeyMaterial")
+        material = self._get_param("PublicKeyMaterial").decode("utf-8")  # type: str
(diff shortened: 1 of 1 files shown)


bblommers (approved):
Thanks for the fix @viren-nadkarni! 

The fact that this 'works' without having to adjust any tests looks a little scary, but in this case, the variable is only used in two places
 1. the `public_key_parse`-utility method, which accepts both strings and bytes
 2. In the `KeyPair`-class, where the variable is stored, but never used.

So that's why it is possible to simply change the type without any side effects.

bpandola:
@bblommers @viren-nadkarni One of the benefits of the new request parsing is never having to manually `encode()` or `decode()`.  This parameter is supposed to be bytes.  Unless one of you can tell me why this was a bug, I'm going to revert this in favor of fixing the downstream type hints (as well as the downstream `if isinstance(key_material, str):` which was only necessary because of the original [incorrect] parameter parsing).

viren-nadkarni:
@bpandola Either approach is fine as long as the actual type is what is hinted :)

PR #9463, opened 2025-11-12
[Pipes] New Methods

Singha22:
Resolves https://github.com/getmoto/moto/issues/9438

### DIFF
--- a/moto/pipes/models.py
+++ b/moto/pipes/models.py
@@ -91,7 +91,7 @@ class EventBridgePipesBackend(BaseBackend):
         tags: Optional[dict[str, str]],
         log_configuration: Optional[dict[str, Any]],
         kms_key_identifier: Optional[str],
-    ) -> tuple[str, str, str, str, str, str]:
+    ) -> Pipe:
         pipe = Pipe(
             name=name,
             account_id=self.account_id,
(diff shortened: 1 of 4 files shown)


moto/pipes/models.py
```diff
@@ -0,0 +1,294 @@
...
+        source_parameters: Optional[dict[str, Any]],
+        enrichment: Optional[str],
+        enrichment_parameters: Optional[dict[str, Any]],
+        target: str,
+        target_parameters: Optional[dict[str, Any]],
+        role_arn: str,
+        tags: Optional[dict[str, str]],
+        log_configuration: Optional[dict[str, Any]],
+        kms_key_identifier: Optional[str],
+    ) -> tuple[str, str, str, str, str, str]:
```
bblommers: It would be easier to understand if we just return the `Pipe` object here. The caller can then immediately understand what data it is getting, and decide which parts of the `Pipe` it needs (arn/name/etc)

moto/pipes/models.py
```diff
@@ -0,0 +1,294 @@
...
+            pipe.name,
+            pipe.desired_state,
+            pipe.current_state,
+            iso_8601_datetime_without_milliseconds(pipe.creation_time),
+            iso_8601_datetime_without_milliseconds(pipe.last_modified_time),
+        )
+
+    def describe_pipe(
+        self, name: str
+    ) -> tuple[
```
bblommers: Same here

moto/pipes/models.py
```diff
@@ -0,0 +1,294 @@
...
+            pipe.target_parameters,
+            pipe.role_arn,
+            pipe.tags,
+            iso_8601_datetime_without_milliseconds(pipe.creation_time),
+            iso_8601_datetime_without_milliseconds(pipe.last_modified_time),
+            pipe.log_configuration,
+            pipe.kms_key_identifier,
+        )
+
+    def delete_pipe(self, name: str) -> tuple[str, str, str, str, str, str]:
```
bblommers, line 122: And same here

tests/test_pipes/test_pipes.py
```diff
@@ -0,0 +1,395 @@
...
+            Source="arn:aws:sqs:us-east-1:123456789012:test-queue",
+            Target="arn:aws:lambda:us-east-1:123456789012:function:test-function",
+        )
+    error = exc_info.value.response["Error"]
+    assert error["Code"] == "ValidationException"
+    assert "RoleArn is a required parameter" in error["Message"]
+
+
+def test_validation_exception_with_field_list():
+    from moto.pipes.exceptions import ValidationException
```
bblommers: **comment** Ideally, all functionality should be tested using the `boto3`-client, to ensure compatibility.

These exceptions are not thrown anywhere yet, but I'm assuming that they will be used in future iterations? 

When that happens, we can re-use these tests to verify the behaviour of the Exceptions using `boto3`.

tests/test_pipes/test_pipes.py
```diff
@@ -0,0 +1,395 @@
...
+    assert pipe["Arn"] == pipe1["Arn"]
+    assert pipe["Name"] == "test-pipe-1"
+    assert pipe["DesiredState"] == "RUNNING"
+    assert pipe["CurrentState"] == "RUNNING"
+    assert pipe["Source"] == "arn:aws:sqs:eu-west-1:123456789012:queue-1"
+    assert pipe["Target"] == "arn:aws:lambda:eu-west-1:123456789012:function:func-1"
+
+
+@mock_aws
+def test_list_pipes_with_pagination():
```
bblommers, line 226: **praise**: Nice to see tests for this!

PR #9464, opened 2025-11-12
fix: Optimize request parsing.

nairb774:
When sending a large `Body` to `sagemaker_runtime.invoke_endpoint` the
response times would just explode. The regular expression used for
sorting the keys had some bad backtracking and performance behavior.

To improve the situation, we can simplify the regex and make use of the
Match object returned from the search to do the splitting. This
implementation is intended to retain the existing behavior but if the
`member` suffix in the first element of the tuple isn't needed then the
magic `+7` could be dropped.

### DIFF
--- a/moto/core/utils.py
+++ b/moto/core/utils.py
@@ -378,6 +378,10 @@ def extract_region_from_aws_authorization(string: str) -> Optional[str]:
     return region
 
 
+# Match: "member." followed by digits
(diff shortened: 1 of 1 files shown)


bpandola:
With the new request parsing for Moto that I've been working on, I don't think this param sort function is even needed anymore... I'll investigate further and update when I have more information.

Thanks for bringing this up and offering a solution.  If you have a sample Sagemaker input that is causing problems, it would be helpful if you could post it here.

PR #9467, opened 2025-11-13
SQS: Enable optional CRC32 header generation for SQS

rogvc-nice:
## Description

This PR introduces a flag to enable (default, current implementation) or disable CRC32 header generation for SQS responses based on an environment variable. 

This change ensures compatibility with certain SDK versions that may have issues with CRC32 validation. For example, it was breaking moto's compatibility with a dotnet project with AWS SDK v4, with errors similar to the issues below:
* https://github.com/aws/aws-sdk-net/issues/119
* https://github.com/aws/aws-sdk-net/discussions/3632

### DIFF
--- a/moto/comprehend/models.py
+++ b/moto/comprehend/models.py
@@ -707,6 +707,10 @@ class ComprehendBackend(BaseBackend):
             **kwargs,
         )
         self.jobs[job.job_id] = job
+
+        if "Tags" in kwargs:
+            self.tagger.tag_resource(job.job_arn, kwargs["Tags"])
+
         return job
 
     def _get_job(self, job_id: str) -> ComprehendJob:
@@ -795,10 +799,6 @@ class ComprehendBackend(BaseBackend):
         return self._list_jobs("DominantLanguageDetection", filter)
 
     def start_entities_detection_job(self, **kwargs: Any) -> ComprehendJob:
-        if "EntityRecognizerArn" not in kwargs:
-            raise InvalidRequestException(
-                "The request is missing the required parameter: EntityRecognizerArn."
-            )
         return self._start_job("EntitiesDetection", **kwargs)
 
     def describe_entities_detection_job(self, job_id: str) -> ComprehendJob:
@@ -824,10 +824,6 @@ class ComprehendBackend(BaseBackend):
         return self._list_jobs("TopicsDetection", filter)
 
     def start_document_classification_job(self, **kwargs: Any) -> ComprehendJob:
-        if "DocumentClassifierArn" not in kwargs:
-            raise InvalidRequestException(
-                "The request is missing the required parameter: DocumentClassifierArn."
-            )
         return self._start_job("DocumentClassification", **kwargs)
 
     def describe_document_classification_job(self, job_id: str) -> ComprehendJob:
--- a/moto/core/serialize.py
+++ b/moto/core/serialize.py
@@ -100,11 +100,13 @@ class SerializationContext:
 
 class TimestampSerializer:
     TIMESTAMP_FORMAT_ISO8601 = "iso8601"
+    TIMESTAMP_FORMAT_ISO8601_ZEROED = "iso8601_zeroed"
     TIMESTAMP_FORMAT_RFC822 = "rfc822"
     TIMESTAMP_FORMAT_UNIX = "unixtimestamp"
 
     ISO8601 = "%Y-%m-%dT%H:%M:%SZ"
     ISO8601_MICRO = "%Y-%m-%dT%H:%M:%S.%fZ"
+    ISO8601_MICRO_ZEROED = "%Y-%m-%dT%H:%M:%S.000Z"
 
     def __init__(self, default_format: str) -> None:
         self.default_format = default_format
@@ -124,6 +126,9 @@ class TimestampSerializer:
         else:
             timestamp_format = self.ISO8601
         return value.strftime(timestamp_format)
+
+    def _timestamp_iso8601_zeroed(self, value: datetime) -> str:
+        return value.strftime(self.ISO8601_MICRO_ZEROED)
 
     @staticmethod
     def _timestamp_unixtimestamp(value: datetime) -> float:
@@ -888,6 +893,8 @@ class QueryJSONSerializer(QuerySerializer):
 
 
 class EC2Serializer(QuerySerializer):
+    DEFAULT_TIMESTAMP_FORMAT = TimestampSerializer.TIMESTAMP_FORMAT_ISO8601_ZEROED
+
     def _serialize_body(self, body: Mapping[str, Any]) -> str:
         body_serialized = xmltodict.unparse(
             body,
--- a/moto/resourcegroupstaggingapi/models.py
+++ b/moto/resourcegroupstaggingapi/models.py
@@ -386,28 +386,56 @@ class ResourceGroupsTaggingAPIBackend(BaseBackend):
 
         # Comprehend
         if self.comprehend_backend:
-            if not resource_type_filters or "comprehend" in resource_type_filters:
-                for document_classifier in self.comprehend_backend.classifiers.values():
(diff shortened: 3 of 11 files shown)


bpandola:
@rogvc-nice Is it possible to disable the CRC32 header generation on real AWS?  If not, I don't think it makes sense to support it in Moto.  In any case, do you actually have a reproducible test case that demonstrates the issue?  From the linked issues you provided, the impact seems negligible.

rogvc-nice:
@bpandola , the issue is that Moto is doing crc32 headers on resources/services that don't use it in regular AWS scenarios. We can't disable it in real AWS because it's not there anymore. From the SQS API, we see that AWS uses MD5 hashes for [sending](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/APIReference/API_SendMessage.html#API_SendMessage_ResponseSyntax) and [receiving](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/APIReference/API_ReceiveMessage.html#API_ReceiveMessage_ResponseSyntax) messages. [Deletion](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/APIReference/API_DeleteMessage.html#API_DeleteMessage_Examples) also does not mention the offending header in the docs.

In one of the linked issues ([ref](https://github.com/aws/aws-sdk-net/discussions/3632)) we see that an AWS dev mentions `x-amz-crc32` is a legacy header that is no longer used by AWS SDK clients (other than potentially DynamoDB?) and from changes to the dotnet AWS SDK from at least oct 2022 (I was upgrading from v3.7.2.117 when I first encountered the error), we can't use SQS moto due to these CRC validation mismatches.

You can verify this behavior by using a barebones c# app that interfaces with moto SQS:

<details>
  <summary>Click me for testing details</summary>

   1. Scaffold the console project
      ```sh
      dotnet new console -n sqs
      cd sqs
      dotnet add package AWSSDK.SQS
      ```
   1. Edit `Program.cs` with test project
      ```csharp
      using Amazon.SQS;
      using Amazon.SQS.Model;

      namespace test_crc32_headers;

      class Program
      {
         static async Task Main(string[] args)
         {
            var awsEndpoint = Environment.GetEnvironmentVariable("AWS_ENDPOINT") ?? "http://localhost:5123";
            var awsRegion = Environment.GetEnvironmentVariable("AWS_REGION") ?? "us-east-1";

            var config = new AmazonSQSConfig();

            if (!string.IsNullOrEmpty(awsEndpoint))
            {
                  Console.WriteLine($"Using custom endpoint: {awsEndpoint}");
                  config.ServiceURL = awsEndpoint;
                  config.AuthenticationRegion = awsRegion;
            }
            else
            {
                  Console.WriteLine($"Using AWS region: {awsRegion}");
                  config.RegionEndpoint = Amazon.RegionEndpoint.GetBySystemName(awsRegion);
            }

            using var client = new AmazonSQSClient(config);

            try
            {
                  Console.WriteLine("1. Creating queue...");
                  var queueName = $"test-queue-{Guid.NewGuid():N}";
                  var createResponse = await client.CreateQueueAsync(new CreateQueueRequest
                  {
                     QueueName = queueName
                  });
                  var queueUrl = createResponse.QueueUrl;
                  Console.WriteLine($"   Queue created: {queueUrl}");

                  Console.WriteLine("2. Sending message...");
                  var sendResponse = await client.SendMessageAsync(new SendMessageRequest
                  {
                     QueueUrl = queueUrl,
                     MessageBody = "Test message for CRC32 header validation"
                  });
                  Console.WriteLine($"   Message sent: {sendResponse.MessageId}");

                  Console.WriteLine("3. Receiving message...");
                  var receiveResponse = await client.ReceiveMessageAsync(new ReceiveMessageRequest
                  {
                     QueueUrl = queueUrl,
                     MaxNumberOfMessages = 1,
                     WaitTimeSeconds = 5
                  });

                  if (receiveResponse.Messages.Count > 0)
                  {
                     var message = receiveResponse.Messages[0];
                     Console.WriteLine($"   Message received: {message.Body}");

                     Console.WriteLine("4. Deleting message...");
                     var deleteResponse = await client.DeleteMessageAsync(new DeleteMessageRequest
                     {
                        QueueUrl = queueUrl,
                        ReceiptHandle = message.ReceiptHandle
                     });
                     Console.WriteLine("   Message deleted successfully");
                  }
                  else
                  {
                     Console.WriteLine("   No messages received");
                  }

                  Console.WriteLine("5. Deleting queue...");
                  await client.DeleteQueueAsync(queueUrl);
                  Console.WriteLine("   Queue deleted");
            }
            catch (Exception ex)
            {
                  Console.WriteLine($"\n*** Error: {ex.Message}");
                  Console.WriteLine($"    Type: {ex.GetType().Name}");
            }
         }
      }

      ```
   1. Run moto

      ```sh
      docker run --rm -p 5123:5000 motoserver/moto:latest
      ```
   1. Run test app
      ```sh
      export AWS_ENDPOINT='http://localhost:5123'
      dotnet run
      ```

   You should see the program failing at the delete step:
   ```log
   Using custom endpoint: http://localhost:5123
   1. Creating queue...
      Queue created: http://localhost:5123/123456789012/test-queue-644e8d7960f3457fb98ed737744eed50
   2. Sending message...
      Message sent: ab7f73b0-8929-484a-9704-441a24de9265
   3. Receiving message...
      Message received: Test message for CRC32 header validation
   4. Deleting message...

   *** Error: CRC value returned with response does not match the computed CRC value for the returned response body.
      Type: IOException
   ```

   If you use the version with these changes, you would do something like

   1. Run moto
      ```sh
      docker run -e MOTO_SQS_DISABLE_CRC32_VALIDATION='true' --rm -p 5123:5000 motoserver/moto:sqs-skip-crc32
      ```
   1. Run test app
      ```sh
      export AWS_ENDPOINT='http://localhost:5123'
      dotnet run
      ```

   You should see the program succeed:
   ```log
   Using custom endpoint: http://localhost:5123
   1. Creating queue...
      Queue created: http://localhost:5123/123456789012/test-queue-c9c5e4c5ab494f9084f02f4372a7042f
   2. Sending message...
      Message sent: ced64807-fc3c-45f2-997e-fcb6e00c17ce
   3. Receiving message...
      Message received: Test message for CRC32 header validation
   4. Deleting message...
      Message deleted successfully
   5. Deleting queue...
      Queue deleted
   ```

</details>

The non-CRC behavior is consistent with regular AWS. Opted for the flag route instead of removing headers altogether to ensure backwards compatibility, while still allowing Moto to be used with modern dotnet projects.

bblommers:
Some GitHub spelunking tells me that this header was introduced in https://github.com/getmoto/moto/pull/1255, to support the XRay SDK.

Considering the [XRay SDK will be EOL](https://github.com/getmoto/moto/issues/9395) in 2027, we can probably remove this header altogether afterwards.

So with that in mind: I would have no problems with an option to disable it now, if it actively blocks users.

Also, as a side note: this is why Moto needs more/better parity tests against AWS! I would love to have a clear picture of which headers are actually send by AWS, how this differs by service and how this changes over time.

rogvc-nice:
@bblommers , I'm happy to get this flag change to an approved state and get it on the earliest possible release since we're using moto and this issue is blocking upgrades for all our apps that use SQS through AWS SDK for dotnet.

What are my next steps here?

bpandola:
Given the additional background information that's been provided, I'm much more in favor of just removing the header altogether now as opposed to some time in the future (with a configuration option between now and then).  Ultimately, I think the .Net SDK's failure to validate the checksum on empty responses is a bug on their side, but the fact that Moto is adding this header to SQS responses in the first place appears to be a bug on our side.

@bblommers I've opened #9473, to be merged pending your comment/approval.

bblommers:
Hi @rogvc-nice , we'll remove the header in the SQS service completely in the next release, so that should unblock you. I expect the release to land somewhere next week, although I can't guarantee an exact date.

I'll close this, considering that #9473 supercedes it. Thank you for the PR and for starting the discussion @rogvc-nice!

PR #9471, opened 2025-11-14
EC2: integrate `Instances` responses with core serializer

bpandola:
Despite a fair bit of attribute adding/renaming in the model, including some of the original (vendored) `boto` classes, this integration ended up being pretty straightforward.

A single test assertion had to be modified where a boolean attribute had been previously hardcoded to `True` in the XML template but is actually `False` in the model (and is now serialized as such).

> [!NOTE]
> This PR errors on the side of preserving existing Moto behavior.  There were a number of hardcoded attribute values of dubious correctness in the XML templates that are now defined in the model.  I have tried to call out these areas with specific comments to help with future debugging/fixing.

### DIFF
--- a/moto/core/serialize.py
+++ b/moto/core/serialize.py
@@ -100,11 +100,13 @@ class SerializationContext:
 
 class TimestampSerializer:
     TIMESTAMP_FORMAT_ISO8601 = "iso8601"
+    TIMESTAMP_FORMAT_ISO8601_ZEROED = "iso8601_zeroed"
     TIMESTAMP_FORMAT_RFC822 = "rfc822"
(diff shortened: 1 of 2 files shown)


moto/ec2/2016-11-15/service-2.moto-extras.json
bpandola, line 1: This is good example of how useful it is to now have the ability to augment/update the Botocore models with Moto "extras".  The EC2 model that ships with Botocore defines the `UserData` parameter as a standard string, even though the documentation is clear that it is a Base64-encoded blob of data.  Here we update the model to ensure that the Moto request parser handles this parameter correctly.

Here's the Botocore code that compensates for the incorrect model:
image

And here's the relevant documentation:
image

PR #9498, opened 2025-11-24
[EC2] Implementing EC2 traffic mirror endpoints

LeekSwan:
This PR implements traffic mirror filters and traffic mirror targets for EC2 resources

Added endpoints:
- create_traffic_mirror_filter
- create_traffic_mirror_target
- describe_traffic_mirror_filter
- describe_traffic_mirror_target

Also implements resource tagging api for both filters and targets


moto/ec2/responses/traffic_mirrors.py
```diff
@@ -0,0 +1,73 @@
...
+
+from ._base_response import EC2BaseResponse
+
+
+class TrafficMirrors(EC2BaseResponse):
+    def create_traffic_mirror_filter(self) -> ActionResult:
+        description = self._get_param("Description")
+        tag_specifications = self._get_param("TagSpecifications")
+        client_token = self._get_param("ClientToken")
+        dry_run = self._get_bool_param("DryRun")
```
bpandola: No need to get the "DryRun" parameter or pass it to the model.  Moto has a built-in helper for this: https://github.com/getmoto/moto/blob/3d3e6856f7a48c4976eafd8000adcda2b3e24dbc/moto/ec2/responses/instances.py#L19

bpandola:
@LeekSwan You still working on this?

LeekSwan:
Hello! Yes I am. Been a bit busy but getting back to it now. Any idea why the checks are failing? They all pass locally and Im having trouble pinpointing the issue.

tests/test_ec2/test_traffic_mirror_resource_group_tagging_api.py
```diff
@@ -0,0 +1,88 @@
...
+        DryRun=False,
+    )
+
+    metadata = resp["ResponseMetadata"]
+    assert metadata["HTTPStatusCode"] == 200
+    assert metadata["RetryAttempts"] == 0
+
+    traffic_mirror_filter = resp["TrafficMirrorFilter"]
+    resource_group_tags = resource_groups_client.get_resources(
+        ResourceARNList=[traffic_mirror_filter["TrafficMirrorFilterId"]],
```
bpandola, line 44: Moto doesn't support this parameter yet, so you end up getting all resources back, which is why the tests are failing (because the ec2 tests are run in parallel and other resources are mixed in from the other tests).

tests/test_ec2/test_traffic_mirror_resource_group_tagging_api.py
```diff
@@ -0,0 +1,88 @@
...
+        DryRun=False,
+    )
+
+    metadata = resp["ResponseMetadata"]
+    assert metadata["HTTPStatusCode"] == 200
+    assert metadata["RetryAttempts"] == 0
+
+    traffic_mirror_target = resp["TrafficMirrorTarget"]
+    resource_group_tags = resource_groups_client.get_resources(
+        ResourceARNList=[traffic_mirror_target["TrafficMirrorTargetId"]],
```
bpandola, line 81: Same here.  Also, I don't think `TrafficMirrorTargetId` is an ARN, is it?  Do these tests work against a real AWS backend?

tests/test_ec2/test_traffic_mirror_resource_group_tagging_api.py
LeekSwan: Got it. Is there someway to add this parameter or is it not feasible atm?

tests/test_ec2/test_traffic_mirror_resource_group_tagging_api.py
LeekSwan: Strange enough, server mode tests ran fine if thats what you mean?

tests/test_ec2/test_traffic_mirror_resource_group_tagging_api.py
bpandola: I mean, I guess that's up to you.  At a glance, I think it might be non-trivial to add, but if you're up for the task go for it.  Otherwise, I would just use one of the other filter parameters that Moto does support, e.g. `ResourceTypeFilters`.

tests/test_ec2/test_traffic_mirror_resource_group_tagging_api.py
bpandola: No, I mean if you ran this against a real AWS backend (not Moto)--if you called `get_resources` with `TrafficMirrorTargetId` in the `ResourceARNList`--would it actually return any results?  I suspect not, since `TrafficMirrorTargetId` is not an ARN... or am I missing something?

bpandola:
Closing due to lack of activity.  @LeekSwan Feel free to re-open if/when you get back to this.

PR #9500, opened 2025-11-25
StepFunctions: start_execution() is only idempotent for running Execu…

bblommers:
## Motivation
Followup to #9397. The `start_execution` is idempotent, but only for running executions. If an execution has finished running (either success or failure), restarting the same execution should still fail.

This PR also adds idempotency to the `start_execution()` method when using the [StepFunctions Parser](https://docs.getmoto.org/en/latest/docs/services/stepfunctions.html), with AWS-validated tests to verify that this now works as expected.

FYI @chriselion, as you raised the original PR:  I believe this change is correct, but please let me know if I've missed anything!


chriselion:
Sounds reasonable to me !

PR #9501, opened 2025-11-26
Autoscaling/base64 encoding

baermat:
With the addtion of `Base64EncodedString` in [this PR](https://github.com/getmoto/moto/pull/9489), the creation of an autoscaling group could fail due to moto attempting to base64 encode a string that was already base64 encoded. This PR addresses this issue

- adds a typecheck
- adds base64 encoded user data to an existing test, ensuring this scenario is covered (the modified test would throw an error prior to the fix).

### DIFF
--- a/moto/ec2/models/launch_templates.py
+++ b/moto/ec2/models/launch_templates.py
@@ -1,5 +1,5 @@
 from collections import OrderedDict
-from typing import Any, Optional
+from typing import Any
 
 from moto.core.common_models import CloudFormationModel
 from moto.core.types import Base64EncodedString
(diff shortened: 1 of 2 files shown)


bpandola:
@baermat Thanks for raising this.  As I noted in the original PR, the `UserData` parameter is tricky to handle because there are just so many ways for it to get into the system. I added the `user_data = Base64EncodedString(user_data)` hack in the model to address a specific use-case, but that wasn't really the right way to go about it (as you have now discovered a scenario in which that line is problematic).  Unfortunately, mypy doesn't really help us here, but that's a separate issue... 
Anyway, I'm going to fix this properly in another PR.  Once we're in the model code, all of our types should be as we expect without the need for any explicit type checks.  
I will keep this PR open until I've opened a suitable replacement.  Thanks again.

PR #9530, opened 2025-12-05
Validate DynamoDB values for number attributes

chriselion:
Resolves https://github.com/getmoto/moto/issues/9522

Adds validation for DynamoDB number attributes:
* value must be a string that's convertable to a python float
  * integer was previously prohibited, float is now also prohibited.
* value must not have any leading or trailing whitespace
* value must be in the ranges described in https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/HowItWorks.NamingRulesDataTypes.html

There may be a few other restrictions that aren't enforced (notably the number of significant digits), but this should hopefully catch a lot of errors before they get to AWS.

---

The failing tests seem unrelated; let me know if you want me to look more closely at them.

### DIFF
--- a/moto/dynamodb/exceptions.py
+++ b/moto/dynamodb/exceptions.py
@@ -382,6 +382,7 @@ class InvalidConversion(JsonRESTError):
         er = "SerializationException"
         super().__init__(er, "NUMBER_VALUE cannot be converted to String")
 
+
 class ValidationException(JsonRESTError):
     def __init__(self, msg: str) -> None:
         er = "ValidationException"
--- a/moto/dynamodb/models/table.py
+++ b/moto/dynamodb/models/table.py
@@ -16,7 +16,7 @@ from moto.dynamodb.exceptions import (
     MockValidationException,
     RangeKeyTooLong,
     SerializationException,
-    ValidationException
+    ValidationException,
 )
 from moto.dynamodb.limits import HASH_KEY_MAX_LENGTH, RANGE_KEY_MAX_LENGTH
 from moto.dynamodb.models.dynamo_type import DynamoType, Item
@@ -561,7 +561,7 @@ class Table(CloudFormationModel):
                     raise SerializationException(
                         "Start of structure or map found where not expected"
                     )
-                
+
     def _validate_number_type(self, value: Any) -> None:
         if isinstance(value, int):
             # TODO int and float values actually raise a ParamValidationError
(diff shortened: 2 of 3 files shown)


tests/test_dynamodb/test_dynamodb_condition_expressions.py
```diff
@@ -292,7 +292,7 @@ def test_condition_expression__attr_doesnt_exist():
     )
 
     client.put_item(
-        TableName=table_name, Item={"forum_name": {"S": "foo"}, "ttl": {"N": "bar"}}
```
chriselion, line 295: This was the only dynamodb test I needed to change

tests/test_dynamodb/exceptions/test_dynamodb_exceptions.py
```diff
@@ -1254,6 +1254,83 @@ def test_put_item_wrong_datatype():
...
+            {"AttributeName": "pk", "KeyType": "HASH"},
+        ],
+        AttributeDefinitions=[
+            {"AttributeName": "pk", "AttributeType": "S"},
+        ],
+        ProvisionedThroughput={"ReadCapacityUnits": 10, "WriteCapacityUnits": 10},
+    )
+
+    # Test putting items with valid values (to make sure we don't break them)
+    valid_values = [
```
chriselion, line 1304: This file like the best file to add the test for checking failing values, but not sure if I should move the check for passing values somewhere else?

moto/dynamodb/exceptions.py
```diff
@@ -383,6 +383,12 @@ def __init__(self) -> None:
         super().__init__(er, "NUMBER_VALUE cannot be converted to String")
 
 
+class ValidationException(JsonRESTError):
```
chriselion, line 386: I saw a similar `ValidationException` in a lot of other modules, but there wasn't one in `dynamodb/` yet.

moto/dynamodb/models/table.py
```diff
@@ -560,6 +562,51 @@ def _validate_item_types(
                         "Start of structure or map found where not expected"
                     )
 
+    def _validate_number_type(self, value: Any) -> None:
+        if isinstance(value, int):
+            # TODO int and float values actually raise a ParamValidationError
+            raise InvalidConversion
+        if isinstance(value, float):
+            raise InvalidConversion
```
chriselion, line 570: This is equivalent to the old behavior. But as noted, I think it deviates from the actual behavior, since that raises a `ParamValidationError`. The actual error is
```
Parameter validation failed:
Invalid type for parameter Item.value.N, value: 42, type: <class 'int'>, valid types: <class 'str'>
```
so I guess we should raise for anything that's not a `str`?

moto/dynamodb/models/table.py
```diff
@@ -560,6 +562,51 @@ def _validate_item_types(
...
+            return
+
+        non_numeric_error = (
+            f"The parameter cannot be converted to a numeric value: {value}"
+        )
+
+        # Quick check - if the value can't be converted to a Python float, consider it invalid
+        try:
+            float_val = float(value)
+        except ValueError:
```
chriselion, line 582: There's a chance that this might cause some false positives, but I figure if it's not a valid python float, it's probably not a valid DynamoDB number either.

moto/dynamodb/models/table.py
```diff
@@ -560,6 +562,51 @@ def _validate_item_types(
...
+        if not math.isfinite(float_val):
+            raise ValidationException(non_numeric_error)
+
+        # More detailed range checks based on these rules:
+        # * Numbers can be positive, negative, or zero. Numbers can have up to 38 digits of precision. Exceeding this results in an exception.
+        # * Positive range: 1E-130 to 9.9999999999999999999999999999999999999E+125
+        # * Negative range: -9.9999999999999999999999999999999999999E+125 to -1E-130
+        # In DynamoDB, numbers are represented as variable length. Leading and trailing zeroes are trimmed.
+        # Source: https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/HowItWorks.NamingRulesDataTypes.html
+        if float_val != 0.0:
```
chriselion, line 598: Note: this covers everybody's favorite weird floating point value, -0.0, too

PR #9531, opened 2025-12-05
Ec2: flow logs type mismatch

baermat:
With update 5.1.17, the parameter "MaxAggregationInterval" for the operation "create_flow_logs" is now apparently passed directly as int, rather than having it converted to a string. This causes a mismatch when validating in the current version when explicitly setting the interval.

Changes:
- change the type of MaxAggregationInterval to int
- add test with explicitly set MaxAggregationInterval (which fails in the current upstream version)
- fix cloudformation test (the cloudformation resource also lists `int` as type in the resource definition docs)

### DIFF
--- a/moto/ec2/responses/flow_logs.py
+++ b/moto/ec2/responses/flow_logs.py
@@ -13,7 +13,7 @@ class FlowLogs(EC2BaseResponse):
         log_destination = self._get_param("LogDestination")
         log_group_name = self._get_param("LogGroupName")
         log_format = self._get_param("LogFormat")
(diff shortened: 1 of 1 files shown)


bpandola (commented):
@baermat For completeness, it's probably best to mark this explicitly as an int in `responses.py`, by using `self._get_int_param` instead of the generic `self._get_param` (which is typed as `Any` and prevents the linter from actually catching the issue you've identified here.)
https://github.com/getmoto/moto/blob/c0d1e5219e1c27643e809e4f56e85c3f19b3de45/moto/ec2/responses/flow_logs.py#L16

PR #9532, opened 2025-12-05
Shield: fix BackendDict service name

bentsku:
While looking at the `shield` service, I've realized that the service name for its `BackendDict` was set to `ec2`. I've tried to check if this was an error or something expected, but this seems to be back from when the service was implemented (#7767), which might mean it was a copy-paste issue. 

This PR sets the service back to what it should be. 

Edit: I now see the test failing due to the region checks, and see this comment: https://github.com/getmoto/moto/pull/7767#discussion_r1648062448

Will look into fixing this, putting back into draft 👀

### DIFF
--- a/moto/shield/models.py
+++ b/moto/shield/models.py
@@ -334,4 +334,6 @@ class ShieldBackend(BaseBackend):
         return self.subscription
 
 
-shield_backends = BackendDict(ShieldBackend, "shield", additional_regions=PARTITION_NAMES)
+shield_backends = BackendDict(
+    ShieldBackend, "shield", additional_regions=PARTITION_NAMES
+)


bblommers:
@bentsku This usually happens because we use `botocore.available_regions(service)` under the hood, and the `shield` service isn't registered there/returns no regions.

Let me know if you want me to do a deep dive though!

bentsku:
> @bentsku This usually happens because we use botocore.available_regions(service) under the hood, and the shield service isn't registered there/returns no regions.
>
> Let me know if you want me to do a deep dive though!

@bblommers ah, nice catch! It seems to indeed be the case here, as shield doesn't seem to be a global service. Do you have an example in how that would be fixed in another service? I've seen `redshift-data` using `redshift` as its service name, but not sure if there would be another way to do that? 

If not, what would you be your preferred way to do it? If you already have an idea and it would be quicker for you to do it than to explain it, feel free to push on the branch 😄
edit: I've seen that I could extend it via `additional_regions`, and I suppose I could pick regions from another service, but not too sure how clean that would be.

bblommers:
@bentsku [Looking at the botocore spec](https://github.com/boto/botocore/blob/e07cddc333fe4fb90efcd5d04324dd83f9cc3a57/botocore/data/endpoints.json#L20591C1-L20617C8), `shield` does look like a global service - maybe this is worth testing.

If it is, we can use the `additional_regions` argument to use the partitions as 'regions', [just like we do in S3](https://github.com/getmoto/moto/blob/c0d1e5219e1c27643e809e4f56e85c3f19b3de45/moto/s3/models.py#L3219C1-L3224C2).

Note that for Moto, this doesn't really matter. We only use the `service_name` for this purpose, so as long as the region names are the same, the exact value is irrelevant.

FWIW, Viren made some similar changes in https://github.com/getmoto/moto/pull/7723, but that was only possible _after_ botocore updated it's spec to include all the regions.

bentsku:
> shield does look like a global service - maybe this is worth testing.

Ah, right, yes does make sense! I could the test using different region, but I'll read more into it and do the appropriate fix then. Thanks for looking into it!

bblommers:
Sorry, closed it by accident. As for a suggested solution: If the service is not global, then the ideal way would probably be to have a second argument, `service_name` and `service_name_for_regions` (or something similar). In most cases they will be the same, but for `shield` we could set that to BackendDict(.., `service_name="shield", service_name_for_regions="ec2")`.

bentsku:
@bblommers you're right, the service is indeed global. 

I've done some tests to verify that it is via the AWS CLI, let me know if you think that is enough:

```bash
# I remove some of the debug logs as they're not relevant, just wanted to show that it would override the endpoint unlike S3 for example
$ aws shield list-protections --region foobar --debug
...
2025-12-05 17:12:06,861 - MainThread - botocore.regions - DEBUG - Using partition endpoint for shield, foobar: aws-global
...
2025-12-05 17:12:06,896 - MainThread - botocore.endpoint - DEBUG - Sending http request: <AWSPreparedRequest stream_output=False, method=POST, url=https://shield.us-east-1.amazonaws.com/, headers={'X-Amz-Target': b'AWSShield_20160616.ListProtections', 'Content-Type': b'application/x-amz-json-1.1', 'User-Agent': b'aws-cli/1.42.31 md/Botocore#1.40.31 ua/2.1 os/macos#25.1.0 md/arch#arm64 lang/python#3.11.6 md/pyimpl#CPython m/b,D,Z,C cfg/retry-mode#legacy botocore/1.40.31', 'X-Amz-Date': b'20251205T161206Z', 'Authorization': b'AWS4-HMAC-SHA256 Credential=<access-key-id>/20251205/us-east-1/shield/aws4_request, SignedHeaders=content-type;host;x-amz-date;x-amz-target, Signature=<signature>', 'amz-sdk-invocation-id': b'57ffb07b-af28-4f6c-bc54-75587c376840', 'amz-sdk-request': b'attempt=1', 'Content-Length': '2'}>

An error occurred (ResourceNotFoundException) when calling the ListProtections operation: The subscription does not exist.
```

I've updated the PR to take this into account, and followed the `route53` model 👍

I've tried to look for tests that would assert than a resource created with a client configured with region A would be accessible from a client with region B, but couldn't really find anything. Technically, in this case, everything would be created in `us-east-1` from the looks of it ([and the endpoints](https://docs.aws.amazon.com/general/latest/gr/shield.html)). 

Would this change be fine as is, or should I add a test for this specific use case? 

edit: sorry for the lint issue, hadn't updated ruff in a while 😓

PR #9539, opened 2025-12-07
Admin: Remove dateutil invocations, replace with datetime/timezone stdlib

bblommers:
The `dateutil`-library is not very actively maintained (if not outright abandoned), and we only use it sparingly.

This PR:
 - Removes any usages where we can easily replace with a stdlib call (like `tzutc()` --> `datetime.timezone.utc`)
 - Replace any calls to `datetime.utcnow` to our custom `utcnow` method, just to remove some deprecation-warnings

There are two places where we still use `dateutil`, specifically the `parse` method:
1. Batch, where we parse the timestamp of Docker logs
2. CodeBuild, where we parse a datetime that we created earlier

The first one may be tricky, as the date format in the Docker logs may be configurable. The second one is definitely tricky, as the logic around date handling is really weird, and may need some refactoring at the same time.

But after those two are handled, we should be able to remove the `dateutil` dependency.

### DIFF
--- a/moto/iot/models.py
+++ b/moto/iot/models.py
@@ -521,7 +521,7 @@ class FakeBillingGroup(CloudFormationModel):
         self.arn = f"arn:{get_partition(self.region_name)}:iot:{self.region_name}:{self.account_id}:billinggroup/{billing_group_name}"
         self.version = 1
         self.things: list[str] = []
-        self.metadata = {"creationDate": datetime.utcnow().isoformat()}
+        self.metadata = {"creationDate": utcnow().isoformat()}
 
     def to_dict(self) -> dict[str, Any]:
         return {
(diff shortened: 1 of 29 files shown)


bpandola:
@bblommers Just a suggestion: for Batch and CodeBuild, you can probably just use [Moto's datetime parser](https://github.com/getmoto/moto/blob/bbc6ac02ee4fba004352bd0d5921108daf1570d1/moto/core/parsers.py#L17).  It's a thin wrapper over Botocore's, which uses `datetutil`; but would allow us to drop `datetutil` as a direct dependency (though of course it will still be brought in transitively).

bblommers:
@bpandola TIL! Thanks for letting me know, I had no idea that `botocore` also uses `dateutil` behind the scenes

Issue #9374, opened 2025-10-15
EC2 update_auto_scaling_group Does Not Update Version when New Version Is Submitted


cloudwitch:
I'm trying to use [update_auto_scaling_group](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/autoscaling/client/update_auto_scaling_group.html) to update my ASGs when there is a new AMI published to the SSM Parameter. However, when writing the unit tests for this I discovered the version does not update on the ASG when using `update_auto_scaling_group` in `moto`. When run against AWS, it updates no problem. I am using a mixed instance policy, but it fails just the same using a regular launch template.

I've included a test script and my output to help with reproduction of the bug.

| Package | Version |
| --- | --- |
| moto | 5.1.14 |
| boto3 | 1.40.52 |
| Python | 3.11 |

Thank you!

## Test script

Run using `python -m unittest test.py`.

```python
"""Tests for moto github issue"""

import os
import unittest
import json

import boto3
from moto import mock_aws


def promote_launch_template_on_asg(asg_client, asg_name: str, launch_template: dict) -> dict:
    """Update mixed instance policy launch template to latest version."""
    asg_client.update_auto_scaling_group(
        AutoScalingGroupName=asg_name,
        MixedInstancesPolicy={
            "LaunchTemplate": {
                "LaunchTemplateSpecification": {
                    "LaunchTemplateName": launch_template["LaunchTemplateName"],
                    "Version": str(launch_template["VersionNumber"]),
                }
            }
        },
    )
    return asg_client.describe_auto_scaling_groups(AutoScalingGroupNames=[asg_name])["AutoScalingGroups"][0]


@mock_aws
class TestMain(unittest.TestCase):
    def setUp(self):
        os.environ["AWS_REGION"] = "us-east-1"
        os.environ["AWS_ACCESS_KEY_ID"] = "testing"
        os.environ["AWS_SECRET_ACCESS_KEY"] = "testing"
        os.environ["AWS_SECURITY_TOKEN"] = "testing"
        os.environ["AWS_SESSION_TOKEN"] = "testing"
        os.environ["AWS_DEFAULT_REGION"] = "us-east-1"
        os.environ["ACCOUNT_NUMBER"] = "123456789012"
        os.environ["ACCOUNT_ID"] = "123456789012"
        self.mock_ec2_client_east = boto3.client("ec2", region_name="us-east-1")
        self.mock_asg_client_east = boto3.client("autoscaling", region_name="us-east-1")
[...]

PR #9543, opened 2025-12-09
EC2: Verify LaunchTemplates tests against AWS

bblommers:
I was looking at #9374, and noticed a distinct lack of `aws_verified` tests in this area - that's why started with verifying that our existing tests run against AWS.

While doing that I found a small bug here, where the `DefaultVersion` wasn't correctly set/computed

### DIFF
--- a/tests/test_ec2/test_launch_templates.py
+++ b/tests/test_ec2/test_launch_templates.py
@@ -1,3 +1,4 @@
+from functools import wraps
 from uuid import uuid4
 
 import boto3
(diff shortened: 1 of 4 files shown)


tests/test_ec2/test_launch_templates.py
```diff
@@ -824,6 +857,8 @@ def test_modify_launch_template_by_name():
 
 @mock_aws
 def test_create_launch_template_with_non_base64_encoded_user_data_fails():
+    # XXX: This doesn't actually fail against AWS.
+    # Maybe it only fails when trying to create an instance from it? Something to play around with
```
bpandola, line 861: This is definitely a change in behavior.  When I was working on #9510, I spent (way too much) time testing all the various scenarios against AWS.  Here was the raw response from my notes:
`b'<?xml version="1.0" encoding="UTF-8"?>\n<Response><Errors><Error><Code>InvalidUserData.Malformed</Code><Message>Invalid BASE64 encoding of user data.</Message></Error></Errors><RequestID>55c3cc24-c745-4477-b01d-8573ee76bfd3</RequestID></Response>'`
But, as you point out: I just tested this again and AWS happily accepts the non-base64-encoded string...  
Buuuut, I was able to get this to fail against real AWS by changing the input to a string that contains obviously invalid base64 characters:
`"UserData": "not base64 encoded:*&^%$"`
So if you make that change, hopefully you can slap the `aws_verified` marker on that test and have it succeed.

bblommers: Ah, interesting. I wonder if AWS simply uses a regex validation for this then.

Either way, I'll probably leave this as-is for now, just to focus on the Autoscaling/Launch templates in general - I'll try to come back to this later :+1:

PR #9552, opened 2025-12-14
fix(ec2): ensure BlockDeviceMapping returns 'attached' status instead 'in-use'

### LINKED ISSUE #9517: EbsInstanceBlockDevice has invalid status field
Hey,

According to the [docs](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_EbsInstanceBlockDevice.html), the possible valid value for the `status` field in `EbsInstanceBlockDevice` is one of `attaching | attached | detaching | detached`.

Using `describe_instances` in Moto (v5.1.18) returns `in-use` for the EBS status field:

```python
with moto.mock_aws():  
    import boto3  
    client = boto3.client('ec2')  
    vpc_response = client.create_vpc(CidrBlock="10.0.0.0/16")  
    vpc_id = vpc_response["Vpc"]["VpcId"]  
  
    subnet_response = client.create_subnet(VpcId=vpc_id, CidrBlock="10.0.0.0/24")  
    subnet_id = subnet_response["Subnet"]["SubnetId"]
  
    instances_response = client.run_instances(  
        ImageId='ami-00000000',  
        MinCount=2,  
        MaxCount=2,  
        InstanceType='t2.micro',  
        SubnetId=subnet_id  
    ) 
  
    response = client.describe_instances()  
    for reservation in response['Reservations']:  
        for instance in reservation['Instances']: 
            # This will print `in-use`, instead of `attached`
            print(instance['BlockDeviceMappings'][0]['Ebs']['Status'])
```

Seems like `in-use` is only valid for the `Volume` [response](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_Volume.html).

jiaming2li:
## Problem
Fixes #9517

Currently, when calling `describe_instances`, the `BlockDeviceMappings.Ebs.Status` field returns the underlying volume status (e.g., `in-use`).
According to AWS documentation and actual behavior, this field should reflect the attachment status (e.g., `attached`), not the volume status.

## Solution
Modified the `block_device_mappings` property in `moto/ec2/models/instances.py`.
It now checks if the volume status is `in-use` and maps it to `attached` for the instance response, ensuring client compatibility.

## Verification
Verified locally with a reproduction script.
- Before fix: Status was `in-use`.
- After fix: Status is `attached` (as expected).

### DIFF
--- a/tests/test_ec2/test_instances.py
+++ b/tests/test_ec2/test_instances.py
@@ -3093,3 +3093,15 @@ def test_run_instances_default_response():
     assert "Tags" not in instance
     assert instance["VirtualizationType"] == "paravirtual"
     assert "VpcId" in instance
+
+
+def test_block_device_status_conversion():
+    """Test EBS block device status conversion."""
+    from moto.ec2.models.instances import Instance
+
+    assert Instance.get_block_device_status("in-use") == "attached"
+
+    assert Instance.get_block_device_status("available") == "detached"
+
+    assert Instance.get_block_device_status("creating") == "creating"
+    assert Instance.get_block_device_status("deleting") == "deleting"


moto/ec2/models/instances.py
```diff
@@ -73,6 +73,27 @@ def __init__(self, options: Optional[dict[str, Any]] = None):
         self.instance_metadata_tags = options.get("InstanceMetadataTags", "disabled")
 
 
+class BlockDeviceWrapper:
```
bblommers: Can we use a regular method here instead? That would be easier to read IMO, and would achieve the same thing.

Something like:
```python
def get_block_status(volume_status):
    if volume_status == "in-use":
        return "attached"
    ...
```

moto/ec2/models/instances.py
```diff
@@ -61,7 +61,7 @@ def __init__(self, message: str = "", code: str = ""):
 
 
 class MetadataOptions:
-    def __init__(self, options: Optional[dict[str, Any]] = None):
+    def __init__(self, options: dict[str, Any] | None = None):
```
bblommers: Moto still supports Python 3.9, and the `|`-operator was only introduced in Python3 .10 - so we're going to have to stick with the `Optional` notation for now

PR #9582, opened 2025-12-22
fix(authorization): update body parameter to use raw_body in authentication method

BrandonLWhite:
This PR fixes https://github.com/getmoto/moto/issues/9580 by changing the expected signature generated as part of `_authenticate_and_authorize_action` to use the original `raw_body` instead of the parsed and transformed body.

### DIFF
--- a/tests/test_dynamodb/test_dynamodb_iam_permission.py
+++ b/tests/test_dynamodb/test_dynamodb_iam_permission.py
@@ -63,15 +63,14 @@ def test_dynamodb_authorization_put_item():
     ]
 
     with enable_iam_authentication():
-        restricted_session = boto3.Session(
+        restricted_dynamo = boto3.resource(
+            "dynamodb",
             region_name="us-east-1",
             aws_access_key_id=credentials["AccessKeyId"],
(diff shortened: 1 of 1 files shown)


BrandonLWhite:
Sorry for the repeated need for CI runs.  I am still figuring out how to do some of these checks locally.
I think I have the region issue taken care of.
However, server mode test still fails for me locally.
```
TEST_SERVER_MODE=true pytest -sv ./tests/test_dynamodb/test_dynamodb_iam_permission.py
```

I am getting `botocore.exceptions.ClientError: An error occurred (UnrecognizedClientException) when calling the PutItem operation: The security token included in the request is invalid.` error.

I'm looking into it, but if anyone has any hints please advise!

BrandonLWhite:
Ok I think I got it worked out.  It seems that `ServerModeMockAWS` doesn't support creating `boto3.Session` instances.  It only patches `boto3.client` and `boto3.resource`.  Therefore, the `endpoint_url` wasn't getting overridden in my test when running with `TEST_SERVER_MODE=true`.

bblommers (approved):
Nice find @BrandonLWhite. Thank you for fixing this and for adding a test case for it!

PR #9585, opened 2025-12-23
Fix DynamoDB filtering for NULL types

chriselion:
Fixes https://github.com/getmoto/moto/issues/9560

There were two separate, but similar problems - both related to treating `{"NULL": True}` as `True` instead of `None`.

The change in `DynamoType.cast_value` looks like it could have broader scope, so let me know if you'd like more tests for it. 

---

Related, I also added a validation on the "value" for NULL attribute - boto checks that type is `boolean`, but the only allowed value is `True`, so I added that.

### DIFF
--- a/moto/dynamodb/models/table.py
+++ b/moto/dynamodb/models/table.py
@@ -561,6 +561,13 @@ class Table(CloudFormationModel):
                     raise SerializationException(
                         "Start of structure or map found where not expected"
                     )
+            elif key == "NULL":
+                if isinstance(value, bool):
+                    if value is not True:
+                        raise ValidationException(
+                            "Null attribute value types must have the value of true"
+                        )
+                # Non-bool values are raise elsewhere as a ParamValidationError
 
     def _validate_number_type(self, value: Any) -> None:
         if isinstance(value, int):
--- a/tests/test_dynamodb/exceptions/test_dynamodb_exceptions.py
+++ b/tests/test_dynamodb/exceptions/test_dynamodb_exceptions.py
@@ -1333,6 +1333,34 @@ def test_put_item_invalid_number_types(value, expected_error_template):
     assert err["Message"] == expected_error_template.format(value=value)
 
 
+@mock_aws
+def test_put_item_invalid_null():
+    """
+    Test that passing a "False" value for a NULL attribute raises an exception.
+    """
+    client = boto3.client("dynamodb", region_name="us-east-1")
+    table_name = f"T{uuid4()}"
+    client.create_table(
+        TableName=table_name,
+        KeySchema=[
+            {"AttributeName": "pk", "KeyType": "HASH"},
+        ],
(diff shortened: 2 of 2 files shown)


tests/test_dynamodb/test_dynamodb.py
```diff
@@ -1333,6 +1333,53 @@ def test_filter_expression():
...
+    # equivalent to (is_visible.exists()) & (is_visible != None)
+    filter_expr = moto.dynamodb.comparisons.get_filter_expression(
+        "(attribute_exists (is_visible) AND is_visible <> :0)",
+        names=None,
+        values={":0": {"NULL": True}},
+    )
+    assert filter_expr.expr(row_no_boolean) is False
+    assert filter_expr.expr(row_null_boolean) is False
+    assert filter_expr.expr(row_boolean_false) is True
+    assert filter_expr.expr(row_boolean_true) is True
```
chriselion, line 1369: This case was broken without the `AttributeValue.expr` fix.

tests/test_dynamodb/test_dynamodb.py
```diff
@@ -1333,6 +1333,53 @@ def test_filter_expression():
...
+
+    # Equivalent to (is_visible.exists()) & (is_visible == True)
+    filter_expr = moto.dynamodb.comparisons.get_filter_expression(
+        "(attribute_exists (is_visible) AND is_visible = :0)",
+        names=None,
+        values={":0": {"BOOL": True}},
+    )
+    assert filter_expr.expr(row_no_boolean) is False
+    assert filter_expr.expr(row_null_boolean) is False
```
chriselion, line 1378: This case was broken without the `DynamoType.cast_value` fix

moto/dynamodb/models/table.py
```diff
@@ -561,6 +561,13 @@ def _validate_item_types(
                     raise SerializationException(
                         "Start of structure or map found where not expected"
                     )
+            elif key == "NULL":
+                if isinstance(value, bool):
+                    if value is not True:
+                        raise ValidationException(
+                            "Null attribute value types must have the value of true"
```
chriselion, line 568: Full exception:
```
False
botocore.exceptions.ClientError: An error occurred (ValidationException) when calling the PutItem operation: One or more parameter values were invalid: Null attribute value types must have the value of true
```
I think we just want the last part here, right?

moto/dynamodb/models/table.py
```diff
@@ -561,6 +561,13 @@ def _validate_item_types(
...
+                        )
+                # Non-bool values are raise elsewhere as a ParamValidationError
```
chriselion, line 570: This is already caught on the client side:
```
botocore.exceptions.ParamValidationError: Parameter validation failed:
Invalid type for parameter Item.value.NULL, value: foobar, type: <class 'str'>, valid types: <class 'bool'>
```

PR #9587, opened 2025-12-24
Remove pylint mention from docs

chriselion:
Pylint was removed in https://github.com/getmoto/moto/pull/8891 (yay!), but the developer docs still say to use it for linting ([here](https://docs.getmoto.org/en/latest/docs/contributing/installation.html)). I removed that, and also cleaned up all the remaining inline comments.

Before:
```shell
$ git grep -i pylint | wc -l
      21
```

After:
```shell
$ git grep -i pylint | wc -l
      1
``` 
(the remaining string is in the [changelog](https://github.com/getmoto/moto/blob/master/CHANGELOG.md#310))


chriselion:
It looks like the test failures are also happening on master (for example, here https://github.com/getmoto/moto/actions/runs/20475745551).

bpandola:
> It looks like the test failures are also happening on master (for example, here https://github.com/getmoto/moto/actions/runs/20475745551).

This should be fixed now.  You can rebase on or merge in `master` to update your branch.

bblommers (approved):
Nice - thanks for the cleanup!

PR #9588, opened 2025-12-24
 Upgrade ruff to 0.14.10 (current latest).

chriselion:
Inspired by https://github.com/getmoto/moto/pull/9587 and https://github.com/getmoto/moto/pull/9585, I decided to see what would happen if I updated to the latest version of ruff. In summary:

* It started complaining about `Optional[Foo]`. I disabled that for now, until you drop python3.9 support (similar to another rule).
* It rewrote `from typing import Callable` to `from collections.abc import Callable`. That looked safe: https://docs.python.org/3.9/library/collections.abc.html#collections.abc.Callable
* It started flagging cases where `zip()` didn't get the `strict` argument passed. This isn't available in python3.9, but setting `target-version = "py39"` in the ruff config avoids it for now.
  * While investigating this, it uncovered a bug in a test case (caused by unequal length args to zip), so I fixed that. I'll add a comment inline.

### DIFF
--- a/moto/backends.py
+++ b/moto/backends.py
@@ -6,7 +6,7 @@ from typing import TYPE_CHECKING, Optional, Union, overload
 import moto
 
 if TYPE_CHECKING:
-    from typing import Literal
+    from typing_extensions import Literal
 
     from moto.acm.models import AWSCertificateManagerBackend
     from moto.acmpca.models import ACMPCABackend


moto/ec2/utils.py
```diff
@@ -30,7 +30,7 @@
 from moto.utilities.utils import md5_hash
 
 if TYPE_CHECKING:
-    from typing_extensions import TypeAlias
+    from typing import TypeAlias
```
chriselion: This shouldn't have been changed, I'll revert it.

tests/test_logs/test_logs_filter_log_events.py
```diff
@@ -26,9 +26,10 @@ def setUp(self) -> None:
         super().setUp()
 
     def test_filter_logs_interleaved(self):
+        timestamp = int(unix_time_millis())
```
chriselion, line 29: This was actually a bug in the test - since the timestamp was old, the events didn't get saved, so no events were returned, so the `zip()` didn't iterate over anything. Adding `strict=True` caused an exception because the lists weren't the same length.

moto/backends.py
```diff
@@ -6,7 +6,7 @@
 import moto
 
 if TYPE_CHECKING:
-    from typing_extensions import Literal
+    from typing import Literal
```
chriselion: This is OK since `Literal` is in py3.9: https://docs.python.org/3.9/library/typing.html#typing.Literal

moto/ec2/utils.py
chriselion: Restored.

pyproject.toml
```diff
@@ -15,6 +15,9 @@ markers = [
     "requires_clean_slate: requires a clean slate, so no parallel testing",
 ]
 
+[tool.ruff]
+target-version = "py39"  # Same as python_requires in setup.cfg
```
chriselion, line 19: This prevents ruff from trying to replace `from typing_extensions import TypeAlias` with `from typing import TypeAlias`. Strangely, you still need to disable the `UP007` and `UP045` rules, even though they should only apply for python>=3.10

chriselion: When do drop 3.9 support, you'll want to change this too. Ruff could infer it from pyproject.toml settings, but looks like it doesn't try to read it from setup.cfg.

tests/test_logs/test_logs_filter_log_events.py
bblommers: Nice catch!

bblommers (changes requested):
Hi @chriselion! I have looked into this before, but (IIRC) gave up because of the issues with Python 3.9. Thanks for persevering and bringing us up to date!

We'll still want to run the tests using Python 3.9 though, so we cannot yet support `strict=True` here.  If can we revert those (`noqa` them all), then I'd be happy to merge this

chriselion:
@bblommers  Argh, I didn't have a chance to look at the test errors, and forgot `strict=` wasn't in 3.9.

How about I disable the lint rule for in the config for now, with a similar comment as the others about 3.9? It's also possible that adding `target-version = "py39"` will stop if from being enforced too (I didn't add that until after I had made the changes). I'll keep the fix for that test regardless.

Is there a timeline for dropping 3.9? Looks like `boto3` will drop it soon, according to warnings in the test logs (for example, [here](https://github.com/getmoto/moto/actions/runs/20510377333/job/58930727694))

bblommers:
> How about I disable the lint rule for in the config for now, with a similar comment as the others about 3.9?

Yep, that works as well! I don't have a strong preference - as long as the tests pass :slightly_smiling_face: 

> Is there a timeline for dropping 3.9?

We tend to only drop support for Python versions after ~= 6 months, to give users plenty of time to upgrade. Given that 3.9 was EOL end of October, we will probably drop it around March/April 2026. So it's a good thing that the comments are there, to remind us to look into these rules!

moto/backends.py
chriselion: Reverting this (even though it's probably safe), since ruff doesn't complain about it after setting the python version.

chriselion:
@bblommers Sounds good, we should be in good change now. I'll check back in a bit and make sure the tests passed.

Let me know if you want a hand with the 3.10 upgrade when the time comes around.

bblommers (approved):
LGTM - thank you so much @chriselion!

PR #9596, opened 2025-12-28
Autoscaling: Improve parity around LaunchTemplates

bblommers:
A ton of Autoscaling tests were rewritten to make sure they behave the same when ran against AWS (and marked with the `aws_parity` to mark them as such.

Related:  #9374 , #9543

### DIFF
--- a/tests/test_autoscaling/__init__.py
+++ b/tests/test_autoscaling/__init__.py
@@ -1,1 +1,78 @@
-# This file is intentionally left blank.
+from contextlib import nullcontext
+from functools import wraps
+from uuid import uuid4
+
+import boto3
+import pytest
+
+from moto import mock_aws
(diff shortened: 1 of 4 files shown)


tests/test_autoscaling/__init__.py
```diff
@@ -1 +1,78 @@
...
+    get_group_name: bool,
+    func,
+    kwargs,
+):
+    autoscaling_client = boto3.client("autoscaling", "us-east-1")
+    kwargs["autoscaling_client"] = autoscaling_client
+
+    autoscaling_group_name = None
+    if create_auto_scaling_group:
+        # TODO - actually create one
```
bblommers, line 47: For now we always create the AutoScalingGroup manually (inside the test), so this isn't necessary yet

PR #9601, opened 2025-12-30
[APIGateway] Enhancement to the apply_patch_operations

ApogeeAC:
This PR adds the ability for end-users to apply updates to their `endpoint_configuration` via the `update_rest_api` call. Per the [docs](https://docs.aws.amazon.com/apigateway/latest/api/patch-operations.html) the UpdateRestApi only supports the replace operation and the value must be one of the three types.

Currently `update_rest_api` does not allow a user to update their endpointConfigurations.


bblommers (approved):
LGTM! Thanks @ApogeeAC

PR #9611, opened 2026-01-03
Core: add `PaginatedResult` result wrapper class

bpandola:
This PR adds a `PaginatedResult` result wrapper class for use with the new request parsing/response serialization pipeline.  This class can be used as a drop-in replacement for `ActionResult` to automatically paginate the wrapped result.  The `PaginatedResult` class leverages the existing Moto paginator utility methods along with Moto's ability to load Botocore models with Moto-specific extras in order to provide an extremely simple/low-code abstraction for adding paging to existing API endpoints.

This PR also includes reference implementations for a few existing backends (EMR, ElastiCache, RDS, and SES).

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -9,6 +9,7 @@ include moto/ec2/resources/latest_amis/*.json
 include moto/ec2/resources/amis.json
 include moto/ec2/resources/ecs/optimized_amis/*.json
 include moto/elasticache/*/*.json
+include moto/emr/*/*.json
 include moto/emr/resources/*.json
 include moto/cognitoidp/resources/*.json
 include moto/dynamodb/parsing/reserved_keywords.txt
--- a/moto/emr/2009-03-31/paginators-1.moto-extras.json
+++ b/moto/emr/2009-03-31/paginators-1.moto-extras.json
@@ -1,0 +1,27 @@
+{
+  "version": 1.0,
+  "merge": {
+    "pagination": {
+      "ListBootstrapActions": {
+        "limit_default": 50,
+        "unique_attribute": "name"
+      },
+      "ListClusters": {
+        "limit_default": 50,
+        "unique_attribute": "id"
+      },
+      "ListInstanceGroups": {
+        "limit_default": 50,
+        "unique_attribute": "id"
+      },
+      "ListInstances": {
+        "limit_default": 50,
+        "unique_attribute": "id"
+      },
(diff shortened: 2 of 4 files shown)


bblommers (approved):
Nice! Using the builtin pagination config from `botocore` should simplify this feature quite a bit for contributors.

**nitpick**: It would be nice to have this updated in [the documentation](https://docs.getmoto.org/en/latest/docs/contributing/development_tips/utilities.html#paginator) as well, to teach people how to use this feature.

**question**: Can we make this behaviour explicit, without having to call the `PaginatedResult`-class? Not burdened by any inside knowledge on how this would work - but theoretically we could just extend the `ActionResult` behaviour to check if the current operation should be paginated (by looking at the `paginators.json` file), and act accordingly.

bpandola:
@bblommers 
> Nice! Using the builtin pagination config from `botocore` should simplify this feature quite a bit for contributors.
> 
With all of these higher-level abstractions, I'm trying to pull everything possible out of each backend that isn't directly related to the service itself.  Absolving contributors from having to think about serialization, pagination, etc. should make it easier to contribute.  That's the hope, anyway.

> **nitpick**: It would be nice to have this updated in [the documentation](https://docs.getmoto.org/en/latest/docs/contributing/development_tips/utilities.html#paginator) as well, to teach people how to use this feature.
> 
No problem.  I did want to avoid having multiple methods documented, so (eventually) my plan is to update the docs once all the existing `@paginate` decorators are migrated to `PaginatedResult`.

> **question**: Can we make this behaviour explicit, without having to call the `PaginatedResult`-class? Not burdened by any inside knowledge on how this would work - but theoretically we could just extend the `ActionResult` behaviour to check if the current operation should be paginated (by looking at the `paginators.json` file), and act accordingly.
>
Short answer: Yes.
Longer answer: I did carefully consider the implicit/explicit tradeoffs when working on this (it's in the Zen of Python after all...).  My feeling was that the added clarity of seeing `PaginatedResult` in a `responses.py` method was worth the tiny amount of extra typing.  It also serves as an explicit check: changing `ActionResult` to `PaginatedResult` says "I've made the required update to the model data and now this can be automatically paginated."  If the entire mechanism was implicit, it would be impossible to tell, from looking at the code alone, whether or not a response was paginated or even intended to be.
(I'm not necessarily opposed to this possibly being made more implicit in the future; I'm just explaining my rationale for why it's not right now.)

PR #9612, opened 2026-01-03
feat[ec2] Implemented subnet cidr reservation functions

peteherman:
- implemented:
  - `create-subnet-cidr-reservation`
  - `get-subnet-cidr-reservations`
  - `delete-subnet-cidr-reservation`
  
The functions are used in the management of prefix/explicit subnet cidr reservations.

These are net-new functions. They were previously not implemented/supported.

### DIFF
--- a/.github/workflows/tests_sdk_ruby.yml
+++ b/.github/workflows/tests_sdk_ruby.yml
@@ -13,7 +13,7 @@ jobs:
     steps:
       - uses: actions/checkout@v6
       - name: Set up Ruby ${{ matrix.ruby-version }}
-        uses: ruby/setup-ruby@ae195bbe749a7cef685ac729197124a48305c1cb
+        uses: ruby/setup-ruby@b90be12699fdfcbee4440c2bba85f6f460446bb0
         with:
           ruby-version: ${{ matrix.ruby-version }}
       - name: Set up Python 3.12
--- a/moto/redshift/models.py
+++ b/moto/redshift/models.py
@@ -649,12 +649,20 @@ class RedshiftBackend(BaseBackend):
         return self.clusters[cluster_id]
 
     def describe_clusters(
-        self, cluster_identifier: Optional[str] = None
+        self,
+        cluster_identifier: Optional[str] = None,
+        tag_keys: Optional[list[str]] = None,
     ) -> list[Cluster]:
         if cluster_identifier:
             if cluster_identifier in self.clusters:
                 return [self.clusters[cluster_identifier]]
             raise ClusterNotFoundError(cluster_identifier)
+        elif tag_keys:
+            return [
+                cluster
+                for cluster in self.clusters.values()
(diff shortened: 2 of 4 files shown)


moto/ec2/models/subnets.py
```diff
@@ -17,21 +18,35 @@
...
 
+class SubnetCidrReservation(TypedDict):
+    SubnetCidrReservationId: str
+    SubnetId: str
+    Cidr: str
+    ReservationType: Literal["explicit"] | Literal["prefix"]
+    OwnerId: str
+    TagSpecifications: list[dict[str, Any]]
+
+
```
bpandola, line 70: I get that there are often advantages to using a TypedDict instead of an object, but in this case I think it's worth following the existing convention and making this a proper object.

moto/ec2/utils.py
```diff
@@ -514,7 +512,7 @@ def filter_internet_gateways(
 
 
 def is_filter_matching(obj: Any, _filter: str, filter_value: Any) -> bool:
-    value = obj.get_filter_value(_filter)
+    value = obj[_filter] if isinstance(obj, dict) else obj.get_filter_value(_filter)
```
bpandola: Case in point: there's no need to add this one-off conditional into the existing code if you make your `SubnetCidrReservation` a proper object.

bpandola (changes requested):
@peteherman This looks good.  I appreciate the comprehensive test coverage.  If you can follow the existing convention with your new model class (SubnetCidrReservation), I'm happy to merge.

moto/ec2/models/subnets.py
peteherman: Thanks for the quick review! I've updated the implementation so `SubnetCidrReservation` is now a `TaggedEC2Resource`. Let me know if this isn't what you had in mind and I can make another change 👍

moto/ec2/exceptions.py
bpandola, line 1: A couple notes here regarding exceptions:

1) The AWS client SDKs add the "An error occurred..." prefix to error messages.  Your exception messages should only include the part after the colon.  If you generate one of these exceptions as is, you'll see that Boto3 shows the duplication: `ClientError("An error occurred (InvalidSubnetCidrReservationID.NotFound) when calling the DeleteSubnetCidrReservation operation: An error occurred (InvalidSubnetCidrReservationID.NotFound) when calling the DeleteSubnetCidrReservation operation: The subnet-cidr-reservation ID 'scr-not-exists' does not exist")`.

2) Your exceptions are using a generic `message` parameter, but they seem to expect a more specific value, e.g. a CIDR block or Reservation Id.  Consider naming these parameters more appropriately.

peteherman: Thanks again for another review! I've just made these changes

bpandola (approved):
Thanks @peteherman!

PR #9625, opened 2026-01-06
Support additional ACM key types (RSA key sizes, ECC)

jtschladen:
Adds support for addition key types in ACM, including more RSA key lengths (3072, 4096) and more ECC curves (secp384r1, secp521r1)

This should put moto at parity with all currently supported ACM key types: https://docs.aws.amazon.com/acm/latest/APIReference/API_Filters.html
> Valid Values: RSA_1024 | RSA_2048 | RSA_3072 | RSA_4096 | EC_prime256v1 | EC_secp384r1 | EC_secp521r1


bblommers (approved):
This is great - thanks for getting us to 100% parity here @jtschladen!

PR #9627, opened 2026-01-06
ResourceGroups: Sync Tasks

max-russ:
Implementing resource group tag sync operations

### DIFF
--- a/.github/workflows/tests_sdk_ruby.yml
+++ b/.github/workflows/tests_sdk_ruby.yml
@@ -13,7 +13,7 @@ jobs:
     steps:
       - uses: actions/checkout@v6
       - name: Set up Ruby ${{ matrix.ruby-version }}
-        uses: ruby/setup-ruby@b90be12699fdfcbee4440c2bba85f6f460446bb0
+        uses: ruby/setup-ruby@675dd7ba1b06c8786a1480d89c384f5620a42647
         with:
           ruby-version: ${{ matrix.ruby-version }}
       - name: Set up Python 3.12
--- a/moto/core/serialize.py
+++ b/moto/core/serialize.py
@@ -1224,10 +1224,20 @@ class ShapeNameAlias(AttributeAliasProvider):
 
     def has_alias(self, key: str) -> bool:
         shape = self.context.shape
-        if shape is not None:
-            if shape.type_name in ["list", "structure"]:
-                if key != shape.name:
-                    return True
+        if shape is None:
+            return False
+        if shape.type_name not in ["list", "structure"]:
+            return False
+        try:
+            if shape.name in shape.parent.members:  # type: ignore[attr-defined]
+                # If the name of the shape conflicts with a sibling key,
+                # we don't want to use it as an alias because it will
+                # pick up the sibling value.
+                return False
+        except AttributeError:
+            pass
+        if key != shape.name:
+            return True
         return False
 
     def get_alias(self, key: str) -> Any:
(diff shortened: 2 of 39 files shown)


max-russ:
Server tests seem to be failing on 404's on resource groups' get_tags calls; I don't see anywhere in my commit where those would have been affected.

moto/resourcegroups/urls.py
```diff
@@ -15,4 +15,8 @@
     "{0}/resources/(?P<resource_arn>[^/]+)/tags$": ResourceGroupsResponse.dispatch,
```
bblommers, line 15: Hi @Peeanio! Server Mode is a bit finicky about it's URL's. Because the ARN now has a slash in them (`..:group/{name})`, we also need to add a dedicated URL to account for the scenario where the URL has an additional slash:

> "{0}/resources/(?P<arn_prefix>[^/]+)/(?P<group_name>[^/]+)/tags$": ResourceGroupsResponse.dispatch,

If you add that URL, the tests should pass

max-russ: Thanks for the advice on where the URLs were off after my ARN change.

moto/resourcegroups/models.py
```diff
@@ -393,5 +462,54 @@ def put_group_configuration(
...
+    def list_tag_sync_tasks(
+        self,
+        next_token: str,
+        filters: Optional[list[dict[str, str]]] = None,
+        max_results: Optional[int] = None,
+    ) -> tuple[list[dict[str, Any]], str]:
+        tag_sync_tasks = []
+        for task in self.tag_sync_tasks.values():
+            tag_sync_tasks.append(task.as_dict())
+        return tag_sync_tasks, next_token
```
bblommers: If users want to use the pagination, it's typically used in a while loop:
```python
tasks = client.list_tag_sync_tasks()
while tasks.has("NextToken"):
    tasks = client.list_tag_sync_tasks(NextToken=..)
```

If we always return the same token, then users will end up in an infinite loop.

Considering we do not support pagination yet, we can just skip this value, and not return it to the user.

moto/resourcegroups/responses.py
```diff
@@ -180,3 +181,42 @@ def put_group_configuration(self) -> str:
...
+        tag_sync_task = self.resourcegroups_backend.get_tag_sync_task(
+            task_arn=task_arn,
+        )
+        return json.dumps(tag_sync_task, default=str)
+
+    def cancel_tag_sync_task(self) -> None:
+        task_arn = self._get_param("TaskArn")
+        self.resourcegroups_backend.cancel_tag_sync_task(
+            task_arn=task_arn,
+        )
```
bblommers, line 196: Nitpick - but for consistency, we should always return something to the user IMHO, even if it's an empty string.

moto/resourcegroups/models.py
max-russ: Thanks for the point out, removed using the token.

moto/resourcegroups/responses.py
max-russ: boto doesn't return anything normally here, and the response has the normal response metadata so matches what the real api returns.
https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/resource-groups/client/cancel_tag_sync_task.html

bblommers:
Looks like GIthub isn't picking up the latest commit - I've reopened the PR to see if that fixes it...

bblommers (approved):
Great - thank you @Peeanio!

PR #9628, opened 2026-01-06
feat(organizations): Add Resource Control Policy support

patricksanders:
This PR adds support for [Resource Control Policies](https://docs.aws.amazon.com/organizations/latest/userguide/orgs_manage_policies_rcps.html) in the Organizations service. RCPs are very similar to SCPs, so I tried to keep things consistent.

With this implementation:
* RCPs will be enabled by default when creating a new organization in the moto backend.
* a default RCP will be created with the same name that AWS uses (`RCPFullAWSAccess`).
* the default RCP will be attached to accounts and OUs created in the organization.
* calls to `organizations.list_policies()` will honor the `Filter` parameter.

I opted to duplicate many of the SCP tests and update them for RCP. I considered parametrizing the existing SCP tests, but after doing a few of them I didn't think the test code deduplication justified the additional complexity. I also added some net-new test cases while I was in there, like validating that the default policies are attached to new accounts/OUs.

In case it's important to the maintainers, this was all done by a human. 😁


moto/organizations/models.py
```diff
@@ -740,9 +772,28 @@ def attach_policy(self, **kwargs: Any) -> None:
         else:
             raise InvalidInputException("You specified an invalid value.")
 
-    def list_policies(self) -> dict[str, Any]:
+    def list_policies(self, **kwargs: Any) -> dict[str, Any]:
```
bblommers:
```suggestion
    def list_policies(self, policy_type: str) -> dict[str, Any]:
```

Can we make the parameter explicit here? It's a minor thing, but that makes it much more readable IMHO

(`policy_type` is just a suggestion, as that is makes more obvious what it does than the generic name `filter` - but I'm open to suggestions here)

moto/organizations/responses.py
```diff
@@ -162,7 +162,9 @@ def attach_policy(self) -> str:
         return "{}"
 
     def list_policies(self) -> str:
-        return json.dumps(self.organizations_backend.list_policies())
+        return json.dumps(
+            self.organizations_backend.list_policies(**self.request_params)
```
bblommers: You can just the utility method `_get_param` to find the value of `Filter`:
```python
policy_type=self._get_param("Filter")
```

moto/organizations/models.py
```diff
@@ -740,9 +772,28 @@ def attach_policy(self, **kwargs: Any) -> None:
...
+        _filter = kwargs["Filter"]
+
+        if not FakePolicy.supported_policy_type(_filter):
+            raise InvalidInputException("You specified an invalid value.")
+
+        if _filter not in [
```
bblommers: Let's not make this a failure mode. If the user supplies an unsupported value, we can just return nothing. In the documentation we can always make it explicit which types are supported.

patricksanders:
Hey @bblommers, thanks for the review! While working on those changes I realized that `list_policies` and `list_policies_for_target` should be paginated, so I added that and applied your recommendations to `list_policies_for_target` as well. I like the explicit args much more than exploding `kwargs`, and making that change gave me a better grasp of how the moto bits work together (`models` <> `responses` in particular).

bblommers (approved):
LGTM! Nice to see so many tests here. 

Thanks for implementing these features in Moto @patricksanders

PR #9635, opened 2026-01-08
Implement IoT list_thing_principals_v2

### LINKED ISSUE #9629: NotImplementedError: The list_thing_principals_v2 action has not been implemented
This method seems to not be implemented yet.

```python
import moto
import boto3
with moto.mock_aws():
    iot = boto3.client("iot")
    iot.list_thing_principals_v2(thingName="thing")
```
```
NotImplementedError: The list_thing_principals_v2 action has not been implemented
```

Note, `list_thing_principals` does work.

Eason09053360:
Closes #9629

## Summary
Implemented the missing `list_thing_principals_v2` method for the IoT backend.

## Changes
- **models.py**: Added `list_thing_principals_v2` by reusing the existing logic from `list_thing_principals`.
- **responses.py**: Added the response handler to format the output according to the V2 API specification (returning `thingPrincipalObjects`).
- **tests**: Added a reproduction test case to verify the implementation.


tests/test_iot/test_issue_repro.py
```diff
@@ -0,0 +1,21 @@
+import boto3
+
+from moto import mock_aws
+
+
+@mock_aws
+def test_list_thing_principals_v2():
```
bblommers: Can we move this test into `test_iot.py` instead? The `list_thing_principals()` method is tested there as well, so it makes sense to add this test next to it IMO

tests/test_iot/test_issue_repro.py
```diff
@@ -0,0 +1,21 @@
...
+    # 2. Create a Thing (required to list principals)
+    client.create_thing(thingName=thing_name)
+
+    # 3. Call the newly implemented V2 function
+    # (This would raise NotImplementedError before our fix)
+    response = client.list_thing_principals_v2(thingName=thing_name)
+
+    # 4. Verify the response contains the V2 specific key
+    assert "thingPrincipalObjects" in response
+    assert response["thingPrincipalObjects"] == []
```
bpandola: Doesn't look like this actually verifies that the thing you created is in the response...

tests/test_iot/test_issue_repro.py
Eason09053360: Thanks for the review! I've updated the test case in tests/test_iot/test_iot.py.

tests/test_iot/test_issue_repro.py
Eason09053360: Thanks for the review!
I confirmed that boto3 filters out the thingName from the response (since it's already known in the request), so I updated the test to verify the principal and ensure the response structure is a dictionary (validating the V2 format)."

bblommers (approved):
LGTM! Thank you for adding this feature to Moto @Eason09053360

PR #9644, opened 2026-01-13
SNS: Fix incomplete conversions of boolean values

viren-nadkarni:
This PR fixes some regressions that appeared after https://github.com/getmoto/moto/pull/9642.

Specifically:
- Occurence of `AttributeError: 'bool' object has no attribute 'lower'` when invoking `SetEndpointAttributes`
- Boolean values incorrectly returned as `True` or `False` (expected `true` and `false`) when invoking `ListEndpointsByPlatformApplication` and `ListPlatformApplications`

### DIFF
--- a/moto/sns/responses.py
+++ b/moto/sns/responses.py
@@ -171,8 +171,10 @@ class SNSResponse(BaseResponse):
     def set_topic_attributes(self) -> ActionResult:
         topic_arn = self._get_param("TopicArn")
         attribute_name = self._get_param("AttributeName")
-        attribute_name = camelcase_to_underscores(attribute_name)
         attribute_value = self._get_param("AttributeValue")
+        if attribute_name == "ContentBasedDeduplication":
+            attribute_value = ensure_boolean(attribute_value)
+        attribute_name = camelcase_to_underscores(attribute_name)
         self.backend.set_topic_attribute(topic_arn, attribute_name, attribute_value)
         return EmptyResult()
 
@@ -359,6 +361,8 @@ class SNSResponse(BaseResponse):
         arn = self._get_param("SubscriptionArn")
         attr_name = self._get_param("AttributeName")
         attr_value = self._get_param("AttributeValue")
+        if attr_name == "RawMessageDelivery":
+            attr_value = ensure_boolean(attr_value)
         self.backend.set_subscription_attributes(arn, attr_name, attr_value)
         return EmptyResult()
 


moto/sns/models.py
```diff
@@ -869,7 +870,7 @@ def set_endpoint_attributes(
     ) -> PlatformEndpoint:
         endpoint = self.get_endpoint(arn)
         if "Enabled" in attributes:
-            attributes["Enabled"] = attributes["Enabled"].lower()
+            attributes["Enabled"] = ensure_boolean(attributes["Enabled"])
```
bpandola: We should be able to delete this conditional entirely, no?  This attribute should already have been coerced to `bool` in `responses.py`.

bpandola (changes requested):
@viren-nadkarni Thank you for closing the loop on this.  I have one minor request and then happy to merge.

P.S. I'd love to be able to run Localstack's test suite against Moto changes to catch regressions like this, but I'm guessing the bulk of your tests are not open source, right?

bblommers:
> P.S. I'd love to be able to run Localstack's test suite against Moto changes to catch regressions like this, but I'm guessing the bulk of your tests are not open source, right?

@bpandola The vast majority of SNS tests is actually open source*: https://github.com/localstack/localstack/tree/main/tests/aws/services/sns 

Don't know how easy it is to run this against Moto though :eyes: 

*For now at least - [this will change in March 2026](https://blog.localstack.cloud/the-road-ahead-for-localstack/#what-is-changing)

viren-nadkarni:
@bpandola Thanks for pushing this through :raised_hands:

bpandola:
> @bpandola The vast majority of SNS tests is actually open source*: https://github.com/localstack/localstack/tree/main/tests/aws/services/sns
> 
> Don't know how easy it is to run this against Moto though 👀
> 
> *For now at least - [this will change in March 2026](https://blog.localstack.cloud/the-road-ahead-for-localstack/#what-is-changing)

@bblommers Yeah, probably just wishful thinking on my part...  It seemed unfortunate to me that they have all this extra test coverage that could (and did in this case) catch a Moto regression, but the feedback loop is manual (and of course leaves them at the mercy of whatever changes we make possibly breaking them).  I felt like there should be some way to automate that check for the benefit of both communities, but that appears to be moot given the link you mentioned and the note I recently saw that they decided to hard fork Moto.

PR #9669, opened 2026-01-26
Adding Amazon Connect support

ghorondo:
This pull request adds support for Amazon Connect instance management and analytics data associations.

Implemented Methods:

- create_instance: Create a new Connect instance.
- describe_instance: Retrieve details for a Connect instance.
- list_instances: List Connect instances
- delete_instance: Delete a Connect instance and clean up analytics associations.
- associate_analytics_data_set: Associate an analytics data set with a Connect instance. Enforces one association per (instance_id, data_set_id) pair.
- disassociate_analytics_data_set: Remove an analytics data association from a Connect instance.
- list_analytics_data_associations: List analytics data associations for a Connect instance, with pagination and optional filtering by data_set_id.
- tag_resource: Add tags to a Connect resource.
- untag_resource: Remove tags from a Connect resource.
- list_tags_for_resource: List tags for a Connect resource.

Unit tests and error handling are included for all new methods.

### DIFF
--- a/.github/workflows/tests_sdk_ruby.yml
+++ b/.github/workflows/tests_sdk_ruby.yml
@@ -13,7 +13,7 @@ jobs:
     steps:
       - uses: actions/checkout@v6
       - name: Set up Ruby ${{ matrix.ruby-version }}
-        uses: ruby/setup-ruby@80740b3b13bf9857e28854481ca95a84e78a2bdf
+        uses: ruby/setup-ruby@90be1154f987f4dc0fe0dd0feedac9e473aa4ba8
         with:
           ruby-version: ${{ matrix.ruby-version }}
       - name: Set up Python 3.12
--- a/moto/core/serialize.py
+++ b/moto/core/serialize.py
@@ -248,7 +248,9 @@ class ResponseSerializer:
         response_dict: ResponseDict = {
             "body": "",
             "headers": {},
-            "status_code": self.DEFAULT_RESPONSE_CODE,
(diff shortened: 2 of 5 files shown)


moto/connect/responses.py
```diff
@@ -0,0 +1,178 @@
...
+    def connect_backend(self) -> ConnectBackend:
+        """Return backend instance specific for this region."""
+        return connect_backends[self.current_account][self.region]
+
+    def _get_instance_id(self) -> str:
+        """Extract instance_id from request path params."""
+        instance_id = self._get_param("InstanceId")
+        return unquote(instance_id) if instance_id else ""
+
+    def _get_int_param_with_fallback(self, name: str) -> Optional[int]:
```
bblommers: **question:** What's the reasoning behind this method? Incoming parameters should always have the same casing, so AFAICT we should be able to just use `self._get_int_param(..)` instead

moto/connect/responses.py
```diff
@@ -0,0 +1,178 @@
...
+            # Try alternate case (maxResults -> MaxResults or vice versa)
+            alt_name = (
+                name[0].upper() + name[1:]
+                if name[0].islower()
+                else name[0].lower() + name[1:]
+            )
+            value = self._get_param(alt_name)
+        return int(value) if value is not None else None
+
+    def _get_str_param_with_fallback(self, name: str) -> Optional[str]:
```
bblommers: Same here - why not use `self._get_param(..)`?

PR #9670, opened 2026-01-26
Speed up DynamoDB query deepcopy while preserving isolation

Will282:
We observed a ~50% speedup in our CI test suite when removing the DynamoDB query deepcopy entirely, however removing it isn't viable since returned items must remain detached from table state. This change optimizes the `__deepcopy__` implementation to preserve that isolation while significantly reducing overhead.

**Changes:**
- Implement custom `__deepcopy__` for `DynamoType` and `Item` that avoids the default pickle-based protocol, bypasses `LimitedSizeDict` size validation during copy, and skips `__init__` re-processing of nested values
- Add test verifying that mutating query results does not affect stored items
- Add parameterized test covering all 10 DynamoDB types (S, N, B, BOOL, NULL, SS, NS, BS, L, M)                                                                                                                   

**Local benchmarks** (not in CI):

| Scenario | Before | After |
|----------|--------|-------|
| 2000 items × 300 iter × 128B | 57.2ms/query | 15.7ms/query |
| 5000 items × 200 iter × 512B | 56.4ms/query | 17.1ms/query |

### DIFF
--- a/moto/dynamodb/models/dynamo_type.py
+++ b/moto/dynamodb/models/dynamo_type.py
@@ -98,6 +98,9 @@ class DynamoType:
         return f"DynamoType: {self.to_json()}"
 
     def __deepcopy__(self, memo: dict[int, Any]) -> "DynamoType":
+        """Custom deepcopy that bypasses the expensive default pickle-based
+        __reduce_ex__ protocol. Safe for DynamoDB AttributeValue trees,
+        which are acyclic."""
         if id(self) in memo:  # pragma: no cover
             return memo[id(self)]  # Circular refs can't occur in DynamoDB data
         result = self.__class__.__new__(self.__class__)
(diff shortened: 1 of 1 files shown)


moto/dynamodb/models/dynamo_type.py
```diff
@@ -97,6 +97,22 @@ def __ge__(self, other: "DynamoType") -> bool:
     def __repr__(self) -> str:
         return f"DynamoType: {self.to_json()}"
 
+    def __deepcopy__(self, memo: dict[int, Any]) -> "DynamoType":
```
bblommers, line 100: Would you mind adding a docstring here to explain why the rationale? The PR description is clear enough - but it would be useful to also have this explanation directly in the code

Will282: Done 👍  I've added docstrings to both `__deepcopy__` overrides explaining the rationale.

PR #9671, opened 2026-01-26
[ResourceGroupsTaggingAPI] Add CloudFront to untag_resources and tag_resources 

ApogeeAC:
- Added CloudFront to ResourceGroupsTaggingApi `tag_resources` and `untag_resources` functions.
- Added supporting tests


ApogeeAC:
@bblommers I could move these to the cloudfront tests directory to consolidate if that is a direction Moto would like to take. I know some resources have begun migrating the tests to `test_***_integration` in their respective directories.

bblommers:
> I could move these to the cloudfront tests directory to consolidate if that is a direction Moto would like to take. I know some resources have begun migrating the tests to test_***_integration in their respective directories.

Yeah, if you could @ApogeeAC , that would be great. The `test_resourcegroupstaggingapi.py` is already getting a bit too unwieldy, so having the tests in a dedicated file would be much more readable IMHO

tests/test_cloudfront/test_cloudfront_dist_tags.py
```diff
@@ -76,3 +76,75 @@ def test_untag_resource_on_existing_distribution():
...
+    rtapi.untag_resources(
+        ResourceARNList=[dist_arn],
+        TagKeys=["k1", "k2"],
+    )
+
+    tags = client.list_tags_for_resource(Resource=dist_arn)["Tags"]["Items"]
+
+    assert {"Key": "k1", "Value": "v1"} not in tags
+    assert {"Key": "k2", "Value": "v2"} not in tags
+    assert {"Key": "NewKey", "Value": "NewValue"} in tags
```
bpandola: The rtapi tests need to go into a `test_cloudfront_integration.py` file because the Resource Groups Tagging backend requires extra dependencies not required by CloudFront.

bpandola (changes requested):

bpandola (approved):
Thanks @ApogeeAC!

Issue #9674, opened 2026-01-27
SageMaker describe_domain does not return DomainSetting attribute


soucema9:
SageMaker describe_domain does not return `DomainSettings` attribute.

It is caused by typo `DomainSetting` instead of `DomainSettings` in https://github.com/getmoto/moto/blob/e59d4f4753a28ed12f07353566a7a43a2169a448/moto/sagemaker/models.py#L2200C14-L2200C27

bpandola:
Closed in #9675

PR #9675, opened 2026-01-27
Fix SageMaker describe_domain does not return DomainSetting attribute

soucema9:
SageMaker describe_domain does not return DomainSettings attribute.

It is caused by typo DomainSetting instead of DomainSettings in https://github.com/getmoto/moto/blob/e59d4f4753a28ed12f07353566a7a43a2169a448/moto/sagemaker/models.py#L2200C14-L2200C27


soucema9:
Fixes #9674

PR #9677, opened 2026-01-29
[Backup] Adding backup vault lock configuration support

ghorondo:
This PR adds the put_backup_vault_lock_configuration and delete_backup_vault_lock_configuration methods for AWS Backup.

### DIFF
--- a/tests/test_backup/test_backup.py
+++ b/tests/test_backup/test_backup.py
@@ -444,6 +444,7 @@ def test_delete_backup_vault_lock_configuration():
 
 @pytest.mark.skipif(
     settings.TEST_SERVER_MODE, reason="Can't freeze time in server mode"
+)
 @freeze_time("2024-01-01")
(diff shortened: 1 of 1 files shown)


tests/test_backup/test_backup.py
```diff
@@ -358,3 +359,110 @@ def test_untag_resource():
...
+    client.create_backup_vault(BackupVaultName="test-vault")
+
+    client.put_backup_vault_lock_configuration(
+        BackupVaultName="test-vault",
+        MinRetentionDays=7,
+        ChangeableForDays=3,
+    )
+
+    # Fast forward past the lock date
+    with freeze_time("2024-01-05"):
```
bblommers, line 461: Freezing the time does not work in ServerMode, as that runs in a separate process (and therefore doesn't know about the freeze). If you disable this test in ServerMode, the tests should pass

PR #9680, opened 2026-01-29
ELBv2: Support internal load balancers with dualstack configuration

simonrw:
# Motivation

Currently `moto` does not allow the creation of ELBv2 "internal" type _dualstack_ load balancers. This is not how AWS behaves - internal dualstack load balancers can be created successfully.

# Changes

- Remove the logical check for internal dualstack load balancers in `set_ip_address_type` (from 2017!)
- Add `ip_address_type` property to `FakeLoadBalancer` so it can be updated from `set_ip_address_type`
- Add validation during `create_load_balancer` that the ip address type is valid
- Update `create_load_balancer` test helper to accept the ip address type
- Update test validating behaviour of `set_ip_address_type` to conform to new implementation
- Add new test verifying that `create_load_balancer` supports creating internal dualstack load balancers
- Add validation that the ip address type is one of the supported values

### DIFF
--- a/tests/test_elbv2/test_elbv2.py
+++ b/tests/test_elbv2/test_elbv2.py
@@ -1318,6 +1318,17 @@ def test_set_ip_address_type():
     client.set_ip_address_type(LoadBalancerArn=arn, IpAddressType="dualstack")
     assert get_ip_address_type(arn) == "dualstack"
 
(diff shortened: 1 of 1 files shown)


bblommers (approved):
LGTM - thanks for the PR @simonrw!

PR #9681, opened 2026-01-29
Refactor S3 request handling to avoid using the `method_dispatch` helper

bpandola:
An effort was made a couple years ago (see #7626 and related PRs) to refactor backends' urls.py/responses.py to avoid using the decorator introduced in #6746.  The issue that decorator fixed (#6745) was legitimate, but better handled via the refactoring.  The S3 backend was the only backend still making use of the `method_dispatch` helper.  This PR refactors the S3 backend to avoid using this method and removes the resulting dead code.

### DIFF
--- a/moto/core/responses.py
+++ b/moto/core/responses.py
@@ -274,32 +274,6 @@ class BaseResponse(_TemplateEnvironmentMixin, ActionAuthenticatorMixin):
     def dispatch(cls, *args: Any, **kwargs: Any) -> Any:  # type: ignore[misc]
         return cls()._dispatch(*args, **kwargs)
 
-    @classmethod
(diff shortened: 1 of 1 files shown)


bpandola:
@bblommers You did the initial refactoring to avoid using `method_dispatch`, so I'm requesting your review here just to make sure there wasn't some reason why you left out S3.  Thanks.

(This only came to my attention because I was doing a little research/prep work for [eventually] ridding the S3 backend of its XML templates.)

bblommers (approved):
Hey @bpandola! IIRC I didn't tackle S3 because I thought it was going to be too complicated. Happy to see that I was wrong, and that we can get rid of this! :partying_face:

PR #9682, opened 2026-01-30
[SQS] [Fix] Implements full support for 1MB SQS message size

henryivesjones:
There is a bug in the current SQS `send_message` implementation that doesn't allow messages > 256KB to be sent. [That limitation has been raised to 1MiB in AWS.](https://aws.amazon.com/about-aws/whats-new/2025/08/amazon-sqs-max-payload-size-1mib/) This fix correctly sets a given queues max message size to the value set when creating the queue.

At present there are two separate places that the maximum size for an sqs message is defined. One of them contains the old value of 256KB. This value is used in a pre-check within the `send_message` method.

```python
        if len(message) > MAXIMUM_MESSAGE_LENGTH:
            raise SQSException(
                "InvalidParameterValue",
                "One or more parameters are invalid. Reason: Message must be shorter than 262144 bytes.",
            )
```

My change consolidate that size max to a single definition and removes this pre-check. There is another check on the `sqs_backend` that enforces the actual size limit for the given queue. The queue model has protections on the attribute being set as well as a default value so I believe that this extra check is unnecessary.

I added a test case that would have caught this bug as well as fixed a test that referenced the `256KB` limit.

### DIFF
--- a/tests/test_sqs/test_sqs.py
+++ b/tests/test_sqs/test_sqs.py
@@ -1447,7 +1447,7 @@ def test_send_large_message_fails():
     err = ex.value.response["Error"]
     assert err["Code"] == "InvalidParameterValue"
     assert err["Message"] == (
-        "One or more parameters are invalid. Reason: Message must be shorter than 1048576 bytes."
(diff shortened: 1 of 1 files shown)


bpandola (approved):
Thanks @henryivesjones

PR #9684, opened 2026-01-30
[ELBv2] Reorg SSL_POLICES to its own file, update calls

ApogeeAC:
As the amount of SSL policies AWS supports continues to grow, the `SSL_POLICIES` var grows with it. Moved it to its own file to cleanup the `responses.py` file.

Also added the latest policies to support any PQ policy calls as currently if a user passes one in Moto throws an error because its not in the supported SSL_POLICIES structure.

Created a `build_policy` helper to build the correct [response](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/elbv2/client/describe_ssl_policies.html) structure with ciphers and priorities.


ApogeeAC:
Need to update `describe_ssl_policies` to include the ciphers and priorities

bblommers (approved):
Makes sense @ApogeeAC - thank you!

PR #9710, opened 2026-02-05
[DynamoDB] Add multi-attribute composite key support for GSIs

### LINKED ISSUE #9545: FEATURE REQUEST: [DynamoDB]add support Multi-attribute key schema for Global Secondary Indexes(GSI)
Recently, DynamoDB supports Multi-attribute key schema in GSI.
I seem this feature is very useful in many cases.

https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/GSI.DesignPattern.MultiAttributeKeys.html

cluebbehusen:
Closes #9545 

## Summary

Adds support for DynamoDB's [multi-attribute composite keys in Global Secondary Indexes](https://aws.amazon.com/about-aws/whats-new/2025/11/amazon-dynamodb-multi-attribute-composite-keys-global-secondary-indexes/), announced November 2025. This feature allows up to 4 attributes each for partition (HASH) and sort (RANGE) keys in GSIs.

- [AWS announcement](https://aws.amazon.com/about-aws/whats-new/2025/11/amazon-dynamodb-multi-attribute-composite-keys-global-secondary-indexes/)
- [AWS blog post](https://aws.amazon.com/blogs/database/multi-key-support-for-global-secondary-index-in-amazon-dynamodb/)
- [Multi-attribute keys pattern docs](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/GSI.DesignPattern.MultiAttributeKeys.html)

## What's implemented

- **Table creation** with multi-attribute GSI key schemas
- **KeyConditionExpression queries** with full validation:
  - Range keys must be specified left-to-right (cannot skip)
  - Only the last range key in the query can use range operators (`<`, `>`, `BETWEEN`, `begins_with`)
  - Earlier range keys must use equality
- **Legacy KeyConditions queries** with the same validation (reuses `validate_schema`)
- **Composite key sorting** across multiple range keys
- **Sparse GSI behavior** — items missing any key attribute are excluded from the index
- **Pagination** with `LastEvaluatedKey` containing all key attributes

## Tests

16 new tests in `tests/test_dynamodb/test_dynamodb_multi_attribute_gsi.py`, all marked `@pytest.mark.aws_verified` and verified against real AWS DynamoDB.

### DIFF
--- a/moto/dynamodb/models/table.py
+++ b/moto/dynamodb/models/table.py
@@ -788,9 +788,7 @@ class Table(CloudFormationModel):
                 )
 
             # Get ALL hash keys from schema (multi-attribute support)
-            index_hash_keys = [
-                key for key in index.schema if key["KeyType"] == "HASH"
-            ]
+            index_hash_keys = [key for key in index.schema if key["KeyType"] == "HASH"]
             if not index_hash_keys:
                 raise MockValidationException(
                     f"Missing Hash Key. KeySchema: {index.name}"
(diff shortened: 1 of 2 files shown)


cluebbehusen:
For test coverage:

1. `get_key()` in `key_condition_expression.py` is now untested since it's effectively dead code (has been replaced by `get_keys()`).
2. The `elif hash_key` fallback in `table.py` is similar to above. It's a fallback so that `Table.query()` still works if called with the old `hash_key` parameter. Both code paths in `responses.py` now always populate `hash_key_conditions`, so this fallback is never hit in practice.

So, in summary, both are dead code *internally*. But if any external consumer of the package is calling internal functions, things could break for them. I erred on the side of caution and preserved them, but, if the policy of this package is to not maintain support for internal functions between versions, I'd be happy to remove them.

The remaining lines uncovered are pre-existing defensive guards that can't be triggered through normal API usage.

Also, apologies for closing and reopening this PR. That was just an inadvertent misclick.

bblommers (approved):
Very cool - thank you for contributing this feature to Moto @cluebbehusen!

This feature seems relatively simple, but there are always a lot of edge cases of course -and having the documentation as part of tests helped me a lot to understand how this is supposed to work. The negative tests are great as well - nothing as frustrating as having a test case pass against Moto, only for it to fail against AWS. 

> if the policy of this package is to not maintain support for internal functions between versions

That is the policy, yes. Don't worry about in this case, though - I'll remove that code in the release PR (scheduled for later today).

PR #9711, opened 2026-02-06
SQS: Fix trailing slash bug

coelias:
New versions of AWS SDK add a trailing `/` at the end of the queue URL making moto to return a 404 as there is no URL match.


bpandola:
Closing due to a lack of response to our request for additional information. It can be reopened when/if more details are provided.

PR #9723, opened 2026-02-08
Prep release 5.1.21

bblommers:
- Updated documentation showing all new features
 - Improve JS tests to show installed versions
 - Removes some dead code from DynamoDB (see https://github.com/getmoto/moto/pull/9710#issuecomment-3860226791)
 - Upgrades DotNet tests to 9.0
 - Automatically use the latest DotNet dependencies for AWS (just in case that could explain #9711)
 - Add a very simple JS test for SQS (just in case that could explain #9711)
 - Randomizes the RDS tests for JS (makes it a little easier to run them continously locally during development)
 - Removes the `wait_for_instance_termination` in Autoscaling tests. This used to be necessary when we used a dedicated subnet, and we needed to ensure that instances were terminated before trying to delete the subnet. With https://github.com/getmoto/moto/pull/9651 we no longer use a dedicated subnet though, so there's also no point in waiting for the instances to be terminated. This should speed up these tests significantly when running them against AWS.

### DIFF
--- a/other_langs/tests_dotnet/s3/UnitTest.cs
+++ b/other_langs/tests_dotnet/s3/UnitTest.cs
@@ -20,7 +20,7 @@ public class UnitTest1
         AmazonS3Client s3Client = new AmazonS3Client(config);
 
         var initial_list = await s3Client.ListBucketsAsync();
-        initial_list.Buckets.Count.Should().Be(0);
+        initial_list.Buckets.Should().BeNull();
 
         var putBucketRequest = new PutBucketRequest
         {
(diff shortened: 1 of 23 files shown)


other_langs/tests_dotnet/s3/UnitTest.cs
```diff
@@ -20,7 +20,7 @@ public async Task TestListBuckets()
         AmazonS3Client s3Client = new AmazonS3Client(config);
 
         var initial_list = await s3Client.ListBucketsAsync();
-        initial_list.Buckets.Count.Should().Be(0);
+        initial_list.Buckets.Should().BeNull();
```
bblommers, line 23: Not a DotNet person, but I'm assuming this is because we now always use the latest version of the `AWSSDK.S3` dependency, which is 4.x (instead of the 3.x we used to use)

PR #9739, opened 2026-02-12
RDS: DBClusterParameterGroup Improvements

MichaelFoleyFZ:
Initial error reported when run rds.list_tags_tags_for_resource 

```
E           botocore.exceptions.ClientError: An error occurred (InvalidParameterValue) when calling the ListTagsForResource operation: Invalid resource name: arn:aws:rds:us-east-1:123456789012:cluster-pg:test-data-db-14
```

- Adds support for getting the tags assigned to an rds cluster param group
- Aligns the creation commands for cluster param group with the param group to make it more consistent 
- Add copy_db_cluster_parameter_group to align with what parameter groups have

### DIFF
--- a/moto/rds/models.py
+++ b/moto/rds/models.py
@@ -3898,10 +3898,10 @@ class RDSBackend(BaseBackend):
                 "The parameter DBParameterGroupFamily must be provided and must not be blank."
             )
         db_cluster_parameter_group = DBClusterParameterGroup(
-            self,
+            backend=self,
             db_cluster_parameter_group_name=db_cluster_parameter_group_name,
-            description=description,
             db_parameter_group_family=db_parameter_group_family,
+            description=description,
             tags=tags,
         )
         self.db_cluster_parameter_groups[db_cluster_parameter_group_name] = (


moto/rds/exceptions.py
```diff
@@ -57,6 +57,22 @@ def __init__(self, db_parameter_group_name: str):
...
+
+class DBParameterGroupAlreadyExistsFault(RDSClientError):
+    def __init__(self, db_parameter_group_name: str):
+        super().__init__(
+            "DBParameterGroupAlreadyExistsFault",
+            f"DB Parameter Group {db_parameter_group_name} already exists.",
+        )
+
+
 class DBParameterGroupAlreadyExistsError(RDSClientError):
```
bpandola: Why are you duplicating these?

moto/rds/exceptions.py
```diff
@@ -57,6 +57,22 @@ def __init__(self, db_parameter_group_name: str):
...
+
+
+class DBParameterGroupAlreadyExistsFault(RDSClientError):
+    def __init__(self, db_parameter_group_name: str):
+        super().__init__(
+            "DBParameterGroupAlreadyExistsFault",
+            f"DB Parameter Group {db_parameter_group_name} already exists.",
+        )
+
+
```
bpandola: What's the reason for duplicating these exceptions?  The Error/Fault naming thing is a holdover from long ago, but the error codes for the existing exception classes should be correct.

moto/rds/models.py
```diff
@@ -3851,19 +3903,30 @@ def describe_orderable_db_instance_options(
         return []
 
     def create_db_cluster_parameter_group(
-        self,
-        group_name: str,
-        family: str,
-        description: str,
+        self, db_cluster_parameter_group_kwargs: dict[str, Any]
```
bpandola: We actually want to move in the other direction here... The RDS backend uses `kwargs` a lot, but we want to move toward stronger typed lists of arguments.  If you want to better align the parameter names with their AWS API counterparts, that's fine; but we don't want to hide arguments in an untyped (Any) dict.

moto/rds/models.py
```diff
@@ -3874,8 +3937,13 @@ def describe_db_cluster_parameter_groups(
...
 
-    def delete_db_cluster_parameter_group(self, group_name: str) -> None:
-        self.db_cluster_parameter_groups.pop(group_name)
+    def delete_db_cluster_parameter_group(
+        self, db_cluster_parameter_group_name: str
+    ) -> DBClusterParameterGroup:
+        if db_cluster_parameter_group_name in self.db_cluster_parameter_groups:
+            return self.db_cluster_parameter_groups.pop(db_cluster_parameter_group_name)
+        else:
+            raise DBParameterGroupNotFoundFault(db_cluster_parameter_group_name)
```
bpandola, line 3927: I get that you're probably just following the existing convention in the `delete_db_parameter_group` method, but let's refactor this into a guard clause that raises an exception (with no `else` statement).

moto/rds/models.py
```diff
@@ -3874,8 +3937,13 @@ def describe_db_cluster_parameter_groups(
             return [self.db_cluster_parameter_groups[group_name]]
         return list(self.db_cluster_parameter_groups.values())
 
-    def delete_db_cluster_parameter_group(self, group_name: str) -> None:
-        self.db_cluster_parameter_groups.pop(group_name)
+    def delete_db_cluster_parameter_group(
+        self, db_cluster_parameter_group_name: str
+    ) -> DBClusterParameterGroup:
```
bpandola: I don't know that there's any point in returning the DBClusterParameterGroup if no downstream call sites actually make use of the value.

moto/rds/models.py
```diff
@@ -4477,6 +4553,10 @@ def name(self, value: str) -> None:
     def resource_id(self) -> str:
         return self.name
 
+    def delete(self, account_id: str, region_name: str) -> None:
+        backend = rds_backends[account_id][region_name]
+        backend.delete_db_cluster_parameter_group(self.name)
+
     def update_cluster_parameters(
```
bpandola, line 4536: If you're gonna add this, it needs to be covered by a test.  If it's not needed for your PR, remove it.

moto/rds/responses.py
```diff
@@ -535,13 +535,9 @@ def remove_from_global_cluster(self) -> ActionResult:
         return ActionResult(result)
 
     def create_db_cluster_parameter_group(self) -> ActionResult:
-        group_name = self.params.get("DBClusterParameterGroupName")
-        family = self.params.get("DBParameterGroupFamily")
-        desc = self.params.get("Description")
+        kwargs = self.params
```
bpandola: Again, even though there is a lot of this in the existing RDS backend, we want to move in the opposite direction. Explicitly defined parameters are better than `**kwargs` (outside of highly specific use cases).

bpandola (changes requested):
@MichaelFoleyFZ I appreciate the spirit of trying to align related methods or refactor for consistency, but we really want to keep changesets to the bare minimum required for the task at hand: in this case, DBClusterParameterGroup tagging and the CopyDBClusterParameterGroup method.  
It's just too much work to review when there are unrelated/unnecessary changes in the diff.  All that said, I am happy to merge if you can pare things back a bit based on my comments. Thanks!

MichaelFoleyFZ:
> @MichaelFoleyFZ I appreciate the spirit of trying to align related methods or refactor for consistency, but we really want to keep changesets to the bare minimum required for the task at hand: in this case, DBClusterParameterGroup tagging and the CopyDBClusterParameterGroup method. It's just too much work to review when there are unrelated/unnecessary changes in the diff. All that said, I am happy to merge if you can pare things back a bit based on my comments. Thanks!

No worries and totally valid, was just hitting each step as I was testing my own work that hit the issues, but should have broken them up. Will fix it up.

Issue #9724, opened 2026-02-08
Cloudformation: support `AWS::KMS::Alias` and `AWS::SSM::Document` resources


ViachaslauKabak:
Could you please add new feature support to moto server. installing CF stack using suggestions from ticket https://github.com/getmoto/moto/issues/9701 i have following error:

```
No Moto CloudFormation support for AWS::KMS::Alias
C:\Python\Python3\Lib\site-packages\moto\cloudformation\parsing.py:343: UserWarning: Tried to parse AWS::KMS::Alias but it's not supported by moto's CloudFormation implementation
  parse_resource(resource_json, resources_map)
```

```
No Moto CloudFormation support for AWS::SSM::Document
C:\Python\Python3\Lib\site-packages\moto\cloudformation\parsing.py:343: UserWarning: Tried to parse AWS::SSM::Document but it's not supported by moto's CloudFormation implementation
  parse_resource(resource_json, resources_map)
```

PR #9744, opened 2026-02-13
KMS: Refactor Aliases into standalone models

bblommers:
There is an outstanding feature request (#9724) to support the `AWS::KMS::Alias` CloudFormation model, and this is really only possible if `Alias` is a fully fledged class.

That's why this PR changes the underlying backend to store aliases as in their own `Alias` class, as children of `Key`s, instead of in a separate `keys_to_aliases` dictionary.

### DIFF
--- a/moto/kms/responses.py
+++ b/moto/kms/responses.py
@@ -277,22 +277,21 @@ class KmsResponse(BaseResponse):
 
         response_aliases = []
 
-        backend_aliases = self.kms_backend.list_aliases()
-        for target_key_id, aliases in backend_aliases.items():
-            for alias_name in aliases:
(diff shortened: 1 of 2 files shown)


moto/kms/responses.py
```diff
@@ -302,11 +301,6 @@ def list_aliases(self) -> str:
                     }
                 )
 
-        if key_id is not None:
-            response_aliases = list(
-                filter(lambda alias: alias["TargetKeyId"] == key_id, response_aliases)
-            )
-
```
bblommers, line 309: Filtering now occurs before we even add the alias to the dictionary, so this is no longer necessary

PR #9725, opened 2026-02-09
S3: delete_object(): support IfMatch

layla19:
When conditionally deleting S3 objects, I ran into the same issue as #8653 but for `delete_object`.

This PR updates the `delete_object` operation using the same mechanism as for `put_object`.

PR #9747, opened 2026-02-14
CI: allow running a subset of the AWS verified tests

bpandola:
Currently, the AWS verified workflow runs once a week.  It can also be run manually as needed, but it only runs against the `master` branch and always runs the entire suite of tests.  When PRs are opened that contain tests marked as `aws_verified`, maintainers don't have an (easy) way to corroborate that claim before merging.  Once merged, it can take up to a week to find out if those tests actually pass against a real AWS backend--and sometimes they don't.

This PR adds input parameters to the `workflow_dispatch` event to allow running specific AWS verified test(s) from a particular git ref.  This change will allow maintainers to run AWS verified tests included in a Pull Request before merging.

Note: I have run this workflow against two existing PRs (#9725 and #9727) that include tests marked as `aws_verified` (see test runs [171](https://github.com/getmoto/moto/actions/runs/22015134144) and [172](https://github.com/getmoto/moto/actions/runs/22015177270)).

### DIFF
--- a/.github/workflows/tests_real_aws.yml
+++ b/.github/workflows/tests_real_aws.yml
@@ -5,9 +5,8 @@ on:
   workflow_dispatch:
     inputs:
       ref:
-        default: "master"
-        description: "The branch, tag or SHA to checkout, e.g. master or refs/pull/<pr_number>/merge"
-        required: false
+        description: "The branch, tag or SHA to checkout, e.g. refs/pull/<pr_number>/merge"
+        required: true
         type: string
       tests_specifier:
         default: "tests"


.github/workflows/tests_real_aws.yml
```diff
@@ -3,6 +3,16 @@ on:
   schedule:
     - cron: '00 6 * * 0'
   workflow_dispatch:
+    inputs:
+      ref:
```
bblommers, line 7: **nitpick:** Can we default this to master?

bpandola: I mean, sure, but the whole point is to run it against something other than `master`...

bblommers: Yeah, I suppose that's fair - it does make more sense to leave it empty :+1:

PR #9756, opened 2026-02-17
[Route53Resolver] Fix missing Id filter support for list_resolver_query_log_configs

ghorondo:
`list_resolver_query_log_configs` supports filtering by `Id` per the AWS API docs, but moto's `ResolverQueryLogConfig.FILTER_NAMES` was  missing it, causing the filter to be rejected as invalid.
  
  https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/route53resolver/client/list_resolver_query_log_configs.html


bpandola (changes requested):
Happy to merge if you can add a test case to cover the new filter.

PR #9759, opened 2026-02-20
fix(SQS): ReceiveMessageWaitTimeSeconds & KmsDataKeyReusePeriodSeconds wrong attributes name

robsonpiere:
This pull request improves the handling and testing of SQS queue attribute type conversions, specifically for integer-valued attributes. It corrects attribute naming inconsistencies and adds comprehensive tests to ensure that string inputs for integer fields are properly converted and stored as integers internally.

**Bug fixes and attribute handling:**

* Corrected attribute names in the `integer_fields` tuple in `moto/sqs/models.py` to match official SQS attribute names, fixing typos and ensuring proper attribute mapping.
  * Corrected the attribute from `ReceiveMessageWaitTime` to `ReceiveMessageWaitTimeSeconds`
  * Corrected the attribute from `KmsDataKeyreusePeriodSeconds ` to `KmsDataKeyReusePeriodSeconds `

**Testing improvements:**

* Added a new parameterized test `test_set_attributes_integer_fields_conversion` in `tests/test_sqs/test_sqs.py` to verify that all integer SQS queue attributes are correctly converted from strings to integers internally and that updates to these attributes are handled properly.
* Imported `camelcase_to_underscores` and `sqs_backends` in the test file to facilitate internal attribute access and verification for the new tests.

When we create queues and subscriptions using Terraform, the ReceiveMessageWaitTimeSeconds in the Moto API was a string rather than an integer.

When you try to send a message, if you have a lambda subscribed to the queue, you'll see this error:

```
  File "/moto/moto/core/responses.py", line 581, in call_action

    response = method()

  File "/moto/moto/sqs/responses.py", line 155, in send_message

    message = self.sqs_backend.send_message(

    

  File "/moto/moto/sqs/models.py", line 878, in send_message

    queue.add_message(message)

    ~~~~~~~~~~~~~~~~~^^^^^^^^^

  File "/moto/moto/sqs/models.py", line 579, in add_message

    messages = backend.receive_message(

    

  File "/moto/moto/sqs/models.py", line 968, in receive_message

    polling_end = unix_time() + wait_seconds_timeout
``` 

It is probably caused because the property was not converted to an integer  in the _set_attributes method due to the wrong name

### DIFF
--- a/tests/test_sqs/test_sqs_integration.py
+++ b/tests/test_sqs/test_sqs_integration.py
@@ -80,7 +80,9 @@ def test_invoke_fake_function_from_sqs_queue():
     sqs = boto3.resource("sqs", region_name="us-east-1")
     queue_name = str(uuid.uuid4())[0:6]
     queue = sqs.create_queue(QueueName=queue_name)
-
+    queue.set_attributes(
+        Attributes={"ReceiveMessageWaitTimeSeconds": "2"}  # Coverage for Moto #9759
+    )
     fn_name = str(uuid.uuid4())[0:6]
     conn = boto3.client("lambda", region_name="us-east-1")
     func = conn.create_function(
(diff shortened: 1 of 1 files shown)


robsonpiere:
> @robsonpiere Looks like this bug was hiding for 8 years waiting to be found.... Good catch! Let's get a test case to cover the scenario and then I'm happy to merge.

Thanks for your feedback. While I was adding the test case, I found another inconsistency in the KmsDataKeyreusePeriodSeconds key.  I fixed it also and added a parametrized test. Let me know if the tests make sense for you.

robsonpiere:
@bpandola, I updated the testes to be skipped in the server mode. Let me know if it works for you

PR #9772, opened 2026-02-22
Migrate 11 AWS services to ActionResult serialization and utcnow() timestamps

### LINKED ISSUE #9763: JSON-protocol services return timestamps as ISO 8601 strings instead of epoch seconds
### Description

Multiple AWS services that use the JSON wire protocol return timestamp fields as ISO 8601 strings (e.g., `"2025-02-21T16:45:27.178871"`) instead of Unix epoch seconds (e.g., `1740153927.178871`).

According to the [AWS JSON protocol specification](https://smithy.io/2.0/aws/protocols/aws-json-1_1-protocol.html), when a shape is of type `timestamp` and no explicit `timestampFormat` trait is set, the default serialization for `json` and `rest-json` protocols is **`unixTimestamp`** (a number representing seconds since the Unix epoch).

This causes AWS SDK clients to fail with errors like:

```
Input timestamp string must be no longer than 20 characters
```

### Affected Services

| # | Service | Protocol | Affected Resources | Affected Fields |
|---|---------|----------|--------------------|-----------------|
| 1 | **Kinesis** | json | Stream | 1 |
| 2 | **Firehose** | json | DeliveryStream | 2 |
| 3 | **Glue** | json | Job, Crawler, SecurityConfiguration, LastCrawlInfo, JobRun, Session, Connection, Workflow, WorkflowRun | 18 |
| 4 | **SageMaker** | json | 27 resource types (Pipeline, TrainingJob, Endpoint, Model, etc.) | 63+ |
| 5 | **DynamoDB Streams** | json | StreamRecord | 1 |
| 6 | **Transcribe** | json | TranscriptionJob, MedicalTranscriptionJob, Vocabulary | 7 |
| 7 | **Step Functions** | json | StateMachine, Execution, Activity | 4 |
| 8 | **SSM** | json | DocumentDescription | 1 |
| 9 | **Service Catalog** | json | PortfolioDetail | 1 |
| 10 | **Kinesis Video** | rest-json | StreamInfo | 1 |
| 11 | **MemoryDB** | json | Node, ShardDetail | 2 |
| 12 | **IoT** | rest-json | BillingGroupMetadata | 1 |

### Root Cause

Timestamp values are stored or serialized using `datetime.strftime()`, `datetime.isoformat()`, or `iso_8601_datetime_with_milliseconds()` instead of `unix_time()` from `moto.core.utils`.

```python
# Pattern 1: strftime
self.creation_time = datetime.now().strftime("%Y-%m-%d %H:%M:%S")

# Pattern 2: isoformat
self.creation_time = datetime.now(timezone.utc).isoformat()

# Pattern 3: iso_8601_datetime_with_milliseconds
self.creation_date = iso_8601_datetime_with_milliseconds()
```

All should be:

```python
from moto.core.utils import unix_time

self.creation_time = unix_time()
```

yusuke-koyoshi:
- fix #9763  

## Summary

  Migrate 11 AWS service implementations from manual `json.dumps()` serialization to the framework-standard `ActionResult`/`EmptyResult` pattern, and replace ad-hoc timestamp generation (`unix_time()`, `.isoformat()`,
  `.strftime()`) with `utcnow()` so that the serializer handles datetime-to-epoch conversion automatically.

  ### Problem

  Many services still use `json.dumps()` in `responses.py` to build HTTP response bodies manually instead of returning `ActionResult` objects. Timestamps are eagerly converted to strings/floats in `models.py`, bypassing the
  serializer's timestamp formatting logic. This causes:

  - Inconsistent timestamp formats across services
  - Timestamps that cannot be mocked with `freeze_time` in tests
  - Duplicated serialization logic that should be handled by the framework

  ### Changes

  **Services migrated (11 total, 32 files, +1429/-1470 lines):**

  | Service | Methods | Key Changes |
  |---------|---------|-------------|
  | Firehose | models only | `unix_time()` → `utcnow()` |
  | KinesisVideo | 6 methods | Full `ActionResult` migration |
  | ServiceCatalog | 12 methods | Full `ActionResult` migration |
  | MemoryDB | 14 methods | Full `ActionResult` migration |
  | Transcribe | 18 methods | Full `ActionResult` migration + type annotation fixes |
  | StepFunctions | 20+ methods | Full `ActionResult` migration |
  | Kinesis | 38 methods | Full `ActionResult` migration + `freeze_time` tests |
  | SSM | 36 methods | Full `ActionResult` migration + `ParameterNotFound` exception |
  | IoT | 80+ methods | Full `ActionResult` migration + `freeze_time` tests |
  | Glue | 76+ methods | Full `ActionResult` migration + `DateTimeEncoder` removal |
  | SageMaker | 100+ methods | Full `ActionResult` migration + `_epoch_to_datetime` helper |

  **Cross-cutting fixes:**

  - **`moto/sagemaker/models.py`**: Fixed copy-paste bug in `list_compilation_jobs` where `creation_time_after` was used instead of `last_modified_time_after`
  - **`moto/sagemaker/utils.py`**: Added `_epoch_to_datetime()` helper to convert epoch floats at the response boundary
  - **`moto/ssm/responses.py`**: Replaced manual `json.dumps` error responses with `ParameterNotFound` exception
  - **`moto/glue/responses.py`**: Fixed `delete_database` to return `EmptyResult()` instead of `ActionResult({})`

  **Type annotation fixes:**
  - `moto/transcribe/models.py`: `Optional[str]` → `Optional[datetime]` for timestamp fields
  - `moto/sagemaker/models.py`: `Optional[str]` → `Optional[datetime]` for `last_execution_time`, `creation_time`, `last_modified_time`
  - `moto/iot/models.py`: `Any` → `dict[str, Union[str, datetime]]` for `to_dict()` return types
  - `moto/glue/models.py`: `dict[str, Any]` → precise Union types for `as_dict()` return types

  ### Test Changes

  - **`tests/test_sagemaker/test_sagemaker_compilation_jobs.py`**: Added `test_list_compilation_jobs_time_filters` — regression test for the `LastModifiedTimeAfter/Before` filter bug (fails with the copy-paste bug)
  - **`tests/test_iot/test_iot_cloudformation.py`**: Strengthened with `freeze_time` + exact value assertions instead of `isinstance` checks
  - **`tests/test_kinesis/test_kinesis.py`**: Replaced `time.sleep` patterns with `freeze_time` for deterministic testing
  - **`tests/test_glue/test_datacatalog.py`**, **`test_workflows.py`**: Added `freeze_time` + timestamp value assertions
  - **`tests/test_sagemaker/test_sagemaker_automljobs.py`**, **`test_sagemaker_hyper_parameter_tuning_job.py`**: Added `freeze_time` for deterministic timestamp tests

### DIFF
--- a/moto/core/serialize.py
+++ b/moto/core/serialize.py
@@ -133,7 +133,9 @@ class TimestampSerializer:
     def _timestamp_unixtimestamp(value: datetime) -> Union[int, float]:
         base_timestamp = calendar.timegm(value.timetuple())
         if value.microsecond:
-            return base_timestamp + value.microsecond / 1_000_000.0
+            # Smithy spec: "Values that are more granular than millisecond
+            # precision SHOULD be truncated to fit millisecond precision."
+            return base_timestamp + (value.microsecond // 1000) / 1000.0
         return base_timestamp
 
     def _timestamp_rfc822(self, value: Union[datetime, float]) -> str:
--- a/tests/test_core/test_serialize.py
+++ b/tests/test_core/test_serialize.py
@@ -157,8 +157,8 @@ def test_datetime_with_microseconds() -> None:
     assert time_str[-1] == "Z"
 
 
-def test_unixtimestamp_preserves_microsecond_precision() -> None:
-    """Verify that unixtimestamp serialization preserves microsecond precision."""
+def test_unixtimestamp_truncates_to_millisecond_precision() -> None:
+    """Verify that unixtimestamp serialization truncates to millisecond precision."""
     model = {
         "metadata": {"protocol": "json", "apiVersion": "2014-01-01"},
         "documentation": "",
@@ -196,9 +196,9 @@ def test_unixtimestamp_preserves_microsecond_precision() -> None:
     body = json.loads(serialized["body"])
     timestamp = body["Timestamp"]
 
-    # Microsecond precision: 123456 µs = 0.123456 s
+    # Millisecond precision: 123456 µs truncated to 123 ms = 0.123 s
     fractional = timestamp - int(timestamp)
-    assert abs(fractional - 0.123456) < 1e-6
+    assert abs(fractional - 0.123) < 1e-6
 
 
 def test_pretty_print_with_short_elements_and_list() -> None:


moto/core/serialize.py
```diff
@@ -130,8 +130,11 @@ def _timestamp_iso8601_zeroed(self, value: datetime) -> str:
         return value.strftime(self.ISO8601_MICRO_ZEROED)
 
     @staticmethod
-    def _timestamp_unixtimestamp(value: datetime) -> float:
-        return int(calendar.timegm(value.timetuple()))
+    def _timestamp_unixtimestamp(value: datetime) -> Union[int, float]:
+        base_timestamp = calendar.timegm(value.timetuple())
+        if value.microsecond:
+            return base_timestamp + value.microsecond / 1_000_000.0
+        return base_timestamp
```
bpandola, line 139: Why not always return a float?  I see your comment `avoid ".0" suffix in serialized output` but it's not clear why... does AWS not serialize the ".0"?

moto/sagemaker/models.py
```diff
@@ -2069,13 +2057,17 @@ def __init__(
...
             "CompletionCriteria": self.auto_ml_problem_type_config[
                 self.auto_ml_problem_type_config_name + "JobConfig"
             ]["CompletionCriteria"],
-            "AutoMLProblemTypeResolvedAttributes": self.auto_ml_problem_type_resolved_attributes,
         }
+        if self.auto_ml_problem_type_resolved_attributes is not None:
+            resolved["AutoMLProblemTypeResolvedAttributes"] = (
+                self.auto_ml_problem_type_resolved_attributes
+            )
+        self.resolved_attributes = resolved
```
bpandola, line 2070: This doesn't seem datetime related... Is this a case where ActionResult serialization exposed that Moto had been serializing something it should not have (or serializing it incorrectly)?

yusuke-koyoshi: Yes, exactly. This is not datetime-related — it's a pre-existing issue that was exposed by the ActionResult migration.

Previously, `auto_ml_problem_type_resolved_attributes` defaulted to `{"SDK_UNKNOWN_MEMBER": {"name": "UnknownMemberName"}}`, which is a boto3 client-side fallback marker for unknown union members — not something a server should return. With manual `json.dumps()` serialization this went unnoticed, but ActionResult uses botocore's serializer which validates against the operation model, so the invalid union key caused issues.

The fix initializes the field to `None` and only includes `AutoMLProblemTypeResolvedAttributes` in the response when it's actually set (for `Tabular` or `TextGeneration` problem types). This matches the [AWS API spec](https://docs.aws.amazon.com/sagemaker/latest/APIReference/API_AutoMLProblemTypeResolvedAttributes.html) where the field is a union type and should be absent for problem types like `ImageClassification`, `TextClassification`, and `TimeSeriesForecasting`.

moto/core/serialize.py
yusuke-koyoshi: The int/float distinction matches the [Smithy protocol compliance tests](https://github.com/smithy-lang/smithy/tree/main/smithy-aws-protocol-tests). The response-side tests expect whole-second timestamps as integers without `.0`:

- [`restJson1/json-structs.smithy#L453`](https://github.com/smithy-lang/smithy/blob/main/smithy-aws-protocol-tests/model/restJson1/json-structs.smithy#L453): `"epochSeconds": 1398796238`
- [`awsJson1_1/kitchen-sink.smithy#L318`](https://github.com/smithy-lang/smithy/blob/main/smithy-aws-protocol-tests/model/awsJson1_1/kitchen-sink.smithy#L318): `"Timestamp": 946845296`

Python's `json.dumps` renders `int` as `1398796238` and `float` as `1398796238.0`, so returning an `int` when there are no fractional seconds avoids the `.0` suffix.

When fractional seconds are present, the [Smithy timestamp spec](https://smithy.io/2.0/spec/protocol-traits.html#timestamp-formats) defines epoch-seconds with "optional millisecond precision" (e.g. `1515531081.123`) and states:

> "Values that are more granular than millisecond precision SHOULD be truncated to fit millisecond precision."

The original implementation was preserving full microsecond precision, so I've updated it to truncate to milliseconds:

```python
# before
return base_timestamp + value.microsecond / 1_000_000.0
# after
return base_timestamp + (value.microsecond // 1000) / 1000.0
```

3593848

yusuke-koyoshi:
@bpandola 

> P.S. This is really competent work and I was impressed with how quickly you got this drafted and submitted. Very curious if this was AI-assisted and, if so, which model was used?

Yes, it was AI-assisted.
The models I primarily use are Opus 4.6 and Sonnet 4.6.
I break down tasks in Claude Code, then iterate through implementation and review cycles. 
The final code review is always done by myself.

PR #9777, opened 2026-02-22
AWSLambda: Use shogo82148 as the default image provider

### LINKED ISSUE #9978: Moto Docker - AWS Lambda image manifest not found
Hi there!

I'm trying to run a Lambda function using Python 3.14 and the Docker version of Moto, but it fails with the following error:

`error running docker: 404 Client Error for http://host.docker.internal:2375/v1.54/images/create?tag=python3.14&fromImage=lambci%2Flambda: Not Found ("manifest for lambci/lambda:python3.14 not found: manifest unknown: manifest unknown")`

This is the Docker compose file I'm using to spin up Moto:

```yaml
services:
  moto-server:
    image: motoserver/moto:5.1.22
    ports:
    - "4566:4566"
    environment:
    - "MOTO_PORT=4566"
    - "DOCKER_HOST=tcp://host.docker.internal:2375"
    - "MOTO_DOCKER_LAMBDA_IMAGE:public.ecr.aws/lambda/python"
    # also tried - "MOTO_DOCKER_LAMBDA_IMAGE:public.ecr.aws/lambda/python3.14"
```

For some reason Moto is ignoring this variable and is trying to pull from the archived lambci/lambda repository.

Is there a way to force Moto to use the image I want?

bblommers:
Until now we've been using `mlupin` as the default provider for Lambda images. However, these images haven't received an update in a long time. If users want to use a recent runtime (anything released in the last 3  years), they need to manually configure a different provider.

Most people are probably using `shogo82148` as the 'custom' provider, as they have been creating up-to-date images for every recent runtime.

[This issue has come up fairly often](https://github.com/getmoto/moto/issues?q=is%3Aissue%20state%3Aclosed%20mlupin), and my response has always been: yes, we should change the default, but it's technically a breaking change, so this should happen in a major version bump.

However: the more I think about it, the more I think that the functionality doesn't actually break. It _changes_, yes - more images are available, and they are more recent. But nothing should actually break, which means that this should be an acceptable change in an non-major release.

The benefit of supporting recent runtimes is also that we can start using the `AWS_ENDPOINT_URL` to redirect AWS requests from inside the Lambda back to the MotoServer. This wasn't possible with `mlupin` images, as they contained older versions of the AWS SDK that did not yet support this environment variable.

Our current solution (and recommendation) is for users to have an if-else clause in their AWSLambda. If under test: set the endpoint url to `motoserver:5000`, else keep it as is. Using the `AWS_ENDPOINT_URL` means that users can do away with these workarounds. We're using this workaround in our tests as well a few times, which I've removed as part of our PR to verify that everything works as it should.

@bpandola I'm thinking of making this part of an upcoming 5.2 release. Do you foresee any problems with that, things that may break that I'm not seeing?

### DIFF
--- a/moto/awslambda/models.py
+++ b/moto/awslambda/models.py
@@ -964,7 +964,6 @@ class LambdaFunction(CloudFormationModel, DockerModel):
             return s
 
     def _invoke_lambda(self, event: str | None = None) -> tuple[str, bool, str]:
-        import docker
         import docker.errors
 
         # Create the LogGroup if necessary, to write the result to
@@ -1000,6 +999,10 @@ class LambdaFunction(CloudFormationModel, DockerModel):
             if settings.is_test_proxy_mode():
                 env_vars["HTTPS_PROXY"] = env_vars["MOTO_HTTP_ENDPOINT"]
                 env_vars["AWS_CA_BUNDLE"] = "/var/task/ca.crt"
+            else:
+                # When the proxy is active, we should keep the original URL
+                # Else: redirect requests to our server
+                env_vars["AWS_ENDPOINT_URL"] = env_vars["MOTO_HTTP_ENDPOINT"]
 
             container = exit_code = None
             log_config = docker.types.LogConfig(type=docker.types.LogConfig.types.JSON)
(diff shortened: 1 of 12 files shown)


bpandola:
@bblommers Seems fine to me (although I'm not an expert in this area of the codebase).  If I'm understanding the proposed change correctly, I agree that this does not warrant a major version bump.

bpandola:
@bblommers Any chance we can get this merged and release 5.2 soon?  Let me know if I can be of assistance. (There are a *lot* of changes since the last release and I want to find out if I broke anything before I make any more. 😆 )

bblommers:
Hey @bpandola! Yeah, absolutely, I've been putting this off for way too long. I'll try to get a release out (early) next week

bblommers:
@bpandola Well, it took a while, as life (and Github Actions reliability...) got in the way, but 5.2.0 is now out. Let me know if you need anything!

Issue #9757, opened 2026-02-18
DynamoDB put_item() incorrect response with new item and ReturnValues='ALL_OLD'


jin-gizmo:
Hello,

I am doing (something like) the following with **moto** (5.1.21) / Python 3.13.12 to create a _new_ item in a DynamoDB table:

```python
table = boto3.resource('dynamodb').Table('abc')
response = table.put_item(Item={'a': 'A'}, ReturnValues='ALL_OLD')
```

The response looks something like this:
```python
{
    'Attributes': {},
    'ResponseMetadata': {
        'HTTPHeaders': { ...  },
        'HTTPStatusCode': 200,
        'RequestId': '...',
        'RetryAttempts': 0,
    },
}
```

In the case of a new entry, DynamoDB itself doesn't include an `Attributes` attribute in the response rather than an empty object.

The boto3 (1.42.51) [put_item](https://docs.aws.amazon.com/boto3/latest/reference/services/dynamodb/table/put_item.html) doco has this which, admittedly, is a bit ambiguous.

> **Attributes** *(dict)* –
> The attribute values as they appeared before the `PutItem` operation, but only if `ReturnValues` is specified as `ALL_OLD` in the request.

Thanks -- Murray

PR #9778, opened 2026-02-23
fix(dynamodb): add test for put_item with ReturnValues='ALL_OLD' on new items

thakoreh:
## Summary

Related to #9757

Adds a test case verifying that `put_item()` with `ReturnValues='ALL_OLD'` returns an empty Attributes map when creating a new item (not replacing an existing one).

## Behavior

Per AWS DynamoDB documentation:
- `ReturnValues='ALL_OLD'` returns the old item attributes if an item was replaced
- For new items (no existing item to replace), returns `{"Attributes": {}}`

## Test Added

`test_put_item_return_values_all_old_new_item` - verifies the empty Attributes response for new items.

### DIFF
--- a/tests/test_dynamodb/test_dynamodb.py
+++ b/tests/test_dynamodb/test_dynamodb.py
@@ -1271,6 +1271,72 @@ def test_put_item_nonexisting_range_key():
     assert ex.value.response["Error"]["Code"] == "ValidationException"
 
 
+@mock_aws
+def test_put_item_returns_no_attributes_for_new_item():
+    """When putting a new item with ReturnValues=ALL_OLD, the response should
+    not contain the Attributes key since there was no previous item."""
+    name = f"T{uuid4()}"
+    client = boto3.client("dynamodb", region_name="us-east-1")
+    client.create_table(
+        TableName=name,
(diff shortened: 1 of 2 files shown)


tests/test_dynamodb/test_dynamodb.py
```diff
@@ -2012,20 +2012,34 @@ def test_put_return_attributes():
...
     r = dynamodb.put_item(
         TableName=table_name,
         Item={"id": {"S": "foo"}, "col1": {"S": "val2"}},
         ReturnValues="ALL_OLD",
     )
     assert r["Attributes"] == {"id": {"S": "foo"}, "col1": {"S": "val1"}}
 
+    # Test 3: Put a NEW item with ReturnValues='ALL_OLD' - should return empty Attributes map
+    # This is the correct AWS behavior: when putting a new item that doesn't exist,
+    # ReturnValues='ALL_OLD' should return {"Attributes": {}} not omit the key
```
bpandola, line 2168: Doesn't this comment contradict #9757, which says "in the case of a new entry, DynamoDB itself doesn't include an Attributes attribute in the response"?  You are saying here that the key should *not* be omitted--were you able to verify that against a real AWS backend?

bpandola:
Closing in favor of #9863

PR #9780, opened 2026-02-24
ec2: Add dry run error handling for fleet creation

nik-localstack:
When `CreateFleet` operation is called with `DryRun=true`, the fleet object should not be created and stored.

# Changes

* Adds `error_on_dryrun` for `create_fleet`
* Adds `ec2_aws_verified` test for the behavior


bblommers (approved):
Nice one - thank you for (still) contributing @nik-localstack!

PR #9785, opened 2026-02-25
feat(cognito-idp): make TOTP MFA work

### LINKED ISSUE #9786: feat: Cognito MFA TOTP support
Currently the TOTP MFA support is "faked" by simply accepting any token passed in by the user.

I've created a PR #9785 to address this by:
* implementing TOTP (using `cryptography`) on the Cognito IdP side and
* ensuring the tests client side, (uses `pyotp`) uses the full OTP flow

However, this support is only partial improvement because the secret used for TOTP is the same for all users.
I'm leaving that to a separate PR as it will require a larger change (need to modify sessions to add the MFA secret)

bdellegrazie:
Implement TOTP MFA using the existing cryptography API. This allows clients to behave correctly and use proper TOTP validation compared with previously where the UserCode was not handled.

Warning: this implementation still uses the fixed constant "secret" and therefore should not be considered for anything other than testing.

A solution implementing "real" TOTP was considered but would need storing the secret temporarily in the session during the auth process but the session object is unfortunately a tuple at this time.

Closes #9786

### DIFF
--- a/.github/workflows/tests_servermode.yml
+++ b/.github/workflows/tests_servermode.yml
@@ -21,7 +21,7 @@ jobs:
       run: |
         pip install build
         python -m build
-        docker run --rm -t --name motoserver -e TEST_SERVER_MODE=true -e MOTO_EC2_LOAD_DEFAULT_AMIS=false -e AWS_SECRET_ACCESS_KEY=server_secret -e AWS_ACCESS_KEY_ID=server_key -v `pwd`:/moto -p 5000:5000 -v /var/run/docker.sock:/var/run/docker.sock python:${{ matrix.python-version }}-slim /moto/scripts/ci_moto_server.sh &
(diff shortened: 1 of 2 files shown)


bdellegrazie:
Added a test for the `cognito_totp` utility and rebased on master

bdellegrazie:
Hi @bblommers,

Thanks for the review!

* I've used `MOTO_COGNITO_IDP_USER_POOL_ENABLE_TOTP` to match the existing Cognito feature flags, happy to change it if you would prefer.
* Added negative test as requested
* Added additional positive test when the feature is "masked" (i.e. not enabled) - this the same as the original `_enabled` test which has been converted to the real TOTP enabled test. I can change it to `_disabled` if you prefer but that may be confusing given TOTP is enabled, just it's verification is bypassed.
* Fixed missing behaviour in `respond_to_auth_challenge` where the TOTP token wasn't being checked at all.
* Made the totp secret a static constant (for the moment)

bdellegrazie:
@bblommers could you please re-review?

I had to delete the test which verifies the old behaviour (TOTP code of anything) because it doesn't work ServerSide.
If there's a marker which can exclude a test specifically on a server-side run, that would help.

bblommers:
Hi @bdellegrazie! There is no marker, but we typically just add an if-statement to the top of the test:

```python
from unittest import SkipTest

from moto import mock_aws, settings

def test():
        raise SkipTest("Can only change setting in DecoratorMode")
```

Everything else LGTM, so if we can keep the test to verify the old behaviour, I'd be happy to merge.

bdellegrazie:
@bblommers I've added the additional negative test. Anything else?

PR #9791, opened 2026-02-26
ec2: Add AWS-managed launch template tags to EC2 instances

nik-localstack:
## Motivation

The only way to verify that the correct launch template is used when creating an instance, is through the tags `"aws:ec2launchtemplate:id"` and `"aws:ec2launchtemplate:version"`

## Changes

* Adds `"aws:ec2launchtemplate:id"` and `"aws:ec2launchtemplate:version"` tags to an ec2 instance when created from a launch template
* Fixes the version selection logic to select the `$Default` version instead of `$Latest`


bblommers (approved):
Thank you!

PR #9796, opened 2026-02-28
Athena: Implement `TagResource` and `UntagResource`

AaronLionsheep:
Athena resources can now be tagged and untagged after resource creation.


AaronLionsheep:
@bblommers will this PR make it into the 5.1.22 release, or will it need to wait for 5.2?

PR #9815, opened 2026-03-05
Enhance MFA setup handling in CognitoIdpBackend

luchesar:
- Updated MFA logic to include a new challenge for MFA setup when MFA is enabled and the user has not verified their token or enabled SMS MFA.
- Added support for returning available MFA options (SOFTWARE_TOKEN_MFA and SMS_MFA) during the MFA setup challenge.
- Changed the default secret code for software token association to a valid value.

Additionally, added a test to verify the MFA setup challenge behavior when MFA is enabled.

### DIFF
--- a/moto/cognitoidp/responses.py
+++ b/moto/cognitoidp/responses.py
@@ -330,7 +330,7 @@ class CognitoIdpResponse(BaseResponse):
         username = self._get_param("Username")
         message_action = self._get_param("MessageAction")
         temporary_password = self._get_param("TemporaryPassword")
-        user = self.backend.admin_create_user(
+        user, temp_password = self.backend.admin_create_user(
             user_pool_id,
             username,
             message_action,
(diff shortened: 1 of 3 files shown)


moto/cognitoidp/responses.py
```diff
@@ -330,15 +330,20 @@ def admin_create_user(self) -> ActionResult:
...
             self._get_param("UserAttributes", []),
         )
 
-        return ActionResult({"User": user.to_json(extended=True)})
+        return_temp_pass = self._get_header("x-amz-moto-return-temporary-password") == "true"
+        headers = {}
+        if return_temp_pass and temp_password:
+            headers["x-moto-temporary-password"] = temp_password
+
+        return ActionResult({"User": user.to_json(extended=True)}, headers=headers)
```
bpandola, line 347: Don't modify ActionResult.  If you need to manually add a header, use `self.response_headers`.

moto/core/responses.py
bpandola, line 1: We don't want to modify "core" functionality unless absolutely necessary.  See my other comment for how to return a custom header.

bpandola:
Closing due to lack of response.  Feel free to reopen if PR comments are addressed.

PR #9819, opened 2026-03-06
Fix Cloudwatch query sorting

### LINKED ISSUE #9818: Cloudwatch log query sorting is the wrong way around
I looks like the Cloudwatch log query sorting is the wrong way round, and has been since it was implemented.

The query obviously doesn't implement everything, and only does timestamp sorting regardless of which field is passed. But when running a query with `sort ... asc` then the results come back newest first and with `sort ... desc` then they come back oldest first.

```python
from uuid import uuid7

import boto3
import pytest
from moto import mock_aws
from moto.core import DEFAULT_ACCOUNT_ID
from moto.logs import logs_backends


@mock_aws
@pytest.mark.parametrize(
    "order, expected_order_func", [("asc", list), ("desc", reversed)]
)
def test_moto_cloudwatch_sort_order(order, expected_order_func):
    cloudwatch = boto3.client("logs")
    log_group = f"test-log-group-{uuid7()}"
    cloudwatch.create_log_group(logGroupName=log_group)
    log_stream = f"test-log-stream-{uuid7()}"
    cloudwatch.create_log_stream(logGroupName=log_group, logStreamName=log_stream)
    # Avoid the "must be recent" filtering and poke it straight into the Moto backend
    logs_backends[DEFAULT_ACCOUNT_ID]["eu-west-1"].groups[log_group].put_log_events(
        log_stream,
        [{"timestamp": i * 1000, "message": f"Message {i}"} for i in range(5)],
    )
    response = cloudwatch.start_query(
        logGroupName=log_group,
        startTime=0,
        endTime=1000,
        queryString=f"fields @timestamp, @message | sort @timestamp {order}",
    )
    results = cloudwatch.get_query_results(queryId=response["queryId"])
    assert results["results"] == [
        [
            {"field": "@ptr", "value": str(i)},
            {"field": "@timestamp", "value": str(i * 1000)},
            {"field": "@message", "value": f"Message {i}"},
        ]
        for i in expected_order_func(range(5))
    ]
```

This appears to be because `sort_reversed()` is [true iff the sort order is `asc`](https://github.com/getmoto/moto/blob/5.1.21/moto/logs/logs_query/query_parser.py#L12-L17) and that value is [passed to the `reversed` parameter in `sorted()`](https://github.com/getmoto/moto/blob/5.1.21/moto/logs/logs_query/__init__.py#L56), which does a _descending_ sort when reversed.

The current unit tests seem to confirm this, because the events are [put into a list in ascending order](https://github.com/getmoto/moto/blob/5.1.21/tests/test_logs/test_logs_query/test_query.py#L22-L28) (`event_1` from 1 second ago is oldest and comes first first) but the [default query (with descending order)](https://github.com/getmoto/moto/blob/5.1.21/tests/test_logs/test_logs_query/test_query.py#L9-L11) asserts that [they're in their insertion order](https://github.com/getmoto/moto/blob/5.1.21/tests/test_logs/test_logs_query/test_query.py#L40-L48) and [the simplified query (with ascending order)](https://github.com/getmoto/moto/blob/5.1.21/tests/test_logs/test_logs_query/test_query.py#L13) expects them [in reversed order](https://github.com/getmoto/moto/blob/5.1.21/tests/test_logs/test_logs_query/test_query.py#L59-L62).

StuartBertram:
This should fix #9818. It inverts the sorting boolean and updates the query tests to match.

I don't think it should impact anything else. And I'm hoping that some automated test runs will confirm that.

### DIFF
--- a/tests/test_logs/test_logs_query/test_boto3.py
+++ b/tests/test_logs/test_logs_query/test_boto3.py
@@ -106,7 +106,7 @@ def test_get_query_results():
         for row in field
         if row["field"] == "@message"
     ]
-    assert messages == ["event nr 2", "event nr 1"]
+    assert messages == ["event nr 1", "event nr 2"]
 
 
 @mock_aws


bblommers:
Hi @StuartBertram! We do  have some automated tests, and it looks like one other test should be fixed with this change - happy to merge once that's done

StuartBertram:
Okay, I think that should fix it. I'd had a quick look at the Boto tests, but I assumed it would be doing the right thing because it was Boto. But I guess it's a mocked function, so it's Boto querying Moto, which hits the function that I've just edited!

PR #9782, opened 2026-02-24
Fix EventBridge list_rule_names_by_target duplicate rule names

thakoreh:
## Description

Fixes issue #9761 - EventBridge events.list_rule_names_by_target() returns duplicate rule names when a rule has multiple targets with the same ARN.

## Changes

- Added  set in  method to deduplicate rules
- Added regression test 

## Test Results

All existing tests pass:
- test_list_rule_names_by_target ✓
- test_list_rule_names_by_target_using_limit ✓

New test passes:
- test_list_rule_names_by_target_with_duplicate_target_arn ✓

PR #9830, opened 2026-03-08
fix(events): deduplicate list_rule_names_by_target results

### LINKED ISSUE #9761: EventBridge events.list_rule_names_by_target() duplicate rule names
Hello,

There seems to be a difference in behaviour for the EventBridge client `list_rule_names_by_target()` between moto (5.1.21) and AWS when an EventBridge rule has multiple targets with the same AWS Lambda function as the TargetArn.

AWS returns the rule name only once (as you'd expect):

```python
>>> from rich.pretty import pprint
>>> import boto3
>>>e=boto3.client('events')
>>> paginator=e.get_paginator('list_rule_names_by_target')
>>> for resp in paginator.paginate(TargetArn='arn:aws:lambda:ap-southeast-2:<ACCT>:function:test'):
...     pprint(resp)
{
    'RuleNames': ['test'],
    'ResponseMetadata': { ... }
}
```

Moto includes the rule name once for each target that has the specified TargetArn.

```python
{
    'RuleNames': ['test', 'test'],
    'ResponseMetadata': { ...  }
}
```

Thanks -- Murray

SergioChan:
## Summary
- ensure `list_rule_names_by_target` returns each matching rule once, even when a rule has multiple targets with the same `TargetArn`
- replace per-target append loop with a per-rule match check so output mirrors AWS behavior
- add regression test covering a single rule configured with duplicate target ARNs

## Testing
- `pytest -q tests/test_events/test_events.py -k "list_rule_names_by_target"` *(fails locally: missing `boto3` dependency in this environment)*
- `python -m py_compile moto/events/models.py tests/test_events/test_events.py`

## Related
Fixes #9761

### DIFF
--- a/moto/events/models.py
+++ b/moto/events/models.py
@@ -1348,9 +1348,8 @@ class EventsBackend(BaseBackend):
         matching_rules = []
 
         for _, rule in event_bus.rules.items():
-            for target in rule.targets:
-                if target["Arn"] == target_arn:
-                    matching_rules.append(rule)
+            if any(target["Arn"] == target_arn for target in rule.targets):
+                matching_rules.append(rule)
 
         return matching_rules
 
--- a/tests/test_events/test_events.py
+++ b/tests/test_events/test_events.py
@@ -292,6 +292,27 @@ def test_list_rule_names_by_target():
 
 
 @mock_aws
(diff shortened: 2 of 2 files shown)


bblommers:
#9782 is already open to fix this bug

bpandola:
I like this implementation better than the other PR. Reopening...

PR #9849, opened 2026-03-11
Fix multipart S3 composite checksum suffix

### LINKED ISSUE #8762: Multipart S3 uploads - composite checksums are not given the -X suffix
As per the discussion referenced below, composite checksums when generated by AWS S3 are given an `-X` suffix to distinguish them from full object checksums.  Moto does not do this, so composite checksums present as full object checksums which causes a problem when retrieving the object using botocore (which attempts to validate the checksum against a full object checksum, inevitably failing validation).

The attached python code can be run against AWS (with correct permissions) or against moto (with the 'mock' argument) to demonstrate the difference of behaviour.  Note the two ExtraArgs options at line 27/28 which can be used to direct AWS towards composite or full object checksum generation.

[reproduce-checksum-generation-problem.txt](https://github.com/user-attachments/files/19606426/reproduce-checksum-generation-problem.txt)

_Originally posted by @sophieilangolyon1-nhs in https://github.com/getmoto/moto/discussions/8760_

SergioChan:
## Summary
- append the multipart part-count suffix (`-N`) to computed multipart S3 checksums when `ChecksumAlgorithm` is used
- add a multipart regression test that verifies Moto now returns the AWS-style composite checksum format for `ChecksumSHA256`
- keep behavior scoped to multipart completion checksum generation only

## Testing
- `python -m pytest tests/test_s3/test_s3_multipart.py -k "checksum_has_composite_suffix"`

## Related
- Fixes #8762


SergioChan:
Pushed a follow-up commit to address the CI failure in `test_copy_key_with_args__using_multipart`.

### What changed
- Updated multipart checksum formatting so the `-<part_count>` suffix is only added when there are **multiple** parts.
- Single-part uploads now return the base checksum value without `-1`.

### Validation
- `python -m pytest tests/test_s3/test_s3_copyobject.py -k test_copy_key_with_args__using_multipart -q`
- `python -m pytest tests/test_s3/test_s3_multipart.py -k test_multipart_upload_checksum_has_composite_suffix -q`

SergioChan:
Thanks @codecov[bot] for the review feedback. I’m working through the requested changes and will follow up with an updated patch/answers shortly.

PR #9851, opened 2026-03-11
fix(dynamodb): reject redundant condition-expression parentheses

### LINKED ISSUE #9574: Dynamodb doesn't catch redundant parenthesis error on condition expression
For a condition expression like: "#a = :b OR (#c = :d AND #e = :f)"

It passes on moto while real aws gives:

Invalid ConditionExpression: The expression has redundant parentheses;

SergioChan:
## Summary
- reject redundant boolean-group parentheses in DynamoDB condition expressions when the grouped clause does not change precedence (for example `a OR (b AND c)`)
- raise `InvalidConditionExpression` with AWS-aligned message `The expression has redundant parentheses;`
- add regression coverage for both the rejected redundant case and an allowed precedence-changing case (`a AND (b OR c)`)

## Testing
- `python -m pytest tests/test_dynamodb/test_dynamodb_condition_expressions.py -k "redundant_parentheses or allows_required_parentheses"`
- `python -m pytest tests/test_dynamodb/test_dynamodb_condition_expressions.py`

## Related
Fixes #9574

### DIFF
--- a/moto/dynamodb/comparisons.py
+++ b/moto/dynamodb/comparisons.py
@@ -4,7 +4,10 @@ from collections.abc import Iterable
 from decimal import Decimal
 from typing import Any, Optional, Union
 
-from moto.dynamodb.exceptions import ConditionAttributeIsReservedKeyword
+from moto.dynamodb.exceptions import (
+    ConditionAttributeIsReservedKeyword,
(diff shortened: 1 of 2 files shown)


bpandola:
This was already submitted (and merged) as #9847

Issue #9518, opened 2025-12-01
FEATURE REQUEST: add support for AWS AgentCore Memory


juliangrueber:
AWS AgentCore Memory is a new service of AWS to store memory for an Agent. It would be great to mock the core functionalities: 

[Create an event](https://docs.aws.amazon.com/bedrock-agentcore/latest/devguide/short-term-create-event.html)

[Get an event](https://docs.aws.amazon.com/bedrock-agentcore/latest/devguide/short-term-get-event.html)

[List events](https://docs.aws.amazon.com/bedrock-agentcore/latest/devguide/short-term-list-events.html)

[Delete an event](https://docs.aws.amazon.com/bedrock-agentcore/latest/devguide/short-term-delete-event.html)

bblommers:
Welcome to Moto @juliangrueber! Marking it as an enhancement to add this service.

PR's are always welcome, if you want to contribute this yourself.

Mide69:
@bblommers I'd like to take this on. I looked at the earlier attempt in #9856, it got a decent chunk of the way there (bedrockagentcore models/responses/tests) before going stale. I'll build a fresh `bedrockagentcore` service module implementing the four documented operations (`create_event`, `get_event`, `list_events`, `delete_event`) against an in-memory store, following the review comments already left on the old PR so it doesn't repeat the same issues.

Could this be assigned to me?

PR #9856, opened 2026-03-12
Support Bedrock Agent Core

andresliszt:
Hello!

First of all thanks for this amazing project. It has done my job easier :). This is my very first attempt of contribution, I wanted to implement #9518.  I took as reference the existing `bedrockagent ` module.  

I left a TODO in the code, because when I create an event, I don't validate that the real memory exists or not. I think (correct me if I'm wrong) this will be possible once this job is done #9776

### DIFF
--- a/moto/bedrock/urls.py
+++ b/moto/bedrock/urls.py
@@ -1,6 +1,7 @@
 """bedrock base URL and path."""
 
 from ..bedrockagent.responses import AgentsforBedrockResponse
+from ..bedrockagentcore.responses import BedrockAgentCoreResponse
 from .responses import BedrockResponse
 
 url_bases = [
@@ -19,6 +20,7 @@ url_paths = {
     "{0}/knowledgebases/(?P<kb_name>[^/]+)/$": AgentsforBedrockResponse.dispatch,
     "{0}/listTagsForResource$": BedrockResponse.dispatch,
     "{0}/logging/modelinvocations$": BedrockResponse.dispatch,
+    "{0}/memories/(?P<memory_id>[^/]+)/events$": BedrockAgentCoreResponse.dispatch,
     "{0}/model-customization-jobs$": BedrockResponse.dispatch,
     "{0}/model-customization-jobs/(?P<jobIdentifier>[^/]+)$": BedrockResponse.dispatch,
     "{0}/model-customization-jobs/(?P<jobIdentifier>[^/]+)/stop$": BedrockResponse.dispatch,
--- a/moto/bedrockagentcore/models.py
+++ b/moto/bedrockagentcore/models.py
@@ -24,7 +24,7 @@ class Event(BaseModel):
         event_timestamp: float,
         payload: list[dict[str, Any]],
         branch: Optional[dict[str, str]],
-        client_token: Optional[str],
+        client_token: str,
         metadata: Optional[dict[str, dict[str, str]]],
     ) -> None:
         self.region_name = region_name
@@ -92,7 +92,7 @@ class BedrockAgentCoreBackend(BaseBackend):
         event_timestamp: float,
         payload: list[dict[str, Any]],
         branch: Optional[dict[str, str]],
-        client_token: Optional[str],
+        client_token: str,
         metadata: Optional[dict[str, dict[str, str]]],
     ) -> Event:
         """Create an event and return the Event model"""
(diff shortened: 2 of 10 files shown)


moto/bedrock/urls.py
```diff
@@ -19,6 +20,7 @@
     "{0}/knowledgebases/(?P<kb_name>[^/]+)/$": AgentsforBedrockResponse.dispatch,
     "{0}/listTagsForResource$": BedrockResponse.dispatch,
     "{0}/logging/modelinvocations$": BedrockResponse.dispatch,
+    "{0}/memories/(?P<memory_id>[^/]+)/events$": BedrockAgentCoreResponse.dispatch,
     "{0}/model-customization-jobs$": BedrockResponse.dispatch,
```
andresliszt, line 24: First doubt, Do I have to register here all the paths from `BedrockAgentCoreResponse`?

moto/bedrockagentcore/models.py
```diff
@@ -0,0 +1,305 @@
...
+        metadata: Optional[dict[str, dict[str, str]]],
+    ) -> Event:
+        """Create an event and return the Event model"""
+        existing_id = self.client_tokens.get(client_token)
+
+        if existing_id:
+            # AWS doc: If this token matches a previous request, AgentCore
+            # ignores the request, but does not return an error.
+            return self.events[existing_id]
+
```
andresliszt, line 105: I read the AWS doc, I interpreted that ignore means returning the same event?

moto/bedrockagentcore/responses.py
```diff
@@ -0,0 +1,105 @@
...
+        super().__init__(service_name="bedrock-agentcore")
+
+    @property
+    def backend(self) -> BedrockAgentCoreBackend:
+        return bedrockagentcore_backends[self.current_account][self.region]
+
+    def create_event(self) -> str:
+        # Memory is not part of the body
+        match = cast(Match[str], MEMORY_PATH_RE.match(self.path))
+        memory_id = match.group("memory_id")
```
andresliszt, line 37: I'm casting because this is guaranteed by boto3, I think there is no need to do `if match`. Note that I've seen other modules where parsing `self.path` is done directly, also assuming that the `path `contains everything. For example in this case would be something like `memory_id = self.path.split("/")[-2]`, but I think a regex is cleaner

moto/bedrock/urls.py
bpandola: No, you shouldn't have to register any of your paths here.  The reason why there are multiple response classes in this file currently is because multiple services use the same signing key (bedrock) and this was someone's attempt to hack around that.  Your service has a different signing key (bedrock-agentcore) and, again, you shouldn't need to modify this file.

moto/bedrockagentcore/urls.py
```diff
@@ -0,0 +1,8 @@
+# urls.py
+from .responses import BedrockAgentCoreResponse
+
+url_bases = [r"https?://bedrock-agentcore\.(.+)\.amazonaws\.com"]
+
+url_paths = {
+    "{0}/.*$": BedrockAgentCoreResponse.dispatch,
```
bpandola, line 7: This catch-all should be fine... I don't think you'll need to add more specific urls here (or elsewhere) for this service to work in server mode.

moto/bedrockagentcore/responses.py
bpandola: Consider opting in to Moto's new request parsing framework to make your life even easier here.  Set `self.automated_parameter_parsing = True` in your initializer and then you'll be able to get all params, regardless of whether they're in the uri or the body, via a simple call to `self._get_param()`, obviating the need for any of the regex stuff.

moto/bedrockagentcore/responses.py
bpandola, line 1: Have a look at the `responses.py` for Bedrock Runtime [here](https://github.com/getmoto/moto/blob/a3e1138d2010147be610a1d3ee6fced9fee3cf44/moto/bedrockruntime/responses.py) for an example of the latest best practices for request parsing/response serialization.  You shouldn't need to manually utilize `json.dumps` or `json.loads` at all.  Just use `self._get_param()` for pulling the parameters you want and return an `ActionResult` wrapping your dictionary results.

moto/bedrockagentcore/models.py
bpandola: Your interpretation is as good as anyone else's at this point.  Make your best guess or, better yet, verify the behavior against a real AWS backend.  If you get it wrong and someone opens an issue in the future to correct it, it's not a big deal.

bpandola (changes requested):
@andresliszt Thanks for getting started on this service.  I've responded to your questions and made a couple comments to (hopefully) aid you on your way.

andresliszt:
> @andresliszt Thanks for getting started on this service. I've responded to your questions and made a couple comments to (hopefully) aid you on your way.

Thanks for your tips! I will be implementing them as soon as possible.

bpandola:
Closing due to lack of activity. Feel free to re-open if/when you get back to this.

PR #9857, opened 2026-03-12
ec2: fix fleet launch template versioning

nik-localstack:
With https://github.com/getmoto/moto/pull/9790 the version resolution of launch template used when starting a fleet was attempted to be fixed but the resolved template wasn't actually used. 

This PR addresses this bug.


bblommers (approved):
LGTM - thanks for the followup @nik-localstack!

PR #9864, opened 2026-03-13
Pipes: add ManagedState and Start/Stop methods

andresliszt:
Hello ! This PR aims to support start/stop methods in the Aws Pipe. Also, I added the `ManagedState `to the `Pipe ` class to have better transitions between states. I changed the `create_pipe` response, now it responds `CREATING` instead of `RUNNING` .

I took as reference batch job implementation, calling the `ManagedState.advance()` method everytime we want to access to a pipe, e.g `describe_pipe`, this ensures that the pipe is at terminal state (e.g `RUNNING/STOPPED`) when we try to access to it.

I'm willing to support the `update_pipe` method if you think I'm in the right direction with this PR.

Thanks!

### DIFF
--- a/moto/pipes/responses.py
+++ b/moto/pipes/responses.py
@@ -5,7 +5,6 @@ from typing import Any
 from urllib.parse import unquote
 
 from moto.core.responses import ActionResult, BaseResponse, EmptyResult
-from moto.core.utils import iso_8601_datetime_without_milliseconds
 
 from .exceptions import ValidationException
 from .models import EventBridgePipesBackend, pipes_backends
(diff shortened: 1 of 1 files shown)


moto/pipes/models.py
```diff
@@ -61,6 +87,10 @@ def __init__(
         self.last_modified_time = utcnow()
         self.state_reason = None
 
+    @property
+    def current_state(self) -> Optional[str]:
+        return self.status
+
```
andresliszt, line 93: This resource is kind unique, status is stored as `current_state` (CurrentState), but the managed state class works with the attribute `status`.

andresliszt:
Thanks @bpandola , is there any documentation on how to run the tests in the moto server? Is it just Docker right?. So far I've been running the tests in my virtual env only. 

Edit: Never mind, I noticed there's a dev containers!

bpandola:
@andresliszt Do whatever works for you.  What works for me when developing locally is to just run two simple commands from the Moto repo root:

Start the server (in one terminal window): `python moto/server.py -H 0.0.0.0 -p 5000`
Run your tests (in another terminal window): `TEST_SERVER_MODE=true pytest -sv -vv  tests/test_pipes`

It's also easy to just run the server command inside an IDE like PyCharm, so you have access to breakpoints and a debgger.

PR #9877, opened 2026-03-15
fix: permit s3 sns raw delivery without message attributes

holbrookab-augment:
## Summary

This fixes an S3 notification bug in Moto when a bucket notification targets SNS and the SNS subscription to SQS uses `RawMessageDelivery=true`.

Moto's S3 notification path publishes to SNS without message attributes. In the raw SQS subscription delivery path, Moto assumed `message_attributes` was always present and iterated `message_attributes.items()`. When S3-originated publishes passed no attributes, that raised an exception and the notification was silently dropped by the S3 notifier's exception-swallowing path.

This change guards the raw delivery path against missing message attributes and adds a regression test covering `S3 -> SNS -> SQS` raw delivery.

## Test plan

- Added a regression test in `tests/test_s3/test_s3_lambda_integration.py`
- Verified the new test fails against the pre-fix code
- Verified the new test passes with the fix
- Ran:

```bash
python -m pytest tests/test_s3/test_s3_lambda_integration.py -k "raw_delivery" -q
python -m pytest tests/test_s3/test_s3_lambda_integration.py -k "sns_sqs" -q
```

## Reproduction

1. Create an S3 bucket notification with a `TopicConfiguration`
2. Subscribe an SQS queue to that SNS topic with `RawMessageDelivery=true`
3. Upload an object to the bucket
4. Before this fix, the raw SQS subscription does not receive the S3 notification
5. After this fix, the raw SQS subscription receives the expected S3 event payload

### DIFF
--- a/tests/test_s3/test_s3_lambda_integration.py
+++ b/tests/test_s3/test_s3_lambda_integration.py
@@ -424,10 +424,7 @@ def test_put_bucket_notification_sns_sqs_raw_delivery():
     assert message_body["Records"][0]["eventName"] == "ObjectCreated:Put"
     assert message_body["Records"][0]["awsRegion"] == REGION_NAME
     assert message_body["Records"][0]["s3"]["configurationId"] == "SomeID"
-    assert (
-        message_body["Records"][0]["s3"]["object"]["key"]
-        == "incoming%2Fmyfile"
-    )
+    assert message_body["Records"][0]["s3"]["object"]["key"] == "incoming%2Fmyfile"
 
 
 @mock_aws


holbrookab-augment:
[From my LLM]
This failing test looks unrelated to the S3/SNS raw-delivery fix in this PR.

The failure is in the pre-existing Docker-backed S3 -> Lambda integration test `test_objectcreated_put__invokes_lambda[...]`, and the logged error is:

`error running docker: 409 ... volume is in use and cannot be removed`

From Moto's source, this appears to come from the Lambda Docker cleanup path in `moto/awslambda/models.py`, where Docker data volumes are keyed by `code_digest` and volume removal currently re-raises `409` conflicts. That seems consistent with a parallel-worker cleanup race rather than a regression from this change.

PR #9878, opened 2026-03-15
EC2/DynamoDB: Fix AWS verified tests

bblommers:
The `test_create_instant_fleet_with_launch_template_overrides` test was failing against AWS with this error:

> [217](https://github.com/getmoto/moto/actions/runs/23104879978/job/67112354960#step:10:1218)
>               assert len(total_instance_ids) == total_capacity
E               AssertionError: assert 1 == 3
E                +  where 1 = len(['i-00a75cc1ec636ac21'])

The underlying issue is that the `create_fleet(..)` request actually returned an error that the chosen availability zone (`us-east-1e`) did not support the chosen instance type (`t3.nano`.. if memory serves me right).

Fix is to make sure that the instances are always placed in availability zone `a`, as that should support all instance types.

Besides that, we now also print the error - if this happens again, we can at least see what's going wrong.

--------------------
Other tests failed because we can only create 5 VPC's at the same time, and some older VPC's were not deleted correctly.

I'm not sure why they were not cleaned up correctly in the first place - but I have deleted them manually.

--------------------
Note that some tests are still flaky/failing, but at least this should reduce the number of issues.

### DIFF
--- a/tests/test_ec2/test_launch_templates.py
+++ b/tests/test_ec2/test_launch_templates.py
@@ -5,7 +5,7 @@ import pytest
 from botocore.client import ClientError
 
 from moto import mock_aws, settings
-from tests import EXAMPLE_AMI_ID, aws_verified
+from tests import EXAMPLE_AMI_ID, allow_aws_request, aws_verified
 from tests.test_ec2 import MINIMAL_LAUNCH_TEMPLATE_DATA, ec2_aws_verified
 
 from .helpers import assert_dryrun_error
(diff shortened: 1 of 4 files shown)


tests/test_ec2/test_launch_templates.py
```diff
@@ -125,7 +125,12 @@ def test_describe_launch_template_versions_without_name_or_id():
 def test_describe_launch_template_versions_with_just_version():
     cli = boto3.client("ec2", region_name="us-east-1")
     resp = cli.describe_launch_template_versions(Versions=["$Latest"])
-    assert resp["LaunchTemplateVersions"] == []
+    if not allow_aws_request():
+        # When running this against AWS, we can't really be sure that there are no LaunchTemplates,
+        # as, other tests that create templates will be running at the same time
+        # We can (and do) simply verify that we can execute the operation without errors
+        # Only against Moto can we know for sure that no LaunchTemplates exist
+        assert resp["LaunchTemplateVersions"] == []
```
bpandola, line 133: We can't know this with Moto either because the ec2 tests run concurrently, right?

bblommers: Oh, you're absolutely right! Interesting that it never fails with Moto then 🤔

bpandola: Well, there's no guarantee that it will fail, but it's possible.  There are lots of flaky EC2 tests that fail sporadically because they make Describe calls without filters and assume only resources created within the test will be returned.

PR #9883, opened 2026-03-17
Support State Machine Alias

andresliszt:
Supports fully StateMachine Aliases


andresliszt:
Where do you get the exact error msg from ? I mean when raising an error, in this case, for the routing configuration. I was looking the aws docs in boto3, but the error msg is not there.

bblommers:
> Where do you get the exact error msg from ? I mean when raising an error, in this case, for the routing configuration. I was looking the aws docs in boto3, but the error msg is not there.

Usually this involves having to run the test against AWS itself, and copying their exact error message.

The important part is getting an error though, to keep parity with AWS, so if the error message is not exactly the same, that's not a major problem.

PR #9887, opened 2026-03-18
Fix MyPy lint when boto3-stubs is present

mdaniel:
# What?

Fix (or sometimes suppress) type annotation problems that seemingly only show up when one has `boto3-stubs[all]` installed in the venv

# Why?

1. `make lint`
1. `pip install 'boto3-stubs[all]'`
1. `make lint || echo :sad-panda:`

Given what the project is aiming for, it in my opinion invaluable having those type helpers installed, but one cannot do that and also know that `make lint` is behaving as it is going to behave for PRs

Some of the complaint is due to `Mapping[str, str]` versus `dict[str, str]`, some of it is because boto3-stubs actually carries formal type declarations for (e.g.) `CreateAccessKeyResponseTypeDef(TypedDict)`

I did not include the actual dependency because I suspected it was omitted on purpose, but I can do so if was merely an oversight

```diff
diff --git a/requirements-dev.txt b/requirements-dev.txt
index eb1c0c1b3..6cb794832 100644
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -13,6 +13,7 @@ prompt_toolkit

 # type stubs that mypy doesn't install automatically
 botocore-stubs
+boto3-stubs[all]
 types-requests
 types-python-dateutil
 types-PyYAML
```

### DIFF
--- a/moto/core/models.py
+++ b/moto/core/models.py
@@ -16,9 +16,9 @@ from typing import (
 )
 from unittest.mock import patch
 
-import boto3
-import botocore
 import responses
+from boto3.resources.base import ResourceMeta
+from botocore.client import BaseClient
 from botocore.config import Config
 from botocore.handlers import BUILTIN_HANDLERS
 
@@ -268,7 +268,7 @@ botocore_stubber = BotocoreStubber()
 BUILTIN_HANDLERS.append(("before-send", botocore_stubber))
 
 
-def patch_client(client: botocore.client.BaseClient) -> None:
+def patch_client(client: BaseClient) -> None:
     """
     Explicitly patch a boto3-client
     """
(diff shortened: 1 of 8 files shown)


bpandola:
I've encountered this myself and it is annoying.  I'm of the opinion that it's not worth making these changes unless you *do* include the dependency, otherwise it will simply just get out of sync again.  I'll let others chime in if there was some reason for leaving these stubs out originally that I'm not aware of, but I'm tentatively in favor of merging.

bblommers:
One argument against install `boto3_stubs` is that it takes up 160+MB :cry: 

@bpandola What do you think about a `requirements-stub.txt` file, that is essentially `requirements-dev.txt` + these stubs? 

Most developers/contributors can get by with installing `requirements-dev.txt`, but the CI (and whoever wants it) can use `requirements-stub.txt`.

Edit: Actually, that's not a great DevX, if regular people might get a linting error that is only reproduced in the CI

bpandola:
@bblommers I agree with your edit--not a fan of a specialized requirements file.  I'm also leaning against this now as I recently installed the package and not only was it... onerous; but I just happened to be working on writing test assertions for modelled error responses, all of which then required `# type: ignore` comments because the stub files don't include anything beyond the basic error metadata.

bpandola:
I appreciate the suggestion here, but after further consideration it's not a constraint we want to add to the project at this point.

PR #9891, opened 2026-03-21
add imdsv2 mocking to instance_metadata endpoints

markjm:
Thank you for this great project. Reporting and fixing a few missing endpoints when IMDSv2 is enabled! Please let me know if im missing something (i read through the contributing guide and think i did everything by the book)

-----

I added a gist with some repro tests here https://gist.github.com/markjm/efae7f71c8dd58d0e3c118097140b0d9 (claude helped with a quick repro creation)

Basically, when IMDSv2 is enabled, there are expected calls to `/latest/api/token` which we want to return a fake token, then also some extra metadata paths which we may care about

Added the imdsv2 path + those extra possible metadata. And added tests for the same.

### DIFF
--- a/moto/moto_server/werkzeug_app.py
+++ b/moto/moto_server/werkzeug_app.py
@@ -211,7 +211,9 @@ class DomainDispatcherApplication:
 
         if path_info.startswith("/moto-api") or path_info == "/favicon.ico":
             host = "moto_api"
-        elif path_info.startswith("/latest/meta-data/"):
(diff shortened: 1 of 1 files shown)


markjm:
I presume most cases under mock wouldnt even try to get credentials, but we have code specifically using `requests` to hit these endpoints to get metadata directly which we would like mocked!

markjm:
thanks @bblommers for kicking off the tests. didnt initially notice the servertests, so fixed that up as well!

PR #9912, opened 2026-03-24
Support ACM certs without CN

charhate:
Problem
                                                                                                                                                                                                                            
 `import_certificate` raises `IndexError: list index out of range` when the certificate has no Common Name (CN) in its subject. The bug is in acm/models.py:
                                                                                                                                                                                                                            
Modern certificates commonly omit CN and rely solely on Subject Alternative Names (SANs). Real AWS ACM accepts such certificates without issue.
                                                                                                                                                                                                                            
  Fix                                                       
  - CN is parsed defensively; falls back to the first SAN when absent (matching real AWS ACM DomainName behaviour), or "" if there are no SANs either                                                                       
  - Issuer CN is also parsed defensively
  - SANs are parsed once in __init__ into self.sans and reused in describe(), removing the duplicate SAN-parsing block that was there before                                                                                
  - domain_names in describe() guards against adding an empty string when both CN and SANs are absent


bpandola (approved):
Thanks @charhate

PR #9915, opened 2026-03-26
Fix EventBridge _send_to_sqs_queue crash when target uses InputTransformer

Scotsoo:
When an EventBridge rule has a target with an `InputTransformer` that forwards to SQS, calling `put_events` crashes with a `KeyError: 'time'`.

`EventTemplateParser.parse()` returns the rendered template as a **string**, which then replaces the original event dict. `_send_to_sqs_queue` unconditionally accesses `event["time"]`, which fails because the event is no longer a dict.

**Fix:**
- Guard the time formatting in `_send_to_sqs_queue` so it only applies when the event is still a dict
- Send transformed string events directly as the SQS message body, bypassing `json.dumps()` which would double-serialize it (wrapping it in extra quotes)

Two integration tests added covering InputTransformer-to-SQS on both the default and custom event buses.

### DIFF
--- a/moto/events/models.py
+++ b/moto/events/models.py
@@ -234,7 +234,7 @@ class Rule(CloudFormationModel):
     ) -> None:
         from moto.sqs import sqs_backends
 
-        if isinstance(event, dict) and "time" in event:
+        if "time" in event:
             event["time"] = iso_8601_datetime_without_milliseconds(
                 utcfromtimestamp(float(event["time"]))  # type: ignore[arg-type]
             )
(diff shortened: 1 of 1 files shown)


bpandola (changes requested):
If an `isinstance` check is necessary here, we've got bigger problems further up the call stack, but as far as I can tell (from a glance) both of your tests will pass without it.  Can you add a test that covers that part of the guard clause or simply remove it?

Scotsoo:
I've removed that check @bpandola, was being over cautious!

PR #9932, opened 2026-03-30
S3: migrate to new response serialization

bpandola:
Moto's S3 backend is a bit of an anomaly, circumventing a lot of the core response flow, which necessitated slightly different integration patterns when migrating to the new ActionResult response serialization.  Given the importance and scope of the S3 backend, I opted to do the migration without making a single update to the model.  Structured dict responses are created in each response method that accurately reproduce the Jinja templates being replaced, including all logic and hardcoded values.

A new `S3Serializer` class has been added to encapsulate the minor customizations required on top of the `rest-xml` serializer for S3 responses.  A couple of serialization bugs that surfaced during the migration have also been fixed, with associated cases added to the protocol test suite.

There is a lot more that can (and will) be done in follow-up PRs.  For example, there is still a lot of direct response header manipulation in `responses.py`, which can be handled automatically by the new serializer.  Error messages are still using Jinja templates, and will need to be migrated as well.  And finally, the S3 backend is still manually parsing input and has not yet been opted in to the new request parsing flow.

### DIFF
--- a/tests/test_s3/test_s3_lock.py
+++ b/tests/test_s3/test_s3_lock.py
@@ -483,6 +483,10 @@ def test_put_object_legal_hold_with_versions():
         VersionId=version_id_1,
         LegalHold={"Status": "ON"},
     )
+    resp = s3_client.get_object_legal_hold(
+        Bucket=bucket_name, Key=key_name, VersionId=version_id_1
+    )
+    assert resp["LegalHold"]["Status"] == "ON"
 
     # put an object on the same key, effectively creating a version 2 of the object
     put_obj_2 = s3_client.put_object(Bucket=bucket_name, Body=b"test", Key=key_name)
@@ -494,6 +498,10 @@ def test_put_object_legal_hold_with_versions():
         VersionId=version_id_2,
         LegalHold={"Status": "ON"},
     )
+    resp = s3_client.get_object_legal_hold(
+        Bucket=bucket_name, Key=key_name, VersionId=version_id_2
+    )
+    assert resp["LegalHold"]["Status"] == "ON"
 
     # assert that the version 1 is locked
     head_obj_1 = s3_client.head_object(


moto/s3/responses.py
```diff
@@ -375,7 +390,7 @@ def _send_response(
         if isinstance(body, str):
             body = body.encode("utf-8")
 
-        if body and "content-type" not in headers:
+        if body and "content-type" not in (header.lower() for header in headers.keys()):
```
bpandola, line 393: The serializer uses "Content-Type" which was resulting in a duplicate header.  This is an inelegant (but localized) fix for a problem that exists throughout Moto.  Eventually, we should ensure that all headers are CaseInsensitive dicts.

moto/s3/responses.py
```diff
@@ -1631,8 +2012,21 @@ def get_object_attributes(self) -> TYPE_RESPONSE:
...
+            "Checksum": {f"Checksum{algo}": value for algo, value in checksum.items()}
+            if checksum
+            else None,
+            "ObjectSize": response_keys["size"],
+            "StorageClass": response_keys["storage_class"],
+        }
+
+        self.data["Action"] = "GetObjectAttributes"
+        status, headers, body = self.serialized(ActionResult(result))
+        headers.update(response_headers)
```
bpandola, line 2028: In a future update, we can just add the `LastModified` field to the result dict.  The serializer will automatically put in the correct response header.

tests/test_core/test_request_passthrough.py
```diff
@@ -49,7 +49,7 @@ def test_passthrough_calls_for_entire_service() -> None:
 
             resp = _aws_request(list_buckets_url)
             assert resp.status_code == 200
-            assert b"<Buckets></Buckets>" in resp.content
+            assert b"ListAllMyBucketsResult" in resp.content
```
bpandola, line 53: S3 (and Moto's response serializer) use short empty elements, e.g. `<Buckets/>`. I changed the assertion here to the root node name, which will always be present regardless of the number of buckets returned.

tests/test_s3/test_server.py
```diff
@@ -109,7 +109,7 @@ def test_s3_server_ignore_subdomain_for_bucketnames():
 
         res = test_client.put("/mybucket", "http://foobaz.localhost:5000/")
         assert res.status_code == 200
-        assert b"mybucket" in res.data
+        assert "mybucket" in res.headers.get("Location")
```
bpandola, line 112: Moto had been returning an XML body for the CreateBucket response that included the bucket name, but AWS (and now Moto with this PR) returns an empty body and the bucket name is in a `Location` header, which we now assert against.

moto/s3/2006-03-01/service-2.moto-extras.json
bpandola, line 1: S3 is a bit of an oddball in that the vast majority of its responses' root tag names do not follow the `rest-xml` guidelines.    Here we supplement the Botocore model with a `resultWrapper` trait for all custom root tag names.

PR #9960, opened 2026-04-05
WAFV2: support `Arn` parameter in `get_web_acl`

GtrMo:
The [`wafv2.get_web_acl`](https://docs.aws.amazon.com/boto3/latest/reference/services/wafv2/client/get_web_acl.html) accepts either an `ARN` or the `(Name, Id, Scope)`. 

This isn't documented (or I couldn't find it), but the ARN takes precedence over the `(Name, Id, Scope)` tuple. Please let me know if there is additional tests needed for this


bpandola (approved):
Thanks @GtrMo

PR #9964, opened 2026-04-07
fix: validate KeyId parameter in KMS decrypt

### LINKED ISSUE #9907: kms decrypt ignores key-id
## Summary

Once a piece of ciphertext has been created by a copy of Moto, it will then subsequently honor that ciphertext going forward, so long as the embedded key-id is available at all on the server

## Current Behavior

```sh
k_id=$(awslocal kms create-key --output text --query KeyMetadata.KeyId)
enc=$(awslocal kms encrypt --key-id $k_id --plaintext "$(echo "hello world" | base64 -w 0)" --output text --query CiphertextBlob)
awslocal kms decrypt --key-id "$(python3 -m uuid -u uuid4)" --ciphertext "$enc" --output text --query Plaintext | base64 --decode
```
emits "hello world"

## Expected Behavior

Unknown --key-id parameters should produce `botocore.exceptions.ClientError: An error occurred (AccessDeniedException) when calling the Decrypt operation`

```sh
k_id=$(awslocal kms create-key --output text --query KeyMetadata.KeyId)
enc=$(awslocal kms encrypt --key-id $k_id --plaintext "$(echo "hello world" | base64 -w 0)" --output text --query CiphertextBlob)
if awslocal kms decrypt --key-id "$(python3 -m uuid -u uuid4)" --ciphertext "$enc"; then
  echo "a random key-id should have terminated the exchange" >&2
  exit 1
fi
```

---

If my mental model is correct, there are two general approaches to fixing this experience:
- have the enforcement be opt-in
- use the original key-id as part of the ciphertext

For the opt-in, the `decrypt`and `re_encrypt` functions could compare the input key-id to the key-id that is currently embedded in the ciphertext. The hazard is that such comparison behavior would need to be retrofitted everywhere the kms backend touches the ciphertext, but would allow those who don't care about enforcement to continue to not concern themselves

In a more strict mode, one could essentially need the "password" of the original key-id to unlock the cipher text, and then the code could go back to behaving as it does now, including continuing to embed the key-id, IV, and tag to avoid having to make the change wide-ranging. It need not be industrial strength cryptography, merely not unconditionally trusting the input key-id

raajheshkannaa:
Fixes #9907

## Summary

- When `KeyId` is passed to the `Decrypt` API, validate it matches the key embedded in the ciphertext
- Raise `AccessDeniedException` if the provided key does not match, consistent with AWS behavior
- `KeyId` remains optional; omitting it preserves existing behavior

## Changes

- `models.py`: Accept optional `key_id` in `decrypt()`, resolve via `any_id_to_key_id()`, and compare against the actual key
- `responses.py`: Read and validate the `KeyId` parameter before passing to the backend
- `test_kms.py`: Add `test_decrypt_validates_key_id` covering correct key, missing key, and wrong key scenarios

### DIFF
--- a/moto/kms/models.py
+++ b/moto/kms/models.py
@@ -15,6 +15,7 @@ from moto.utilities.tagging_service import TaggingService
 from moto.utilities.utils import get_partition
 
 from .exceptions import (
+    AccessDeniedException,
     InvalidKeyUsageException,
     KMSInvalidMacException,
     ValidationException,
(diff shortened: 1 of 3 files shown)


bblommers (approved):
LGTM! Thanks for adding support for this feature @raajheshkannaa

PR #10008, opened 2026-05-03
fix(lambda): preserve Unix file permissions in zip2tar

gavin913-lss:
The `zip2tar` utility in Lambda does not preserve Unix file permissions from ZIP external attributes. When ZIP entries have executable permissions (e.g., Go/Rust `bootstrap` binaries), the resulting tar loses those permissions, causing "permission denied" errors at runtime.

Fix: Read Unix permissions from `zipinfo.external_attr >> 16` and apply to `TarInfo.mode` when present.


bpandola:
AI spam (that doesn't fix the issue or pass quality checks).

PR #10011, opened 2026-05-04
Check if protocol is in the service's `supported_protocols` before switching to it

edgarrmondragon:
Fixes

- https://github.com/getmoto/moto/issues/10010


bpandola:
@edgarrmondragon I appreciate the attempt at a fix here, but I'm going to go with something more S3-specific (#10012).

edgarrmondragon:
> @edgarrmondragon I appreciate the attempt at a fix here, but I'm going to go with something more S3-specific (#10012).

Thanks!

PR #10014, opened 2026-05-05
Route53: update_health_check() should handle falsy values correctly

### LINKED ISSUE #10013: Bug: Impossible to re-enable health checks in version 5.2
My Route53 health check tests using moto, where I disable and re-enable health checks, started failing after upgrading to moto 5.2. I think I found the reason, or at least I found a working fix (for me).

The issue seem to have been introduced in https://github.com/getmoto/moto/pull/9875, which switched Route53 from manual `xmltodict.parse()` to `automated_parameter_parsing = True` with `self._get_param()`. The new framework parses parameters using the AWS service model, which means booleans are now delivered to `models.py` as actual Python bool values instead of strings.

The bug at https://github.com/getmoto/moto/blob/master/moto/route53/models.py#L869-L870:
```
if health_check_args.get("disabled"):          # broken: False is falsy
    health_check.disabled = health_check_args.get("disabled")
```

In 5.1, `xmltodict` returned the string "false" for <Disabled>false</Disabled>. A non-empty string is truthy in Python, so the assignment ran.

In 5.2, `self._get_param("Disabled")` returns actual False. if False: never executes, the field is never updated. Health checks can be disabled but not re-enabled.

Fix that solved the problem for me when trying it locally:
```
if health_check_args.get("disabled") is not None:
    health_check.disabled = health_check_args.get("disabled")
```

I think the same pattern is broken for inverted and enable_sni in the same function for the same reason.

bblommers:
Fixes #10013 

Values should always be updated if they are not None. The user only mentioned the boolean values not being set, but this applies to all falsy values - `[]`, `0` are also valid values.

### DIFF
--- a/moto/route53/models.py
+++ b/moto/route53/models.py
@@ -848,31 +848,31 @@ class Route53Backend(BaseBackend):
         if not health_check:
             raise NoSuchHealthCheck(health_check_id)
 
-        if health_check_args.get("ip_address"):
+        if health_check_args.get("ip_address") is not None:
             health_check.ip_address = health_check_args.get("ip_address")
-        if health_check_args.get("port"):
+        if health_check_args.get("port") is not None:
             health_check.port = health_check_args.get("port")
-        if health_check_args.get("resource_path"):
+        if health_check_args.get("resource_path") is not None:
             health_check.resource_path = health_check_args.get("resource_path")
-        if health_check_args.get("fqdn"):
+        if health_check_args.get("fqdn") is not None:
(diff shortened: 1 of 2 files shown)


bpandola (approved):
Thanks for fixing this, @bblommers!  These boolean strings have been a real pain during the request parsing/response serialization refactor, but I still think it's important that we convert them.  Unfortunately, the three in question here had no test coverage and I failed to spot them.  I appreciate you jumping on this so quickly!

PR #10021, opened 2026-05-12
fix(lambda): preserve Unix file permissions in zip2tar conversion

### LINKED ISSUE #9947: zip2tar in Lambda does not preserve Unix file permissions, causing permission denied when executing Lambda handler
## Description

`zip2tar` in `moto/awslambda/models.py` converts a deployment ZIP to a TAR stream,
but does not copy Unix file permissions from `zipinfo.external_attr` to `tarinfo.mode`.
`tarfile.TarInfo` defaults to a mode without the execute bit,
so Lambda handler binaries (e.g. `bootstrap` in Go/Rust runtimes) cannot be executed.

## Moto version

5.1.22

## Steps to reproduce

1. Create a Lambda function using a ZIP that contains a binary with execute permission (e.g. `bootstrap` with mode `0o755`)
2. Invoke the Lambda function via Moto

## Expected behavior

The Lambda function is invoked successfully.

## Actual behavior

Lambda invocation fails with: `fork/exec /var/task/bootstrap: permission denied`

## Root cause

In `moto/awslambda/models.py`, `zip2tar` creates `TarInfo` without setting `mode`:

```python
tarinfo = tarfile.TarInfo(name=zipinfo.filename)
tarinfo.size = zipinfo.file_size
tarinfo.mtime = calendar.timegm(zipinfo.date_time) - timeshift
# zipinfo.external_attr (which stores Unix permissions) is never read
```

## Proposed fix

```python
tarinfo.mtime = calendar.timegm(zipinfo.date_time) - timeshift
if zipinfo.external_attr >> 16:
    tarinfo.mode = zipinfo.external_attr >> 16
```

algojogacor:
## Summary

Fixes #9947

Lambda functions packaged as ZIP archives containing binaries with execute permissions (e.g., Go/Rust `bootstrap` binaries) fail with `fork/exec /var/task/bootstrap: permission denied` because `zip2tar` does not copy Unix file permissions from the ZIP archive to the TAR stream.

## Root Cause

In `moto/awslambda/models.py`, `zip2tar` creates `TarInfo` objects without setting `mode`. `tarfile.TarInfo` defaults to a mode without the execute bit (`0o644`), so binaries lose their execute permission during the ZIP-to-TAR conversion.

The Unix permissions are stored in `zipinfo.external_attr` (upper 16 bits), but this field is never read by the current implementation.

## Fix

Extract Unix permissions from `zipinfo.external_attr >> 16` and apply them to `tarinfo.mode`, preserving the original file permissions from the ZIP archive. The `if zipinfo.external_attr >> 16` guard skips entries without Unix permission data (e.g., entries created on Windows).

## Changes

- `moto/awslambda/models.py`: 2 lines added — extract and apply Unix permissions in `zip2tar`

## Testing

- **Go/Rust Lambda with executable bootstrap binary**: handler invoked successfully without permission error ✅
- **Python Lambda (no binary)**: no regression, existing behavior preserved ✅
- **ZIP created on Windows (external_attr = 0)**: guard clause skips, no unintended mode set ✅

### DIFF
--- a/moto/awslambda/models.py
+++ b/moto/awslambda/models.py
@@ -92,8 +92,14 @@ def zip2tar(zip_bytes: bytes) -> io.BytesIO:
             tarinfo = tarfile.TarInfo(name=zipinfo.filename)
             tarinfo.size = zipinfo.file_size
             tarinfo.mtime = calendar.timegm(zipinfo.date_time) - timeshift
-            if zipinfo.external_attr >> 16:
-                tarinfo.mode = zipinfo.external_attr >> 16
+            tarinfo.mode = zipinfo.external_attr >> 16
+            # Ensure minimum read bits for all users.
+            # Lambda extracts these files to /var/task and needs
+            # at least 0444 (read) for files and 0555 (read+execute) for dirs.
+            if zipinfo.is_dir():
+                tarinfo.mode |= 0o555
+            else:
+                tarinfo.mode |= 0o444
             infile = zipf.open(zipinfo.filename)
             tarf.addfile(tarinfo, infile)
 
--- a/moto/awslambda/models.py
+++ b/moto/awslambda/models.py
@@ -92,6 +92,8 @@ def zip2tar(zip_bytes: bytes) -> io.BytesIO:
             tarinfo = tarfile.TarInfo(name=zipinfo.filename)
             tarinfo.size = zipinfo.file_size
             tarinfo.mtime = calendar.timegm(zipinfo.date_time) - timeshift
+            if zipinfo.external_attr >> 16:
+                tarinfo.mode = zipinfo.external_attr >> 16
             infile = zipf.open(zipinfo.filename)
             tarf.addfile(tarinfo, infile)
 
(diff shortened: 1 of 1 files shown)


algojogacor:
## Fix: zip2tar minimum permission enforcement

The root cause of the `PermissionError` reading `/var/task/index.py` is that `zipinfo.external_attr >> 16` can be `0` for Lambda ZIP files built on systems without Unix permission tracking (e.g., Windows, certain CI pipelines). The original code:

```python
if zipinfo.external_attr >> 16:
    tarinfo.mode = zipinfo.external_attr >> 16
```

would silently skip setting `tarinfo.mode` when the upper 16 bits are zero, leaving it at the default `0o000`. When the tar is extracted, files have no read permissions at all.

### Changes
1. **Always assign** `tarinfo.mode = zipinfo.external_attr >> 16` (even if 0)
2. **OR in minimum permission bits** after assignment:
   - Files: `|= 0o444` (read for owner, group, world)
   - Directories: `|= 0o555` (read+execute for owner, group, world)

This preserves original Unix permissions from ZIPs that have them, while guaranteeing that Lambda can always read the extracted files.

bpandola (approved):

PR #10029, opened 2026-05-20
fix(route53): Ensure that alias type records do not return TTL prop

pgibson1-godaddy:
`TTL` is a mandatory input when creating most record types with alias records being an exception to this rule. Alias records use the TTL of their target and therefore don't take `TTL` as an input nor do they return that property in list operations.


bpandola (approved):
Thanks @pgibson1-godaddy

PR #10030, opened 2026-05-20
fix(acm): Preserve default cert_options when partial options are provided

pgibson1-godaddy:
Current behaviour is that the entire certificate options object is updated with whatever the user passes in during creation but this doesn't respect the defaults and can result in HTTP 500 errors later.

For example creating a certificate with just `CertificateTransparencyLoggingPreference` set results in a `KeyError` when listing the cert as moto tries to read the non-existent `Export` key.


bpandola (approved):
Thanks @pgibson1-godaddy

PR #10031, opened 2026-05-20
Support full parameter ARN lookups in SSM get_parameters

### LINKED ISSUE #9396: Testing of cross-account SSM parameters doesn't seem to be working
I'm trying to write a test case that tests fetching SSM parameter information across accounts. I'm mocking this up by setting MOTO_ACCOUNT_ID to create SSM parameters in the target account, creating the RAM share to let it view across accounts. I seem to be encountering a few issues here:
```
    monkeypatch.setenv('MOTO_ACCOUNT_ID', '222222222222')
    response = ssm.put_parameter(
        Name='/Parameter1',
        Value='Value1',
        Type='String',
        Tier='Advanced'
    )
    print(f"Created Parameter #1: {response}")
    response = ssm.put_parameter(
        Name='/Parameter2',
        Value='Value2',
        Type='String',
        Tier='Advanced'
    )
    print(f"Created Parameter #2: {response}")
    ram.create_resource_share(
        name='ShareTest',
        resourceArns=[
            "arn:aws:ssm:ap-southeast-2:222222222222:parameter/Parameter1",
            "arn:aws:ssm:ap-southeast-2:222222222222:parameter/Parameter2"
        ],
        principals=[
            '111111111111'
        ]
    )

    monkeypatch.setenv('MOTO_ACCOUNT_ID', '111111111111')
    # Do testing
```
- get_parameters doesn't seem to support fetching using the full ARN (it looks like get_parameter was fixed by #7748, but this fix didn't touch related methods).
- get_parameter works by ARN when I don't change the MOTO_ACCOUNT_ID environment variable but breaks otherwise (this could be correct behaviour, as there is an issue creating a RAM resource share for these parameters - but see below).
- create_resource_share fails with `You cannot share the selected resource type. ` despite the parameters being configured as 'Advanced'. One potentially related observation here is that 'Tier' isn't returned by the put_parameter API call despite it being expected according to the Boto3 documentation (which suggests to me that this isn't currently being captured, and only Advanced tier parameters can be used with RAM according to AWS documentation).

mokashang:
Fixes #9396

## Problem

`SimpleSystemManagerBackend.get_parameter()` (single lookup) strips the SSM ARN prefix before resolving a parameter, so it accepts both a bare name (`/test`) and a full ARN (`arn:aws:ssm:<region>:<account>:parameter/test`). This was added so callers can fetch a parameter by its ARN.

`get_parameters()` (the batch variant used by the `GetParameters` API) was never updated to match. It guards the lookup with:

```python
if name.split(":")[0] in self._parameters:
```

When `name` is a full ARN, `name.split(":")[0]` is `"arn"`, which is never a stored parameter, so the ARN is skipped and ends up in the response's `InvalidParameters` list. The reporter in #9396 hit this: `get_parameter` works by ARN but `get_parameters` does not.

## Fix

Strip the ARN prefix the same way `get_parameter()` does before performing the membership check, so batch lookups by ARN resolve consistently with single lookups. The original name is still passed through to `get_parameter()` (which handles version/label suffixes) and used as the result key, so `InvalidParameters` detection in the response layer is unaffected.

## Testing

- Added `test_get_parameters_by_arn`, parametrized over a bare name and a full ARN, mirroring the existing `test_get_parameter` regression test. It fails before the change (the ARN form is returned under `InvalidParameters`) and passes after.
- `pytest tests/test_ssm/` — 180 passed.
- `black --check` and `ruff check` clean on the changed files; `mypy moto/ssm/models.py` reports no issues.

### DIFF
--- a/moto/ssm/models.py
+++ b/moto/ssm/models.py
@@ -1879,7 +1879,10 @@ class SimpleSystemManagerBackend(BaseBackend):
             )
 
         for name in set(names):
-            if name.split(":")[0] in self._parameters:
+            lookup_name = name
+            if lookup_name.startswith(self.ssm_prefix):
+                lookup_name = lookup_name.replace(self.ssm_prefix, "")
+            if lookup_name.split(":")[0] in self._parameters:
                 try:
                     param = self.get_parameter(name)
 
--- a/tests/test_ssm/test_ssm.py
+++ b/tests/test_ssm/test_ssm.py
@@ -662,6 +662,31 @@ def test_get_parameter(param):
     )
 
 
+@pytest.mark.parametrize(
+    "param",
+    ["/test", "arn:aws:ssm:us-east-1:123456789012:parameter/test"],
+)
+@mock_aws
+def test_get_parameters_by_arn(param):
+    # Setup
+    client = boto3.client("ssm", region_name=SSM_REGION)
+    client.put_parameter(
+        Name="/test", Description="A test parameter", Value="value", Type="String"
+    )
+
+    # Execute
+    response = client.get_parameters(Names=[param], WithDecryption=False)
+
+    # Verify
+    assert response["InvalidParameters"] == []
+    assert len(response["Parameters"]) == 1
+    parameter = response["Parameters"][0]
+    assert parameter["Name"] == "/test"
+    assert parameter["Value"] == "value"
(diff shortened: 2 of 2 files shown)


bpandola:
AI generated and doesn't address the main concern raised in the linked issue.

PR #10032, opened 2026-05-20
fix(s3): adjust SQS bucket notification timestamp format

milas:
Fix S3 bucket notifications for SQS messages to have timestamps in the exactly format expected by AWS.

This is mentioned in the AWS docs[1] (already linked in the source): 

> "eventTime":"The time, in ISO-8601 format (for example, 1970-01-01T00:00:00.000Z) when Amazon S3 finished processing the request"

In moto terms, this is our friend `iso_8601_datetime_with_milliseconds`

[1]: https://docs.aws.amazon.com/AmazonS3/latest/userguide/notification-content-structure.html

### DIFF
--- a/moto/s3/notifications.py
+++ b/moto/s3/notifications.py
@@ -1,16 +1,13 @@
 import copy
 import json
-from datetime import datetime
 from enum import Enum
 from typing import TYPE_CHECKING, Any
 from urllib.parse import quote_plus
 
-from moto.core.utils import unix_time
+from moto.core.utils import iso_8601_datetime_with_milliseconds, unix_time
 
 if TYPE_CHECKING:
     from moto.s3.models import FakeBucket
-
-_EVENT_TIME_FORMAT = "%Y-%m-%dT%H:%M:%S.%f"
 
 
 class S3NotificationEvent(str, Enum):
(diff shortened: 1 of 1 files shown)


bpandola (changes requested):
@milas Was there a sequence of API calls resulting in an error that led you to this PR?  If so, it'd be good to capture that in a test.

milas:
Ah, apologies for vagueness. It was a discovered via downstream expectation when using motoserver!

Specifically, I was using moto for an integration test w/ [Vector's S3 source](https://vector.dev/docs/reference/configuration/sources/aws_s3/) (which uses SQS bucket notifications).

The Rust type's deserializer is strict and refuses unless it's got the `Z` at the end etc, so it refuses to handle the event notification messages currently. With these changes, it works as expected 🎉 

```rust
// https://docs.aws.amazon.com/AmazonS3/latest/dev/notification-content-structure.html
#[derive(Clone, Debug, Deserialize, Serialize)]
#[serde(rename_all = "PascalCase")]
pub struct S3Event {
    pub records: Vec<S3EventRecord>,
}

#[derive(Clone, Debug, Deserialize, Serialize)]
#[serde(rename_all = "camelCase")]
pub struct S3EventRecord {
    pub event_version: S3EventVersion,
    pub event_source: String,
    pub aws_region: String,
    pub event_name: S3EventName,
    pub event_time: DateTime<Utc>,

    pub s3: S3Message,
}
```
https://github.com/vectordotdev/vector/blob/2084d534f269bb1fa2d508bbc28a634d57d4dc32/src/sources/aws_s3/sqs.rs#L1004C1-L1021C2

bpandola (approved):

PR #10038, opened 2026-05-26
Refactor: ResourceGroupsTaggingAPI

### LINKED ISSUE #9544: FEATURE REQUEST: Add support for EC2, RDS, and ASG resources to resourcegroupstaggingapi tag and untag resources
Writing tags with this api is not currently supported for these services (https://docs.getmoto.org/en/latest/docs/services/resourcegroupstaggingapi.html). But it would be extremely helpful for some tests I would like to write to ensure consistent tagging accross multiple services.

The services listed are just the 3 that I am currently working with, but support for a wider range of services than just these would be very helpful. Not sure if there's a convenient way for moto to generalize the implementation here or not.

bpandola:
Allow service backends to opt in to the Resource Groups Tagging API via a simple interface, obviating the need for the RGTAPI backend to directly import participating backends or manage per-service tagging logic.

Migrates EC2, Lambda, RDS, and S3 as the first consumers to illustrate the benefits and ensure this new method works for key AWS services.  Follow-on PRs will migrate the rest of the existing services, greatly simplifying the RGTAPI backend.

## Key Benefits

- **Eliminates Boilerplate** — Services implement tagging in their own backend; RGTAPI discovers them automatically instead of hardcoding each one.
- **Centralizes Filtering Logic** — Tag filtering and matching now live in one place instead of scattered across services.
- **Auto-Discovery** — `iter_taggable_backends()` dynamically finds all tagging-capable services; adding support requires zero changes to the RGTAPI backend.
- **Better Separation** — RGTAPI handles the higher-level concerns, services handle their own details; both sides of the tagging contract become easier to understand and update.

Closes #9544

### DIFF
--- a/moto/autoscaling/models.py
+++ b/moto/autoscaling/models.py
@@ -3,11 +3,13 @@ from __future__ import annotations
 import itertools
 import math
 from collections import OrderedDict
+from collections.abc import Iterator
 from datetime import datetime
 from typing import Any
 
 from moto.core.base_backend import BackendDict, BaseBackend
 from moto.core.common_models import BaseModel, CloudFormationModel
+from moto.core.resource_tagging import TaggableResourcesMixin, TaggedResource
 from moto.core.types import Base64EncodedString
 from moto.core.utils import utcnow
 from moto.ec2 import ec2_backends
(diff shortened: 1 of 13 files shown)


bpandola:
Hey @bblommers, I know you've had some thoughts about this subject over the years.  I've given it a lot of thought myself, and I think I've come up with a good approach.  Let me know what you think!

bblommers (approved):
Hi @bpandola, thanks for tackling this! Using a Mixin-class to support this functionality is a great approach - very nice to see the service-specific logic removed from the RT-api 💯

PR #10044, opened 2026-05-31
ResourceGroupsTaggingAPI: finalize integration with TaggableResourcesMixin

bpandola:
This changeset completes the major refactor of the **Resource Groups Tagging API** backend (initiated in #10038) to dramatically simplify `moto/resourcegroupstaggingapi/models.py` by migrating tagging support to individual service backends via the new `TaggableResourcesMixin` pattern--replacing hundreds of lines of imperative, service-specific code with a clean, extensible abstraction.

> [!NOTE]
> It's possible that the `tag_resource` method on `TaggableResourcesMixin` may clash with an existing `tag_resource` method already defined on a service backend class.
>
>When this happens we:
> - use the method signature from `TaggableResourcesMixin` as the canonical interface
> - translate/adapt any `tag_resource` method signature differences in `responses.py`
>
> This allows the mixin's tagging interface to be consistent across all services while still accommodating tagging requests made with an individual service's native client API.

### DIFF
--- a/moto/acm/models.py
+++ b/moto/acm/models.py
@@ -3,7 +3,7 @@ import calendar
 import datetime
 import ipaddress
 import re
-from collections.abc import Iterable
+from collections.abc import Iterable, Iterator
 from typing import Any
 
 import cryptography.hazmat.primitives.asymmetric.rsa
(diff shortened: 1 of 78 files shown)


bpandola (commented):
The missing coverage was/is mostly code that used to be in the Resource Groups Tagging API backend--and was not covered there--that has now been moved to an individual service backend and is still not covered by a test. I did add some tests to cover missing lines where possible, but I'm going to leave things short of 100%. Connect Campaigns, in particular, never had any test coverage and doesn't seem to have ever worked in its previous state, but I figured it best to migrate it as is and let any bugs surface when/if people try to use it rather than to delete that code altogether.

PR #10068, opened 2026-06-16
Transfer: Omit unset optional fields from describe responses

### LINKED ISSUE #9901: [Transfer] Server and Connector optional fields should not be returned in response
https://docs.aws.amazon.com/boto3/latest/reference/services/transfer/client/describe_server.html
https://docs.aws.amazon.com/transfer/latest/APIReference/API_DescribedServer.html

In practice, it appears that optional values that are not set shouldn't be included in the responses.
Fields such as LoggingRole and others should be omitted from responses

https://github.com/getmoto/moto/blob/master/moto/transfer/types.py

somaz94:
AWS Transfer omits optional fields that were never set from `DescribeServer` / `DescribeConnector` responses, but moto was returning them as `null` (the model `to_dict()` output is serialized verbatim). This adds a small recursive `prune_unset` helper applied in `Server.to_dict()` and `Connector.to_dict()` so unset scalars (e.g. `LoggingRole`, `Certificate`, `SecurityPolicyName`) and empty nested blocks (`EndpointDetails`, `IdentityProviderDetails`, `As2Config`, ...) are dropped, matching the real API.

Validation:
- `pytest tests/test_transfer/` — 18 passed (2 new tests asserting describe_server / describe_connector omit unset optional fields; all 16 existing tests pass unchanged)
- `ruff check` / `ruff format --check` — clean
- `mypy moto/transfer/` — no issues found

Resolves #9901

### DIFF
--- a/moto/transfer/types.py
+++ b/moto/transfer/types.py
@@ -6,6 +6,25 @@ from moto.core.common_models import BaseModel
 from moto.moto_api._internal import mock_random
 
 
+def prune_unset(value: Any) -> Any:
+    """Recursively drop unset fields (None / empty dict / empty list).
+
+    AWS Transfer omits optional fields that were never set from its Describe*
+    responses rather than returning them as null, so the serialized model must
+    do the same.
+    """
+    if isinstance(value, dict):
+        pruned = {}
+        for key, val in value.items():
+            cleaned = prune_unset(val)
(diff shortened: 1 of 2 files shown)


bblommers (approved):
LGTM! Thanks for this PR @somaz94

PR #10070, opened 2026-06-18
ECS: support capacityProviderStrategy in RunTask

### LINKED ISSUE #8722: [ECS] Does RunTask support capacityProviderStrategy?
Hi,

I'm trying to test a call to ecs.RunTask with a capacityProviderStrategy argument (and no launchType). The capacityProvider specified is using an autoScalingGroup. This is supposed to use autoscaling to create instances for the tasks, but I'm getting `Exception: No instances found in cluster test-cluster` when I do this. Looking at the code it seems like the runTask method doesn't use the capacityProviderStrategy argument, so I'm suspecting that this mode of operation is not supported. Is that correct?

UTKARSH698:
## Summary

`ECS.RunTask` did not accept `capacityProviderStrategy`. The parameter was silently dropped, and because no `launchType` was supplied the backend fell through to EC2 placement and raised `No instances found in cluster ...`. Fixes #8722.

## Changes

- `responses.py`: parse `capacityProviderStrategy` and pass it to the backend.
- `models.py`:
  - Accept `capacity_provider_strategy` in `run_task`.
  - Raise `InvalidParameterException` when both `launchType` and `capacityProviderStrategy` are given (matches AWS).
  - Fall back to the cluster's `defaultCapacityProviderStrategy` when neither is supplied.
  - Route `FARGATE`/`FARGATE_SPOT` providers to Fargate placement and user-defined (ASG-backed) providers to EC2 container instances.
  - Record `capacityProviderName` on the task so it is returned in the response.

## Tests

Added to `tests/test_ecs/test_ecs.py`:
- `test_run_task_with_capacity_provider_strategy_fargate`
- `test_run_task_with_capacity_provider_strategy_ec2`
- `test_run_task_with_launch_type_and_capacity_provider_strategy_fails`
- `test_run_task_uses_cluster_default_capacity_provider_strategy`

All ECS tests pass locally (the only failures are pre-existing CloudFormation tests that require the optional `openapi_spec_validator` dependency). `ruff check` is clean.

### DIFF
--- a/moto/ecs/models.py
+++ b/moto/ecs/models.py
@@ -314,6 +314,7 @@ class Task(ManagedState, BaseModel):
         tags: list[dict[str, str]] | None = None,
         networking_configuration: dict[str, Any] | None = None,
         platform_version: str | None = None,
+        capacity_provider_name: str | None = None,
     ):
         # Configure ManagedState
         # https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task-lifecycle.html
@@ -344,6 +345,7 @@ class Task(ManagedState, BaseModel):
         self.started_by = started_by
         self.tags = tags or []
         self.launch_type = launch_type
+        self.capacity_provider_name = capacity_provider_name
         self.platform_version = platform_version
         self.stopped_reason = ""
         self.resource_requirements = resource_requirements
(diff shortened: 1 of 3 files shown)


UTKARSH698:
The failing CI jobs here are unrelated to this change. The only failing test is `tests/test_awslambda/test_lambda_eventsourcemapping.py::test_get_event_source_mapping`, which fails with `ParamValidationError: Invalid length for parameter UUID, value: 1, valid min length: 36` — a botocore validation change, not anything in the ECS code.

The same test is currently failing on `master` (HEAD `865295a`) with the identical error, so it is pre-existing and independent of this PR. All ECS tests, including the four new `capacityProviderStrategy` tests, pass across Python 3.10–3.14.

bblommers (approved):
LGTM - thanks @UTKARSH698! Nice and clean implementation, and nice to see that there are tests for all possible scenarios as well.

PR #10094, opened 2026-07-04
SFN parser: convert PascalCase task parameters to boto casing (fixes startExecution.sync)

### LINKED ISSUE #10076: StepFunctions parser: states:::states:startExecution.sync fails with KeyError 'Input' (PascalCase params not converted to boto casing)
## Summary

The `arn:aws:states:::states:startExecution.sync` / `.sync:2` service integration (a parent state machine starting a child state machine) fails under the Step Functions **parser** backend with `States.Runtime` / cause `'Input'`.

Root cause: the parser passes the ASL **PascalCase** parameter keys (`Input`, `StateMachineArn`) straight to `boto3`'s `start_execution`, whose members are **lowercase** (`stateMachineArn` / `name` / `input` / `traceHeader`). botocore's serializer then raises `KeyError: 'Input'`.

This appears to be a regression introduced when the parser was ported from LocalStack (#9142): the base `StateTaskService._normalise_parameters` / `_normalise_response` — which in LocalStack convert the request/response between SFN PascalCase and the boto member casing — were reduced to `pass` in moto, while the helper methods they used to call (`_to_boto_request` / `_from_boto_response`) were kept but are now never invoked.

It goes unnoticed because the SFN→SNS / →DynamoDB / →SQS integration tests use services whose boto members are *already* PascalCase (`TopicArn`, `Message`, `TableName`), so verbatim passthrough happens to work. There is no SFN→SFN (`startExecution.sync`) integration test, and `start_execution` is the case where SFN's own API members are lowercase.

## How to reproduce

```python
import json, time
import boto3
from moto.server import ThreadedMotoServer
from moto.core.config import default_user_config

# Enable the ASL interpreter (parser backend)
default_user_config["stepfunctions"]["execute_state_machine"] = True

srv = ThreadedMotoServer(port=5000)
srv.start()
sfn = boto3.client("stepfunctions", endpoint_url="http://localhost:5000",
                   region_name="us-east-1", aws_access_key_id="x", aws_secret_access_key="x")
role = "arn:aws:iam::123456789012:role/sfn"

child = sfn.create_state_machine(
    name="child", roleArn=role,
    definition=json.dumps({"StartAt": "P", "States": {"P": {"Type": "Pass", "End": True}}}),
)["stateMachineArn"]

parent = sfn.create_state_machine(
    name="parent", roleArn=role,
    definition=json.dumps({
        "StartAt": "CallChild",
        "States": {
            "CallChild": {
                "Type": "Task",
                "Resource": "arn:aws:states:::states:startExecution.sync:2",
                "Parameters": {"Input": {"foo": "bar"}, "StateMachineArn": child},
                "End": True,
            }
        },
    }),
)["stateMachineArn"]

exe = sfn.start_execution(stateMachineArn=parent, name="r1", input="{}")["executionArn"]
for _ in range(40):
    d = sfn.describe_execution(executionArn=exe)
    if d["status"] != "RUNNING":
        break
    time.sleep(0.25)
print("STATUS:", d["status"])   # FAILED
print("ERROR:", d.get("error")) # States.Runtime
print("CAUSE:", d.get("cause")) # An error occurred while executing the state 'CallChild' ... 'Input'
srv.stop()
```

## Expected

The parent execution reaches `SUCCEEDED` (the child runs and the `sync:2` task returns its describe-execution output), as it does on real AWS Step Functions.

## Actual

```
STATUS: FAILED
ERROR: States.Runtime
CAUSE: An error occurred while executing the state 'CallChild' (entered at the event id #4). 'Input'
```

The underlying exception (visible if you log `ExecutionState._from_error`) is:

```
File ".../botocore/serialize.py", line 473, in _serialize_type_structure
    member_shape = members[member_key]
                   ~~~~~~~^^^^^^^^^^^^
KeyError: 'Input'
```

from `state_task_service_sfn.py::_eval_service_task` → `getattr(sfn_client, "start_execution")(**normalised_parameters)` with `normalised_parameters = {"Input": "...", "StateMachineArn": "..."}`.

## Pointers

- `moto/stepfunctions/parser/asl/component/state/exec/state_task/service/state_task_service.py` — base `_normalise_parameters` (and `_normalise_response`) are `pass`; the `_to_boto_request` / `_from_boto_response` helpers are present but never called.
- Compare LocalStack's equivalent base `_normalise_parameters`, which looks up the boto `input_shape` and calls `_to_boto_request(parameters, input_shape)` to convert the casing before the boto call.
- `state_task_service_sfn.py::_normalise_parameters` only JSON-stringifies `Input`; it relies on the (now-missing) base normalisation to fix key casing.

## Environment

- moto 5.1.22
- boto3 1.42.93 / botocore 1.42.93
- Python 3.12.13

koteshyelamati:
Fixes #10076

**Stacked on #10093** — the first commit here is that PR's deepcopy fix; this PR's own changes are the last two commits. Once #10093 merges, this diff reduces to `state_task_service.py` + the new integration test.

The `arn:aws:states:::states:startExecution.sync[:2]` service integration (a parent state machine starting a child state machine) failed under the Step Functions parser backend with `States.Runtime` / `KeyError: 'Input'`: the parser passed the ASL PascalCase parameter keys (`Input`, `StateMachineArn`) straight to boto3's `start_execution`, whose members are lowercase.

This is a regression from the LocalStack port (#9142): the base `_normalise_parameters` / `_normalise_response` — which convert the request/response between SFN PascalCase and the boto member casing via the operation model shapes — were reduced to `pass`, while the `_to_boto_request` / `_from_boto_response` helpers they call were kept but never invoked. It went unnoticed because the existing SNS/SQS/DynamoDB integration tests use services whose boto members are already PascalCase, so verbatim passthrough happened to work.

Changes:

- Restore the base `_normalise_parameters` / `_normalise_response` from the LocalStack implementation, using `botocore.session` to load the operation model (LocalStack uses its own `load_service`).
- Convert `datetime` values in normalised responses to epoch seconds, matching the AWS JSON protocol representation of timestamps. Without this, the `.sync` task result (the child's DescribeExecution output, which contains `startDate`/`stopDate`) cannot be JSON-serialised into the parent execution's output.
- Add an SFN-to-SFN integration test: a parent state machine runs `startExecution.sync:2` on a child and the parent reaches `SUCCEEDED` with the child's output in its task result. The test fails with the reported `KeyError: 'Input'` without the fix.

All existing parser integration tests (SNS/SQS/DynamoDB/choice/map/parallel/idempotency) pass with the restored normalisation.


bpandola (commented):
Running your integration test against a real AWS backend fails for us.  See log here: https://github.com/getmoto/moto/actions/runs/28776578462/job/85321890332#step:10:108

koteshyelamati:
Thanks for running the AWS integration test, @bpandola! I checked the CI log — could you paste the specific assertion/error from the failure? The test creates real state machines and waits for a sync execution, so I want to understand whether it's a timing issue (the test's poll interval), a permissions/region gap, or a correctness problem with the normalisation logic itself. Happy to adjust the test or the fix once I know what's actually failing.

Issue #10077, opened 2026-06-29
StepFunctions parser: second StartExecution on the same state machine fails with 'cannot pickle _thread.RLock' (deepcopy of executions)


tinovyatkin:
## Summary

Under the Step Functions **parser** backend (`stepfunctions.execute_state_machine=True`), starting the **same state machine a second time** fails with a 500 / `TypeError: cannot pickle '_thread.RLock' object`.

`StepFunctionsParserBackend.start_execution` does `state_machine_clone = copy.deepcopy(state_machine)` ([`moto/stepfunctions/parser/models.py#L175`](https://github.com/getmoto/moto/blob/master/moto/stepfunctions/parser/models.py#L175)). After the first execution, `state_machine.executions` contains an `Execution` whose `exec_worker` is a live `ExecutionWorker` (a `threading.Thread` plus an `Environment` holding a `_thread.RLock`). `deepcopy` traverses the `executions` list and tries to copy that lock, which is unpicklable.

This also makes the `arn:aws:states:::states:startExecution.sync[:2]` service integration unusable (a parent SM starting a child SM triggers a second `start_execution` on the backend), but the bug is more general — it happens for **any** second `StartExecution` on a state machine that has already run.

## How to reproduce

```python
import json, time, os
import boto3
from moto.server import ThreadedMotoServer
from moto.core.config import default_user_config

default_user_config["stepfunctions"]["execute_state_machine"] = True
os.environ["MOTO_PORT"] = "5000"
srv = ThreadedMotoServer(port=5000); srv.start()
sfn = boto3.client("stepfunctions", endpoint_url="http://localhost:5000",
                   region_name="us-east-1", aws_access_key_id="x", aws_secret_access_key="x")

arn = sfn.create_state_machine(
    name="sm", roleArn="arn:aws:iam::123456789012:role/sfn",
    definition=json.dumps({"StartAt": "P", "States": {"P": {"Type": "Pass", "End": True}}}),
)["stateMachineArn"]

e1 = sfn.start_execution(stateMachineArn=arn, name="run1", input="{}")["executionArn"]
time.sleep(1)
print("run1:", sfn.describe_execution(executionArn=e1)["status"])   # SUCCEEDED

# Second execution of the SAME state machine:
e2 = sfn.start_execution(stateMachineArn=arn, name="run2", input="{}")["executionArn"]  # 500
time.sleep(1)
print("run2:", sfn.describe_execution(executionArn=e2)["status"])
srv.stop()
```

## Expected

[...]

PR #10095, opened 2026-07-05
SFN parser: support AWS JSONPath filter dialect (&&/|| and empty filter results)

### LINKED ISSUE #10078: StepFunctions parser: jsonpath_ng backend doesn't implement AWS JSONPath filter dialect (&&/|| parse error; empty filter raises instead of [])
## Summary

The Step Functions **parser** backend evaluates ASL JSONPath through `jsonpath_ng.ext`, which does not implement AWS Step Functions' JSONPath dialect. Two gaps make the classic service integrations unusable for real-world (CDK-generated) state machines:

1. **`&&` / `||` filter conjunctions.** AWS SFN filter expressions use `&&` / `||` (e.g. `$.items[?(@.a == true && @.b == true)]`). `jsonpath_ng` only accepts a single `&` / `|` and raises `JsonPathParserError: Parse error ... near token &` on the doubled form. This shape is emitted directly by `aws-cdk-lib`'s `StateMachine` (e.g. inside `States.JsonToString(...)`), so it appears in real deployed ASL.

2. **Empty filter result raises instead of `[]`.** On AWS, a `[?()]` filter that matches nothing yields `[]`. moto's `extract_json` raises `NoSuchJsonPathError` for it (it returns `[]` only for slice/wildcard paths, not filters), surfacing as `States.Runtime`.

Both block the `arn:aws:states:::states:startExecution.sync` path for any non-trivial scenario, because the always-present input-preparation states filter file lists with exactly these constructs.

## Reproduce

`&&` parse failure (end-to-end):

```python
import json, time, os
import boto3
from moto.server import ThreadedMotoServer
from moto.core.config import default_user_config

default_user_config["stepfunctions"]["execute_state_machine"] = True
os.environ["MOTO_PORT"] = "5000"
srv = ThreadedMotoServer(port=5000); srv.start()
sfn = boto3.client("stepfunctions", endpoint_url="http://localhost:5000",
                   region_name="us-east-1", aws_access_key_id="x", aws_secret_access_key="x")
definition = {
    "StartAt": "Filter",
    "States": {"Filter": {
        "Type": "Pass",
        "Parameters": {"selected.$": "$.items[?(@.keep == true && @.ready == true)]"},
        "End": True,
    }},
}
arn = sfn.create_state_machine(name="f", roleArn="arn:aws:iam::123456789012:role/sfn",
                               definition=json.dumps(definition))["stateMachineArn"]
exe = sfn.start_execution(stateMachineArn=arn, name="r1",
                          input=json.dumps({"items": [{"keep": True, "ready": True}]}))["executionArn"]
for _ in range(40):
    d = sfn.describe_execution(executionArn=exe)
    if d["status"] != "RUNNING": break
    time.sleep(0.25)
print(d["status"], d.get("cause"))   # FAILED  JsonPathParserError(Parse error at 1:26 near token & (&))
srv.stop()
```

Empty-filter facet (isolated; single `&` so it parses, matches nothing):

```python
from moto.stepfunctions.parser.asl.utils.json_path import extract_json
extract_json("$.items[?(@.k == 'NONE')]", {"items": [{"k": "a"}]})
# -> raises NoSuchJsonPathError; AWS Step Functions yields []
```

## Suggested fix

Rather than special-case each divergence, consider swapping the JSONPath engine for one that implements the AWS dialect. [`python-jsonpath`](https://github.com/jg-rp/python-jsonpath) handles both cases out of the box:

```python
import jsonpath  # python-jsonpath
data = {"items": [{"keep": True, "ready": True}, {"keep": True, "ready": False}]}
jsonpath.findall("$.items[?(@.keep == true && @.ready == true)]", data)  # -> [{'keep': True, 'ready': True}]
jsonpath.findall("$.items[?(@.k == 'NONE')]", data)                      # -> []  (no error)
```

`extract_json`'s current AWS-specific post-processing (singleton-array unpack, the #7825 context-Index special case, scalar-vs-list) maps cleanly onto `python-jsonpath`'s `JSONPathEnvironment.compile(path).singular_query()`: a singular query returns the single value (or "not found" for a missing definite path), a non-singular query returns the list of matches. That replaces the `isinstance(match.path, Index)` heuristics and `_is_singleton_array_access` regex with the library's own notion of a singular path.

(For now we monkeypatch `json_path.extract_json` onto `python-jsonpath` in our offline test harness; we'd happily drop it once moto evaluates AWS JSONPath natively.)

## Environment

- moto 5.1.22 (parser backend, `execute_state_machine=True`)
- jsonpath-ng 1.8.0 (latest; does not implement AWS `&&`/`||`)
- Python 3.12

koteshyelamati:
Fixes #10078

The Step Functions parser backend evaluates ASL JSONPath through `jsonpath_ng.ext`, which does not implement two features of AWS Step Functions' JSONPath dialect. Both appear in real-world (CDK-generated) state machines and surface as `States.Runtime` errors:

1. **`&&` / `||` filter conjunctions.** AWS SFN filter expressions use `&&`/`||` (e.g. `$.items[?(@.a == true && @.b == true)]`); jsonpath_ng only accepts single `&` and raises `JsonPathParserError` on the doubled form, and its filter grammar has no disjunction at all.
2. **Empty filter results raise instead of yielding `[]`.** On AWS, a filter that matches nothing produces `[]`; `extract_json` raised `NoSuchJsonPathError` (it returned `[]` only for slice/wildcard paths).

Rather than swapping the JSONPath engine (a larger dependency decision), this implements the dialect on top of jsonpath_ng inside `extract_json`:

- `&&` is translated to jsonpath_ng's `&`, skipping string literals.
- `||` is handled by expanding the filter into one path per disjunct (respecting AWS's precedence: `&&` binds tighter than `||`), evaluating each, and merging matches back into document order with duplicates removed. Nothing changes for paths without `||`.
- Filters that match nothing now return `[]`, consistent with the existing slice/wildcard behaviour.

Operators inside string literals (e.g. `[?(@.v == 'x && y')]`) are preserved by quote-aware scanning in both transformations.

Added unit tests covering conjunctions, disjunctions (including precedence, chaining and document order), quoted literals, empty filter results, and the pre-existing behaviours (definite paths still raise `NoSuchJsonPathError` when missing, wildcard/index access unchanged). All existing parser test suites pass.

Together with #10093 and #10094 this completes the set of three issues (#10076, #10077, #10078) blocking the `states:startExecution.sync` path for CDK-generated state machines.


tinovyatkin:
I would not recommend to merge this as is - patching just these two cases without extending testing corpus is future error prone. I believe swapping whole jsonpath engine is more future proof solution here.

PR #10106, opened 2026-07-14
Complete the sfn versioning API

andresliszt:
Hey there: Added these methods to stepfunctions

- `list_state_machine_versions` returns only published versions, ordered newest-first.
- `delete_state_machine_version` is idempotent when the version does not exist, but raises `ConflictException` if any alias references the version.
- `publish_state_machine_version` returns `stateMachineVersionArn` and `creationDate`.
- `list_state_machine_versions` and `publish_state_machine_version` raise `StateMachineDoesNotExist` for unknown state machine ARNs.
- `create_state_machine` rejects `versionDescription` unless `publish=True`, raising `ValidationException`.

Also simplifies a little bit the code creating a helper method `_get_state_machine(self, arn: str) -> StateMachine` which was repeated a lot across the file.

Simplfies the tests using  `_get_default_role `and `simple_definition`

### DIFF
--- a/moto/stepfunctions/exceptions.py
+++ b/moto/stepfunctions/exceptions.py
@@ -31,6 +31,11 @@ class StateMachineDoesNotExist(AWSError):
     STATUS = 400
 
 
+class ConflictException(AWSError):
+    TYPE = "ConflictException"
+    STATUS = 400
+
+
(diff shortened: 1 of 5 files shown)


andresliszt:
Hmm lint is failing, I will review it!

PR #10110, opened 2026-07-15
Fix core decorator typing for mypy 2.3.0 (isclass now narrows to type[object])

### LINKED ISSUE #10109: Newest mypy causing build issues
mypy changes the isclass annotation causing the build issues

Build Error:
`
Running MyPy...
mypy --install-types --non-interactive
moto/core/models.py:68:41: error: Argument 1 to "_decorate_class" of "MockAWS" has incompatible type "type[object]"; expected "Callable[P, T]"  [arg-type]
Found 2 errors in 1 file (checked 1146 source files)
moto/core/models.py:192:16: error: Incompatible return value type (got "type[object]", expected "Callable[P, T]")  [return-value]
make: *** [Makefile:25: lint] Error 1
`

isclass annotation changed from `TypeIs[type[Any]]` to `TypeIs[type[object]]` in 2.3.0.

[mypy change](https://github.com/python/mypy/blame/82e8498c19494edebee08dd615470443ec53fffc/mypy/typeshed/stdlib/inspect.pyi#L220)

ApogeeAC:
mypy's latest release changed isclass annotation from `TypeIs[type[Any]]` to `TypeIs[type[object]]`

_decorate_class now has an accurate signature: it takes a class and returns a class (type[T] -> type[T]), where previously both were typed as Callable.
Removed the assert inspect.isclass(klass)  # Keep mypy happy line. It was originally added as a workaround for older mypy, and under mypy 2.3.0 it actively re-triggers the problem by re-narrowing the type.
Added a single cast() where __call__ returns the decorated class, telling mypy what's true at runtime but inexpressible in its type system: a class is also a callable that produces its instances.

Suggested change would fix #10109

### DIFF
--- a/moto/core/models.py
+++ b/moto/core/models.py
@@ -12,6 +12,7 @@ from typing import (
     TYPE_CHECKING,
     Any,
     TypeVar,
+    cast,
 )
 from unittest.mock import patch
 
@@ -65,7 +66,7 @@ class MockAWS(AbstractContextManager["MockAWS"]):
         self, func: "Callable[P, T]", reset: bool = True, remove_data: bool = True
     ) -> "Callable[P, T]":
         if inspect.isclass(func):
-            return self._decorate_class(func)
+            return cast("Callable[P, T]", self._decorate_class(func))
         return self._decorate_callable(func, reset, remove_data)
 
     def __enter__(self) -> "MockAWS":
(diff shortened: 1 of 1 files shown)


ApogeeAC:
@bblommers / @bpandola
Happy to adjust the approach here if you'd prefer something different — this was the minimal fix I found, but I'm not attached to the specifics. Feel free to push changes directly or leave suggestions and I'll turn them around 👍

ApogeeAC:
Will mark ready for review once the checks are completed!

bblommers (approved):
LGTM! Thanks for the fix @ApogeeAC 🙏

PR #10115, opened 2026-07-16
Fix describe_instances group-id filter for instances in multiple security groups

### LINKED ISSUE #10114: describe_instances group-id filter only matches an instance's first security group

### Reproduction

```python
import boto3
from moto import mock_aws

@mock_aws
def run():
    def q(gid):
run()
```

### Cause

winklemad:
Fixes #10114.

### Problem
`describe_instances` filtered by `group-id` / `instance.group-id` / `instance.group-name` only matches an instance's **first** security group, so an instance in several groups is wrongly excluded when filtered by any of its other groups (repro in #10114). Real AWS matches via any attached group.

### Cause
`get_object_value` (`moto/ec2/utils.py`), when a filter attribute traverses a list (`security_groups.id`), returned the first element's value instead of all of them.

### Fix
Collect the value from every list element. `instance_value_in_filter_values` already intersects lists, so the requested group now matches via any of the instance's security groups; `passes_igw_filter_dict` never traverses a list, so it is unaffected.

### Tests
Added `test_get_instances_filtering_by_group_id_multiple_security_groups` (matches via first + second SG + group-name; a non-member group returns nothing). Fails before, passes after; `test_instances.py` filter tests stay green (25 passed).

### DIFF
--- a/moto/ec2/utils.py
+++ b/moto/ec2/utils.py
@@ -348,10 +348,12 @@ def get_object_value(obj: Any, attr: str) -> Any:
         elif isinstance(val, dict):
             val = val[key]
         elif isinstance(val, list):
-            for item in val:
-                item_val = get_object_value(item, key)
-                if item_val:
-                    return item_val
+            # Collect the value from every element (e.g. all of an instance's
+            # security groups), not just the first, so a filter matches when the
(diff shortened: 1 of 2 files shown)


bpandola:
Superseded by #10170

PR #10125, opened 2026-07-20
RDS: fix test using invalid RoleArn length

ApogeeAC:
Botocore 1.43.51 tightened RoleArn validation on AddRoleToDBCluster to require a minimum length of 20 characters, so the test's placeholder value "a-different-role" (16 characters) now fails client-side before it can exercise the intended DBClusterRoleAlreadyExists error.

- Changelog: https://github.com/boto/botocore/blob/1.43.51/CHANGELOG.rst
- Release diff: https://github.com/boto/botocore/compare/1.43.50...1.43.51 

Failure example in build:
image


bblommers (approved):
🙏

PR #10131, opened 2026-07-26
CloudTrail: Fix start_logging/stop_logging to accept trail ARN

### LINKED ISSUE #10107: CloudTrail start_logging/stop_logging raise KeyError when called with a trail ARN
# CloudTrail `start_logging`/`stop_logging` raise `KeyError` when called with a trail ARN

## The bug

`CloudTrailBackend.start_logging` (and `stop_logging`) look the trail up with a bare dict access keyed by **name**, but AWS's `StartLogging`/`StopLogging` accept **either the trail name or its ARN** in the `Name` parameter. When called with an ARN, moto raises an unhandled `KeyError`, which surfaces to the client as an HTTP 500.

This breaks the Terraform `aws_cloudtrail` resource against moto: the provider uses the trail ARN as the resource ID and calls `StartLogging` with that ARN, so `terraform apply` fails. Because moto returns a 500, the AWS SDK/provider treats it as a transient error and **retries until timeout**, so it appears to hang.

## To Reproduce

```python
import json
import boto3
from moto import mock_aws


@mock_aws
def test_start_logging_by_arn():
    s3 = boto3.client("s3", region_name="us-east-1")
    s3.create_bucket(Bucket="trail-bucket")
    s3.put_bucket_policy(
        Bucket="trail-bucket",
        Policy=json.dumps(
            {
                "Version": "2012-10-17",
                "Statement": [
                    {
                        "Sid": "AWSCloudTrailWrite",
                        "Effect": "Allow",
                        "Principal": {"Service": "cloudtrail.amazonaws.com"},
                        "Action": "s3:PutObject",
                        "Resource": "arn:aws:s3:::trail-bucket/AWSLogs/123456789012/*",
                    }
                ],
            }
        ),
    )

    ct = boto3.client("cloudtrail", region_name="us-east-1")
    arn = ct.create_trail(Name="my-trail", S3BucketName="trail-bucket")["TrailARN"]

    ct.start_logging(Name="my-trail")  # OK — resolved by name
    ct.start_logging(Name=arn)         # raises KeyError (AWS accepts the ARN here)


test_start_logging_by_arn()
```

Output:

```
KeyError: 'arn:aws:cloudtrail:us-east-1:123456789012:trail/my-trail'
```

## Expected behavior

`start_logging`/`stop_logging` should resolve the trail by **name or ARN** (matching real AWS, and consistent with `get_trail_status`), and raise `TrailNotFoundException` — not a `KeyError`/500 — if the trail genuinely doesn't exist.

## Root cause

In `moto/cloudtrail/models.py`, trails are stored keyed by name:

```python
self.trails[name] = trail        # create_trail
```

but `start_logging`/`stop_logging` do a raw dict lookup with no ARN fallback:

```python
def start_logging(self, name: str) -> None:
    trail = self.trails[name]    # KeyError when `name` is an ARN
    trail.start_logging()

def stop_logging(self, name: str) -> None:
    trail = self.trails[name]
    trail.stop_logging()
```

By contrast, `get_trail_status` (and the `get_trail` helper) already resolve both forms:

```python
trail = next(
    (t for t in all_trails if t.trail_name == name or t.arn == name), None
)
```

So the ARN-tolerant path already exists in the backend — it just isn't used by `start_logging`/`stop_logging`. (`get_trail_status` was made ARN-tolerant in #4544 / #4547, which didn't touch these two methods.) The existing tests only ever call `start_logging(Name=<plain name>)`, which is why this went uncaught.

## Suggested fix

Route `start_logging`/`stop_logging` through the existing ARN-tolerant `get_trail` helper (or replicate its `trail_name == name or arn == name` match), and raise `TrailNotFoundException` on a miss.

## Environment

- moto **5.2.2** (also present on `master` at time of writing)
- Reproduced both directly via boto3 (snippet above) and via the Terraform `aws_cloudtrail` resource (`enable_logging = true`) against `moto server`.

Mide69:
CloudTrail: Fix start_logging/stop_logging to accept trail ARN

Closes [#10107](https://github.com/getmoto/moto/issues/10107)

start_logging/stop_logging did a raw self.trails[name] dict lookup, which raises KeyError when called with a trail ARN; AWS accepts both name and ARN for these operations, and get_trail_status already handles this correctly. Routes both methods through the existing ARN-tolerant get_trail() helper instead, matching the pattern already used by update_trail, put_event_selectors, etc. Unknown trails now correctly raise TrailNotFoundException.

Added tests covering start/stop by ARN and the not-found case for both methods. The full test_cloudtrail suite (45 tests) passes, and the ruff check/ruff format --check/mypy all clean.

### DIFF
--- a/moto/cloudtrail/models.py
+++ b/moto/cloudtrail/models.py
@@ -335,11 +335,11 @@ class CloudTrailBackend(BaseBackend):
         return self.describe_trails(include_shadow_trails=True)
 
     def start_logging(self, name: str) -> None:
-        trail = self.trails[name]
+        trail = self.get_trail(name_or_arn=name)
         trail.start_logging()
 
     def stop_logging(self, name: str) -> None:
-        trail = self.trails[name]
+        trail = self.get_trail(name_or_arn=name)
         trail.stop_logging()
 
     def delete_trail(self, name: str) -> None:
--- a/tests/test_cloudtrail/test_cloudtrail.py
+++ b/tests/test_cloudtrail/test_cloudtrail.py
@@ -309,6 +309,31 @@ def test_get_trail_status_after_starting_and_stopping():
 
 
 @mock_aws
+def test_start_and_stop_logging_by_arn():
+    client = boto3.client("cloudtrail", region_name="eu-west-3")
+    _, resp, trail_name = create_trail_simple(region_name="eu-west-3")
+    trail_arn = resp["TrailARN"]
+
+    client.start_logging(Name=trail_arn)
+    status = client.get_trail_status(Name=trail_name)
+    assert status["IsLogging"] is True
+
+    client.stop_logging(Name=trail_arn)
+    status = client.get_trail_status(Name=trail_name)
+    assert status["IsLogging"] is False
+
+
+@pytest.mark.parametrize("method", ["start_logging", "stop_logging"])
+@mock_aws
+def test_start_and_stop_logging_unknown_trail(method):
+    client = boto3.client("cloudtrail", region_name="eu-west-3")
+    with pytest.raises(ClientError) as exc:
+        getattr(client, method)(Name="does-not-exist")
+    err = exc.value.response["Error"]
+    assert err["Code"] == "TrailNotFoundException"
+
+
+@mock_aws
 def test_get_trail_status_multi_region_not_from_the_home_region():
     # CloudTrail client
     client_us_east_1 = boto3.client("cloudtrail", region_name="us-east-1")


ApogeeAC:
Duplicates #10119

Mide69:
Just noticed [#10119](https://github.com/getmoto/moto/pull/10119) was opened before this one (Jul 18 vs my Jul 26) and implements essentially the same fix. Closing this in favor of that one to avoid asking a maintainer to pick between duplicates -thanks for flagging it, @ApogeeAC.

PR #10145, opened 2026-07-31
EC2: match all security groups when filtering describe_instances

### LINKED ISSUE #10114: describe_instances group-id filter only matches an instance's first security group

Sanjays2402:
Closes #10114

`describe_instances` filtered by `group-id`, `instance.group-id` or `instance.group-name` only matched an instance's *first* security group, so an instance in several groups was wrongly excluded when filtered by any of the others.

`get_object_value()` returned as soon as the first list element produced a value, so for a list-valued attribute like `security_groups` only that element's value reached `instance_value_in_filter_values()`. It now collects the value of every element and returns the full list (or `None` when empty), which the filter comparison already handles via its list branch.

`test_get_instances_filtering_by_secondary_instance_group` covers the three affected filters plus the first group; it fails on master and passes with the fix.

This change was prepared with AI assistance; the regression test was run locally and fails without the fix.

### DIFF
--- a/moto/ec2/utils.py
+++ b/moto/ec2/utils.py
@@ -348,10 +348,12 @@ def get_object_value(obj: Any, attr: str) -> Any:
         elif isinstance(val, dict):
             val = val[key]
         elif isinstance(val, list):
+            item_values = []
             for item in val:
                 item_val = get_object_value(item, key)
                 if item_val:
-                    return item_val
+                    item_values.append(item_val)
+            return item_values or None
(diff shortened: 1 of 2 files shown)


ApogeeAC:
Duplicates #10115

bpandola:
Superseded by #10170

PR #10165, opened 2026-08-04
fix: group-id/group-name filters only matched an instance's first security group

### LINKED ISSUE #10114: describe_instances group-id filter only matches an instance's first security group

charuishere:
get_object_value() in moto/ec2/utils.py returns early on the first
match when walking a list attribute (e.g. an instance's security
groups), so filtering describe_instances by group-id or group-name
only ever matched an instance's first security group. Instances in
multiple groups were silently excluded when filtered by any group
other than the first.

Fixed it to collect all matches into a list instead of returning
early - the calling code (instance_value_in_filter_values) already
expected a list back.

Added tests for both group-id and group-name filters with an
instance in multiple security groups. Confirmed both fail against
the old code (IndexError) and pass with the fix.

Fixes #10114

### DIFF
--- a/moto/ec2/utils.py
+++ b/moto/ec2/utils.py
@@ -348,10 +348,12 @@ def get_object_value(obj: Any, attr: str) -> Any:
         elif isinstance(val, dict):
             val = val[key]
         elif isinstance(val, list):
+            values = []
             for item in val:
                 item_val = get_object_value(item, key)
                 if item_val:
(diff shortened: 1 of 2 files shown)


bpandola:
Superseded by #10170

PR #10175, opened 2026-08-11
fix(dynamodb): validate Select and honor Scan COUNT

### LINKED ISSUE #10174: [DynamoDB] Query and Scan do not enforce Select semantics
Moto does not fully enforce DynamoDB's `Select` semantics for `Query` and `Scan`. This allows invalid requests to succeed and makes `Scan` with `Select="COUNT"` include `Items`.

This is present in the latest Moto release, `5.2.2`, and on the current `master` branch at 7dc0838503a3da50ac83980dc82c07572bb5212a.

The following code creates a table with a `KEYS_ONLY` GSI and demonstrates both behaviors:

```python
import boto3
from moto import mock_aws


@mock_aws
def reproduce():
    client = boto3.client("dynamodb", region_name="us-east-1")

    client.create_table(
        TableName="select-repro",
        AttributeDefinitions=[
            {"AttributeName": "pk", "AttributeType": "S"},
            {"AttributeName": "gsi_pk", "AttributeType": "S"},
        ],
        KeySchema=[{"AttributeName": "pk", "KeyType": "HASH"}],
        GlobalSecondaryIndexes=[
            {
                "IndexName": "test_gsi",
                "KeySchema": [{"AttributeName": "gsi_pk", "KeyType": "HASH"}],
                "Projection": {"ProjectionType": "KEYS_ONLY"},
            }
        ],
    )

    try:
        client.query(
            TableName="select-repro",
            IndexName="test_gsi",
            Select="ALL_ATTRIBUTES",
            KeyConditionExpression="gsi_pk = :value",
            ExpressionAttributeValues={":value": {"S": "group-1"}},
        )
    except ClientError as exc:
        print(exc.response["Error"]["Code"])
    else:
        print("Query accepted")

    response = client.scan(
        TableName="select-repro",
        Select="COUNT",
    )
    print("Items" in response)


reproduce()
```

**Expected Result**

The query should raise a `ValidationException` because `ALL_ATTRIBUTES` is not supported on a GSI whose projection type is not `ALL`.

The scan response should contain `Count` and `ScannedCount`, but not `Items`.

**Actual**

Moto accepts the query and prints:

```text
Query accepted
```

Moto also includes `Items` in the scan response, so the second value printed is:

```text
True
```

cluebbehusen:
DynamoDB enforces validation rules for `Select` on `Query` and `Scan`, but Moto silently accepts invalid combinations and includes `Items` in `Scan` responses when `Select="COUNT"`.

This PR adds 19 AWS-verified test cases covering GSI projection restrictions, invalid `ProjectionExpression` and `AttributesToGet` combinations, case-sensitive `Select` values, and `COUNT` response handling. Each case was confirmed to fail against Moto and pass against DynamoDB before the implementation was updated. The shared validation and response-handling changes make those same cases pass against Moto.

Fixes #10174.

### DIFF
--- a/moto/dynamodb/responses.py
+++ b/moto/dynamodb/responses.py
@@ -36,6 +36,12 @@ from .exceptions import (
 )
 
 TRANSACTION_MAX_ITEMS = 25
+SELECT_VALUES = [
+    "SPECIFIC_ATTRIBUTES",
+    "COUNT",
+    "ALL_ATTRIBUTES",
+    "ALL_PROJECTED_ATTRIBUTES",
+]
 
 
 def include_consumed_capacity(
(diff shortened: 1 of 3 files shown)


bpandola (approved):
@cluebbehusen I appreciate the high quality effort here, especially all of the AWS verified tests.  Thanks!

Issue #10167, opened 2026-08-06
resourcegroupstaggingapi get_resources() returns duplicate entries in server mode (not reproducible with @mock_aws)


Jbat1Jumper:
## Environment
- moto: 5.2.2 (installed via `moto[server]`)
- Python: 3.12.3
- boto3: 1.40.49
- Reproduced with `moto_server` (HTTP server mode). NOT reproducible with the `@mock_aws` decorator — see below.

## How to reproduce

Against a running `moto_server` (default settings, `MOTO_ACCOUNT_ID` unset or set, doesn't matter):

```python
import boto3

ENDPOINT = "http://localhost:4566"
REGION = "us-east-1"

sns = boto3.client("sns", endpoint_url=ENDPOINT, region_name=REGION)
rgt = boto3.client("resourcegroupstaggingapi", endpoint_url=ENDPOINT, region_name=REGION)

topic_arn = sns.create_topic(Name="repro-topic")["TopicArn"]
sns.tag_resource(ResourceArn=topic_arn, Tags=[{"Key": "instance", "Value": "test-instance"}])

result = rgt.get_resources(
    ResourceTypeFilters=["sns"],
    TagFilters=[{"Key": "instance", "Values": ["test-instance"]}],
)
for entry in result["ResourceTagMappingList"]:
    print(entry["ResourceARN"])
print(len(result["ResourceTagMappingList"]), "entries")
```

Actual output (server mode):
```
arn:aws:sns:us-east-1:000000000000:repro-topic
arn:aws:sns:us-east-1:000000000000:repro-topic
2 entries
```

Expected: 1 entry — the topic was only created/tagged once.

Notably, this does NOT reproduce with the decorator

The same create → tag → get_resources sequence, run in-process with @mock_aws instead of against moto_server, returns exactly one clean entry:

```python
import boto3
from moto import mock_aws

with mock_aws():
    sns = boto3.client("sns", region_name="us-east-1")
    topic_arn = sns.create_topic(Name="repro-topic")["TopicArn"]
    sns.tag_resource(ResourceArn=topic_arn, Tags=[{"Key": "instance", "Value": "test-instance"}])

    rgt = boto3.client("resourcegroupstaggingapi", region_name="us-east-1")
    result = rgt.get_resources(
        ResourceTypeFilters=["sns"],
        TagFilters=[{"Key": "instance", "Values": ["test-instance"]}],
    )
    print(len(result["ResourceTagMappingList"]), "entries")  # -> 1
```

[...]

PR #10179, opened 2026-08-11
Core: guard against duplicate references in `BackendDict._instances`

Jbat1Jumper:
Adds a regression test for #10167. Every service backend registers itself in a global list (BackendDict._instances) the first time each AWS account uses it, but nothing checks whether it's already there. Once a second account has touched a service within a long-lived moto_server process, resourcegroupstaggingapi.get_resources() returns every other account's resources twice. This only shows up in server mode, since @mock_aws resets that list before and after every test.


bpandola (approved):
Thanks @Jbat1Jumper

Issue #10176, opened 2026-08-11
DynamoDB: Incorrect item size calculation for Number attributes


hannes-ucsc:
*(Posted by Claude Code)*

## Description

Moto's DynamoDB `DynamoType.size()` method computes the size of a Number attribute as `len(str(value))` — the length of the decimal string representation. Real DynamoDB uses a more compact encoding where the size is approximately `ceil(significant_digits / 2) + 1` bytes, as [documented by AWS](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/CapacityUnitCalculations.html).

For example, a Unix timestamp like `1786461547` (10 digits):
- Moto computes: `len("1786461547")` = **10 bytes**
- Real DynamoDB uses: `ceil(10 / 2) + 1` = **6 bytes**

This means moto's 400 KB item size check is more restrictive than real DynamoDB's. Items that DynamoDB accepts can be rejected by moto with `ValidationException: Item size has exceeded the maximum allowed size`.

## Steps to reproduce

```python
import boto3
from moto import mock_aws

@mock_aws
def test():
    client = boto3.client("dynamodb", region_name="us-east-1")
    client.create_table(
        TableName="test",
        KeySchema=[{"AttributeName": "pk", "KeyType": "HASH"}],
        AttributeDefinitions=[{"AttributeName": "pk", "AttributeType": "S"}],
        BillingMode="PAY_PER_REQUEST",
    )

    # This is accepted by real DynamoDB but rejected by moto
    client.put_item(
        TableName="test",
        Item={
            "pk": {"S": "x" * 40},
            "data": {"B": b"x" * 409530},
            "ttl": {"N": "1786461547"},
        },
    )

test()
```

## Root cause

In `moto/dynamodb/models/dynamo_type.py`, `DynamoType.size()`:

```python
def size(self) -> int:
    if self.is_number():
        value_size = len(str(self.value))  # <-- should be ceil(digits / 2) + 1
```

## Expected behavior

Number attribute size should be computed as `ceil(significant_digits / 2) + 1` to match DynamoDB's actual encoding, per the AWS documentation on [item size calculations](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/CapacityUnitCalculations.html).

PR #10182, opened 2026-08-12
dynamodb: fix Number attribute size calculation to use significant digits

ylemiesa57:
Found this one while poking around moto's DynamoDB code, seemed like a good self-contained bug to learn the item-size logic a bit better.

Issue: #10176

## What I changed

`DynamoType.size()` was computing a Number attribute's size as `len(str(value))`, basically just the length of the decimal string. Real DynamoDB actually uses a more compact encoding based on significant digits, roughly 1 byte per two significant digits plus 1 byte, with leading and trailing zeros trimmed first. This is documented here: https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/CapacityUnitCalculations.html

So moto was overcounting the size of most numbers, which made its 400KB item-size check stricter than the real thing and could reject items that real DynamoDB would accept. The issue's example was a 10-digit timestamp like `1786461547`, sized as 10 bytes by moto instead of the documented 6.

I added a small `_number_size()` helper that strips the sign, drops the decimal point, trims leading/trailing zeros, and applies the "1 byte per 2 significant digits + 1 byte" formula, and swapped it in for the old `len(str(value))` call.

## Testing

- Added `tests/test_dynamodb/models/test_dynamo_type_size.py` with parametrized cases checked against AWS's own worked examples (27 -> 2 bytes, -27 -> 3 bytes, 461 -> 3 bytes, zero-trimming cases, the 38-significant-digit max), plus a direct regression test for the timestamp from the issue and a NumberSet sum-of-members check. 12 new tests, all passing.
- Ran the full `tests/test_dynamodb/` suite (718 tests) locally, all still passing, including the existing `test_item_size_is_under_400KB` test.
- `ruff check` / `ruff format --check` on the touched files come back clean (double-checked the pre-existing lint warnings elsewhere in `dynamo_type.py` are unrelated to this change, they're on upstream master too).
- `mypy` on the touched file: no issues.

One honest caveat: I didn't try to reproduce the exact `put_item` call from the issue end-to-end, since moto intentionally sets its internal size cap a bit below the real 400KB/409600-byte limit as a safety margin (there's a comment in `LimitedSizeDict` about this), so the issue's specific byte counts don't line up perfectly with moto's cap either way. The actual bug, the wrong Number size formula, is what I focused on fixing and testing directly.

Let me know if I should tweak anything, still getting familiar with this codebase.

### DIFF
--- a/moto/dynamodb/models/dynamo_type.py
+++ b/moto/dynamodb/models/dynamo_type.py
@@ -22,6 +22,24 @@ deserializer = TypeDeserializer()
 serializer = TypeSerializer()
 
 
+def _number_size(raw_value: str) -> int:
+    """
+    Approximate the on-disk size of a DynamoDB Number attribute, matching
+    AWS's documented item-size rules:
+    https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/CapacityUnitCalculations.html
+
+    "The size of a number is approximately (1 byte per two significant
+    digits) + (1 byte)." Leading and trailing zeroes are trimmed before
+    counting significant digits, 0 itself has no significant digits, and
+    negative numbers use an extra byte for the sign.
+    """
+    is_negative = raw_value.startswith("-")
+    digits_only = raw_value.lstrip("-").replace(".", "")
+    significant_digits = digits_only.strip("0")
+    digit_count = len(significant_digits)
+    return -(-digit_count // 2) + (2 if is_negative else 1)
+
+
 class DDBType:
     """
     Official documentation at https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_AttributeValue.html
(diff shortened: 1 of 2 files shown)


moto/dynamodb/models/dynamo_type.py
```diff
@@ -214,7 +232,7 @@ def child_attr(self, key: int | str | None) -> DynamoType | None:
 
     def size(self) -> int:
         if self.is_number():
-            value_size = len(str(self.value))
+            value_size = _number_size(str(self.value))
```
hannes-ucsc, line 235:
```suggestion
            value_size = _number_size(self.value)
```

Consider pushing the string conversion into the function. The fact that numbers are stored as strings is a DynamoDB implementation detail that the *function* should emulate, not the *caller*.

hannes-ucsc (changes requested):

PR #10187, opened 2026-08-14
fix(dynamodb): BETWEEN filter incorrectly excludes attribute value 0

Sohel2309:
## Bug

DynamoDB `BETWEEN` filtering incorrectly excludes items whose attribute
value is exactly `0`, even when `0` is within the specified range.

## Root cause

`FuncBetween.expr()` checked the tested attribute using bare Python
truthiness (`and attr`). Since `Decimal("0")` is falsy, an attribute
whose value is zero was incorrectly treated as having no value.

The `start` and `end` operands already use zero-safe checks for this
exact case.

## Fix

Apply the same zero-safe value check to the attribute being tested.

## Testing

- Added regression coverage for attribute value `0`
- Confirmed the regression test fails on the unfixed code
- Confirmed the targeted tests pass after the fix
- `3/3` relevant tests pass
- `git diff --check` passes

### DIFF
--- a/moto/dynamodb/comparisons.py
+++ b/moto/dynamodb/comparisons.py
@@ -1229,16 +1229,17 @@ class FuncBetween(Func):
         start = self.start.expr(item)
         attr = self.attr.expr(item)
         end = self.end.expr(item)
-        # Need to verify whether start has a valid value
-        # Can't just check  'if start', because start could be 0, which is a valid number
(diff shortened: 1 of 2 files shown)


bpandola:
The issue here was indirectly addressed in #10272, but I'll merge this PR for the additional test coverage.

bpandola (approved):

PR #10210, opened 2026-08-27
Admin: migrate `cfnresponse.py` to package resource

bpandola:
Package `cfnresponse.py` as a resource text file, avoiding the (unnecessary) Python import and the need to modify the original AWS code to suit Moto's linting rules.

This PR (combined with the recently merged #10208) also allows Moto to remove the `packages` subdirectory, which had to be explicitly ignored in some helper scripts.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,6 +1,7 @@
 include README.md LICENSE AUTHORS.md
 include requirements.txt requirements-dev.txt
 include moto/py.typed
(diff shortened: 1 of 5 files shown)


bblommers (approved):
Makes sense! We lose the automatic linting on this file of course, but considering it is never going to change, that is not a big deal - and the fact that we can now get rid of the `packages` module is well worth it.

Thanks for the clean up @bpandola

PR #10214, opened 2026-08-31
Add S3 GetBucketPolicyStatus support

OmniNomadLLC:
## What

Implements `GetBucketPolicyStatus` for S3.

## Why

Currently the call falls through to a generic empty response: `get_bucket_policy_status` returns `PolicyStatus: {}` regardless of whether the bucket has a policy, and never raises `NoSuchBucketPolicy`. We ran into this while testing a security-audit tool against moto (a tool that checks whether buckets are public), where the empty response is indistinguishable from a private bucket.

## How

- `NoSuchBucketPolicy` when the bucket has no policy (matches AWS).
- `IsPublic` is computed from the stored policy: an Allow statement with a wildcard principal (`"*"`, `{"AWS": "*"}` or a list containing `"*"`) and no `Condition` block makes the policy public. Statements carrying any `Condition` are treated as non-public, which matches how AWS treats access-restricting condition keys; the full AWS evaluation of which condition keys still count as public is intentionally out of scope and documented as such in the helper.
- Dispatch and `ACTION_MAP` entries follow the existing `GetBucketPolicy`/`GetPublicAccessBlock` patterns; the response uses the `ActionResult` serializer.

## Tests

Four new tests in `tests/test_s3/test_s3_bucket_policy.py`: no policy raises `NoSuchBucketPolicy`; wildcard-principal policy reports `IsPublic: True`; scoped principal reports `False`; wildcard with a `Condition` reports `False`. Full `-k "policy or public_access"` subset passes (25 tests), `ruff check` and `ruff format` clean.

### DIFF
--- a/moto/s3/responses.py
+++ b/moto/s3/responses.py
@@ -89,6 +89,7 @@ from .models import (
 from .utils import (
     ARCHIVE_STORAGE_CLASSES,
     bucket_name_from_url,
+    bucket_policy_is_public,
     compute_checksum,
     cors_matches_origin,
     metadata_from_headers,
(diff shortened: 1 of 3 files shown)


bpandola (changes requested):
Add tests that directly call the new utility method with various policy documents to cover all the branches and I will merge. Thanks.

PR #10234, opened 2026-09-02
SQS: Raise PurgeQueueInProgress on repeated purge within 60 seconds

### LINKED ISSUE #10233: SQS: purge_queue never raises PurgeQueueInProgress within the 60 second window
SQS `purge_queue` never raises `PurgeQueueInProgress`. Real SQS allows only one `PurgeQueue` operation per queue every 60 seconds; a second call inside that window fails with error code `AWS.SimpleQueueService.PurgeQueueInProgress`. Moto happily purges any number of times, so code that handles (or tests for) the purge cooldown behaves differently against moto than against AWS.

The 60 second rule is in the PurgeQueue API reference ("Only one `PurgeQueue` operation is allowed every 60 seconds"), and the botocore service model declares the `PurgeQueueInProgress` error shape on the operation: "Indicates that the specified queue previously received a `PurgeQueue` request within the last 60 seconds (the time it can take to delete the messages in the queue)."
https://docs.aws.amazon.com/AWSSimpleQueueService/latest/APIReference/API_PurgeQueue.html

Repro:

```python
import boto3
from moto import mock_aws

@mock_aws
def repro():
    client = boto3.client("sqs", region_name="us-east-1")
    queue_url = client.create_queue(QueueName="t1")["QueueUrl"]
    client.purge_queue(QueueUrl=queue_url)
    client.purge_queue(QueueUrl=queue_url)  # succeeds; AWS raises PurgeQueueInProgress
    print("second purge succeeded")

repro()
```

Against real AWS the second call raises `ClientError` with code `AWS.SimpleQueueService.PurgeQueueInProgress` and message `Only one PurgeQueue operation on t1 is allowed every 60 seconds.`; against moto it prints `second purge succeeded`.

Versions: moto 5.2.4.dev (master), boto3 1.40.x, Python 3.12.

I have a fix ready (track the last purge time on the queue and raise the documented error inside the window) and will open a PR. One note for the discussion: this makes moto stricter, so a user test suite that purges the same queue repeatedly without mocking time would start seeing the same error real AWS gives; happy to gate it behind a setting instead if you prefer.

haeganm:
Fixes #10233

SQS allows only one `PurgeQueue` operation per queue every 60 seconds; a second call inside the window fails with `AWS.SimpleQueueService.PurgeQueueInProgress`. Moto allowed unlimited consecutive purges. The 60 second rule is documented in the PurgeQueue API reference and the `PurgeQueueInProgress` error shape is declared on the operation in the botocore service model.

Changes:

- `Queue` now records `last_purged_at`, and `purge_queue` raises the new `PurgeQueueInProgress` exception when called again within 60 seconds, following the existing SQS exception conventions (legacy query-compat error code, same shape as `QueueDoesNotExist`).
- Two tests: a plain double-purge test asserting the error code and message (runs in server mode too), and a freezegun test asserting the purge succeeds again once 60 seconds have passed (skipped in server mode like the other time-manipulation tests in the file).

Checked that nothing else in moto or its test suite purges a queue twice within a window, so no existing behavior breaks. Note this makes moto stricter in a way that matches AWS: a user suite that purges the same queue repeatedly in under 60 seconds of real time will now get the same error AWS would give (freezegun sidesteps it). If you would rather gate this behind a setting, happy to rework.

Testing: `pytest tests/test_sqs/` passes (the two `test_sqs_integration.py` Lambda-in-Docker tests fail identically on unmodified master in my environment, no Docker); the new tests fail before the fix and pass after; `ruff check` and `ruff format --check` clean with the pinned ruff 0.14.10.

### DIFF
--- a/moto/sqs/exceptions.py
+++ b/moto/sqs/exceptions.py
@@ -34,6 +34,14 @@ class QueueAlreadyExists(SQSException):
         super().__init__("QueueAlreadyExists", message)
 
 
+class PurgeQueueInProgress(SQSException):
+    def __init__(self, queue_name: str) -> None:
+        super().__init__(
+            "AWS.SimpleQueueService.PurgeQueueInProgress",
+            f"Only one PurgeQueue operation on {queue_name} is allowed every 60 seconds.",
+        )
+
+
 class EmptyBatchRequest(SQSException):
     def __init__(self, action: str = "Send") -> None:
         super().__init__(
--- a/moto/sqs/models.py
+++ b/moto/sqs/models.py
@@ -32,6 +32,7 @@ from .exceptions import (
     MessageAttributesInvalid,
     MissingParameter,
     OverLimit,
+    PurgeQueueInProgress,
     QueueAlreadyExists,
     QueueDoesNotExist,
     ReceiptHandleIsInvalid,
@@ -253,6 +254,7 @@ class Queue(CloudFormationModel):
         self._pending_messages: set[Message] = set()
         self.deleted_messages: set[str] = set()
         self._messages_lock = Condition()
+        self.last_purged_at: float | None = None
 
         now = unix_time()
         self.created_timestamp = now
@@ -1136,6 +1138,10 @@ class SQSBackend(BaseBackend, TaggableResourcesMixin):
 
     def purge_queue(self, queue_name: str) -> None:
         queue = self.get_queue(queue_name)
+        now = unix_time()
+        if queue.last_purged_at is not None and now - queue.last_purged_at < 60:
+            raise PurgeQueueInProgress(queue.name)
+        queue.last_purged_at = now
         queue._messages = []
         queue._pending_messages = set()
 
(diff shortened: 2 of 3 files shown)


Ashishkosana (approved):
Nice, focused fidelity fix — matches the documented SQS rule that only one `PurgeQueue` is allowed per queue every 60 seconds, and the error code/shape lines up with the botocore model (`AWS.SimpleQueueService.PurgeQueueInProgress`) the same way `QueueDoesNotExist` does.

What looks good:
- Recording `last_purged_at` on the operation itself (not on whether messages existed) is the right AWS-shaped behavior.
- Exception subclass follows existing SQS conventions.
- Tests cover both the in-window failure (including the message string) and the after-60s success path, with the server-mode skip consistent with other freezegun tests in this file.
- Calling out the “moto gets stricter / suites that double-purge fast will now fail like AWS” tradeoff in the PR body is exactly the right note for maintainers.

One optional nit (non-blocking): `purge_queue` still clears `_messages` / `_pending_messages` without taking `_messages_lock`, while receive paths wait on that Condition. That predates this PR, but if you touch this area again it might be worth aligning the purge write with the lock so a concurrent receive can’t observe a half-cleared queue. Not required for this change.

CI looks green from here. LGTM.

bpandola (approved):

Issue #8585, opened 2025-02-11
Moto Server allows deploying invalid CFN templates


jabalazs:
# Replication Steps
1. Run `moto_server`
2. Deploy an invalid template:
```
$ aws --endpoint-url http://127.0.0.1:5000 cloudformation create-stack --stack-name test-stack --template-body file://template.json
```

where `template.json` is:

```json
{
  "Resources": {
    "MyTopic": {
      "Type": "AWS::SNS::Topic",
      "Properties": {
        "TopicName": "MySNSTopic",
        "asdfasdf": true,
        "qweqweqwe": true
      },
      "Foo": {"Bar": 0}
    }
  }
}
```

3. The template gets successfully deployed:
```bash
$ aws --endpoint-url http://127.0.0.1:5000 cloudformation get-template --stack-name test-stack
{
    "TemplateBody": {
        "Resources": {
            "MyTopic": {
                "Type": "AWS::SNS::Topic",
                "Properties": {
                    "TopicName": "MySNSTopic",
                    "asdfasdf": true,
                    "qweqweqwe": true
                },
                "Foo": {
                    "Bar": 0
                }
            }
        }
    }
}
```

# Expectation
When trying to deploy an invalid template we should get an error. For example running cfn-lint returns:
```bash
$ cfn-lint template.json
E3002 Additional properties are not allowed ('asdfasdf' was unexpected)
final_template.json:7:9

E3002 Additional properties are not allowed ('qweqweqwe' was unexpected)
final_template.json:8:9

E3001 Additional properties are not allowed ('Foo' was unexpected)
final_template.json:10:7
```

bblommers:
Hi @jabalazs, thanks for raising this.

The first issue should be a simple enhancement. We already use the `cfn_lint` library when the user explicitly calls `validate_template` - it makes sense to also call this validation when creating a stack.

The second, to validate the supplied parameters, is _probably_ also doable. The `botocore` library already lists all parameters for each request, so we should be able to extract that information. This is just a bit more work, as I don't think we have anything in place for that yet.

PR's are welcome, if you want to take a stab at either and/or both!

argerrard:
If no one is currently looking at this enhancement, I'll look into a PR for the first issue.

Mide69:
[...]

PR #10235, opened 2026-09-03
CloudFormation: raise ValidationError when template has no Resources key

Mide69:
## Summary
- `create_stack()` crashed with an unhandled `KeyError: 'Resources'` when a template had no top-level `Resources` key, instead of returning an AWS-style `ValidationError`.
- `ResourceMap.__init__` (moto/cloudformation/parsing.py) now raises a `ValidationError` with the message `Template format error: At least one Resources member must be defined.`, matching real AWS CloudFormation behavior.

Found while implementing an opt-in cfn-lint validation feature for #8585 — this is a pre-existing bug on master, unrelated to that PR.

## Repro (before fix)
```python
import boto3, json
from moto import mock_aws

@mock_aws
def repro():
    client = boto3.client("cloudformation", region_name="us-east-1")
    bad_template = json.dumps({"AWSTemplateFormatVersion": "2010-09-09", "Description": "Stack 1"})
    client.create_stack(StackName="test-stack", TemplateBody=bad_template)

repro()  # raised KeyError: 'Resources'
```

## Test plan
- [x] Added `test_create_stack_fail_missing_resources_key` in `tests/test_cloudformation/test_cloudformation_stack_crud.py`, asserting `create_stack` raises a `ClientError` with code `ValidationError` and the AWS-matching message.
- [x] Verified repro script now raises a proper `ClientError` instead of crashing.
- [x] Ran `test_cloudformation_stack_crud.py` and `test_cloudformation_stack_integration.py` — no new failures introduced (pre-existing unrelated EC2-mock failures confirmed present on `master` via a clean worktree check).
- [x] `ruff check` and `mypy` pass clean on both touched files.

### DIFF
--- a/moto/cloudformation/parsing.py
+++ b/moto/cloudformation/parsing.py
@@ -529,6 +529,10 @@ class ResourceMap(collections_abc.Mapping):  # type: ignore[type-arg]
         cross_stack_resources: dict[str, "Export"] | None,
     ):
         self._template = template
+        if template != {} and "Resources" not in template:
+            raise ValidationError(
+                message="Template format error: At least one Resources member must be defined."
+            )
         self._resource_json_map: dict[str, Any] = (
             template["Resources"] if template != {} else {}
         )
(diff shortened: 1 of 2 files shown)


bblommers (approved):
Makes sense! Thanks for the fix @Mide69

PR #10250, opened 2026-09-08
Payment Cryptography: stop echoing the request token from list calls

### LINKED ISSUE #10249: Payment Cryptography: list calls echo the request NextToken, so pagination never terminates
`list_keys`, `list_aliases` and `list_tags_for_resource` all return the caller's own `NextToken` in the response. A client resuming from a token gets it handed straight back, and botocore stops the loop:

```python
import boto3
from moto import mock_aws

@mock_aws
def repro():
    client = boto3.client("payment-cryptography", region_name="us-east-1")
    client.create_key(
        Exportable=True,
        KeyAttributes={
            "KeyUsage": "TR31_P0_PIN_ENCRYPTION_KEY",
            "KeyClass": "SYMMETRIC",
            "KeyAlgorithm": "TDES_3KEY",
            "KeyModesOfUse": {"Encrypt": True, "Decrypt": True},
        },
    )
    list(client.get_paginator("list_keys").paginate(
        PaginationConfig={"StartingToken": "resume-token"}))

repro()
# PaginationError: Error during pagination: The same next token was
# received twice: {'NextToken': 'resume-token'}
```

In `moto/paymentcryptography/models.py` the parameter comes back out untouched:

```python
return [key.to_dict() for key in keys], next_token
```

Every key fits in one page here, so `None` is the accurate answer. This is the same thing #10236 reported in kafka, in a service that still has it.

chiruu12:
Fixes #10249

`list_keys`, `list_aliases` and `list_tags_for_resource` each returned the `next_token` they were given. Resuming a listing from a token handed it back unchanged, which botocore reports as `The same next token was received twice`.

All three now return `None`. Every result fits in one page, so that is the accurate answer, and it matches what `list_clusters` in kafka was changed to in #10237.

Six tests. Five fail on the parent commit, two of them with the `PaginationError` above. The sixth pins that `list_keys` still filters on `KeyState` and passes either way.

`tests/test_paymentcryptography/test_paymentcryptography.py`: 39 passed, with `ruff`, `ruff format --check` and `mypy` clean. `test_server.py` does not collect without `flask` installed; that is the same on `master`.

### DIFF
--- a/moto/paymentcryptography/models.py
+++ b/moto/paymentcryptography/models.py
@@ -198,13 +198,17 @@ class PaymentCryptographyControlPlaneBackend(BaseBackend):
         keys = list(self.keys.values())
         if key_state:
             keys = [key for key in keys if key.key_state == key_state]
-        return [key.to_dict() for key in keys], next_token
+        # None, not the caller's own token. Every key comes back in one
+        # page, and handing the incoming token back made botocore raise
+        # "The same next token was received twice" on a resumed listing.
+        return [key.to_dict() for key in keys], None
 
     def list_tags_for_resource(
         self, resource_arn: str, next_token: Optional[str], max_results: Optional[int]
     ) -> tuple[list[dict[str, str]], Optional[str]]:
         tags = self.tagger.list_tags_for_resource(resource_arn)["Tags"]
-        return tags, next_token
+        # See list_keys: the incoming token must not come back out.
+        return tags, None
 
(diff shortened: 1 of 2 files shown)


bpandola (commented):
The tests included in this PR (and in #10237) will fail (invalid token error) if proper pagination code is ever added.  The appropriate fix here is to implement proper pagination or simply remove the offending code altogether.  I've chosen the latter for now (see #10260).

PR #10262, opened 2026-09-17
SWF: validate required workflow execution parameters

### LINKED ISSUE #10263: SWF: missing workflow execution members raise unhandled exceptions
`DescribeWorkflowExecution` and `GetWorkflowExecutionHistory` index required request fields before handling missing values. Omitting `execution.runId` raises an unhandled `KeyError`, which can surface as an internal server error. Clients that bypass boto3's local parameter validation can reach this path.

Reproduced against `d22033db2b0762e15cfbc4947ea53809dfd880e7`:

```python
import boto3
from moto import mock_aws

with mock_aws():
    client = boto3.client(
        "swf",
        region_name="us-east-1",
        config=Config(parameter_validation=False),
    )
    client.describe_workflow_execution(
        domain="test-domain",
        execution={"workflowId": "workflow-id"},
    )
```

Moto raises `KeyError: 'runId'`. Replacing the method with `get_workflow_execution_history` produces the same failure. Omitting `workflowId`, supplying an empty execution object, or omitting required top-level fields exposes similar gaps.

Verified both operations against live AWS SWF in `us-east-1` on September 17, 2026, using signed JSON requests. A missing or null `execution.runId` returns HTTP 400 with:

```json
{
  "__type": "com.amazon.coral.validate#ValidationException",
  "message": "1 validation error detected: Value null at 'execution.runId' failed to satisfy constraint: Member must not be null"
}
```

AWS also returns field-specific `ValidationException` messages for missing/null `domain`, `execution`, and `execution.workflowId`, and combines the messages when multiple required fields are absent. Incorrect JSON types return `SerializationException`; missing members should be distinguished from those type errors.

A fix and HTTP regression coverage are proposed in #10262.

AndreKurait:
Closes #10263.

`DescribeWorkflowExecution` and `GetWorkflowExecutionHistory` index required fields before validating them. A request missing `execution.runId` raises an unhandled `KeyError` when the client bypasses boto3's parameter validation.

Validate `domain`, `execution`, and its two IDs before looking up the workflow. Use the existing `SWFValidationException` for missing/null members, including combined messages when several fields are absent. Return checked strings from `_check_string` so the shared helper needs no casts.

Verified both operations against live AWS SWF in `us-east-1` on September 17, 2026. All 20 missing/null cases match AWS's HTTP 400 status, full `ValidationException` type, and exact message. Another 12 HTTP cases cover invalid JSON types using the existing Flask test-client pattern.

Validation: all 214 SWF tests pass, along with mypy for SWF and repository-pinned Ruff lint and formatting checks. On the pre-PR base, 29 of the 32 HTTP cases fail.


AndreKurait:
@bpandola i fixed the linting issue by merging from upstream. That was fix #10271 which is why my build was failing.

This time i've ran the actions on my own fork
https://github.com/AndreKurait/moto/actions/runs/36017184013

PR #10264, opened 2026-09-19
EC2: implement AMI Disable/EnableImage

heldersepu:
## WHAT
Implementing `disable_image` and `enable_image`  for EC2

## WHY
These two looked like simple methods that we needed and implementation was missing 

## TESTS
Unit test added, see: `test_ami_enable_disable`


bpandola (approved):

heldersepu:
thanks @bpandola for the assist with the format and guard clauses

heldersepu:
@bpandola When you can please take a look at this other PR: https://github.com/getmoto/moto/pull/10270
I'm not sure why the tests are failing there, locally seems to be fine

PR #10270, opened 2026-09-20
DMS: implement modify_replication_instance

heldersepu:
## WHAT
Implementing `modify_replication_instance` for DMS

## WHY
This is a methods that I need and implementation is missing 

## TESTS
Unit test added, see: `test_modify_replication_instance`

## REFERENCES
- https://docs.aws.amazon.com/dms/latest/APIReference/API_ModifyReplicationInstance.html#DMS-ModifyReplicationInstance-request-AllowMajorVersionUpgrade
- https://docs.aws.amazon.com/boto3/latest/reference/services/dms/client/modify_replication_instance.html
- https://github.com/hashicorp/terraform-provider-aws/issues/14438#issuecomment-670580415


heldersepu:
Linting brought up an interesting case:
`Argument 1 to "parse" has incompatible type "str | None"; expected "str"  [arg-type]`

According to the AWS documentation:
https://docs.aws.amazon.com/boto3/latest/reference/services/dms/client/create_replication_instance.html
>EngineVersion (string) –
>
>The engine version number of the replication instance.
>
>If an engine version number is not specified when a replication instance is created, the default is the latest engine version available.

If I understood correctly that means the engine version will not be `None`  ... but what is the latest?
I think a full implementation will require the method `describe_engine_versions` that would be the way to get latest

heldersepu:
Strange that the TestNDeploy fails I tested the same locally and it pass:

`/pytest -sv -rs --cov=moto --cov-report xml ./tests/ --ignore tests/test_acm --ignore tests/test_amp --ignore tests/test_awslambda --ignore tests/test_batch --ignore tests/test_dynamodb --ignore tests/test_ec2 --ignore tests/test_s3/ --ignore tests/test_sqs`

image

bpandola:
> If I understood correctly that means the engine version will not be `None` ... but what is the latest? I think a full implementation will require the method `describe_engine_versions` that would be the way to get latest

The easiest thing would be to just provide a default value in the `ReplicationInstance` constructor for `engine_version` if it's `None`.  If you want to do that and/or implement `DescribeEngineVersions` in a separate PR, feel free!

bpandola:
> Strange that the TestNDeploy fails I tested the same locally and it pass:
> 
> `/pytest -sv -rs --cov=moto --cov-report xml ./tests/ --ignore tests/test_acm --ignore tests/test_amp --ignore tests/test_awslambda --ignore tests/test_batch --ignore tests/test_dynamodb --ignore tests/test_ec2 --ignore tests/test_s3/ --ignore tests/test_sqs`
> 

The `packaging` library is a dev requirement but isn't a Moto requirement, which is why the tests passed locally but failed in server mode (where the library is not included in the container).

I've pushed a commit to make this work with Moto's own `LooseVersion` implementation.

bpandola (approved):

heldersepu:
> I've pushed a commit to make this work with Moto's own `LooseVersion` implementation.

Thanks again, I keep this in mind in the future

PR #10273, opened 2026-09-21
CodeDeploy: apply includeOnlyStatuses without a deployment group

chiruu12:
`list_deployments` accepts `includeOnlyStatuses` and ignores it unless `deploymentGroupName` is also passed, so a caller filtering on status alone gets every deployment back.

The status check sits inside the `if deployment_group_name:` block in `matches_filters`, so it only runs on the branch where a group was named.

Against current master, with three deployments that are all in `Created`:

    list_deployments(includeOnlyStatuses=["Failed"])                      -> 3 deployments
    list_deployments(applicationName=..., deploymentGroupName=...,
                     includeOnlyStatuses=["Failed"])                      -> 0 deployments

The first call should return nothing. Real CodeDeploy applies `includeOnlyStatuses` on its own, and the API does not require an application or a group alongside it.

Moving the check out one level fixes it. The new test fails on master with `assert ['50b4b9d9-...'] == []` and passes with the change, and it covers both directions: a status nothing matches returns empty, and the status the deployments actually have returns them.

`externalId` and `createTimeRange` on the same call are still unused. Those are separate filters that were never wired up rather than a filter that breaks depending on another argument, so I have left them out of this change.

tests/test_codedeploy passes, 27 tests. ruff and mypy are clean on `moto/codedeploy`. There is a pre-existing PIE808 in `tests/test_codedeploy/test_codedeploy.py:410` that I have not touched.


bpandola (changes requested):
No need to change the overall logic here.  The `return True` is the default and should stay.  You need only unindent the `if` in question to get it out from under the (unrelated) `deployment_group_name` block.

bpandola (approved):

